PR #2189, opened 2015-03-16
React to Authenticate(schemes) being removed

HaoK:


dougbu:
:shipit: thx for getting this in @HaoK

PR #2194, opened 2015-03-17
Updates to the extensibility for validator providers

rynowak:
This change removes reflection from validator providers, and instead
relies on cached metadata in in the modelmetadata.

In general this means that our MVPs don't need to cache anything, they
just look at the metadata and create what they need.

In the case of data-annotations, we update the model details provider to
add validation attributes to the modelmetadata. This would allow someone
to replace the DataAnnotationsValidatorProvider, but still use the
metadata in these attributes.

The change to the IModelValidatorProvider api (to use a context) is
intended to minimize allocations. Currently each validator provider needs
to return a list so you end up with N+1 lists (N validators + a final list
to compine them all). This change will let us just create the final list
(and a small context object). This is a very very high traffic API so it
seemed worth doing.

There's also some general massaging of namespaces and file locations.


test/Microsoft.AspNet.Mvc.FunctionalTests/ModelMetadataAttributeTest.cs
```diff
@@ -98,6 +98,8 @@ public class ModelMetadataAttributeTest
 
             var content = new StringContent(input, Encoding.UTF8, "application/json");
 
+            System.Diagnostics.Debugger.Launch();
+
             var url = "http://localhost/ModelMetadataTypeValidation/ValidateProductViewModelIncludingMetadata";
```
rynowak: Will remove :laughing:

test/WebSites/ModelBindingWebSite/Controllers/VehicleController.cs
```diff
@@ -37,7 +37,7 @@ public class VehicleController : Controller
         [HttpPut("/api/vehicles/{id}")]
         [Produces("application/json")]
         public object UpdateVehicleApi(
-            [Range(1, 500)] int id,
```
rynowak: This wasn't really ever intended to work, but we did it.... There's a few of these that need to be cleaned up.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DefaultModelValidatorProvider.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
+{
+    public class DefaultModelValidatorProvider : IModelValidatorProvider
```
dougbu: how many validators do we have in the system?  do we really need more than `IValidatorProvider` (or `IValidationProvider`) and `DefaultModelValidatorProvider` (do the or yourself)?

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DefaultModelValidatorProvider.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        public void GetValidators(ModelValidatorProviderContext context)
```
dougbu: similarly `ValidatorProviderContext` or ...

test/WebSites/ModelBindingWebSite/Controllers/VehicleController.cs
dougbu: `[Range]` is not our attribute and, like `[Validation]` and many a subclass, is completely able to target parameters.  aren't you introducing a regression from MVC 5?

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataMemberModelValidatorProvider.cs
```diff
@@ -6,49 +6,40 @@
...
-    public class DataMemberModelValidatorProvider : AssociatedValidatorProvider
+    public class DataMemberModelValidatorProvider : IModelValidatorProvider
     {
-        protected override IEnumerable<IModelValidator> GetValidators(ModelMetadata metadata,
-                                                                      IEnumerable<object> attributes)
+        public void GetValidators(ModelValidatorProviderContext context)
         {
             // Types cannot be required; only properties can
-            if (metadata.ContainerType == null || string.IsNullOrEmpty(metadata.PropertyName))
+            if (context.ModelMetadata.MetadataKind != ModelMetadataKind.Property)
```
dougbu: why make this more specific than it was?

dougbu:
:watch: for less ugly names and new restrictions in the system.  (fyi expect non-design comments will be about `[NotNull]` and some odd spacing)

test/WebSites/ModelBindingWebSite/Controllers/VehicleController.cs
rynowak: This started working by coincidence when we added `GetMetadataForParameter`. We never really talked about this or intended it :laughing: @harshgMSFT is currently hard at work removing `GetMetadataForParameter`

MVC never had support for this. 

Yes, `DataAnnotations` attributes can be placed on parameters, but MVC and WebAPI have never supported this - they are only supported on model property and types.

Try out the following in MVC 4
```
public class HomeController : Controller
    {
        //
        // GET: /Home/

        public ActionResult Index([Range(0, 5)] int id)
        {
            return View();
        }

        public ActionResult WithModel(Model model)
        {
            return View();
        }

        public class Model
        {
            [Range(0, 5)]
            public int Id { get; set; }
        }
    }
```

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DefaultModelValidatorProvider.cs
rynowak: We have 10-15 validator implementations, and 3 providers.

1). Default provider - this is for extensibility - if you add an attribute that **is-a** `IModelValidator` this handles it for you - this is consistent with some of our designs for other metadata+provider concepts (filters, action constraints).
2). DataAnnotations provider, handles `ValidationAttribute` and `IValidatableObject`
3). DataMember provider, handles `[DataMember(IsRequired = true)]`

Would like to replace 3 with something better, but for now this is the best way to adopt the current validation system to the new provider extensibility.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataMemberModelValidatorProvider.cs
rynowak: What do you mean more specific? `metadata.ContainerType == null` and `context.ModelMetadata.MetadataKind != ModelMetadataKind.Property` are equivalent.

rynowak:
**If you're game I'd like to propose a moratorium on **new** names for this PR.** We have several more concept that we should take a look at that aren't changing in this PR. I think your suggestions are where we're going to end up for these classes, but I'd rather us look at the bigger picture before making that conclusion.

The stuff that's here is consistent with the current names.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataMemberModelValidatorProvider.cs
dougbu: never mind; thought this was where `[Validation]` on parameters stopped working.

dougbu:
> I'd like to propose a moratorium on new names

it's going to be murder but I'll try

rynowak:
WOrking on removing the design tag.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/IModelValidatorProvider.cs
```diff
@@ -1,12 +1,12 @@
 // Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System.Collections.Generic;
+using Microsoft.AspNet.Mvc.ModelBinding.Validation;
 
 namespace Microsoft.AspNet.Mvc.ModelBinding
```
pranavkm: Shouldn't this go in to the `Validation` namespace?

rynowak: Yeah probably :laughing:

rynowak:
@dougbu updated.

The namespace changes had a cascading effect, but I'd rather us just get this over with than do it halfsies.

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DataMemberBindingMetadataProvider.cs
```diff
@@ -0,0 +1,45 @@
...
+
+            // isDataContract == true iff the container type has at least one DataContractAttribute
+            var containerType = context.Key.ContainerType.GetTypeInfo();
+            var isDataContract = containerType.GetCustomAttribute<DataContractAttribute>() != null;
+            if (isDataContract)
+            {
+                // We don't need to add a validator, just to set IsRequired = true. The validation
+                // system will do the right thing.
+                context.BindingMetadata.IsRequired = true;
+            }
```
rynowak: This thing used to be a validator provider and it would return a no-op 'required' validator. Just setting `IsRequired = true` does the same thing. So I decided to do that instead.

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/ModelAttributes.cs
```diff
@@ -26,7 +26,7 @@ public static IEnumerable<object> GetAttributesForParameter(ParameterInfo parame
         {
             // Return the parameter attributes first.
             var parameterAttributes = parameter.GetCustomAttributes();
-            var typeAttributes = parameter.ParameterType.GetTypeInfo().GetCustomAttributes();
+            var typeAttributes = ModelAttributes.GetAttributesForType(parameter.ParameterType);
```
rynowak: Oops.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsModelValidatorProvider.cs
```diff
@@ -42,35 +41,24 @@ public class DataAnnotationsModelValidatorProvider : AssociatedValidatorProvider
             get { return _attributeFactories; }
         }
 
-        private static bool AddImplicitRequiredAttributeForValueTypes
+        public void GetValidators(ModelValidatorProviderContext context)
         {
-            get { return _addImplicitRequiredAttributeForValueTypes; }
-            set { _addImplicitRequiredAttributeForValueTypes = value; }
-        }
```
rynowak: `AddImplicitRequiredAttributeForValueTypes` was unused.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DefaultObjectValidator.cs
```diff
@@ -49,7 +49,7 @@ public void Validate([NotNull] ModelValidationContext modelValidationContext)
             string modelKey,
             ModelExplorer modelExplorer, 
             ValidationContext validationContext, 
-            IEnumerable<IModelValidator> validators)
+            IList<IModelValidator> validators)
```
rynowak: New api returns a list - this lets us simplify the code a bit.

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DataAnnotationsMetadataDetailsProvider.cs
```diff
@@ -167,5 +167,13 @@ public void GetDisplayMetadata([NotNull] DisplayMetadataProviderContext context)
                 displayMetadata.TemplateHint = "HiddenInput";
             }
         }
+
+        public void GetValidationMetadata([NotNull] ValidationMetadataProviderContext context)
```
harshgMSFT: docs

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DataMemberBindingMetadataProvider.cs
harshgMSFT: :+1:

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/ValidationMetadata.cs
```diff
@@ -8,5 +10,6 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
     /// </summary>
     public class ValidationMetadata
     {
+        public IList<object> ValiatorMetadata { get; } = new List<object>();
```
harshgMSFT: docs

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadata.cs
```diff
@@ -356,5 +356,14 @@ public override string TemplateHint
                 return DisplayMetadata.TemplateHint;
             }
         }
+
+        /// <inheritdoc />
+        public override IReadOnlyList<object> ValidatorMetadata
+        {
+            get
+            {
+                return new ReadOnlyCollection<object>(ValidationMetadata.ValiatorMetadata);
```
harshgMSFT: creating a new collection everytime for something that is already cached? should we consider caching this as well?

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/AssociatedValidatorProvider.cs
```diff
@@ -1,54 +0,0 @@
-// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-using System;
```
harshgMSFT: :heart_eyes:

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DefaultObjectValidator.cs
```diff
@@ -84,7 +84,10 @@ public void Validate([NotNull] ModelValidationContext modelValidationContext)
                 // the same for all the elements of the array, we do not do GetValidators for each element,
                 // instead we just pass them over. See ValidateElements function.
                 var validatorProvider = validationContext.ModelValidationContext.ValidatorProvider;
-                validators = validatorProvider.GetValidators(modelExplorer.Metadata);
+                var validatorProviderContext = new ModelValidatorProviderContext(modelExplorer.Metadata);
```
harshgMSFT: would it be cleaner to add a new method for these 3 lines?

test/Microsoft.AspNet.Mvc.Core.Test/OptionDescriptors/DefaultValidationProviderProviderTest.cs
```diff
@@ -52,9 +53,8 @@ public TestModelValidationProvider(ITestService service)
 
             public ITestService Service { get; private set; }
 
-            public IEnumerable<IModelValidator> GetValidators(ModelMetadata metadata)
+            public void GetValidators(ModelValidatorProviderContext context)
             {
-                throw new NotImplementedException();
```
harshgMSFT: I think this would be indirectly ensuring that this never gets called. is this interesting? should we consider keeping this or replacing with an alternate mechanism/

test/Microsoft.AspNet.Mvc.TestCommon/TestModelValidatorProvider.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
+{
+    //
```
harshgMSFT: remove or add comments.

src/Microsoft.AspNet.Mvc/MvcOptionsSetup.cs
```diff
@@ -62,11 +63,13 @@ public static void ConfigureMvc(MvcOptions options)
 
             // Set up metadata providers
             options.ModelMetadataDetailsProviders.Add(new DefaultBindingMetadataProvider());
+            options.ModelMetadataDetailsProviders.Add(new DefaultValidationMetadataProvider());
             options.ModelMetadataDetailsProviders.Add(new DataAnnotationsMetadataDetailsProvider());
```
harshgMSFT: see this type is still called `DataAnnotationsMetadataDetailsProvider` ...and not `DataAnnotationsMetadataProvider`

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Validation/DefaultModelValidatorProviderTest.cs
```diff
@@ -0,0 +1,264 @@
...
+            var validatorProvider = TestModelValidatorProvider.CreateDefaultProvider();
+
+            var metadata = metadataProvider.GetMetadataForType(typeof(ValidatableObject));
+            var context = new ModelValidatorProviderContext(metadata);
+
+            // Act
+            validatorProvider.GetValidators(context);
+
+            // Assert
+            var validators = context.Validators;
```
harshgMSFT: nit: merge 28 and 30

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Validation/DataAnnotationsModelValidatorProviderTest.cs
```diff
@@ -172,41 +178,15 @@ public void IValidatableObjectGetsAValidator()
...
-
-            // Assert
-            Assert.Single(validators);
-        }
-#endif
-
-        // Integration with metadata system
-
-        [Fact]
-        public void DoesNotReadPropertyValue()
```
harshgMSFT: Was this test not interesting? Looks like this ensured that we do not inspect the value while fetching the validator?

harshgMSFT:
:watch: mainly for the comment on DefaultModelMetadata.cs

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DataAnnotationsMetadataDetailsProvider.cs
rynowak: fixed

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/ValidationMetadata.cs
rynowak: Fixed

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadata.cs
rynowak: Yeah I meant to clean this up but I forgot - wil fix. BTW `ReadOnlyCollection` is a wrapper, it doesn't do any copying.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DefaultObjectValidator.cs
rynowak: No, the pattern is slightly more verbose, but it should stay as-is. This pattern allows us to build in more kinds of optimizations in the future if we need to (reusing the context for one).

test/Microsoft.AspNet.Mvc.TestCommon/TestModelValidatorProvider.cs
rynowak: removed

src/Microsoft.AspNet.Mvc/MvcOptionsSetup.cs
rynowak: `DataAnnotationsMetadataDetailsProvider` was originally the pattern here, but I think I like your suggestion better. Will change it.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Validation/DefaultModelValidatorProviderTest.cs
rynowak: don't care to

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Validation/DataAnnotationsModelValidatorProviderTest.cs
rynowak: Correct. It's now **impossible** for this to happen. Nothing in this code path has access to the model. It's actually been this way since the first MM change I made.

rynowak:
@harshgMSFT updated

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadata.cs
harshgMSFT: Is this fixed?

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/ValidationMetadata.cs
harshgMSFT: not yet :)

src/Microsoft.AspNet.Mvc/MvcOptionsSetup.cs
harshgMSFT: Is this fixed?

harshgMSFT:
:shipit:

PR #2199, opened 2015-03-18
React to hosting changes

HaoK:
Putting back a precomp fallback SP for now since Hosting.Create is gone, UseServices is dead, so switched over to ConfigureServices everywhere

cc @davidfowl @rynowak @dougbu

### DIFF
--- a/samples/MvcSample.Web/Startup.cs
+++ b/samples/MvcSample.Web/Startup.cs
@@ -1,8 +1,8 @@
 ﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-#if DNX451
 using System;
+#if DNX451
 using System.IO;
 using Autofac;
 #endif
@@ -18,6 +18,7 @@ using Microsoft.Framework.DependencyInjection.Autofac;
 using Microsoft.Framework.Runtime;
 using Microsoft.Framework.Runtime.Infrastructure;
 #endif
+using System.ComponentModel;
 using MvcSample.Web.Filters;
 using MvcSample.Web.Services;
 
@@ -64,7 +65,7 @@ namespace MvcSample.Web
                 });
 
                 // Create the autofac container
-                ContainerBuilder builder = new ContainerBuilder();
+                var builder = new ContainerBuilder();
 
                 // Create the container and use the default application services as a fallback
                 AutofacRegistration.Populate(
@@ -73,7 +74,7 @@ namespace MvcSample.Web
 
                 builder.RegisterModule<MonitoringModule>();
 
-                IContainer container = builder.Build();
+                var container = builder.Build();
 
                 return container.Resolve<IServiceProvider>();
             }
--- a/samples/MvcSample.Web/Startup.cs
+++ b/samples/MvcSample.Web/Startup.cs
@@ -18,7 +18,6 @@ using Microsoft.Framework.DependencyInjection.Autofac;
 using Microsoft.Framework.Runtime;
 using Microsoft.Framework.Runtime.Infrastructure;
 #endif
-using System.ComponentModel;
 using MvcSample.Web.Filters;
 using MvcSample.Web.Services;
 
@@ -26,6 +25,10 @@ namespace MvcSample.Web
 {
     public class Startup
     {
+#if DNX451
+        private bool _autoFac;
+#endif
+
         public IServiceProvider ConfigureServices(IServiceCollection services)
         {
             services.AddCaching();
@@ -44,7 +47,9 @@ namespace MvcSample.Web
             });
 
 #if DNX451
-            // REVIEW: This is hella hacky
+            // Fully-qualify configuration path to avoid issues in functional tests. Just "config.json" would be fine
+            // but Configuration uses CallContextServiceLocator.Locator.ServiceProvider to get IApplicationEnvironment.
+            // Functional tests update that service but not in the static provider.
             var applicationEnvironment = services.BuildServiceProvider().GetRequiredService<IApplicationEnvironment>();
             var configurationPath = Path.Combine(applicationEnvironment.ApplicationBasePath, "config.json");
 
@@ -56,7 +61,7 @@ namespace MvcSample.Web
             if (configuration.TryGet("DependencyInjection", out diSystem) &&
                 diSystem.Equals("AutoFac", StringComparison.OrdinalIgnoreCase))
             {
-
+                _autoFac = true;
                 services.ConfigureRazorViewEngine(options =>
                 {
                     var expander = new LanguageViewLocationExpander(
@@ -91,20 +96,7 @@ namespace MvcSample.Web
             app.UseFileServer();
 
 #if DNX451
-            // Fully-qualify configuration path to avoid issues in functional tests. Just "config.json" would be fine
-            // but Configuration uses CallContextServiceLocator.Locator.ServiceProvider to get IApplicationEnvironment.
-            // Functional tests update that service but not in the static provider.
-            var applicationEnvironment = app.ApplicationServices.GetRequiredService<IApplicationEnvironment>();
-            var configurationPath = Path.Combine(applicationEnvironment.ApplicationBasePath, "config.json");
-
-            // Set up configuration sources.
-            var configuration = new Configuration()
-                .AddJsonFile(configurationPath)
-                .AddEnvironmentVariables();
-
-            string diSystem;
-            if (configuration.TryGet("DependencyInjection", out diSystem) &&
-                diSystem.Equals("AutoFac", StringComparison.OrdinalIgnoreCase))
+            if (_autoFac)
             {
                 app.UseMiddleware<MonitoringMiddlware>();
             }
--- a/samples/MvcSample.Web/Startup.cs
+++ b/samples/MvcSample.Web/Startup.cs
@@ -16,10 +16,10 @@ using Microsoft.Framework.DependencyInjection;
 #if DNX451
 using Microsoft.Framework.DependencyInjection.Autofac;
 using Microsoft.Framework.Runtime;
+using Microsoft.Framework.Runtime.Infrastructure;
 #endif
 using MvcSample.Web.Filters;
 using MvcSample.Web.Services;
-
 
 namespace MvcSample.Web
 {
@@ -43,9 +43,9 @@ namespace MvcSample.Web
             });
 
 #if DNX451
-            //var applicationEnvironment = app.ApplicationServices.GetRequiredService<IApplicationEnvironment>();
-            //var configurationPath = Path.Combine(applicationEnvironment.ApplicationBasePath, "config.json");
-            var configurationPath = "config.json";
+            // REVIEW: This is hella hacky
+            var applicationEnvironment = services.BuildServiceProvider().GetRequiredService<IApplicationEnvironment>();
+            var configurationPath = Path.Combine(applicationEnvironment.ApplicationBasePath, "config.json");
 
             // Set up configuration sources.
             var configuration = new Configuration()
--- a/samples/MvcSample.Web/Startup.cs
+++ b/samples/MvcSample.Web/Startup.cs
@@ -25,6 +25,65 @@ namespace MvcSample.Web
 {
     public class Startup
     {
+        public IServiceProvider ConfigureServices(IServiceCollection services)
+        {
+            services.AddCaching();
+            services.AddSession();
+
+            services.AddMvc();
+            services.AddSingleton<PassThroughAttribute>();
+            services.AddSingleton<UserNameService>();
+            services.AddTransient<ITestService, TestService>();
+
+            services.ConfigureMvc(options =>
+            {
+                options.Filters.Add(typeof(PassThroughAttribute), order: 17);
+                options.AddXmlDataContractSerializerFormatter();
+                options.Filters.Add(new FormatFilterAttribute());
+            });
+
+#if DNX451
+            //var applicationEnvironment = app.ApplicationServices.GetRequiredService<IApplicationEnvironment>();
+            //var configurationPath = Path.Combine(applicationEnvironment.ApplicationBasePath, "config.json");
+            var configurationPath = "config.json";
+
+            // Set up configuration sources.
+            var configuration = new Configuration()
+                .AddJsonFile(configurationPath)
+                .AddEnvironmentVariables();
+            string diSystem;
+            if (configuration.TryGet("DependencyInjection", out diSystem) &&
+                diSystem.Equals("AutoFac", StringComparison.OrdinalIgnoreCase))
+            {
+
+                services.ConfigureRazorViewEngine(options =>
+                {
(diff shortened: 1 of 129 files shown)


src/Microsoft.AspNet.Mvc/RazorPreCompileModule.cs
```diff
@@ -68,10 +58,12 @@ public virtual void BeforeCompile(IBeforeCompileContext context)
...
                 context.ProjectContext.ProjectDirectory);
-            serviceCollection.AddInstance<IApplicationEnvironment>(precompilationApplicationEnvironment);
 
-            var serviceProvider = serviceCollection.BuildServiceProvider();
-            var viewCompiler = new RazorPreCompiler(serviceProvider, context, _memoryCache, compilationSettings)
+            var replacedServices = new ServiceCollection();
+            replacedServices.AddMvc();
+            replacedServices.AddInstance<IApplicationEnvironment>(precompilationApplicationEnvironment);
+
+            var viewCompiler = new RazorPreCompiler(new WrappingServiceProvider(_appServices, replacedServices), context, _memoryCache, compilationSettings)
```
dougbu: wrap long line

samples/MvcSample.Web/Startup.cs
```diff
@@ -25,6 +25,65 @@ namespace MvcSample.Web
...
+            services.ConfigureMvc(options =>
+            {
+                options.Filters.Add(typeof(PassThroughAttribute), order: 17);
+                options.AddXmlDataContractSerializerFormatter();
+                options.Filters.Add(new FormatFilterAttribute());
+            });
+
+#if DNX451
+            //var applicationEnvironment = app.ApplicationServices.GetRequiredService<IApplicationEnvironment>();
+            //var configurationPath = Path.Combine(applicationEnvironment.ApplicationBasePath, "config.json");
```
dougbu: - this code was necessary to get the functional tests working reliably when they invoked this web site.  why is it no longer necessary here?
- if fully-qualifying the configuration path isn't needed here, why did you leave it in `Configure(IApplicationBuilder)`?
- do not check in commented-out code

src/Microsoft.AspNet.Mvc/RazorPreCompileModule.cs
dougbu: or use a `var` for the `WrappingServiceProvider`

src/Microsoft.AspNet.Mvc/RazorPreCompileModule.cs
```diff
@@ -84,6 +76,27 @@ public void AfterCompile(IAfterCompileContext context)
         {
         }
 
+        // REVIEW: Welcome back to life delegating SP!!!
```
dougbu: ick!  we just got rid of this hack.  why doesn't Hosting support the approaches MVC requires?

src/Microsoft.AspNet.Mvc/RazorPreCompileModule.cs
```diff
@@ -84,6 +76,27 @@ public void AfterCompile(IAfterCompileContext context)
...
+        private class WrappingServiceProvider : IServiceProvider
+        {
+            private readonly IServiceProvider _inner;
+            private readonly IServiceProvider _outer;
```
dougbu: I have no idea what these names mean without reading the code.  use something like `_override` and `_fallback`

test/WebSites/ActionConstraintsWebSite/Startup.cs
```diff
@@ -9,18 +9,19 @@ namespace ActionConstraintsWebSite
...
-                {
-                    options.AddXmlDataContractSerializerFormatter();
-                });
+                options.AddXmlDataContractSerializerFormatter();
             });
+        }
+
+        public void Configure(IApplicationBuilder app)
+        {
+            var configuration = app.GetTestConfiguration();
```
dougbu: nit: bet you don't need the `configuration` variable anywhere

test/WebSites/AutofacWebSite/Startup.cs
```diff
@@ -11,23 +11,24 @@ namespace AutofacWebSite
...
+            services.AddMvc();
+            services.AddTransient<HelloWorldBuilder>();
 
-            app.UseServices(services =>
-            {
-                services.AddMvc();
-                services.AddTransient<HelloWorldBuilder>();
+            var builder = new ContainerBuilder();
+            AutofacRegistration.Populate(builder,
+                                         services);
```
dougbu: nit: please unwrap this line; very short

test/WebSites/ControllersFromServicesWebSite/Startup.cs
```diff
@@ -16,34 +16,37 @@ namespace ControllersFromServicesWebSite
...
-                        .WithControllersAsServices(
-                         new[]
-                         {
+            services.AddMvc()
+                    .WithControllersAsServices(
+                     new[]
+                     {
                             typeof(TimeScheduleController).GetTypeInfo().Assembly
-                         });
+                     });
```
dougbu: fyi no need to line up the periods; just hit enter after `services` or `services.AddMvc()` and VS should do the Right Thing:tm:

test/WebSites/RazorCompilerCacheWebSite/Startup.cs
```diff
@@ -9,17 +9,19 @@ namespace RazorCompilerCacheWebSite
 {
     public class Startup
     {
+        // Set up application services
+        public void ConfigureServices(IServiceCollection services)
+        {
+            // Add MVC services to the services container
```
dougbu: nit: this comment is getting a little old by this point.  says nothing `services.AddMvc()` doesn't

test/WebSites/RazorCompilerCacheWebSite/Startup.cs
```diff
@@ -9,17 +9,19 @@ namespace RazorCompilerCacheWebSite
 {
     public class Startup
     {
+        // Set up application services
```
dougbu: isn't this actually setting up _request_ services?

test/WebSites/RoutingWebSite/Startup.cs
```diff
@@ -8,18 +8,19 @@ namespace RoutingWebSite
...
+        }
 
-                services.AddScoped<TestResponseGenerator>();
-            });
+        public void Configure(IApplicationBuilder app)
+        {
+            var configuration = app.GetTestConfiguration();
 
-            app.UseErrorReporter();
+           app.UseErrorReporter();
```
dougbu: nit: accidental change here

src/Microsoft.AspNet.Mvc/RazorPreCompileModule.cs
dougbu: put another way: we were using `HostingServices.Create()` and it was less evil / ugly than this

src/Microsoft.AspNet.Mvc/RazorPreCompileModule.cs
```diff
@@ -84,6 +76,27 @@ public void AfterCompile(IAfterCompileContext context)
...
+                _outer = replacedServices.BuildServiceProvider();
+            }
+
+            public object GetService(Type serviceType)
+            {
+                return _outer.GetService(serviceType) ?? _inner.GetService(serviceType);
+            }
+        }
+
+
```
dougbu: nit: remove extra blank line

dougbu:
:watch:

test/WebSites/ActionConstraintsWebSite/Startup.cs
HaoK: I don't but it does magic I thought

test/WebSites/RazorCompilerCacheWebSite/Startup.cs
HaoK: no distinction anymore with hosting changes

test/WebSites/ActionConstraintsWebSite/Startup.cs
dougbu: I meant the line could just be `app.GetTestConfiguration();` (the method does magic; the variable does nothing)

HaoK: I don't disagree but that seems like something that should be done as a full pass over the test code, personally I think you should rename the method

dougbu: triage rejected the rename bug you had me file.  but you're moving an unused variable around and I wasn't suggesting additional changes to lines you weren't already touching.

HaoK: My point is that this is an issue all over the test code base, I don't see the value in fixing one place, all that's going to do is make any future FindReplace All operations harder.

If this is something you think is worth fixing, I think it should be done as one PR doing a global replace all removing the var from all tests at once.

HaoK:
Updated with changes, 7 test failures remaining
- Mostly TagHelpers, and RazorCompilerCacheTest.CompilerCache_IsNotInitializedUntilFirstViewRequest

test/Microsoft.AspNet.Mvc.FunctionalTests/ActionResultTests.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
```
davidfowl: left over from Debugger.Launch

src/Microsoft.AspNet.Mvc/RazorPreCompileModule.cs
davidfowl: You don't actually need this services to do precompilation.

davidfowl: I filed a bug for this. The RazorPreCompiler not use ISP in the ctor.

HaoK:
Tests are all passing now

samples/MvcSample.Web/Startup.cs
```diff
@@ -16,15 +16,75 @@
...
+
+            services.ConfigureMvc(options =>
+            {
+                options.Filters.Add(typeof(PassThroughAttribute), order: 17);
+                options.AddXmlDataContractSerializerFormatter();
+                options.Filters.Add(new FormatFilterAttribute());
+            });
+
+#if DNX451
+            // REVIEW: This is hella hacky
```
dougbu: - replace this comment with a copy of the one in `Configure(IApplicationBuilder app)`
- this is not at all hacky compared to the previous approach of overriding the `static IServiceProvider` in all functional tests
- must remain as is until aspnet/Configuration#159 is fixed

samples/MvcSample.Web/Startup.cs
```diff
@@ -16,15 +16,75 @@
...
+            var applicationEnvironment = services.BuildServiceProvider().GetRequiredService<IApplicationEnvironment>();
+            var configurationPath = Path.Combine(applicationEnvironment.ApplicationBasePath, "config.json");
+
+            // Set up configuration sources.
+            var configuration = new Configuration()
+                .AddJsonFile(configurationPath)
+                .AddEnvironmentVariables();
+            string diSystem;
+            if (configuration.TryGet("DependencyInjection", out diSystem) &&
+                diSystem.Equals("AutoFac", StringComparison.OrdinalIgnoreCase))
```
dougbu: can the code from line 48 to here be done once?  that is, does Hosting create a single `Startup` instance and therefore support use of a `bool _isAutoFac;` field?

samples/MvcSample.Web/Startup.cs
```diff
@@ -16,15 +16,75 @@
...
+            {
+
```
dougbu: nit: remove extra blank line

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
```
dougbu: ??

samples/MvcSample.Web/Startup.cs
```diff
@@ -16,15 +16,75 @@
 #if DNX451
 using Microsoft.Framework.DependencyInjection.Autofac;
 using Microsoft.Framework.Runtime;
+using Microsoft.Framework.Runtime.Infrastructure;
 #endif
+using System.ComponentModel;
```
dougbu: what brought this in?

src/Microsoft.AspNet.Mvc/RazorViewEngineOptionsSetup.cs
```diff
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Diagnostics;
```
dougbu: ??

test/Microsoft.AspNet.Mvc.FunctionalTests/BestEffortLinkGenerationTest.cs
```diff
@@ -15,6 +16,8 @@ public class BestEffortLinkGenerationTest
     {
         private const string SiteName = nameof(BestEffortLinkGenerationWebSite);
         private readonly Action<IApplicationBuilder> _app = new BestEffortLinkGenerationWebSite.Startup().Configure;
+        private readonly Action<IServiceCollection> _configureServices = new BestEffortLinkGenerationWebSite.Startup().ConfigureServices;
+
```
dougbu: nit: remove extra blank lines added after a number of these new fields

test/Microsoft.AspNet.Mvc.FunctionalTests/TestHelper.cs
```diff
@@ -81,10 +113,11 @@ public static TestServer CreateServer(Action<IApplicationBuilder> builder, strin
                 applicationBasePath,
                 applicationWebSiteName);
             services.AddInstance<IApplicationEnvironment>(environment);
+            services.AddInstance<IHostingEnvironment>(new HostingEnvironment(environment));
```
dougbu: from the way you've updated the functional tests, it appears this is the only required change to this class.  plus likely the `using` updates.  revert adding the new, unused `CreateServer`, the `private` method renaming, and the minor change below.

dougbu:
:watch:

test/Microsoft.AspNet.Mvc.FunctionalTests/TestHelper.cs
dougbu: @HaoK please revert the changes to this file that were not required.  we don't need unused helper methods lying around.

src/Microsoft.AspNet.Mvc/RazorPreCompileModule.cs
dougbu: why did this "REVIEW" comment get checked in?  we don't do TODOs in the code

dougbu:
4b5dd19

PR #2202, opened 2015-03-19
React to aspnet/Razor#169 changes

ajaybhargavb:
@dougbu @NTaylorMullen


dougbu:
:shipit: as long as you can build from root with aspnet/Razor#169 bits installed (slightly surprised we only have one test like this)

NTaylorMullen:
Beautiful, :shipit:

Issue #1529, opened 2014-11-08
List of assembly names in DefaultAssemblyProvider needs to be kept up to date


pranavkm:
The list of assemblies in DefaultAssemblyProvider is out of date and 
* Is missing the WebAPI shim and Tag helpers
* Includes `Microsoft.AspNet.Mvc.Rendering` which no longer exists

We should determine if there is a way to keep this list up to date by adding a unit test. One way might be to look at the dependencies of `Microsoft.AspNet.Mvc` and use that to verify this - although I think it doesn't include TagHelpers and WebAPICompatShim.

ianhong:
1c6800beab42c69f8d5d724b6530305783bd2ec1

PR #2208, opened 2015-03-19
List of assembly names in DefaultAssemblyProvider needs to be kept up to date

ianhong:
@rynowak 

This is a PR for #1529. Please review this.

Thanks

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/DefaultAssemblyProvider.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/DefaultAssemblyProvider.cs
@@ -23,18 +23,21 @@ namespace Microsoft.AspNet.Mvc
         /// Gets the set of assembly names that are used as root for discovery of
         /// MVC controllers, view components and views.
         /// </summary>
-        // DefaultControllerTypeProvider uses CandidateAssemblies to determine if the base type of a POCO controller
-        // lives in an assembly that references MVC. CandidateAssemblies excludes all assemblies from the
-        // ReferenceAssemblies set. Consequently adding WebApiCompatShim to this set would cause the ApiController to
-        /// fail this test.
+        /// <see cref="DefaultControllerTypeProvider"/> uses <see cref="CandidateAssemblies"/> to determine
+        /// if the base type of a POCO controller lives in an assembly that references MVC.
+        /// <see cref="CandidateAssemblies"/> excludes all assemblies from the <see cref="ReferenceAssemblies"/> set.
+        /// Consequently adding WebApiCompatShim to this set would cause the ApiController to fail this test.
         protected virtual HashSet<string> ReferenceAssemblies { get; } = new HashSet<string>(StringComparer.Ordinal)
         {
             "Microsoft.AspNet.Mvc",
+            "Microsoft.AspNet.Mvc.Common",
             "Microsoft.AspNet.Mvc.Core",
             "Microsoft.AspNet.Mvc.ModelBinding",
             "Microsoft.AspNet.Mvc.Razor",
             "Microsoft.AspNet.Mvc.Razor.Host",
             "Microsoft.AspNet.Mvc.TagHelpers",
+            "Microsoft.AspNet.Mvc.Xml",
+            "Microsoft.AspNet.PageExecutionInstrumentation.Interfaces",
         };
 
         /// <inheritdoc />
--- a/src/Microsoft.AspNet.Mvc.Core/DefaultAssemblyProvider.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/DefaultAssemblyProvider.cs
@@ -23,14 +23,13 @@ namespace Microsoft.AspNet.Mvc
         /// Gets the set of assembly names that are used as root for discovery of
         /// MVC controllers, view components and views.
         /// </summary>
-        /// <see cref="DefaultControllerTypeProvider"/> uses <see cref="CandidateAssemblies"/> to determine
-        /// if the base type of a POCO controller lives in an assembly that references MVC.
-        /// <see cref="CandidateAssemblies"/> excludes all assemblies from the <see cref="ReferenceAssemblies"/> set.
-        /// Consequently adding WebApiCompatShim to this set would cause the ApiController to fail this test.
+        /// DefaultControllerTypeProvider uses CandidateAssemblies to determine if the base type of a POCO controller
+        /// lives in an assembly that references MVC. CandidateAssemblies excludes all assemblies from the
+        /// ReferenceAssemblies set. Consequently adding WebApiCompatShim to this set would cause the ApiController to
+        /// fail this test.
         protected virtual HashSet<string> ReferenceAssemblies { get; } = new HashSet<string>(StringComparer.Ordinal)
         {
             "Microsoft.AspNet.Mvc",
-            "Microsoft.AspNet.Mvc.Common",
             "Microsoft.AspNet.Mvc.Core",
             "Microsoft.AspNet.Mvc.ModelBinding",
             "Microsoft.AspNet.Mvc.Razor",
--- a/src/Microsoft.AspNet.Mvc.Core/DefaultAssemblyProvider.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/DefaultAssemblyProvider.cs
@@ -23,10 +23,10 @@ namespace Microsoft.AspNet.Mvc
         /// Gets the set of assembly names that are used as root for discovery of
         /// MVC controllers, view components and views.
         /// </summary>
-        /// DefaultControllerTypeProvider uses CandidateAssemblies to determine if the base type of a POCO controller
-        /// lives in an assembly that references MVC. CandidateAssemblies excludes all assemblies from the
-        /// ReferenceAssemblies set. Consequently adding WebApiCompatShim to this set would cause the ApiController to
-        /// fail this test.
+        // DefaultControllerTypeProvider uses CandidateAssemblies to determine if the base type of a POCO controller
+        // lives in an assembly that references MVC. CandidateAssemblies excludes all assemblies from the
+        // ReferenceAssemblies set. Consequently adding WebApiCompatShim to this set would cause the ApiController to
+        // fail this test.
         protected virtual HashSet<string> ReferenceAssemblies { get; } = new HashSet<string>(StringComparer.Ordinal)
         {
             "Microsoft.AspNet.Mvc",
--- a/test/Microsoft.AspNet.Mvc.Core.Test/DefaultAssemblyProviderTests.cs
+++ b/test/Microsoft.AspNet.Mvc.Core.Test/DefaultAssemblyProviderTests.cs
@@ -104,31 +104,39 @@ namespace Microsoft.AspNet.Mvc.Core
         {
             // Arrange
             var provider = new MvcAssembliesTestingProvider();
-            var allAssemblies = provider.AllLoadedAssemblies;
-            var expected = provider.ExpectedReferenceAssemblies;
+
+            var expected = new HashSet<string>
+                {
+                    "Microsoft.AspNet.Mvc",
+                    "Microsoft.AspNet.Mvc.Core",
+                    "Microsoft.AspNet.Mvc.ModelBinding",
+                    "Microsoft.AspNet.Mvc.Razor.Host",
+                    "Microsoft.AspNet.Mvc.Razor",
+                    "Microsoft.AspNet.Mvc.TagHelpers",
+                    "Microsoft.AspNet.Mvc.Xml",
+                    "Microsoft.AspNet.PageExecutionInstrumentation.Interfaces",
+                };
 
             // Act
-            var referenceAssemblies = provider.DefaultAssemblyProviderReferenceAssemblies;
+            var referenceAssemblies = provider.ReferenceAssemblies;
 
             // Assert
-            Assert.Contains("Microsoft.AspNet.Mvc", allAssemblies);
             Assert.True(expected.SetEquals(referenceAssemblies));
         }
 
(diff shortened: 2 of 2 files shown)


src/Microsoft.AspNet.Mvc.Core/DefaultAssemblyProvider.cs
```diff
@@ -23,18 +23,21 @@ public DefaultAssemblyProvider(ILibraryManager libraryManager)
...
         /// MVC controllers, view components and views.
         /// </summary>
-        // DefaultControllerTypeProvider uses CandidateAssemblies to determine if the base type of a POCO controller
-        // lives in an assembly that references MVC. CandidateAssemblies excludes all assemblies from the
-        // ReferenceAssemblies set. Consequently adding WebApiCompatShim to this set would cause the ApiController to
-        /// fail this test.
+        /// <see cref="DefaultControllerTypeProvider"/> uses <see cref="CandidateAssemblies"/> to determine
+        /// if the base type of a POCO controller lives in an assembly that references MVC.
+        /// <see cref="CandidateAssemblies"/> excludes all assemblies from the <see cref="ReferenceAssemblies"/> set.
+        /// Consequently adding WebApiCompatShim to this set would cause the ApiController to fail this test.
```
rynowak: Revert this.

This isn't user-visible documentation, it's a comment for us. Adding the docs constructs like `<see .../>` has no effect.

src/Microsoft.AspNet.Mvc.Core/DefaultAssemblyProvider.cs
```diff
@@ -23,18 +23,21 @@ public DefaultAssemblyProvider(ILibraryManager libraryManager)
...
         protected virtual HashSet<string> ReferenceAssemblies { get; } = new HashSet<string>(StringComparer.Ordinal)
         {
             "Microsoft.AspNet.Mvc",
+            "Microsoft.AspNet.Mvc.Common",
```
rynowak: This isn't an assembly as-such, it's a shared source package. Remove this.

test/Microsoft.AspNet.Mvc.Core.Test/DefaultAssemblyProviderTests.cs
```diff
@@ -164,5 +197,81 @@ public NullAssemblyProvider(ILibraryManager libraryManager) : base(libraryManage
...
+                        "Microsoft.AspNet.Mvc.Razor.Host",
+                        "Microsoft.AspNet.Mvc.Razor",
+                        "Microsoft.AspNet.Mvc.TagHelpers",
+                        "Microsoft.AspNet.Mvc.Xml",
+                        "Microsoft.AspNet.PageExecutionInstrumentation.Interfaces",
+                    };
+                }
+            }
+
+            public IEnumerable<string> DefaultAssemblyProviderReferenceAssemblies
```
rynowak: If all you're doing is changing the visibility, then use `new`

test/Microsoft.AspNet.Mvc.Core.Test/DefaultAssemblyProviderTests.cs
```diff
@@ -164,5 +197,81 @@ public NullAssemblyProvider(ILibraryManager libraryManager) : base(libraryManage
...
+                    // The following assemblies are not reachable from Microsoft.AspNet.Mvc
+                    mvcAssemblies.Add("Microsoft.AspNet.Mvc.TagHelpers");
+                    mvcAssemblies.Add("Microsoft.AspNet.Mvc.Xml");
+                    mvcAssemblies.Add("Microsoft.AspNet.PageExecutionInstrumentation.Interfaces");
+
+                    return new HashSet<string>(mvcAssemblies.Distinct());
+                }
+            }
+
+            public HashSet<string> ExpectedReferenceAssemblies
```
rynowak: I would expect to see this in the test itself where it's more visible.

test/Microsoft.AspNet.Mvc.Core.Test/DefaultAssemblyProviderTests.cs
```diff
@@ -164,5 +197,81 @@ public NullAssemblyProvider(ILibraryManager libraryManager) : base(libraryManage
...
+
+            public IEnumerable<string> DefaultAssemblyProviderReferenceAssemblies
+            {
+                get
+                {
+                    return base.ReferenceAssemblies;
+                }
+            }
+
+            public IEnumerable<string> AllLoadedAssemblies
```
rynowak: This isn't really loaded assemblies, it's the available libraries.

test/Microsoft.AspNet.Mvc.Core.Test/DefaultAssemblyProviderTests.cs
```diff
@@ -164,5 +197,81 @@ public NullAssemblyProvider(ILibraryManager libraryManager) : base(libraryManage
...
+        private class MvcAssembliesTestingProvider : DefaultAssemblyProvider
+        {
+            private static readonly ILibraryManager _libraryManager = GetLibraryManager();
+            private static readonly string _mvcName = "Microsoft.AspNet.Mvc";
+            private static IEnumerable<string> _assemblies; 
+
+            public MvcAssembliesTestingProvider() : base(_libraryManager)
+            { }
+
+            public HashSet<string> LoadedMvcReferenceAssemblies
```
rynowak: This seems really fragile. You want to get the library reference for `"Microsoft.AspNet.Mvc"` and then see what IT references. 

The goal is to **generate** the list that you have below in `ExpectedReferenceAssemblies`.

rynowak:
:watch:

ianhong:
@rynowak 

I updated per your comments. Please review this once more.

src/Microsoft.AspNet.Mvc.Core/DefaultAssemblyProvider.cs
```diff
@@ -23,9 +23,9 @@ public DefaultAssemblyProvider(ILibraryManager libraryManager)
         /// Gets the set of assembly names that are used as root for discovery of
         /// MVC controllers, view components and views.
         /// </summary>
-        // DefaultControllerTypeProvider uses CandidateAssemblies to determine if the base type of a POCO controller
-        // lives in an assembly that references MVC. CandidateAssemblies excludes all assemblies from the
-        // ReferenceAssemblies set. Consequently adding WebApiCompatShim to this set would cause the ApiController to
+        /// DefaultControllerTypeProvider uses CandidateAssemblies to determine if the base type of a POCO controller
+        /// lives in an assembly that references MVC. CandidateAssemblies excludes all assemblies from the
+        /// ReferenceAssemblies set. Consequently adding WebApiCompatShim to this set would cause the ApiController to
```
rynowak: Still should be just 2 slashes. These aren't doc comments

test/Microsoft.AspNet.Mvc.Core.Test/DefaultAssemblyProviderTests.cs
```diff
@@ -164,5 +205,100 @@ public NullAssemblyProvider(ILibraryManager libraryManager) : base(libraryManage
...
+
+            public new HashSet<string> ReferenceAssemblies
+            {
+                get
+                {
+                    return base.ReferenceAssemblies;
+                }
+            }
+
+            private static HashSet<string> GetAllDepedencies(string libraryName, HashSet<string> dependencies)
```
rynowak: `GetAllDepedencies` -> `GetAllDependencies`

test/Microsoft.AspNet.Mvc.Core.Test/DefaultAssemblyProviderTests.cs
```diff
@@ -164,5 +205,100 @@ public NullAssemblyProvider(ILibraryManager libraryManager) : base(libraryManage
...
+            {
+                var directDependencies = _libraryManager.GetLibraryInformation(libraryName).Dependencies;
+
+                if (directDependencies != null && directDependencies.Count() > 0)
```
rynowak: `Count() > 0` is not needed. This will do the same thing without it. Think before you `Count()` -- its almost always a code smell.

test/Microsoft.AspNet.Mvc.Core.Test/DefaultAssemblyProviderTests.cs
```diff
@@ -164,5 +205,100 @@ public NullAssemblyProvider(ILibraryManager libraryManager) : base(libraryManage
...
+
+            public HashSet<string> LoadableReferenceAssemblies
+            {
+                get
+                {
+                    if (_mvcReferenceAssemblies != null)
+                    {
+                        return _mvcReferenceAssemblies;
+                    }
+
```
rynowak: why do you care about caching and statics here - this is test code.. just write it in a simple and clear way.

test/Microsoft.AspNet.Mvc.Core.Test/DefaultAssemblyProviderTests.cs
```diff
@@ -164,5 +205,100 @@ public NullAssemblyProvider(ILibraryManager libraryManager) : base(libraryManage
...
+                    var dependencies = new HashSet<string>() { _mvcName };
+                    dependencies = GetAllDepedencies(_mvcName, dependencies);
```
rynowak: Why do you return something here?

test/Microsoft.AspNet.Mvc.Core.Test/DefaultAssemblyProviderTests.cs
```diff
@@ -164,5 +205,100 @@ public NullAssemblyProvider(ILibraryManager libraryManager) : base(libraryManage
...
+                    .Where(n => n.StartsWith(_mvcName))
+                    .Except(exceptionalAssebmlies)
+                    .ToList();
+
+                // The following assemblies are not reachable from Microsoft.AspNet.Mvc
+                mvcAssemblies.Add("Microsoft.AspNet.Mvc.TagHelpers");
+                mvcAssemblies.Add("Microsoft.AspNet.Mvc.Xml");
+                mvcAssemblies.Add("Microsoft.AspNet.PageExecutionInstrumentation.Interfaces");
+
+                System.Diagnostics.Debugger.Launch();
```
rynowak: Remove before checkin

test/Microsoft.AspNet.Mvc.Core.Test/DefaultAssemblyProviderTests.cs
```diff
@@ -98,6 +99,46 @@ public void CandidateAssemblies_ReturnsEmptySequenceWhenReferenceAssembliesIsNul
             Assert.Empty(nullProviderCandidates.Select(a => a.Name));
         }
 
+        [Fact]
+        public void ReferenceAssemblies_ReturnsExpectedReferenceAssemblies()
```
rynowak: Needs some comments about what these tests are for, explain that you will want to modify the test if you've added a new dependency.

rynowak:
:watch:

test/Microsoft.AspNet.Mvc.Core.Test/DefaultAssemblyProviderTests.cs
ianhong: Oops.. sorry.

ianhong:
@rynowak 
I updated this PR, it took some time to find an another way to get ILibraryManager instead of relying on HostingServices.Create.

Thanks

rynowak:
:shipit:

test/Microsoft.AspNet.Mvc.Core.Test/DefaultAssemblyProviderTests.cs
```diff
@@ -98,6 +100,51 @@ public void CandidateAssemblies_ReturnsEmptySequenceWhenReferenceAssembliesIsNul
             Assert.Empty(nullProviderCandidates.Select(a => a.Name));
         }
 
+        // This test verifies DefaultAssemblyProvider.ReferenceAssemblies returns the assembly list that we intended.
+        // If ReferenceAssemblies needs to be changed, the expected list in this test should be updated together.
+        [Fact]
+        public void ReferenceAssemblies_ReturnsExpectedReferenceAssemblies()
```
Eilon: Can this test ever possibly catch a bug? What change could someone be making where this test would provide any meaningful information? I feel like adding a comment to the product source code would be just as useful because there's nothing "computed" here - it's just verifying a static list.

Eilon: And, for whatever reason, the assemblies in the unit test are listed in a different order from the product code (not that it matters, because the sets are equal, but it makes me even more suspicious about this).

davidfowl: This is a pretty weird test.

ianhong: This test is for catching the case where the list in the product code is accidently changed.  Normally, when we automate the tests for a package, we would add this type of test to check its assets. But if you think this type of test is useless from the unit test perspective, let me know. I'll send a PR to remove.

Eilon: Suppose someone inadvertently adds another assembly to the list. Their immediate next step is to just "fix" the unit test and the product is then broken. As such, I think a big giant warning comment in the product source code would yield a better result.

If there's a way to programmatically generate the list, or somehow verify that the list doesn't include anything that can't possibly be correct, that would be great. But comparing one static list to another static list doesn't do much.

ianhong:
1c6800beab42c69f8d5d724b6530305783bd2ec1

PR #2210, opened 2015-03-19
Fix build break in jsonpatch branch

kirthik:
@rynowak - could you please take a look. This fix will make jsonpatch branch to build and tests to pass.


rynowak:
:shipit:

Issue #2195, opened 2015-03-18
Transition MVC TagHelpers to use TargetElementAttributes attribute targeting.


NTaylorMullen:
Now that `TagHelper`s can target attributes we should transition our existing `TagHelper`s to utilize this new "required" attributes functionality.

One example: `ValidationSummaryTagHelper` shouldn't target "div" elements then no-op if the attribute doesn't exist: It should target the `asp-validation-summary` attribute on any HTML element.

This should also enable us to no longer have ambiguity between "should I be a tag helper" for all of our tag helpers.

dougbu:
it's likely views would perform better if tag helpers were not created only to no-op.  however we should measure things to confirm this is worthwhile since the MVC tag helpers are doing simple checks that work today.

specifically:
> ValidationSummaryTagHelper shouldn't target "div" elements then no-op if the attribute doesn't exist: It should target the asp-validation-summary attribute on any HTML element.

"any" is too broad.  we likely want to target `<div>` and `<span>` and that's about it.  in any case, we should strive to preserve the chosen behaviour and limit work on this issue to the implementation choice.  deal with reevaluating the behaviours separately.

> This should also enable us to no longer have ambiguity between "should I be a tag helper" ...

could you clarify this statement?  is this another point about changing the current MVC tag helper behaviours?

danroth27:
@DamianEdwards Let us know if you think we need to pull this back into the current milestone.

DamianEdwards:
We should do this now

NTaylorMullen:
@dougbu The ambiguity bit I mentioned will drastically improve `TagHelper` tooling. For instance, today in the pre-bits for `TagHelper` tooling every `div` tag is highlighted because of the validation summary maybe being a `TagHelper` on div elements. Same thing goes for all other `TagHelper`s.

dougbu:
@NTaylorMullen thanks, that was much clearer

ajaybhargavb:
Part 1 - https://github.com/aspnet/Mvc/commit/bfdeda797df054a4ae8a6fa9250e65f07cd63a1e
Part 2 - https://github.com/aspnet/Mvc/commit/e2058905ec30facdf41d1eca2360be3c4b6b7314
Part 3 - https://github.com/aspnet/Mvc/commit/fd5e07dd0323812263b9fc0160ad56be3a4f3f6c

PR #2211, opened 2015-03-19
TagHelpers attribute targeting - part 1

ajaybhargavb:
Issue #2195 
@NTaylorMullen @dougbu

### DIFF
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
@@ -15,6 +15,7 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
     /// <summary>
     /// <see cref="ITagHelper"/> implementation targeting &lt;input&gt; elements with an <c>asp-for</c> attribute.
     /// </summary>
+    [TargetElement("input", Attributes = ForAttributeName)]
     public class InputTagHelper : TagHelper
     {
         private const string ForAttributeName = "asp-for";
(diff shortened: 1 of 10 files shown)


src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
```diff
@@ -121,93 +122,79 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
                 output.CopyHtmlAttribute(nameof(Value), context);
             }
 
-            if (For == null)
+            // Note null or empty For.Name is allowed because TemplateInfo.HtmlFieldPrefix may be sufficient.
+            // IHtmlGenerator will enforce name requirements.
+            var metadata = For.Metadata;
+            var modelExplorer = For.ModelExplorer;
+            if (metadata == null)
```
rynowak: There seem to be more changes going on here than just the attribute targeting, whatsis?

src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
```diff
@@ -121,93 +122,79 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
...
             {
-                // Regular HTML <input/> element. Just make sure Format wasn't specified.
-                if (Format != null)
-                {
-                    throw new InvalidOperationException(Resources.FormatInputTagHelper_UnableToFormat(
```
NTaylorMullen: Should be able to remove this resource.

src/Microsoft.AspNet.Mvc.TagHelpers/LabelTagHelper.cs
```diff
@@ -32,34 +33,31 @@ public class LabelTagHelper : TagHelper
         /// <remarks>Does nothing if <see cref="For"/> is <c>null</c>.</remarks>
         public override async Task ProcessAsync(TagHelperContext context, TagHelperOutput output)
         {
-            if (For != null)
+            var tagBuilder = Generator.GenerateLabel(ViewContext,
+                                                        For.ModelExplorer,
```
NTaylorMullen: spacing, just put all parameters at 1 indent (hit enter before ViewContext and wherever it lands is good). This way it's easier to maintain.

src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
NTaylorMullen: This was previous code, the diff is just misleading. Since we require the `asp-for` attribute to be provided there was no longer a need to check if it was provided.

NTaylorMullen:
solid, :shipit:

rynowak:
:shipit: from me if @NTaylorMullen is happy

src/Microsoft.AspNet.Mvc.TagHelpers/ValidationMessageTagHelper.cs
```diff
@@ -8,10 +8,10 @@
 namespace Microsoft.AspNet.Mvc.TagHelpers
 {
     /// <summary>
-    /// <see cref="ITagHelper"/> implementation targeting &lt;span&gt; elements with an <c>asp-validation-for</c>
+    /// <see cref="ITagHelper"/> implementation targeting any HTML element with an <c>asp-validation-for</c>
     /// attribute.
     /// </summary>
-    [TargetElement("span")]
+    [TargetElement(Attributes = ValidationForAttributeName)]
```
dougbu: this change in behaviour should be separate.  discuss w/ @DamianEdwards as to whether we need to make this change.

src/Microsoft.AspNet.Mvc.TagHelpers/ValidationSummaryTagHelper.cs
```diff
@@ -8,10 +8,10 @@
 namespace Microsoft.AspNet.Mvc.TagHelpers
 {
     /// <summary>
-    /// <see cref="ITagHelper"/> implementation targeting &lt;div&gt; elements with an <c>asp-validation-summary</c>
+    /// <see cref="ITagHelper"/> implementation targeting any HTML element with an <c>asp-validation-summary</c>
     /// attribute.
     /// </summary>
-    [TargetElement("div")]
+    [TargetElement(Attributes = ValidationSummaryAttributeName)]
```
dougbu: this change in behaviour should be separate. discuss w/ @DamianEdwards as to whether we need to make this change.

note as well that there aren't that many block elements in HTML.  `<div>` was one of the few where splitting out the list made sense.

test/Microsoft.AspNet.Mvc.FunctionalTests/Compiler/Resources/TagHelpersWebSite.Employee.Create.Invalid.html
```diff
@@ -69,7 +69,7 @@
 
             <div class="form-group">
                 <div class="col-md-offset-2 col-md-10">
-                    <input class="btn btn-default" type="submit" value="Create" />
+                    <input type="submit" value="Create" class="btn btn-default" />
```
dougbu: why did this change?

dougbu:
:watch: doesn't look ready

test/Microsoft.AspNet.Mvc.FunctionalTests/Compiler/Resources/TagHelpersWebSite.Employee.Create.Invalid.html
NTaylorMullen: Used to be a `TagHelper` and it no longer is so the attribute ordering is no longer affected.

ajaybhargavb:
Created a follow up issue https://github.com/aspnet/Mvc/issues/2220

ajaybhargavb:
bfdeda797df054a4ae8a6fa9250e65f07cd63a1e

PR #2223, opened 2015-03-20
Fixing some inaccurate test names and comments

rynowak:
These tests had the wrong names and comments.

### DIFF
--- a/test/Microsoft.AspNet.Mvc.FunctionalTests/ModelBindingFromHeaderTest.cs
+++ b/test/Microsoft.AspNet.Mvc.FunctionalTests/ModelBindingFromHeaderTest.cs
@@ -295,7 +295,7 @@ namespace Microsoft.AspNet.Mvc.FunctionalTests
 
         // This model sets a value for 'Title', and the model binder won't trounce it.
         //
-        // There's no validation error because we validate the initialized value.
+        // There's a validation error because we validate the value in the request (not present).
         [Fact]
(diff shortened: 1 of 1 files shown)


test/Microsoft.AspNet.Mvc.FunctionalTests/ModelBindingFromHeaderTest.cs
```diff
@@ -325,11 +325,12 @@ public class ModelBindingFromHeaderTest
...
 
-        // This model uses default value for 'Title'.
+        // This model uses [DefaultValueAttribute(...)] for 'Title', and the model binder won't 
+        // trounce it.
         //
-        // There's no validation error because we validate the default value.
+        // There's a validation error because we validate the value in the request (not present).
         [Fact]
-        public async Task FromHeader_BindHeader_ToModel_NoValues_DefaultValue_NoValidationError()
+        public async Task FromHeader_BindHeader_ToModel_NoValues_DefaultValueAttribute_ValidationError()
```
dougbu: nit: why bother w/ the `Attribute` part of the name?

dougbu:
:shipit:

test/Microsoft.AspNet.Mvc.FunctionalTests/ModelBindingFromHeaderTest.cs
rynowak: I added that to be more explicit. When I've said 'default value' in the past, it's been mistaken for a value set in the constructor of the model.

Issue #2141, opened 2015-03-08
Script & Link tag helpers should support generating cache-busting file version hash in URL


DamianEdwards:
The `ScriptTagHelper` and `LinkTagHelper` should support an option that enables the generation of a file version hash of the linked file contents placed in the URL for the purposes of busting client-side cache. The hash should be placed in the query-string (key `v`). The helpers should render the hash by default.

This feature existed in `System.Web.Optimization` so there might be some code we can use from there.

### Example
``` HTML
<link src="~/lib/bootstrap/css/bootstrap.min.css" rel="stylesheet" asp-file-version="true" />
<script src="~/lib/jquery/jquery.min.js" asp-file-version="true"></script>
```

BrianVallelunga:
Any chance we'll get the cache-busting feature for images too?

DamianEdwards:
@BrianVallelunga great idea! The `<img />` tag should support globbing too. Logged #2249 to track it

sornaks:
https://github.com/aspnet/Mvc/commit/0e783ace5898c761435541aa937085a8593a75d0

PR #2225, opened 2015-03-20
Issue #2141 - Script & Link tag helpers should support generating cache-busting file version hash in URL.

sornaks:
#2141

### DIFF
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
@@ -88,7 +88,8 @@ namespace Microsoft.AspNet.Mvc.TagHelpers.Internal
             {
                 if (!context.AllAttributes.ContainsKey(attribute) ||
                     context.AllAttributes[attribute] == null ||
-                    string.IsNullOrWhiteSpace(context.AllAttributes[attribute] as string))
+                    (typeof(string).IsAssignableFrom(context.AllAttributes[attribute].GetType()) &&
+                    string.IsNullOrWhiteSpace(context.AllAttributes[attribute] as string)))
                 {
                     // Missing attribute!
                     missingAttributes.Add(attribute);
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
@@ -51,13 +51,10 @@ namespace Microsoft.AspNet.Mvc.TagHelpers.Internal
             var fileInfo = _fileProvider.GetFileInfo(resolvedPath);
             if (!fileInfo.Exists)
             {
-                if (_requestPathBase.HasValue)
+                if (_requestPathBase.HasValue &&
+                    resolvedPath.StartsWith(_requestPathBase.Value, StringComparison.OrdinalIgnoreCase))
                 {
-                    if (resolvedPath.StartsWith(_requestPathBase.Value, StringComparison.OrdinalIgnoreCase))
-                    {
-                        resolvedPath = resolvedPath.Substring(_requestPathBase.Value.Length + 1);
-                    }
-
+                    resolvedPath = resolvedPath.Substring(_requestPathBase.Value.Length);
                     fileInfo = _fileProvider.GetFileInfo(resolvedPath);
                 }
 
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
@@ -1,0 +1,88 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Security.Cryptography;
+using Microsoft.AspNet.FileProviders;
+using Microsoft.AspNet.Http;
+using Microsoft.AspNet.WebUtilities;
+using Microsoft.Framework.Caching.Memory;
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.AspNet.Mvc.TagHelpers.Internal
+{
+    /// <summary>
+    /// Provides version hash for a specified file.
+    /// </summary>
+    public class FileVersionProvider
+    {
+        private const string VersionKey = "v";
+        private readonly IFileProvider _fileProvider;
+        private readonly IMemoryCache _cache;
+        private readonly PathString _requestPathBase;
+
+        /// <summary>
+        /// Creates a new instance of <see cref="FileVersionProvider"/>.
+        /// </summary>
+        /// <param name="fileProvider">The file provider to get and watch files.</param>
+        /// <param name="applicationName">Name of the application.</param>
+        /// <param name="cache"><see cref="IMemoryCache"/> where versioned urls of files are cached.</param>
+        public FileVersionProvider(
+            [NotNull] IFileProvider fileProvider,
+            [NotNull] IMemoryCache cache,
+            [NotNull] PathString requestPathBase)
+        {
+            _fileProvider = fileProvider;
+            _cache = cache;
+            _requestPathBase = requestPathBase;
+        }
+
+        /// <summary>
+        /// Adds version query parameter to the specified file path.
+        /// </summary>
+        /// <param name="path">The path of the file to which version should be added.</param>
+        /// <returns>Path containing the version query string.</returns>
+        /// <remarks>
+        /// The version query string is appended as with the key "v".
+        /// </remarks>
+        public string AddFileVersionToPath([NotNull] string path)
+        {
+            var resolvedPath = path;
+            var fileInfo = _fileProvider.GetFileInfo(resolvedPath);
+            if (!fileInfo.Exists)
+            {
+                if (_requestPathBase.HasValue)
+                {
+                    if (resolvedPath.StartsWith(_requestPathBase.Value, StringComparison.OrdinalIgnoreCase))
+                    {
+                        resolvedPath = resolvedPath.Substring(_requestPathBase.Value.Length + 1);
+                    }
+
+                    fileInfo = _fileProvider.GetFileInfo(resolvedPath);
+                }
+
+                if (!fileInfo.Exists)
+                {
+                    // if the file is not in the current server.
+                    return path;
+                }
+            }
+
+            return _cache.GetOrSet(path, cacheGetOrSetContext =>
+            {
+                var trigger = _fileProvider.Watch(resolvedPath);
+                cacheGetOrSetContext.AddExpirationTrigger(trigger);
+                return QueryHelpers.AddQueryString(path, VersionKey, GetHashForFile(fileInfo));
+            });
+        }
+
+        private string GetHashForFile(IFileInfo fileInfo)
+        {
+            using (var sha256 = SHA256.Create())
+            {
+                var hash = sha256.ComputeHash(fileInfo.CreateReadStream());
+                return WebEncoders.Base64UrlEncode(hash);
+            }
+        }
+    }
+}
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
@@ -29,6 +29,7 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
     [TargetElement("link", Attributes = FallbackTestClassAttributeName)]
     [TargetElement("link", Attributes = FallbackTestPropertyAttributeName)]
     [TargetElement("link", Attributes = FallbackTestValueAttributeName)]
+    [TargetElement("link", Attributes = FileVersionAttributeName)]
     public class LinkTagHelper : TagHelper
     {
         private const string HrefIncludeAttributeName = "asp-href-include";
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
@@ -5,12 +5,12 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
-using System.Text;
 using Microsoft.AspNet.Hosting;
 using Microsoft.AspNet.Mvc.TagHelpers.Internal;
 using Microsoft.AspNet.Razor.Runtime.TagHelpers;
 using Microsoft.Framework.Caching.Memory;
 using Microsoft.Framework.Logging;
+using Microsoft.Framework.Runtime;
 using Microsoft.Framework.WebEncoders;
 
 namespace Microsoft.AspNet.Mvc.TagHelpers
@@ -40,8 +40,14 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
         private const string FallbackTestPropertyAttributeName = "asp-fallback-test-property";
         private const string FallbackTestValueAttributeName = "asp-fallback-test-value";
         private const string FallbackJavaScriptResourceName = "compiler/resources/LinkTagHelper_FallbackJavaScript.js";
+        private const string FileVersionAttributeName = "asp-file-version";
+        private const string HrefAttributeName = "href";
+
+        private FileVersionProvider _fileVersionProvider;
 
         private static readonly ModeAttributes<Mode>[] ModeDetails = new[] {
+            // Regular src with file version alone
+            ModeAttributes.Create(Mode.FileVersion, new[] { FileVersionAttributeName }),
             // Globbed Href (include only) no static href
             ModeAttributes.Create(Mode.GlobbedHref, new [] { HrefIncludeAttributeName }),
             // Globbed Href (include & exclude), no static href
@@ -77,14 +83,18 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
         private enum Mode
         {
             /// <summary>
+            /// Just adding a file version for the generated urls.
+            /// </summary>
+            FileVersion = 0,
+            /// <summary>
             /// Just performing file globbing search for the href, rendering a separate &lt;link&gt; for each match.
             /// </summary>
-            GlobbedHref = 0,
+            GlobbedHref = 1,
             /// <summary>
             /// Rendering a fallback block if primary stylesheet fails to load. Will also do globbing for both the
             /// primary and fallback hrefs if the appropriate properties are set.
             /// </summary>
-            Fallback = 1,
+            Fallback = 2,
         }
 
         /// <summary>
@@ -107,6 +117,15 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
         /// </summary>
         [HtmlAttributeName(FallbackHrefAttributeName)]
         public string FallbackHref { get; set; }
+
+        /// <summary>
+        /// Value indicating if file version should be appended to the href urls.
+        /// </summary>
+        /// <remarks>
+        /// If <c>true</c> then a query string "v" with the encoded content of the file is added.
+        /// </remarks>
+        [HtmlAttributeName(FileVersionAttributeName)]
+        public bool? FileVersion { get; set; }
 
         /// <summary>
         /// A comma separated list of globbed file patterns of CSS stylesheets to fallback to in the case the primary
@@ -149,7 +168,8 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
         [HtmlAttributeName(FallbackTestValueAttributeName)]
         public string FallbackTestValue { get; set; }
 
-        // Properties are protected to ensure subclasses are correctly activated. Internal for ease of use when testing.
+        // Properties are protected to ensure subclasses are correctly activated.
+        // Internal for ease of use when testing.
         [Activate]
         protected internal ILoggerFactory LoggerFactory { get; set; }
 
@@ -194,9 +214,10 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
 
             var builder = new DefaultTagHelperContent();
 
-            if (mode == Mode.Fallback && string.IsNullOrEmpty(HrefInclude))
+            if (mode == Mode.Fallback && string.IsNullOrEmpty(HrefInclude) || mode == Mode.FileVersion)
             {
-                // No globbing to do, just build a <link /> tag to match the original one in the source file
+                // No globbing to do, just build a <link /> tag to match the original one in the source file.
+                // Or just add file version to the link tag.
                 BuildLinkTag(attributes, builder);
             }
             else
@@ -218,14 +239,14 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
         {
             // Build a <link /> tag for each matched href as well as the original one in the source file
             string staticHref;
-            attributes.TryGetValue("href", out staticHref);
+            attributes.TryGetValue(HrefAttributeName, out staticHref);
 
             EnsureGlobbingUrlBuilder();
             var urls = GlobbingUrlBuilder.BuildUrlList(staticHref, HrefInclude, HrefExclude);
 
             foreach (var url in urls)
             {
-                attributes["href"] = HtmlEncoder.HtmlEncode(url);
+                attributes[HrefAttributeName] = url;
                 BuildLinkTag(attributes, builder);
             }
         }
@@ -233,10 +254,19 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
         private void BuildFallbackBlock(TagHelperContent builder)
         {
             EnsureGlobbingUrlBuilder();
-            var fallbackHrefs = GlobbingUrlBuilder.BuildUrlList(FallbackHref, FallbackHrefInclude, FallbackHrefExclude);
+            var fallbackHrefs =
+                GlobbingUrlBuilder.BuildUrlList(FallbackHref, FallbackHrefInclude, FallbackHrefExclude).ToArray();
 
-            if (fallbackHrefs.Any())
+            if (fallbackHrefs.Length > 0)
             {
+                if (ShouldAddFileVersion())
+                {
+                    for (var i=0; i < fallbackHrefs.Length; i++)
+                    {
+                        fallbackHrefs[i] = _fileVersionProvider.AddFileVersionToPath(fallbackHrefs[i]);
+                    }
+                }
+
                 builder.Append(Environment.NewLine);
 
                 // Build the <meta /> tag that's used to test for the presence of the stylesheet
@@ -269,17 +299,46 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
             }
         }
 
-        private static void BuildLinkTag(IDictionary<string, string> attributes, TagHelperContent builder)
+        private void EnsureFileVersionProvider()
         {
+            if (_fileVersionProvider == null)
+            {
+                _fileVersionProvider = new FileVersionProvider(
+                    HostingEnvironment.WebRootFileProvider,
+                    Cache,
+                    ViewContext.HttpContext.Request.PathBase);
+            }
+        }
+
+        private void BuildLinkTag(IDictionary<string, string> attributes, TagHelperContent builder)
+        {
+            EnsureFileVersionProvider();
             builder.Append("<link ");
 
             foreach (var attribute in attributes)
             {
-                builder.Append(
-                    string.Format(CultureInfo.InvariantCulture, "{0}=\"{1}\" ", attribute.Key, attribute.Value));
+                var attributeValue = attribute.Value;
+                if (string.Equals(attribute.Key, HrefAttributeName, StringComparison.OrdinalIgnoreCase))
+                {
+                    attributeValue = HtmlEncoder.HtmlEncode(
+                        ShouldAddFileVersion() ?
+                            _fileVersionProvider.AddFileVersionToPath(attributeValue) :
+                            attributeValue);
+                }
+
+                builder
+                    .Append(attribute.Key)
+                    .Append("=\"")
+                    .Append(attributeValue)
+                    .Append("\" ");
             }
 
             builder.Append("/>");
+        }
+
+        private bool ShouldAddFileVersion()
+        {
+            return FileVersion ?? false;
         }
     }
 }
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Globalization;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Hosting;
@@ -11,6 +10,7 @@ using Microsoft.AspNet.Mvc.TagHelpers.Internal;
 using Microsoft.AspNet.Razor.Runtime.TagHelpers;
 using Microsoft.Framework.Caching.Memory;
 using Microsoft.Framework.Logging;
+using Microsoft.Framework.Runtime;
 using Microsoft.Framework.WebEncoders;
 
 namespace Microsoft.AspNet.Mvc.TagHelpers
@@ -36,8 +36,13 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
         private const string FallbackSrcExcludeAttributeName = "asp-fallback-src-exclude";
         private const string FallbackTestExpressionAttributeName = "asp-fallback-test";
         private const string SrcAttributeName = "src";
+        private const string FileVersionAttributeName = "asp-file-version";
+
+        private FileVersionProvider _fileVersionProvider;
 
         private static readonly ModeAttributes<Mode>[] ModeDetails = new[] {
+            // Regular src with file version alone
+            ModeAttributes.Create(Mode.FileVersion, new[] { FileVersionAttributeName }),
             // Globbed src (include only)
             ModeAttributes.Create(Mode.GlobbedSrc, new [] { SrcIncludeAttributeName }),
             // Globbed src (include & exclude)
@@ -67,14 +72,18 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
         private enum Mode
         {
             /// <summary>
+            /// Just adding a file version for the generated urls.
+            /// </summary>
+            FileVersion = 0,
+            /// <summary>
             /// Just performing file globbing search for the src, rendering a separate &lt;script&gt; for each match.
             /// </summary>
-            GlobbedSrc = 0,
+            GlobbedSrc = 1,
             /// <summary>
             /// Rendering a fallback block if primary javascript fails to load. Will also do globbing for both the
             /// primary and fallback srcs if the appropriate properties are set.
             /// </summary>
-            Fallback = 1
+            Fallback = 2
         }
 
         /// <summary>
@@ -97,6 +106,15 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
         /// </summary>
         [HtmlAttributeName(FallbackSrcAttributeName)]
         public string FallbackSrc { get; set; }
+
+        /// <summary>
+        /// Value indicating if file version should be appended to src urls.
+        /// </summary>
+        /// <remarks>
+        /// A query string "v" with the encoded content of the file is added.
+        /// </remarks>
+        [HtmlAttributeName(FileVersionAttributeName)]
+        public bool? FileVersion { get; set; }
 
         /// <summary>
         /// A comma separated list of globbed file patterns of JavaScript scripts to fallback to in the case the
@@ -137,11 +155,11 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
         [Activate]
         protected internal IMemoryCache Cache { get; set; }
 
-        // Internal for ease of use when testing.
-        protected internal GlobbingUrlBuilder GlobbingUrlBuilder { get; set; }
-
         [Activate]
         protected internal IHtmlEncoder HtmlEncoder { get; set; }
+
+        // Internal for ease of use when testing.
+        protected internal GlobbingUrlBuilder GlobbingUrlBuilder { get; set; }
 
         /// <inheritdoc />
         public override async Task ProcessAsync(TagHelperContext context, TagHelperOutput output)
@@ -167,9 +185,10 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
             var builder = new DefaultTagHelperContent();
             var originalContent = await context.GetChildContentAsync();
 
-            if (mode == Mode.Fallback && string.IsNullOrEmpty(SrcInclude))
+            if (mode == Mode.Fallback && string.IsNullOrEmpty(SrcInclude) || mode == Mode.FileVersion)
             {
                 // No globbing to do, just build a <script /> tag to match the original one in the source file
+                // Or just add file version to the script tag.
                 BuildScriptTag(originalContent, attributes, builder);
             }
             else
@@ -194,14 +213,14 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
         {
             // Build a <script> tag for each matched src as well as the original one in the source file
             string staticSrc;
-            attributes.TryGetValue("src", out staticSrc);
+            attributes.TryGetValue(SrcAttributeName, out staticSrc);
 
             EnsureGlobbingUrlBuilder();
             var urls = GlobbingUrlBuilder.BuildUrlList(staticSrc, SrcInclude, SrcExclude);
 
             foreach (var url in urls)
             {
-                attributes["src"] = HtmlEncoder.HtmlEncode(url);
+                attributes[SrcAttributeName] = url;
                 var content =
                     string.Equals(url, staticSrc, StringComparison.OrdinalIgnoreCase) ? originalContent : null;
                 BuildScriptTag(content, attributes, builder);
@@ -211,6 +230,7 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
         private void BuildFallbackBlock(IDictionary<string, string> attributes, DefaultTagHelperContent builder)
         {
             EnsureGlobbingUrlBuilder();
+            EnsureFileVersionProvider();
 
             var fallbackSrcs = GlobbingUrlBuilder.BuildUrlList(FallbackSrc, FallbackSrcInclude, FallbackSrcExclude);
 
@@ -226,9 +246,13 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
                 {
                     builder.Append("<script");
 
-                    if (!attributes.ContainsKey("src"))
+                    if (!attributes.ContainsKey(SrcAttributeName))
                     {
-                        AppendSrc(builder, "src", src);
+                        AppendAttribute(
+                            builder,
+                            SrcAttributeName, 
+                            HtmlEncoder.HtmlEncode(ShouldAddFileVersion() ? _fileVersionProvider.AddFileVersionToPath(src) : src),
+                            escapteQuotes: true);
                     }
 
                     foreach (var attribute in attributes)
@@ -238,21 +262,22 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
                             var encodedKey = JavaScriptStringEncoder.Default.JavaScriptStringEncode(attribute.Key);
                             var encodedValue = JavaScriptStringEncoder.Default.JavaScriptStringEncode(attribute.Value);
 
-                            builder.Append(string.Format(
-                                CultureInfo.InvariantCulture,
-                                " {0}=\\\"{1}\\\"",
-                                encodedKey,
-                                encodedValue));
+                            AppendAttribute(builder, encodedKey, encodedValue, escapteQuotes: true);
                         }
                         else
                         {
-                            AppendSrc(builder, attribute.Key, src);
+                            AppendAttribute(
+                                builder,
+                                attribute.Key,
+                                HtmlEncoder.HtmlEncode(
+                                    ShouldAddFileVersion() ? _fileVersionProvider.AddFileVersionToPath(src) : src),
+                                escapteQuotes: true);
                         }
                     }
 
                     builder.Append("><\\/script>");
                 }
-                
+
                 builder.Append("\"));</script>");
             }
         }
@@ -268,17 +293,37 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
             }
         }
 
-        private static void BuildScriptTag(
+        private void EnsureFileVersionProvider()
+        {
+            if (_fileVersionProvider == null)
+            {
+                _fileVersionProvider = new FileVersionProvider(
+                    HostingEnvironment.WebRootFileProvider,
+                    Cache,
+                    ViewContext.HttpContext.Request.PathBase);
+            }
+        }
+
+        private void BuildScriptTag(
             TagHelperContent content,
             IDictionary<string, string> attributes,
             TagHelperContent builder)
         {
+            EnsureFileVersionProvider();
             builder.Append("<script");
 
             foreach (var attribute in attributes)
             {
-                builder.Append(
-                    string.Format(CultureInfo.InvariantCulture, " {0}=\"{1}\"", attribute.Key, attribute.Value));
+                string attributeValue = attribute.Value;
+                if (string.Equals(attribute.Key, SrcAttributeName, StringComparison.OrdinalIgnoreCase))
+                {
+                    attributeValue = HtmlEncoder.HtmlEncode(
+                        ShouldAddFileVersion() ?
+                            _fileVersionProvider.AddFileVersionToPath(attribute.Value) :
+                            attributeValue);
+                }
+
+                AppendAttribute(builder, attribute.Key, attributeValue, escapteQuotes: false);
             }
 
             builder.Append(">")
@@ -286,15 +331,20 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
                    .Append("</script>");
         }
 
-        private void AppendSrc(TagHelperContent content, string srcKey, string srcValue)
+        private bool ShouldAddFileVersion()
         {
-            // Append src attribute in the original place and replace the content the fallback content
-            // No need to encode the key because we know it is "src".
-            content.Append(" ")
-                   .Append(srcKey)
-                   .Append("=\\\"")
-                   .Append(HtmlEncoder.HtmlEncode(srcValue))
-                   .Append("\\\"");
+            return FileVersion ?? false;
         }
+
+        private void AppendAttribute(TagHelperContent content, string key, string value, bool escapteQuotes)
+        {
+            content
+                .Append(" ")
+                .Append(key)
+                .Append(escapteQuotes ? "=\\\"" : "=\"")
+                .Append(value)
+                .Append(escapteQuotes ? "\\\"" : "\"");
+        }
+
     }
 }
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
@@ -27,6 +27,7 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
     [TargetElement("script", Attributes = FallbackSrcIncludeAttributeName)]
     [TargetElement("script", Attributes = FallbackSrcExcludeAttributeName)]
     [TargetElement("script", Attributes = FallbackTestExpressionAttributeName)]
+    [TargetElement("script", Attributes = FileVersionAttributeName)]
     public class ScriptTagHelper : TagHelper
     {
         private const string SrcIncludeAttributeName = "asp-src-include";
--- a/test/Microsoft.AspNet.Mvc.TagHelpers.Test/Internal/FileVersionProviderTest.cs
+++ b/test/Microsoft.AspNet.Mvc.TagHelpers.Test/Internal/FileVersionProviderTest.cs
@@ -96,17 +96,16 @@ namespace Microsoft.AspNet.Mvc.TagHelpers.Internal
         }
 
         [Theory]
-        [InlineData("/hello/world", "/hello/world")]
-        [InlineData("/testApp/hello/world", "/hello/world")]
-        [InlineData("~/testApp/hello/world", "/hello/world")]
-        [InlineData("/hello/world", "/hello/world")]
-        public void SetsValueInCache(string filePath, string watchPath)
+        [InlineData("/hello/world", "/hello/world", null)]
+        [InlineData("/testApp/hello/world", "/hello/world", "/testApp")]
+        [InlineData("/hello/world", "/hello/world", null)]
+        public void SetsValueInCache(string filePath, string watchPath, string requestPathBase)
         {
             // Arrange
             var trigger = new Mock<IExpirationTrigger>();
-            var hostingEnvironment = GetMockHostingEnvironment(filePath);
+            var hostingEnvironment = GetMockHostingEnvironment(filePath, requestPathBase != null);
             Mock.Get(hostingEnvironment.WebRootFileProvider)
-                .Setup(f => f.Watch(It.IsAny<string>())).Returns(trigger.Object);
+                .Setup(f => f.Watch(watchPath)).Returns(trigger.Object);
 
             object cacheValue = null;
             var cache = new Mock<IMemoryCache>();
@@ -130,7 +129,7 @@ namespace Microsoft.AspNet.Mvc.TagHelpers.Internal
             var fileVersionProvider = new FileVersionProvider(
                 hostingEnvironment.WebRootFileProvider,
                 cache.Object,
-                GetRequestPathBase());
+                GetRequestPathBase(requestPathBase));
 
             // Act
             var result = fileVersionProvider.AddFileVersionToPath(filePath);
--- a/test/Microsoft.AspNet.Mvc.TagHelpers.Test/Internal/FileVersionProviderTest.cs
+++ b/test/Microsoft.AspNet.Mvc.TagHelpers.Test/Internal/FileVersionProviderTest.cs
@@ -1,0 +1,215 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using System.Text;
+using Microsoft.AspNet.FileProviders;
+using Microsoft.AspNet.Hosting;
+using Microsoft.AspNet.Http;
+using Microsoft.Framework.Caching.Memory;
+using Microsoft.Framework.Expiration.Interfaces;
+using Moq;
+using Xunit;
+
+namespace Microsoft.AspNet.Mvc.TagHelpers.Internal
+{
+    public class FileVersionProviderTest
+    {
+        [Theory]
+        [InlineData("/hello/world", "/hello/world?v=f4OxZX_x_FO5LcGBSKHWXfwtSx-j1ncoSt3SABJtkGk")]
+        [InlineData("/hello/world?q=test", "/hello/world?q=test&v=f4OxZX_x_FO5LcGBSKHWXfwtSx-j1ncoSt3SABJtkGk")]
+        [InlineData("/hello/world?q=foo&bar", "/hello/world?q=foo&bar&v=f4OxZX_x_FO5LcGBSKHWXfwtSx-j1ncoSt3SABJtkGk")]
+        public void AddsVersionToFiles_WhenCacheIsAbsent(string filePath, string expected)
+        {
+            // Arrange
+            var hostingEnvironment = GetMockHostingEnvironment(filePath);
+            var fileVersionProvider = new FileVersionProvider(
+                hostingEnvironment.WebRootFileProvider,
+                GetMockCache(),
+                GetRequestPathBase());
+
+            // Act
+            var result = fileVersionProvider.AddFileVersionToPath(filePath);
+
+            // Assert
+            Assert.Equal(expected, result);
+        }
+
+        [Theory]
+        [InlineData("/testApp/hello/world", true, "/testApp")]
+        [InlineData("/testApp/foo/bar/hello/world", true, "/testApp/foo/bar")]
+        [InlineData("/test/testApp/hello/world", false, "/testApp")]
+        public void AddsVersionToFiles_PathContainingAppName(
+            string filePath,
+            bool pathStartsWithAppBase,
+            string requestPathBase)
+        {
+            // Arrange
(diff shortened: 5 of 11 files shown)


src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -98,6 +108,12 @@ private enum Mode
         public string FallbackHref { get; set; }
 
         /// <summary>
+        /// Value indicating if file version should be appended to the href urls.
+        /// </summary>
+        [HtmlAttributeName(FileVersionAttributeName)]
+        public string FileVersion { get; set; }
```
NTaylorMullen: Should add remarks stating what gets appended to the href urls.

NTaylorMullen: `bool`

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -183,7 +200,11 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
 
             var builder = new DefaultTagHelperContent();
 
-            if (mode == Mode.Fallback && string.IsNullOrEmpty(HrefInclude))
+            if (mode == Mode.FileVersion)
+            {
+                BuildLinkTag(attributes, builder);
+            }
+            else if (mode == Mode.Fallback && string.IsNullOrEmpty(HrefInclude))
```
NTaylorMullen: Can remove the first `if`:

```C#
if ((mode == Mode.Fallback && 
    string.IsNullOrEmpty(HrefInclude)) ||
    mode == Mode.FileVersion)
```

Will have to update the comment below.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -258,17 +288,80 @@ private void EnsureGlobbingUrlBuilder()
...
+                        string.Format(CultureInfo.InvariantCulture, "{0}=\"{1}\" ", attribute.Key, attribute.Value));
+                }
             }
 
             builder.Append("/>");
         }
+
+        private void AppendHref(TagHelperContent content, string hrefKey, string hrefValue)
+        {
+            string formatter = "{0}=\"{1}";
```
NTaylorMullen: `var`

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -258,17 +288,80 @@ private void EnsureGlobbingUrlBuilder()
...
+            string fileVersion = string.Empty;
```
NTaylorMullen: `var`

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -258,17 +288,80 @@ private void EnsureGlobbingUrlBuilder()
...
+            if (FileVersion == "true")
```
NTaylorMullen: `if (FileVersion)` once it's `bool`

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -258,17 +288,80 @@ private void EnsureGlobbingUrlBuilder()
...
+                    CultureInfo.InvariantCulture,
+                    formatter,
+                    hrefKey,
+                    HtmlEncoder.HtmlEncode(hrefValue),
+                    fileVersion));
+        }
+
+        private string GetFileContentEncodedString(string filePath)
+        {
+            var fileContents = ReadFileContents(HostingEnvironment.WebRootFileProvider.GetFileInfo(filePath));
```
NTaylorMullen: You don't need to get file contents. The hash can be computed from the file stream.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -258,17 +288,80 @@ private void EnsureGlobbingUrlBuilder()
...
+            if (fileContents != null)
+            {
+                using (var sha256 = SHA256.Create())
+                {
+                    var hash = sha256.ComputeHash(System.Text.Encoding.Unicode.GetBytes(fileContents));
```
NTaylorMullen:
```C#
private string GetFileContentEncodedString(string filePath)
{
    var fileInfo = HostingEnvironment.WebRootFileProvider.GetFileInfo(filePath);
    if (fileInfo.Exists)
    {
        using (var sha256 = SHA256.Create())
        {
            var hash = sha256.ComputeHash(fileInfo.CreateReadStream());
            return WebEncoders.Base64UrlEncode(hash);
        }
    }

    return string.Empty;
}
```

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -258,17 +288,80 @@ private void EnsureGlobbingUrlBuilder()
...
+                    return WebEncoders.Base64UrlEncode(hash);
+                }
+            }
+
+            return string.Empty;
+        }
+
+        private string ReadFileContents(IFileInfo fileInfo)
```
NTaylorMullen: Can remove this

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -258,17 +288,80 @@ private void EnsureGlobbingUrlBuilder()
...
+
+        private void AppendHref(TagHelperContent content, string hrefKey, string hrefValue)
+        {
+            string formatter = "{0}=\"{1}";
+            string fileVersion = string.Empty;
+            if (FileVersion == "true")
+            {
+                fileVersion = GetFileContentEncodedString(hrefValue);
+                // This is the case if there was any exception thrown when reading the file.
+                if (fileVersion != string.Empty)
```
NTaylorMullen: Just have the default case return `null`.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
sornaks: :+1:

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
NTaylorMullen: Or better: 
```C#
content.Append(
    string.Format(
        CultureInfo.InvariantCulture, 
        "{0}=\"{1}", 
        hrefKey, 
        HtmlEncoder.HtmlEncode(hrefValue)));
```

Left more comments below.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -258,17 +288,80 @@ private void EnsureGlobbingUrlBuilder()
...
+                {
+                    formatter += "?v={2}";
```
NTaylorMullen:
```C#
content
    .Append("?v=")
    .Append(fileVersion);
```

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -258,17 +288,80 @@ private void EnsureGlobbingUrlBuilder()
...
+                }
+            }
+
+            formatter += "\" ";
+
+            content.Append(
```
NTaylorMullen: delete

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -258,17 +288,80 @@ private void EnsureGlobbingUrlBuilder()
...
+            {
+                fileVersion = GetFileContentEncodedString(hrefValue);
+                // This is the case if there was any exception thrown when reading the file.
+                if (fileVersion != string.Empty)
+                {
+                    formatter += "?v={2}";
+                }
+            }
+
+            formatter += "\" ";
```
NTaylorMullen: `content.Append("\" ");`

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
NTaylorMullen: Can move declaration to where it's assigned below.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -258,17 +288,80 @@ private void EnsureGlobbingUrlBuilder()
...
-        private static void BuildLinkTag(IDictionary<string, string> attributes, TagHelperContent builder)
+        private void BuildLinkTag(IDictionary<string, string> attributes, TagHelperContent builder)
         {
             builder.Append("<link ");
 
             foreach (var attribute in attributes)
             {
-                builder.Append(
-                    string.Format(CultureInfo.InvariantCulture, "{0}=\"{1}\" ", attribute.Key, attribute.Value));
+                if (attribute.Key == "href")
```
NTaylorMullen: `if (string.Equals(attribute.Key, "href", StringComparison.OrdinalIgnoreCase))`

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -222,10 +243,19 @@ private void BuildGlobbedLinkTags(IDictionary<string, string> attributes, TagHel
         private void BuildFallbackBlock(TagHelperContent builder)
         {
             EnsureGlobbingUrlBuilder();
-            var fallbackHrefs = GlobbingUrlBuilder.BuildUrlList(FallbackHref, FallbackHrefInclude, FallbackHrefExclude);
+            var fallbackHrefs =
+                GlobbingUrlBuilder.BuildUrlList(FallbackHref, FallbackHrefInclude, FallbackHrefExclude).ToArray();
 
             if (fallbackHrefs.Any())
             {
+                if (FileVersion == "true")
```
NTaylorMullen: `if (FileVersion)` once it's `bool`

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -222,10 +243,19 @@ private void BuildGlobbedLinkTags(IDictionary<string, string> attributes, TagHel
         private void BuildFallbackBlock(TagHelperContent builder)
         {
             EnsureGlobbingUrlBuilder();
-            var fallbackHrefs = GlobbingUrlBuilder.BuildUrlList(FallbackHref, FallbackHrefInclude, FallbackHrefExclude);
+            var fallbackHrefs =
+                GlobbingUrlBuilder.BuildUrlList(FallbackHref, FallbackHrefInclude, FallbackHrefExclude).ToArray();
```
NTaylorMullen: Why `.ToArray()`?

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -258,17 +288,80 @@ private void EnsureGlobbingUrlBuilder()
...
+            }
+
+            formatter += "\" ";
+
+            content.Append(
+                string.Format(
+                    CultureInfo.InvariantCulture,
+                    formatter,
+                    hrefKey,
+                    HtmlEncoder.HtmlEncode(hrefValue),
```
NTaylorMullen: Why were you encoding before? We don't for other attributes (updated my snippet that takes care of the `hrefValue` above)

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -222,10 +243,19 @@ private void BuildGlobbedLinkTags(IDictionary<string, string> attributes, TagHel
...
+            var fallbackHrefs =
+                GlobbingUrlBuilder.BuildUrlList(FallbackHref, FallbackHrefInclude, FallbackHrefExclude).ToArray();
 
             if (fallbackHrefs.Any())
             {
+                if (FileVersion == "true")
+                {
+                    for (var i=0; i < fallbackHrefs.Count(); i++)
+                    {
+                        fallbackHrefs[i] += "?v=" + GetFileContentEncodedString(fallbackHrefs[i]);
```
NTaylorMullen: Hmm we need to unify this logic, we shouldn't be doing this in two places.

Add a helper method called something like `string ResolveVersionedHref(string href)`. We can then use this in the `AppendHref` method that I left snippets in. In those snippets you can modify them to write out the "resolved" href value `if (FileVersion)`.

Ultimately we want to try and avoid strings as much as possible but for this specific case we can make due for the sake of unifying logic.

Once the helper method is completed this should be:

```C#
if (FileVersion)
{
    for (var i = fallbackHrefs.Count() - 1; i--; i >= 0)
    {
        fallbackHrefs[i] = ResolveVersionedHref(fallbackHrefs[i]);
    }
}
```

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -258,17 +288,80 @@ private void EnsureGlobbingUrlBuilder()
...
         {
             builder.Append("<link ");
 
             foreach (var attribute in attributes)
             {
-                builder.Append(
-                    string.Format(CultureInfo.InvariantCulture, "{0}=\"{1}\" ", attribute.Key, attribute.Value));
+                if (attribute.Key == "href")
+                {
+                    AppendHref(builder, attribute.Key, attribute.Value);
```
NTaylorMullen: The more I look at this I see that it can be simplified even further, ignore my comments in the `AppendHref` method, it can be ignored.

This piece of the foreach can be as follows:

```C#
foreach (var attribute in attributes)
{
    var attributeValue = attribute.Value;
    if (string.Equals(attribute.Key, "href", StringComparison.OrdinalIgnoreCase) &&
        FileVersion)
    {
        attributeValue = ResolveVersionedHref(attributeValue);
    }

    builder.Append(
        string.Format(CultureInfo.InvariantCulture, "{0}=\"{1}\" ", attribute.Key, attributeValue));
}
```

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
NTaylorMullen: Ignore all my comments on `AppendHref`, found a better way: https://github.com/aspnet/Mvc/pull/2225/files#r26881513

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
NTaylorMullen: Also we shouldn't be using `?v=`, it should be `#v=`

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -272,8 +293,15 @@ private void EnsureGlobbingUrlBuilder()
 
             foreach (var attribute in attributes)
             {
-                builder.Append(
-                    string.Format(CultureInfo.InvariantCulture, " {0}=\"{1}\"", attribute.Key, attribute.Value));
+                if (attribute.Key.Equals(SrcAttributeName, StringComparison.OrdinalIgnoreCase))
+                {
+                    AppendSrc(builder, attribute.Key, attribute.Value, escapeQuotes: false);
```
NTaylorMullen: Gonna wait to look at this file, apply the comments in `LinkTagHelper` to `ScriptTagHelper` as well and i'll look over them in the next iteration.

NTaylorMullen:
Will re-visit tests in next iteration after:

- File version defaults to `true` based on issue requirement double checked with @DamianEdwards that this is the intended functionality.
- File version url bit should be `?v=` OR `&v=` instead of always `?v=`  Should add tests to validate this too

NTaylorMullen:
Talked with @DamianEdwards about caching and he said we should definitely be caching the disk reads.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
dougbu: also should have a constant for `"href"` and use it here and elsewhere in the file

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
sornaks: I am not sure what I'm missing, but only string attributes are getting set. If I change it to bool, it never gets set.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
sornaks: Hmmm, we should also handle URLs with `#` in them. Like `foo/site#test` should become `foo/site?v=some-code#test`.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
NTaylorMullen: What you're most likely seeing is due to pre-compilation essentially caching your page (assuming you're using pre-compilation, it's on by default in TagHelperSample.Web), just tried this on my box and it worked as expected but I ran into a pre-compilation caching issue because of it. 

/cc @DamianEdwards

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
sornaks: Hmmm, I thought this was more readable since we have different sections for different modes..

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
sornaks: Because I want to persist the changes of individual items in the enumerable..

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
NTaylorMullen: ehh it's duplicate code, made me think twice as to why it was different

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
NTaylorMullen: Yup, good call.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
NTaylorMullen: :+1:

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
DamianEdwards: Thought about this some more and decided it should be `false` by default actually.

Also, this has to be a `bool`

dougbu: @DamianEdwards might be more HTML-ish if we converted it to a bool using `var addFileVersion = FileVersion != null;`.  is that better or worse?

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -138,7 +154,8 @@ private enum Mode
         [HtmlAttributeName(FallbackTestValueAttributeName)]
         public string FallbackTestValue { get; set; }
 
-        // Properties are protected to ensure subclasses are correctly activated. Internal for ease of use when testing.
+        // Properties are protected to ensure subclasses are correctly activated.
+        // Internal for ease of use when testing.
```
dougbu: nit: why did this change? don't churn for churn's sake or due to a personal preference

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -214,7 +235,7 @@ private void BuildGlobbedLinkTags(IDictionary<string, string> attributes, TagHel
 
             foreach (var url in urls)
             {
-                attributes["href"] = HtmlEncoder.HtmlEncode(url);
+                attributes["href"] = url;
```
DamianEdwards: Why is this no longer encoded?

dougbu: why did this change?  if URLs are getting double-encoded, submit a quick fix and add a test for that -- separately

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
DamianEdwards: Why would pre-compilation change bound property values? It shouldn't change anything to do with execution of the page.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -258,17 +288,80 @@ private void EnsureGlobbingUrlBuilder()
...
-                builder.Append(
-                    string.Format(CultureInfo.InvariantCulture, "{0}=\"{1}\" ", attribute.Key, attribute.Value));
+                if (attribute.Key == "href")
+                {
+                    AppendHref(builder, attribute.Key, attribute.Value);
+                }
+                else
+                {
+                    builder.Append(
+                        string.Format(CultureInfo.InvariantCulture, "{0}=\"{1}\" ", attribute.Key, attribute.Value));
```
dougbu: why create a `string` here?  would be more straightforward as a chained set of `Append()` calls

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
DamianEdwards: Think of the modes as primarily for detecting whether the helper will run or not. The modes overlap and combine when matched, so splitting the actual logic based on matched modes isn't ideal. Basically, what @NTaylorMullen said :smile:

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -281,15 +309,62 @@ private void EnsureGlobbingUrlBuilder()
                    .Append("</script>");
         }
 
-        private void AppendSrc(TagHelperContent content, string srcKey, string srcValue)
+        private void AppendSrc(TagHelperContent content, string srcKey, string srcValue, bool escapeQuotes)
```
dougbu: what does not escaping quotes for the `src` attribute value have to do with adding the file version?  fine to do a couple of small unrelated changes while you have the file open but this PR seems to mix at least three user-visible changes.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
DamianEdwards: We need to cache generation of the hash and invalidate it based on a file changed trigger. Code is all there in the `GlobbingUrlBuilder` class.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
DamianEdwards: @dougbu much worse. This is a Tag Helper property, it should be typed as such.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -222,10 +243,19 @@ private void BuildGlobbedLinkTags(IDictionary<string, string> attributes, TagHel
...
             EnsureGlobbingUrlBuilder();
-            var fallbackHrefs = GlobbingUrlBuilder.BuildUrlList(FallbackHref, FallbackHrefInclude, FallbackHrefExclude);
+            var fallbackHrefs =
+                GlobbingUrlBuilder.BuildUrlList(FallbackHref, FallbackHrefInclude, FallbackHrefExclude).ToArray();
 
             if (fallbackHrefs.Any())
             {
+                if (FileVersion == "true")
+                {
+                    for (var i=0; i < fallbackHrefs.Count(); i++)
```
DamianEdwards: `fallbackHrefs.Length` as you made it an array. Calling `.Count()` on each iteration should be a red flag.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
NTaylorMullen: It doesn't change the bound properties, it just gives the appearance that things aren't binding because it's not recompiling the pages (they're cached). Going to look into our caching habbits around precompilation today

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
DamianEdwards: This needs to handle and no-op for paths that aren't in the site too, e.g. http://some-cdn.com/boostrap/bootstrap.css.
We can't create file versions for files we can't properly map back to readable files on the server.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
DamianEdwards: You need to handle cases where the `filePath` is prefixed with the app path base too, e.g. site is being hosted in `/myapp` and the path passed is `/myapp/css/site.css`. The `WebRootFileProvider` won't know about that path so you'll have to trim the path base from the beginning first.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
DamianEdwards: We shouldn't be doing manual URL building logic here if possible. See if the types in `Microsoft.AspNet.Http` help with this logic, e.g. `PathString`, `QueryString`, `FragmentString`. Look at `Microsoft.AspNet.Http.Extensions.UriHelper` to help with splitting the URL up into its typed parts. It doesn't look like we have the functionality in there to add an additional query value and then emit the full URL again though, so you may have to add that to `Microsoft.AspNet.WebUtilities.QueryHelpers` or similar. Check with @loudej

sornaks:
Updated the PR. Have some tests to add. Will do that tomorrow.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
```diff
@@ -0,0 +1,103 @@
...
+using Microsoft.Framework.Runtime;
+
+namespace Microsoft.AspNet.Mvc.TagHelpers.Internal
+{
+    /// <summary>
+    /// Provides version hash for a specified file.
+    /// </summary>
+    public class FileVersionProvider
+    {
+        private const string versionKey = "v";
```
NTaylorMullen: consts should be pascal cased, even when private.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
```diff
@@ -88,7 +88,7 @@ public static class AttributeMatcher
             {
                 if (!context.AllAttributes.ContainsKey(attribute) ||
                     context.AllAttributes[attribute] == null ||
-                    string.IsNullOrWhiteSpace(context.AllAttributes[attribute] as string))
+                    string.IsNullOrWhiteSpace(context.AllAttributes[attribute].ToString()))
```
NTaylorMullen: ?

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
```diff
@@ -0,0 +1,103 @@
...
+
+using System;
+using System.Security.Cryptography;
+using Microsoft.AspNet.FileProviders;
+using Microsoft.AspNet.Hosting;
+using Microsoft.AspNet.WebUtilities;
+using Microsoft.Framework.Caching.Memory;
+using Microsoft.Framework.Runtime;
+
+namespace Microsoft.AspNet.Mvc.TagHelpers.Internal
```
NTaylorMullen: Note: For internal classes you don't need to worry about documenting them.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
```diff
@@ -0,0 +1,103 @@
...
+        private const string versionKey = "v";
+
+        /// <summary>
+        /// Creates a new instance of <see cref="FileVersionProvider"/>.
+        /// </summary>
+        /// <param name="fileProvider">The file provider to get and watch files.</param>
+        /// <param name="applicationName">Name of the applicaiton.</param>
+        /// <param name="cache">Cache where versioned urls of files are cached.</param>
+        public FileVersionProvider(
+            IFileProvider fileProvider,
```
NTaylorMullen: `[NotNull]` for this and `applicationName`

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
```diff
@@ -0,0 +1,103 @@
...
+
+        /// <summary>
+        /// Adds version query parameter to the specified file path.
+        /// </summary>
+        /// <param name="filePath">The path of the file to which version should be added.</param>
+        /// <returns>Path containing the version query string.</returns>
+        /// <remarks>
+        /// The version query string is appended as with the key "v".
+        /// </remarks>
+        public string AddVersionToFilePath(string filePath)
```
NTaylorMullen: "FilePath" is somewhat misleading, better: `AddFileVersionToPath`.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
```diff
@@ -0,0 +1,103 @@
...
+        {
+            var fileInfo = FileProvider.GetFileInfo(filePath);
+            if (!fileInfo.Exists)
+            {
+                if (filePath.Contains(ApplicationName))
```
NTaylorMullen: You'll have to do a little bit more. 

- `Contains` covers `/` and `~/` starting cases but reallly leaves open situations when the `ApplicationName` is just randomly used in the URL (it can have false-positives which is realllly bad): Use `StartsWith`/`Substring`. This means you'll also need to strip out beginning `/` or `~/`.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
```diff
@@ -0,0 +1,103 @@
...
+        {
+            FileProvider = fileProvider;
+            ApplicationName = applicationName;
+            Cache = cache;
+        }
+
+        /// <summary>
+        /// The <see cref="IFileProvider"/> to get and watch files.
+        /// </summary>
+        public IFileProvider FileProvider { get; }
```
NTaylorMullen: Any reason all these properties can't be `private readonly` fields?

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
```diff
@@ -0,0 +1,103 @@
...
+            }
+
+            if (Cache != null)
+            {
+                return Cache.GetOrSet(filePath, cacheGetOrSetContext =>
+                {
+                    var trigger = FileProvider.Watch(filePath);
+                    cacheGetOrSetContext.AddExpirationTrigger(trigger);
+
+                    return Cache.Set(
```
NTaylorMullen: Can't you just return: `QueryHelpers.AddQueryString(filePath, versionKey, GetHashForFile(fileInfo, filePath))`? From my understanding returning will perform the set for you

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -98,6 +106,15 @@ private enum Mode
         public string FallbackHref { get; set; }
 
         /// <summary>
+        /// Value indicating if file version should be appended to the href urls.
+        /// </summary>
+        /// <remarks>
+        /// A query string "v" with the encoded content of the file is added.
+        /// </remarks>
+        [HtmlAttributeName(FileVersionAttributeName)]
+        public bool FileVersion { get; set; } = false;
```
NTaylorMullen: no need to set here, it'll default to `false` if it's not set.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -159,6 +177,11 @@ private enum Mode
 
         // Internal for ease of use when testing.
         protected internal GlobbingUrlBuilder GlobbingUrlBuilder { get; set; }
+        
+        protected FileVersionProvider FileVersionProvider { get; set; }
```
NTaylorMullen: `// Internal for ease of use when testing.`

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -159,6 +177,11 @@ private enum Mode
...
+
+        [Activate]
+        protected internal IApplicationEnvironment ApplicationEnvironment { get; set; }
```
NTaylorMullen: Move this activated property up with the other activated properties.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -222,10 +246,19 @@ private void BuildGlobbedLinkTags(IDictionary<string, string> attributes, TagHel
...
             EnsureGlobbingUrlBuilder();
-            var fallbackHrefs = GlobbingUrlBuilder.BuildUrlList(FallbackHref, FallbackHrefInclude, FallbackHrefExclude);
+            var fallbackHrefs =
+                GlobbingUrlBuilder.BuildUrlList(FallbackHref, FallbackHrefInclude, FallbackHrefExclude).ToArray();
 
             if (fallbackHrefs.Any())
             {
+                if (FileVersion)
+                {
+                    for (var i=0; i < fallbackHrefs.Count(); i++)
```
NTaylorMullen: `.Length`

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -257,15 +290,39 @@ private void EnsureGlobbingUrlBuilder()
                     ViewContext.HttpContext.Request.PathBase);
             }
         }
+        private void EnsureFileVersionProvider()
```
NTaylorMullen: nit: add line inbetween `}` before this

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -257,15 +290,39 @@ private void EnsureGlobbingUrlBuilder()
...
+        private void BuildLinkTag(IDictionary<string, string> attributes, TagHelperContent builder)
         {
+            EnsureFileVersionProvider();
             builder.Append("<link ");
 
             foreach (var attribute in attributes)
             {
-                builder.Append(
-                    string.Format(CultureInfo.InvariantCulture, "{0}=\"{1}\" ", attribute.Key, attribute.Value));
+                var versionAttributeValue = attribute.Value;
```
NTaylorMullen: Most of the time it wont be an `href`, just name this `attributeValue`

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -94,6 +102,15 @@ private enum Mode
         public string FallbackSrc { get; set; }
 
         /// <summary>
+        /// Value indicating if file version should be appended to src urls.
+        /// </summary>
+        /// <remarks>
+        /// A query string "v" with the encoded content of the file is added.
+        /// </remarks>
+        [HtmlAttributeName(FileVersionAttributeName)]
+        public bool FileVersion { get; set; } = false;
```
NTaylorMullen: no need to force it to `false` here, it'll default to `false` if it's not set.

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -221,9 +245,16 @@ private void BuildFallbackBlock(IDictionary<string, string> attributes, DefaultT
...
 
-                    if (!attributes.ContainsKey("src"))
+                    if (!attributes.ContainsKey(SrcAttributeName))
                     {
-                        AppendSrc(builder, "src", src);
+                        builder
+                            .Append(" ")
+                            .Append(SrcAttributeName)
+                            .Append("=")
+                            .Append("\\\"")
```
NTaylorMullen: This can be a part of the previous append

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -221,9 +245,16 @@ private void BuildFallbackBlock(IDictionary<string, string> attributes, DefaultT
                 {
                     builder.Append("<script");
 
-                    if (!attributes.ContainsKey("src"))
+                    if (!attributes.ContainsKey(SrcAttributeName))
                     {
-                        AppendSrc(builder, "src", src);
```
NTaylorMullen: Why not still use helper and have it do all this magic?

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
NTaylorMullen: Can go in previous append.

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -233,15 +264,24 @@ private void BuildFallbackBlock(IDictionary<string, string> attributes, DefaultT
...
+                                .Append(" ")
+                                .Append(encodedKey)
+                                .Append("=")
+                                .Append("\\\"")
+                                .Append(encodedValue)
+                                .Append("\\\"");
                         }
                         else
                         {
-                            AppendSrc(builder, attribute.Key, src);
```
NTaylorMullen: Ditto comments as above

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -233,15 +264,24 @@ private void BuildFallbackBlock(IDictionary<string, string> attributes, DefaultT
...
-                            builder.Append(string.Format(
-                                CultureInfo.InvariantCulture,
-                                " {0}=\\\"{1}\\\"",
-                                encodedKey,
-                                encodedValue));
+                            builder
+                                .Append(" ")
+                                .Append(encodedKey)
+                                .Append("=")
+                                .Append("\\\"")
```
NTaylorMullen: Can go in append above.

NTaylorMullen:
:watch: Didn't look at tests, waiting for you to complete them.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
sornaks: Now that we have a bool html attribute, `as string` would return `null` whereas `ToString()` will return `true` or `false`.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
sornaks: No. I followed the pattern in `GlobbingUrlBuilder`.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
sornaks: No. In fact it can be `private`.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
dougbu: don't have a `bool` _HTML_ attribute but do have a bound attribute that cannot be checked here because it is not `Nullable<T>`.  must either make the property `bool?` or treat the corresponding mode as having no required attributes.  @DamianEdwards ?

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -31,12 +32,15 @@ public class ScriptTagHelper : TagHelper
...
+        private const string FileVersionAttributeName = "asp-file-version";
 
         private static readonly ModeAttributes<Mode>[] ModeDetails = new[] {
+            // Regular src with file version alone
+            ModeAttributes.Create(Mode.FileVersion, new[] { FileVersionAttributeName }),
             // Globbed src (include only)
             ModeAttributes.Create(Mode.GlobbedSrc, new [] { SrcIncludeAttributeName }),
             // Globbed src (include & exclude)
-            ModeAttributes.Create(Mode.GlobbedSrc, new [] { SrcIncludeAttributeName, SrcExcludeAttributeName }),
+            ModeAttributes.Create(Mode.GlobbedSrc, new [] { SrcIncludeAttributeName, SrcExcludeAttributeName}),
```
dougbu: nit: what changed here? why?

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
sornaks: Even if it is `Nullable<T>` we would still need this change. And we should not modify the mode to have no required attributes because we would end up executing the matching code-path even for simple script sources like `<script src="test.js"></script>`

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
```diff
@@ -88,7 +88,8 @@ public static class AttributeMatcher
             {
                 if (!context.AllAttributes.ContainsKey(attribute) ||
                     context.AllAttributes[attribute] == null ||
-                    string.IsNullOrWhiteSpace(context.AllAttributes[attribute] as string))
+                    (typeof(string).IsAssignableFrom(context.AllAttributes[attribute].GetType()) &&
+                    string.IsNullOrWhiteSpace(context.AllAttributes[attribute] as string)))
```
dougbu: this will hit an NRE if `context.AllAttributes[attribute]` is `null`!

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
```diff
@@ -0,0 +1,88 @@
...
+        /// <returns>Path containing the version query string.</returns>
+        /// <remarks>
+        /// The version query string is appended as with the key "v".
+        /// </remarks>
+        public string AddFileVersionToPath(string path)
+        {
+            var fileInfo = _fileProvider.GetFileInfo(path);
+            if (!fileInfo.Exists)
+            {
+                if (path.StartsWith("/" + _applicationName) ||
```
NTaylorMullen: Specify comparer for both of these.

NTaylorMullen: What happens if it just starts with `_applicationName`?

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
sornaks: That case would be short circuited.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
```diff
@@ -0,0 +1,88 @@
...
+                    path.StartsWith("~/" + _applicationName))
+                {
+                    fileInfo = _fileProvider.GetFileInfo(path.Split(
```
NTaylorMullen: Split is really wasteful, do an approach like this: 

https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Razor.Host/ViewHierarchyUtility.cs#L61-L69

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -66,16 +70,22 @@ public class LinkTagHelper : TagHelper
...
+            GlobbedHref = 1,
             /// <summary>
             /// Rendering a fallback block if primary stylesheet fails to load. Will also do globbing for both the
             /// primary and fallback hrefs if the appropriate properties are set.
             /// </summary>
-            Fallback = 1,
+            Fallback = 2,
         }
 
+        private FileVersionProvider FileVersionProvider { get; set; }
```
NTaylorMullen: `private readonly`

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -62,16 +65,22 @@ public class ScriptTagHelper : TagHelper
...
+            GlobbedSrc = 1,
             /// <summary>
             /// Rendering a fallback block if primary javascript fails to load. Will also do globbing for both the
             /// primary and fallback srcs if the appropriate properties are set.
             /// </summary>
-            Fallback = 1
+            Fallback = 2
         }
 
+        private FileVersionProvider FileVersionProvider { get; set; }
```
NTaylorMullen: ditto as in linktaghelper

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
sornaks: Haven't made this a `readonly` field cos I am not assigning this in the ctor.. Should I add a ctor and initialize it there?

test/Microsoft.AspNet.Mvc.TagHelpers.Test/Internal/FileVersionProviderTest.cs
```diff
@@ -0,0 +1,168 @@
...
+
+            // Act
+            var result = fileVersionProvider.AddFileVersionToPath(filePath);
+
+            // Assert
+            Assert.Equal("/hello/world?v=f4OxZX_x_FO5LcGBSKHWXfwtSx-j1ncoSt3SABJtkGk", result);
+        }
+
+        [Fact]
+        public void AddsVersionToFiles_WhenAppNameIsInUrl()
```
NTaylorMullen: `WhenAppNameIsInUrl` => `WhenUrlStartsWithAppName`. 

Need to add tests for
- `/appname`
- `~/appname`
- `appname
- `foo/appname`
- `/foo/appname`
- `~/foo/appname`

test/Microsoft.AspNet.Mvc.TagHelpers.Test/Internal/FileVersionProviderTest.cs
```diff
@@ -0,0 +1,168 @@
...
+using Microsoft.Framework.Caching.Memory;
+using Microsoft.Framework.Expiration.Interfaces;
+using Moq;
+using Xunit;
+
+namespace Microsoft.AspNet.Mvc.TagHelpers.Internal
+{
+    public class FileVersionProviderTest
+    {
+        [Fact]
```
NTaylorMullen: Also add tests that cover urls with `#` `?` and `&` and mixtures.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
NTaylorMullen: `private` only then

test/Microsoft.AspNet.Mvc.TagHelpers.Test/Internal/FileVersionProviderTest.cs
sornaks: The `QueryHelpers` does not handle `#` for now. https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNet.WebUtilities/QueryHelpers.cs#L21
I could file a follow up and fix this. @DamianEdwards what do you think?

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
dougbu: :ok:
so the intent here is to treat _any_ value for non-`string` properties as provided?  I guess that makes sense for `bool?` but I'm wondering if it's the correct behaviour in general.  not sure about this...

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
```diff
@@ -0,0 +1,88 @@
...
+        private const string VersionKey = "v";
+        private readonly IFileProvider _fileProvider;
+        private readonly string _applicationName;
+        private readonly IMemoryCache _cache;
+
+        /// <summary>
+        /// Creates a new instance of <see cref="FileVersionProvider"/>.
+        /// </summary>
+        /// <param name="fileProvider">The file provider to get and watch files.</param>
+        /// <param name="applicationName">Name of the applicaiton.</param>
```
dougbu: typo in "applicaiton"

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
```diff
@@ -0,0 +1,88 @@
...
+        /// <param name="cache">Cache where versioned urls of files are cached.</param>
```
dougbu: nit: "Cache" -> `<see cref="IMemoryCache"/>`

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
```diff
@@ -0,0 +1,88 @@
...
+        public FileVersionProvider(
+            [NotNull] IFileProvider fileProvider,
+            [NotNull] string applicationName,
+            IMemoryCache cache)
```
dougbu: why is an instance of this class instantiated when caller does not have a `IMemoryCache` instance?  what is the scenario when file hashes should be generated but not cached?

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
dougbu: nit: can't see how the property (and its generated methods) adds anything over a `private` field

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
sornaks: When you are unable to activate the `IMemoryCache` for some reason?

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -98,6 +108,15 @@ private enum Mode
         public string FallbackHref { get; set; }
 
         /// <summary>
+        /// Value indicating if file version should be appended to the href urls.
+        /// </summary>
+        /// <remarks>
+        /// A query string "v" with the encoded content of the file is added.
```
dougbu: `If <c>true</c> then a query string ...`

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -207,14 +231,14 @@ private void BuildGlobbedLinkTags(IDictionary<string, string> attributes, TagHel
...
-            attributes.TryGetValue("href", out staticHref);
+            attributes.TryGetValue(HrefAttributeName, out staticHref);
 
             EnsureGlobbingUrlBuilder();
             var urls = GlobbingUrlBuilder.BuildUrlList(staticHref, HrefInclude, HrefExclude);
 
             foreach (var url in urls)
             {
-                attributes["href"] = HtmlEncoder.HtmlEncode(url);
+                attributes[HrefAttributeName] = url;
```
dougbu: again, why is this no longer HTML encoded here?

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -222,10 +246,19 @@ private void BuildGlobbedLinkTags(IDictionary<string, string> attributes, TagHel
         private void BuildFallbackBlock(TagHelperContent builder)
         {
             EnsureGlobbingUrlBuilder();
-            var fallbackHrefs = GlobbingUrlBuilder.BuildUrlList(FallbackHref, FallbackHrefInclude, FallbackHrefExclude);
+            var fallbackHrefs =
+                GlobbingUrlBuilder.BuildUrlList(FallbackHref, FallbackHrefInclude, FallbackHrefExclude).ToArray();
 
             if (fallbackHrefs.Any())
```
dougbu: have an `Array`, use its `Length` and not Linq

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
sornaks: Because it is taken care in `BuildLinkTag`

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -66,16 +70,22 @@ public class LinkTagHelper : TagHelper
         private enum Mode
         {
             /// <summary>
+            /// Just adding a file version for the src.
```
dougbu: comment is misleading because you're adding file versions to all generated URLs

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -132,12 +150,15 @@ private enum Mode
...
-        // Internal for ease of use when testing.
-        protected internal GlobbingUrlBuilder GlobbingUrlBuilder { get; set; }
-
         [Activate]
         protected internal IHtmlEncoder HtmlEncoder { get; set; }
 
+        [Activate]
+        protected internal IApplicationEnvironment ApplicationEnvironment { get; set; }
+
+        // Internal for ease of use when testing.
```
dougbu: why is this property `protected`?

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -221,9 +244,13 @@ private void BuildFallbackBlock(IDictionary<string, string> attributes, DefaultT
...
                     builder.Append("<script");
 
-                    if (!attributes.ContainsKey("src"))
+                    if (!attributes.ContainsKey(SrcAttributeName))
                     {
-                        AppendSrc(builder, "src", src);
+                        AppendAttribute(
+                            builder,
+                            SrcAttributeName, 
+                            HtmlEncoder.HtmlEncode(ShouldAddFileVersion() ? FileVersionProvider.AddFileVersionToPath(src) : src),
```
dougbu: nit: wrap

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -281,15 +329,20 @@ private void EnsureGlobbingUrlBuilder()
...
+        private bool ShouldAddFileVersion()
         {
-            // Append src attribute in the original place and replace the content the fallback content
-            // No need to encode the key because we know it is "src".
-            content.Append(" ")
-                   .Append(srcKey)
-                   .Append("=\\\"")
-                   .Append(HtmlEncoder.HtmlEncode(srcValue))
-                   .Append("\\\"");
+            return FileVersion ?? false;
```
dougbu: nit: would be clearer to inline this expression the few places where this method is called

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
dougbu: activation doesn't fail without everything failing.

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
sornaks: I tried that. It gets more confusing when you try to use this along with ternary operators. Hence I separated this. I think this is a lot more readable.

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
sornaks: It had 121 characters.

sornaks:
Updated.. Please have a look.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -66,14 +72,18 @@ public class LinkTagHelper : TagHelper
         private enum Mode
         {
             /// <summary>
+            /// Adding file version to the generated urls.
```
dougbu: previous description was more correct i.e. use "Just adding a file version for the src." because modes overlap and could add file versions in other modes.

dougbu: my comment here was that you were changing more than "src" URLs, not that the other bit was misleading.

dougbu:
:shipit: when @NTaylorMullen is happy

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        /// <summary>
+        /// Adds version query parameter to the specified file path.
+        /// </summary>
+        /// <param name="path">The path of the file to which version should be added.</param>
+        /// <returns>Path containing the version query string.</returns>
+        /// <remarks>
+        /// The version query string is appended as with the key "v".
+        /// </remarks>
+        public string AddFileVersionToPath(string path)
```
NTaylorMullen: `[NotNull]`

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
```diff
@@ -0,0 +1,92 @@
...
+        {
+            var resolvedPath = path;
+            var fileInfo = _fileProvider.GetFileInfo(path);
+            if (!fileInfo.Exists)
+            {
+                if (resolvedPath.StartsWith("/" + _applicationName, StringComparison.OrdinalIgnoreCase))
```
NTaylorMullen: Instead of doing string concats, just do a `StartsWIth` for `/` and `substring(1)`. Can then use `Ordinal` for comparer.

sornaks: That would also make `/fooController/bar' to `fooController/bar` or `~/fooController/bar' to `fooController/bar` which is not really necessary..

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
```diff
@@ -0,0 +1,92 @@
...
+                {
+                    resolvedPath = resolvedPath.Substring(_applicationName.Length + 1);
+                }
+                else if (resolvedPath.StartsWith("~/" + _applicationName, StringComparison.OrdinalIgnoreCase))
+                {
+                    resolvedPath = resolvedPath.Substring(_applicationName.Length + 2);
+                }
+                else if (resolvedPath.StartsWith(_applicationName, StringComparison.OrdinalIgnoreCase))
```
NTaylorMullen: This can then just be an `if`.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
NTaylorMullen: Hmm I see, alright leave it.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
NTaylorMullen: ignore

NTaylorMullen:
It looks good, just want to check on one thing with @DamianEdwards before we get it in.

NTaylorMullen:
:shipit: we'll grab damian later.

NTaylorMullen:
Add two tests that flow the request app base from `LinkTagHelper`/`ScriptTagHelper` all the way through to the version provider. :watch: Almost done.

NTaylorMullen:
:shipit:

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -233,10 +255,19 @@ private void BuildGlobbedLinkTags(IDictionary<string, string> attributes, TagHel
...
-            var fallbackHrefs = GlobbingUrlBuilder.BuildUrlList(FallbackHref, FallbackHrefInclude, FallbackHrefExclude);
+            var fallbackHrefs =
+                GlobbingUrlBuilder.BuildUrlList(FallbackHref, FallbackHrefInclude, FallbackHrefExclude).ToArray();
 
-            if (fallbackHrefs.Any())
+            if (fallbackHrefs.Length > 0)
             {
+                if (ShouldAddFileVersion())
+                {
+                    for (var i=0; i < fallbackHrefs.Length; i++)
```
pranavkm: Formatting.

sornaks:
Thanks guys! https://github.com/aspnet/Mvc/commit/0e783ace5898c761435541aa937085a8593a75d0

Issue #2179, opened 2015-03-15
Error with complex Layouts and Sections.


AerisG222:
### Functional impact
The view fails with the following error:

InvalidOperationException: The following sections have been defined but have not been rendered: 'scripts_footer'.

### Minimal repro steps
I have tried to create a small sample that builds in complexity, from working views to a minimal test that fails.  This sample is available here:  [https://github.com/AerisG222/nested_layout_test/](https://github.com/AerisG222/nested_layout_test/).

The README on the repo contains steps showing basic layouts working, then the one that fails.

What appears to be causing the issue is where sections are defined across the layouts.  In the case that is failing, the layout page adds new sections that are referenced by the view, but are not in the parent layouts.  This seems to interfere with also being able to target the sections that are defined on the parent layout.

### Expected result
I was expecting all views to display resulting html w/o the exception

### Actual result
View fails with the exception from above

### Technical Details
Note that I am trying this from Linux, using a mono nightly build from 20150311212449, and the dev aspnet feed (with dnx).   here are version numbers in case this is helpful:

dnx --version
1.0.0-beta4-11271

dnvm list

Active Version              Runtime Location             Alias
------ -------              ------- --------             -----
  *    1.0.0-beta4-11271    mono    ~/.dnx/runtimes      default

mono --version
Mono JIT compiler version 4.1.0 (tarball Wed Mar 11 17:55:29 EDT 2015)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           __thread
	SIGSEGV:       altstack
	Notifications: epoll
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	LLVM:          supported, not enabled.
	GC:            sgen

Thanks,
Mike

danroth27:
@dougbu Please investigate what the impact of this issue is. Thx.

PR #2226, opened 2015-03-20
Fix for supporting nested sections in layouts

ajaybhargavb:
Issue #2179 
@rynowak @pranavkm

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
@@ -2,7 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Mvc.Rendering;
 using Microsoft.AspNet.PageExecutionInstrumentation;
@@ -168,6 +170,8 @@ namespace Microsoft.AspNet.Mvc.Razor
             // A layout page can specify another layout page. We'll need to continue
             // looking for layout pages until they're no longer specified.
             var previousPage = RazorPage;
+            var unrenderedSections = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+
             while (!string.IsNullOrEmpty(previousPage.Layout))
             {
                 if (!bodyWriter.IsBuffering)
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
```diff
@@ -190,12 +194,23 @@ private bool EnableInstrumentation
                 layoutPage.RenderBodyDelegate = bodyWriter.CopyTo;
                 bodyWriter = await RenderPageAsync(layoutPage, context, executeViewStart: false);
 
-                // Verify that RenderBody is called, or that RenderSection is called for all sections
-                layoutPage.EnsureBodyAndSectionsWereRendered();
+                // Verify that RenderBody is called
+                layoutPage.EnsureBodyWasRendered();
+
+                unRenderedSections.ExceptWith(layoutPage.RenderedSections);
```
rynowak: `unRenderedSections` -> `unrenderedSections`

src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
```diff
@@ -190,12 +194,23 @@ private bool EnableInstrumentation
...
+                unRenderedSections.UnionWith(layoutPage.PreviousSectionWriters.Keys.Except(layoutPage.RenderedSections,
```
rynowak: If you reverse the order of these operations I don't think you'd have to do `Except` twice

rynowak:
:shipit:

ajaybhargavb:
Updated with tests.

src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
```diff
@@ -190,12 +194,22 @@ private bool EnableInstrumentation
                 layoutPage.RenderBodyDelegate = bodyWriter.CopyTo;
                 bodyWriter = await RenderPageAsync(layoutPage, context, executeViewStart: false);
 
-                // Verify that RenderBody is called, or that RenderSection is called for all sections
-                layoutPage.EnsureBodyAndSectionsWereRendered();
+                // Verify that RenderBody is called
+                layoutPage.EnsureBodyWasRendered();
+
+                unrenderedSections.UnionWith(layoutPage.PreviousSectionWriters.Keys);
```
pranavkm: This might not address sections that are redefined but not rendered in the outer definition.

```C#
// NestedLayout
@{ Layout = "OuterLayout"; }
@section scripts {
   // stuff
}

// OuterLayout
@{ Layout = "BaseLayout"; }
@section scripts { 
  // Don't render inner scripts
}

// BaseLayout
@RenderSection("scripts");
```

In this scenario, the `scripts` section defined by `InnerLayout` is never rendered but the view would happily succeed. Perhaps we shouldn't be flattening the results in `unrenderedSections`?

Issue #2027, opened 2015-02-17
Provide option to extend select list to contain current value


dougbu:
Today, if a user passes a select list containing `{ 1, 2, 3, }` to `@Html.DropDownList()`, related methods, or the `<select>` tag helper when the expression value is `4`, the generated `<select/>` will look like nothing is selected.  This is true whether or not `4` would cause a validation error on submission.

Suggest automatically extending the select list to `{ 1, 2, 3, 4, }`, then select and disable the corresponding `<option/>` element.  For back-compatibility and to cover the tag helper case where the Razor source already contains an explicit `<option/>` for `4`, this should be opt-in.  That is, add `bool appendCurrentValue` parameters and `asp-append-current-value` attributes to the relevant HTML and tag helpers.

FYI the MVC 5 `enum` helpers (`@Html.EnumDropDownListFor()`) unconditionally ensured the current `enum` value was included in the generated `<option/>` elements and selected.  Could reuse some of that code (though not all of the semantics -- bit quirky and `enum`-specific in retrospect).

yishaigalatzer:
I'm not sure why we worry about back compat here. Even if there is a knob (which I'm not sure why it should even be there) it should be off by default

dougbu:
I agree with you on this behaviour being off by default @yishaigalatzer.  opt-in was the limit of my "back-compatibility" concern.

I also agree we shouldn't need a "knob" anywhere higher than an individual DDL call, LB call, or `<select/>` tag.  nothing broader than new parameters for the HTML helpers and a new bound attribute for the `<select/>` tag helper.

yishaigalatzer:
I don't know why a knob is needed at all

-----Original Message-----

dougbu:
offline decisions:
- when invoked using an HTML helper, always generate an additional `<option/>` element when the current value would not otherwise be shown
 - this is the correct behaviour and makes the `null` and "unknown" cases visibly distinct
- when invoked using the `<select/>` tag helper, have the contained `<option/>` tag helpers communicate their selection back (base on aspnet/Razor#238 or use existing `ViewContext.FormContext.FormData` approach)
 - change the `<select/>` tag helper to execute children prior to making a decision on generating the additional `<option/>` element
 - make this final decision based in part on whether an `<option/>` tag helper indicated it was selected.
[...]

Issue #438, opened 2014-05-10
Add `Html.GetEnumSelectList()` helpers


NTaylorMullen:
The first select helper `DropDownList` does not include any special-cases for `enum` types.  Need to support at least the scenarios that `EnumHelpers` and the `EnumDropDownList()` and `EnumDropDownListFor()` helpers enable in legacy MVC.

dougbu:
design is fairly simple: when unable to find a select list for a non-`[Flags]` `enum`, generate a collection of `SelectListItem`s based on the `enum`'s type.  this generation will be analogous to `EnumHelper.GetSelectList(Type)` from MVC 5.  (that class' handling of the current expression result will _not_ be included.  #2027 covers a similar but broader enhancement.)

minor complication for `<select/>` tag helper is that the Razor source may contain `<option/>` elements covering at least some of the `enum` fields.  place a `SelectedInformation` instance into the `FormContext` instead of the current `ICollection<string>` collection of selected values.  this allows contained `<option/>` tag helpers to inform the `<select/>` tag helper of explicit values.  the `<select/>` tag helper will use this data to eliminate duplicates.  also reqruires the `<select/>` tag helper to call `TagHelperContext.GetChildContentAsync()`.

much of the above is shared with the #602 and #2027 fixes.

TL;DR
this enhancement will not support `[Flags]` enums.  `@Html.ListBox[For]()` does not support non-collections.  and, while the form value provider correctly maps multiple values from a `<select/>` element to an `object` array, model binding accepts only the first element of that array.  so, no work to enable part of a scenario.

details of preconditions:
- `DefaultHtmlGenerator.GenerateSelect()` is called with `selectList: null`
- no `IEnumerable<SelectListItem>` is found in `ViewData`
- `ModelMetadata.ModelType.GetTypeInfo().IsEnum` or `Nullable.GetUnderlyingType(ModelMetadata.ModelType).GetTypeInfo().IsEnum`
- `enumType.GetTypeInfo().GetCustomAttribute<FlagsAttribute>() == null`

dougbu:
based on conversations with @Eilon, have simplified the API surface to a helper method that returns a select list for a given `enum` type.  e.g. `public IEnumerable<SelectListItem> GetSelectList<TEnum>() where TEnum : struct`.  so basically redo the `EnumHelper` method from MVC 5.

could include `... GetSelectList(Type enumType)` if people are passionate about that. 

[...]

PR #2229, opened 2015-03-21
Add `enum`-related details to `ModelMetadata`

dougbu:
- #438 part 1/2 and #2027 part 1/3 (all small)
- needs tests and XML doc corrections (broken references)

### DIFF
--- a/src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DataAnnotationsMetadataProvider.cs
+++ b/src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DataAnnotationsMetadataProvider.cs
@@ -1,8 +1,10 @@
 ﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.ComponentModel.DataAnnotations;
 using System.Linq;
+using System.Reflection;
 using Microsoft.Framework.Internal;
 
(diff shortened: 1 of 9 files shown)


rynowak:
Looks to be on the right track. I have some idea about how to clean up the cachability of this, but it can wait until later.

dougbu:
Updated to include tests and removed [Design] label

@rynowak any objection to the third commit I added?

rynowak:
> @rynowak any objection to the last commit I added (USE_REAL_RESOURCES)?

No objection, I just don't see what it fixes. We're testing that we call the correct overload of `DisplayAttribute` so it will use reflection to call a static property on a static class. It seems super-insignificant to me what that static property does for the purposes of the test (we mock things constantly).

I also don't see why/how we'd enable `USE_REAL_RESOURCES` as a define in the future. Basically IMO this just creates more ceremony on something we'd do 100% of the way in a perfect world. Due to limitations in our tests, we went 95% and to me that's good enough.

If you feel like this is worth it go ahead. :shipit: from me on the changes that actually related to the enum feature.

dougbu:
@rynowak to be clear, there's no "limitations" in our tests in this area.  I added `USE_REAL_RESOURCES` only to make it easy (a one-line addition to project.json) to test any workaround for the `[Display]` limitation we might come up with.  I agree that if we were successful, we'd remove `TestResources` and update the few places where we use `USE_REAL_RESOURCES`.

dougbu:
commit 7dd3afe

rynowak:
@dougbu :ok_hand: if you think it's better got for it!

PR #2230, opened 2015-03-21
Fixes #583 Generating a unique cookie name for an application.

### LINKED ISSUE #583: FileResult

harshgMSFT:
@GrabYourPitchforks .. please let me know if this does not sound right.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/AntiForgery/AntiForgery.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/AntiForgery/AntiForgery.cs
@@ -1,10 +1,15 @@
 // Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+using System.Linq;
+using System.Security.Cryptography;
+using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNet.DataProtection;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Mvc.Rendering;
+using Microsoft.AspNet.WebUtilities;
 using Microsoft.Framework.Internal;
 using Microsoft.Framework.OptionsModel;
 using Microsoft.Framework.WebEncoders;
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNet.Mvc.Core/AntiForgery/AntiForgery.cs
```diff
@@ -24,9 +28,13 @@ public sealed class AntiForgery
                            [NotNull] IDataProtectionProvider dataProtectionProvider,
                            [NotNull] IAntiForgeryAdditionalDataProvider additionalDataProvider,
                            [NotNull] IOptions<MvcOptions> mvcOptions,
-                           [NotNull] IHtmlEncoder htmlEncoder)
+                           [NotNull] IHtmlEncoder htmlEncoder,
+                           IOptions<DataProtectionOptions> dataProtectionOptions)
         {
             var config = mvcOptions.Options.AntiForgeryOptions;
+            var applicationId = dataProtectionOptions.Options.ApplicationDiscriminator ?? string.Empty;
```
rynowak: Should we use something stronger than `string.Empty` here?

src/Microsoft.AspNet.Mvc.Core/AntiForgery/AntiForgery.cs
```diff
@@ -109,5 +117,15 @@ public void SetCookieTokenAndHeader([NotNull] HttpContext context)
...
             _worker.SetCookieTokenAndHeader(context);
         }
+
+        private string ComputeCookieName(string applicationId)
+        {
+            using (var sha256 = SHA256.Create())
+            {
+                var hash = sha256.ComputeHash(Encoding.UTF8.GetBytes(applicationId));
+                var subHash = hash.Take(4).ToArray();
+                return Convert.ToBase64String(subHash);
```
rynowak: Just 4 bytes?

src/Microsoft.AspNet.Mvc.Core/AntiForgery/AntiForgery.cs
harshgMSFT: nope, the cookie name is not the secret, its the payload that is.

src/Microsoft.AspNet.Mvc.Core/AntiForgery/AntiForgery.cs
harshgMSFT: To keep the name short and avoid it being unnecessarily long.

harshgMSFT:
Updated

test/Microsoft.AspNet.Mvc.FunctionalTests/AntiForgeryTestHelper.cs
```diff
@@ -51,13 +51,25 @@ public static string RetrieveAntiForgeryToken(string htmlContent, string actionU
...
+                Key = cookieKey,
+                Value = cookieData
+            };
+        }
+
+        public class CookieMetadata
+        {
+            public string Key { get; set; }
+
+            public string Value { get; set; }
```
rynowak: Just re-use `KeyValuePair` :laughing:

rynowak:
:shipit:

Issue #2213, opened 2015-03-19
Tooling: Directives are not inheriteted via _GlobalImport.cshtml


NTaylorMullen:
Due to the change here: https://github.com/aspnet/Mvc/commit/cfd5630cf21d4d17ce9dc39554d57f5abb88a8a0 directives are no longer inherited in the tooling scenario because we short-circuit for rooted paths (tooling passes in rooted paths) in the `ViewStartUtility`, now `ViewHierarchyUtility`. 

This seems to have regressed at the beginning of December.

dougbu:
how are our tests passing?  or what part of the scenario is missing from our test suites?

NTaylorMullen:
Tooling provides a rooted path for the source file where runtime provides an app relative path.

dougbu:
OIC we were simply not testing the "design-time" scenarios correctly

NTaylorMullen:
0f1292abf6857c0a2d3a678a28c5fe8a03babf1d

Workign on tests.

PR #2232, opened 2015-03-21
Add test to validate tooling path normalization.

NTaylorMullen:
- Tooling passes in rooted paths when asking for a RazorParser for a file. This was problematic when resolving inherited code trees and ultimately this commit tests that.

#2213

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor.Host/MvcRazorHost.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor.Host/MvcRazorHost.cs
@@ -5,6 +5,7 @@ using System.Collections.Generic;
 using System.IO;
 using Microsoft.AspNet.FileProviders;
 using Microsoft.AspNet.Mvc.Razor.Directives;
+using Microsoft.AspNet.Mvc.Razor.Internal;
 using Microsoft.AspNet.Razor;
 using Microsoft.AspNet.Razor.Generator;
 using Microsoft.AspNet.Razor.Generator.Compiler;
@@ -36,31 +37,13 @@ namespace Microsoft.AspNet.Mvc.Razor
         // This field holds the type name without the generic decoration (MyBaseType)
         private readonly string _baseType;
         private readonly ICodeTreeCache _codeTreeCache;
+        private readonly RazorPathNormalizer _pathNormalizer;
         private ChunkInheritanceUtility _chunkInheritanceUtility;
 
-#if NET45
-        /// <summary>
-        /// Initializes a new instance of <see cref="MvcRazorHost"/> with the specified
-        /// <param name="root"/>.
-        /// </summary>
-        /// <param name="root">The path to the application base.</param>
-        // Note: This constructor is used by tooling and is created once for each
-        // Razor page that is loaded. Consequently, each loaded page has its own copy of
-        // the CodeTreeCache, but this ok - having a shared CodeTreeCache per application in tooling
-        // is problematic to manage.
-        public MvcRazorHost(string root) :
-            this(new DefaultCodeTreeCache(new PhysicalFileProvider(root)))
-        {
-            ApplicationRoot = root;
-        }
-#endif
-        /// <summary>
-        /// Initializes a new instance of <see cref="MvcRazorHost"/> using the specified <paramref name="fileProvider"/>.
-        /// </summary>
-        /// <param name="fileProvider">A <see cref="IFileProvider"/> rooted at the application base path.</param>
-        public MvcRazorHost(ICodeTreeCache codeTreeCache)
+        internal MvcRazorHost(ICodeTreeCache codeTreeCache, RazorPathNormalizer pathNormalizer)
             : base(new CSharpRazorCodeLanguage())
         {
+            _pathNormalizer = pathNormalizer;
             _baseType = BaseType;
             _codeTreeCache = codeTreeCache;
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNet.Mvc.Razor.Host/DesignTimeRazorPathNormalizer.cs
```diff
@@ -0,0 +1,29 @@
...
+        {
+            _applicationRoot = applicationRoot;
+        }
+
+        public override string NormalizePath([NotNull] string path)
+        {
+            // Need to convert path to application relative (rooted paths are passed in during design time).
+            if (Path.IsPathRooted(path))
+            {
+                path = path.Substring(_applicationRoot.Length);
```
rynowak: worth verifying that `_applicationRoot` ends with a trailing slash? Probably good to verify and throw `StartsWith` since we'll do silly invalid things if the arguments are correct.

test/Microsoft.AspNet.Mvc.Razor.Host.Test/MvcRazorHostTest.cs
```diff
@@ -254,6 +286,23 @@ private static Stream GetResourceStream(string resourceName)
                 generatedLocation: new MappingLocation(generatedLocation, contentLength));
         }
 
+        private class PathValidatingChunkInheritanceUtility : ChunkInheritanceUtility
```
rynowak: Might be simpler to subclass `MvcRazorHost` in the tests and `new DecorateRazorParser` or just verify the data that you're looking for in the return values. Then you don't need any 'test only' code in the product.

rynowak:
:watch:

src/Microsoft.AspNet.Mvc.Razor.Host/DesignTimeRazorPathNormalizer.cs
NTaylorMullen: :+1:

test/Microsoft.AspNet.Mvc.Razor.Host.Test/MvcRazorHostTest.cs
NTaylorMullen: The only piece that gets the source file name where the path being relative is important is the `ChunkInheritanceUtility` so adding this class is necessary. The other alternative was adding a semi-unit semi-functional test which took the entire route of resolving a view based on a rooted path; to fully test that approach without exploding took a decent amount of work/replacement.

NTaylorMullen:
Updated.

test/Microsoft.AspNet.Mvc.Razor.Host.Test/MvcRazorHostTest.cs
rynowak: K, no big concenrs about the testability changes, just wanted to suggest an idea.

src/Microsoft.AspNet.Mvc.Razor.Host/DesignTimeRazorPathNormalizer.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        public DesignTimeRazorPathNormalizer([NotNull] string applicationRoot)
+        {
+            _applicationRoot = applicationRoot;
+        }
+
+        public override string NormalizePath([NotNull] string path)
+        {
+            // Need to convert path to application relative (rooted paths are passed in during design time).
+            if (Path.IsPathRooted(path) && path.StartsWith(_applicationRoot, StringComparison.Ordinal))
```
rynowak: There's still no check for a trailing slash here. These checks would pass in a case where `applicationroot = c:\foo` which seems super innocent on the surface. Then you'd end up returning `\bar\baz.cshtml` from this, which is not what you want.

rynowak:
:shipit:  after adding a check for trailing slash

src/Microsoft.AspNet.Mvc.Razor.Host/DesignTimeRazorPathNormalizer.cs
dougbu: in what case would `_applicationRoot` be empty or otherwise _not_ a rooted path?  not sure the first part of this condition helps

src/Microsoft.AspNet.Mvc.Razor.Host/MvcRazorHost.cs
```diff
@@ -115,10 +98,27 @@ public MvcRazorHost(ICodeTreeCache codeTreeCache)
 
 #if NET45
         /// <summary>
-        /// The path to the application root.
+        /// Initializes a new instance of <see cref="MvcRazorHost"/> with the specified
+        /// <param name="root"/>.
```
dougbu: `<paramref`

src/Microsoft.AspNet.Mvc.Razor.Host/MvcRazorHost.cs
```diff
@@ -115,10 +98,27 @@ public MvcRazorHost(ICodeTreeCache codeTreeCache)
...
         /// </summary>
-        public string ApplicationRoot { get; }
+        /// <param name="root">The path to the application base.</param>
+        // Note: This constructor is used by tooling and is created once for each
+        // Razor page that is loaded. Consequently, each loaded page has its own copy of
+        // the CodeTreeCache, but this ok - having a shared CodeTreeCache per application in tooling
+        // is problematic to manage.
+        public MvcRazorHost(string root) :
+            this(new DefaultCodeTreeCache(new PhysicalFileProvider(root)), new DesignTimeRazorPathNormalizer(root))
```
dougbu: nit: we normally keep the colon and constructor call together on one line i.e. `: this(...`

src/Microsoft.AspNet.Mvc.Razor.Host/DesignTimeRazorPathNormalizer.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.AspNet.Mvc.Razor.Internal
```
dougbu: nit: should be in an Internal directory

src/Microsoft.AspNet.Mvc.Razor.Host/RazorPathNormalizer.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.AspNet.Mvc.Razor.Internal
```
dougbu: nit: should be in an Internal directory

src/Microsoft.AspNet.Mvc.Razor.Host/DesignTimeRazorPathNormalizer.cs
NTaylorMullen: @rynowak That's not this path normalizations concern. The `ChunkInheritanceUtility` and other layers understand that format.

@dougbu We're ensuring tooling doesn't change their behavior.

dougbu: >  tooling doesn't change their behavior

then perhaps `Debug.Assert(path.StartsWith(_applicationRoot, StringComparison.Ordinal))`?

my point is the method will not change behaviour if you remove the `Path.IsPathRooted(path)` check

dougbu:
:shipit: after fixing the doc comment and once @rynowak is pleased as punch

Issue #1736, opened 2014-12-20
Case sensitive filesystems and Razor view searches


callumj:
### Title
When using a case sensitive file system the Razor View Engine will be unable to find the controller's view folder unless the path matches case in the directory structure.

### Functional impact
Marginal, just requires that developers ensure the use of controller names in paths matches the same case as the directory structure.

### Minimal repro steps
1. Deploy a MVC application on a case sensitive file system (Mono on Linux) Example app used: https://github.com/callumj/vnext-mvc-sample
2. Create a controller with a view such as `HomeController#Create`, this should have a view `Home/Create.cshtml`
3. Navigate to `/home/Create`
4. MVC will be unable to locate `Home/Create.cshtml` because it has been told to try `home/Create.cshtml`.

### Expected result
It should locate the view irregardless of the case.

### Actual result
View not found

yishaigalatzer:
We should consider the following things as well when fixing this:
1. Controller
2. Action
3. Area
4. View Expander sample

markrendle:
This is not "marginal", it's literally the first thing I ran into working under Linux and it's *really* annoying. I lost an hour chasing it down.

pranavkm:
Fixed in 54c1fed254d100d075500147f4b44368f391d99f

PR #2233, opened 2015-03-21
Issue #1736 - Razor view searches in case-sensitive filesystems.

sornaks:
#1736


src/Microsoft.AspNet.Mvc.Razor/RazorViewEngine.cs
```diff
@@ -228,6 +228,29 @@ public virtual IEnumerable<string> AreaViewLocationFormats
             return new RazorPageResult(pageName, searchedLocations);
         }
 
+        private string GetNormalizedRouteValue(string key, ActionContext context)
+        {
+            if (context.ActionDescriptor.AttributeRouteInfo == null)
+            {
+                var match = context.ActionDescriptor.RouteConstraints.SingleOrDefault(r => r.RouteKey == key);
```
rynowak: For safety we want to do a case-insensitive comparison here. `RouteValueDefaults` is already case-insensitive.

rynowak: The additional thing you have to check with RouteConstraints is the keytype. You only want to normalize if it's a requiredmatch (or whatever). If it's a deny then the value must be null, If it's a catch-all you need to use the route value.

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngine.cs
```diff
@@ -228,6 +228,29 @@ public virtual IEnumerable<string> AreaViewLocationFormats
             return new RazorPageResult(pageName, searchedLocations);
         }
 
+        private string GetNormalizedRouteValue(string key, ActionContext context)
```
rynowak: This is absolutely worth documenting in the comments what the bug we're fixing is here and why it's so important.

test/Microsoft.AspNet.Mvc.Razor.Test/RazorViewEngineTest.cs
```diff
@@ -594,7 +594,109 @@ public void FindPage_ReturnsSearchedLocationsIfPageCannotBeFound()
...
 
-       [Fact]
+        [Theory]
+        // Looks in RouteValueDefaults
+        [InlineData(true)]
+        // Looks in RouteConstraints
+        [InlineData(false)]
+        public void FindPage_SelectsActionCaseInsensitively(bool isAttributeRouted)
+        {
+            // The RouteData contains "Foo", and the ActionDescriptor contains "foo"
```
rynowak: Reverse these imo, that would make it much more like the actual scenario users reported.

test/Microsoft.AspNet.Mvc.Razor.Test/RazorViewEngineTest.cs
```diff
@@ -594,7 +594,109 @@ public void FindPage_ReturnsSearchedLocationsIfPageCannotBeFound()
...
+            };
+
+            var page = Mock.Of<IRazorPage>();
+            var pageFactory = new Mock<IRazorPageFactory>();
+            pageFactory.Setup(p => p.CreateInstance("/Views/foo/layout.cshtml"))
+                       .Returns(page)
+                       .Verifiable();
+
+            var viewEngine = CreateViewEngine(pageFactory.Object);
+            var routes = new Dictionary<string, string>()
```
rynowak: this name is pretty unclear

test/Microsoft.AspNet.Mvc.Razor.Test/RazorViewEngineTest.cs
```diff
@@ -594,7 +594,109 @@ public void FindPage_ReturnsSearchedLocationsIfPageCannotBeFound()
...
+            // The RouteData contains "Foo", and the ActionDescriptor contains "foo"
+            // which matches the case of the constructor thus searching in the appropriate location.
+            // Arrange
+            var routeValues = new Dictionary<string, object>
+            {
+                { "controller", "Foo" }
+            };
+
+            var page = Mock.Of<IRazorPage>();
+            var pageFactory = new Mock<IRazorPageFactory>();
```
rynowak: Make it strict so it throws if you give it the wrong value.

test/Microsoft.AspNet.Mvc.Razor.Test/RazorViewEngineTest.cs
```diff
@@ -594,7 +594,109 @@ public void FindPage_ReturnsSearchedLocationsIfPageCannotBeFound()
...
+            var viewEngine = CreateViewEngine(pageFactory.Object);
+            var routes = new Dictionary<string, string>()
+            {
+                { "controller", "foo" }
+            };
+
+            var context = GetActionContextWithRoutes(routeValues, routes, isAttributeRouted);
+
+            // Act
+            var result = viewEngine.FindPage(context, "layout");
```
rynowak: pick something else other than layout - like `Details`.

test/Microsoft.AspNet.Mvc.Razor.Test/RazorViewEngineTest.cs
```diff
@@ -594,7 +594,109 @@ public void FindPage_ReturnsSearchedLocationsIfPageCannotBeFound()
...
+            {
+                { "controller", "bar" }
+            };
+            var page = Mock.Of<IRazorPage>();
+
+            var viewEngine = CreateViewEngine();
+            var context = GetActionContextWithRoutes(routeValues, routes, isAttributeRouted);
+
+            // Act
+            var result = viewEngine.FindPage(context, "bar");
```
rynowak: page name is the same as the controller name?

test/Microsoft.AspNet.Mvc.Razor.Test/RazorViewEngineTest.cs
```diff
@@ -594,7 +594,109 @@ public void FindPage_ReturnsSearchedLocationsIfPageCannotBeFound()
...
+
+            // Assert
+            Assert.Equal("bar", result.Name);
+            Assert.Null(result.Page);
+            Assert.Equal(expected, result.SearchedLocations);
+        }
+
+        [Fact]
         public void AreaViewLocationFormats_ContainsExpectedLocations()
```
rynowak: I also want to see a case for areas where the route data gets normalized.

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngine.cs
```diff
@@ -228,6 +228,29 @@ public virtual IEnumerable<string> AreaViewLocationFormats
...
+            {
+                object match;
+                context.ActionDescriptor.RouteValueDefaults.TryGetValue(key, out match);
+                if (match != null)
+                {
+                    return match.ToString();
+                }
+            }
+
+            return context.RouteData.Values.GetValueOrDefault<string>(key);
```
rynowak: should we compare the route data **value** against the AD value case-insensitively as a check? I can't think of any super-valid case where this matters, but it will make sure we're not doing more than we intended to.

rynowak:
:watch:

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngine.cs
pranavkm: Does this need to be `SingleOrDefault` as opposed to `FirstOrDefault`? `SingleOrDefault` will throw if you have multiple entries which match the constraint. If we want to explicitly deal with this scenario, we should throw a better exception.

kichalla:
I think we should even make changes to view location cache (as the key is case sensitive)?
https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Razor/DefaultViewLocationCache.cs#L48

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngine.cs
sornaks: Hmm, not sure if I got this comment. Will talk to you once you are here..

sornaks:
Updated.. Please have a look.

src/Microsoft.AspNet.Mvc.Razor/Internal/RouteUtility.cs
```diff
@@ -0,0 +1,51 @@
...
+
+namespace Microsoft.AspNet.Mvc.Razor.Internal
+{
+    /// <summary>
+    /// Utility for normalizing route values.
+    /// </summary>
+    public class RouteUtility
+    {
+        /// <summary>
+        /// In a case sensitive file system, Views with a case different from the URL will not be found
```
pranavkm: Perhaps this should belong in the `remarks` section? The summary should be more along the lines of 

> Gets the value for a route key from the route constraints or route value defaults of an `<see cref="ActionDescriptor"/>` if it exists there, otherwise from the `RouteData`. 

(or something along those lines)

src/Microsoft.AspNet.Mvc.Razor/Internal/RouteUtility.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// if we are looking only in the route data. Hence looking at the ActionDescriptor as well.
+        /// </summary>
+        /// <param name="key">The key to lookup.</param>
+        /// <param name="context">The key is searched in this <see cref="ActionContext"/>.</param>
+        /// <returns>The value corresponding to the key.</returns>
+        public static string GetNormalizedRouteValue(string key, ActionContext context)
```
pranavkm: Consider having `ActionContext` be the first param?

src/Microsoft.AspNet.Mvc.Razor/Internal/RouteUtility.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            if (context.ActionDescriptor.AttributeRouteInfo == null)
```
pranavkm: Maybe add a comment here stating the scenario (attribute routed versus not?)

src/Microsoft.AspNet.Mvc.Razor/Internal/RouteUtility.cs
```diff
@@ -0,0 +1,51 @@
...
+                        return null;
+                    }
+
+                    return match.RouteValue;
+                }
+            }
+            else
+            {
+                object match;
+                context.ActionDescriptor.RouteValueDefaults.TryGetValue(key, out match);
```
pranavkm:
```C#
if (context.ActionDescriptor.RouteValueDefaults.TryGetValue(key, out match))
{
   return match.ToString();
}
```

src/Microsoft.AspNet.Mvc.Razor/DefaultViewLocationCache.cs
```diff
@@ -49,7 +50,9 @@ internal static string GenerateKey(ViewLocationExpanderContext context)
         {
             var keyBuilder = new StringBuilder();
             var routeValues = context.ActionContext.RouteData.Values;
-            var controller = routeValues.GetValueOrDefault<string>(RazorViewEngine.ControllerKey);
+            var controller = RouteUtility.GetNormalizedRouteValue(
+                RazorViewEngine.ControllerKey,
+                context.ActionContext);
 
             // format is "{viewName}:{isPartial}:{controllerName}:{areaName}:"
             keyBuilder.Append(context.ViewName)
```
kichalla: I think the view name here is a user-supplied one? if yes, there are chances the case could be different depending on different areas of code executing this view (even though the chances are remote i think)...one alternative is to make the dictionary key to be `OrdinalIgnoreCase` and you can avoid doing all the other changes in this class...

pranavkm: It's developer supplied as opposed to application user \ browser supplied. Changing the cache key doesn't make a significant difference if the value ends up being the wrong case resulting a failure in lookup.

pranavkm:
Updated.

src/Microsoft.AspNet.Mvc.Razor/DefaultViewLocationCache.cs
```diff
@@ -49,7 +49,9 @@ internal static string GenerateKey(ViewLocationExpanderContext context)
         {
             var keyBuilder = new StringBuilder();
             var routeValues = context.ActionContext.RouteData.Values;
-            var controller = routeValues.GetValueOrDefault<string>(RazorViewEngine.ControllerKey);
+            var controller = RouteValueUtility.GetNormalizedRouteValue(
```
rynowak: Gross

rynowak: Why not just have this be a static on RazorViewEngine - you already have the keys as statics on that class.

src/Microsoft.AspNet.Mvc.Razor/Extensions/DictionaryExtensions.cs
```diff
@@ -1,26 +0,0 @@
...
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-using Microsoft.AspNet.Mvc;
-using Microsoft.Framework.Internal;
-
-namespace System.Collections.Generic
-{
-    internal static class DictionaryExtensions
-    {
-        public static T GetValueOrDefault<T>([NotNull] this IDictionary<string, object> dictionary,
```
rynowak: YAY

src/Microsoft.AspNet.Mvc.Razor/Internal/RouteValueUtility.cs
```diff
@@ -0,0 +1,64 @@
...
+        /// <see cref="ActionDescriptor.RouteValueDefaults"/> for attribute routes and
+        /// <see cref="ActionDescriptor.RouteConstraints"/> for traditional routes to get route values produces
+        /// consistently cased results.
+        /// </remarks>
+        public static string GetNormalizedRouteValue(ActionContext context, string key)
+        {
+            var actionDescriptor = context.ActionDescriptor;
+            if (actionDescriptor.AttributeRouteInfo != null)
+            {
+                // For attribute routes, use an inline constraint if available.
```
rynowak: This comment doesn't make any sense.

src/Microsoft.AspNet.Mvc.Razor/Internal/RouteValueUtility.cs
```diff
@@ -0,0 +1,64 @@
...
+                object match;
+                if (actionDescriptor.RouteValueDefaults.TryGetValue(key, out match))
+                {
+                    return match.ToString();
```
rynowak: Should we verify that this isn't doing more than changing the case? For instance if the use manually edits route values (wierd idea, but someone will try it).

rynowak:
:watch: Tests look good, I have a few small comments about the implementation.

src/Microsoft.AspNet.Mvc.Razor/DefaultViewLocationCache.cs
pranavkm: Sure.

pranavkm: Although the only issue is the keys are `internal` and maybe this method has some utility in user code - for instance in a custom view location expander.

src/Microsoft.AspNet.Mvc.Razor/Internal/RouteValueUtility.cs
pranavkm: Thought as much.

src/Microsoft.AspNet.Mvc.Razor/Internal/RouteValueUtility.cs
pranavkm: Would that mean it's something along the lines of

```C#
if (actionDescriptor.RouteValueDefaults.TryGetValue(key, out match))
{
   var stringValue = match.ToString();
   if (string.Equals(stringValue, routeDataValue, StringComparison.OrdinalIgnoreCase))
   { 
       return stringValue; 
    }
}
```

rynowak: :+1:

pranavkm:
Updated.

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngine.cs
```diff
@@ -133,6 +135,62 @@ public virtual IEnumerable<string> AreaViewLocationFormats
...
+        /// <param name="key">The route key to lookup.</param>
+        /// <returns>The value corresponding to the key.</returns>
+        /// <remarks>
+        /// The casing of a route value in <see cref="ActionContext.RouteData"/> is determined by the client.
+        /// This making constructing paths for view locations in a case sensitive file system unreliable. Using the
+        /// <see cref="ActionDescriptor.RouteValueDefaults"/> for attribute routes and
+        /// <see cref="ActionDescriptor.RouteConstraints"/> for traditional routes to get route values produces
+        /// consistently cased results.
+        /// </remarks>
+        internal static string GetNormalizedRouteValue(ActionContext context, string key)
```
rynowak: Why is this internal?

rynowak:
:shipit:

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngine.cs
pranavkm: Seems odd for it to be public (the two keys are internal too FYI).

Issue #1712, opened 2014-12-12
MutableObjectModelBinder.GetPropertyValidationInfo should not reflect over `ModelType.Properties`


pranavkm:
https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.ModelBinding/Binders/MutableObjectModelBinder.cs#L305-L306

In the MutableObjectModelBinder we attempt to look for `BindingBehaviorAttribute` on properties. As part of this, we iterate over all properties on the `ModelType`. This is problematic in

1. There is a mismatch in the properties we discover as part of  `ModelMetadata` and the ones that are   discovered here. For instance `ModelMetadata` ignores indexers versus this does not. Consequently a model with indexers causes an exception to be thrown at https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.ModelBinding/Binders/MutableObjectModelBinder.cs#L309.

2. Reflecting on the type as opposed to using property metadata seems to be against the established pattern of reading cached metadata.

rynowak:
Triage: This is a follow-on to the model-metadata work and should be considered for RC.

rynowak:
f3679f214e85b64bc4d7387e34753f4b8e01beb8

PR #2234, opened 2015-03-22
Part 1 of a fix for #1712 - ModelBinding and validation should not use

rynowak:
reflection.

This change moves the logic to process BindingBehaviorAttribute into the
model metadata system, and exposes it via that CanBeBound and IsRequired
properties.

This removes the need in the MutableObjectModelBinder to reflect over
properties and look at for this attribute.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.ModelBinding/Binders/MutableObjectModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.ModelBinding/Binders/MutableObjectModelBinder.cs
@@ -273,14 +273,20 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
 
         protected virtual IEnumerable<ModelMetadata> GetMetadataForProperties(ModelBindingContext bindingContext)
         {
-            var validationInfo = GetPropertyValidationInfo(bindingContext);
-            var newPropertyFilter = GetPropertyFilter();
-            return bindingContext.ModelMetadata.Properties
-                                 .Where(propertyMetadata =>
-                                    newPropertyFilter(bindingContext, propertyMetadata.PropertyName) &&
-                                    (validationInfo.RequiredProperties.Contains(propertyMetadata.PropertyName) ||
-                                    !validationInfo.SkipProperties.Contains(propertyMetadata.PropertyName)) &&
-                                    CanUpdateProperty(propertyMetadata));
+            var propertyBindingInfo = GetPropertyBindingInfo(bindingContext);
+
+            var propertyFilter = GetPropertyFilter();
+            return 
+                bindingContext.ModelMetadata
+                .Properties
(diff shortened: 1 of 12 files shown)


src/Microsoft.AspNet.Mvc.ModelBinding/Binders/MutableObjectModelBinder.cs
```diff
@@ -273,14 +273,20 @@ protected virtual void EnsureModel(ModelBindingContext bindingContext)
...
+                bindingContext.ModelMetadata
+                .Properties
+                .Where(propertyMetadata =>
+                {
+                    return 
+                        propertyFilter(bindingContext, propertyMetadata.PropertyName) &&
+                        (propertyBindingInfo.RequiredProperties.Contains(propertyMetadata.PropertyName) ||
+                        !propertyBindingInfo.SkipProperties.Contains(propertyMetadata.PropertyName)) &&
+                        CanUpdateProperty(propertyMetadata);
+                });
```
rynowak: I had to rename some of these things, and this formatting was driving me nuts. Still don't like it so much.

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/MutableObjectModelBinder.cs
```diff
@@ -310,21 +316,12 @@ private static bool TryGetPropertyDefaultValue(PropertyInfo propertyInfo, out ob
             }
         }
 
-        internal static PropertyValidationInfo GetPropertyValidationInfo(ModelBindingContext bindingContext)
+        private static PropertyValidationInfo GetPropertyBindingInfo(ModelBindingContext bindingContext)
```
rynowak: In some past version this was internal for testing (possibly in webapi), but there are no tests that use it in the current codebase.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/ModelMetadataProviderTest.cs
```diff
@@ -621,14 +665,6 @@ public DataTypeWithCustomEditFormat() : base("Custom datatype")
...
         }
 
-        private void ActionWithoutBindAttribute(User param)
-        {
-        }
-
-        private void ActionWithBindAttribute([Bind(new string[] { "IsAdmin" }, Prefix = "ParameterPrefix")] User param)
-        {
-        }
-
```
rynowak: unused, this should have been removed as part of removing `GetMetadataForParameter`

rynowak:
Hitting the pause button here. Making validation respect our `ModelMetadata.IsRequired` property might cause some trouble. Will have a discussion and come up with another PR.

Issue #602, opened 2014-06-02
Provide default select lists for `bool` and `bool?` expressions


dougbu:
Similar to #438 should provide automatic (well, static) select lists for `bool` and `bool?` expressions.  This amounts to using `DefaultDisplayTemplates.TriStateValues()` (and something similar for non-`Nullable` Booleans) when the user does not pass a select list  to `@Html.DropDownList()` and `@Html.ListBox()`.

Current implementation gives up when no select list is passed or found in `ViewData`.

dougbu:
Triage: this helper is fairly trivial but simple for customers to do themselves.  suggest we not fix this one for RC.

danroth27:
@dougbu Could you please add a quick code sample to this bug that shows what a user would have to do to solve this problem?

dougbu:
I wouldn't recommend using a `<select/>` for any regular `bool`.  just use a checkbox.  `<select/>` is possible but would look strange.

for `bool?`, assuming you want localization and have the resources in your resx file:
``` c#
var selectListForNullableBool = new List<SelectListItem>
{
    new SelectListItem
    {
        Text = Resources.Common_NotSet,
        Value = string.Empty,
    },
    new SelectListItem
    {
        Text = Resources.Common_True,
        Value = "true",
    },
    new SelectListItem
    {
        Text = Resources.Common_False,
        Value = "false",
    },
};
```
the first `SelectListItem` isn't :100:% required when using `Html.DropDownList[For]()` because you can call that method as follows:
``` c#
@Html.DropDownListFor(m => m.BoolQuest, selectList, optionLabel: Resources.Common_NotSet)
```

PR #2235, opened 2015-03-22
Add `Html.GetSelectList()` helpers

dougbu:
- #438 part 2/2


src/Microsoft.AspNet.Mvc.Core/Rendering/Html/HtmlHelper.cs
```diff
@@ -351,6 +351,35 @@ public string DisplayText(string expression)
...
         /// <inheritdoc />
+        public IEnumerable<SelectListItem> GetSelectList<TEnum>() where TEnum : struct
+        {
+            var type = typeof(TEnum);
+            var metadata = MetadataProvider.GetMetadataForType(type);
+            if (!metadata.IsEnum || metadata.IsFlagsEnum)
+            {
+                throw new ArgumentException(
+                    Resources.FormatHtmlHelper_TypeNotSupported(type.FullName),
+                    nameof(TEnum));
```
rynowak: This should be a better string and make it clear that this is for enums.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/HtmlHelper.cs
```diff
@@ -1016,5 +1045,22 @@ protected virtual string GenerateValue(string expression, object value, string f
         {
             return _htmlGenerator.GetClientValidationRules(ViewContext, modelExplorer, expression);
         }
+
+        protected virtual IEnumerable<SelectListItem> GetSelectList([NotNull] ModelMetadata metadata)
```
rynowak: protected is like public, should we validate here? or perhaps name this something like `GetEnumSelectList`?

src/Microsoft.AspNet.Mvc.Core/Rendering/IHtmlHelper.cs
```diff
@@ -374,7 +375,35 @@ public interface IHtmlHelper
...
+        IEnumerable<ModelClientValidationRule> GetClientValidationRules(
+            ModelExplorer modelExplorer,
+            string expression);
+
+        /// <summary>
+        /// Returns a select list for the given <typeparamref name="TEnum"/>.
+        /// </summary>
+        /// <typeparam name="TEnum">Type to generate a select list for.</typeparam>
+        /// <returns>
+        /// An IEnumerable{SelectListItem} containing the select list for the given <typeparamref name="TEnum"/>.
```
rynowak: should IEnumerable be in a `<see ... />`? (you're the expert on this).

test/Microsoft.AspNet.Mvc.Core.Test/Rendering/HtmlHelperSelectTest.cs
```diff
@@ -858,6 +864,352 @@ public void ListBoxFor_WithUnreleatedExpression_GeneratesExpectedValue()
...
+        {
+            // Arrange
+            var metadataProvider = TestModelMetadataProvider.CreateDefaultProvider();
+            var htmlHelper = new TestHtmlHelper(metadataProvider);
+
+            // Act & Assert
+            ExceptionAssert.ThrowsArgument(
+                () => htmlHelper.GetSelectList<EnumWithFlags>(),
+                "TEnum",
+                $"The type '{ typeof(EnumWithFlags).FullName }' is not supported.");
```
rynowak: 1538 PranavPoints have been awarded :+1:

test/Microsoft.AspNet.Mvc.Core.Test/Rendering/HtmlHelperSelectTest.cs
```diff
@@ -858,6 +864,352 @@ public void ListBoxFor_WithUnreleatedExpression_GeneratesExpectedValue()
...
+            // Arrange
+            var metadataProvider = TestModelMetadataProvider.CreateDefaultProvider();
+            var metadata = metadataProvider.GetMetadataForType(typeof(EnumWithFields));
+            var htmlHelper = new TestHtmlHelper(metadataProvider);
+
+            // Act
+            var result = htmlHelper.GetSelectList<EnumWithFields>();
+
+            // Assert
+            Assert.Equal(metadata.ModelType, htmlHelper.Metadata.ModelType);
```
rynowak: wait, how does this happen? I'm confused about how `htmlHelper.Metadata` get its type

rynowak: Oh I see, it's  capturing thing you built for the tests. Carry on then.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/HtmlHelper.cs
dougbu: hmm, think I named this before decided to close #602.  without `bool?`, I can't think of another type this method could support.  I'll rename all the overloads to `GetEnumSelectList`.

src/Microsoft.AspNet.Mvc.Core/Rendering/IHtmlHelper.cs
dougbu: good catch

test/Microsoft.AspNet.Mvc.Core.Test/Rendering/HtmlHelperSelectTest.cs
```diff
@@ -858,6 +864,352 @@ public void ListBoxFor_WithUnreleatedExpression_GeneratesExpectedValue()
...
+            var htmlHelper = new TestHtmlHelper(metadataProvider);
+
+            // Act
+            var result = htmlHelper.GetSelectList(typeof(EnumWithFields));
+
+            // Assert
+            Assert.Equal(metadata.ModelType, htmlHelper.Metadata.ModelType);
+
+            Assert.Same(htmlHelper.SelectListItems, result);            // No replacement of the underlying List
+            VerifySelectList(htmlHelper.CopiedSelectListItems, result); // No change to the (mutable) items
```
rynowak: I really don't understand what this is doing. This test (and `GetSelectListTEnum_WrapsGetSelectListModelMetadata`) just seems to me to verify the class you wrote to test this feature.

test/Microsoft.AspNet.Mvc.Core.Test/Rendering/HtmlHelperSelectTest.cs
```diff
@@ -858,6 +864,352 @@ public void ListBoxFor_WithUnreleatedExpression_GeneratesExpectedValue()
...
+
+            [Display(Name = "menos uno")]
+            MinusOne = -1,
+
+#if USE_REAL_RESOURCES
+            [Display(Name = nameof(Test.Resources.DisplayAttribute_Name), ResourceType = typeof(Test.Resources))]
+#else
+            [Display(Name = nameof(TestResources.DisplayAttribute_Name), ResourceType = typeof(TestResources))]
+#endif
+            MinusTwo = -2,
```
rynowak: Should we have a case where an enum field has a `[Display(...)]` but doesn't set the name?

rynowak:
:watch: looks mostly good, I had a few small questions

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/HtmlHelper.cs
dougbu: the `<exception/>` doc comments aren't good enough for you?  :smile_cat: 

how about I add `Type must be an enum that does not have an associated FlagsAttribute.` to the resource?  I'd pass in `nameof(Enum).ToLowerInvariant()` and `nameof(FlagsAttribute)`

test/Microsoft.AspNet.Mvc.Core.Test/Rendering/HtmlHelperSelectTest.cs
dougbu: :ok:

test/Microsoft.AspNet.Mvc.Core.Test/Rendering/HtmlHelperSelectTest.cs
dougbu: these two tests are attempting to verify that `GetSelectList<TEnum>()` and `GetSelectList(Type)` do nothing but verification on their own i.e. they're pass-through wrappers.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/HtmlHelper.cs
rynowak: :+1:

rynowak:
:shipit: consider me signed off after fixing the things you agreed to fix.

dougbu:
commit 570b1e5

PR #2236, opened 2015-03-22
Fix ModelMetadata.IsRequired

rynowak:
The DataAnnotationsMetadataProvider was being too aggressive in setting
IsRequired, so we never actually executed the fallback logic here. Found
this while working on removing some reflection code from the validator,
and wanted to split it out because it's simple.


test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DataMemberBindingMetadataProviderTest.cs
```diff
@@ -90,7 +90,7 @@ public void ClassWithDataMemberIsRequiredTrueWithoutDataContract_NoValidator()
         private class ClassWithDataMemberIsRequiredTrueWithoutDataContract
         {
             [DataMember(IsRequired = true)]
-            public int TheProperty { get; set; }
+            public string TheProperty { get; set; }
```
rynowak: Changing to keep away from the default behavior for value types.

dougbu:
nits:
- your commit description ("aggressive" in particular) is a bit too general.  the issue here is that `DataAnnotationsMetadataProvider` was interfering with the default value provided in `DefaultModelMetadata`.
- could have `DataMemberBindingMetadataProvider` short-circuit if `IsRequired` is already `true`.  extra check in that class doesn't add anything but doesn't really hurt either.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/ModelMetadataProviderTest.cs
```diff
@@ -586,6 +587,69 @@ public void BinderTypeProviders_FirstAttributeHasPrecedence()
...
+        {
+            // Arrange
+            var attributes = new object[]
+            {
+            };
+
+            var provider = CreateProvider(attributes);
+
+            // Act
+            var metadata = provider.GetMetadataForProperty(typeof(string), "Length");
```
dougbu: nice -- a required read-only property :smile:

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/ModelMetadataProviderTest.cs
```diff
@@ -621,14 +685,6 @@ public DataTypeWithCustomEditFormat() : base("Custom datatype")
...
         }
 
-        private void ActionWithoutBindAttribute(User param)
-        {
-        }
-
-        private void ActionWithBindAttribute([Bind(new string[] { "IsAdmin" }, Prefix = "ParameterPrefix")] User param)
-        {
-        }
-
```
dougbu: I take it these were unused?

dougbu:
- please also add tests similar to `GetBindingDetails_NoRequiredAttribute_IsRequiredLeftAlone` for `DataMemberRequiredValidationMetadataProvider`
- nit: `IsReadOnly` handling in `DataAnnotationsMetadataProvider` looks correct but isn't tested as thoroughly and could hit similar issues to `IsRequired`.  if you have time, please add a `GetBindingDetails_NoEditableAttribute_IsReadOnlyLeftAlone` test too.

dougbu:
:watch: for one or two more tests

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/ModelMetadataProviderTest.cs
rynowak: Yes, these should have been removed along with `GetMetadataForParameter`

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DataMemberRequiredBindingMetadataProvider.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
     /// <summary>
     /// An <see cref="IBindingMetadataProvider"/> for <see cref="DataMemberAttribute.IsRequired"/>.
     /// </summary>
-    public class DataMemberRequiredValidationMetadataProvider : IBindingMetadataProvider
+    public class DataMemberRequiredBindingMetadataProvider : IBindingMetadataProvider
```
rynowak: This name was wrong. Fixed the tests too.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DataMemberBindingMetadataProviderTest.cs
```diff
@@ -1,96 +0,0 @@
...
-
-using System.Runtime.Serialization;
-using Xunit;
-
-namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
-{
-    public class DataMemberBindingMetadataProviderTest
-    {
-        [Fact]
-        public void ClassWithoutAttributes_NotRequired()
```
rynowak: Moved these existing tests to `ModeMetadataProviderTest` because they are actually integration tests. Added new unit tests to the renamed version of this file which cover the same cases.

rynowak:
@dougbu ping

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DataMemberBindingValidationMetadataProviderTest.cs
```diff
@@ -0,0 +1,168 @@
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Runtime.Serialization;
+using Xunit;
+
+namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
+{
+    public class DataMemberBindingMetadataProviderTest
```
dougbu: need to rename the file

dougbu: also, shouldn't this class be named `DataMemberRequiredBindingMetadataProviderTest` ?

dougbu:
:shipit: w/ one rename

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DataMemberBindingValidationMetadataProviderTest.cs
rynowak: :+1: on both counts, sorry thought I fixed all of this.

rynowak:
784021cf85f695d52479634afd8108bd5862ee07

PR #2237, opened 2015-03-22
Refactor `IHtmlGenerator.GenerateSelect()` to enable #2027 fix

dougbu:
- #2027 part 2/3
- add `IHtmlGenerator.GetCurrentValues()` method
 - allow `<option/>` tag helpers to run before `<select/>` helper generation
 - bring together bits of `GenerateSelect()` and `UpdateSelectListItemsWithDefaultValue()`
- add doc comments for `IHtmlGenerator.GenerateSelect()` methods


src/Microsoft.AspNet.Mvc.Core/Rendering/Html/HtmlHelper.cs
```diff
@@ -691,8 +691,8 @@ protected virtual string GenerateDisplayText(ModelExplorer modelExplorer)
                 ViewContext,
                 modelExplorer,
                 optionLabel,
-                expression: expression,
-                selectList: selectList,
+                expression,
+                selectList,
```
dougbu: minor nit.  this call changes in part 3/3 as well.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/SelectTagHelperTest.cs
```diff
@@ -243,12 +243,9 @@ public class SelectTagHelperTest
             Assert.Equal(expectedTagName, output.TagName);
 
             Assert.NotNull(viewContext.FormContext?.FormData);
-            var keyValuePair = Assert.Single(
+            Assert.Single(
                 viewContext.FormContext.FormData,
                 entry => entry.Key == SelectTagHelper.SelectedValuesFormDataKey);
-            Assert.NotNull(keyValuePair.Value);
-            var selectedValues = Assert.IsAssignableFrom<ICollection<string>>(keyValuePair.Value);
-            Assert.InRange(selectedValues.Count, 0, 1);
```
dougbu: not worth checking anymore.  `GetCurrentValues()` returns `null` in some cases where `GenerateSelect()` would set the `out selectedValues` argument to an empty collection.  (empty collection should now occur only if (a) a value is an empty `string`, (b) `allowMultiple` is `true` and a value is `null`, or (c) a value is a `string` that is neither a defined `enum` name nor an integer.)

dougbu:
fyi I'm working on adding `DefaultHtmlGenerator` tests to cover `GetCurrentValues()`.  done otherwise

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -377,16 +380,16 @@ public virtual TagBuilder GenerateAntiForgery([NotNull] ViewContext viewContext)
             bool allowMultiple,
             object htmlAttributes)
         {
-            ICollection<string> ignored;
+            var currentValues = GetCurrentValues(viewContext, modelExplorer, expression, allowMultiple);
```
rynowak: sorry, n00bz0r. What's current values mean in this context? Are they the selected values?

rynowak:
I took a look through this and saw nothing that surprised me, just some refactoring. So a very weak :shipit: from me. Probably want to have someone more familiar than me look if possible.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
dougbu: current values of the expression -- whatever the method can find in ModelState, ViewData, the model, or the expression result.  "selected values" isn't quite correct because we aren't (yet) generating a `<select/>` or setting `SelectListItem.Selected`.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -746,18 +725,136 @@ public virtual TagBuilder GenerateAntiForgery([NotNull] ViewContext viewContext)
...
+            {
+                if (modelExplorer == null)
+                {
+                    // Html.DropDownList() and Html.ListBox() helper case.
+                    rawValue = viewContext.ViewData.Eval(expression);
+                    if (rawValue is IEnumerable<SelectListItem>)
+                    {
+                        // This ViewData item contains the fallback selectList collection for GenerateSelect().
+                        // Do not try to use this collection.
+                        rawValue = null;
```
NTaylorMullen: Kind of misleading, this bit essentially means `return null` because we immediately short-circuit.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -746,18 +725,136 @@ public virtual TagBuilder GenerateAntiForgery([NotNull] ViewContext viewContext)
...
+                    }
+                }
+                else
+                {
+                    // <select/>, Html.DropDownListFor() and Html.ListBoxFor() helper case. Do not use ViewData.
+                    rawValue = modelExplorer.Model;
+                }
+            }
+
+            if (rawValue == null)
```
NTaylorMullen: This can be moved inside the `if (rawValue == null)` block above.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -746,18 +725,136 @@ public virtual TagBuilder GenerateAntiForgery([NotNull] ViewContext viewContext)
...
+
+            // Convert raw value to a collection.
+            IEnumerable rawValues;
+            if (allowMultiple)
+            {
+                rawValues = rawValue as IEnumerable;
+                if (rawValues == null || rawValues is string)
+                {
+                    throw new InvalidOperationException(
+                        Resources.FormatHtmlHelper_SelectExpressionNotEnumerable("expression"));
```
NTaylorMullen: `nameof(expression)`

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -746,18 +725,136 @@ public virtual TagBuilder GenerateAntiForgery([NotNull] ViewContext viewContext)
...
+                if (value == null)
+                {
+                    continue;
+                }
+
+                if (!isEnum)
+                {
+                    // Add original or converted string.
+                    var stringValue = (value as string) ?? Convert.ToString(value, CultureInfo.CurrentCulture);
+                    if (!string.IsNullOrEmpty(stringValue))
```
NTaylorMullen: Curious: Why are `null`/empty values ignored and dissapear into the ether?

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -746,18 +725,136 @@ public virtual TagBuilder GenerateAntiForgery([NotNull] ViewContext viewContext)
...
+                    {
+                        currentValues.Add(stringValue);
+                    }
+
+                    // No enum special-case required.
+                    continue;
+                }
+
+                var enumValue = value as Enum;
+                if (enumValue == null)
```
NTaylorMullen: Why do we need to check if its `null`? Can't we depend on the `isEnum` property?

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -746,18 +725,136 @@ public virtual TagBuilder GenerateAntiForgery([NotNull] ViewContext viewContext)
...
+                    if (!string.IsNullOrEmpty(stringValue))
+                    {
+                        currentValues.Add(stringValue);
+                    }
+
+                    // No enum special-case required.
+                    continue;
+                }
+
+                var enumValue = value as Enum;
```
NTaylorMullen: `(Enum)value` if we can depend on the `isEnum` value

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -990,6 +1087,37 @@ internal static object GetModelStateValue(ViewContext viewContext, string key, T
...
 
+        private static Enum ConvertEnumFromInteger(object value, Type targetType)
+        {
+            try
+            {
+                return (Enum)Enum.ToObject(targetType, value);
+            }
+            catch (Exception exception)
+            {
+                if (exception is FormatException || exception.InnerException is FormatException)
```
NTaylorMullen: What populates the `InnerException` here?

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
dougbu: `Enum.ToObject()`

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
dougbu: nah, the `return` is quite visible

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
dougbu: minor but whatever

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
dougbu: :ok:

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
dougbu: - turns out this special case doesn't exist in MVC 5.  checking on the one at line 808.
- fyi `string.Empty` will result in `<option value=""></option>` and should round-trip as `null` instead unless `ConvertEmptyStringToNull` is `false` (an unusual case).  similarly `null` can round-trip as `string.Empty` in some cases.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
dougbu: can't depend on `isEnum` because that relates to the target property, not the `ModelState` or `ViewData` source.  `value` may be a `string`, an `int` (or some integral type), or something with an explicit conversion to `TEnum`.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
dougbu: can't depend on `isEnum`

dougbu:
Updated.

Now focused on #2240 i.e. we've closed (or soon will close) #2027 and I'll update the commit descriptions to reflect the different focus.  Aims are now to
- get rid of the ugly `out` parameter,
- move associated code into one place, and
- address #2240.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -746,18 +725,130 @@ public virtual TagBuilder GenerateAntiForgery([NotNull] ViewContext viewContext)
...
+            modelExplorer = modelExplorer ??
+                ExpressionMetadataProvider.FromStringExpression(expression, viewContext.ViewData, _metadataProvider);
+
+            var enumNames = modelExplorer.Metadata.EnumNamesAndValues;
+            var isEnum = modelExplorer.Metadata.IsEnum;
+            var innerType =
+                Nullable.GetUnderlyingType(modelExplorer.Metadata.ModelType) ?? modelExplorer.Metadata.ModelType;
+
+            // Convert raw value collection to strings.
+            var currentValues = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```
dougbu: previous `Ordinal` was inconsistent with the original code as well as MVC 5

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -746,18 +725,130 @@ public virtual TagBuilder GenerateAntiForgery([NotNull] ViewContext viewContext)
...
+            foreach (var value in rawValues)
```
dougbu: loop got simpler once it didn't have to generate just one `HashSet` entry per `value`

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -746,18 +725,130 @@ public virtual TagBuilder GenerateAntiForgery([NotNull] ViewContext viewContext)
...
+                // those names directly.
+                if (enumNames == null || !enumNames.ContainsKey(stringValue.Trim()))
+                {
+                    currentValues.Add(stringValue);
+                }
+
+                // Remainder handles isEnum cases. Convert.ToString() returns field names for enum values but select
+                // list may (well, should) contain integer values.
+                var enumValue = value as Enum;
+                if (isEnum && enumValue == null && value != null)
```
NTaylorMullen: should rename `isEnum` to something more specific. Realllllly confusing seeing this statement next to `enumValue == null` for someone who's unfamiliar with all of the inner workings of metadata.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -746,18 +725,130 @@ public virtual TagBuilder GenerateAntiForgery([NotNull] ViewContext viewContext)
...
+                    var valueType = value.GetType();
+                    if (typeof(long).IsAssignableFrom(valueType) || typeof(ulong).IsAssignableFrom(valueType))
+                    {
+                        // E.g. user added an int to a ViewData entry and called a string-based HTML helper.
+                        enumValue = ConvertEnumFromInteger(value, innerType);
+                    }
+                    else
+                    {
+                        // E.g. got a string from ModelState.
+                        if (!string.IsNullOrEmpty(stringValue))
```
NTaylorMullen: `else if`

NTaylorMullen:
:shipit:

dougbu:
/cc @pranavkm 
fyi I've addressed @NTaylorMullen's last couple of comments

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -746,18 +722,137 @@ public virtual TagBuilder GenerateAntiForgery([NotNull] ViewContext viewContext)
...
+                    currentValues.Add(stringValue);
+                }
+
+                // Remainder handles isEnum cases. Convert.ToString() returns field names for enum values but select
+                // list may (well, should) contain integer values.
+                var enumValue = value as Enum;
+                if (isTargetEnum && enumValue == null && value != null)
+                {
+                    var valueType = value.GetType();
+                    if (typeof(long).IsAssignableFrom(valueType) || typeof(ulong).IsAssignableFrom(valueType))
```
pranavkm: Does this need to account for other int-like types (short / ushort / byte etc?)

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -990,6 +1085,37 @@ internal static object GetModelStateValue(ViewContext viewContext, string key, T
             return UnobtrusiveValidationAttributesGenerator.GetValidationAttributes(clientRules);
         }
 
+        private static Enum ConvertEnumFromInteger(object value, Type targetType)
+        {
+            try
+            {
+                return (Enum)Enum.ToObject(targetType, value);
+            }
+            catch (Exception exception)
```
pranavkm:
```C#
catch (Exception exception) 
when (exception is FormatException || exception.InnerException is FormatException)
{
  return null;
}

File: src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs [hunk 17]
```diff
@@ -1060,48 +1186,26 @@ private static string GetInputTypeString(InputType inputType)
...
-
-            selectedValues = new HashSet<string>(values, StringComparer.OrdinalIgnoreCase);
-
             // Perform deep copy of selectList to avoid changing user's Selected property values.
             var newSelectList = new List<SelectListItem>();
             foreach (SelectListItem item in selectList)
             {
-                var newItem = new SelectListItem
+                var value = item.Value ?? item.Text;
+                var selected = currentValues.Contains(value);
```
pranavkm: Shouldn't this specify the casing? The old code would have performed a case insensitive lookup versus the new code relies on the collection's `Contains` implementation.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
dougbu: that's the reason for `IsAssignableFrom()`.  for example the tests use `int` and conversion to the appropriate `TEnum` works fine.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
dougbu: cute but not worth changing

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
dougbu: the collection comes from `GetCurrentValues()` and that provides a `HashSet` that ignores case.  it is possible to be explicit here (and in `OptionTagHelper`) using the Linq extension method.  but it isn't clear to me why we'd override a developer who _decides_ to provide a case-sensitive collection.

pranavkm: `GetCurrentValues` is virtual and `IReadOnlyCollection` does not guarantee the case-insensitive contract. Changing to use the LInq extension sounds like the correct behavior. You could also get rid of using `IReadOnlyCollection<string>` (I don't think it's a commonly exposed interface on our API surface) and use `IEnumerable<string>` if you make the change.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
pranavkm: Seems like the right fix though? Why not use a feature that's specifically designed for this scenario?

pranavkm:
:watch:

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
pranavkm: :ok:

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
dougbu: :ok:

dougbu:
@pranavkm updated

pranavkm:
:shipit:

dougbu:
9ac6ebd

PR #2241, opened 2015-03-23
[Design] Removing ComplexModelDTOModelBinder

harshgMSFT:
Not much design here other than simplification resulting from removal of complex model dto model binder. 
Reasoning: Highly coupled, very hard to extend, hard to understand and maintain due to the flow. 
Also used to model bind an unrelated model `ComplexDTOModel` which caused to break the parent-child relationship of model binding contexts.


harshgMSFT:
cc: @dougbu , @rynowak

rynowak:
I like the idea of it, I really don't like the idea of making changes here until we've got confidence that we're enforcing the right behavior with our tests. For one we lose the ability to do comparisons between our code and the original.

rynowak:
Would be good to also review https://aspnet.codeplex.com/downloads/get/110748 for any interesting tidbits.

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/MutableObjectModelBinder.cs
```diff
@@ -236,24 +231,37 @@ private static bool CanUpdateReadOnlyProperty(Type propertyType)
...
+                    bindingContext,
+                    propertyModelName,
+                    propertyMetadata);
 
-            childContext.Model = dto;
+                var modelBindingResult =
+                    await bindingContext.OperationBindingContext.ModelBinder.BindModelAsync(propertyBindingContext);
+                if (modelBindingResult == null)
+                {
+                    // Could not bind. Add a result so MutableObjectModelBinder will check for [DefaultValue].
```
dougbu: nit: you're in `MutableObjectModelBinder` :smiley_cat: think you want to refer to `ProcessResult()`

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/MutableObjectModelBinder.cs
```diff
@@ -236,24 +231,37 @@ private static bool CanUpdateReadOnlyProperty(Type propertyType)
             return true;
         }
 
-        private async Task<ModelBindingResult> CreateAndPopulateDto(
+        private async Task<IDictionary<ModelMetadata, ModelBindingResult>> BindProperties(
```
pranavkm: `BindPropertiesAsync`

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/MutableObjectModelBinder.cs
```diff
@@ -404,7 +414,7 @@ internal void ProcessDto(ModelBindingContext bindingContext, ComplexModelDto dto
 
             // For each property that ComplexModelDtoModelBinder attempted to bind, call the setter, recording
```
pranavkm: Comment no longer makes sense.

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/MutableObjectModelBinder.cs
```diff
@@ -404,7 +414,7 @@ internal void ProcessDto(ModelBindingContext bindingContext, ComplexModelDto dto
...
             // exceptions as necessary.
-            foreach (var entry in dto.Results)
+            foreach (var entry in result)
             {
                 var dtoResult = entry.Value;
```
pranavkm: Variables probably need to be renamed.

pranavkm:
Seems sound otherwise.

dougbu:
@harshgMSFT suggest closing this PR

harshgMSFT:
Closing.. will create a new one if we decide to fix this.

PR #2242, opened 2015-03-24
Stop passing around IServiceProvider in RazorPreCompiler

### LINKED ISSUE #2204: Stop passing around IServiceProvider in RazorPreCompiler
https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Razor/Razor/PreCompileViews/RazorPreCompiler.cs#L30

We should always pass explicit dependencies and not assume a populated ISP. It makes it hard to understand what dependencies are actually required.

pranavkm:
Fixes #2204

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor/Razor/PreCompileViews/RazorPreCompiler.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/Razor/PreCompileViews/RazorPreCompiler.cs
@@ -14,9 +14,7 @@ using Microsoft.AspNet.Razor.Runtime.TagHelpers;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.Framework.Caching.Memory;
-using Microsoft.Framework.DependencyInjection;
 using Microsoft.Framework.Internal;
-using Microsoft.Framework.OptionsModel;
 using Microsoft.Framework.Runtime;
 using Microsoft.Framework.Runtime.Roslyn;
 
@@ -24,32 +22,16 @@ namespace Microsoft.AspNet.Mvc.Razor
 {
     public class RazorPreCompiler
     {
-        private readonly IFileProvider _fileProvider;
-
         public RazorPreCompiler(
-            [NotNull] IServiceProvider designTimeServiceProvider,
-            [NotNull] IBeforeCompileContext compileContext,
-            [NotNull] IMemoryCache precompilationCache,
-            [NotNull] CompilationSettings compilationSettings)
-            : this(
-                compileContext,
-                designTimeServiceProvider.GetRequiredService<IAssemblyLoadContextAccessor>(),
-                designTimeServiceProvider.GetRequiredService<IOptions<RazorViewEngineOptions>>(),
-                precompilationCache,
-                compilationSettings)
-        {
-        }
-
-        private RazorPreCompiler(
             [NotNull] IBeforeCompileContext compileContext,
             [NotNull] IAssemblyLoadContextAccessor loadContextAccessor,
-            [NotNull] IOptions<RazorViewEngineOptions> optionsAccessor,
+            [NotNull] IFileProvider fileProvider,
             [NotNull] IMemoryCache precompilationCache,
             [NotNull] CompilationSettings compilationSettings)
         {
             CompileContext = compileContext;
             LoadContext = loadContextAccessor.GetLoadContext(GetType().GetTypeInfo().Assembly);
-            _fileProvider = optionsAccessor.Options.FileProvider;
+            FileProvider = fileProvider;
             CompilationSettings = compilationSettings;
             PreCompilationCache = precompilationCache;
             TagHelperTypeResolver = new PrecompilationTagHelperTypeResolver(CompileContext, LoadContext);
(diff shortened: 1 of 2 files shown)


davidfowl:
:shipit: !!! Do this with more things in general.

src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilationOptionsProviderExtension.cs
```diff
@@ -32,5 +32,24 @@ public static class CompilationOptionsProviderExtension
                                                               applicationEnvironment.Configuration)
                                           .ToCompilationSettings(applicationEnvironment.RuntimeFramework);
         }
+
+        /// <summary>
+        /// Parses the <see cref="ICompilerOptions"/> for a project specified by <see cref="IProjectContext"/>
```
dougbu: `<see cref="IProjectContext"/>` -> `<paramref name="projectContext"/>`

src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilationOptionsProviderExtension.cs
```diff
@@ -32,5 +32,24 @@ public static class CompilationOptionsProviderExtension
...
+        /// and returns a <see cref="CompilationSettings"/> used for Roslyn compilation.
+        /// </summary>
+        /// <param name="compilerOptionsProvider">
+        /// A <see cref="ICompilerOptionsProvider"/> that reads compiler options.
+        /// </param>
+        /// <param name="projectContext">The <see cref="IProjectContext"/> for the application being compiled.</param>
+        /// <returns>The <see cref="CompilationSettings"/> for the application being compiled.</returns>
+        public static CompilationSettings GetCompilationSettings(
+            [NotNull] this ICompilerOptionsProvider compilerOptionsProvider,
+            [NotNull] IProjectContext projectContext)
```
dougbu: why is this a `public` extension method and not a `private static` method in `RazorPreCompileModule`?

src/Microsoft.AspNet.Mvc/RazorPreCompileModule.cs
```diff
@@ -76,76 +64,5 @@ public virtual void BeforeCompile(IBeforeCompileContext context)
         public void AfterCompile(IAfterCompileContext context)
         {
         }
-
-        // REVIEW: Welcome back to life delegating SP!!!
```
dougbu: :dancers:

src/Microsoft.AspNet.Mvc.Razor/Razor/PreCompileViews/RazorPreCompiler.cs
```diff
@@ -60,6 +42,8 @@ public class RazorPreCompiler
         /// </summary>
         public bool GenerateSymbols { get; set; }
 
+        protected IFileProvider FileProvider { get; }
```
dougbu: how does making this a property relate to the bug you're fixing?

dougbu:
:watch: mostly for a couple of answers

src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilationOptionsProviderExtension.cs
pranavkm: The use case was for users to be use these extensions in their custom precompilation modules - it was primarily the reason we have the other extension method.

src/Microsoft.AspNet.Mvc.Razor/Razor/PreCompileViews/RazorPreCompiler.cs
pranavkm: Minor cleanup.

src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilationOptionsProviderExtension.cs
dougbu: if that's the case, better / more discoverable as `protected` methods in our base precompilation module

src/Microsoft.AspNet.Mvc.Razor/Razor/PreCompileViews/RazorPreCompiler.cs
dougbu: :ok:

src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilationOptionsProviderExtension.cs
pranavkm: Perhaps. When the code was written, the refactoring prevented doing this - I think it's in better shape now to allow this

pranavkm:
Updated.

src/Microsoft.AspNet.Mvc/RazorPreCompileModule.cs
```diff
@@ -77,75 +65,14 @@ public void AfterCompile(IAfterCompileContext context)
...
-            public FrameworkName RuntimeFramework
-            {
-                get
-                {
-                    return _originalApplicationEnvironment.RuntimeFramework;
-                }
-            }
+            return compilerOptionsProvider.GetCompilerOptions(projectContext.Name,
+                                                              projectContext.TargetFramework,
+                                                              projectContext.Configuration)
```
dougbu: nit: will be easier to maintain if you just hit <Enter> after the open paren

dougbu:
:shipit:

HaoK:
Yay, now I don't have to keep fixing this code whenever hosting changes :)

davidfowl:
@pranavkm Can you get this in?

pranavkm:
Sure. The bug was marked for beta5 - I was under the impression I had to wait until we branched. I'll get this in.

PR #2243, opened 2015-03-24
ModelMetadataIdentity cleanup

rynowak:
Adding Equals and GetHashCode implementations - the lack of these results
in a lot of boxing.

Removing dead code, not possible to create a model metadata for a
parameter anymore.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadataProvider.cs
+++ b/src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadataProvider.cs
@@ -167,10 +167,52 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
 
         private class TypeCache : ConcurrentDictionary<ModelMetadataIdentity, DefaultMetadataDetailsCache>
         {
+            public TypeCache()
+                : base(ModelMetadataIdentityComparer.Instance)
+            {
+            }
         }
 
         private class PropertiesCache : ConcurrentDictionary<ModelMetadataIdentity, DefaultMetadataDetailsCache[]>
         {
+            public PropertiesCache()
+                : base(ModelMetadataIdentityComparer.Instance)
+            {
+            }
+        }
+
+        private class ModelMetadataIdentityComparer : IEqualityComparer<ModelMetadataIdentity>
+        {
+            public static readonly ModelMetadataIdentityComparer Instance = new ModelMetadataIdentityComparer();
+
+            public bool Equals(ModelMetadataIdentity x, ModelMetadataIdentity y)
+            {
+                return
+                    x.ContainerType == y.ContainerType &&
+                    x.ModelType == y.ModelType &&
+                    x.Name == y.Name;
+            }
+
+            public int GetHashCode(ModelMetadataIdentity obj)
+            {
+                unchecked
+                {
+                    var hash = 17;
+                    hash = hash * 23 + obj.ModelType.GetHashCode();
+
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadataProvider.cs
```diff
@@ -167,10 +167,52 @@ protected virtual DefaultMetadataDetailsCache CreateTypeCacheEntry([NotNull] Mod
...
+            }
+        }
+
+        private class ModelMetadataIdentityComparer : IEqualityComparer<ModelMetadataIdentity>
+        {
+            public static readonly ModelMetadataIdentityComparer Instance = new ModelMetadataIdentityComparer();
+
+            public bool Equals(ModelMetadataIdentity x, ModelMetadataIdentity y)
+            {
+                return
```
dougbu: any use cases where `null` checks would be required?

rynowak: It's a struct...

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/ModelMetadataIdentity.cs
```diff
@@ -26,21 +25,6 @@ public static ModelMetadataIdentity ForType([NotNull] Type modelType)
         }
 
         /// <summary>
-        /// Creates a <see cref="ModelMetadataIdentity"/> for the provided <see cref="ParameterInfo"/>.
```
harshgMSFT: I thought i had removed these .. :( ... thanks for cleaning up.

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadataProvider.cs
```diff
@@ -167,10 +167,52 @@ protected virtual DefaultMetadataDetailsCache CreateTypeCacheEntry([NotNull] Mod
...
+                    x.ContainerType == y.ContainerType &&
+                    x.ModelType == y.ModelType &&
+                    x.Name == y.Name;
+            }
+
+            public int GetHashCode(ModelMetadataIdentity obj)
+            {
+                unchecked
```
harshgMSFT: isn't it by default unchecked?

rynowak: No, default is checked

harshgMSFT: From MSDN `Expressions that contain non-constant terms are unchecked by default at compile time and run time.`
https://msdn.microsoft.com/en-us/library/a569z7k8.aspx

rynowak:
Any other feedback or is this gtg after removing `checked`? (Not that I'm in a hurry, going to wait for branching to complete first)

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadataProvider.cs
dougbu: forgot that (:sheep:-ish)

dougbu:
:shipit: if @harshgMSFT is happy

harshgMSFT:
:shipit: once `unchecked` is removed.

PR #2244, opened 2015-03-24
Quick fix: Correct invalid HTML in some of our tag helper-related views

dougbu:
- few end tags for void elements e.g. `</input>`
- fair number of self-closing flow elements e.g. `<label .../>`
- most-visible change is correction of `<textarea>` in tag helper sample
- no product code changes


dougbu:
fyi only; will submit shortly

PR #2251, opened 2015-03-24
Quick fixes: Clean up test code

dougbu:
- remove useless `configuration` variables and `Configuration` instances
- remove "Review" code comment (the only product code change)
- reduce repeated code in `TestHelper` for functional tests

### DIFF
--- a/test/WebSites/ActionConstraintsWebSite/Startup.cs
+++ b/test/WebSites/ActionConstraintsWebSite/Startup.cs
@@ -21,7 +21,7 @@ namespace ActionConstraintsWebSite
 
         public void Configure(IApplicationBuilder app)
         {
-            var configuration = app.GetTestConfiguration();
+            app.GetTestConfiguration();
 
             app.UseErrorReporter();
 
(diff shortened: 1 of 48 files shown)


dougbu:
FYI only; will push changes once builds settle

test/WebSites/ActionConstraintsWebSite/Startup.cs
```diff
@@ -21,7 +21,7 @@ public void ConfigureServices(IServiceCollection services)
 
         public void Configure(IApplicationBuilder app)
         {
-            var configuration = app.GetTestConfiguration();
+            app.GetTestConfiguration();
```
rynowak: So what does this do now? presumably magic? Can we name it accordingly `app.AddTestServices()` or something reasonable?

dougbu: renamed it to `UseCultureReplacer()` since it adds just one middleware

dougbu:
@rynowak updated

rynowak:
@dougbu solid :shipit: once we get the all-clear.

dougbu:
fc1017f

PR #2253, opened 2015-03-24
Add sample and functional test for custom inline constraints

ianhong:
@harshgMSFT and @rynowak 

I added these tests. Please review them and let me know if you want to see any more tests. 

Thanks,


test/Microsoft.AspNet.Mvc.FunctionalTests/InlineConstraintTests.cs
```diff
@@ -478,6 +478,35 @@ public class InlineConstraintTests
...
 
+        [Theory]
+        // Testing custom inline constraint configured via MapRoute
+        [InlineData("software", HttpStatusCode.OK, "software From Software Controller Index")]
+        // Testing custom inline constraint updated in ConstraintMap
+        [InlineData("hardware", HttpStatusCode.NotFound, null)]
+        [InlineData("HARDWARE", HttpStatusCode.OK, "HARDWARE From Hardware Controller Index")]
+        // Testing custom inline constraint configured with Attribute Routing
+        [InlineData("consumable", HttpStatusCode.OK, "consumable From Consumable Controller Index")]
+        public async Task CustomInlineConstraint_Add_Update(
```
ianhong: There is no test for removing constraints from ConstraintMap. I think that is not a good candidate for functional tests.  Once a constraint is removed, and it is still used in the Web Site, it will simply return 500 error. So, I think I need to create another Web site only for that scenario. I think there should be no special case from custom inline constraints' perspective. 
But, please let me know if you think we need to add a test for that scenario too.

rynowak: Think about if it can be tested via a unit-test. We want to avoid creating functional tests for scenarios that throw unless it's important.

test/WebSites/InlineConstraintsWebSite/Constraints/ItemTypeConstraint.cs
```diff
@@ -0,0 +1,42 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Routing;
+
+namespace InlineConstraintsWebSite.Constraints
+{
+    public class ItemTypeConstraint : IRouteConstraint
```
rynowak: Can we think of something else interesting to do here? (Maybe look on stackoverflow for an example). This constraint is trivial so it's not super useful as a scenario.

test/WebSites/InlineConstraintsWebSite/Startup.cs
```diff
@@ -12,6 +15,27 @@ public class Startup
...
+
+            services.Configure<RouteOptions>(
+                routeOptions =>
+                {
+                    if (routeOptions.ConstraintMap.ContainsKey("servicetype"))
+                    {
+                        routeOptions.ConstraintMap["servicetype"] =
+                            typeof(ItemTypeUpperCaseConstraint);
+                    }
+                });
```
rynowak: would be cleaner to combine all of these in a statement lambda.

why do you need `if (routeOptions.ConstraintMap.ContainsKey("servicetype"))`?

test/WebSites/InlineConstraintsWebSite/Controllers/InlineConstraints_SoftwareController.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using Microsoft.AspNet.Mvc;
+
+namespace InlineConstraintsWebSite.Controllers
+{
+    public class InlineConstraints_SoftwareController : Controller
+    {
+        public IActionResult Index(string type)
+        {
+            return View("ItemType", type + " From Software Controller");
```
rynowak: Why return views here instead of just content? This seems like it makes it harder to verify with no benefit.

test/Microsoft.AspNet.Mvc.FunctionalTests/InlineConstraintTests.cs
ianhong: @rynowak 
I checked the Routing repo, but I couldn't find the unit tests for ConstraintMap.Remove. I'll add to the Routing repo an  unit test for that once this PR is checked in.

test/Microsoft.AspNet.Mvc.FunctionalTests/InlineConstraintTests.cs
```diff
@@ -478,6 +478,35 @@ public class InlineConstraintTests
...
+            var server = TestHelper.CreateServer(_app, SiteName, _configureServices);
+            var client = server.CreateClient();
+
+            // Act
+            var response = await client.GetAsync("http://localhost/producttype/index/" + type);
+            var body = await response.Content.ReadAsStringAsync();
+
+            // Assert
+            Assert.Equal(expectedResult, response.StatusCode);
+            if(expectedResult == HttpStatusCode.OK)
```
Eilon: Formatting

test/Microsoft.AspNet.Mvc.FunctionalTests/InlineConstraintTests.cs
```diff
@@ -478,6 +478,35 @@ public class InlineConstraintTests
...
+        // Testing custom inline constraint configured via MapRoute
+        [InlineData("software", HttpStatusCode.OK, "software From Software Controller Index")]
+        // Testing custom inline constraint updated in ConstraintMap
+        [InlineData("hardware", HttpStatusCode.NotFound, null)]
+        [InlineData("HARDWARE", HttpStatusCode.OK, "HARDWARE From Hardware Controller Index")]
+        // Testing custom inline constraint configured with Attribute Routing
+        [InlineData("consumable", HttpStatusCode.OK, "consumable From Consumable Controller Index")]
+        public async Task CustomInlineConstraint_Add_Update(
+            string type,
+            HttpStatusCode expectedResult,
```
Eilon: expectedResult --> expectedStatusCode ?

test/WebSites/InlineConstraintsWebSite/Constraints/ItemTypeConstraint.cs
Eilon: A classic one is something like ISBN10 and ISBN13. It's just a line or two of boring math to compute, and showing how to support both ISBN formats could be neat.

test/WebSites/InlineConstraintsWebSite/Startup.cs
ianhong: Without the if statement, it will add the constraint to the constraint map. Since this sample or test is about updating an existing constraint, I checked if the key exists first.

test/WebSites/InlineConstraintsWebSite/Constraints/ItemTypeConstraint.cs
ianhong: Thanks @Eilon 
I was looking at Stackoverflow, I think ISBN should work. :-)

test/Microsoft.AspNet.Mvc.FunctionalTests/InlineConstraintTests.cs
harshgMSFT: I do not like the fact that you had to special case here, Is it really needed to get a 404 for the updated inline constraint, there could be an action which matches the updated inline constraint. That way you wont have to special case here and you would be having a simpler Theory's inline data.

test/WebSites/InlineConstraintsWebSite/Controllers/InlineConstraints_HardwareController.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using Microsoft.AspNet.Mvc;
+
+namespace InlineConstraintsWebSite.Controllers
+{
+    public class InlineConstraints_HardwareController : Controller
+    {
+        public IActionResult Index(string type)
+        {
+            return View("ItemType", type + " From Hardware Controller");
```
harshgMSFT: same as @rynowak's comment below

test/WebSites/InlineConstraintsWebSite/Startup.cs
```diff
@@ -1,9 +1,12 @@
 ﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using InlineConstraintsWebSite.Constraints;
 using Microsoft.AspNet.Builder;
+using Microsoft.AspNet.Routing;
 using Microsoft.AspNet.Routing.Constraints;
 using Microsoft.Framework.DependencyInjection;
+using System.Collections.Generic;
```
harshgMSFT: sort usings.

test/WebSites/InlineConstraintsWebSite/Startup.cs
harshgMSFT: huh?, if you want to test an update scenario you can update an already existing one, you don't have to add and then update a constraint, for example if you wanted to update `exists` to mean something else. 
And lets combine all these adds as one lambda.

test/WebSites/InlineConstraintsWebSite/Startup.cs
```diff
@@ -38,6 +72,7 @@ public void Configure(IApplicationBuilder app)
                 routes.MapRoute("areaExists", "area-exists/{controller=Home}/{action=Index}");
                 routes.MapRoute("areaWithoutExists-area", "area-withoutexists/{area}/{controller=Home}/{action=Index}");
                 routes.MapRoute("areaWithoutExists", "area-withoutexists/{controller=Home}/{action=Index}");
+
```
harshgMSFT: nit: remove.

harshgMSFT:
:watch:

test/Microsoft.AspNet.Mvc.FunctionalTests/InlineConstraintTests.cs
ianhong: I already had a test data matching the unpdated inline constraint, so 404 is not necessary, but I still addded that to show or test it more explicitly. Let me know if you still don't like this.

test/WebSites/InlineConstraintsWebSite/Startup.cs
ianhong: I think updating an existing constraint is not a good idea. this same WebSite is used in many existing tests, and I don't like to disturb them. I'd like to keep this for that reason.

test/WebSites/InlineConstraintsWebSite/Startup.cs
```diff
@@ -12,6 +15,27 @@ public class Startup
         // Set up application services
         public void ConfigureServices(IServiceCollection services)
         {
+
+            services.Configure<RouteOptions>(
```
kichalla: nit: you can also do `services.ConfigureRouting(`

ianhong:
This PR is not ready for another iteration. I just push to move another dev machine. I'll let you know when it is ready to review.

ianhong:
@harshgMSFT and @rynowak 

I updated this PR for your and @Eilon's comments.

Thanks,

test/WebSites/InlineConstraintsWebSite/Constraints/IsbnDigitScheme10Constraint.cs
```diff
@@ -0,0 +1,67 @@
...
+using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Routing;
+
+namespace InlineConstraintsWebSite.Constraints
+{
+    public class IsbnDigitScheme10Constraint : IRouteConstraint
+    {
+        private readonly bool _turnOn;
+
+        public IsbnDigitScheme10Constraint(bool turnOn)
```
Eilon: Maybe I missed something, but what exactly is this `turnOn` parameter?

I think I was envisioning something different: a single `isbn` constraint that takes a parameter of either 10 or 13 specifying the format. I don't think we need two different constraints.

test/WebSites/InlineConstraintsWebSite/Constraints/IsbnDigitScheme10Constraint.cs
```diff
@@ -0,0 +1,67 @@
...
+            if (!values.TryGetValue(routeKey, out value))
+            {
+                return false;
+            }
+
+            var isbnNumber = value as string;
+
+            if (isbnNumber == null
+                || isbnNumber.Length != 10
+                || isbnNumber.Any(n => !char.IsDigit(n)))
```
Eilon: `char.IsDigit` returns `true` for *any* type of decimal digit, including all the characters here: http://www.fileformat.info/info/unicode/category/Nd/list.htm

I think here you'd want to specifically check `0` - `9`.

Eilon: ISBN10 also allows the character `X` for the check digit. How is that handled here?

Eilon: See this: http://en.wikipedia.org/wiki/International_Standard_Book_Number#ISBN-10_check_digits

test/WebSites/InlineConstraintsWebSite/Constraints/IsbnDigitScheme10Constraint.cs
```diff
@@ -0,0 +1,67 @@
...
+            {
+                return false;
+            }
+
+            var sum = 0;
+            Func<char, int> convertToInt = (char n) => (int)char.GetNumericValue(n);
```
Eilon: Similar to earlier Unicode comment, just doing plain old ASCII math is best: just subtract `0` from the char to get its numeric value because only ASCII digits are allowed anyway.

test/WebSites/InlineConstraintsWebSite/Startup.cs
```diff
@@ -12,6 +13,26 @@ public class Startup
...
+                routeOptions => routeOptions.ConstraintMap.Add(
+                    "IsbnDigitScheme10",
+                    typeof(IsbnDigitScheme10Constraint)));
+
+            services.ConfigureRouting(
+                routeOptions => routeOptions.ConstraintMap.Add(
+                    "IsbnDigitScheme13",
+                    typeof(IsbnDigitScheme10Constraint)));
+
+            services.ConfigureRouting(
```
Eilon: Not sure I understand the purpose of this extra configuration... can you add a comment that explains its purpose?

ianhong: @Eilon I'll add a comments like this is for testing the ConstraintMap for update. I talked to @rynowak  before and that was the one of the intentions of the bug.

test/WebSites/InlineConstraintsWebSite/Constraints/IsbnDigitScheme10Constraint.cs
ianhong: The reason why I have two different constraints is for testing updating scenario for constraintmap. And, I also didn't want to update an existing constraint since this WebSite is used for the rest of the functional tests too. 
In this case, turnOn isn't useful, but.. this is just to show how users can pass a parameter to the constructor.

Let me know if you think we don't have to test the update scenario or...we don't have to show an example for passing a parameter to the constraint's constructor.

Eilon: Hmm then maybe I'd make it a more realistic example, such as an option that allows optional hyphens in the number or something. So with `allowDashes=true` it would allow `123-456-789-X` and with `allowDashes=false` it would require `123456789X`.

test/WebSites/InlineConstraintsWebSite/Startup.cs
Eilon: OK. I think a comment would be very helpful here because the purpose is not obvious.

ianhong:
@Eilon, Update per your comments!

test/WebSites/InlineConstraintsWebSite/Constraints/IsbnDigitScheme10Constraint.cs
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        public bool Match(
+            HttpContext httpContext,
+            IRouter route,
+            string routeKey,
+            IDictionary<string, object> values,
+            RouteDirection routeDirection)
+        {
+            var isbnRegExpression = @"^(\d{9})(\d|X)\z";
```
Eilon: Instead of `\z` use `$` to match end-of-string (it's odd to mix `^` and `\z` because `^`/`$` and `\A`/`\Z` are matching sets - never mixed).

test/WebSites/InlineConstraintsWebSite/Constraints/IsbnDigitScheme10Constraint.cs
```diff
@@ -0,0 +1,78 @@
...
+
+            if (_allowDashes)
+            {
+                isbnRegExpression = @"^(\d{1})[-]\d{3}[-]\d{5}[-][\d|X]\z";
```
Eilon: There are very few rules about where dashes are allowed in ISBNs.

test/WebSites/InlineConstraintsWebSite/Constraints/IsbnDigitScheme10Constraint.cs
```diff
@@ -0,0 +1,78 @@
...
+            {
+                return false;
+            }
+
+            var sum = 0;
+            Func<char, int> convertToInt = (char n) => (int)n - (int)'0';
+
+            if (_allowDashes)
+            {
+                isbnNumber = isbnNumber = isbnNumber.Replace("-", string.Empty);
```
Eilon: Setting `isbnNumber` twice?

test/WebSites/InlineConstraintsWebSite/Constraints/IsbnDigitScheme10Constraint.cs
```diff
@@ -0,0 +1,78 @@
...
+
+            object value;
+
+            if (!values.TryGetValue(routeKey, out value))
+            {
+                return false;
+            }
+
+            var isbnNumber = value as string;
+            var isbnRegEx = new Regex(isbnRegExpression, RegexOptions.IgnoreCase);
```
Eilon: Because there are only ever two regexes and they're both constant, can just store both variations statically as Regex objects and just choose which one to use at runtime. Also, I think you can just call `.IsMatch()` instead of calling `Match(...).Success`.

test/WebSites/InlineConstraintsWebSite/Constraints/IsbnDigitScheme13Constraint.cs
```diff
@@ -0,0 +1,63 @@
...
+            object value;
+
+            if (!values.TryGetValue(routeKey, out value))
+            {
+                return false;
+            }
+
+            var isbnNumber = value as string;
+
+            Func<char, bool> isValidIsbnChar = (char c) =>
```
Eilon: With chars I think you can just say `c >= '0' && c <= '9'` no? Also don't need to declare a lambda for this - just stick it inline.

ianhong:
@Eilon 
I updated per your comments. I removed regex since the allowed positions of '-' in isbn numbers made it too complex.

test/Microsoft.AspNet.Mvc.FunctionalTests/InlineConstraintTests.cs
```diff
@@ -478,10 +478,30 @@ public class InlineConstraintTests
             Assert.Equal(HttpStatusCode.NotFound, response.StatusCode);
         }
 
+        [Theory]
+        // Testing custom inline constraint updated in ConstraintMapg
```
Eilon: Typo

test/WebSites/InlineConstraintsWebSite/Constraints/IsbnDigitScheme13Constraint.cs
```diff
@@ -0,0 +1,57 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Routing;
+
+namespace InlineConstraintsWebSite.Constraints
+{
+    public class IsbnDigitScheme13Constraint : IRouteConstraint
```
Eilon: I believe ISBN13 also has optional dashes. At least the examples I found seemed to sometimes have dashes (but with 5 segments, instead of 4).

test/WebSites/InlineConstraintsWebSite/Constraints/IsbnDigitScheme13Constraint.cs
```diff
@@ -0,0 +1,57 @@
...
+
+            if (isbnNumber == null
+                || isbnNumber.Length != 13
+                || isbnNumber.Any(n => n < '0' || n > '9'))
+            {
+                return false;
+            }
+
+            var sum = 0;
+            var multipliedBy = new int[] { 1, 3 };
```
Eilon: Can store this in a `private static readonly`.

test/WebSites/InlineConstraintsWebSite/Constraints/IsbnDigitScheme13Constraint.cs
ianhong: @Eilon 
Since this is not a ISBN check sample, and we also show how to use the constraint parameter with the ISBN10 checker, I intentinally didn't include this logic here to make this sample simple. 
Please let me know if you still want to add that.

ianhong:
@Eilon 
I update the sample to use private static readony.

Thanks

Eilon:
:shipit: thanks!

ianhong:
21031a3aa82ac6618da6320701be7ea0af0fdf34

PR #2254, opened 2015-03-24
Unrendered sections does not throw when redefined and rendered in nested

### LINKED ISSUE #2252: Unrendered sections does not throw when redefined and rendered in nested layout.
Follow up #2226 
For sections that are redefined but not rendered in the outer definition.
```C#
// NestedLayout
@{ Layout = "OuterLayout"; }
@section scripts {
   // stuff
}

// OuterLayout
@{ Layout = "BaseLayout"; }
@section scripts { 
  // Don't render inner scripts
}

// BaseLayout
@RenderSection("scripts");
```
In this scenario, the scripts section defined by NestedLayout is never rendered but the view would happily succeed.

cc' @pranavkm

pranavkm:
layout.

Fixes #2252

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
+using System.Linq;
 using System.Security.Claims;
 using System.Text;
 using System.Threading.Tasks;
@@ -47,15 +48,6 @@ namespace Microsoft.AspNet.Mvc.Razor
                 }
 
                 return ViewContext.HttpContext;
-            }
-        }
-
-        /// <inheritdoc />
-        public IEnumerable<string> RenderedSections
-        {
-            get
-            {
-                return _renderedSections;
             }
         }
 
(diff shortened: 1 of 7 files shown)


pranavkm:
cc @ajaybhargavb

ajaybhargavb:
Not fully related to this PR: Just checked MVC 5 supports having `RenderBody` inside a section which is rendered in another layout. Do we plan to have the same support? Right now our code will throw when there is no explicit `RenderBody` call.

pranavkm:
Can't think of any reason why this might be problematic. @Eilon, any thoughts on @ajaybhargavb's scenarios?

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -735,6 +736,22 @@ public void EnsureBodyWasRendered()
...
         }
 
+        /// <inheritdoc />
+        public void EnsureSectionsWereRendered()
+        {
+            // If PreviousSectionWriters is set, ensure all defined sections were rendered.
+            if (PreviousSectionWriters != null)
+            {
+                var sectionsNotRendered = PreviousSectionWriters.Keys.Except(_renderedSections,
+                                                                             StringComparer.OrdinalIgnoreCase);
```
dougbu: nit: stylish, yes; maintainable indentation, no

Eilon:
@ajaybhargavb so you're saying in MVC 5 a "random" section could call RenderBody() and that would work? If that's the case I'm guessing that was quite unintentional... If there's a breaking change in that scenario I'm totally fine shipping with that for now and if it turns out someone cared we can revisit... but I'd be surprised.

ajaybhargavb:
Actually it doesn't throw even when there is no `RenderBody()` call. It is ignored. It throws only for unrendered sections. @pranavkm, was this behavior change intended in MVC 6?

Eilon:
I *think* RenderBody was meant to be optional because there might not be an actual body, so requiring rendering it would be bad. Some users always use explicit sections, whereas some use body+sections, and some use body only.

pranavkm:
Does seem like a change in behavior - we only required bodies to be rendered if no sections were defined. I can address this as part of the change.

ajaybhargavb:
:+1:

Eilon:
Ohh yeah that's right there was that logic.

pranavkm:
Updated.

ajaybhargavb:
:shipit:

pranavkm:
cc @dougbu

dougbu:
:shipit: (didn't ask for you to wait for me)

PR #2255, opened 2015-03-24
Cleanup: Moving files from src/Microsoft.AspNet.Mvc.Razor/Razor/PreCompi...

pranavkm:
...leViews ->

src/Microsoft.AspNet.Mvc.Razor/Precompilation

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilationFailedException.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilationFailedException.cs
@@ -7,7 +7,7 @@ using System.Linq;
 using Microsoft.Framework.Runtime;
 using Microsoft.Framework.Internal;
 
-namespace Microsoft.AspNet.Mvc.Razor
+namespace Microsoft.AspNet.Mvc.Razor.Compilation
 {
     /// <summary>
     /// An <see cref="Exception"/> thrown when accessing the result of a failed compilation.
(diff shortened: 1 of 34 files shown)


Eilon:
Should the namespaces be updated to reflect the path? Namespaces are supposed to match paths (they didn't before the file move, but they should match if we're cleaning this up).

pranavkm:
They are all in the same namespace `Microsoft.AspNet.Mvc.Razor` and don't use the folder hierarchy for namespacing. I could change that too - although we have precedence against having folder based namespaces in other code in Mvc \ Razor.

The primary motivation for doing this change was to remove things that were nested two layers deep.

Eilon:
These should use folder hierarchy as namespace hierarchy. Why would we make an exception here? If it's too many files in a folder, then it's too many files in the namespace.

pranavkm:
We've been using folders to keep projects logically segmented without actually namespacing them accordingly. For instance:

* https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/ActionConstraints/ActionConstraintContext.cs#L7
* https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.ModelBinding/Binders/ArrayModelBinder.cs#L8
* https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilationResult.cs#L8

Should I file a separate bug to update all of these \ move them to the root directory if we don't want to namespace based on folder hierarchy?

Eilon:
Just fix the one at hand now. The other ones are also just wrong :smile: Can file a bug on the other ones and we'll deal with it later.

pranavkm:
:ok:

pranavkm:
Updated.

Eilon:
:shipit:

PR #2257, opened 2015-03-24
Prevent RazorErrorExtensions.ToDiagnostics from throwing when it

pranavkm:
encounters SourceLocation.Undefined \ negative error lengths


src/Microsoft.AspNet.Mvc.Razor/Precompilation/RazorErrorExtensions.cs
```diff
@@ -20,15 +22,19 @@ public static Diagnostic ToDiagnostics([NotNull] this RazorError error, [NotNull
...
-            var linePositionEnd = new LinePosition(error.Location.LineIndex,
-                                                   error.Location.CharacterIndex + error.Length);
-            var linePositionSpan = new LinePositionSpan(linePositionStart, linePositionEnd);
-
-            var location = Location.Create(filePath, textSpan, linePositionSpan);
+            var location = error.Location;
+            if (location == SourceLocation.Undefined)
+            {
+                location = SourceLocation.Zero;
+            }
```
dougbu: this seems like a localized fix for a more general problem.  seem to see `Undefined` locations most often in cases where the error is in another file.  so we need to pass more information along and emit diagnostics showing the actual locations, not the start of the current file.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/RazorErrorExtensions.cs
```diff
@@ -20,15 +22,19 @@ public static Diagnostic ToDiagnostics([NotNull] this RazorError error, [NotNull
...
+            var length = Math.Max(0, error.Length);
```
dougbu: in what scenarios other than when `location == SourceLocation.Undefined` could the `error.Length` be negative?

test/Microsoft.AspNet.Mvc.Razor.Test/Precompilation/RazorErrorExtensionsTest.cs
```diff
@@ -0,0 +1,66 @@
...
+                    { SourceLocation.Undefined, 0 },
+                    { SourceLocation.Zero, -1 },
+                    { SourceLocation.Zero, 0 },
+                };
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(ToDiagnostic_DoesNotThrowWhenRazorErrorLocationIsZeroOrUndefinedData))]
+        public void ToDiagnostic_DoesNotThrowWhenRazorErrorLocationIsZeroOrUndefined(
```
dougbu: nit: `DoesNotThrow` is as meaningless as the assertion that got pulled out of Xunit.  suggest `Succeeds`

test/Microsoft.AspNet.Mvc.Razor.Test/Precompilation/RazorErrorExtensionsTest.cs
```diff
@@ -0,0 +1,66 @@
...
+            Assert.Equal(0, span.EndLinePosition.Line);
+            Assert.Equal(0, span.EndLinePosition.Character);
+        }
+
+        [Fact]
+        public void ToDiagnostic_ConvertsRazorErrorLocation_ToSourceLineMappings()
+        {
+            // Arrange
+            var sourceLocation = new SourceLocation(30, 10, 1);
+            var error = new RazorError("some message", sourceLocation, 5);
```
dougbu: nit: named parameters would make these two lines easier to read

src/Microsoft.AspNet.Mvc.Razor/Precompilation/RazorErrorExtensions.cs
dougbu: put another way, the current changes should be a defence-in-depth _part_ of a more general fix that gets the errors pointing to their correct locations.

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.Razor/Precompilation/RazorErrorExtensions.cs
pranavkm: https://github.com/aspnet/Razor/issues/334

src/Microsoft.AspNet.Mvc.Razor/Precompilation/RazorErrorExtensions.cs
pranavkm: No idea, but seemed like a sensible thing to do since we don't do an error check for `length` in `RazorError`.

pranavkm:
Sorry, I deleted the base branch which closed this PR. I'll send a new one.

Issue #2271, opened 2015-03-25
Adding @using in _GlobalImports shows squigglies in VS


rustd:
I have the following in _GlobalImports.cshtml
```C#
@using NoAut
@using Microsoft.Framework.OptionsModel
@addTagHelper "*, Microsoft.AspNet.Mvc.TagHelpers"
```

In my View I am using it as 
```C#
@inject IOptions<AppSettings> AppSettings
```

In VS I get squiggles in my Views. It works at runtime. Chatted with Taylor and he will look into this

NTaylorMullen:
/cc @danroth27 @Eilon @DamianEdwards Found why this is happening. It's an easy fix and is ultimately caused by the same issue that resulted in the directives `@tagHelperPrefix`, `@addTagHelper` and `@removeTagHelper` no longer working in `_GloblImport.cshtml` during design time: just need to apply the fix in an additional location. 

I'd love to get this one into release - super low impact since it only affects tooling.

NTaylorMullen:
Super quick fix to this out for PR now: assigned to beta5, feel free to move milestone if you decide there's enough value in getting it into beta4.

Eilon:
Clearing the milestone so we can triage in the morning.

NTaylorMullen:
9402f1485ccc245f874d5d530f8c8a1ad54f3ced

PR #2272, opened 2015-03-25
Normalize paths for MvcRazorHost DecorateCodeBuilder's ChunkInheritanceU...

NTaylorMullen:
...tility call.

- This solves tooling trying to resolve ChunkInheritanceUtility bits for chunks: Aka can't inherit `@using`, `@inject` etc. from _GlobalImport.cshtml
- Added test to validate paths are normalized for DecorateCodeBuilder.

#2271

### DIFF
--- a/test/Microsoft.AspNet.Mvc.Razor.Host.Test/MvcRazorHostTest.cs
+++ b/test/Microsoft.AspNet.Mvc.Razor.Host.Test/MvcRazorHostTest.cs
@@ -47,6 +47,39 @@ namespace Microsoft.AspNet.Mvc.Razor
             Assert.Equal("src/file.cshtml", chunkInheritanceUtility.InheritedCodeTreePagePath, StringComparer.Ordinal);
         }
 
+        [Theory]
+        [InlineData("//")]
+        [InlineData("C:/")]
+        [InlineData(@"\\")]
+        [InlineData(@"C:\")]
+        public void DecorateCodeBuilder_DesignTimeRazorPathNormalizer_NormalizesChunkInheritanceUtilityPaths(
+            string rootPrefix)
+        {
+            // Arrange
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNet.Mvc.Razor.Host.Test/MvcRazorHostTest.cs
```diff
@@ -47,6 +47,39 @@ public class MvcRazorHostTest
...
+        [Theory]
+        [InlineData("//")]
+        [InlineData("C:/")]
+        [InlineData(@"\\")]
+        [InlineData(@"C:\")]
+        public void DecorateCodeBuilder_DesignTimeRazorPathNormalizer_NormalizesChunkInheritanceUtilityPaths(
+            string rootPrefix)
+        {
+            // Arrange
+            var rootedAppPath = $"{rootPrefix}SomeComputer/Location/Project/";
```
pranavkm: Could you add a test where the `rootedFilePath` and `rootedAppPath` don't have the same line ending formats

For instance: `filePath`: "/foo/bar/baz.cshtml", `appPath`: "foo\bar\baz.cshtml"

NTaylorMullen: Ah, as of right I don't believe that can ever happen as far as tooling is concerned. What scenario did you envision causing that behavior?

pranavkm: Ok. Just concerned given that the types are available to user code and consequently they could be used in this manner.

pranavkm:
:shipit:

NTaylorMullen:
9402f1485ccc245f874d5d530f8c8a1ad54f3ced
b0b0836395e23dd9c02cd6b2450aabc373cb60a0

PR #2277, opened 2015-03-26
Change casing of the "compiler" folder

victorhurdugaci:
Due to a recent change in resource patterns, the `compiler\resources` folder must be lower case. This is causing the current build failure in MVC.

I am not 100% sure but after this change you might have to force sync on Windows to pick the new changes.

cc @troydai @NTaylorMullen


troydai:
:shipit:

pranavkm:
:shipit:

NTaylorMullen:
:shipit:

PR #2281, opened 2015-03-28
Let VS do its thing

dougbu:
- newline after BOM and slight reordering of Mvc.sln
- add `Service/Include` properties to some test .xproj files
 - also add BOM in a few cases


dougbu:
@BillHiebert will newer VS versions keep making changes like this?  if yes, I'll get it in.  if no, guess we'll try to avoid mixing these temp changes into other commits...

davidfowl:
FWIW, I'm seeing this on dnx as well with the latest VS builds. It's pretty damn annoying.

dougbu:
trying to find the thread but think @BillHiebert or someone said this related to letting VS know there's an associated test project.  the upshot might have been along the lines of "well, these files are for VS".

probably means we need an aspnet/DNX PR similar to this one.  but let's wait 'til we hear...

davidfowl:
Seems like it might be every test project

dougbu:
found the [thread](https://github.com/aspnet/aspnet.xunit/pull/5#issuecomment-85245665).  helpful though @BillHiebert didn't mention permanence.

@sayedihashimi asked
> @billhiebert do you know how the Service item is added to the project file?

and @BillHiebert responded
> It is a VS thing. It is a way to have a project which is using a specific service load the service when the project opens. This avoids the service having to be loaded for all projects.

wow, and someone named @davidfowl had earlier commented
> It's ok. The project file exists for crap like this.

rynowak:
:shipit: :shipit: :shipit:

dougbu:
ok @rynowak one sec ...

PR #2282, opened 2015-03-29
React to hosting changes

HaoK:
MVC changes for https://github.com/aspnet/Hosting/pull/210

cc @davidfowl


test/Microsoft.AspNet.Mvc.FunctionalTests/TestHelper.cs
```diff
@@ -114,7 +117,10 @@ public static TestServer CreateServer(Action<IApplicationBuilder> builder, strin
                 applicationBasePath,
                 applicationWebSiteName);
             services.AddInstance<IApplicationEnvironment>(environment);
-            services.AddInstance<IHostingEnvironment>(new HostingEnvironment(environment));
+            var hostingEnvironment = new HostingEnvironment();
```
HaoK: Maybe we should consider a helper to initialize all 3 of these like the ctor used to with appEnv, hostingEnv.Initialize(applicaitonBasePath) that will set WebRoot/FileProvider?

dougbu: probably a good idea.  that should avoid the ugly use here of `Microsoft.AspNet.Hosting.Internal` and `HostingUtilities`.

dougbu:
:watch: for a simpler functional test setup

HaoK:
@dougbu @rynowak FYI looks like you should be able to simplify TestHelper further once this set of hosting changes go in, I think you can probably get rid of your own instead of HostingEnvironment since its fully get/settable now.  Also with IStartupFilter, you probably can automatically inject the culture replacer as part of test setup instead of having 450 instances of that line of code :)

test/Microsoft.AspNet.Mvc.FunctionalTests/TestHelper.cs
```diff
@@ -114,7 +114,9 @@ public static TestServer CreateServer(Action<IApplicationBuilder> builder, strin
                 applicationBasePath,
                 applicationWebSiteName);
             services.AddInstance<IApplicationEnvironment>(environment);
```
HaoK: Perhaps replacing the application environment might no longer be needed now that hosting enviroment doesn't depend on it

PR #2285, opened 2015-03-29
Rename "dnu bundle" to "dnu publish"

ChengTian:
https://github.com/aspnet/dnx/issues/1474

@muratg @davidfowl

### DIFF
--- a/test/WebSites/LoggingWebSite/project.json
+++ b/test/WebSites/LoggingWebSite/project.json
@@ -4,7 +4,7 @@
     "exclude": [
         "wwwroot"
     ],
-    "bundleExclude": [
+    "publishExclude": [
         "**.kproj",
         "**.user",
         "**.vspscc"


test/WebSites/LoggingWebSite/project.json
```diff
@@ -4,7 +4,7 @@
     "exclude": [
         "wwwroot"
     ],
-    "bundleExclude": [
+    "publishExclude": [
```
dougbu: last time we renamed project.json properties, we added warnings for the old ones but supported them for a while.  is that not happening for this change?

davidfowl:
:shipit:

test/WebSites/LoggingWebSite/project.json
davidfowl: Yes it'll happen

muratg:
:shipit:

PR #2288, opened 2015-03-29
MVC portion of aspnet/Razor#335

dougbu:
- set correct `MarkAsHtmlEncodedMethodName` value in `MvcRazorHost`
- handle `HtmlString` values in `TagHelperOutput.Attributes` in `RazorPage`
 - special-case double-quotes in `HtmlString` values
 - add `static WriteTo()` method for use in tag helpers
- handle non-`string` `output.Attributes` values in tag helpers
- provide a `TagHelperContent.Append(object, ...)` extension method
- add `LinkTagHelper.Href` and `ScriptTagHelper.Src` properties
 - avoid Razor HTML-encoding these attribute values before their use
- add `JavaScriptEncoder` properties in `LinkTagHelper` and `ScriptTagHelper`
 - allow encoding testing without unit testing the default encoder
- handle MVC and Razor changes for this bug in existing tests
- add functional tests of encodings
- add test encoders to TestCommon project

nits:
- correct `InputTagHelper` to pass `type=""` through unchanged
- set correct `TagHelperContentTypeName` value in `MvcRazorHost`
- remove unnecessary `FormTagHelper.Method` and `OptionTagHelper.Selected` properties
- remove complex ternaries and `ShouldAddFileVersion()` methods
- add a few debug assertions
- fix some odd wrapping
- remove or `#if`-out unused `using`s
- remove trailing whitespace

todo:
- find a reliable way to handle bound attributes that are URLs
 - currently HTML- or JavaScript-encoded, incorrectly escaping `&` in HTML case

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor.Host/MvcRazorHost.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor.Host/MvcRazorHost.cs
@@ -3,7 +3,9 @@
 
 using System.Collections.Generic;
 using System.IO;
+#if NET45
 using Microsoft.AspNet.FileProviders;
+#endif
 using Microsoft.AspNet.Mvc.Razor.Directives;
 using Microsoft.AspNet.Mvc.Razor.Internal;
 using Microsoft.AspNet.Razor;
@@ -18,6 +20,8 @@ namespace Microsoft.AspNet.Mvc.Razor
     public class MvcRazorHost : RazorEngineHost, IMvcRazorHost
     {
         private const string BaseType = "Microsoft.AspNet.Mvc.Razor.RazorPage";
+        private const string HtmlHelperPropertyName = "Html";
+
         private static readonly string[] _defaultNamespaces = new[]
         {
             "System",
@@ -28,7 +32,7 @@ namespace Microsoft.AspNet.Mvc.Razor
         };
         private static readonly Chunk[] _defaultInheritedChunks = new[]
         {
-            new InjectChunk("Microsoft.AspNet.Mvc.Rendering.IHtmlHelper<TModel>", "Html"),
+            new InjectChunk("Microsoft.AspNet.Mvc.Rendering.IHtmlHelper<TModel>", HtmlHelperPropertyName),
             new InjectChunk("Microsoft.AspNet.Mvc.IViewComponentHelper", "Component"),
             new InjectChunk("Microsoft.AspNet.Mvc.IUrlHelper", "Url"),
         };
@@ -76,6 +80,8 @@ namespace Microsoft.AspNet.Mvc.Razor
                     ScopeManagerBeginMethodName = nameof(TagHelperScopeManager.Begin),
                     ScopeManagerEndMethodName = nameof(TagHelperScopeManager.End),
 
+                    TagHelperContentTypeName = nameof(TagHelperContent),
+
                     // Can't use nameof because RazorPage is not accessible here.
                     CreateTagHelperMethodName = "CreateTagHelper",
                     StartTagHelperWritingScopeMethodName = "StartTagHelperWritingScope",
@@ -83,6 +89,9 @@ namespace Microsoft.AspNet.Mvc.Razor
 
                     WriteTagHelperAsyncMethodName = "WriteTagHelperAsync",
                     WriteTagHelperToAsyncMethodName = "WriteTagHelperToAsync",
+
+                    // Can't use nameof because IHtmlHelper is (also) not accessible here.
+                    MarkAsHtmlEncodedMethodName = HtmlHelperPropertyName + ".Raw",
                 })
             {
                 ResolveUrlMethodName = "Href",
--- a/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
@@ -294,11 +294,10 @@ namespace Microsoft.AspNet.Mvc.Razor
 
                 foreach (var attribute in tagHelperOutput.Attributes)
                 {
-                    var value = HtmlEncoder.HtmlEncode(attribute.Value);
                     writer.Write(' ');
                     writer.Write(attribute.Key);
                     writer.Write("=\"");
-                    writer.Write(value);
+                    WriteTo(writer, HtmlEncoder, attribute.Value, escapeQuotes: true);
                     writer.Write('"');
                 }
 
@@ -366,26 +365,67 @@ namespace Microsoft.AspNet.Mvc.Razor
         /// </remarks>
         public virtual void WriteTo([NotNull] TextWriter writer, object value)
         {
-            if (value != null && value != HtmlString.Empty)
+            WriteTo(writer, HtmlEncoder, value, escapeQuotes: false);
+        }
+
+        /// <summary>
+        /// Writes the specified <paramref name="value"/> with HTML encoding to given <paramref name="writer"/>.
+        /// </summary>
+        /// <param name="writer">The <see cref="TextWriter"/> instance to write to.</param>
+        /// <param name="encoder">The <see cref="IHtmlEncoder"/> to use when encoding <paramref name="value"/>.</param>
+        /// <param name="value">The <see cref="object"/> to write.</param>
+        /// <param name="escapeQuotes">
+        /// If <c>true</c> escapes double quotes in a <paramref name="value"/> of type <see cref="HtmlString"/>.
+        /// Otherwise writes <see cref="HtmlString"/> values as-is.
+        /// </param>
+        /// <remarks>
+        /// <paramref name="value"/>s of type <see cref="HtmlString"/> are written without encoding and the
+        /// <see cref="HelperResult.WriteTo(TextWriter)"/> is invoked for <see cref="HelperResult"/> types.
+        /// For all other types, the encoded result of <see cref="object.ToString"/> is written to the
+        /// <paramref name="writer"/>.
+        /// </remarks>
+        public static void WriteTo(
+            [NotNull] TextWriter writer,
+            [NotNull] IHtmlEncoder encoder,
+            object value,
+            bool escapeQuotes)
+        {
+            if (value == null || value == HtmlString.Empty)
             {
-                var helperResult = value as HelperResult;
-                if (helperResult != null)
-                {
-                    helperResult.WriteTo(writer);
-                }
-                else
-                {
-                    var htmlString = value as HtmlString;
-                    if (htmlString != null)
-                    {
-                        writer.Write(htmlString);
-                    }
-                    else
-                    {
-                        WriteTo(writer, value.ToString());
-                    }
-                }
+                return;
             }
+
+            var helperResult = value as HelperResult;
+            if (helperResult != null)
+            {
+                helperResult.WriteTo(writer);
+                return;
+            }
+
+            var htmlString = value as HtmlString;
+            if (htmlString != null)
+            {
+                if (escapeQuotes)
+                {
+                    // In this case the text likely came directly from the Razor source. Since the original string is
+                    // an attribute value that may have been quoted with single quotes, must handle any double quotes
+                    // in the value. Writing the value out surrounded by double quotes.
+                    //
+                    // Do not combine following condition with check of escapeQuotes; htmlString.ToString() can be
+                    // expensive when the HtmlString is created with a StringCollectionTextWriter.
+                    var stringValue = htmlString.ToString();
+                    if (stringValue.Contains("\""))
+                    {
+                        writer.Write(stringValue.Replace("\"", "&quot;"));
+                        return;
+                    }
+                }
+
+                htmlString.WriteTo(writer);
+                return;
+            }
+
+            WriteTo(writer, encoder, value.ToString());
         }
 
         /// <summary>
@@ -395,9 +435,14 @@ namespace Microsoft.AspNet.Mvc.Razor
         /// <param name="value">The <see cref="string"/> to write.</param>
         public virtual void WriteTo([NotNull] TextWriter writer, string value)
         {
+            WriteTo(writer, HtmlEncoder, value);
+        }
+
+        private static void WriteTo(TextWriter writer, IHtmlEncoder encoder, string value)
+        {
             if (!string.IsNullOrEmpty(value))
             {
-                HtmlEncoder.HtmlEncode(value, writer);
+                encoder.HtmlEncode(value, writer);
             }
         }
 
@@ -764,34 +809,6 @@ namespace Microsoft.AspNet.Mvc.Razor
             if (PreviousSectionWriters == null)
             {
                 throw new InvalidOperationException(Resources.FormatRazorPage_MethodCannotBeCalled(methodName));
-            }
-        }
-
-        private class TagHelperContentWrapperTextWriter : TextWriter
-        {
-            public TagHelperContentWrapperTextWriter(Encoding encoding)
-            {
-                Content = new DefaultTagHelperContent();
-                Encoding = encoding;
-            }
-
-            public TagHelperContent Content { get; }
-
-            public override Encoding Encoding { get; }
-
-            public override void Write(string value)
-            {
-                Content.Append(value);
-            }
-
-            public override void Write(char value)
-            {
-                Content.Append(value.ToString());
-            }
-
-            public override string ToString()
-            {
-                return Content.ToString();
             }
         }
     }
--- a/src/Microsoft.AspNet.Mvc.Razor/TagHelperContentWrapperTextWriter.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/TagHelperContentWrapperTextWriter.cs
@@ -1,0 +1,62 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.IO;
+using System.Text;
+using Microsoft.AspNet.Razor.Runtime.TagHelpers;
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.AspNet.Mvc.Razor
+{
+    /// <summary>
+    /// <see cref="TextWriter"/> implementation which writes to a <see cref="TagHelperContent"/> instance.
+    /// </summary>
+    public class TagHelperContentWrapperTextWriter : TextWriter
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TagHelperContentWrapperTextWriter"/> class.
+        /// </summary>
+        /// <param name="encoding">The <see cref="Encoding"/> in which output is written.</param>
+        public TagHelperContentWrapperTextWriter([NotNull] Encoding encoding)
+            : this(encoding, new DefaultTagHelperContent())
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TagHelperContentWrapperTextWriter"/> class.
+        /// </summary>
+        /// <param name="encoding">The <see cref="Encoding"/> in which output is written.</param>
+        /// <param name="content">The <see cref="TagHelperContent"/> to write to.</param>
+        public TagHelperContentWrapperTextWriter([NotNull] Encoding encoding, [NotNull] TagHelperContent content)
+        {
+            Content = content;
+            Encoding = encoding;
+        }
+
+        /// <summary>
+        /// The <see cref="TagHelperContent"/> this <see cref="TagHelperContentWrapperTextWriter"/> writes to.
+        /// </summary>
+        public TagHelperContent Content { get; }
+
+        /// <inheritdoc />
+        public override Encoding Encoding { get; }
+
+        /// <inheritdoc />
+        public override void Write(string value)
+        {
+            Content.Append(value);
+        }
+
+        /// <inheritdoc />
+        public override void Write(char value)
+        {
+            Content.Append(value.ToString());
+        }
+
+        /// <inheritdoc />
+        public override string ToString()
+        {
+            return Content.ToString();
+        }
+    }
+}
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/FormTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/FormTagHelper.cs
@@ -27,7 +27,7 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
         // Protected to ensure subclasses are correctly activated. Internal for ease of use when testing.
         [Activate]
         protected internal IHtmlGenerator Generator { get; set; }
-        
+
         /// <summary>
         /// The name of the action method.
         /// </summary>
@@ -41,12 +41,7 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
         public string Controller { get; set; }
 
         /// <summary>
-        /// The HTTP method for processing the form, either GET or POST.
-        /// </summary>
-        public string Method { get; set; }
-
-        /// <summary>
-        /// Whether the anti-forgery token should be generated. 
+        /// Whether the anti-forgery token should be generated.
         /// </summary>
         /// <value>Defaults to <c>false</c> if user provides an <c>action</c> attribute; <c>true</c> otherwise.</value>
         [HtmlAttributeName(AntiForgeryAttributeName)]
@@ -84,12 +79,6 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
                 // User is using the FormTagHelper like a normal <form> tag. Anti-forgery default should be false to
                 // not force the anti-forgery token on the user.
                 antiForgeryDefault = false;
-
-                // Restore method attribute.
-                if (Method != null)
-                {
-                    output.CopyHtmlAttribute(nameof(Method), context);
-                }
             }
             else
             {
@@ -98,7 +87,7 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
                                                         Action,
                                                         Controller,
                                                         routeValues,
-                                                        Method,
+                                                        method: null,
                                                         htmlAttributes: null);
 
                 if (tagBuilder != null)
@@ -120,7 +109,8 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
 
         // TODO: https://github.com/aspnet/Razor/issues/89 - We will not need this method once #89 is completed.
         private static Dictionary<string, object> GetRouteValues(
-            TagHelperOutput output, IEnumerable<KeyValuePair<string, string>> routePrefixedAttributes)
+            TagHelperOutput output,
+            IEnumerable<KeyValuePair<string, object>> routePrefixedAttributes)
         {
             Dictionary<string, object> routeValues = null;
             if (routePrefixedAttributes.Any())
@@ -128,10 +118,11 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
                 // Prefixed values should be treated as bound attributes, remove them from the output.
                 output.RemoveRange(routePrefixedAttributes);
 
-                // Generator.GenerateForm does not accept a Dictionary<string, string> for route values.
+                // Remove prefix from keys and convert all values to strings. HtmlString and similar classes are not
+                // meaningful to routing.
                 routeValues = routePrefixedAttributes.ToDictionary(
                     attribute => attribute.Key.Substring(RouteAttributePrefix.Length),
-                    attribute => (object)attribute.Value);
+                    attribute => (object)attribute.Value.ToString());
             }
 
             return routeValues;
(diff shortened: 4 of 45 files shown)


dougbu:
/cc @Eilon @NTaylorMullen @DamianEdwards 

note issue mentioned at bottom of description.  probably need to avoid that in release branch too (?)

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -366,26 +365,70 @@ public virtual void Write(object value)
...
+            var htmlString = value as HtmlString;
+            if (htmlString != null)
+            {
+                if (escapeQuotes)
                 {
-                    var htmlString = value as HtmlString;
-                    if (htmlString != null)
+                    // In this case the text likely came directly from the Razor source. Since the original string is
+                    // an attribute value that may have been quoted with single quotes, must handle any double quotes
+                    // in the value. Writing the value out surrounded by double quotes.
```
dougbu: didn't need to worry about this case before, when we were overzealous about HTML encoding

src/Microsoft.AspNet.Mvc.TagHelpers/FormTagHelper.cs
```diff
@@ -98,7 +87,7 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
                                                         Action,
                                                         Controller,
                                                         routeValues,
-                                                        Method,
+                                                        method: null,
```
dougbu: don't need to pass user's `method` to the `IHtmlGenerator`; final result is the same as before because `TagBuilder` attributes do not overwrite anything already in the `TagHelperOutput`.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -99,6 +99,15 @@ private enum Mode
         }
 
         /// <summary>
+        /// Address of the linked resource.
+        /// </summary>
+        /// <remarks>
+        /// Passed through to the generated HTML in all cases.
+        /// </remarks>
+        [HtmlAttributeName(HrefAttributeName)]
+        public string Href { get; set; }
```
dougbu: must be bound because helper needs the un-encoded value and encoding semantics must be identical to `asp-fallback-href` and so on.  same for `ScriptTagHelper.Src`.

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
```diff
@@ -24,18 +25,19 @@ public static class TagHelperOutputExtensions
...
-                attribute.Key.Equals(attributeName, StringComparison.OrdinalIgnoreCase));
-
-            if (!tagHelperOutput.Attributes.ContainsKey(entry.Key))
+            if (!tagHelperOutput.Attributes.ContainsKey(attributeName))
             {
-                tagHelperOutput.Attributes.Add(entry.Key, entry.Value.ToString());
+                // We look for the original attribute so we can restore the exact attribute name the user typed.
+                // Approach also ignores changes made to tagHelperOutput[attributeName].
+                var entry = context.AllAttributes.First(
+                    attribute => attribute.Key.Equals(attributeName, StringComparison.OrdinalIgnoreCase));
```
dougbu: minor cleanup: didn't need to happen unless the condition was `true`

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -315,16 +341,23 @@ private void EnsureFileVersionProvider()
...
+                    if (FileVersion == true)
+                    {
+                        attributeStringValue = _fileVersionProvider.AddFileVersionToPath(attributeStringValue);
+                    }
+
+                    attributeValue = attributeStringValue;
                 }
 
-                AppendAttribute(builder, attribute.Key, attributeValue, escapteQuotes: false);
+                AppendAttribute(builder, attribute.Key, attributeValue, escapeQuotes: false);
```
dougbu: minor cleanup: fix typo

test/Microsoft.AspNet.Mvc.FunctionalTests/Compiler/Resources/MvcTagHelpersWebSite.MvcTagHelper_Home.Link.html
```diff
@@ -31,14 +31,14 @@
...
+    <link rel="stylesheet" href="/site.css" /><link rel="stylesheet" href="/sub/site2.css" />
 
     <!-- Globbed link tag with existing file and static href should dedupe -->
-    <link href="/site.css" rel="stylesheet" />
+    <link rel="stylesheet" href="/site.css" />
 
     <!-- Fallback to static href -->
-    <link href="/site.min.css" rel="stylesheet" data-extra="test" />
-<meta name="x-stylesheet-fallback-test" class="hidden" /><script>!function(a,b,c){var d,e=document,f=e.getElementsByTagName("SCRIPT"),g=f[f.length-1].previousElementSibling,h=e.defaultView&&e.defaultView.getComputedStyle?e.defaultView.getComputedStyle(g):g.currentStyle;if(h&&h[a]!==b)for(d=0;d<c.length;d++)e.write('<link rel="stylesheet" href="'+c[d]+'"/>')}("visibility","hidden",["\/site.css"]);</script>
+    <link rel="stylesheet" data-extra="test" title="&quot;the&quot; title" href="/site.min.css?a=b&amp;c=d" />
```
dougbu: don't want `&amp;` here.  see TODO item in PR description.  note this encoding is likely not a problem when using static files for CSS and script resources.

test/Microsoft.AspNet.Mvc.FunctionalTests/Compiler/Resources/MvcTagHelpersWebSite.MvcTagHelper_Home.Link.html
```diff
@@ -31,14 +31,14 @@
...
+<meta name="x-stylesheet-fallback-test" class="hidden" /><script>!function(a,b,c){var d,e=document,f=e.getElementsByTagName("SCRIPT"),g=f[f.length-1].previousElementSibling,h=e.defaultView&&e.defaultView.getComputedStyle?e.defaultView.getComputedStyle(g):g.currentStyle;if(h&&h[a]!==b)for(d=0;d<c.length;d++)e.write('<link rel="stylesheet" href="'+c[d]+'"/>')}("visibility","hidden",["\/site.css?a=b\u0026c=d"]);</script>
```
dougbu: also note JavaScript-encoding of the URL is correct here

test/Microsoft.AspNet.Mvc.FunctionalTests/Compiler/Resources/MvcTagHelpersWebSite.MvcTagHelper_Home.Script.html
```diff
@@ -7,34 +7,34 @@
...
-    <script src="/site.js" data-foo="foo-data1">
+    <script src="/site.js" data-foo="foo-data1" title="&lt;the title>">
         // Regular script with comment in body, and extra properties.
     </script>
 
-    <script src="/blank.js" data-foo="foo-data2">
+    <script data-foo="foo-data2" title="&lt;the title>" src="/blank.js?a=b&amp;c=d">
         // TagHelper script with comment in body, and extra properties.
     </script>
-<script>(false||document.write("<script src=\"/site.js\" data-foo=\"foo-data2\"><\/script>"));</script>
```
dougbu: `src=\"/site.js\"` here indicates a slightly different bug in previous version of the `<script/>` tag helper: unlike `<link/>` tag helper, did not JavaScript-encode fallback URLs.

test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/MvcTagHelpersWebSite.MvcTagHelper_Home.OrderUsingHtmlHelpers.Encoded.html
```diff
@@ -0,0 +1,84 @@
+﻿
+
+
+<html>
+<head>
+    <meta charset="utf-8" />
+    <title></title>
+</head>
+<body>
+    <form action="HtmlEncode[[/MvcTagHelper_Order/Submit]]" method="HtmlEncode[[post]]">
```
dougbu: hmm, we over-encode generated URLs as well -- even using HTML helpers.  lack of a customer complaint probably means we can ignore my "todo" item for now.

test/Microsoft.AspNet.Mvc.Razor.Test/RazorPageTest.cs
```diff
@@ -658,9 +658,10 @@ public void IsSectionDefined_ThrowsIfPreviousSectionWritersIsNotRegistered()
 
             // Assert
             var buffer = writer.BufferedWriter.Buffer;
-            Assert.Equal(2, buffer.BufferEntries.Count);
+            Assert.Equal(3, buffer.BufferEntries.Count);
             Assert.Equal("Hello world", buffer.BufferEntries[0]);
-            Assert.Same(stringCollectionWriter.Buffer.BufferEntries, buffer.BufferEntries[1]);
+            Assert.Equal("text1", buffer.BufferEntries[1]);
+            Assert.Equal("text2", buffer.BufferEntries[2]);
```
dougbu: previously, `RazorPage` unnecessarily concatenated the buffer into a new string

test/Microsoft.AspNet.Mvc.TagHelpers.Test/FormTagHelperTest.cs
```diff
@@ -261,52 +258,6 @@ public class FormTagHelperTest
         }
 
         [Theory]
-        [InlineData("my-action")]
-        [InlineData("http://www.contoso.com")]
-        [InlineData("my/action")]
-        public async Task ProcessAsync_RestoresBoundAttributesIfActionIsSpecified(string htmlAction)
```
dougbu: scenario no longer exists

test/Microsoft.AspNet.Mvc.FunctionalTests/Compiler/Resources/MvcTagHelpersWebSite.MvcTagHelper_Home.Link.html
```diff
@@ -7,19 +7,19 @@
...
     <!-- Globbed link tag with existing file and exclude -->
-    <link rel="stylesheet" href="/site.css" /><link rel="stylesheet" href="/sub/site2.css" />
+    <link rel="stylesheet" title="&quot;the&quot; title" href="/site.css" /><link rel="stylesheet" title="&quot;the&quot; title" href="/sub/site2.css" />
 
     <!-- Globbed link tag missing include -->
     <link rel="stylesheet" />
 
     <!-- Globbed link tag missing include but with static href -->
-    <link href="/site.css" rel="stylesheet" />
+    <link rel="stylesheet" href="/site.css" />
```
dougbu: reordering happens because `href` is now bound.  I'll revert changes like this when I fix aspnet/Razor#225, which is mostly about moving bound attributes to the end.

src/Microsoft.AspNet.Mvc.Razor.Host/MvcRazorHost.cs
```diff
@@ -83,6 +89,9 @@ internal MvcRazorHost(ICodeTreeCache codeTreeCache, RazorPathNormalizer pathNorm
 
                     WriteTagHelperAsyncMethodName = "WriteTagHelperAsync",
                     WriteTagHelperToAsyncMethodName = "WriteTagHelperToAsync",
+
+                    // IHtmlHelper<TModel> is also not accessible here.
```
NTaylorMullen: `Can't use nameof because IHtmlHelper<TModel>.Raw is not accessible here.` - Had to read the comment over several times and look at previous configurations to "sort of" understand what it meant.

dougbu: :ok:

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -294,11 +294,10 @@ public TagHelperContent EndTagHelperWritingScope()
 
                 foreach (var attribute in tagHelperOutput.Attributes)
                 {
-                    var value = HtmlEncoder.HtmlEncode(attribute.Value);
                     writer.Write(' ');
                     writer.Write(attribute.Key);
                     writer.Write("=\"");
-                    writer.Write(value);
+                    WriteTo(writer, HtmlEncoder, attribute.Value, escapeQuotes: true);
```
NTaylorMullen: - `escapeQuotes` is currently only relevant if inside of an attribute value. The `WriteTo` method having an extra `escapeQuotes` parameter seems like the wrong place.  Should create a new `HtmlString` (if its an `HtmlString`) with the modified `attribute.Value` with escaped quotes here, then pass through to `WriteTo`. That way we aren't touching a piece of code that everything else calls into.
- Why pass through the `HtmlEncoder` when it's always just using the pages anyways? Suggest not passing through and just using the pages `HtmlEncoder`.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
NTaylorMullen: Add a comment.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
dougbu: - why is `HtmlString` a better location for this?  that class is used all over the place whereas this `WriteTo` method is called three times.
- this `WriteTo` method is `static` and so cannot reference an instance property of the page.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
dougbu: nah.  not much more to say and existing "Passed through" comment matches what we have other places we bind a regular HTML attribute.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -321,25 +338,27 @@ private void BuildLinkTag(IDictionary<string, string> attributes, TagHelperConte
...
                 if (string.Equals(attribute.Key, HrefAttributeName, StringComparison.OrdinalIgnoreCase))
                 {
-                    attributeValue = HtmlEncoder.HtmlEncode(
-                        ShouldAddFileVersion() ?
-                            _fileVersionProvider.AddFileVersionToPath(attributeValue) :
-                            attributeValue);
+                    var attributeStringValue = attributeValue as string;
+
+                    // "href" values come from bound attributes and globbing. Must always be non-null strings.
+                    Debug.Assert(attributeStringValue != null);
```
NTaylorMullen: If multiple `TagHelper`s run on the `Link` tag and one runs previously and sets the `TagHelperOutput.Attributes["href"]` to `null` or something other than `string` wont this fail?

dougbu: nope.  code sets `TagHelperOutput.Attributes["href"]` itself.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
NTaylorMullen: - Hmm? I think you misunderstood what I meant. Don't think that should exist inside of `HtmlString`, instead of doing `WriteTo(writer, HtmlEncoder, attribute.Value, escapeQuotes: true);` you'd do:
```C#
var attributeValue = attribute.Value as HtmlString;
if (attributeValue != null && attributeValue.Value.Contains("\"", StringComparison.Ordinal))
{
    attributeValue = new HtmlString(attributeValue.Value.Replace("\"", "&quote;"));
    WriteTo(writer, attributeValue);
}
else
{
    WriteTo(writer, attribute.Value);
}
```
- :ok:

dougbu: - huh?  that code would need to exist two of the three places that call this method.  better to have one special case than two.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
NTaylorMullen: Hu? What about this call point: https://github.com/aspnet/Mvc/blob/dougbu/bound.not.encoded.razor.335/src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs#L240

Not seeing where we ensure it's been set.

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -243,36 +262,43 @@ private void BuildFallbackBlock(IDictionary<string, string> attributes, DefaultT
...
+                if (!attributes.ContainsKey(SrcAttributeName))
+                {
+                    // Need this entry to place each fallback source.
+                    attributes.Add(SrcAttributeName, null);
+                }
+
                 foreach (var src in fallbackSrcs)
                 {
-                    builder.Append("<script");
+                    // "src" values come from bound attributes and globbing. Must always be non-null.
```
NTaylorMullen: These `src`s are coming from the `GlobbingUrlBuilder`

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
dougbu: https://github.com/aspnet/Mvc/blob/dougbu/bound.not.encoded.razor.335/src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs#L213

and if `Src` is not bound but appears in the `Attributes` dictionary, we're sunk anyhow.  that upends assumptions made in most of the MVC tag helpers.

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
dougbu: right. that takes its inputs from bound attributes and performs globbing on them.  further `FallbackSrc` isn't globbed at all.  the comment is correct.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
NTaylorMullen: That doesn't protect against a previously running `TagHelper` doing `output.Attributes["href"] = ...` when the source looks like:

`<link asp-file-version="true" />`

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -315,16 +341,23 @@ private void EnsureFileVersionProvider()
...
                 if (string.Equals(attribute.Key, SrcAttributeName, StringComparison.OrdinalIgnoreCase))
                 {
-                    attributeValue = HtmlEncoder.HtmlEncode(
-                        ShouldAddFileVersion() ?
-                            _fileVersionProvider.AddFileVersionToPath(attribute.Value) :
-                            attributeValue);
+                    var attributeStringValue = attributeValue as string;
+
+                    // "src" values come from bound attributes and globbing. Must always be non-null strings.
+                    Debug.Assert(attributeStringValue != null);
```
NTaylorMullen: Same concern noted here: https://github.com/aspnet/Mvc/pull/2288/files#r27361694

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
dougbu: again
>  if Src is not bound but appears in the Attributes dictionary, we're sunk anyhow. that upends assumptions made in most of the MVC tag helpers.

NTaylorMullen: That's fine but this is a `Debug.Assert`... We shouldn't be `Debug.Assert`ing for situations that can occur.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
NTaylorMullen: Add a helper, it's really odd seeing it in the `WriteTo`

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperContentExtensions.cs
```diff
@@ -0,0 +1,74 @@
...
+                RazorPage.WriteTo(writer, encoder, value, escapeQuotes: true);
+            }
+
+            return content;
+        }
+
+        // Must be kept consistent with RazorPage.TagHelperContentWrapperTextWriter.
+        private class ContentWrapperTextWriter : TextWriter
+        {
+            public ContentWrapperTextWriter(TagHelperContent content, Encoding encoding)
```
NTaylorMullen: Is encoding used at all in the base implementation in `TextWriter`? We don't use it at all and it pollutes a lot of the call sites for the above `Append` extension method.

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -332,20 +365,27 @@ private void EnsureFileVersionProvider()
...
+        private void AppendAttribute(TagHelperContent content, string key, object value, bool escapeQuotes)
         {
             content
                 .Append(" ")
-                .Append(key)
-                .Append(escapteQuotes ? "=\\\"" : "=\"")
-                .Append(value)
-                .Append(escapteQuotes ? "\\\"" : "\"");
+                .Append(key);
+            if (escapeQuotes)
```
NTaylorMullen: Maybe a better name than `escapeQuotes`? Just because it also determines if we encode the `value`.

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperContentExtensions.cs
```diff
@@ -0,0 +1,74 @@
...
+        /// <param name="value">The <see cref="object"/> to write.</param>
+        /// <returns><paramref name="content"/> after the write operation has completed.</returns>
+        /// <remarks>
+        /// <paramref name="value"/>s of type <see cref="Rendering.HtmlString"/> are written without encoding and the
+        /// <see cref="HelperResult.WriteTo(TextWriter)"/> is invoked for <see cref="HelperResult"/> types.
+        /// For all other types, the encoded result of <see cref="object.ToString"/> is written to the
+        /// <paramref name="writer"/>.
+        /// </remarks>
+        public static TagHelperContent Append(
+            this TagHelperContent content,
```
NTaylorMullen: `[NotNull]` attributes on these guys :smile:

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
dougbu: no. `WriteTo` is the helper.  otherwise end up special-casing `HtmlString` in two layers.

the real fix is to know how attribute values were quoted in the original Razor source.  until we have that, this workaround is sufficient.

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperContentExtensions.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+            using (var writer = new ContentWrapperTextWriter(content, encoding))
+            {
+                RazorPage.WriteTo(writer, encoder, value, escapeQuotes: true);
+            }
+
+            return content;
+        }
+
+        // Must be kept consistent with RazorPage.TagHelperContentWrapperTextWriter.
```
NTaylorMullen: Should just make a singular class used in both locations. Can be in the internal namespace.

test/Microsoft.AspNet.Mvc.FunctionalTests/AntiForgeryTestHelper.cs
```diff
@@ -16,7 +16,8 @@ public static string RetrieveAntiForgeryToken(string htmlContent, string actionU
         {
             return RetrieveAntiForgeryTokens(
                 htmlContent,
-                attribute => attribute.Value.EndsWith(actionUrl, StringComparison.OrdinalIgnoreCase))
+                attribute => (attribute.Value.EndsWith(actionUrl, StringComparison.OrdinalIgnoreCase) ||
```
NTaylorMullen: nit: no need for surrounding parenthesi

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -366,26 +365,70 @@ public virtual void Write(object value)
...
+            {
+                if (escapeQuotes)
                 {
-                    var htmlString = value as HtmlString;
-                    if (htmlString != null)
+                    // In this case the text likely came directly from the Razor source. Since the original string is
+                    // an attribute value that may have been quoted with single quotes, must handle any double quotes
+                    // in the value. Writing the value out surrounded by double quotes.
+                    var stringValue = htmlString.ToString();
+                    if (stringValue.Contains("\""))
```
NTaylorMullen: Can combine this bit into the above `if` statement and then only have a single else `htmlString.WriteTo(writer);` instead of two.

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
dougbu: better name than `escapteQuotes` and about as good as it's going to get

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperContentExtensions.cs
dougbu: :ok:

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperContentExtensions.cs
dougbu: nah.  your request here is contradictory to request to consolidate with `RazorPage.TagHelperContentWrapperTextWriter`.  in any case we only have two calls to this extension method at the moment.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
dougbu: `htmlString.ToString()` is very inefficient when the `HtmlString` has been instantiated using a `StringCollectionTextWriter`.  bad idea to move that into the mainline code path.

perhaps a `return` and fall-through would work.  will play a bit...

test/Microsoft.AspNet.Mvc.FunctionalTests/AntiForgeryTestHelper.cs
dougbu: :ok:

dougbu:
updated

src/Microsoft.AspNet.Mvc.Razor/TagHelperContentWrapperTextWriter.cs
```diff
@@ -0,0 +1,62 @@
...
+{
+    /// <summary>
+    /// <see cref="TextWriter"/> implementation which writes to a <see cref="TagHelperContent"/> instance.
+    /// </summary>
+    public class TagHelperContentWrapperTextWriter : TextWriter
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TagHelperContentWrapperTextWriter"/> class.
+        /// </summary>
+        /// <param name="encoding">The <see cref="Encoding"/> in which output is written.</param>
```
NTaylorMullen: It's not though :/, the encoding has no effect on what gets written to the `TagHelperContent`. Now that this is consolidated, is the encoding necessary? It's a requirement of the abstract `TextWriter` class but does not come into play for what we're using `TagHelperContentWrapperTextWriter` for.

dougbu: this approximates how the `Encoding` property in `TextWriter` is documented.  as you say, it is part of the class' requirements.  the value should match the rest of our `TextWriter`s so future uses (there are very few today) work correctly.

in any case, I'm not revamping `TagHelperContentWrapperTextWriter` in this way as part of this PR.  this is for Beta4.

NTaylorMullen:
:shipit:

dougbu:
5882cdb

PR #2293, opened 2015-03-30
Remove caching from QueryStringValueProviderFactory

### LINKED ISSUE #2258: [Cleanup] QueryStringValueProviderFactory does not need to cache in the HttpContext.Items
The pattern is based off of WebAPI's code where they had to store it in the context to avoid reparsing the query. Since we're dealing with a parsed QueryString, we can avoid having to cache the value provider in the `HttpContext`,

pranavkm:
Fixes #2258


kichalla:
:shipit:

PR #2296, opened 2015-03-30
Prevent RazorErrorExtensions.ToDiagnostics from throwing when it encounters SourceLocation.Undefined \ negative error lengths

pranavkm:


pranavkm:
Old PR: https://github.com/aspnet/Mvc/pull/2257

Eilon:
:shipit:

src/Microsoft.AspNet.Mvc.Razor/Precompilation/RazorErrorExtensions.cs
```diff
@@ -20,15 +22,19 @@ public static Diagnostic ToDiagnostics([NotNull] this RazorError error, [NotNull
...
-                                                   error.Location.CharacterIndex + error.Length);
-            var linePositionSpan = new LinePositionSpan(linePositionStart, linePositionEnd);
-
-            var location = Location.Create(filePath, textSpan, linePositionSpan);
+            var location = error.Location;
+            if (location == SourceLocation.Undefined)
+            {
+                location = SourceLocation.Zero;
+            }
+            var length = Math.Max(0, error.Length);
```
dougbu: suggest using a fixed and short length when `location == SourceLocation.Zero`.  end up highlighting a random amount of text in the current (unrelated) file as-is.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/RazorErrorExtensions.cs
```diff
@@ -20,15 +22,19 @@ public static Diagnostic ToDiagnostics([NotNull] this RazorError error, [NotNull
...
 
-            return Diagnostic.Create(descriptor, location);
+            var textSpan = new TextSpan(location.AbsoluteIndex, length);
+            var linePositionStart = new LinePosition(location.LineIndex, location.CharacterIndex);
+            var linePositionEnd = new LinePosition(location.LineIndex,
+                                                   location.CharacterIndex + length);
```
dougbu: nit: can this fit on one line? if not wrap first argument as well.

dougbu:
:shipit: w/ one suggested change

src/Microsoft.AspNet.Mvc.Razor/Precompilation/RazorErrorExtensions.cs
pranavkm: The editor ends up placing the cursor at the start of the error column when we do set the length to 0. Seems like an ok-ish behavior - so I'm leaving this as is.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/RazorErrorExtensions.cs
pranavkm: :ok:

src/Microsoft.AspNet.Mvc.Razor/Precompilation/RazorErrorExtensions.cs
dougbu: I was talking about the case where `error.Length > 0` and `location == SourceLocation.Zero` (might have been `SourceLocation.Undefined` a few lines earlier).  I would prefer `length == 0` for this case, e.g.
``` c#
var location = error.Location;
if (location == SourceLocation.Undefined)
{
  location = SourceLocation.Zero;
}

var length = Math.Max(0, error.Length);
if (location == SourceLocation.Zero)
{
  length = 0;
}
```

PR #2306, opened 2015-03-31
[Perf] Fully cache model metadata

rynowak:
This change caches the actual model metadata instances. Some profiling
showed we didn't go far enough, we were allocating a lot of ModelMetadata
+ ModelPropertyCollection instances.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultMetadataDetails.cs
+++ b/src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultMetadataDetails.cs
@@ -7,21 +7,19 @@ using System.Collections.Generic;
 namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
 {
     /// <summary>
-    /// A cache of metadata objects for a <see cref="DefaultModelMetadata"/>.
+    /// Holds associated metadata objects for a <see cref="DefaultModelMetadata"/>.
     /// </summary>
     /// <remarks>
-    /// These instances are shared by all <see cref="DefaultModelMetadata"/> instances representing
-    /// the same <see cref="Type"/>, property, or parameter. Any modifications to the data must be
-    /// thread-safe for multiple readers and writers.
+    /// Any modifications to the data must be thread-safe for multiple readers and writers.
     /// </remarks>
-    public class DefaultMetadataDetailsCache
+    public class DefaultMetadataDetails
     {
         /// <summary>
-        /// Creates a new <see cref="DefaultMetadataDetailsCache"/>.
+        /// Creates a new <see cref="DefaultMetadataDetails"/>.
         /// </summary>
         /// <param name="key">The <see cref="ModelMetadataIdentity"/>.</param>
         /// <param name="attributes">The set of model attributes.</param>
-        public DefaultMetadataDetailsCache(ModelMetadataIdentity key, IReadOnlyList<object> attributes)
+        public DefaultMetadataDetails(ModelMetadataIdentity key, IReadOnlyList<object> attributes)
         {
             Key = key;
             Attributes = attributes;
@@ -46,6 +44,11 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
         /// Gets or sets the <see cref="ModelMetadataIdentity"/>
         /// </summary>
         public ModelMetadataIdentity Key { get; }
+
+        /// <summary>
+        /// Gets or sets the <see cref="ModelMetadata"/> entries for the model properties.
+        /// </summary>
+        public ModelMetadata[] Properties { get; set; }
 
         /// <summary>
         /// Gets or sets a property accessor delegate to get the property value from a model object.
--- a/src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultMetadataDetails.cs
+++ b/src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultMetadataDetails.cs
@@ -31,17 +31,17 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
         public IReadOnlyList<object> Attributes { get; }
 
         /// <summary>
-        /// Gets or sets the <see cref="Metadata.BindingMetadata"/>
+        /// Gets or sets the <see cref="Metadata.BindingMetadata"/>.
         /// </summary>
         public BindingMetadata BindingMetadata { get; set; }
 
         /// <summary>
-        /// Gets or sets the <see cref="Metadata.DisplayMetadata"/>
+        /// Gets or sets the <see cref="Metadata.DisplayMetadata"/>.
         /// </summary>
         public DisplayMetadata DisplayMetadata { get; set; }
 
         /// <summary>
-        /// Gets or sets the <see cref="ModelMetadataIdentity"/>
+        /// Gets or sets the <see cref="ModelMetadataIdentity"/>.
         /// </summary>
         public ModelMetadataIdentity Key { get; }
 
--- a/src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadata.cs
+++ b/src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadata.cs
@@ -16,7 +16,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
     {
         private readonly IModelMetadataProvider _provider;
         private readonly ICompositeMetadataDetailsProvider _detailsProvider;
-        private readonly DefaultMetadataDetailsCache _cache;
+        private readonly DefaultMetadataDetails _cache;
 
         private ReadOnlyDictionary<object, object> _additionalValues;
         private bool? _isReadOnly;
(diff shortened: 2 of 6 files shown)


src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadataProvider.cs
```diff
@@ -52,85 +60,67 @@ public virtual ModelMetadata GetMetadataForType([NotNull] Type modelType)
...
         /// </remarks>
-        protected virtual DefaultMetadataDetailsCache[] CreatePropertyCacheEntries([NotNull] ModelMetadataIdentity key)
+        protected virtual DefaultMetadataDetails[] CreatePropertyDetails([NotNull] ModelMetadataIdentity key)
         {
-            var propertyHelpers = PropertyHelper.GetProperties(key.ModelType);
+            var propertyHelpers = PropertyHelper.GetVisibleProperties(key.ModelType);
+            if (propertyHelpers == null)
+            {
+                System.Diagnostics.Debugger.Launch();
+            }
```
rynowak: will undo :laughing:

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadataProvider.cs
```diff
@@ -173,12 +159,17 @@ public TypeCache()
...
-                : base(ModelMetadataIdentityComparer.Instance)
+            public ModelMetadataCacheEntry(ModelMetadata metadata, DefaultMetadataDetails details)
             {
+                Metadata = metadata;
+                Details = details;
             }
+
+            public ModelMetadata Metadata { get; private set; }
+
+            public DefaultMetadataDetails Details { get; private set; }
```
rynowak: I know someone is going to suggest that I change this code in a small way in order to earn more PranavPoints. Let me preempt you.... this is a struct.

pranavkm: :100:

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultMetadataDetails.cs
```diff
@@ -7,21 +7,19 @@
...
-        /// Creates a new <see cref="DefaultMetadataDetailsCache"/>.
+        /// Creates a new <see cref="DefaultMetadataDetails"/>.
         /// </summary>
         /// <param name="key">The <see cref="ModelMetadataIdentity"/>.</param>
         /// <param name="attributes">The set of model attributes.</param>
-        public DefaultMetadataDetailsCache(ModelMetadataIdentity key, IReadOnlyList<object> attributes)
+        public DefaultMetadataDetails(ModelMetadataIdentity key, IReadOnlyList<object> attributes)
         {
             Key = key;
             Attributes = attributes;
```
dougbu: old nit: `BindingMetadata` and two properties after it lack periods in their summaries

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadata.cs
```diff
@@ -16,7 +16,7 @@ public class DefaultModelMetadata : ModelMetadata
     {
         private readonly IModelMetadataProvider _provider;
         private readonly ICompositeMetadataDetailsProvider _detailsProvider;
-        private readonly DefaultMetadataDetailsCache _cache;
+        private readonly DefaultMetadataDetails _cache;
```
dougbu: nit: `_details` ?

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadata.cs
```diff
@@ -29,11 +29,11 @@ public class DefaultModelMetadata : ModelMetadata
         /// </summary>
         /// <param name="provider">The <see cref="IModelMetadataProvider"/>.</param>
         /// <param name="detailsProvider">The <see cref="ICompositeMetadataDetailsProvider"/>.</param>
-        /// <param name="cache">The <see cref="DefaultMetadataDetailsCache"/>.</param>
+        /// <param name="cache">The <see cref="DefaultMetadataDetails"/>.</param>
         public DefaultModelMetadata(
             [NotNull] IModelMetadataProvider provider,
             [NotNull] ICompositeMetadataDetailsProvider detailsProvider,
-            [NotNull] DefaultMetadataDetailsCache cache)
+            [NotNull] DefaultMetadataDetails cache)
```
dougbu: nit: `cache` isn't correct here either

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadataProvider.cs
```diff
@@ -24,6 +24,8 @@ public class DefaultModelMetadataProvider : IModelMetadataProvider
         public DefaultModelMetadataProvider(ICompositeMetadataDetailsProvider detailsProvider)
         {
             DetailsProvider = detailsProvider;
+
+            _cacheEntryFactory = CreateCacheEntry;
```
dougbu: curious: why not use `CreateCacheEntry` directly when calling `_typeCache.GetOrAdd()`?

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DefaultModelMetadataProviderTest.cs
```diff
@@ -86,6 +86,7 @@ public void GetMetadataForProperties_Cached()
             // Assert
             for (var i = 0; i < metadata1.Length; i++)
             {
+                Assert.Same(metadata1[0], metadata2[1]);
```
dougbu: why is this assertion inside the loop?

dougbu: actually, why is this assertion valid?  the two properties have different names, attributes, and more

dougbu:
:watch: just for test issue

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadataProvider.cs
```diff
@@ -36,15 +38,21 @@ public virtual IEnumerable<ModelMetadata> GetMetadataForProperties([NotNull]Type
         {
             var key = ModelMetadataIdentity.ForType(modelType);
 
-            var propertyEntries = _propertiesCache.GetOrAdd(key, CreatePropertyCacheEntries);
-
-            var properties = new ModelMetadata[propertyEntries.Length];
-            for (var i = 0; i < properties.Length; i++)
+            var cacheEntry = _typeCache.GetOrAdd(key, _cacheEntryFactory);
+            if (cacheEntry.Details.Properties == null)
```
pranavkm: Does this need a comment saying this would result in safe races?

pranavkm:
:shipit: when @dougbu signs off

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DefaultModelMetadataProviderTest.cs
rynowak: I didn't run the tests, I was having package issues.

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadataProvider.cs
rynowak: Method to delegate conversion allocates (per the spec). That means that every call, every cache lookup, we 'new' a small object. I wouldn't always do this optimization but it did shop up as part of the profiling that led me to make this change.

There's a cool VS Extension that will bring this kind of thing to your attention https://visualstudiogallery.msdn.microsoft.com/f9b47b93-8675-4ae0-9c52-5da8027c4bb8

It will show you in the editor anywhere where the code you wrote is going to allocate without you writing 'new'. I don't recommend fixing the things it shows you 99% of the time, but it's a good educational tool

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadata.cs
rynowak: :+1:

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadata.cs
rynowak: :+1:

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultMetadataDetails.cs
rynowak: :+1: will fix.

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadataProvider.cs
rynowak: Yeah, seems like a good idea.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DefaultModelMetadataProviderTest.cs
rynowak: Updated this code:

```
// Act
            var properties1 = provider.GetMetadataForProperties(typeof(ModelType)).Cast<DefaultModelMetadata>().ToArray();
            var properties2 = provider.GetMetadataForProperties(typeof(ModelType)).Cast<DefaultModelMetadata>().ToArray();

            // Assert
            Assert.Equal(properties1.Length, properties2.Length);
            for (var i = 0; i < properties1.Length; i++)
            {
                Assert.Same(properties1[i], properties2[i]);
                Assert.Same(properties1[i].Attributes, properties2[i].Attributes);
                Assert.Same(properties1[i].BindingMetadata, properties2[i].BindingMetadata);
                Assert.Same(properties1[i].DisplayMetadata, properties2[i].DisplayMetadata);
                Assert.Same(properties1[i].ValidationMetadata, properties2[i].ValidationMetadata);
            }
```

All the assertions for the individual properties (BindingMetadata, etc) are still important - they verify that we cache these.

dougbu:
:shipit:

PR #2312, opened 2015-04-01
Use Hosting.Interfaces

rynowak:
/cc @davidfowl 

Kiran is working on removing the dependency on logging, but for now we still use it. We were getting it through the reference to Hosting before.


davidfowl:
Logging.Interfaces

rynowak:
> Logging.Interfaces

> Kiran is working on removing the dependency on logging, but for now we still use it. We were getting it through the reference to Hosting before.

L2ReadBro

davidfowl:
Until then :smile:

rynowak:
Any other useful feedback? All the logging stuff we did earlier depends on classes in `.Logging`, and it's already in progress being redone.

davidfowl:
I dont think you depend on logging you only depend on interfaces (in Core)

rynowak:
https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Framework.Logging/ReflectionBasedLogValues.cs
https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Framework.Logging/LoggerFactory.cs

davidfowl:
MVC uses LoggerFactory in Core? Something is terribly wrong there.

davidfowl:
Anyways :shipit:

rynowak:
> MVC uses LoggerFactory in Core? Something is terribly wrong there.

Yep. I just found that. It's being used to log something that we're going to remove as part of logging cleanup so that will go away too.

Issue #2276, opened 2015-03-26
Storing objects (non-string things) in TempDataDictionary does not round trip correctly


pranavkm:
### Scenario

```C#
public class FeedbackController : Controller
{
    private const string _userKey = "user-data";
    
    [HttpPost]
    public IActionResult SubmitApplication(string name, int age)
    {
        // Do db stuff
        var user = new User { Id = 100, Name = name, Age = age };
        TempData[_userKey] = user;
        return RedirectToAction("WelcomeUser");
    }
    
    public string WelcomeUser()
    {
        var user = (User)TempData[_userKey];
        return "Welcome " + user.Name;
    }
}
```

**Expected** (per behavior in Mvc 5)
Ouputs Hello {userName}

**Actual**
Unable to cast object of type 'Newtonsoft.Json.Linq.JObject' to type 'User'.

pranavkm:
It looks like we Json serialize values in `TempDataDictionary` as opposed to using binary serialization in Mvc 5. Consequently complex types don't deserialize correctly. Perhaps we should consider disallowing storing non-primitive value types + strings in our default implementation if the (totally reasonable) intention is to avoid binary serialization.

pranavkm:
Minor note: In-memory session state is simply holding on to the object in memory, not serializing them. Consequently, you can store objects that cannot be deserialized in it

```C#
public class User
{
   public User(int id) {}
   public string Name { get; set; }
}
```

This is to counter @danroth27's suggestion to use Json.Net's type name handling behavior to serialize types.

danroth27:
@pranavkm Consider restricting to only primitives, dictionaries of primitives and arrays of primitives for now.

pranavkm:
I think this was meant to be assigned to @ajaybhargavb. Let's talk about this before you start working on it.

Fido789:
In some early beta it was working as expected

`var messages = TempData[PageMessagesKey] as IList<PageMessageModel>;`

then in beta4 I had to deserialize objects from TempData like this

`var messages = (TempData[PageMessagesKey] as Newtonsoft.Json.Linq.JArray)?.ToObject<List<PageMessageModel>>();`

but it was working. And in beta5 it is not working at all. I don't think it's going to the right direction. How can I use TempData to store objects in beta5?

pranavkm:
cc @Eilon. Could you respond to @Fido789 on what the suggested approach is if a user wants to stash complex data types?

Eilon:
[...]

PR #2317, opened 2015-04-02
Serialize only simple types to session in TempData

ajaybhargavb:
Issue - #2276 
@pranavkm

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/Resources.resx
+++ b/src/Microsoft.AspNet.Mvc.Core/Resources.resx
@@ -451,4 +451,13 @@
   <data name="ModelType_WrongType" xml:space="preserve">
     <value>The model's runtime type '{0}' is not assignable to the type '{1}'.</value>
   </data>
+  <data name="TempData_CannotSerializeToSession" xml:space="preserve">
+    <value>The '{0}' cannot serialize an object of type '{1}' to session state.</value>
+  </data>
+  <data name="TempData_CannotDeserializeToken" xml:space="preserve">
+    <value>Cannot deserialize {0} of type '{1}'.</value>
+  </data>
+  <data name="TempData_CannotSerializeDictionary" xml:space="preserve">
+    <value>The '{0}' cannot serialize a dictionary with a key of type '{1}' to session state.</value>
+  </data>
 </root>
--- a/src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
@@ -2,12 +2,18 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
+using System.Linq;
+using System.Reflection;
 using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Mvc.Core;
 using Microsoft.Framework.Internal;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Bson;
+using Newtonsoft.Json.Linq;
 
 namespace Microsoft.AspNet.Mvc
 {
@@ -16,8 +22,34 @@ namespace Microsoft.AspNet.Mvc
     /// </summary>
     public class SessionStateTempDataProvider : ITempDataProvider
     {
-        private static JsonSerializer jsonSerializer = new JsonSerializer();
-        private static string TempDataSessionStateKey = "__ControllerTempData";
+        private const string TempDataSessionStateKey = "__ControllerTempData";
+        private readonly JsonSerializer _jsonSerializer = JsonSerializer.Create(
+            new JsonSerializerSettings()
+            {
+                TypeNameHandling = TypeNameHandling.None
+            });
+
+        private static readonly MethodInfo _convertArrayMethodInfo = typeof(SessionStateTempDataProvider).GetMethod(
+            nameof(ConvertArray), BindingFlags.Static | BindingFlags.NonPublic);
+        private static readonly MethodInfo _convertDictMethodInfo = typeof(SessionStateTempDataProvider).GetMethod(
+            nameof(ConvertDictionary), BindingFlags.Static | BindingFlags.NonPublic);
+
+        private static readonly ConcurrentDictionary<Type, Func<JArray, object>> _arrayConverters =
+            new ConcurrentDictionary<Type, Func<JArray, object>>();
+        private static readonly ConcurrentDictionary<Type, Func<JObject, object>> _dictionaryConverters =
+            new ConcurrentDictionary<Type, Func<JObject, object>>();
+
+        private static readonly Dictionary<JTokenType, Type> _tokenTypeLookup = new Dictionary<JTokenType, Type>
+        {
+            { JTokenType.String, typeof(string) },
+            { JTokenType.Integer, typeof(int) },
+            { JTokenType.Boolean, typeof(bool) },
+            { JTokenType.Float, typeof(float) },
+            { JTokenType.Guid, typeof(Guid) },
+            { JTokenType.Date, typeof(DateTime) },
+            { JTokenType.TimeSpan, typeof(TimeSpan) },
+            { JTokenType.Uri, typeof(Uri) },
+        };
 
         /// <inheritdoc />
         public virtual IDictionary<string, object> LoadTempData([NotNull] HttpContext context)
@@ -37,8 +69,67 @@ namespace Microsoft.AspNet.Mvc
                 using (var memoryStream = new MemoryStream(value))
                 using (var writer = new BsonReader(memoryStream))
                 {
-                    tempDataDictionary = jsonSerializer.Deserialize<Dictionary<string, object>>(writer);
+                    tempDataDictionary = _jsonSerializer.Deserialize<Dictionary<string, object>>(writer);
                 }
+
+                var convertedDictionary = new Dictionary<string, object>(tempDataDictionary, StringComparer.OrdinalIgnoreCase);
+                foreach (var item in tempDataDictionary)
+                {
+                    var jArrayValue = item.Value as JArray;
+                    if (jArrayValue != null && jArrayValue.Count > 0)
+                    {
+                        var arrayType = jArrayValue[0].Type;
+                        Type returnType;
+                        if (_tokenTypeLookup.TryGetValue(arrayType, out returnType))
+                        {
+                            var arrayConverter = _arrayConverters.GetOrAdd(returnType, type =>
+                            {
+                                return (Func<JArray, object>)_convertArrayMethodInfo.MakeGenericMethod(type).CreateDelegate(typeof(Func<JArray, object>));
+                            });
+                            var result = arrayConverter(jArrayValue);
+
+                            convertedDictionary[item.Key] = result;
+                        }
+                        else
+                        {
+                            var message = Resources.FormatTempData_CannotDeserializeToken(nameof(JToken), arrayType);
+                            throw new InvalidOperationException(message);
+                        }
+                    }
+                    else
+                    {
+                        var jObjectValue = item.Value as JObject;
+                        if (jObjectValue == null)
+                        {
+                            continue;
+                        }
+                        else if (!jObjectValue.HasValues)
+                        {
+                            convertedDictionary[item.Key] = null;
+                            continue;
+                        }
+
+                        var jTokenType = jObjectValue.Properties().First().Value.Type;
+                        Type valueType;
+                        if (_tokenTypeLookup.TryGetValue(jTokenType, out valueType))
+                        {
+                            var dictionaryConverter = _dictionaryConverters.GetOrAdd(valueType, type =>
+                            {
+                                return (Func<JObject, object>)_convertDictMethodInfo.MakeGenericMethod(type).CreateDelegate(typeof(Func<JObject, object>));
+                            });
+                            var result = dictionaryConverter(jObjectValue);
+
+                            convertedDictionary[item.Key] = result;
+                        }
+                        else
+                        {
+                            var message = Resources.FormatTempData_CannotDeserializeToken(nameof(JToken), jTokenType);
+                            throw new InvalidOperationException(message);
+                        }
+                    }
+                }
+
+                tempDataDictionary = convertedDictionary;
 
                 // If we got it from Session, remove it so that no other request gets it
                 session.Remove(TempDataSessionStateKey);
@@ -59,13 +150,19 @@ namespace Microsoft.AspNet.Mvc
             var hasValues = (values != null && values.Count > 0);
             if (hasValues)
             {
+                foreach (var item in values.Values)
+                {
+                    // We want to allow only simple types to be serialized in session.
+                    EnsureObjectCanBeSerialized(item);
+                }
+
                 // Accessing Session property will throw if the session middleware is not enabled.
                 var session = context.Session;
                 
                 using (var memoryStream = new MemoryStream())
                 using (var writer = new BsonWriter(memoryStream))
                 {
-                    jsonSerializer.Serialize(writer, values);
+                    _jsonSerializer.Serialize(writer, values);
                     session[TempDataSessionStateKey] = memoryStream.ToArray();
                 }
             }
@@ -79,6 +176,66 @@ namespace Microsoft.AspNet.Mvc
         private static bool IsSessionEnabled(HttpContext context)
         {
             return context.GetFeature<ISessionFeature>() != null;
+        }
+
+        internal void EnsureObjectCanBeSerialized(object item)
+        {
+            var itemType = item.GetType();
+            Type actualType = null;
+
+            if (itemType.IsArray)
+            {
+                itemType = itemType.GetElementType();
+            }
+            else if (itemType.GetTypeInfo().IsGenericType)
+            {
+                if (itemType.ExtractGenericInterface(typeof(IList<>)) != null)
+                {
+                    var genericTypeArguments = itemType.GetGenericArguments();
+                    Debug.Assert(genericTypeArguments.Length == 1, "IList<T> has one generic argument");
+                    actualType = genericTypeArguments[0];
+                }
+                else if (itemType.ExtractGenericInterface(typeof(IDictionary<,>)) != null)
+                {
+                    var genericTypeArguments = itemType.GetGenericArguments();
+                    Debug.Assert(genericTypeArguments.Length == 2, "IDictionary<TKey, TValue> has two generic arguments");
+                    // Throw if the key type of the dictionary is not string.
+                    if (genericTypeArguments[0] != typeof(string))
+                    {
+                        var message = Resources.FormatTempData_CannotSerializeDictionary(
+                            typeof(SessionStateTempDataProvider).FullName, genericTypeArguments[0]);
+                        throw new InvalidOperationException(message);
+                    }
+                    else
+                    {
+                        actualType = genericTypeArguments[1];
+                    }
+                }
+            }
+
+            actualType = actualType ?? itemType;
+            if (!TypeHelper.IsSimpleType(actualType))
+            {
+                var underlyingType = Nullable.GetUnderlyingType(actualType) ?? actualType;
+                var message = Resources.FormatTempData_CannotSerializeToSession(
+                    typeof(SessionStateTempDataProvider).FullName, underlyingType);
+                throw new InvalidOperationException(message);
+            }
+        }
+
+        private static IList<TVal> ConvertArray<TVal>(JArray array)
+        {
+            return array.Values<TVal>().ToArray();
+        }
+
+        private static IDictionary<string, TVal> ConvertDictionary<TVal>(JObject jObject)
+        {
+            var convertedDictionary = new Dictionary<string, TVal>(StringComparer.Ordinal);
+            foreach (var item in jObject)
+            {
+                convertedDictionary.Add(item.Key, jObject.Value<TVal>(item.Key));
+            }
+            return convertedDictionary;
         }
     }
 }
--- a/test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
+++ b/test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using Microsoft.AspNet.Http;
 using Moq;
@@ -76,6 +77,151 @@ namespace Microsoft.AspNet.Mvc
             });
         }
 
+        public static TheoryData<object, Type> InvalidTypes
+        {
+            get
+            {
+                return new TheoryData<object, Type>
+                {
+                    { new object(), typeof(object) },
+                    { new object[3], typeof(object) },
+                    { new TestItem(), typeof(TestItem) },
+                    { new List<TestItem>(), typeof(TestItem) },
+                    { new Dictionary<string, TestItem>(), typeof(TestItem) },
+                };
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(InvalidTypes))]
+        public void EnsureObjectCanBeSerialized_InvalidType_Throws(object value, Type type)
+        {
+            // Arrange
+            var testProvider = new SessionStateTempDataProvider();
+
+            // Act & Assert
+            var exception = Assert.Throws<InvalidOperationException>(() =>
+            {
+                testProvider.EnsureObjectCanBeSerialized(value);
+            });
+            Assert.Equal($"The '{typeof(SessionStateTempDataProvider).FullName}' cannot serialize an object of type '{type}' to session state.",
+                exception.Message);
+        }
+
+        public static TheoryData<object, Type> InvalidDictionaryTypes
+        {
+            get
+            {
+                return new TheoryData<object, Type>
+                {
+                    { new Dictionary<int, string>(), typeof(int) },
+                    { new Dictionary<Uri, Guid>(), typeof(Uri) },
+                    { new Dictionary<object, string>(), typeof(object) },
+                    { new Dictionary<TestItem, TestItem>(), typeof(TestItem) }
+                };
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(InvalidDictionaryTypes))]
+        public void EnsureObjectCanBeSerialized_InvalidDictionaryType_Throws(object value, Type type)
+        {
+            // Arrange
+            var testProvider = new SessionStateTempDataProvider();
+
+            // Act & Assert
+            var exception = Assert.Throws<InvalidOperationException>(() =>
+            {
+                testProvider.EnsureObjectCanBeSerialized(value);
+            });
+            Assert.Equal($"The '{typeof(SessionStateTempDataProvider).FullName}' cannot serialize a dictionary with a key of type '{type}' to session state.",
+                exception.Message);
+        }
+
+        public static TheoryData<object> ValidTypes
+        {
+            get
+            {
+                return new TheoryData<object>
+                {
+                    { 10 },
+                    { new int[]{ 10, 20 } },
+                    { "FooValue" },
+                    { new Uri("http://Foo") },
+                    { Guid.NewGuid() },
+                    { new List<string> { "foo", "bar" } },
+                    { new DateTimeOffset() },
+                    { 100.1m },
+                    { new Dictionary<string, int>() },
+                    { new Uri[] { new Uri("http://Foo"), new Uri("http://Bar") } }
+                };
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(ValidTypes))]
+        public void EnsureObjectCanBeSerialized_ValidType_DoesNotThrow(object value)
+        {
+            // Arrange
+            var testProvider = new SessionStateTempDataProvider();
+
+            // Act & Assert (Does not throw)
+            testProvider.EnsureObjectCanBeSerialized(value);
+        }
+
+        [Fact]
+        public void SaveAndLoad_SimpleTypesCanBeStoredAndLoaded()
+        {
+            // Arrange
+            var testProvider = new SessionStateTempDataProvider();
+            var inputGuid = Guid.NewGuid();
+            var inputDictionary = new Dictionary<string, string>
+            {
+                { "Hello", "World" },
+            };
+            var input = new Dictionary<string, object>
+            {
+                { "string", "value" },
+                { "int", 10 },
+                { "bool", false },
+                { "DateTime", new DateTime() },
+                { "Guid", inputGuid },
+                { "List`string", new List<string> { "one", "two" } },
+                { "Dictionary", inputDictionary },
+                { "EmptyDictionary", new Dictionary<string, int>() }
+            };
+            var context = GetHttpContext(new TestSessionCollection(), true);
+
+            // Act
+            testProvider.SaveTempData(context, input);
+            var TempData = testProvider.LoadTempData(context);
+
+            // Assert
+            var stringVal = Assert.IsType<string>(TempData["string"]);
+            Assert.Equal("value", stringVal);
+            var intVal = Convert.ToInt32(TempData["int"]);
+            Assert.Equal(10, intVal);
+            var boolVal = Assert.IsType<bool>(TempData["bool"]);
+            Assert.Equal(false, boolVal);
+            var datetimeVal = Assert.IsType<DateTime>(TempData["DateTime"]);
+            Assert.Equal(new DateTime().ToString(), datetimeVal.ToString());
+            var guidVal = Assert.IsType<Guid>(TempData["Guid"]);
+            Assert.Equal(inputGuid.ToString(), guidVal.ToString());
+            var list = (IList<string>)TempData["List`string"];
+            Assert.Equal(2, list.Count);
+            Assert.Equal("one", list[0]);
+            Assert.Equal("two", list[1]);
+            var dictionary = Assert.IsType<Dictionary<string, string>>(TempData["Dictionary"]);
+            Assert.Equal("World", dictionary["Hello"]);
+            var emptyDictionary = (IDictionary<string, int>)TempData["EmptyDictionary"];
+            Assert.Null(emptyDictionary);
+        }
+
+        private class TestItem
+        {
+            public int DummyInt { get; set; }
+        }
+
         private HttpContext GetHttpContext(ISessionCollection session, bool sessionEnabled=true)
         {
             var httpContext = new Mock<HttpContext>();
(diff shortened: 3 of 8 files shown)


pranavkm:
cc @Eilon

test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
```diff
@@ -76,6 +76,44 @@ public void Save_NullSession_NonEmptyDictionary_Throws()
             });
         }
 
+        [Theory]
+        [MemberData(nameof(InvalidTypes))]
+        public void Save_InvalidType_Throws(string key, object value, Type type)
```
kichalla: Can you add test to verify that all the types mentioned in `TypeHelper.IsSimpleType` can be deserialized...I am specifically interested about `struct` and also not sure about `Uri`

https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Common/TypeHelper.cs#L68

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
```diff
@@ -16,8 +17,8 @@ namespace Microsoft.AspNet.Mvc
     /// </summary>
     public class SessionStateTempDataProvider : ITempDataProvider
     {
-        private static JsonSerializer jsonSerializer = new JsonSerializer();
-        private static string TempDataSessionStateKey = "__ControllerTempData";
+        private string TempDataSessionStateKey = "__ControllerTempData";
```
pranavkm: Why did this change?

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
```diff
@@ -16,8 +17,8 @@ namespace Microsoft.AspNet.Mvc
...
+        private JsonSerializer jsonSerializer = new JsonSerializer();
```
pranavkm: `readonly`?

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
```diff
@@ -76,9 +80,39 @@ public virtual void SaveTempData([NotNull] HttpContext context, IDictionary<stri
             }
         }
 
-        private static bool IsSessionEnabled(HttpContext context)
+        private bool IsSessionEnabled(HttpContext context)
```
pranavkm: Undo?

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
ajaybhargavb: Actually @kichalla suggested that since this is a Singleton service, there is no point in having them as static.

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
```diff
@@ -76,9 +80,39 @@ public virtual void SaveTempData([NotNull] HttpContext context, IDictionary<stri
...
+            foreach (var item in values.Values)
+            {
+                var itemType = item.GetType();
+                Type[] actualTypes = null;
+
+                if (itemType.IsArray)
+                {
+                    itemType = itemType.GetElementType();
+                }
+                else if (TypeHelper.IsCollectionType(itemType))
```
pranavkm: I think this might return true for non-generic collections (like `List`). Maybe just do a check here specifically for `IList<>` and `IDictionary<,>`?

test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
```diff
@@ -76,6 +76,44 @@ public void Save_NullSession_NonEmptyDictionary_Throws()
             });
         }
 
+        [Theory]
```
pranavkm: Needs tests that verify `EnsureObjectCanBeSerialized` succeeds for the types we expect it to work with.

test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
```diff
@@ -76,6 +76,44 @@ public void Save_NullSession_NonEmptyDictionary_Throws()
...
+
+            // Act & Assert
+            var exception = Assert.Throws<InvalidOperationException>(() =>
+            {
+                testProvider.SaveTempData(
+                    GetHttpContext(session: null, sessionEnabled: false),
+                    new Dictionary<string, object> { { key, value } }
+                );
+            });
+            Assert.Equal(string.Format("The type {0} cannot be serialized to Session.", type), exception.Message);
```
pranavkm: $"The type {type} cannot be serialized to Session."

src/Microsoft.AspNet.Mvc.Core/Resources.resx
```diff
@@ -451,4 +451,7 @@
   <data name="ModelType_WrongType" xml:space="preserve">
     <value>The model's runtime type '{0}' is not assignable to the type '{1}'.</value>
   </data>
+  <data name="TempData_CannotSerializeToSession" xml:space="preserve">
+    <value>The type {0} cannot be serialized to Session.</value>
```
pranavkm: Maybe call out what type specifically cannot deserialize here:

> The type {0} cannot be serialized to the `{1}` by '{2}'.

{1} -> nameof(TempDataDictionary), {2} -> typeof(SessionStateTempDataProvider).FullName

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
```diff
@@ -59,6 +60,9 @@ public virtual void SaveTempData([NotNull] HttpContext context, IDictionary<stri
             var hasValues = (values != null && values.Count > 0);
             if (hasValues)
             {
+                // We want to allow only primitive types to be serialized in session.
```
pranavkm: primitive types isn't accurate - simple types?

pranavkm:
We need functional tests that verify these values round trip correctly - if I throw in a `IList<string>` for instance. There's a bunch of other types that `IsSimpleType` allows (`decimal`, `DateTime`, `Guid` etc). @Eilon are we ok with that as long as it round trips correctly?

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
pranavkm: I'm partial to the `JsonSerializer` being instance-level, but keys don't benefit from that. In fact this should probably be a `const` since it's private.

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
pranavkm: Naming convention: `private readonly JsonSerializer _jsonSerializer`

test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
ajaybhargavb: Added.

ajaybhargavb:
Updated.

src/Microsoft.AspNet.Mvc.Core/Resources.resx
```diff
@@ -451,4 +451,7 @@
   <data name="ModelType_WrongType" xml:space="preserve">
     <value>The model's runtime type '{0}' is not assignable to the type '{1}'.</value>
   </data>
+  <data name="TempData_CannotSerializeToSession" xml:space="preserve">
+    <value>The type {0} cannot be serialized to Session by '{1}'.</value>
```
pranavkm: See https://github.com/aspnet/Mvc/pull/2317#discussion_r27625942

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
pranavkm: Can this be undone?

test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
```diff
@@ -76,6 +76,70 @@ public void Save_NullSession_NonEmptyDictionary_Throws()
...
+        public void EnsureObjectCanBeSerialized_ValidType_DoesNotThrow(string key, object value)
+        {
+            // Arrange
+            var testProvider = new SessionStateTempDataProvider();
+
+            // Act & Assert (Does not throw)
+            testProvider.EnsureObjectCanBeSerialized(new Dictionary<string, object> { { key, value } });
+        }
+
+        public static TheoryData<string, object, Type> InvalidTypes
```
pranavkm: Formatting nit: member data needs to preceed the actual test.

test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
```diff
@@ -76,6 +76,70 @@ public void Save_NullSession_NonEmptyDictionary_Throws()
...
+            {
+                return new TheoryData<string, object>
+                {
+                    { "int", 10 },
+                    { "IntArray", new int[]{ 10, 20 } },
+                    { "string", "FooValue" },
+                    { "SimpleDict", new Dictionary<string, int>() },
+                    { "Uri", new Uri("http://Foo") },
+                    { "Guid", Guid.NewGuid() },
+                    { "SimpleList", new List<string> { "foo", "bar" } }
```
pranavkm: Can you also add `DateTimeOffset`, `TimeSpan` and `decimal` and Array of non-primitive types (e.g. array of `Uri`s)

test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
```diff
@@ -76,6 +76,70 @@ public void Save_NullSession_NonEmptyDictionary_Throws()
...
+        {
+            get
+            {
+                return new TheoryData<string, object, Type>
+                {
+                    { "Object", new object(), typeof(object) },
+                    { "ObjectArray", new object[3], typeof(object) },
+                    { "TestItem", new TestItem(), typeof(TestItem) },
+                    { "ListTestItem", new List<TestItem>(), typeof(TestItem) },
+                    { "DictTestItem", new Dictionary<string, TestItem>(), typeof(TestItem) }
```
pranavkm: Maybe also add `Dictionary<object, string>` and `Dictionary<TestItem, TestItem>` here

pranavkm:
:shipit: it from me. Get sign off from @Eilon before you check this in

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
```diff
@@ -16,8 +18,12 @@ namespace Microsoft.AspNet.Mvc
     /// </summary>
     public class SessionStateTempDataProvider : ITempDataProvider
     {
-        private static JsonSerializer jsonSerializer = new JsonSerializer();
-        private static string TempDataSessionStateKey = "__ControllerTempData";
+        private const string TempDataSessionStateKey = "__ControllerTempData";
+        private readonly JsonSerializer _jsonSerializer = JsonSerializer.Create(
+            new JsonSerializerSettings()
+            {
+                TypeNameHandling = TypeNameHandling.Auto
```
Eilon: Need to consult with @blowdart to use this because it is generally not allowed to use any `TypeNameHandling` value other than `None`. I don't know if we can guarantee that the data is coming from a trusted source. For example, suppose that the session state provider is a custom one that stores all session data in a cookie - that is no longer trusted data because it can be tampered. Our default session state providers can't be tampered because they are stored in trusted environments (local machine memory, Azure Cache, etc.).

blowdart: Indeed. You can't ship this. Use of Auto is banned because it enables RCE. Unless you can ensure every input is signed, and validated you need to find another way to do this.

ajaybhargavb:
Updated. @Eilon @pranavkm

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
```diff
@@ -16,8 +21,31 @@ namespace Microsoft.AspNet.Mvc
     /// </summary>
     public class SessionStateTempDataProvider : ITempDataProvider
     {
-        private static JsonSerializer jsonSerializer = new JsonSerializer();
-        private static string TempDataSessionStateKey = "__ControllerTempData";
+        private const string TempDataSessionStateKey = "__ControllerTempData";
+        private readonly JsonSerializer _jsonSerializer = JsonSerializer.Create(
+            new JsonSerializerSettings()
+            {
+                TypeNameHandling = TypeNameHandling.None
```
Eilon: @blowdart FYI - be happy :smile:

blowdart: When am I ever HAPPY? I am content.

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
```diff
@@ -16,8 +21,31 @@ namespace Microsoft.AspNet.Mvc
...
+            });
+
+        private static readonly MethodInfo _convertArrayMethodInfo = typeof(SessionStateTempDataProvider).GetMethod(
+            nameof(ConvertArray), BindingFlags.NonPublic | BindingFlags.Static, null, new[] { typeof(JArray) }, null);
+
+        private readonly ConcurrentDictionary<Type, Func<JArray, object>> _arrayConverters =
+            new ConcurrentDictionary<Type, Func<JArray, object>>();
+
+        private static Dictionary<JTokenType, Type> _arrayTypeLookup = new Dictionary<JTokenType, Type>
```
Eilon: `readonly`

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
```diff
@@ -37,7 +65,27 @@ public class SessionStateTempDataProvider : ITempDataProvider
...
                 {
-                    tempDataDictionary = jsonSerializer.Deserialize<Dictionary<string, object>>(writer);
+                    tempDataDictionary = _jsonSerializer.Deserialize<Dictionary<string, object>>(writer);
+                }
+                foreach (var item in tempDataDictionary.ToList())
+                {
+                    var jArrayValue = item.Value as JArray;
+                    if (jArrayValue != null && jArrayValue.Count > 0)
+                    {
+                        Type returnType = null;
```
Eilon: Don't need to initialize this.

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
```diff
@@ -37,7 +65,27 @@ public class SessionStateTempDataProvider : ITempDataProvider
...
+                        _arrayTypeLookup.TryGetValue(jArrayValue[0].Type, out returnType);
+                        if (returnType != null)
```
Eilon: Can just do the `if()` on the return value of TryGetValue, e.g.: `if (_foo.TryGetValue(x, out y)) { ... do stuff with y ... }`.

Eilon:
What about dictionaries? Is it easy to support those?

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
```diff
@@ -37,7 +65,27 @@ public class SessionStateTempDataProvider : ITempDataProvider
                 using (var memoryStream = new MemoryStream(value))
                 using (var writer = new BsonReader(memoryStream))
                 {
-                    tempDataDictionary = jsonSerializer.Deserialize<Dictionary<string, object>>(writer);
+                    tempDataDictionary = _jsonSerializer.Deserialize<Dictionary<string, object>>(writer);
+                }
+                foreach (var item in tempDataDictionary.ToList())
```
pranavkm: Could you create a new dictionary rather than writing to this one (preferably one with `StringComparer.OrdinalIgnoreCase` semantics? Seems cleaner than doing a `ToList` here.

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
pranavkm: This should probably throw if the mapping type could not be found. Silently ignoring might suppress issues in our filtering behavior.

test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
```diff
@@ -76,6 +76,76 @@ public void Save_NullSession_NonEmptyDictionary_Throws()
             });
         }
 
+        public static TheoryData<object, Type> InvalidTypes
```
pranavkm: Need unit tests verifying deserialization works

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
```diff
@@ -16,8 +21,31 @@ namespace Microsoft.AspNet.Mvc
...
+        private readonly JsonSerializer _jsonSerializer = JsonSerializer.Create(
+            new JsonSerializerSettings()
+            {
+                TypeNameHandling = TypeNameHandling.None
+            });
+
+        private static readonly MethodInfo _convertArrayMethodInfo = typeof(SessionStateTempDataProvider).GetMethod(
+            nameof(ConvertArray), BindingFlags.NonPublic | BindingFlags.Static, null, new[] { typeof(JArray) }, null);
+
+        private readonly ConcurrentDictionary<Type, Func<JArray, object>> _arrayConverters =
```
pranavkm: Could you add a note here - https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc/MvcServices.cs#L169 - saying that it needs to be a singleton because we have things that are cached on it?

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
```diff
@@ -16,8 +21,31 @@ namespace Microsoft.AspNet.Mvc
...
+            new ConcurrentDictionary<Type, Func<JArray, object>>();
+
+        private static Dictionary<JTokenType, Type> _arrayTypeLookup = new Dictionary<JTokenType, Type>
+        {
+            { JTokenType.String, typeof(string) },
+            { JTokenType.Integer, typeof(int) },
+            { JTokenType.Boolean, typeof(bool) },
+            { JTokenType.Float, typeof(float) },
+            { JTokenType.Guid, typeof(Guid) },
+            { JTokenType.Object, typeof(object) },
```
pranavkm: This shouldn't be allowed. We wouldn't have serialized an object array.

test/WebSites/TempDataWebSite/Controllers/HomeController.cs
```diff
@@ -41,5 +43,49 @@ public IActionResult PeekTempData()
...
+        {
+            var value1 = TempData["key1"].ToString();
+            var value2 = Convert.ToInt32(TempData["key2"]);
+            var value3 = (IList<string>)TempData["key3"];
+            var value4 = (DateTime)TempData["key4"];
+            var value5 = (Guid)TempData["key5"];
+            return $"{value1} {value2.ToString()} {value3.Count.ToString()} {value4.ToString()} {value5.ToString()}";
+        }
+
+        public string SetTempDataInvalidType()
```
pranavkm: What's this for?

ajaybhargavb: This is being used in this test https://github.com/aspnet/Mvc/pull/2317/files#diff-c244bc4de89efad52c8eb4342c2ea7c9R192 to verify it throws for types that can't be serialized.

ajaybhargavb:
Updated with support for dictionaries. @Eilon @pranavkm

blowdart:
In the dictionary case is the key string, or can it be any of the types you support?

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
```diff
@@ -16,8 +21,34 @@ namespace Microsoft.AspNet.Mvc
...
+            });
+
+        private static readonly MethodInfo _convertArrayMethodInfo = typeof(SessionStateTempDataProvider).GetMethod(
+            nameof(ConvertArray), BindingFlags.Static | BindingFlags.NonPublic);
+        private static readonly MethodInfo _convertDictMethodInfo = typeof(SessionStateTempDataProvider).GetMethod(
+            nameof(ConvertDict), BindingFlags.Static | BindingFlags.NonPublic);
+
+        private static readonly ConcurrentDictionary<Type, Func<JArray, object>> _arrayConverters =
+            new ConcurrentDictionary<Type, Func<JArray, object>>();
+        private static readonly ConcurrentDictionary<Type, Func<JObject, object>> _dictConverters =
```
pranavkm: `dictionaryConverter`

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
```diff
@@ -37,9 +68,62 @@ public class SessionStateTempDataProvider : ITempDataProvider
...
+                }
+
+                var convertedDictionary = new Dictionary<string, object>(tempDataDictionary, StringComparer.OrdinalIgnoreCase);
+                foreach (var item in tempDataDictionary)
+                {
+                    var jArrayValue = item.Value as JArray;
+                    if (jArrayValue != null && jArrayValue.Count > 0)
+                    {
+                        Type returnType;
+                        if (_tokenTypeLookup.TryGetValue(jArrayValue[0].Type, out returnType))
```
pranavkm: Could you refactor this out in to a variable:
`var arrayType = jArrayValue[0].Type;`

ajaybhargavb:
It can only be string because `JObject` implements `IDictionary<string, JToken>`

blowdart:
Good :)

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
```diff
@@ -80,5 +170,60 @@ private static bool IsSessionEnabled(HttpContext context)
...
+                    itemType.ExtractGenericInterface(typeof(IDictionary<,>)) != null)
+                {
+                    actualTypes = itemType.GetGenericArguments();
+                }
+            }
+
+            actualTypes = actualTypes ?? new Type[] { itemType };
+
+            // Throw if the key type of the dictionary is not string.
+            if (actualTypes.Length > 1 && actualTypes[0] != typeof(string))
```
pranavkm: Do this early on as part of detecting the `IDictionary<,>`. You can avoid having `Type[] actualTypes` and just have one.

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
```diff
@@ -80,5 +170,60 @@ private static bool IsSessionEnabled(HttpContext context)
...
+                }
+            }
+        }
+
+        private static IList<TVal> ConvertArray<TVal>(JArray array)
+        {
+            return array.Values<TVal>().ToArray();
+        }
+
+        private static IDictionary<string, T> ConvertDict<T>(JObject jObject)
```
pranavkm: `TVal`

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
```diff
@@ -80,5 +170,60 @@ private static bool IsSessionEnabled(HttpContext context)
...
+        {
+            var convertedDict = new Dictionary<string, T>();
```
pranavkm: `StringComparer.Ordinal`

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
```diff
@@ -80,5 +170,60 @@ private static bool IsSessionEnabled(HttpContext context)
...
+            foreach (var item in jObject)
+            {
+                convertedDict.Add(item.Key, jObject.Value<T>(item.Key));
```
pranavkm: Shouldn't this be `item.Value`?

test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
```diff
@@ -76,6 +77,144 @@ public void Save_NullSession_NonEmptyDictionary_Throws()
...
+        {
+            get
+            {
+                return new TheoryData<object, Type>
+                {
+                    { new object(), typeof(object) },
+                    { new object[3], typeof(object) },
+                    { new TestItem(), typeof(TestItem) },
+                    { new List<TestItem>(), typeof(TestItem) },
+                    { new Dictionary<string, TestItem>(), typeof(TestItem) },
```
pranavkm: Also need to add a `Dictionary<non-string, simple type>` here

test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
```diff
@@ -76,6 +77,144 @@ public void Save_NullSession_NonEmptyDictionary_Throws()
...
+                { "bool", false },
+                { "DateTime", new DateTime() },
+                { "Guid", inputGuid },
+                { "List`string", new List<string> { "one", "two" } },
+                { "Dictionary", inputDict },
+            };
+            var context = GetHttpContext(new TestSessionCollection(), true);
+
+            // Act
+            //System.Diagnostics.Debugger.Launch();
```
pranavkm: Remove

test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
```diff
@@ -76,6 +77,144 @@ public void Save_NullSession_NonEmptyDictionary_Throws()
...
+        {
+            // Arrange
+            var testProvider = new SessionStateTempDataProvider();
+
+            // Act & Assert (Does not throw)
+            testProvider.EnsureObjectCanBeSerialized(value);
+        }
+
+        [Fact]
+        public void SaveAndLoad_WorksAsExpected()
```
pranavkm: Maybe call this `SimpleTypesCanBeStoredAndLoaded` - it's slightly less vague than AsExpected.

test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
```diff
@@ -76,6 +77,144 @@ public void Save_NullSession_NonEmptyDictionary_Throws()
...
+            var context = GetHttpContext(new TestSessionCollection(), true);
+
+            // Act
+            //System.Diagnostics.Debugger.Launch();
+            testProvider.SaveTempData(context, input);
+            var TempData = testProvider.LoadTempData(context);
+
+            // Assert
+            Assert.Equal("value", TempData["string"]);
+            Assert.Equal(10, Convert.ToInt32(TempData["int"]));
```
pranavkm: This should look like:
```C#
var result = Assert.IsType<int>(TempData["int"]);
Assert.Equal(10, result);
```

pranavkm: Similarly for everything else here.

test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
```diff
@@ -87,12 +226,63 @@ private HttpContext GetHttpContext(ISessionCollection session, bool sessionEnabl
...
             {
                 httpContext.Setup(h => h.GetFeature<ISessionFeature>()).Returns(Mock.Of<ISessionFeature>());
-                httpContext.Setup(h => h.Session[It.IsAny<string>()]);
             }
             return httpContext.Object;
         }
+
+        private class TestSessionCollection : ISessionCollection
+        {
+            private Dictionary<string, byte[]> _innerDict = new Dictionary<string, byte[]>();
```
pranavkm: `_innerDictionary`

test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
```diff
@@ -87,12 +226,63 @@ private HttpContext GetHttpContext(ISessionCollection session, bool sessionEnabl
...
+            }
             if (sessionEnabled)
             {
                 httpContext.Setup(h => h.GetFeature<ISessionFeature>()).Returns(Mock.Of<ISessionFeature>());
-                httpContext.Setup(h => h.Session[It.IsAny<string>()]);
             }
             return httpContext.Object;
         }
+
+        private class TestSessionCollection : ISessionCollection
```
pranavkm: @Tratcher, is there an easy way to re-use `SessionCollection` for test purposes?

pranavkm:
:watch:

test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
Tratcher: Not as is. Feel free to request such on https://github.com/aspnet/HttpAbstractions/issues/272

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
ajaybhargavb: No. It takes the key as argument and returns the converted value.

test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
ajaybhargavb: Changed for everything except `int` and `List`.
The serializer assumes all `int` as `long` and returns `long` so directly casting doesn't work.
And for `List` directly casting works but `Assert.IsType` fails.
```C#
(IList<string>)TempData["List`string"] - works
Assert.IsType<IList<string>>(tempData["List`string"]) - fails

**Comment by ajaybhargavb** [comment]
Updated.

File: src/Microsoft.AspNet.Mvc.Core/Resources.resx [hunk 36]
```diff
@@ -451,4 +451,13 @@
...
+  </data>
+  <data name="TempData_CannotDeserializeToken" xml:space="preserve">
+    <value>Cannot deserialize JToken of type {0}.</value>
+  </data>
+  <data name="TempData_CannotSerializeDictionary" xml:space="preserve">
+    <value>The dictionary with TKey {0} cannot be serialized to Session by '{1}'.</value>
```
Eilon: `The '{0}' cannot serialize a dictionary with a key of type '{1}' to session state.` (Swapped `{0}` and `{1}` too.)

src/Microsoft.AspNet.Mvc.Core/Resources.resx
```diff
@@ -451,4 +451,13 @@
   <data name="ModelType_WrongType" xml:space="preserve">
     <value>The model's runtime type '{0}' is not assignable to the type '{1}'.</value>
   </data>
+  <data name="TempData_CannotSerializeToSession" xml:space="preserve">
+    <value>The type {0} cannot be serialized to Session by '{1}'.</value>
```
Eilon: `The '{0}' cannot serialize an object of type '{1}' to session state.` (Swapped `{0}` and `{1}` too.)

Eilon:
Just some small string comments, then :shipit: !

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
```diff
@@ -80,5 +171,66 @@ private static bool IsSessionEnabled(HttpContext context)
...
+            var itemType = item.GetType();
+            Type actualType = null;
+
+            if (itemType.IsArray)
+            {
+                itemType = itemType.GetElementType();
+            }
+            else if (itemType.GetTypeInfo().IsGenericType)
+            {
+                if (itemType.ExtractGenericInterface(typeof(IList<>)) != null ||
```
pranavkm: Just split this out:

```C#
if (itemType.ExtractGenericInterface(typeof(IList<>)))
{ 
      Debug.Asser(itemType.GetGenericArguments().Length == 1, "IList<T> has one argument");
       itemType = itemType.GetGenericArguments()[0];
}
if (itemType.ExtractGenericInterface(typeof(IDictionary<>)))
{
    var genericTypeArguments = itemType.GetGenericArguments();
   Debug.Assert(genericTypeArguments.Length == 0);
   ...
}
```

src/Microsoft.AspNet.Mvc.Core/Resources.resx
```diff
@@ -451,4 +451,13 @@
   <data name="ModelType_WrongType" xml:space="preserve">
     <value>The model's runtime type '{0}' is not assignable to the type '{1}'.</value>
   </data>
+  <data name="TempData_CannotSerializeToSession" xml:space="preserve">
+    <value>The '{0}' cannot serialize an object of type '{1}' to session state.</value>
+  </data>
+  <data name="TempData_CannotDeserializeToken" xml:space="preserve">
+    <value>Cannot deserialize JToken of type {0}.</value>
```
pranavkm: Let's make `JToken` a parameter.

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
```diff
@@ -37,9 +69,63 @@ public class SessionStateTempDataProvider : ITempDataProvider
...
+                            var message = Resources.FormatTempData_CannotDeserializeToken(arrayType);
+                            throw new InvalidOperationException(message);
+                        }
+                    }
+                    else
+                    {
+                        var jObjectValue = item.Value as JObject;
+                        if (jObjectValue == null || !jObjectValue.HasValues)
+                        {
+                            continue;
```
pranavkm: Should we assign a null value when `jObjectValue.HasValues == false`? It will avoid throwing a cast exception in user code:

```C#
TempData["key"] = new Dictionary<string, string>();

...
var myDictionary = (IDictionary<string, string>)TempData["key"]; // will throw now
```

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
```diff
@@ -37,9 +69,63 @@ public class SessionStateTempDataProvider : ITempDataProvider
...
+                        }
+
+                        var jTokenType = jObjectValue.Properties().First().Value.Type;
+                        Type valueType;
+                        if (_tokenTypeLookup.TryGetValue(jTokenType, out valueType))
+                        {
+                            var dictConverter = _dictionaryConverters.GetOrAdd(valueType, type =>
```
pranavkm: `dictionaryConverter`

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
```diff
@@ -80,5 +172,65 @@ private static bool IsSessionEnabled(HttpContext context)
...
+                throw new InvalidOperationException(message);
+            }
+        }
+
+        private static IList<TVal> ConvertArray<TVal>(JArray array)
+        {
+            return array.Values<TVal>().ToArray();
+        }
+
+        private static IDictionary<string, TVal> ConvertDict<TVal>(JObject jObject)
```
pranavkm: `ConvertDictionary` (let's not abbreviate that specific term)

pranavkm:
:shipit: once the remaining comments are addressed.

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
Eilon: I guess it depends on how you'd get into this state, no? What would the user have tried to store in TempData such that they would end up with `!HasValue`?

pranavkm: An empty dictionary would cause this (I think) - it's not particularly far fetched to run into that scenario.

PR #2328, opened 2015-04-03
* Simplify MvcOptions

### LINKED ISSUE #2266: Remove facades for accessing Options<T> and pass options to the invoker
From API review:

Reduce the numbers of services declared by passing options directly to the invoker. There's a bunch of facades (`IModelBinderProvider`, `IOutputFormatterProvider`, etc...) which abstract over options and really just provide access to a collection.

### LINKED ISSUE #2269: Simplify MvcOptions
From API review:

We want to move away from `OptionDescriptor<T>`. The desire here is to simplify the lifecycle for the various components that go into options - in general they should be instances/singletons. 

There may be some cases where the non-singleton makes sense (`IViewEngine` for one), in that case evaluate the best pattern (factory, type activation, service). 

We want to keep extension methods that we've built where the do something useful. In general the properties on options will be an `IList<T>` and will be easier to work with overall.

pranavkm:
* Remove facades for accessing Options<T> and pass options to the invoker

Fixes #2266
Fixes #2269


src/Microsoft.AspNet.Mvc.Core/ModelBinders/BodyModelBinder.cs
```diff
@@ -19,39 +17,25 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
...
-            _bodyValidationExcludeFiltersProvider = bodyValidationExcludeFiltersProvider;
         }
 
         /// <inheritdoc />
         protected async override Task<ModelBindingResult> BindModelCoreAsync([NotNull] ModelBindingContext bindingContext)
         {
-            var formatters = _bindingContext.Value.InputFormatters;
+            var requestServices = bindingContext.OperationBindingContext.HttpContext.RequestServices;
+
+            var formatterSelector = requestServices.GetRequiredService<IInputFormatterSelector>();
```
pranavkm: Started a thread - https://github.com/aspnet/Mvc/issues/2330 - to discuss if we could change the contract for formatters. It might allow us to get away from doing `GetRequiredService` here.

src/Microsoft.AspNet.Mvc.Core/MvcOptions.cs
```diff
@@ -22,16 +22,16 @@ public class MvcOptions
         public MvcOptions()
         {
             Conventions = new List<IApplicationModelConvention>();
-            ModelBinders = new List<ModelBinderDescriptor>();
+            ModelBinders = new List<IModelBinder>();
             ViewEngines = new List<ViewEngineDescriptor>();
```
pranavkm: cc @lodejard / @davidfowl 

I left it as `ViewEngineDescriptor` on `MvcOptions`. `RazorViewEngine` requires 4 services and it feels a bit dirty to be doing `GetRequiredService` to get to those as part of execution. Thoughts?

rynowak:
This is a really gigantic PR, is it possible to break up at this point?

src/Microsoft.AspNet.Mvc.Core/MvcOptions.cs
davidfowl: Why do you have to call GetRequiredService? https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc/MvcOptionsSetup.cs should take in the dependencies it needs to new up the razor view engine right?

rynowak: A bunch of stuff it uses is currently scoped. We should simplify the rest of these and then find the right solution for the view engine by looking at all the components and what needs to be stateful and why (not as part of this PR).

pranavkm: @davidfowl, that's certainly a possibility. There are types like the `IMvcRazorHost` that are created as part of the `RazorViewEngine`'s object graph that are intended to be transient. We'd have to tweak those to ensure they're ok with longer lifespans or change how we consume them.

pranavkm:
@rynowak, let me give it a try. I can start by un-deleting the files to start with - perhaps it might make the diffs less daunting.

pranavkm:
Ok - Sending it out in parts - this contains most of the code changes without the file deletions. Think this should be manageable.

src/Microsoft.AspNet.Mvc.Core/ModelBinders/BodyModelBinder.cs
harshgMSFT: This seems ugly to get all these 4 services as part of execution. Is this the general pattern for anything that needs scoped/instance services?

src/Microsoft.AspNet.Mvc.Core/ControllerActionInvoker.cs
```diff
@@ -23,20 +23,20 @@ public class ControllerActionInvoker : FilterActionInvoker
...
             [NotNull] ControllerActionDescriptor descriptor,
-            [NotNull] IInputFormattersProvider inputFormatterProvider,
+            [NotNull] IReadOnlyList<IInputFormatter> inputFormatters,
             [NotNull] IControllerActionArgumentBinder controllerActionArgumentBinder,
-            [NotNull] IModelBinderProvider modelBinderProvider,
-            [NotNull] IModelValidatorProviderProvider modelValidatorProviderProvider,
-            [NotNull] IValueProviderFactoryProvider valueProviderFactoryProvider,
+            [NotNull] IReadOnlyList<IModelBinder> modelBinders,
+            [NotNull] IReadOnlyList<IModelValidatorProvider> modelValidatorProviders,
+            [NotNull] IReadOnlyList<IValueProviderFactory> valueProviderFactories,
```
rynowak: Bonus RyanPoints for using `IReadOnlyList` :heart_eyes:

src/Microsoft.AspNet.Mvc.Core/FilterActionInvoker.cs
```diff
@@ -206,14 +207,10 @@ private IFilter[] GetFilters()
...
-
-            context.ValidatorProviders = new List<IModelValidatorProvider>(
-                _modelValidatorProviderProvider.ModelValidatorProviders);
-
-            context.ValueProviderFactories = new List<IValueProviderFactory>(
-                _valueProviderFactoryProvider.ValueProviderFactories);
+            context.InputFormatters = new List<IInputFormatter>(_inputFormatters);
+            context.ModelBinders = new List<IModelBinder>(_modelBinders);
+            context.ValidatorProviders = new List<IModelValidatorProvider>(_modelValidatorProviders);
+            context.ValueProviderFactories = new List<IValueProviderFactory>(_valueProviderFactories);
```
rynowak: beautiful

src/Microsoft.AspNet.Mvc.Core/ModelBinders/BodyModelBinder.cs
pranavkm: Hopefully addressing the layering might solve this. We had a chat with @loudej and he thought this was ok. However I think there's an ongoing discussion \ work item about visiting the usage of `GetRequiredServices` in our framework, so this should get covered as part of it.

rynowak:
:shipit:

Issue #2264, opened 2015-03-25
Make output formatters available to ResourceExecutingContext


rynowak:
From API review: 

We want to expose output formatters to resource filters for consistency -- they have access to input formatters.

harshgMSFT:
58a5ad22797f5dc3589a341832715dc188b0b3e9

PR #2332, opened 2015-04-03
[Fix:#2264] Adding OutputFormatters to resource filters 

harshgMSFT:


src/Microsoft.AspNet.Mvc.Core/ActionResults/ObjectResult.cs
```diff
@@ -273,10 +273,14 @@ private IEnumerable<IOutputFormatter> GetDefaultFormatters(ActionContext context
...
+                var actionBindingContext = context.HttpContext
                                     .RequestServices
-                                    .GetRequiredService<IOutputFormattersProvider>()
-                                    .OutputFormatters;
+                                    .GetRequiredService<IScopedInstance<ActionBindingContext>>()
+                                    .Value;
+
+                // In scenarios where there is a resource filter which directly shortcircuits using an ObjectResult.
+                // actionBindingContext is not setup yet and is null.
+                formatters = actionBindingContext?.OutputFormatters ?? new List<IOutputFormatter>();
```
rynowak: Should we fall back to options in this case? Seems better than always giving a 406.

rynowak:
:watch: code seems fine, this will have to get rebased on top of pranav's changes of course.

Let's discuss the scenario where a resource filter short-circuits.

src/Microsoft.AspNet.Mvc.Core/ActionResults/ObjectResult.cs
harshgMSFT: sure that seems fine, however that would mean another call to `GetRequiredService`

harshgMSFT:
@rynowak updated

src/Microsoft.AspNet.Mvc.Core/ControllerActionInvoker.cs
```diff
@@ -31,9 +32,10 @@ public class ControllerActionInvoker : FilterActionInvoker
             [NotNull] IScopedInstance<ActionBindingContext> actionBindingContextAccessor,
             [NotNull] ITempDataDictionary tempData)
             : base(
-                  actionContext, 
+                  actionContext,
                   filterProviders,
-                  inputFormatters, 
+                  inputFormatters,
+                  outputFormatters,
```
rynowak: Just as an FYI - this will break entropy, and you will have to update it.

rynowak:
:shipit:

harshgMSFT:
58a5ad22797f5dc3589a341832715dc188b0b3e9

Issue #2307, opened 2015-03-31
[Perf] Change MutableObjectModelBinder.CanBindType to use ModelMetadata.IsComplex


rynowak:
For small models, this ends up being about .6% of the request. We already have this as part of model metadata, so it's a trivial change.

PR #2333, opened 2015-04-03
Fixes 2307

harshgMSFT:


dougbu:
:shipit: but update commit description to include a detail or two e.g. what's changing.  also reference the correct issue -- #2307.  2370 doesn't (yet) exist.

rynowak:
Please update https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.ModelBinding/ModelMetadata.cs#L255 to cache the result, that's the point of the bug :laughing: 

`ModelMetadata` instance are cached, so doing it in a private field is fine.

harshgMSFT:
@rynowak updated.

Eilon:
:shipit: from me. Also please update the PR title so it's easier to find in the history.

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/MutableObjectModelBinder.cs
```diff
@@ -18,7 +18,7 @@ public class MutableObjectModelBinder : IModelBinder
         public virtual async Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindingContext)
```
pranavkm: Unrelated, but mark this as `[NotNull]`

pranavkm:
:shipit:

rynowak:
:shipit:

harshgMSFT:
0c0fa10ed0687e5e7885c4bda147858bc486ac05

PR #2334, opened 2015-04-03
Resource renames

victorhurdugaci:
The names of the generated resources will change. This change addresses that. The new names are `<projectname>.<folder_path_with_dots>.<resource_name>`

Please review: @pranavkm @davidfowl @dougbu @DamianEdwards

### DIFF
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
@@ -32,6 +32,8 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
     [TargetElement("link", Attributes = FileVersionAttributeName)]
     public class LinkTagHelper : TagHelper
     {
+        private static readonly string Namespace = typeof(LinkTagHelper).Namespace;
+
         private const string HrefIncludeAttributeName = "asp-href-include";
         private const string HrefExcludeAttributeName = "asp-href-exclude";
         private const string FallbackHrefAttributeName = "asp-fallback-href";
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -40,7 +40,7 @@ public class LinkTagHelper : TagHelper
         private const string FallbackTestClassAttributeName = "asp-fallback-test-class";
         private const string FallbackTestPropertyAttributeName = "asp-fallback-test-property";
         private const string FallbackTestValueAttributeName = "asp-fallback-test-value";
-        private const string FallbackJavaScriptResourceName = "compiler/resources/LinkTagHelper_FallbackJavaScript.js";
+        private const string FallbackJavaScriptResourceName = "Microsoft.AspNet.Mvc.TagHelpers.compiler.resources.LinkTagHelper_FallbackJavaScript.js";
```
pranavkm: Use namespace from type - don't hardcode it here.

pranavkm:
:watch:

dougbu:
@victorhurdugaci 
- which is the base change you're reacting to here?
- what is the timing for the set of changes?
- suggest this change should not impact the generated files under test/Microsoft.AspNet.Mvc.Razor.Host.Test/TestFiles.  instead change `MvcRazorHostTest.RunRuntimeTest()` and `...RunDesignTimeTest()` to use the same filenames as before (nput files are still read from disk) but check them against the renamed resources.

dougbu:
:watch:

victorhurdugaci:
@dougbu 

    which is the base change you're reacting to here?

https://github.com/aspnet/dnx/pull/1615

    what is the timing for the set of changes?

sometime this week
    
    suggest this change should not impact the generated files under ...

good point. updated

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -40,7 +42,7 @@ public class LinkTagHelper : TagHelper
         private const string FallbackTestClassAttributeName = "asp-fallback-test-class";
         private const string FallbackTestPropertyAttributeName = "asp-fallback-test-property";
         private const string FallbackTestValueAttributeName = "asp-fallback-test-value";
-        private const string FallbackJavaScriptResourceName = "compiler/resources/LinkTagHelper_FallbackJavaScript.js";
+        private readonly string FallbackJavaScriptResourceName = Namespace + ".compiler.resources.LinkTagHelper_FallbackJavaScript.js";
```
dougbu: nit: pls wrap this line

dougbu:
:shipit:

Issue #2259, opened 2015-03-25
MutableObjectModelBinder should avoid using PropertyInfo.SetValue to set property values


pranavkm:
We should use cached delegates created using `PropertyHelper.MakeFastPropertySetters` to set property values instead. Note: We create these delegates as part of creating `DefaultMetadataDetailsCache` for property entries today.

rynowak:
@Eilon @danroth27 this is something we wanted to do in a follow up to the MMP changes and didn't quite get to yet. There's a few more places to cleanup also beyond the setter in modelbinding.

PR #2342, opened 2015-04-06
Use fast property setter in MutableObjectModelBinder

ajaybhargavb:
Issue - #2259

### DIFF
--- a/src/Microsoft.AspNet.Mvc.ModelBinding/Binders/MutableObjectModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.ModelBinding/Binders/MutableObjectModelBinder.cs
@@ -480,7 +480,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             {
                 try
                 {
-                    property.SetValue(bindingContext.Model, value);
+                    propertyMetadata.PropertySetter(bindingContext.Model, value);
                 }
                 catch (Exception ex)
                 {
--- a/src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultMetadataDetails.cs
+++ b/src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultMetadataDetails.cs
@@ -51,9 +51,9 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
         public ModelMetadata[] Properties { get; set; }
 
         /// <summary>
-        /// Gets or sets a property accessor delegate to get the property value from a model object.
+        /// Gets or sets a property getter delegate to get the property value from a model object.
         /// </summary>
-        public Func<object, object> PropertyAccessor { get; set; }
+        public Func<object, object> PropertyGetter { get; set; }
 
         /// <summary>
         /// Gets or sets a property setter delegate to set the property value on a model object.
--- a/src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadata.cs
+++ b/src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadata.cs
@@ -284,9 +284,13 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
                     {
                         _isReadOnly = BindingMetadata.IsReadOnly;
                     }
+                    else if (MetadataKind == ModelMetadataKind.Type)
+                    {
+                        _isReadOnly = false;
+                    }
                     else
                     {
-                        _isReadOnly = _details.PropertySetter != null;
+                        _isReadOnly = _details.PropertySetter == null;
                     }
                 }
 
@@ -405,6 +409,24 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
                 }
 
                 return _validatorMetadata;
+            }
+        }
+
+        /// <inheritdoc />
+        public override Func<object, object> PropertyGetter
+        {
+            get
+            {
+                return _details.PropertyGetter;
+            }
+        }
+
+        /// <inheritdoc />
+        public override Action<object, object> PropertySetter
+        {
+            get
+            {
+                return _details.PropertySetter;
             }
         }
     }
--- a/src/Microsoft.AspNet.Mvc.ModelBinding/ModelMetadata.cs
+++ b/src/Microsoft.AspNet.Mvc.ModelBinding/ModelMetadata.cs
@@ -298,5 +298,15 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         {
             return DisplayName ?? PropertyName ?? ModelType.Name;
         }
+
+        /// <summary>
+        /// Gets or sets a property getter delegate to get the property value from a model object.
+        /// </summary>
+        public abstract Func<object, object> PropertyGetter { get; }
+
+        /// <summary>
+        /// Gets or sets a property setter delegate to set the property value on a model object.
+        /// </summary>
+        public abstract Action<object, object> PropertySetter { get; }
     }
 } 
--- a/test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DefaultModelMetadataTest.cs
+++ b/test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DefaultModelMetadataTest.cs
@@ -411,8 +411,93 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
             Assert.Equal(firstPropertiesEvaluation, secondPropertiesEvaluation);
         }
 
+        [Fact]
+        public void IsReadOnly_ReturnsFalse_ForType()
+        {
+            // Arrange
+            var detailsProvider = new EmptyCompositeMetadataDetailsProvider();
+            var provider = new DefaultModelMetadataProvider(detailsProvider);
+
+            var key = ModelMetadataIdentity.ForType(typeof(int[]));
+            var cache = new DefaultMetadataDetails(key, new object[0]);
+
+            var metadata = new DefaultModelMetadata(provider, detailsProvider, cache);
+
+            // Act
+            var isReadOnly = metadata.IsReadOnly;
+
+            // Assert
+            Assert.False(isReadOnly);
+        }
+
+        [Fact]
+        public void IsReadOnly_ReturnsTrue_ForPrivateSetProperty()
+        {
+            // Arrange
+            var detailsProvider = new EmptyCompositeMetadataDetailsProvider();
+            var provider = new DefaultModelMetadataProvider(detailsProvider);
+
+            var key = ModelMetadataIdentity.ForType(typeof(TypeWithProperties));
+            var cache = new DefaultMetadataDetails(key, new object[0]);
+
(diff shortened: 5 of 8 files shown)


src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadata.cs
```diff
@@ -286,7 +286,7 @@ public override bool IsReadOnly
                     }
                     else
                     {
-                        _isReadOnly = _details.PropertySetter != null;
+                        _isReadOnly = _details.PropertySetter == null;
```
rynowak: Yep, looks like we had both of these checks backwards.

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadata.cs
```diff
@@ -407,5 +407,23 @@ public override IReadOnlyList<object> ValidatorMetadata
...
+            }
+        }
+
+        /// <inheritdoc />
+        public override Action<object, object> PropertySetter
+        {
+            get
+            {
+                return _details.PropertySetter;
+            }
```
rynowak: Would like to see some unit tests for when these are populated and not. Take a look at `ModelMetadataProviderTest` which has more of the integration type tests.

rynowak:
:watch: for tests

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/MutableObjectModelBinder.cs
```diff
@@ -480,7 +480,7 @@ internal void ProcessDto(ModelBindingContext bindingContext, ComplexModelDto dto
             {
                 try
                 {
-                    property.SetValue(bindingContext.Model, value);
+                    propertyMetadata.PropertySetter(bindingContext.Model, value);
```
pranavkm: There's code on Line 430 that does reflection to fetch all properties. Could you file a follow up work item to remove using reflection and use model metadata?

rynowak: This is already tracked (assigned me to me)

ajaybhargavb:
Updated.

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadata.cs
```diff
@@ -284,9 +284,13 @@ public override bool IsReadOnly
                     {
                         _isReadOnly = BindingMetadata.IsReadOnly;
                     }
+                    else if (MetadataKind == ModelMetadataKind.Property)
+                    {
+                        _isReadOnly = _details.PropertySetter == null;
+                    }
```
rynowak: This block should probably come first (ignore `BindingMetadata.IsReadOnly` if it's Type metadata).

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DefaultModelMetadataTest.cs
```diff
@@ -411,8 +411,72 @@ public void PropertiesEnumerationEvaluatedOnce()
             Assert.Equal(firstPropertiesEvaluation, secondPropertiesEvaluation);
         }
 
+        [Fact]
+        public void IsReadOnly_returnsFalse_ForType()
```
rynowak: `returnsFalse`

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/ModelMetadataProviderTest.cs
```diff
@@ -696,11 +696,53 @@ public void IsRequired_DataMemberIsRequiredTrueWithoutDataContract_False()
...
+            // Act
+            var metadata = metadataProvider.GetMetadataForProperty(
+                typeof(ClassWithPrivateSetProperty),
+                nameof(ClassWithPrivateSetProperty.StringProperty));
+
+            // Assert
+            Assert.Null(metadata.PropertySetter);
+            Assert.NotNull(metadata.PropertyAccessor);
+        }
+
```
rynowak: How about the getter?

ajaybhargavb: Added a test.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/ModelMetadataProviderTest.cs
```diff
@@ -696,11 +696,75 @@ public void IsRequired_DataMemberIsRequiredTrueWithoutDataContract_False()
...
+            var type = typeof(ClassWithPrivateGetProperty);
+            var propertyName = nameof(ClassWithPrivateGetProperty.StringProperty);
+
+            // Act & Assert
+            var exception = Assert.Throws<ArgumentException>(() =>
+            {
+                metadataProvider.GetMetadataForProperty(type, propertyName);
+            });
+            Assert.Equal($"The property {type}.{propertyName} could not be found.{Environment.NewLine}Parameter name: propertyName",
+                exception.Message);
```
rynowak: Just do `GetMetadataForType` and then check that the property is null, will be clearer.

rynowak:
:shipit:

pranavkm:
:shipit:

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DefaultModelMetadataTest.cs
```diff
@@ -411,8 +411,72 @@ public void PropertiesEnumerationEvaluatedOnce()
...
+            Assert.False(isReadOnly);
+        }
+
         private void ActionMethod(string input)
         {
         }
+
+        private class Foo
+        {
+            public string Bar { private set; get; }
```
Eilon: Recommend renaming this to be something like PublicGetPrivateSetProperty and PublicGetPublicSetProperty to make the rest of the code easier to read.

BTW are scenarios like protected properties and/or protected getters/setters interesting here?

src/Microsoft.AspNet.Mvc.ModelBinding/ModelMetadata.cs
```diff
@@ -289,5 +289,15 @@ public string GetDisplayName()
         {
             return DisplayName ?? PropertyName ?? ModelType.Name;
         }
+
+        /// <summary>
+        /// Gets or sets a property accessor delegate to get the property value from a model object.
+        /// </summary>
+        public abstract Func<object, object> PropertyAccessor { get; }
```
Eilon: Rename to `PropertyGetter`? (Matches `PropertySetter` better.)

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DefaultModelMetadataTest.cs
ajaybhargavb: I don't think so. The condition is for them to not be private. So it doesn't matter if they are protected or public.

Eilon: You mean the condition is for them *to be public* right? We presumably want to ignore protected/private/internal?

ajaybhargavb: Right now the PropertySetter is being set for anything that is not private set (similarly for the getter). Should we change it to allow only public? https://github.com/aspnet/Mvc/pull/2342/files#diff-8a66a6db79787dca4cf9119d59910ec1R128

Eilon: Hmm that looks suspicious to me. @rynowak @harshgMSFT any thoughts on what the check should be?

rynowak: The right behavior is whatever MVC5 did. As for the reflection APIs, they don't always give top billing to the other access modifiers than public or private. I seems to recall a few cases where private means non-public.

Eilon: Agreed.

ajaybhargavb: Checked MVC 5's behavior. Anything not public is considered ReadOnly. Will update the condition to reflect the same.

ajaybhargavb:
Updated @Eilon

Eilon:
Looks good! :shipit:

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultMetadataDetails.cs
```diff
@@ -53,7 +53,7 @@ public DefaultMetadataDetails(ModelMetadataIdentity key, IReadOnlyList<object> a
         /// <summary>
         /// Gets or sets a property accessor delegate to get the property value from a model object.
```
harshgMSFT: property getter delegate

src/Microsoft.AspNet.Mvc.ModelBinding/ModelMetadata.cs
```diff
@@ -299,5 +299,15 @@ public string GetDisplayName()
         {
             return DisplayName ?? PropertyName ?? ModelType.Name;
         }
+
+        /// <summary>
+        /// Gets or sets a property accessor delegate to get the property value from a model object.
```
harshgMSFT: property getter delegate

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DefaultModelMetadataTest.cs
```diff
@@ -411,8 +411,93 @@ public void PropertiesEnumerationEvaluatedOnce()
...
+            Assert.False(isReadOnly);
+        }
+
         private void ActionMethod(string input)
         {
         }
+
+        private class Foo
+        {
+            public string PublicGetPrivateSetProperty { private set; get; }
```
harshgMSFT: nit: feels a bit weird looking at order of get and set reversed. Upto you if you want to fix it.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DefaultModelMetadataTest.cs
```diff
@@ -411,8 +411,93 @@ public void PropertiesEnumerationEvaluatedOnce()
...
+
+            // Assert
+            Assert.False(isReadOnly);
+        }
+
         private void ActionMethod(string input)
         {
         }
+
+        private class Foo
```
harshgMSFT: even though this is a private class consider naming it with a better name.

harshgMSFT:
fix comments and optional nits then :shipit:

PR #2345, opened 2015-04-06
Fixes 2304:Decouple Client Validation from IModelValidatorProvider

harshgMSFT:
The fix splits client validation and model validation into two separate hierarchies.
Introduced ClientModelValidatorProvider in MvcOptions, which can be iterated to produce IClientModelValidators.
As a result of this, HtmlGenerator code can be free of ActionBindingContext and directly consumes options.
This also means that we do not modify the client validations during resource filters.


rynowak:
:heart_eyes:

src/Microsoft.AspNet.Mvc.Core/MvcOptions.cs
```diff
@@ -120,6 +122,12 @@ public int MaxModelValidationErrors
         public IList<ModelValidatorProviderDescriptor> ModelValidatorProviders { get; }
 
         /// <summary>
+        /// Gets a list of <see cref="IClientModelValidatorProvider"/>s used by 
+        /// <see cref="ModelBinding.Validation.CompositeClientModelValidatorProvider"/>.
```
rynowak: > Gets a list of <see cref="IClientModelValidatorProvider"/>s used by

Gets a list of <see cref="IClientModelValidatorProvider"/>s instances ...

IMO seems safe to leave out the part about `CompositeClientModelValidatorProvider`.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -25,7 +26,7 @@ public class DefaultHtmlGenerator : IHtmlGenerator
             typeof(DefaultHtmlGenerator).GetTypeInfo().GetDeclaredMethod(nameof(ConvertEnumFromString));
 
         private readonly AntiForgery _antiForgery;
-        private readonly IScopedInstance<ActionBindingContext> _bindingContextAccessor;
+        private readonly IClientModelValidatorProvider _clientModelValidatorProvider;
```
rynowak: @pranavkm is doing another change to get rid of these types of facades. I think we're ok to just use options directly here and not bother with the composite.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/CompositeClientModelValidatorProvider.cs
```diff
@@ -0,0 +1,36 @@
...
+        public IReadOnlyList<IClientModelValidatorProvider> ValidatorProviders { get; }
+
+        /// <inheritdoc />
+        public void GetValidators(ClientValidatorProviderContext context)
+        {
+            foreach (var validatorProvider in ValidatorProviders)
+            {
+                validatorProvider.GetValidators(context);
+            }
+        }
```
rynowak: I don't feel like this class is really needed since we're removing these kinds of things already. Let's discuss if you think it's important to have.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
harshgMSFT: I know, however @pranavkm  is removing things like descriptors and `IModelValidatorProviderProvider`, this interface is still going to be there. And the usage here helps in abstarcting the iteration to `CompositeClientModelValidatorProvider`

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidatorProvider.cs
```diff
@@ -0,0 +1,105 @@
...
+            DataAnnotationsClientModelValidationFactory(ValidationAttribute attribute);
+
+        // Factories for validation attributes
+        private static readonly DataAnnotationsClientModelValidationFactory _defaultAttributeFactory =
+            (attribute) => new DataAnnotationsClientModelValidator(attribute);
+
+        private readonly Dictionary<Type, DataAnnotationsClientModelValidationFactory> _attributeFactories =
+            BuildAttributeFactoriesDictionary();
+
+        internal Dictionary<Type, DataAnnotationsClientModelValidationFactory> AttributeFactories
```
rynowak: Are these internal things used in tests?

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidatorProvider.cs
```diff
@@ -0,0 +1,105 @@
...
+                    factory = _defaultAttributeFactory;
+                }
+
+                context.Validators.Add(factory(attribute));
+            }
+        }
+
+        private static Dictionary<Type, DataAnnotationsClientModelValidationFactory> BuildAttributeFactoriesDictionary()
+        {
+            var dict = new Dictionary<Type, DataAnnotationsClientModelValidationFactory>();
```
rynowak: :laughing:

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/CompositeClientModelValidatorProvider.cs
harshgMSFT: Only reason why its there: 
1. Not repeating iteration in tests and in src. 
Other than that I don't care if this stays.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidatorProvider.cs
harshgMSFT: yes

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidatorProvider.cs
```diff
@@ -0,0 +1,105 @@
...
+        private static void AddValidationAttributeAdapter(
+            Dictionary<Type, DataAnnotationsClientModelValidationFactory> dictionary,
+            Type validationAttributeType,
+            DataAnnotationsClientModelValidationFactory factory)
+        {
+            if (validationAttributeType != null)
+            {
+                dictionary.Add(validationAttributeType, factory);
+            }
+        }
```
rynowak: Would this be simpler to write as a dictionary initializer? Those didn't exist when this code was originally written. Your call if you think it's better or not.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsModelValidator.cs
```diff
@@ -9,7 +9,7 @@
 
 namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
 {
-    public class DataAnnotationsModelValidator : IModelValidator, IClientModelValidator
+    public class DataAnnotationsModelValidator : IModelValidator
```
rynowak: yay

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidator.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// <summary>
+        /// The <see cref="ValidationAttribute"/> associated with this instance.
+        /// </summary>
+        public ValidationAttribute Attribute { get; private set; }
+
+        /// <inheritdoc />
+        public virtual IEnumerable<ModelClientValidationRule> GetClientValidationRules(
+            [NotNull] ClientModelValidationContext context)
+        {
+            var customValidator = Attribute as IClientModelValidator;
```
rynowak: Is this needed because of `DataTypeAttributeAdapter`? should we simplify this further by just making that class be added as a client validator?

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Validation/DataAnnotationsClientModelValidatorProviderTest.cs
```diff
@@ -0,0 +1,155 @@
...
+                                                                    "WithAttribute");
+
+            var providerContext = new ClientValidatorProviderContext(metadata);
+
+            // Act
+            provider.GetValidators(providerContext);
+
+            // Assert
+            var validator = Assert.Single(providerContext.Validators);
+            var adapter = Assert.IsType<RequiredAttributeAdapter>(validator);
```
rynowak: This seems to fly in the face of the name of the test.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/ICompositeClientModelValidatorProvider.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
+{
+    /// <summary>
+    /// Aggregate of <see cref="IClientModelValidatorProvider"/>s that delegates to its underlying providers.
+    /// </summary>
+    public interface ICompositeClientModelValidatorProvider : IClientModelValidatorProvider
```
rynowak: Is this used anywhere?

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidatorProvider.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        /// <inheritdoc />
+        public void GetValidators(ClientValidatorProviderContext context)
+        {
+            foreach (var attribute in context.ValidatorMetadata.OfType<ValidationAttribute>())
+            {
+                DataAnnotationsClientModelValidationFactory factory;
+                if (!_attributeFactories.TryGetValue(attribute.GetType(), out factory))
+                {
+                    factory = _defaultAttributeFactory;
```
pranavkm: We could probably refactor this code without having to specify the default factory.
```C#
{
    valdiator = factory(attribute);
}
else
{
   validator = new DataAnnotationsClientModelValidator(attribute);
}

context.Validators.Add(validator);

File: src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs (hunk 2 above)
Comment (rynowak) [inline src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs]: Part of the work of removing the `OptionDescriptor<T>` is to remove all of the facades that abstract away accessing options, and for our code to just use options directly. What you're doing here isn't exactly that, so it's probably OK if it stays.

File: src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidatorProvider.cs [hunk 12]
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        private static void AddValidationAttributeAdapter(
+            Dictionary<Type, DataAnnotationsClientModelValidationFactory> dictionary,
+            Type validationAttributeType,
+            DataAnnotationsClientModelValidationFactory factory)
+        {
+            if (validationAttributeType != null)
+            {
+                dictionary.Add(validationAttributeType, factory);
```
pranavkm: Should be able to inline this and remove the null check.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidatorProvider.cs
```diff
@@ -0,0 +1,105 @@
...
+namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
+{
+    /// <summary>
+    /// An implementation of <see cref="IClientModelValidatorProvider"/> which providers client validators
+    /// for attributes which derive from <see cref="ValidationAttribute"/>. It also provides
+    /// a validator for types which implement <see cref="IClientModelValidator"/>.
+    /// The logic to support <see cref="IClientModelValidator"/>
+    /// is implemented in <see cref="DataAnnotationsClientModelValidator"/>.
+    /// </summary>
+    public class DataAnnotationsClientModelValidatorProvider : IClientModelValidatorProvider
```
pranavkm: In terms of layering - can all of the client validation move to Rendering? It doesn't have much to do with model binding or the validation system in there now, does it?

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/ICompositeClientModelValidatorProvider.cs
dougbu: why does this marker interface exist?

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataTypeAttributeAdapter.cs
```diff
@@ -12,7 +12,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
     /// A validation adapter that is used to map <see cref="DataTypeAttribute"/>'s to a single client side validation
     /// rule.
     /// </summary>
-    public class DataTypeAttributeAdapter : DataAnnotationsModelValidator
+    public class DataTypeAttributeAdapter : DataAnnotationsClientModelValidator
```
Eilon: Just a thought - should the now client-only validator adapters move into a separate namespace? Like `.ClientValidators` or `.Validators.Client`?

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
harshgMSFT: leaving this as is.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/CompositeClientModelValidatorProvider.cs
harshgMSFT: as in the case of `DefaultHtmlGenerator` leaving this as is. Also noteworthy here is the fact that we already have a composite for the model validator piece as well.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidatorProvider.cs
harshgMSFT: `dict`-> `dictionary`

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidatorProvider.cs
harshgMSFT: done.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidator.cs
harshgMSFT: > Is this needed because of DataTypeAttributeAdapter
No, `DataTypeAttributeAdapter` is already an `IClientModelValidator'.

You would need the cast as the type `ValidationAttribute`.
Moreover it good to have this as there as someone can create a custom attribute which does not implement `IClientModelValidator`.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/ICompositeClientModelValidatorProvider.cs
harshgMSFT: Not used anywhere, its only there for completeness. ( there is a `ICompositeModelValidatorProvider`). 
I am fine removing it if people are not ok.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidatorProvider.cs
harshgMSFT: done,

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/ICompositeClientModelValidatorProvider.cs
dougbu: don't leave unused junk lying around; it'll just confuse future generations

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidatorProvider.cs
harshgMSFT: As discussed yesterday not in this PR.

pranavkm: :+1:

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidator.cs
pranavkm: Couldn't we just filter it early on though - as part of the provider? I'd be ok doing this as part of a separate PR - but feels like it should be the responsibility of the validator provider to unwrap these attributes rather than the validator.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidatorProvider.cs
```diff
@@ -0,0 +1,102 @@
...
+                    factory = _defaultAttributeFactory;
+                }
+
+                context.Validators.Add(factory(attribute));
+            }
+        }
+
+        private static Dictionary<Type, DataAnnotationsClientModelValidationFactory> BuildAttributeFactoriesDictionary()
+        {
+            return new Dictionary<Type, DataAnnotationsClientModelValidationFactory>()
```
pranavkm: :100:

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidator.cs
harshgMSFT: We could but that would mean this class exposes IClientModelValidator instead of a validationattribute. ... which is kind of weird.. .lets discuss and close on this offline.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Validation/DataAnnotationsClientModelValidatorProviderTest.cs
harshgMSFT: whats the issue here? can you clarify?

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
pranavkm: Do we care about allowing `IClientModelValidators` to be changed via `ResourceFilters` (similar to your `IOutputFormatter` change)?

rynowak: > Do we care about allowing IClientModelValidators to be changed via ResourceFilters (similar to your IOutputFormatter change)?

We don't IMO. The problem is that client validators are for the *next* action, not the current action. This was one of the reasons for the change.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidator.cs
harshgMSFT: Thought more about this, keeping this facade is better as the consumers have a clear interface ( they do not have to deal with adapter classes directly).  Moreover this class acts as a base class for all of the adapters and provide the data annotations attribute and the error message. 
The Provider can still ensure that only stuff which is `IClientModelValidator` gets passed to it.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidatorProvider.cs
harshgMSFT: done.

harshgMSFT: There is another piece that I wanted to do (as discussed) this should expose the dictionary in some way will be doing this in a followup PR.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataTypeAttributeAdapter.cs
harshgMSFT: Good idea, we can do that once we do the moving of this code base to under rendering etc. Let me know if you want to do this right away.

harshgMSFT:
updated.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataTypeAttributeAdapter.cs
rynowak: I think it's a good idea to do *iff* we reorganize this code further.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/CompositeClientModelValidatorProvider.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// </summary>
+        /// <param name="providers">
+        /// A collection of <see cref="IClientModelValidatorProvider"/> instances.
+        /// </param>
+        public CompositeClientModelValidatorProvider([NotNull] IEnumerable<IClientModelValidatorProvider> providers)
+        {
+            ValidatorProviders = new List<IClientModelValidatorProvider>(providers);
+        }
+
+        public IReadOnlyList<IClientModelValidatorProvider> ValidatorProviders { get; }
```
rynowak: Docs

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/CompositeModelValidatorProvider.cs
```diff
@@ -7,9 +7,9 @@
 namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
 {
     /// <summary>
-    /// Default implementation for <see cref="ICompositeModelValidatorProvider"/>.
+    /// Aggregate of <see cref="IModelValidatorProvider"/>s that delegates to its underlying providers.
     /// </summary>
-    public class CompositeModelValidatorProvider : ICompositeModelValidatorProvider
+    public class CompositeModelValidatorProvider
```
rynowak: This should still be an `IModelValidatorProvider`.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidator.cs
```diff
@@ -0,0 +1,56 @@
...
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Linq;
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
+{
+    /// <summary>
+    /// A default implementation of <see cref="IClientModelValidator"/>.
```
rynowak: It's not really a default, it's specifically for the DataAnnotations attributes

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidator.cs
rynowak: As discussed, let's handled this with a separate 'default' provider. Then it's not tied to anything data annotations.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidatorOfTAttribute.cs
```diff
@@ -0,0 +1,33 @@
...
+
+using System.ComponentModel.DataAnnotations;
+
+namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
+{
+    /// <summary>
+    /// A generic implementation of <see cref="DataAnnotationsClientModelValidator"/>.
+    /// </summary>
+    /// <typeparam name="TAttribute">An attribute which implements <see cref="ValidationAttribute"/>.</typeparam>
+    public class DataAnnotationsClientModelValidator<TAttribute> : DataAnnotationsClientModelValidator
```
rynowak: As discussed, I think we can make this abstract.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Validation/DataAnnotationsClientModelValidatorProviderTest.cs
rynowak: `GetValidators_DoesNotAddRequiredAttribute_IfAttributeIsSpecifiedExplicitly`

I think I get it. The point is that a **single** required attribute is added.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -35,13 +36,14 @@ public class DefaultHtmlGenerator : IHtmlGenerator
...
             [NotNull] AntiForgery antiForgery,
-            [NotNull] IScopedInstance<ActionBindingContext> bindingContextAccessor,
+            [NotNull] IOptions<MvcOptions> optionsAccessor,
             [NotNull] IModelMetadataProvider metadataProvider,
             [NotNull] IUrlHelper urlHelper,
             [NotNull] IHtmlEncoder htmlEncoder)
         {
             _antiForgery = antiForgery;
-            _bindingContextAccessor = bindingContextAccessor;
+            var clientValidatorProviders = optionsAccessor.Options.ClientModelValidatorProviders.ToArray();
```
dougbu: nit: why is this here (versus in the `CompositeClientModelValidatorProvider` constructor)?

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/ClientValidatorProviderContext.cs
```diff
@@ -0,0 +1,49 @@
...
+
+namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
+{
+    /// <summary>
+    /// A context for <see cref="IClientModelValidatorProvider"/>.
+    /// </summary>
+    public class ClientValidatorProviderContext
+    {
+        /// <summary>
+        /// Creates a new <see cref="ClientModelValidationContext"/>.
```
dougbu: `<see cref="ClientValidatorProviderContext"/>`

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/ClientValidatorProviderContext.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// </summary>
+        /// <param name="modelMetadata">The <see cref="ModelBinding.ModelMetadata"/>.</param>
```
dougbu: nit: the `ModelMetadata` for what?

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/ClientValidatorProviderContext.cs
```diff
@@ -0,0 +1,49 @@
...
+            get
+            {
+                return ModelMetadata.ValidatorMetadata;
+            }
+        }
+
+        /// <summary>
+        /// Gets the list of <see cref="IClientModelValidator"/> instances. <see cref="IClientModelValidatorProvider"/>
+        /// instances should add validators to this list when
+        /// <see cref="IClientModelValidatorProvider.GetValidators(ClientValidatorProviderContext)"/>
```
dougbu: unless the method is currently ambiguous, remove `(ClientValidatorProviderContext)`

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/CompositeClientModelValidatorProvider.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
+{
+    /// <summary>
+    /// Aggregate of <see cref="IClientModelValidatorProvider"/>s that delegates to its underlying providers..
```
dougbu: remove second period

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
harshgMSFT: Composite does not care about where someone is getting it from?

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/CompositeClientModelValidatorProvider.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// </summary>
+    public class CompositeClientModelValidatorProvider
```
dougbu: doesn't this implement `IClientModelValidatorProvider`?

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidatorOfTAttribute.cs
dougbu: what's the scenario where a validator would inherit from `DataAnnotationsClientModelValidator` and not `DataAnnotationsClientModelValidator<TAttribute>`?  we have a few cases of that today but they look like minor bugs.

if this scenario doesn't exist, could remove the base `DataAnnotationsClientModelValidator` class.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidatorOfTAttribute.cs
```diff
@@ -0,0 +1,33 @@
...
+        where TAttribute : ValidationAttribute
+    {
+
+        /// <summary>
+        /// Create a new instance of <see cref="DataAnnotationsClientModelValidator"/>.
+        /// </summary>
+        /// <param name="attribute">An attribute which implements <see cref="ValidationAttribute"/>.</param>
```
dougbu: - classes inherit from other classes, they don't implement them
- here `TAttribute` is much more important than `ValidationAttribute`

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
dougbu: sorry, my comment wasn't clear.  my complaint was about the `ToArray()` call: `DefaultHtmlGenerator` looks at `ClientModelValidatorProviders` once and calls the `CompositeClientModelValidatorProvider` once.  so if there's a (perf) reason for `ToArray()`, the concern is within `CompositeClientModelValidatorProvider`.  likely don't need it at all.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidatorOfTAttribute.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
+{
+    /// <summary>
+    /// A generic implementation of <see cref="DataAnnotationsClientModelValidator"/>.
+    /// </summary>
+    /// <typeparam name="TAttribute">An attribute which inherits <see cref="ValidationAttribute"/>.</typeparam>
```
dougbu: nit: `inherits from`

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidatorOfTAttribute.cs
```diff
@@ -0,0 +1,50 @@
...
+    public abstract class DataAnnotationsClientModelValidator<TAttribute> : IClientModelValidator
+        where TAttribute : ValidationAttribute
+    {
+        /// <summary>
+        /// Create a new instance of <see cref="DataAnnotationsClientModelValidator"/>.
+        /// </summary>
+        /// <param name="attribute">An attribute which inherits <see cref="ValidationAttribute"/>.</param>
```
dougbu: nit: `from` may be missing a couple of places

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidatorProvider.cs
```diff
@@ -0,0 +1,103 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Linq;
+
+namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
+{
+    /// <summary>
+    /// An implementation of <see cref="IClientModelValidatorProvider"/> which providers client validators
```
dougbu: `providers` -> `provides`

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsModelValidatorProvider.cs
```diff
@@ -19,100 +19,18 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
     /// </summary>
```
dougbu: `<summary/>` needs to be updated

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DefaultClientModelValidatorProvider.cs
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+
+namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
+{
+    /// <summary>
+    /// A default implementation of <see cref="IClientModelValidatorProvider"/> which providers client validators
```
dougbu: `providers` -> `provides`

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidatorProvider.cs
```diff
@@ -0,0 +1,103 @@
...
+            {
+                IClientModelValidator validator;
+                DataAnnotationsClientModelValidationFactory factory;
+                if (_attributeFactories.TryGetValue(attribute.GetType(), out factory))
+                {
+                    validator = factory(attribute);
+                }
+                else
+                {
+                    validator = attribute as IClientModelValidator;
```
dougbu: isn't this redundant now that you've added `DefaultClientModelValidatorProvider`?

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsModelValidatorProvider.cs
```diff
@@ -19,100 +19,18 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
...
-                {
-                    factory = _defaultAttributeFactory;
-                }
-
-                context.Validators.Add(factory(attribute));
+                context.Validators.Add(new DataAnnotationsModelValidator(attribute));
             }
 
             // Produce a validator if the type supports IValidatableObject
             if (typeof(IValidatableObject).IsAssignableFrom(context.ModelMetadata.ModelType))
```
dougbu: nit: strange to have `IValidatableObject` handled in this class while implementations of `IClientModelValidator` are handled in their own provider.  suggest `DataAnnotationsModelValidatorProvider` should also be made specific to `ValidationAttribute`

dougbu: hmm, this might not be a nit...

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsModelValidator.cs
dougbu: where is this class now used?  if adapters aren't needed at all server-side, remove

test/Microsoft.AspNet.Mvc.Core.Test/Rendering/DefaultHtmlGeneratorTest.cs
```diff
@@ -584,10 +584,13 @@ private static IHtmlGenerator GetGenerator(IModelMetadataProvider metadataProvid
                 mvcOptionsAccessor.Object,
                 htmlEncoder,
                 dataOptionsAccessor.Object);
-
```
dougbu: nit: this was my favourite blank line in the whole repo :frowning:

dougbu:
:watch:

also
- are CIs failing because you haven't rebased recently?
- will `ModelMetadata` additions we discussed today be included in the next iteration?

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsModelValidator.cs
Eilon: Not sure I understand. Adapters are by definition server components... they translate server validator metadata into client validator metadata.

Then again, the key point of this change was to remove the interface from this type... everything else is just fallout from that. The purpose of this type is to take Data Annotations attributes and turn them into server validators.

Or was this comment about something else?

harshgMSFT: @dougbu  
Huh? I do not understand our comment, this is not about adapters, this is about normal server side validation. the class that has been removed is `DataAnnotationsClientModelValidator`

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsModelValidatorProvider.cs
harshgMSFT: Those are two different things. `IClientModelValidator` has got nothing to do with data annotations and that is why it is handled in a separate provider. However `IValidatableObject` is a concept defined in `System.ComponentModel.DataAnnotations` and it is correct to handle it in `DataAnnotationsModelValidatorProvider`and not a separate one. 
The server side equivalent of `IClientModelValidator` (i.e `IModelValidator`) is already handled in a separate provider `DefaultModelValidatorProvider`.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidatorProvider.cs
harshgMSFT: yeah, this else should have been removed.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsModelValidatorProvider.cs
Eilon: @harshgMSFT yeah that sounds right to me.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsModelValidator.cs
dougbu: looked like most references to `DataAnnotationsModelValidator` or `DataAnnotationsModelValidator<T>` now use `DataAnnotationsClientModelValidator` and `DataAnnotationsClientModelValidator<T>`.  so I was asking what uses remained.

now see `DataAnnotationsModelValidatorProvider` still exists and is still enabled.  so my comment isn't actionable.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsModelValidatorProvider.cs
dougbu: :ok:

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsModelValidator.cs
harshgMSFT: still not sure .. are you saying ignore the comment :smile:

dougbu: yes, nevermind

harshgMSFT:
>are CIs failing because you haven't rebased recently?
Yes, but they seem to be failing even post a rebase, I will check and update before I merge.

> will ModelMetadata additions we discussed today be included in the next iteration?

What changes the required changes are already there in `DefaultValidationMetadataProvider`.

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultValidationMetadataProvider.cs
```diff
@@ -19,6 +19,11 @@ public void GetValidationMetadata([NotNull] ValidationMetadataProviderContext co
             {
                 context.ValidationMetadata.ValiatorMetadata.Add(attribute);
```
dougbu: not this issue and probably not this PR but please correct typo in `ValidatorMetadata` (missing 'd')

src/Microsoft.AspNet.Mvc/MvcOptionsSetup.cs
```diff
@@ -1,89 +0,0 @@
-﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
```
dougbu: why was this file removed?

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Validation/DataAnnotationsModelValidatorProviderTest.cs
```diff
@@ -37,111 +37,6 @@ public void GetValidators_ReturnsValidatorForIValidatableObject()
 #endif
 
         [Fact]
-        public void GetValidators_DoesNotAddRequiredAttribute_ForNonNullableValueTypes_IfAttributeIsSpecifiedExplicitly()
```
dougbu: where are the replacements for the tests removed from this class?  don't see a `DataAnnotationsClientModelValidatorProviderTest` class.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Validation/DataAnnotationsModelValidatorTest.cs
```diff
@@ -220,55 +220,6 @@ public void ValidateReturnsMemberNameIfItIsDifferentFromDisplayName()
 #endif
 
         [Fact]
-        public void GetClientValidationRules_ReturnsEmptyRuleSet()
```
dougbu: where are the replacements for these tests?  don't see a `DataAnnotationsClientModelValidatorTest` class.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/CompositeClientModelValidatorProvider.cs
dougbu: looks like @rynowak's comment hasn't been addressed

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsModelValidatorProvider.cs
dougbu: thx

dougbu:
:watch: looks like a couple of files are missing

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Validation/DataAnnotationsModelValidatorTest.cs
harshgMSFT: yeah somehow https://github.com/aspnet/Mvc/commit/700ef25a8131a984d83f67439a8b6bb43c80304a#diff-8d20817bae0a4a0e2f12498e17d231ca
was deleted.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Validation/DataAnnotationsModelValidatorProviderTest.cs
harshgMSFT: yeah somehow
https://github.com/aspnet/Mvc/commit/700ef25a8131a984d83f67439a8b6bb43c80304a#diff-8d20817bae0a4a0e2f12498e17d231ca
was deleted.

src/Microsoft.AspNet.Mvc.Core/MvcOptions.cs
```diff
@@ -115,6 +116,11 @@ public int MaxModelValidationErrors
         public IList<IModelValidatorProvider> ModelValidatorProviders { get; }
 
         /// <summary>
+        /// Gets a list of <see cref="IClientModelValidatorProvider"/> instances.
+        /// </summary>
+        public IList<IClientModelValidatorProvider> ClientModelValidatorProviders { get; set; }
```
pranavkm: This should be `get` only.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Validation/DataAnnotationsModelValidatorTest.cs
harshgMSFT: There are no replacement tests for these, the `DataAnnotationsClientModelValidator` is no longer there. The remaining class `DataAnnotationsClientModelValidator<T>` is abstract and the functionality is well tested by individual adapters.

src/Microsoft.AspNet.Mvc.Core/MvcOptions.cs
harshgMSFT: :+1:

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Validation/DataAnnotationsModelValidatorTest.cs
dougbu: where is `GetErrorMessage()` checked?  what confirms that the client validator provider successfully returns an empty rule set for an unknown `ValidationAttribute`?  what confirms that provider handles the `IClientModelValidator` case in general (not just for `[Remote]`)?

harshgMSFT: >where is GetErrorMessage() checked?
The individual adapters eg look at `StringLengthAttributeAdapterTest`. 
I do not see much value in testing this for the abstract base class. Dont feel strongly against it though. If you do feel strongly for it. I will add a new one.  (Please note the test I have removed had nothing to do with GetErrorMessage.

> what confirms that the [data annotations] client validator provider successfully returns an empty rule set for an unknown ValidationAttribute

It doesnt return an empty rule set. There is no client validator added in such a scenario.

>what confirms that the other provider handles the IClientModelValidator case in general (not just for [Remote])?

look at `DefaultModelValidatorProviderTest`.

harshgMSFT:
Hopefully this is the last one.--- Branch Updated.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/CompositeClientModelValidatorProvider.cs
dougbu: add docs

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -35,13 +36,14 @@ public class DefaultHtmlGenerator : IHtmlGenerator
         /// </summary>
         public DefaultHtmlGenerator(
             [NotNull] AntiForgery antiForgery,
-            [NotNull] IScopedInstance<ActionBindingContext> bindingContextAccessor,
+            [NotNull] IOptions<MvcOptions> optionsAccessor,
```
dougbu: pls add `<param/>` descriptions

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidatorOfTAttribute.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
+{
+    /// <summary>
+    /// An implementation of <see cref="IClientModelValidator"/> which understands data annotation attributes.
+    /// </summary>
+    /// <typeparam name="TAttribute">An attribute which inherits from <see cref="ValidationAttribute"/>.</typeparam>
```
dougbu: - `The type of the attribute.`
- don't think mentioning the type restriction is worthwhile.  but if you feel strongly about it, `... Must be a <see cref="ValidationAttribute"/> subclass.`

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidatorOfTAttribute.cs
```diff
@@ -0,0 +1,50 @@
...
+    public abstract class DataAnnotationsClientModelValidator<TAttribute> : IClientModelValidator
+        where TAttribute : ValidationAttribute
+    {
+        /// <summary>
+        /// Create a new instance of <see cref="DataAnnotationsClientModelValidator{TAttribute}"/>.
+        /// </summary>
+        /// <param name="attribute">An attribute which inherits from <see cref="ValidationAttribute"/>.</param>
```
dougbu: `The <typeparamref name="TAttribute"/> instance to validate.`

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidatorOfTAttribute.cs
```diff
@@ -0,0 +1,50 @@
...
+        public DataAnnotationsClientModelValidator(TAttribute attribute)
+        {
+            Attribute = attribute;
+        }
+
+        /// <summary>
+        /// Gets the attribute associated with this instance.
```
dougbu: minor: attribute -> `<typeparamref name="TAttribute"/>`

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidatorProvider.cs
```diff
@@ -0,0 +1,96 @@
...
+using System.Linq;
+
+namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
+{
+    /// <summary>
+    /// An implementation of <see cref="IClientModelValidatorProvider"/> which provides client validators
+    /// for attributes which derive from <see cref="ValidationAttribute"/>. It also provides
+    /// a validator for types which implement <see cref="IClientModelValidator"/>.
+    /// The logic to support <see cref="IClientModelValidator"/>
+    /// is implemented in <see cref="DataAnnotationsClientModelValidator{}"/>.
```
dougbu: `DataAnnotationsClientModelValidator{}` -> `DataAnnotationsClientModelValidator{TAttribute}`

harshgMSFT: There is no {TAttribute} defined in this class and the xml-docs-test barfs while building.

dougbu: bit strange since `DataAnnotationsClientModelValidator` class doesn't exist and empty curly braces should also cause an error.  try removing `{}` then.

harshgMSFT: That doesnt work. This is the only thing that works.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Validation/DefaultModelClientValidatorProviderTest.cs
```diff
@@ -0,0 +1,268 @@
...
+            var metadata = metadataProvider.GetMetadataForType(typeof(ModelValidatorAttributeOnClass));
+            var context = new ModelValidatorProviderContext(metadata);
+
+            // Act
+            validatorProvider.GetValidators(context);
+
+            // Assert
+            var validators = context.Validators;
+
+            var validator = Assert.IsType<CustomModelValidatorAttribute>(Assert.Single(validators));
```
dougbu: nit: two line version of these assertions is easier on the eyes

dougbu:
:watch: for a few doc changes and additions

harshgMSFT:
@dougbu  updated.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/CompositeClientModelValidatorProvider.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// <param name="providers">
+        /// A collection of <see cref="IClientModelValidatorProvider"/> instances.
+        /// </param>
+        public CompositeClientModelValidatorProvider([NotNull] IEnumerable<IClientModelValidatorProvider> providers)
+        {
+            ValidatorProviders = new List<IClientModelValidatorProvider>(providers);
+        }
+
+        /// <summary>
+        /// Gets a list of <see cref="IClientModelValidatorProvider"/>.
```
dougbu: add `instances`

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/CompositeModelValidatorProvider.cs
```diff
@@ -22,6 +22,9 @@ public CompositeModelValidatorProvider([NotNull] IEnumerable<IModelValidatorProv
             ValidatorProviders = new List<IModelValidatorProvider>(providers);
         }
 
+        /// <summary>
+        /// Gets the list of <see cref="IModelValidatorProvider"/>.
```
dougbu: add `instances`

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidatorOfTAttribute.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// Create a new instance of <see cref="DataAnnotationsClientModelValidator{TAttribute}"/>.
+        /// </summary>
+        /// <param name="attribute">The <typeparamref name="TAttribute"/> instance to validate.</param>
+        public DataAnnotationsClientModelValidator(TAttribute attribute)
+        {
+            Attribute = attribute;
+        }
+
+        /// <summary>
+        /// Gets the type of the <typeparamref name="TAttribute"/> associated with this instance.
```
dougbu: no, property is not of type `Type`.  `Gets the <typeparamref name="TAttribute"/> instance to validate.` or maybe just `Gets the <typeparamref name="TAttribute"/> instance.`

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidatorProvider.cs
dougbu: weird but :ok:

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/IClientModelValidatorProvider.cs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
+{
+    /// <summary>
+    /// Provides a collection of <see cref="IClientModelValidator"/>.
```
dougbu: add `instances` (or perhaps `s` since you do that later in this interface)

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Validation/DataAnnotationsClientModelValidatorProviderTest.cs
```diff
@@ -0,0 +1,154 @@
...
+    {
+        private readonly IModelMetadataProvider _metadataProvider = TestModelMetadataProvider.CreateDefaultProvider();
+
+        [Fact]
+        public void GetValidators_DoesNotAddRequiredAttribute_IfAttributeIsSpecifiedExplicitly()
+        {
+            // Arrange
+            var provider = new DataAnnotationsClientModelValidatorProvider();
+            var metadata = _metadataProvider.GetMetadataForProperty(typeof(DummyRequiredAttributeHelperClass),
+                                                                    "WithAttribute");
```
dougbu: side note: this is test code but don't get into the habit of wrapping lines this way. (a) just hit `<Enter>` (don't force the next person to enter all the spaces) and (b) if you wrap one argument, wrap them all.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Validation/DefaultModelClientValidatorProviderTest.cs
dougbu: put another way, one assertion per line

dougbu:
:shipit: with final (doc comment) fixes

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/IClientModelValidatorProvider.cs
harshgMSFT: DONE

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DataAnnotationsClientModelValidatorOfTAttribute.cs
harshgMSFT: oops my bad.

PR #2347, opened 2015-04-07
Add serviceable attribute to projects.

NTaylorMullen:


src/Microsoft.AspNet.JsonPatch/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Reflection;
+
+[assembly: AssemblyMetadata("Serviceable", "True")]
```
pranavkm: Out of curiosity, is the casing for `True` supposed to look like this?

pranavkm:
:shipit:

src/Microsoft.AspNet.Mvc.Razor.Host/Properties/AssemblyInfo.cs
```diff
@@ -1,6 +1,8 @@
-﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
```
dougbu: is this another BOM change?

src/Microsoft.AspNet.JsonPatch/Properties/AssemblyInfo.cs
Eilon: Yes

Eilon:
:shipit:

src/Microsoft.AspNet.Mvc.Razor.Host/Properties/AssemblyInfo.cs
NTaylorMullen: Actually we're removing the BOM. Found that our assemblyinfo.cs files are extreeemly inconsistent (almost seems like a 50/50 split) with their bom's, decided to remove all of them.

PR #2349, opened 2015-04-07
Remove duplicate tests

dougbu:
- duplication in data sets confuses xUnit


dougbu:
FYI only; I'll merge this quick fix shortly

Issue #2261, opened 2015-03-25
Add default editor templates for all numeric types, including short, ushort, and double, etc.


weitzhandler:
Hi.

Please have a look at [this](https://aspnetwebstack.codeplex.com/workitem/2) post.

It addresses what the `System.ComponentModel.DataAnnotations.DataType` attribute is here for.

Anyway, as per the [closing note](https://aspnetwebstack.codeplex.com/workitem/2#ClosedDiv), only the following numeric types are supported:
>type=number
int
byte
sbyte
uint
long
ulong
DataType enum
PhoneNumber (tel)
EmailAddress (email)
Url (url)
Date (date)
Time (time)
DateTime (datetime)

I'd like to ask you to please include short in the numeric category as well.

danroth27:
@dougbu Can you check if short is actually supported?

dougbu:
@weitzhandler server-side we simply execute any `ValidationAttribute` or subclass we find.  client side we do two things:
1. special-case some attributes for Unobtrusive client-side validation e.g. `[EmailAddress]` and `[Compare]`
2. if you use `@Html.Editor()`, `@Html.EditorFor()`, or the `<input/>` tag helper, generate `<input/>` elements with the appropriate `type` attribute based on `[DataType]`.  this allows browsers to validate the inputs.

I suspect the scenario you've opened this issue about is 2. in my list -- where we generate `<input type="number" .../>` and where we don't.  your `type=number` list above matches what we have in both MVC 5 and 6.

please confirm your request is for helper calls such as `@Html.EditorFor(m => m.ShortProperty)` to generate `<input type="number" .../>`.

weitzhandler:
@dougbu 
Yes that's exactly what I meant.

dougbu:
updated subject to reflect what would need to change here.

@weitzhandler In the meantime you could create a Shared/EditorTemplates/Short.cshtml file containing for example
``` c#
@model short
@Html.TextBoxFor(m => m, htmlAttributes: new { type = "number" })
```

weitzhandler:
I'd also recommend adding those templates for `ushort` and all the other CLR primitive numeric types.

weitzhandler:
@Eilon and @dougbu  [Here](https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs)'s your friend. And BTW, I don't see any reason not to have `decimal` as an `<input type="number" />` as well.

[...]

PR #2350, opened 2015-04-07
Automatically use `type="number"` for expressions of more types

dougbu:
- #2261
- include `short`, `ushort`, `float`, `double`

- correct typo in `<input/>` tag helper; ignored calculated format
 - only one test for `<input/>`'s calculated format :frowning:
- fill some `Editor*()` and `<input/>` tag helper test gaps

nit: clean up some trailing whitespace


test/Microsoft.AspNet.Mvc.Core.Test/Rendering/DefaultEditorTemplatesTest.cs
```diff
@@ -28,48 +28,56 @@ public class DefaultEditorTemplatesTest
             {
                 return new TheoryData<string, string>
                 {
-                    { null, "__TextBox__" },
-                    { string.Empty, "__TextBox__" },
-                    { "EmailAddress", "__TextBox__" },
-                    { "emailaddress", "__TextBox__" },
+                    { null, "__TextBox__ class='text-box single-line'" },
```
pranavkm: Are these due to the `format` fix?

dougbu: no.  `StubbyHtmlHelper` previously ignored the passed `htmlAttributes`.  added that so the chosen `class` and `type` values come through.

the `format` fix led only to the one line functional test change

test/Microsoft.AspNet.Mvc.TagHelpers.Test/InputTagHelperTest.cs
```diff
@@ -685,6 +842,19 @@ private class Model
             public NestedModel NestedModel { get; set; }
 
             public bool IsACar { get; set; }
+
+            [DataType(DataType.Date)]
+            public DateTime Date { get; set; }
```
pranavkm: Maybe name this `DateWithDataType` or something along those lines so it's easier to read as part of `MemberData`?

pranavkm:
:shipit:

test/Microsoft.AspNet.Mvc.TagHelpers.Test/InputTagHelperTest.cs
dougbu: fyi decided to leave this alone because the `DateTimeLocal` and `Time` properties could also be named `DateTimeWithDataType` or `DateTimeOffsetWithDataType`.  got too confusing.  the properties are all named after how they're intended to be used.

dougbu:
73e8fc1

PR #2353, opened 2015-04-07
Removing IInputFormatterSelector

pranavkm:
Follow up work item to #2269


src/Microsoft.AspNet.Mvc.Core/ModelBinders/BodyModelBinder.cs
```diff
@@ -48,7 +47,7 @@ public BodyModelBinder()
                 return new ModelBindingResult(model: null, key: bindingContext.ModelName, isModelSet: false);
             }
 
-            object model = null;
+            object model;
```
rynowak: You can safely move this all of the way into the `try`

rynowak:
:shipit:

pranavkm:
Updated with some dead code cleanup + new functional tests

src/Microsoft.AspNet.Mvc.Core/ModelBinders/BodyModelBinder.cs
```diff
@@ -48,34 +47,21 @@ public BodyModelBinder()
...
+                return new ModelBindingResult(model, key: string.Empty, isModelSet: true);
             }
             catch (Exception ex)
             {
-                model = GetDefaultValueForType(bindingContext.ModelType);
                 bindingContext.ModelState.AddModelError(bindingContext.ModelName, ex);
 
                 // This model binder is the only handler for the Body binding source.
                 // Always tell the model binding system to skip other model binders i.e. return non-null.
                 return new ModelBindingResult(model: null, key: bindingContext.ModelName, isModelSet: false);
```
kichalla: This was my bad. I think the first parameter here was supposed to be assigned the default value for type model calculated above?

pranavkm: * For action parameters, `MethodInfo.Invoke` seems to take care of coercing null value types to default values for value types (this is verified by test). 
* For value type properties that are annotated with `[FromBody]`, we end up getting a null ref as part of the property setter. Tracking this via https://github.com/aspnet/Mvc/issues/2357. 

That said, it seems like fixing it specifically for `BodyModelBinder` is incorrect and we should have a single top level piece of code that addresses this, not individual binders.

kichalla: :+1:

kichalla:
:shipit:

test/Microsoft.AspNet.Mvc.FunctionalTests/InputFormatterTests.cs
```diff
@@ -127,6 +127,43 @@ public class InputFormatterTests
         }
 
         [Theory]
+        [InlineData("\"I'm a JSON string!\"")]
+        [InlineData("true")]
+        public async Task JsonInputFormatter_ReturnsDefaultValue_ForValueTypes(string input)
```
harshgMSFT: How about an empty string content? that should be treated as a null value and a model state error should be added by JsonInputformatter

harshgMSFT:
:watch:  on one test case.

pranavkm:
Updated.

harshgMSFT:
:shipit:

PR #2360, opened 2015-04-08
[JsonPatch]Adding errors in ModelState

kirthik:
Added an extension method for `ApplyTo` in MVC that takes `ModelState` as a parameter.

@rynowak - Could you please let me know how this looks overall? I will add tests in subsequent iterations.


samples/MvcSample.Web/Controllers/JsonPatchController.cs
```diff
@@ -34,8 +35,13 @@ public IActionResult Patch([FromBody] JsonPatchDocument<Customer> patchDoc)
                 }
             };
 
-            patchDoc.ApplyTo(customer);
+            patchDoc.ApplyTo(customer, ModelState);
 
+            if (!ModelState.IsValid)
+            {
+                var entry = ModelState.SingleOrDefault(m => m.Key.Contains("JsonPatchDocument"));
+                return new ObjectResult(entry.Value.Errors[0].ErrorMessage);
```
rynowak: We'd want this to work with `return BadRequest(ModelState)` without any massaging.

src/Microsoft.AspNet.JsonPatch/Adapters/IObjectAdapter.cs
```diff
@@ -1,16 +1,19 @@
...
-        void Remove(Microsoft.AspNet.JsonPatch.Operations.Operation<T> operation, T objectToApplyTo);
-        void Replace(Microsoft.AspNet.JsonPatch.Operations.Operation<T> operation, T objectToApplyTo);
-        void Test(Microsoft.AspNet.JsonPatch.Operations.Operation<T> operation, T objectToApplyTo);
-    }
+        void Add(Operation<T> operation, T objectToApplyTo, Action<string> action);
+        void Copy(Operation<T> operation, T objectToApplyTo, Action<string> action);
+        void Move(Operation<T> operation, T objectToApplyTo, Action<string> action);
+        void Remove(Operation<T> operation, T objectToApplyTo, Action<string> action);
+        void Replace(Operation<T> operation, T objectToApplyTo, Action<string> action);
+        void Test(Operation<T> operation, T objectToApplyTo, Action<string> action);
```
rynowak: Not sure that changing the API of `IObjectAdapater` is the best way to go here

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -81,16 +81,21 @@ public ObjectAdapter(IContractResolver contractResolver)
...
         /// Add is used by various operations (eg: add, copy, ...), yet through different operations;
         /// This method allows code reuse yet reporting the correct operation on error
         /// </summary>
-        private void Add(string path, object value, T objectToApplyTo, Operation<T> operationToReport)
+        private void Add(
+            string path,
+            object value,
+            T objectToApplyTo,
+            Operation<T> operationToReport,
+            Action<string> action = null)
```
rynowak: why default? That says to me that there's a bunch of code calling this without an error action.

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -81,16 +81,21 @@ public ObjectAdapter(IContractResolver contractResolver)
         /// </summary>
         /// <param name="operation">The add operation</param>
         /// <param name="objectApplyTo">Object to apply the operation to</param>
-        public void Add(Operation<T> operation, T objectToApplyTo)
+        public void Add(Operation<T> operation, T objectToApplyTo, Action<string> action = null)
```
rynowak: The error reporting action could be a member of the `ObjectAdapter` instance, rather than a part of the interface.

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -471,20 +496,22 @@ public void Test(Operation<T> operation, T objectToApplyTo)
 
                     if (array.Count <= positionInPathAsInteger)
                     {
-                        throw new JsonPatchException<T>(
+                        LogError(
+                            objectToApplyTo,
                             operation,
                             Resources.FormatInvalidIndexForArrayProperty(operation.op, operation.path),
-                            objectToApplyTo);
+                            action);
```
rynowak: Should we stop if we encounter an error or keep going? That's the risk of changing thing to not-throw.

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -647,14 +680,32 @@ private bool IsNonStringArray(JsonPatchProperty patchProperty)
...
+            string errorMessage,
+            Action<string> action = null)
+        {
+            if (action != null)
+            {
+                action(errorMessage);
+            }
+            else
+            {
+                throw new JsonPatchException<T>(operation, errorMessage, objectToApplyTo);
```
rynowak: Should we introduce a type for this (`JsonPatchError`) rather than just a string?

src/Microsoft.AspNet.JsonPatch/TypedJsonPatchDocument.cs
```diff
@@ -357,13 +357,17 @@ public void ApplyTo(T objectToApplyTo)
             ApplyTo(objectToApplyTo, new ObjectAdapter<T>(ContractResolver));
         }
 
-        public void ApplyTo(T objectToApplyTo, IObjectAdapter<T> adapter)
+        public void ApplyTo(T objectToApplyTo, Action<string> action)
         {
+            ApplyTo(objectToApplyTo, new ObjectAdapter<T>(ContractResolver), action);
+        }
 
+        public void ApplyTo(T objectToApplyTo, IObjectAdapter<T> adapter, Action<string> action = null)
```
rynowak: We don't use default parameter values in public APIs (in general).

rynowak: Should we have another API where we take a prefix to use for model-state? (Take a look at how input formatters add errors).

Would it be possible to find all/most of the errors during deserialization? Worth thinking about?

src/Microsoft.AspNet.Mvc/JsonPatchExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// </summary>
+        /// <param name="patchDoc">The <see cref="JsonPatchDocument{T}"/>.</param>
+        /// <param name="objectToApplyTo">The entity on which <see cref="JsonPatchDocument{T}"/> is applied.</param>
+        /// <param name="modelState">The <see cref="ModelStateDictionary"/> to add errors.</param>
+        public static void ApplyTo<T>(
+            [NotNull] this JsonPatchDocument<T> patchDoc,
+            [NotNull] T objectToApplyTo,
+            [NotNull] ModelStateDictionary modelState) where T : class
+        {
+            var modelStateError = new ModelStateError(modelState, patchDoc.GetType().Name);
```
rynowak: You can just use a lambda here, not sure this class is generally reusable.

rynowak:
:watch: - it's a good start.

Eilon:
I have some general confusion with this, though it's not necessarily specific to this PR. There's a lot of stuff named "T" and "action" and it's not clear to me from the APIs what the purpose is of these generic arguments and parameters. Can they be given more specific names?

kirthik:
@Eilon - Are these ok  - `T` changed to `TModel` and `action` changed to `logError`?

src/Microsoft.AspNet.Mvc/JsonPatchExtensions.cs
kirthik: Makes sense. I will change it.

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
kirthik: What would `JsonPatchError` have? What value would it add?

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
kirthik: Yes we should stop. I will add a `return`.

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
kirthik: Synced up offline. Will fix it.

Eilon:
`T` -> `TModel` sounds fine. Maybe `action` to `logErrorAction`?

Eilon:
(Can do it in a separate PR)

kirthik:
@rynowak - Updated PR. Could you please take a look?
@Eilon  - I changed `action` to `logErrorAction` in this PR. I will change `T` to `TModel` in a separate PR.

src/Microsoft.AspNet.JsonPatch/Adapters/IObjectAdapter.cs
```diff
@@ -1,16 +1,19 @@
...
-        void Replace(Microsoft.AspNet.JsonPatch.Operations.Operation<T> operation, T objectToApplyTo);
-        void Test(Microsoft.AspNet.JsonPatch.Operations.Operation<T> operation, T objectToApplyTo);
-    }
+        void Add(Operation<T> operation, T objectToApplyTo);
+        void Copy(Operation<T> operation, T objectToApplyTo);
+        void Move(Operation<T> operation, T objectToApplyTo);
+        void Remove(Operation<T> operation, T objectToApplyTo);
+        void Replace(Operation<T> operation, T objectToApplyTo);
+        void Test(Operation<T> operation, T objectToApplyTo);
+	}
```
Eilon: Tab :open_mouth:

src/Microsoft.AspNet.Mvc/JsonPatchExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+
+namespace Microsoft.AspNet.Mvc
+{
+    /// <summary>
+    /// Extensions for <see cref="JsonPatchDocument{T}"/>
+    /// </summary>
+    public static class JsonPatchExtensions
+    {
+        /// <summary>
+        /// Applies json patch operations on object and logs errors in <see cref="ModelStateDictionary"/>.
```
Eilon: `json` -> `JSON`

Eilon:
Aside from the two small comments, :shipit: 

We do need to do an overall API review of JsonPatch once it's stabilized. I think some of the namespaces and type names can be cleaned up a bit.

samples/MvcSample.Web/Controllers/JsonPatchController.cs
```diff
@@ -34,8 +35,12 @@ public IActionResult Patch([FromBody] JsonPatchDocument<Customer> patchDoc)
                 }
             };
 
-            patchDoc.ApplyTo(customer);
+            patchDoc.ApplyTo(customer, ModelState);
 
+            if (!ModelState.IsValid)
+            {
+                return HttpBadRequest(ModelState);
+            }
```
harshgMSFT: nit: add a new line between this and return.

src/Microsoft.AspNet.JsonPatch/Adapters/IObjectAdapter.cs
```diff
@@ -1,16 +1,19 @@
 ﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
```
harshgMSFT: is this required?

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -16,9 +17,12 @@ public class ObjectAdapter<T> : IObjectAdapter<T> where T : class
     {
         public IContractResolver ContractResolver { get; set; }
 
-        public ObjectAdapter(IContractResolver contractResolver)
+        public Action<JsonPatchError<T>> LogErrorAction { get; set; }
```
harshgMSFT: docs?

harshgMSFT: all public apis on this class.

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -90,7 +94,11 @@ public void Add(Operation<T> operation, T objectToApplyTo)
         /// Add is used by various operations (eg: add, copy, ...), yet through different operations;
         /// This method allows code reuse yet reporting the correct operation on error
         /// </summary>
-        private void Add(string path, object value, T objectToApplyTo, Operation<T> operationToReport)
+        private void Add(
```
harshgMSFT: it looked better on one line

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -139,7 +147,11 @@ private void Add(string path, object value, T objectToApplyTo, Operation<T> oper
 
                     var conversionResult = PropertyHelpers.ConvertToActualType(genericTypeOfArray, value);
 
-                    CheckIfPropertyCanBeSet(conversionResult, objectToApplyTo, operationToReport, path);
+                    CheckIfPropertyCanBeSet(
```
harshgMSFT: it looked better on one line

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
harshgMSFT: do you need a public setter?

src/Microsoft.AspNet.Mvc/JsonPatchExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+        public static void ApplyTo<T>(
+            [NotNull] this JsonPatchDocument<T> patchDoc,
+            [NotNull] T objectToApplyTo,
+            [NotNull] ModelStateDictionary modelState,
+            string prefix) where T : class
+        {
+            if (prefix == null)
+            {
+                prefix = string.Empty;
+            }
```
harshgMSFT: nit : put a newline.

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
kirthik: Will add docs and remove public setter.

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
kirthik: I will fix it.

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
kirthik: Will fix it.

src/Microsoft.AspNet.JsonPatch/Adapters/IObjectAdapter.cs
kirthik: No.

src/Microsoft.AspNet.Mvc/JsonPatchExtensions.cs
kirthik: Will fix it.

src/Microsoft.AspNet.JsonPatch/Adapters/IObjectAdapter.cs
kirthik: Will fix it.

Eilon:
:shipit:

samples/MvcSample.Web/Controllers/JsonPatchController.cs
```diff
@@ -34,7 +35,12 @@ public IActionResult Patch([FromBody] JsonPatchDocument<Customer> patchDoc)
                 }
             };
 
-            patchDoc.ApplyTo(customer);
+            patchDoc.ApplyTo(customer, ModelState);
+
+            if (!ModelState.IsValid)
+            {
+                return HttpBadRequest(ModelState);
+            }
```
rynowak: Much better :+1:

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -7,18 +7,32 @@
...
+        /// <summary>
+        /// Gets or sets the <see cref="IContractResolver"/>.
+        /// </summary>
         public IContractResolver ContractResolver { get; set; }
 
-        public ObjectAdapter(IContractResolver contractResolver)
+        /// <summary>
+        /// Action for logging <see cref="JsonPatchError{T}"/>.
+        /// </summary>
+        public Action<JsonPatchError<T>> LogErrorAction { get; private set; }
```
rynowak: If you want to you can leave out the `private set` and earn PranavPoints to spend on cool prizes. Also properties go after constructors. Since you're doing other cleanup here, might as well do that too.

src/Microsoft.AspNet.Mvc/JsonPatchExtensions.cs
```diff
@@ -0,0 +1,53 @@
...
+
+            patchDoc.ApplyTo(objectToApplyTo, jsonPatchError =>
+            {
+                modelState.TryAddModelError(prefix + patchDoc.GetType().Name, jsonPatchError.ErrorMessage);
```
rynowak: This isn't really right is it? Wouldn't it be `<prefix>.<something>`.

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -179,7 +201,11 @@ private void Add(string path, object value, T objectToApplyTo, Operation<T> oper
                     value);
 
                 // Is conversion successful
-                CheckIfPropertyCanBeSet(conversionResultTuple, objectToApplyTo, operationToReport, path);
+                CheckIfPropertyCanBeSet(
+                    conversionResultTuple,
+                    objectToApplyTo,
+                    operationToReport,
+                    path);
```
rynowak: Does this throw? Should it have short circuiting behavior now?

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -622,17 +665,19 @@ public void Copy(Operation<T> operation, T objectToApplyTo)
...
             }
             if (patchProperty.Property.Ignored)
             {
-                throw new JsonPatchException<T>(
+                LogError(new JsonPatchError<T>(
+                    objectToApplyTo,
                     operation,
-                    Resources.FormatCannotUpdateProperty(propertyPath),
-                    objectToApplyTo);
+                    Resources.FormatCannotUpdateProperty(propertyPath)));
```
rynowak: This one doesn't have a `return` -- should it?

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -651,10 +696,22 @@ private bool IsNonStringArray(JsonPatchProperty patchProperty)
         {
             if (!result.CanBeConverted)
             {
-                throw new JsonPatchException<T>(
+                LogError(new JsonPatchError<T>(
+                    objectToApplyTo,
                     operation,
-                    Resources.FormatInvalidValueForProperty(result.ConvertedInstance, path),
-                    objectToApplyTo);
+                    Resources.FormatInvalidValueForProperty(result.ConvertedInstance, path)));
```
rynowak: This one doesn't have a return -- should it?

rynowak:
:watch: design looks :100: - ready for minor fixes and tests

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
kirthik: Added it.

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
kirthik: Added it.

kirthik:
@rynowak - Added tests and fixed minor issue mentioned above. Could you please take a look?

test/Microsoft.AspNet.JsonPatch.Test/NestedObjectTests.cs
```diff
@@ -1748,5 +1892,10 @@ public void MoveToEndOfListWithSerialization()
             Assert.Equal(0, doc.IntegerValue);
             Assert.Equal(new List<int>() { 1, 2, 3, 5 }, doc.SimpleDTO.IntegerList);
         }
+
+        private void LogErrorMessage(JsonPatchError<SimpleDTOWithNestedDTO> patchError)
+        {
+            ErrorMessage = patchError.ErrorMessage;
+        }
```
rynowak: I typically just avoid statefulness on the test class itself. Your call if you like it better by my or your way :laughing: 

I'd do this like so:

```
public class TestErrorLogger
{
    public string ErrorMessage { get; set; }
}

    private void LogErrorMessage(JsonPatchError<SimpleDTOWithNestedDTO> patchError)
    {
        ErrorMessage = patchError.ErrorMessage;
    }

var logger = new TestErrorLogger();
patchDoc.ApplyTo(doc, logger.LogErrorMessage);

```

rynowak: You can see the problem here with `Operation<T>` being generic. It's viral

Mvc.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.22726.0
+VisualStudioVersion = 14.0.22806.0
```
harshgMSFT: nit : 806 is bad .. update your VS :smile:

test/Microsoft.AspNet.Mvc.FunctionalTests/JsonPatchTest.cs
```diff
@@ -0,0 +1,146 @@
...
+                        "[{ \"op\": \"add\", \"path\": \"Customer/Orders/5\", " +
+                        "\"value\": { \"OrderName\": \"Name5\" }}]",
+                        "{\"Patch.JsonPatchDocument`1\":[\"For operation 'add' on array property at path " +
+                        "'Customer/Orders/5', the index is larger than the array size.\"]}"
+                    },
+                    new object[] {
+                        "http://localhost/JsonPatch/JsonPatchWithModelState",
+                        "[{ \"op\": \"add\", \"path\": \"Customer/Orders/5\", " +
+                        "\"value\": { \"OrderName\": \"Name5\" }}]",
+                        "{\"JsonPatchDocument`1\":[\"For operation 'add' on array property at path " +
```
rynowak: Can we do better than `JsonPatchDocument` for the error? I would think that's we'd use the property path that failed - or at least the inner-type of the document.

samples/MvcSample.Web/Views/JsonPatch/Index.cshtml
```diff
@@ -27,6 +27,28 @@
                 });
             });
 
+            $('#invalidaddorder').on("click", function () {
```
harshgMSFT: nit: should we camel case? `invalidAddOrder` ?

src/Microsoft.AspNet.JsonPatch/Adapters/IObjectAdapter.cs
```diff
@@ -1,16 +1,21 @@
...
+using Microsoft.AspNet.JsonPatch.Operations;
+
 namespace Microsoft.AspNet.JsonPatch.Adapters
 {
-    public interface IObjectAdapter<T>
-      where T : class
+    /// <summary>
+    /// Defines the operations that can be performed on a JSON patch document.
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
```
harshgMSFT: Better name than `T` ? Also missing docs.

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -7,21 +7,35 @@
...
+    /// <inheritdoc />
     public class ObjectAdapter<T> : IObjectAdapter<T> where T : class
     {
-        public IContractResolver ContractResolver { get; set; }
-
-        public ObjectAdapter(IContractResolver contractResolver)
+        /// <summary>
+        /// Initializes a new instance of <see cref="ObjectAdapter{T}"/>.
+        /// </summary>
+        /// <param name="contractResolver"></param>
```
harshgMSFT: docs.

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -7,21 +7,35 @@
...
+        public ObjectAdapter(IContractResolver contractResolver, Action<JsonPatchError<T>> logErrorAction)
         {
             ContractResolver = contractResolver;
+            LogErrorAction = logErrorAction;
         }
 
         /// <summary>
+        /// Gets or sets the <see cref="IContractResolver"/>.
+        /// </summary>
+        public IContractResolver ContractResolver { get; set; }
```
harshgMSFT: Do you need the set?

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -622,18 +681,24 @@ public void Copy(Operation<T> operation, T objectToApplyTo)
...
+                LogError(new JsonPatchError<T>(
+                    objectToApplyTo,
                     operation,
-                    Resources.FormatPropertyDoesNotExist(propertyPath),
-                    objectToApplyTo);
+                    Resources.FormatPropertyDoesNotExist(propertyPath)));
+
+                return false;
             }
             if (patchProperty.Property.Ignored)
```
harshgMSFT: nit : a newline before this would make this even better :)

src/Microsoft.AspNet.JsonPatch/JsonPatchError.cs
```diff
@@ -0,0 +1,30 @@
...
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.AspNet.JsonPatch
+{
+    /// <summary>
+    /// Captures error message and the related entity and operation that caused it.
+    /// </summary>
+    public class JsonPatchError<T> where T : class
+    {
+        public JsonPatchError(
```
harshgMSFT: docs for this and the properties.

src/Microsoft.AspNet.JsonPatch/JsonPatchError.cs
```diff
@@ -0,0 +1,30 @@
...
+            [NotNull] T affectedObject,
+            [NotNull] Operation<T> operation,
+            [NotNull] string errorMessage)
+        {
+            AffectedObject = affectedObject;
+            Operation = operation;
+            ErrorMessage = errorMessage;
+        }
+
+        public T AffectedObject { get; private set; }
```
harshgMSFT: you can skip `private set` for all the properties.

src/Microsoft.AspNet.JsonPatch/Operations/TypedOperation.cs
```diff
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
```
harshgMSFT: required?

src/Microsoft.AspNet.JsonPatch/TypedJsonPatchDocument.cs
```diff
@@ -354,12 +354,16 @@ public JsonPatchDocument<T> Copy<TProp>(Expression<Func<T, TProp>> from, Express
 
         public void ApplyTo(T objectToApplyTo)
         {
-            ApplyTo(objectToApplyTo, new ObjectAdapter<T>(ContractResolver));
+            ApplyTo(objectToApplyTo, new ObjectAdapter<T>(ContractResolver, null));
```
harshgMSFT: named parameter for null.

src/Microsoft.AspNet.Mvc/JsonPatchExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+        public static void ApplyTo<T>(
+            [NotNull] this JsonPatchDocument<T> patchDoc,
+            [NotNull] T objectToApplyTo,
+            [NotNull] ModelStateDictionary modelState) where T : class
+        {
+            patchDoc.ApplyTo(objectToApplyTo, modelState, prefix: null);
+        }
+
+        /// <summary>
+        /// Applies json patch operations on object and logs errors in <see cref="ModelStateDictionary"/>.
```
harshgMSFT: be consistent i think this should be `Applies JSON  patch`

src/Microsoft.AspNet.Mvc/JsonPatchExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// </summary>
+        /// <param name="patchDoc">The <see cref="JsonPatchDocument{T}"/>.</param>
+        /// <param name="objectToApplyTo">The entity on which <see cref="JsonPatchDocument{T}"/> is applied.</param>
+        /// <param name="modelState">The <see cref="ModelStateDictionary"/> to add errors.</param>
+        public static void ApplyTo<T>(
+            [NotNull] this JsonPatchDocument<T> patchDoc,
+            [NotNull] T objectToApplyTo,
+            [NotNull] ModelStateDictionary modelState) where T : class
+        {
+            patchDoc.ApplyTo(objectToApplyTo, modelState, prefix: null);
```
harshgMSFT: we have so far relied on prefix to be empty in case of `TryUpdateModel` etc, should we do the same here?

src/Microsoft.AspNet.Mvc/JsonPatchExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+        public static void ApplyTo<T>(
+            [NotNull] this JsonPatchDocument<T> patchDoc,
+            [NotNull] T objectToApplyTo,
+            [NotNull] ModelStateDictionary modelState,
+            string prefix) where T : class
+        {
+            var key = patchDoc.GetType().Name;
+            if (!string.IsNullOrEmpty(prefix))
+            {
+                key = prefix + "." + patchDoc.GetType().Name;
```
harshgMSFT: you can rewrite it as 
`key = prefix + "." +  key;` you do not need to re compute `patchDoc.GetType().Name`

test/Microsoft.AspNet.JsonPatch.Test/NestedObjectTests.cs
```diff
@@ -1092,6 +1212,30 @@ public void ReplaceInListInvalidPositionTooSmallWithSerialization()
...
+            patchDoc.Replace<int>(o => o.SimpleDTO.IntegerList, 5, -1);
+
+            // Act
+            patchDoc.ApplyTo(doc, LogErrorMessage);
+
+            // Assert
+            Assert.Equal("Property does not exist at path '/simpledto/integerlist/-1'.", ErrorMessage);
+        }
+
+
```
harshgMSFT: extra  line.

test/Microsoft.AspNet.JsonPatch.Test/NestedObjectTests.cs
harshgMSFT: :+1 with @rynowak here.

test/Microsoft.AspNet.JsonPatch.Test/NestedObjectTests.cs
```diff
@@ -316,6 +318,31 @@ public void AddToListInvalidPositionTooLargeWithSerialization()
...
+            // create patch
+            var patchDoc = new JsonPatchDocument<SimpleDTOWithNestedDTO>();
+            patchDoc.Add<int>(o => o.SimpleDTO.IntegerList, 4, 4);
+
+            // Act
+            patchDoc.ApplyTo(doc, LogErrorMessage);
+
+            //Assert
+            Assert.Equal("For operation 'add' on array property at path '/simpledto/integerlist/4', the index is " +
+                "larger than the array size.", ErrorMessage);
```
harshgMSFT: if you are splitting up lines, make sure that each argument gets a newline.

test/Microsoft.AspNet.JsonPatch.Test/NestedObjectTests.cs
```diff
@@ -1047,6 +1144,29 @@ public void ReplaceInListInvalidInvalidPositionTooLargeWithSerialization()
...
+                }
+            };
+
+            // create patch
+            var patchDoc = new JsonPatchDocument<SimpleDTOWithNestedDTO>();
+            patchDoc.Replace<int>(o => o.SimpleDTO.IntegerList, 5, 3);
+
+            // Act
+            patchDoc.ApplyTo(doc, LogErrorMessage);
+            // Assert
```
harshgMSFT: new line before this.

test/Microsoft.AspNet.JsonPatch.Test/NestedObjectTests.cs
```diff
@@ -1047,6 +1144,29 @@ public void ReplaceInListInvalidInvalidPositionTooLargeWithSerialization()
         }
 
         [Fact]
+        public void ReplaceInListInvalidInvalidPositionTooLarge_LogError()
```
harshgMSFT: `ReplaceInListInvalid_PositionTooLarge_LogError` would it make more sense?
nit: Also in general  would changing the suffix of the tests form `LogError` to `LogsError` make sense?

src/Microsoft.AspNet.JsonPatch/Adapters/IObjectAdapter.cs
kirthik: I am going to send a separate PR where I change `T` all across to `TModel`

test/WebSites/JsonPatchWebSite/Startup.cs
```diff
@@ -0,0 +1,36 @@
...
+        }
+
+        public void Configure(IApplicationBuilder app)
+        {
+            app.UseCultureReplacer();
+
+            // Set up file serving for JavaScript files.
+            app.UseFileServer();
+
+            app.UseErrorReporter();
```
harshgMSFT: do you need this ? I think the general guidance is not to use this.

test/WebSites/JsonPatchWebSite/Controllers/JsonPatchController.cs
```diff
@@ -0,0 +1,64 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using JsonPatchWebSite.Models;
+using Microsoft.AspNet.JsonPatch;
+using Microsoft.AspNet.Mvc;
+
+namespace JsonPatchWebSite.Controllers
+{
+    public class JsonPatchController : Controller
```
harshgMSFT: general guidance for any functional test case is to use attribute routing.

test/Microsoft.AspNet.Mvc.Test/JsonPatchExtensionsTest.cs
```diff
@@ -0,0 +1,47 @@
...
+using Xunit;
+
+namespace Microsoft.AspNet.Mvc
+{
+    public class JsonPatchExtensionsTest
+    {
+        [Fact]
+        public void ApplyTo_JsonPatchDocument_ModelState()
+        {
+            var operation = new Operation<Customer>("add", "Customer/CustomerId", null, "TestName");
```
harshgMSFT: use a named parameter for null.

harshgMSFT: Missing // Act Assert etc.

test/Microsoft.AspNet.Mvc.Test/JsonPatchExtensionsTest.cs
```diff
@@ -0,0 +1,47 @@
...
+            var patchDoc = new JsonPatchDocument<Customer>();
+            patchDoc.Operations.Add(operation);
+
+            var modelState = new ModelStateDictionary();
+
+            patchDoc.ApplyTo(new Customer(), modelState);
+
+            Assert.Equal("Property does not exist at path 'Customer/CustomerId'.",
+                modelState["JsonPatchDocument`1"].Errors[0].ErrorMessage);
```
harshgMSFT: Should you also verify that there is only one error?

test/Microsoft.AspNet.Mvc.Test/JsonPatchExtensionsTest.cs
```diff
@@ -0,0 +1,47 @@
...
+            var operation = new Operation<Customer>("add", "Customer/CustomerId", null, "TestName");
+            var patchDoc = new JsonPatchDocument<Customer>();
+            patchDoc.Operations.Add(operation);
+
+            var modelState = new ModelStateDictionary();
+
+            patchDoc.ApplyTo(new Customer(), modelState, "jsonpatch");
+
+            Assert.Equal("Property does not exist at path 'Customer/CustomerId'.",
+                modelState["jsonpatch.JsonPatchDocument`1"].Errors[0].ErrorMessage);
```
harshgMSFT: verify that this has one error: essentially 
```c#
var  error = Assert.Single(modelState["jsonpatch.JsonPatchDocument`1"].Errors);
Assert.Equal("Property does not exist at path 'Customer/CustomerId'.", error.ErrorMessage);
```

test/Microsoft.AspNet.Mvc.FunctionalTests/JsonPatchTest.cs
```diff
@@ -0,0 +1,146 @@
...
+            // Act
+            var response = await client.SendAsync(request);
+
+            // Assert
+            var body = await response.Content.ReadAsStringAsync();
+            var customer = JsonConvert.DeserializeObject<Customer>(body);
+            Assert.Equal("Name2", customer.Orders[2].OrderName);
+        }
+
+
```
harshgMSFT: remove extra line.

harshgMSFT:
:watch: looks good overall. Waiting on cosmetic fixes.

src/Microsoft.AspNet.Mvc/JsonPatchExtensions.cs
kirthik: Will change to `string.Empty` to be consistent.

src/Microsoft.AspNet.Mvc/JsonPatchExtensions.cs
kirthik: Yes. I will change it.

test/Microsoft.AspNet.Mvc.Test/JsonPatchExtensionsTest.cs
kirthik: Will do that.

test/Microsoft.AspNet.Mvc.Test/JsonPatchExtensionsTest.cs
kirthik: will do that.

kirthik:
@harshgMSFT , @rynowak : updated PR

harshgMSFT:
:shipit: when @rynowak  is happy.

rynowak:
:shipit:

kirthik:
checked in - 3962c597f3619d24c04822dd7e492ca9253ba5a3

RehanSaeed:
The overload of `ApplyTo` with `ModelState` throws exceptions:
- `ArgumentException` when the patch document is invalid e.g. typo in the `op` property.
- `InvalidOperationException` when using a value of `test` for the `op` property.
I expected it to add an error into the `ModelState` instead. Right now, I'm having to wrap it in a try-catch and add `ModelState` errors manually.

RehanSaeed:
I've resorted to writing the following extension method:

```
public static class JsonPatchDocumentExtensions
{
    public static void SafeApplyTo<T>(this JsonPatchDocument<T> patch, T value, ModelStateDictionary modelState)
        where T : class
    {
        try
        {
            patch.ApplyTo(value, modelState);
        }
        catch (ArgumentException argumentException)
        {
            modelState.AddModelError(
                typeof(T).Name,
                argumentException.Message);
        }
        catch (NotSupportedException notSupportedException)
        {
            modelState.AddModelError(
                typeof(T).Name,
                notSupportedException.Message);
        }
    }
}
```

Unit tests:

```
public class JsonPatchDocumentExtensionsTest
{
    [Fact]
    public void SafeApplyTo_ValidPatchParameters_NoErrorsAdded()
    {
        var patch = new JsonPatchDocument<Model>();
        patch.Operations.Add(new Operation<Model>("replace", "/Property", null, "Foo"));
        var model = new Model();

        patch.SafeApplyTo(model, modelState);

    }

    [Theory]
    [InlineData("replace", "/PropertyNotFound", null, "Foo", "The property at path '/PropertyNotFound' could not be removed.")]
    [InlineData("test", "/Property", null, "Foo", "The test operation is not supported.")]
    [InlineData("invalid", "/Property", null, "Foo", "Requested value 'invalid' was not found.")]
    public void SafeApplyTo_PatchParameters_AddsModelStateError(
        string op,
        string path,
        string from,
        string value,
        string error)
    {
        var patch = new JsonPatchDocument<Model>();
        patch.Operations.Add(new Operation<Model>(op, path, from, value));
        var model = new Model();

        patch.SafeApplyTo(model, modelState);

        Assert.Equal(1, modelState.ErrorCount);
        Assert.Equal(nameof(Model), modelState.First().Key);
        Assert.Equal(1, modelState.First().Value.Errors.Count);
        Assert.Equal(error, modelState.First().Value.Errors.First().ErrorMessage);
    }

    public class Model
    {
        public string Property { get; set; }
    }
}
```

rynowak:
@RehanSaeed - can you open a separate issue so we can address this?

RehanSaeed:
@rynowak Done https://github.com/aspnet/Mvc/issues/5463

PR #2362, opened 2015-04-08
Adding a unit test to verify if CacheTagHelper gets disabled when setting the enabled flag to false.

sornaks:
/cc: @dougbu @NTaylorMullen

### DIFF
--- a/test/Microsoft.AspNet.Mvc.TagHelpers.Test/CacheTagHelperTest.cs
+++ b/test/Microsoft.AspNet.Mvc.TagHelpers.Test/CacheTagHelperTest.cs
@@ -243,7 +243,6 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
             var id = "unique-id";
             var childContent = "original-child-content";
             var cache = new Mock<IMemoryCache>();
-            cache.CallBase = true;
             cache.Setup(c => c.Set(
                 /*key*/ It.IsAny<string>(),
                 /*link*/ It.IsAny<IEntryLink>(),
@@ -283,7 +282,6 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
             var id = "unique-id";
             var childContent = "original-child-content";
             var cache = new Mock<IMemoryCache>();
-            cache.CallBase = true;
             cache.Setup(c => c.Set(
                 /*key*/ It.IsAny<string>(),
                 /*link*/ It.IsAny<IEntryLink>(),
@@ -311,6 +309,47 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
             Assert.Empty(tagHelperOutput.PostContent.GetContent());
             Assert.True(tagHelperOutput.IsContentModified);
             Assert.Equal(childContent, tagHelperOutput.Content.GetContent());
+            cache.Verify(c => c.Set(
+                /*key*/ It.IsAny<string>(),
+                /*link*/ It.IsAny<IEntryLink>(),
+                /*state*/ It.IsAny<object>(),
+                /*create*/ It.IsAny<Func<ICacheSetContext, object>>()),
+                Times.Once);
+        }
+
+        [Fact]
+        public async Task ProcessAsync_DoesNotUseCache_OnPassingFalseWhenEnabled()
+        {
+            // Arrange
+            var cache = new Mock<IMemoryCache>();
+            cache.Setup(c => c.Set(
+                /*key*/ It.IsAny<string>(),
+                /*link*/ It.IsAny<IEntryLink>(),
+                /*state*/ It.IsAny<object>(),
+                /*create*/ It.IsAny<Func<ICacheSetContext, object>>()))
+                .Returns(new DefaultTagHelperContent().SetContent("ok"))
+                .Verifiable();
+            object cacheResult;
+            cache.Setup(c => c.TryGetValue(It.IsAny<string>(), It.IsAny<IEntryLink>(), out cacheResult))
+                .Returns(false);
+            var tagHelperContext = GetTagHelperContext("unique-id", "original-child-content");
+            var tagHelperOutput = new TagHelperOutput("cache", new Dictionary<string, object>());
+            var cacheTagHelper = new CacheTagHelper
+            {
+                ViewContext = GetViewContext(),
+                MemoryCache = cache.Object,
+                Enabled = true
+            };
+
+            // Act
+            await cacheTagHelper.ProcessAsync(tagHelperContext, tagHelperOutput);
+
+            // set enabled to false in the cache tag helper.
+            cacheTagHelper.Enabled = false;
+            await cacheTagHelper.ProcessAsync(tagHelperContext, tagHelperOutput);
+
+            // Assert
+            // if cache was not disabled this would have been called more than once.
             cache.Verify(c => c.Set(
                 /*key*/ It.IsAny<string>(),
                 /*link*/ It.IsAny<IEntryLink>(),


test/Microsoft.AspNet.Mvc.TagHelpers.Test/CacheTagHelperTest.cs
```diff
@@ -320,6 +318,47 @@ public void GenerateKey_WithMultipleVaryByOptions_CreatesCombinedKey()
         }
 
         [Fact]
+        public async Task ProcessAsync_DoesNotUseCache_OnPassingFalseWhenEnabled()
```
NTaylorMullen: `ProcessAsync_DoesNotUseCache_OnPassingFalseWhenEnabled` => `ProcessAsync_DoesNotUseCache_WhenEnabledFalse`

test/Microsoft.AspNet.Mvc.TagHelpers.Test/CacheTagHelperTest.cs
```diff
@@ -320,6 +318,47 @@ public void GenerateKey_WithMultipleVaryByOptions_CreatesCombinedKey()
...
+            {
+                ViewContext = GetViewContext(),
+                MemoryCache = cache.Object,
+                Enabled = true
+            };
+
+            // Act
+            await cacheTagHelper.ProcessAsync(tagHelperContext, tagHelperOutput);
+
+            // set enabled to false in the cache tag helper.
```
NTaylorMullen: self explanatory, no need for comment.

NTaylorMullen:
:shipit:

test/Microsoft.AspNet.Mvc.TagHelpers.Test/CacheTagHelperTest.cs
```diff
@@ -320,6 +318,47 @@ public void GenerateKey_WithMultipleVaryByOptions_CreatesCombinedKey()
...
         [Fact]
+        public async Task ProcessAsync_DoesNotUseCache_OnPassingFalseWhenEnabled()
+        {
+            // Arrange
+            var cache = new Mock<IMemoryCache>();
+            cache.Setup(c => c.Set(
+                /*key*/ It.IsAny<string>(),
+                /*link*/ It.IsAny<IEntryLink>(),
+                /*state*/ It.IsAny<object>(),
+                /*create*/ It.IsAny<Func<ICacheSetContext, object>>()))
```
dougbu: nit: mock setups in MVC tend to use trailing comments e.g. `It.IsAny<object>(),    // state`

test/Microsoft.AspNet.Mvc.TagHelpers.Test/CacheTagHelperTest.cs
```diff
@@ -283,7 +282,6 @@ public void GenerateKey_WithMultipleVaryByOptions_CreatesCombinedKey()
             var id = "unique-id";
             var childContent = "original-child-content";
             var cache = new Mock<IMemoryCache>();
-            cache.CallBase = true;
```
dougbu: why are these lines changing?

test/Microsoft.AspNet.Mvc.TagHelpers.Test/CacheTagHelperTest.cs
```diff
@@ -320,6 +318,47 @@ public void GenerateKey_WithMultipleVaryByOptions_CreatesCombinedKey()
...
+            await cacheTagHelper.ProcessAsync(tagHelperContext, tagHelperOutput);
+
+            // Assert
+            // if cache was not disabled this would have been called more than once.
+            cache.Verify(c => c.Set(
+                /*key*/ It.IsAny<string>(),
+                /*link*/ It.IsAny<IEntryLink>(),
+                /*state*/ It.IsAny<object>(),
+                /*create*/ It.IsAny<Func<ICacheSetContext, object>>()),
+                Times.Once);
```
dougbu: why is this method called at all when caching is disabled?

test/Microsoft.AspNet.Mvc.TagHelpers.Test/CacheTagHelperTest.cs
```diff
@@ -320,6 +318,47 @@ public void GenerateKey_WithMultipleVaryByOptions_CreatesCombinedKey()
...
+            cache.Setup(c => c.Set(
+                /*key*/ It.IsAny<string>(),
+                /*link*/ It.IsAny<IEntryLink>(),
+                /*state*/ It.IsAny<object>(),
+                /*create*/ It.IsAny<Func<ICacheSetContext, object>>()))
+                .Returns(new DefaultTagHelperContent().SetContent("ok"))
+                .Verifiable();
+            object cacheResult;
+            cache.Setup(c => c.TryGetValue(It.IsAny<string>(), It.IsAny<IEntryLink>(), out cacheResult))
+                .Returns(false);
```
dougbu: does test fail without this setup?  expected cache to be completely unused when tag helper is disabled.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/CacheTagHelperTest.cs
sornaks: The test case is about having it enabled initially and then making a call with disabled.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/CacheTagHelperTest.cs
sornaks: CallBase is not needed. It is an interface..

test/Microsoft.AspNet.Mvc.TagHelpers.Test/CacheTagHelperTest.cs
NTaylorMullen: It's called once for the first `ProcessAsync`

test/Microsoft.AspNet.Mvc.TagHelpers.Test/CacheTagHelperTest.cs
```diff
@@ -320,6 +318,47 @@ public void GenerateKey_WithMultipleVaryByOptions_CreatesCombinedKey()
...
+            var tagHelperContext = GetTagHelperContext("unique-id", "original-child-content");
+            var tagHelperOutput = new TagHelperOutput("cache", new Dictionary<string, object>());
+            var cacheTagHelper = new CacheTagHelper
+            {
+                ViewContext = GetViewContext(),
+                MemoryCache = cache.Object,
+                Enabled = true
```
dougbu: OIC.  this is a bizarre test.  highly unlikely `Enabled` will change after initialization, never after `ProcessAsync` has run.

Eilon: Hmm right. Is it even possible for a tag helper's property to change after being initialized? (In any sane way.)

sornaks: I dont think it is. So I was trying to test a scenario where one cache tag helper had some value in the cache and a successive cache tag helper tried to access the cache with `enabled=false` in which case the cache must not be used. I am going to write a functional test for this instead (without any time bounds).

dougbu:
> I am going to write a functional test for this instead

@sornaks should this PR be closed now?

sornaks:
Yeah. I'll send that in a separate PR.

Issue #2267, opened 2015-03-25
Move responsibility for setting ModelState.MaxAllowedErrors into model binding


rynowak:
From API review:

Currently this code is in MvcRouteHandler which is not where we want it. It should be in the model binding entry point so that it's enforced for all calls into model binding.

sornaks:
https://github.com/aspnet/Mvc/commit/064c01cf2b022e365f82e5729a616e17b7f54d9a

PR #2363, opened 2015-04-09
#2267 - Moving responsibility for setting ModelState.MaxAllowedErrors into a more appropriate location.

sornaks:
/cc: @pranavkm @rynowak

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ControllerActionInvoker.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ControllerActionInvoker.cs
@@ -30,7 +30,8 @@ namespace Microsoft.AspNet.Mvc.Core
             [NotNull] IReadOnlyList<IModelValidatorProvider> modelValidatorProviders,
             [NotNull] IReadOnlyList<IValueProviderFactory> valueProviderFactories,
             [NotNull] IScopedInstance<ActionBindingContext> actionBindingContextAccessor,
-            [NotNull] ITempDataDictionary tempData)
+            [NotNull] ITempDataDictionary tempData,
+            int maxAllowedErrorsInModelState)
             : base(
                   actionContext,
                   filterProviders,
@@ -39,7 +40,8 @@ namespace Microsoft.AspNet.Mvc.Core
                   modelBinders, 
                   modelValidatorProviders, 
                   valueProviderFactories,
-                  actionBindingContextAccessor)
+                  actionBindingContextAccessor,
+                  maxAllowedErrorsInModelState)
         {
             _descriptor = descriptor;
             _controllerFactory = controllerFactory;
--- a/src/Microsoft.AspNet.Mvc.Core/ControllerActionInvoker.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ControllerActionInvoker.cs
@@ -31,7 +31,7 @@ namespace Microsoft.AspNet.Mvc.Core
             [NotNull] IReadOnlyList<IValueProviderFactory> valueProviderFactories,
             [NotNull] IScopedInstance<ActionBindingContext> actionBindingContextAccessor,
             [NotNull] ITempDataDictionary tempData,
-            int maxAllowedErrorsInModelState)
+            int maxModelValidationErrors)
             : base(
                   actionContext,
                   filterProviders,
@@ -41,7 +41,7 @@ namespace Microsoft.AspNet.Mvc.Core
                   modelValidatorProviders, 
                   valueProviderFactories,
                   actionBindingContextAccessor,
-                  maxAllowedErrorsInModelState)
+                  maxModelValidationErrors)
         {
             _descriptor = descriptor;
             _controllerFactory = controllerFactory;
--- a/src/Microsoft.AspNet.Mvc.Core/ControllerActionInvokerProvider.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ControllerActionInvokerProvider.cs
@@ -22,6 +22,7 @@ namespace Microsoft.AspNet.Mvc.Core
         private readonly IReadOnlyList<IValueProviderFactory> _valueProviderFactories;
         private readonly IScopedInstance<ActionBindingContext> _actionBindingContextAccessor;
         private readonly ITempDataDictionary _tempData;
+        private readonly int _maxAllowedErrorsInModelState;
 
         public ControllerActionInvokerProvider(
             IControllerFactory controllerFactory,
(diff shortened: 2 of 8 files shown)


test/Microsoft.AspNet.Mvc.Core.Test/ControllerActionInvokerTest.cs
```diff
@@ -1958,18 +1959,37 @@ public void CreateActionResult_ReturnsObjectContentResult(Type type, object inpu
...
+            var expected = 147;
+            var filter = new MockAuthorizationFilter(expected);
+            var invoker = CreateInvoker(
+                filter,
+                actionThrows: false,
+                tempData: null,
+                maxAllowedErrorsInModelState: expected);
+
+            // Act & Assert
+            // The authorization filter asserts if MaxAllowedErrors was set to the right value.
```
harshgMSFT: nit: could have used a mock with a callback to assert. That way the assert would have been in the test case and would be clear. Also you would not need to be a specific filter (authorization filter), it could have been a simple IFilter.

src/Microsoft.AspNet.Mvc.Core/ControllerActionInvoker.cs
```diff
@@ -30,7 +30,8 @@ public class ControllerActionInvoker : FilterActionInvoker
             [NotNull] IReadOnlyList<IModelValidatorProvider> modelValidatorProviders,
             [NotNull] IReadOnlyList<IValueProviderFactory> valueProviderFactories,
             [NotNull] IScopedInstance<ActionBindingContext> actionBindingContextAccessor,
-            [NotNull] ITempDataDictionary tempData)
+            [NotNull] ITempDataDictionary tempData,
+            int maxAllowedErrorsInModelState)
```
dougbu: `maxModelValidationErrors` might be a bit more clear (consistent with the option name)

dougbu:
:shipit: after fixing the parameter / field names and @harshgMSFT signs off

src/Microsoft.AspNet.Mvc.Core/ControllerActionInvoker.cs
harshgMSFT: :+1:

src/Microsoft.AspNet.Mvc.Core/ControllerActionInvokerProvider.cs
```diff
@@ -67,7 +69,8 @@ public void OnProvidersExecuting([NotNull] ActionInvokerProviderContext context)
                                     _modelValidatorProviders,
                                     _valueProviderFactories,
                                     _actionBindingContextAccessor,
-                                    _tempData);
+                                    _tempData,
+                                    _maxAllowedErrorsInModelState);
```
pranavkm: This is different from what we'd discussed - but you could assign the value here rather than flow it two additional classes: `context.ActionContext.ModelStateDictionary = _maxModelValidationErrors;;`.It's less obvious, but it's fewer lines of code change. @rynowak,  thoughts?

rynowak:
This bug is/was stale and this was already added to the modeling binding code: https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs#L92

`DefaultControllerActionArgumentBinder` is a better place to do this re: separation of concerns anyway. We already have a reasonable default for `ModelState.MaxModelErrors` so a user would have to work hard to avoid this.

Suggest just removing the code from `MvcRouteHandler`.

harshgMSFT:
@rynowak I thought the whole idea was to do it right before  filter invocation so that `ActionContext.ModelState` has the correct value to begin with, doing it later (as we do it now) would mean that any filter that overrode the value would not be effective. i.e it would be reset.

Eilon:
Confused... doesn't DI create the action invoker? If so, how does it work with there being a new `int` parameter?

rynowak:
@harshgMSFT - That sounds like a reasonable reason - it just kinda conflicts with options in the first place. Why provide this obscure setting in options if it's possible to change in a filter? Why provide the ability to change it in a filter if it's part of options?  Anyone relying on this setting being a precise tool to optimise around is doing something bizzare.

@Eilon invokers are created per-request from a factory - the factory is DI-ed and that's the one mediating between options and the `ModelState`

Eilon:
@rynowak oh, of course, thanks :smile:

pranavkm:
> Why provide the ability to change it in a filter if it's part of options?

There's also the secondary issue of the user performing model binding as part of one the early filters which now need to respect the set value.

src/Microsoft.AspNet.Mvc.Core/ControllerActionInvokerProvider.cs
rynowak: Nah, do it in the invoker not in the factory.

src/Microsoft.AspNet.Mvc.Core/MvcRouteHandler.cs
```diff
@@ -93,9 +93,6 @@ public VirtualPathData GetVirtualPath([NotNull] VirtualPathContext context)
 
             var actionContext = new ActionContext(context.HttpContext, context.RouteData, actionDescriptor);
 
-            var optionsAccessor = services.GetRequiredService<IOptions<MvcOptions>>();
-            actionContext.ModelState.MaxAllowedErrors = optionsAccessor.Options.MaxModelValidationErrors;
-
```
rynowak: There's more code to delete from `DefaultControllerActionArgumentBinder`

test/Microsoft.AspNet.Mvc.Core.Test/ControllerActionInvokerTest.cs
sornaks: Not doing this nit because I think not having a mock here is more understandable and straight-forward.

sornaks:
Updated the PR.

harshgMSFT:
:shipit:

src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
```diff
@@ -89,7 +89,6 @@ private void ActivateProperties(object controller, Type containerType, Dictionar
             IDictionary<string, object> arguments,
             IEnumerable<ParameterDescriptor> parameterMetadata)
         {
-            modelState.MaxAllowedErrors = _options.MaxModelValidationErrors;
```
rynowak: We can completely remove options from this class now.

rynowak:
:watch:

sornaks:
Updated..

rynowak:
:shipit:

sornaks:
Checked in. Thanks! https://github.com/aspnet/Mvc/commit/064c01cf2b022e365f82e5729a616e17b7f54d9a

Issue #2337, opened 2015-04-04
Automatically chose `type="file"` when generating `<input/>` elements for `IFormFile` expressions


dougbu:
`@Html.Editor()` and the `<input/>` tag helper choose appropriate `type` attribute values based in part on the given expression's type.  these helpers do not currently handle `IFormFile` or `IEnumerable<IFormFile>` properties automatically.  fixing this is pretty simple and would improve the end-to-end file upload scenario.

PR #2366, opened 2015-04-09
Added support for file type in input taghelper and htmlhelper

ajaybhargavb:
Issue - #2337 
@dougbu @NTaylorMullen

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultEditorTemplates.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultEditorTemplates.cs
@@ -5,11 +5,12 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Globalization;
-using System.Linq;
 using System.Text;
 using Microsoft.AspNet.Mvc.Core;
 using Microsoft.AspNet.Mvc.ModelBinding;
+using Microsoft.AspNet.Mvc.Rendering.Internal;
 using Microsoft.Framework.DependencyInjection;
+using Microsoft.Framework.Internal;
 
 namespace Microsoft.AspNet.Mvc.Rendering
 {
@@ -374,6 +375,20 @@ namespace Microsoft.AspNet.Mvc.Rendering
             return GenerateTextBox(htmlHelper, inputType: "number");
         }
 
+        public static string FileInputTemplate([NotNull] IHtmlHelper htmlHelper)
+        {
+            return GenerateTextBox(htmlHelper, inputType: "file");
+        }
+
+        public static string FileCollectionInputTemplate([NotNull] IHtmlHelper htmlHelper)
+        {
+            var htmlAttributes =
+                CreateHtmlAttributes(htmlHelper, className: "text-box single-line", inputType: "file");
+            htmlAttributes["multiple"] = "multiple";
+
+            return GenerateTextBox(htmlHelper, htmlHelper.ViewData.TemplateInfo.FormattedModelValue, htmlAttributes);
+        }
+
         private static void ApplyRfc3339DateFormattingIfNeeded(IHtmlHelper htmlHelper, string format)
         {
             if (htmlHelper.Html5DateRenderingMode != Html5DateRenderingMode.Rfc3339)
@@ -405,6 +420,11 @@ namespace Microsoft.AspNet.Mvc.Rendering
             var htmlAttributes =
                 CreateHtmlAttributes(htmlHelper, className: "text-box single-line", inputType: inputType);
 
+            return GenerateTextBox(htmlHelper, value, htmlAttributes);
+        }
+
+        private static string GenerateTextBox(IHtmlHelper htmlHelper, object value, object htmlAttributes)
+        {
             return htmlHelper.TextBox(
                 current: null,
                 value: value,
--- a/src/Microsoft.AspNet.Mvc.Core/Rendering/Internal/TemplateRenderer.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Rendering/Internal/TemplateRenderer.cs
@@ -7,16 +7,19 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Mvc.Core;
+using Microsoft.AspNet.Mvc.ModelBinding;
 using Microsoft.Framework.DependencyInjection;
 using Microsoft.Framework.Internal;
 
-namespace Microsoft.AspNet.Mvc.Rendering
+namespace Microsoft.AspNet.Mvc.Rendering.Internal
 {
-    internal class TemplateRenderer
+    public class TemplateRenderer
     {
-        private static readonly string DisplayTemplateViewPath = "DisplayTemplates";
-        private static readonly string EditorTemplateViewPath = "EditorTemplates";
+        private const string DisplayTemplateViewPath = "DisplayTemplates";
+        private const string EditorTemplateViewPath = "EditorTemplates";
+        public const string IEnumerableOfIFormFileName = "IEnumerable`" + nameof(IFormFile);
 
         private static readonly Dictionary<string, Func<IHtmlHelper, string>> _defaultDisplayActions =
             new Dictionary<string, Func<IHtmlHelper, string>>(StringComparer.OrdinalIgnoreCase)
@@ -62,6 +65,8 @@ namespace Microsoft.AspNet.Mvc.Rendering
                 { typeof(decimal).Name, DefaultEditorTemplates.DecimalTemplate },
                 { typeof(string).Name, DefaultEditorTemplates.StringTemplate },
                 { typeof(object).Name, DefaultEditorTemplates.ObjectTemplate },
+                { typeof(IFormFile).Name, DefaultEditorTemplates.FileInputTemplate },
+                { IEnumerableOfIFormFileName, DefaultEditorTemplates.FileCollectionInputTemplate },
             };
 
         private ViewContext _viewContext;
@@ -136,7 +141,7 @@ namespace Microsoft.AspNet.Mvc.Rendering
                 metadata.DataTypeName
             };
 
-            foreach (string templateHint in templateHints.Where(s => !string.IsNullOrEmpty(s)))
+            foreach (var templateHint in templateHints.Where(s => !string.IsNullOrEmpty(s)))
             {
                 yield return templateHint;
             }
@@ -146,14 +151,27 @@ namespace Microsoft.AspNet.Mvc.Rendering
             var modelType = _viewData.ModelExplorer.ModelType;
             var fieldType = Nullable.GetUnderlyingType(modelType) ?? modelType;
 
-            yield return fieldType.Name;
+            foreach (var typeName in GetTypeNames(_viewData.ModelExplorer.Metadata, fieldType))
+            {
+                yield return typeName;
+            }
+        }
+
+        public static IEnumerable<string> GetTypeNames(ModelMetadata modelMetadata, Type fieldType)
+        {
+            // Not returning type name here for IEnumerable<IFormFile> since we will be returning
+            // a more specific name, IEnumerableOfIFormFileName.
+            if (typeof(IEnumerable<IFormFile>) != fieldType)
+            {
+                yield return fieldType.Name;
+            }
 
             if (fieldType == typeof(string))
             {
                 // Nothing more to provide
                 yield break;
             }
-            else if (!metadata.IsComplexType)
+            else if (!modelMetadata.IsComplexType)
             {
                 // IsEnum is false for the Enum class itself
                 if (fieldType.IsEnum())
@@ -167,36 +185,44 @@ namespace Microsoft.AspNet.Mvc.Rendering
                 }
 
                 yield return "String";
+                yield break;
             }
-            else if (fieldType.IsInterface())
+            else if (!fieldType.IsInterface())
             {
-                if (typeof(IEnumerable).IsAssignableFrom(fieldType))
-                {
-                    yield return "Collection";
-                }
-
-                yield return "Object";
-            }
-            else
-            {
-                var isEnumerable = typeof(IEnumerable).IsAssignableFrom(fieldType);
-
+                var type = fieldType;
                 while (true)
                 {
-                    fieldType = fieldType.BaseType();
-                    if (fieldType == null)
+                    type = type.BaseType();
+                    if (type == null || type == typeof(object))
                     {
                         break;
                     }
 
-                    if (isEnumerable && fieldType == typeof(Object))
-                    {
-                        yield return "Collection";
-                    }
-
-                    yield return fieldType.Name;
+                    yield return type.Name;
                 }
             }
+
+            if (typeof(IEnumerable).IsAssignableFrom(fieldType))
+            {
+                if (typeof(IEnumerable<IFormFile>).IsAssignableFrom(fieldType))
+                {
+                    yield return IEnumerableOfIFormFileName;
+
+                    // Specific name has already been returned, now return the generic name.
+                    if (typeof(IEnumerable<IFormFile>) == fieldType)
+                    {
+                        yield return fieldType.Name;
+                    }
+                }
+
+                yield return "Collection";
+            }
+            else if (typeof(IFormFile) != fieldType && typeof(IFormFile).IsAssignableFrom(fieldType))
+            {
+                yield return nameof(IFormFile);
+            }
+
+            yield return "Object";
         }
 
         private static IHtmlHelper MakeHtmlHelper(ViewContext viewContext, ViewDataDictionary viewData)
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
@@ -2,11 +2,12 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Mvc.ModelBinding;
 using Microsoft.AspNet.Mvc.Rendering;
+using Microsoft.AspNet.Mvc.Rendering.Internal;
 using Microsoft.AspNet.Razor.Runtime.TagHelpers;
 
 namespace Microsoft.AspNet.Mvc.TagHelpers
@@ -47,6 +48,8 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
                 { nameof(Boolean), InputType.CheckBox.ToString().ToLowerInvariant() },
                 { nameof(Decimal), InputType.Text.ToString().ToLowerInvariant() },
                 { nameof(String), InputType.Text.ToString().ToLowerInvariant() },
+                { nameof(IFormFile), "file" },
+                { TemplateRenderer.IEnumerableOfIFormFileName, "file" },
             };
 
         // Mapping from <input/> element's type to RFC 3339 date and time formats.
@@ -271,13 +274,22 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
                 format = GetFormat(modelExplorer, inputTypeHint, inputType);
             }
 
+            object htmlAttributes = null;
+            if (string.Equals(inputType, "file") && string.Equals(inputTypeHint, TemplateRenderer.IEnumerableOfIFormFileName))
+            {
+                htmlAttributes = new Dictionary<string, object>
+                {
+                    { "multiple", "multiple" }
+                };
+            }
+
             return Generator.GenerateTextBox(
                 ViewContext,
                 modelExplorer,
                 For.Name,
                 value: modelExplorer.Model,
                 format: format,
-                htmlAttributes: null);
+                htmlAttributes: htmlAttributes);
         }
 
         // Get a fall-back format based on the metadata.
@@ -353,55 +365,9 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
                 }
             }
 
-            yield return fieldType.Name;
-
-            if (fieldType == typeof(string))
+            foreach (string typeName in TemplateRenderer.GetTypeNames(modelExplorer.Metadata, fieldType))
             {
-                // Nothing more to provide
-                yield break;
-            }
-            else if (!modelExplorer.Metadata.IsComplexType)
-            {
-                // IsEnum is false for the Enum class itself
-                if (fieldType.IsEnum())
-                {
-                    // Same as fieldType.BaseType.Name in this case
-                    yield return "Enum";
-                }
-                else if (fieldType == typeof(DateTimeOffset))
-                {
-                    yield return "DateTime";
-                }
-
-                yield return "String";
-            }
-            else if (fieldType.IsInterface())
-            {
-                if (typeof(IEnumerable).IsAssignableFrom(fieldType))
-                {
-                    yield return "Collection";
-                }
-
-                yield return "Object";
-            }
-            else
-            {
-                var isEnumerable = typeof(IEnumerable).IsAssignableFrom(fieldType);
-                while (true)
-                {
-                    fieldType = fieldType.BaseType();
-                    if (fieldType == null)
-                    {
-                        break;
-                    }
-
-                    if (isEnumerable && fieldType == typeof(Object))
-                    {
-                        yield return "Collection";
-                    }
-
-                    yield return fieldType.Name;
-                }
+                yield return typeName;
             }
         }
     }
--- a/test/Microsoft.AspNet.Mvc.TagHelpers.Test/InputTagHelperTest.cs
+++ b/test/Microsoft.AspNet.Mvc.TagHelpers.Test/InputTagHelperTest.cs
@@ -5,8 +5,10 @@ using System;
 using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
 using System.Threading.Tasks;
+using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Mvc.ModelBinding;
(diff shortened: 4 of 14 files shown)


src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultEditorTemplates.cs
```diff
@@ -374,6 +374,22 @@ public static string NumberInputTemplate(IHtmlHelper htmlHelper)
             return GenerateTextBox(htmlHelper, inputType: "number");
         }
 
+        public static string FileInputTemplate(IHtmlHelper htmlHelper)
```
NTaylorMullen: Not sure why none of the methods in this class have `[NotNull]` but we should probably add them for each of the methods you added.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultEditorTemplates.cs
```diff
@@ -374,6 +374,22 @@ public static string NumberInputTemplate(IHtmlHelper htmlHelper)
...
+        {
+            return GenerateTextBox(htmlHelper, inputType: "file");
+        }
+
+        public static string FileCollectionInputTemplate(IHtmlHelper htmlHelper)
+        {
+            var htmlAttributes = new Dictionary<string, object>()
```
NTaylorMullen: nit: no need for `()`

src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
```diff
@@ -194,6 +197,11 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
 
             if (tagBuilder != null)
             {
+                if (inputType == "file" && inputTypeHint == nameof(IFormFileCollection))
```
NTaylorMullen: `string.Equals(inputType....)`

NTaylorMullen: Why doesn't the `GenerateTextBox` method take care of this?

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultEditorTemplates.cs
```diff
@@ -374,6 +374,22 @@ public static string NumberInputTemplate(IHtmlHelper htmlHelper)
...
+            {
+                { "multiple", "" }
+            };
+            htmlHelper.ViewData[HtmlAttributeKey] = htmlAttributes;
+
+            return FileInputTemplate(htmlHelper);
```
dougbu: need to use a different helper / approach, not `GenerateTextBox`.  as-is you're overwriting whatever was passed in the `@Html.Editor(... additionalViewData: X)` argument and using that to get additional information down to `GenerateTextBox`.

suggest instead
``` c#
  var htmlAttributes =
                CreateHtmlAttributes(htmlHelper, className: "text-box single-line", inputType: "file");
  htmlAttributes["multiple"] = "multiple";
  return htmlHelper.TextBox(
                current: null,
                value: value,
                format: null,
                htmlAttributes: htmlAttributes)
                    .ToString();
```

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultEditorTemplates.cs
```diff
@@ -374,6 +374,22 @@ public static string NumberInputTemplate(IHtmlHelper htmlHelper)
...
+        public static string FileInputTemplate(IHtmlHelper htmlHelper)
+        {
+            return GenerateTextBox(htmlHelper, inputType: "file");
+        }
+
+        public static string FileCollectionInputTemplate(IHtmlHelper htmlHelper)
+        {
+            var htmlAttributes = new Dictionary<string, object>()
+            {
+                { "multiple", "" }
```
dougbu: if this remains (not sure it's needed w/ change below), value must be "multiple" and not the empty string.  we generate XHTML-compliant HTML.

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultEditorTemplates.cs
NTaylorMullen: Sorry didn't see that it was `Design`

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/TemplateRenderer.cs
```diff
@@ -168,6 +171,14 @@ private IEnumerable<string> GetViewNames()
...
                 yield return "String";
             }
+            else if (typeof(IFormFile).IsAssignableFrom(fieldType))
+            {
+                yield return nameof(IFormFile);
+            }
+            else if (typeof(IEnumerable<IFormFile>).IsAssignableFrom(fieldType))
+            {
+                yield return nameof(IFormFileCollection);
+            }
```
dougbu: this gives the interfaces too high a priority.  fine for the `fieldType.IsInterface()` case but should for example yield `FormFile` before `IFormFile` if the property is of that type or a subtype.  need to add these checks immediately before where this method yields "Collection" and "Object".  for example,
``` c#
            else if (fieldType.IsInterface())
            {
                if (typeof(IEnumerable).IsAssignableFrom(fieldType))
                {
                    if (typeof(IEnumerable<IFormFile>).IsAssignableFrom(fieldType))
                    {
                        yield return nameof(IFormFileCollection);
                    }

                    yield return "Collection";
                }
            ...
```

will likely lead to a few duplicate lines but that's better than not returning the expected types.

src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
```diff
@@ -194,6 +197,11 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
...
+                {
+                    tagBuilder.Attributes.Add("multiple", "");
```
dougbu: value must be "multiple"

src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
```diff
@@ -375,6 +383,14 @@ private static IEnumerable<string> GetInputTypeHints(ModelExplorer modelExplorer
...
                 yield return "String";
             }
+            else if (typeof(IFormFile).IsAssignableFrom(fieldType))
+            {
+                yield return nameof(IFormFile);
+            }
+            else if (typeof(IEnumerable<IFormFile>).IsAssignableFrom(fieldType))
+            {
+                yield return nameof(IFormFileCollection);
+            }
```
dougbu: same issue w/ precedence here

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/TemplateRenderer.cs
dougbu: also, this method _always_ yields "Collection", "String", or "Object" even if more-specific special cases exist in `_defaultEditorActions`.  that's required not just because `_defaultDisplayActions` contains many fewer keys (e.g. because depending on such information makes this method brittle).  don't break that invariant.

dougbu:
:watch:

ajaybhargavb:
Updated.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/TemplateRenderer.cs
```diff
@@ -172,14 +175,31 @@ private IEnumerable<string> GetViewNames()
             {
                 if (typeof(IEnumerable).IsAssignableFrom(fieldType))
                 {
+                    if (typeof(IEnumerable<IFormFile>).IsAssignableFrom(fieldType))
```
NTaylorMullen: You repeat this logic twice for `IFormFile`/`IFormFileCollection`. In both cases they take precedence over  the interfaces/class types. Can you instead place them as their own else if statements before the `else if (fieldType.IsInterface())`? That way the logic is only ever done once and essentially does the same thing.

ajaybhargavb: You mean do this? https://github.com/aspnet/Mvc/commit/4558a333dca48212d685d07c7469fd442ab8d4a2#diff-2f73dc4b7cc96312c3a65b0c24251c9dR178

I had this in my previous commit. I changed it after discussing with @dougbu regarding this comment https://github.com/aspnet/Mvc/pull/2366#discussion_r28095216

NTaylorMullen: Ignore this.

ajaybhargavb:
Updated.

src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
```diff
@@ -271,13 +274,22 @@ private TagBuilder GenerateTextBox(ModelExplorer modelExplorer, string inputType
                 format = GetFormat(modelExplorer, inputTypeHint, inputType);
             }
 
+            object htmlAttributes = null;
+            if (inputType == "file" && inputTypeHint == nameof(IFormFileCollection))
```
NTaylorMullen: `string.Equals` for these guys.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/TemplateRenderer.cs
```diff
@@ -167,36 +170,38 @@ private IEnumerable<string> GetViewNames()
...
             {
-                var isEnumerable = typeof(IEnumerable).IsAssignableFrom(fieldType);
-
+                var type = fieldType;
                 while (true)
                 {
-                    fieldType = fieldType.BaseType();
-                    if (fieldType == null)
+                    type = type.BaseType();
+                    if (type == null || type == typeof(Object))
```
NTaylorMullen: `Object` => `object` (since you're touching it)

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/TemplateRenderer.cs
```diff
@@ -167,36 +170,38 @@ private IEnumerable<string> GetViewNames()
...
+                    yield return type.Name;
+                }
+            }
 
-                    yield return fieldType.Name;
+            if (typeof(IEnumerable).IsAssignableFrom(fieldType))
+            {
+                if (typeof(IEnumerable<IFormFile>).IsAssignableFrom(fieldType))
+                {
+                    yield return nameof(IFormFileCollection);
```
dougbu: odd for this to return `nameof(IFormFileCollection)` and not `typeof(IEnumerable<IFormFile>).Name`.  let's not introduce an inconsistency that needs to be explained.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/TemplateRenderer.cs
```diff
@@ -167,36 +170,38 @@ private IEnumerable<string> GetViewNames()
...
                 }
+
+                yield return "Collection";
+            }
+            else if (typeof(IFormFile).IsAssignableFrom(fieldType))
+            {
+                yield return nameof(IFormFile);
```
dougbu: if the property is of type `IFormFile`, this will lead to a duplicate entry in the returned collection.  think the only duplication today occurs when the property is of type `object`.  admittedly at the code of complicating the conditions (e.g. `typeof(IFormFile) != fieldType && typeof(IFormFile).IsAssignableFrom(fieldType)`), worth avoiding the duplicate here too.  see for example the special case for `string`, precisely to avoid a duplicate entry.

after fixing the return string for `IEnumerable<IFormFile>`, will also need to deal with potential duplication when `typeof(IEnumerable<IFormFile>) == fieldType`

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/TemplateRenderer.cs
```diff
@@ -167,36 +170,38 @@ private IEnumerable<string> GetViewNames()
                 }
```
dougbu: a couple of lines earlier (https://github.com/aspnet/Mvc/blob/file-tag-helper/src/Microsoft.AspNet.Mvc.Core/Rendering/Html/TemplateRenderer.cs#L152) to the end of the method are identical to the code in `InputTagHelper.GetInputTypeHints()`.  suggest adding a `public static GetTypeNames(Type type)` method in this class, avoiding the duplication.

admit that refactoring could have happened earlier but you're rearranging the code now...

src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
```diff
@@ -374,35 +386,38 @@ private static IEnumerable<string> GetInputTypeHints(ModelExplorer modelExplorer
                 }
```
dougbu: as mentioned in `TemplateRenderer`, suggest removing this code and using a shared helper.  (when I made my offline point about keeping the two methods' structure aligned, I didn't realize how much was identical)

dougbu:
:watch: at a high level this looks good.  getting close to time for do the work for realz.

ajaybhargavb:
Updated with tests. @dougbu @NTaylorMullen

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/TemplateRenderer.cs
```diff
@@ -7,16 +7,19 @@
...
 
-namespace Microsoft.AspNet.Mvc.Rendering
+namespace Microsoft.AspNet.Mvc.Rendering.Internal
 {
-    internal class TemplateRenderer
+    public class TemplateRenderer
     {
         private static readonly string DisplayTemplateViewPath = "DisplayTemplates";
         private static readonly string EditorTemplateViewPath = "EditorTemplates";
+        public const string IEnumerableOfIFormFileName = "IEnumerable`IFormFile`";
```
NTaylorMullen: `public static readonly string IEnumerableOfIFormFileName = typeof(IEnumerable<IFormFile>).Name;`

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/TemplateRenderer.cs
```diff
@@ -141,19 +146,31 @@ private IEnumerable<string> GetViewNames()
...
+            var fieldType = _viewData.ModelExplorer.ModelType;
+
             // We don't want to search for Nullable<T>, we want to search for T (which should handle both T and
             // Nullable<T>).
-            var modelType = _viewData.ModelExplorer.ModelType;
-            var fieldType = Nullable.GetUnderlyingType(modelType) ?? modelType;
+            fieldType = Nullable.GetUnderlyingType(fieldType) ?? fieldType;
 
-            yield return fieldType.Name;
+            foreach (string typeName in GetTypeNames(_viewData.ModelExplorer, fieldType))
```
NTaylorMullen: `var`

NTaylorMullen:
Gonna have to wait on @dougbu for sign off on this one, that type hint resolution loop is rough to follow.

I've added as much as I could to the PR, consider me signed off.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/TemplateRenderer.cs
Eilon: Could at least hard-code the `"IEnumerable"` part - that type name will never change :smile: And at that point seriously just use "+" instead of crazy string interpolation.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/TemplateBuilder.cs
```diff
@@ -5,9 +5,9 @@
 using Microsoft.AspNet.Mvc.ModelBinding;
 using Microsoft.Framework.Internal;
 
-namespace Microsoft.AspNet.Mvc.Rendering
+namespace Microsoft.AspNet.Mvc.Rendering.Internal
```
dougbu: looks like this project contains a few `internal` namespaces already and they follow the pattern `Microsoft.AspNet.Mvc.Internal.XYZ`.  so these should instead be in `Microsoft.AspNet.Mvc.Internal.Rendering`.

files should also be moved to the `src/Microsoft.AspNet.Mvc.Core/Internal/Rendering` folder.

Eilon: Eh, no, I don't think so. The common (and logical) pattern is to add a `.Internal` namespace in the namespace you're in. Otherwise you end up with a super weird hierarchy.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/TemplateRenderer.cs
dougbu: :+1: on the `static readonly` part.

but this mostly an arbitrary string and I'm not sure it matters here whether `IFormFile` is renamed in the future.  as-is or use `nameof`, either is fine.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/TemplateRenderer.cs
```diff
@@ -141,19 +146,31 @@ private IEnumerable<string> GetViewNames()
                 yield return templateHint;
             }
 
+            var fieldType = _viewData.ModelExplorer.ModelType;
+
             // We don't want to search for Nullable<T>, we want to search for T (which should handle both T and
             // Nullable<T>).
-            var modelType = _viewData.ModelExplorer.ModelType;
-            var fieldType = Nullable.GetUnderlyingType(modelType) ?? modelType;
+            fieldType = Nullable.GetUnderlyingType(fieldType) ?? fieldType;
```
dougbu: why did the block from line 149 to here change?  not worth the churn

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/TemplateRenderer.cs
```diff
@@ -141,19 +146,31 @@ private IEnumerable<string> GetViewNames()
...
 
-            yield return fieldType.Name;
+            foreach (string typeName in GetTypeNames(_viewData.ModelExplorer, fieldType))
+            {
+                yield return typeName;
+            }
+        }
+
+        public static IEnumerable<string> GetTypeNames(ModelExplorer modelExplorer, Type fieldType)
```
dougbu: pass `ModelMetadata metadata`, not the whole `ModelExplorer`.  only need the metadata.

src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
```diff
@@ -346,62 +358,12 @@ private static IEnumerable<string> GetInputTypeHints(ModelExplorer modelExplorer
             var fieldType = modelExplorer.ModelType;
             if (typeof(bool?) != fieldType)
             {
-                var underlyingType = Nullable.GetUnderlyingType(fieldType);
-                if (underlyingType != null)
-                {
-                    fieldType = underlyingType;
-                }
-            }
```
dougbu: again, leave this block alone.  readable and does what it says.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/InputTagHelperTest.cs
```diff
@@ -679,16 +683,25 @@ public class InputTagHelperTest
...
-                .Setup(mock => mock.GenerateTextBox(
-                    tagHelper.ViewContext,
-                    tagHelper.For.ModelExplorer,
-                    tagHelper.For.Name,
-                    null,                                   // value
-                    expectedFormat,
-                    null))                                  // htmlAttributes
-                .Returns(tagBuilder)
-                .Verifiable();
+            .Setup(mock => mock.GenerateTextBox(
```
dougbu: nit: this should have remained indented

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/TemplateRenderer.cs
dougbu: thx

src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
dougbu: thx

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/TemplateRenderer.cs
dougbu: @Eilon I think you just lost some @pranavkm Points :sob:

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/TemplateBuilder.cs
dougbu: :ok_hand: then we have some work to do cleaning up the `src/Microsoft.AspNet.Mvc.Core/Internal` folder

in any case @ajaybhargavb the files in the new `Microsoft.AspNet.Mvc.Rendering.Internal` namespace should move to another folder

dougbu:
:watch: very close

dougbu:
please add a few tests of the new `GetTypeNames()` method.  coverage of `Display()` and `Editor()` helpers isn't great and we definitely lack anything like old tests of `GetViewNames()` in http://aspnetwebstack.codeplex.com/SourceControl/latest#test/System.Web.Mvc.Test/Html/Test/TemplateHelpersTest.cs  my suggestion is to use those tests as inspiration, not to copy everything.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/TemplateRenderer.cs
Eilon: True story: There is a conversion between @pranavkm points and @Eilon points but I won't say what the conversion rate is.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/TemplateBuilder.cs
ajaybhargavb: Moved.

ajaybhargavb:
Updated @dougbu

src/Microsoft.AspNet.Mvc.Core/Rendering/Internal/TemplateRenderer.cs
```diff
@@ -7,16 +7,19 @@
...
 
-namespace Microsoft.AspNet.Mvc.Rendering
+namespace Microsoft.AspNet.Mvc.Rendering.Internal
 {
-    internal class TemplateRenderer
+    public class TemplateRenderer
     {
         private static readonly string DisplayTemplateViewPath = "DisplayTemplates";
         private static readonly string EditorTemplateViewPath = "EditorTemplates";
+        public static readonly string IEnumerableOfIFormFileName = "IEnumerable`" + nameof(IFormFile);
```
Eilon: Hmm weird why are these all `static readonly` instead of `const`? I checked and `const` works with `nameof` (because it's resolved at compile time).

ajaybhargavb: I am confused. I changed this from `const` based on the feedback here. https://github.com/aspnet/Mvc/pull/2366#discussion_r28301603

Eilon: Oh I think that might be because it was erroneously assumed that this is public, whereas it really isn't. This member is effectively internal (it used to be C# internal but now it's in the Internal namespace) so we can use const's all we want.

@NTaylorMullen @dougbu did I miss anything here on the reasoning?

dougbu: not from my perspective.  I was thinking `public` not `public with special circumstances`.

src/Microsoft.AspNet.Mvc.Core/Rendering/Internal/TemplateRenderer.cs
```diff
@@ -136,24 +141,38 @@ private IEnumerable<string> GetViewNames()
                 metadata.DataTypeName
             };
 
-            foreach (string templateHint in templateHints.Where(s => !string.IsNullOrEmpty(s)))
+            foreach (var templateHint in templateHints.Where(s => !string.IsNullOrEmpty(s)))
             {
                 yield return templateHint;
             }
 
+
```
dougbu: don't add extra blank line

test/Microsoft.AspNet.Mvc.TagHelpers.Test/InputTagHelperTest.cs
```diff
@@ -686,7 +712,7 @@ public class InputTagHelperTest
                     tagHelper.For.Name,
                     null,                                   // value
                     expectedFormat,
-                    null))                                  // htmlAttributes
+                    htmlAttributes ?? null))                // htmlAttributes
```
dougbu: remove `?? null`; same without it except nobody will as "why?"

dougbu:
:sihpit: after addressing @Eilon's comment and my less-important ones

src/Microsoft.AspNet.Mvc.Core/Rendering/Internal/TemplateRenderer.cs
NTaylorMullen: Nope didn't miss anything. Whenever there's a `public` class **I** always assume `static readonly`. For internal namespaced classes, could definitely choose not too. 

I like to go with the approach that's more flexible and leads to less head scratching/searching when re-reading code down the line (as long as its impact is not significant). And heck maybe it's just me that finds it easier/quicker to assume `public` => `public static readonly` instead of `const` but then again here we are in this long convo :smile: 

<- Indifferent for this situation.

PR #2368, opened 2015-04-09
React to HttpResponse.Challenge API changes

HaoK:
https://github.com/aspnet/HttpAbstractions/pull/254

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ActionResults/ChallengeResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ActionResults/ChallengeResult.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Collections.Generic;
+using System.Linq;
 using Microsoft.AspNet.Http.Authentication;
 using Microsoft.Framework.Internal;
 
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNet.Mvc.Core.Test/ActionResults/ChallengeResultTest.cs
```diff
@@ -31,7 +31,7 @@ public void ChallengeResult_Execute()
             result.ExecuteResult(actionContext);
 
             // Assert
-            httpResponse.Verify(c => c.Challenge(null, (IEnumerable<string>)new string[] { }), Times.Exactly(1));
+            httpResponse.Verify(c => c.Challenge(null, ""), Times.Exactly(1));
```
rynowak: We should have a new test to cover the other behavior (no schemes)?

src/Microsoft.AspNet.Mvc.Core/ActionResults/ChallengeResult.cs
```diff
@@ -47,7 +48,17 @@ public ChallengeResult(IList<string> authenticationSchemes, AuthenticationProper
         public override void ExecuteResult([NotNull] ActionContext context)
         {
             var response = context.HttpContext.Response;
-            response.Challenge(Properties, AuthenticationSchemes);
+            if (AuthenticationSchemes.Any())
```
davidfowl: Count > 0

HaoK:
Updated with PR feedback

src/Microsoft.AspNet.Mvc.Core/ActionResults/ChallengeResult.cs
```diff
@@ -47,7 +47,17 @@ public ChallengeResult(IList<string> authenticationSchemes, AuthenticationProper
         public override void ExecuteResult([NotNull] ActionContext context)
         {
             var response = context.HttpContext.Response;
-            response.Challenge(Properties, AuthenticationSchemes);
+            if (AuthenticationSchemes.Count > 0)
```
rynowak: Will this every be null? Might want a `?.` in there #PranavPoints

rynowak:
:shipit:

HaoK:
bee20973c770c0bf6af27f8aa50c45206e89111e

Issue #2280, opened 2015-03-27
JsonPatch: Cannot do an "add" patch to IList<Anything>


NTaylorMullen:
When attempting to patch `People` in the object below our patching mechanism throws.

```C#
public class Users
{
    public IList<string> People { get; } = new List<string>();
}
```

Note this happens for any type of `IList`

rynowak:
@NTaylorMullen please include sample code of the controller/action and the jsonpatch document being sent to the server so we can verify.

@kirthik - there may be an issue with the detection for 'is this an array or a string'...

NTaylorMullen:
```C#
public class HomeController : Controller
{
    public static Users Users = new Users();

    // GET: /<controller>/
    public IActionResult Index()
    {
        return View();
    }

    [HttpPatch]
    public string DoWork([FromBody] JsonPatchDocument<Users> patchDoc)
    {
        patchDoc.ApplyTo(Users);

        return JsonConvert.SerializeObject(Users);
    }
}

public class Users
{
    public IList<string> People { get; set; } = new List<string>();
}
```

```JavaScript
var data = [
    { "op": "add", "path": "People/1", "value": "Taylor"},
];

$.ajax({
    url: 'Home/DoWork',
    data: JSON.stringify(data),
    type: 'PATCH',
    dataType: 'json',
    contentType: 'application/json-patch+json'
});
```

kirthik:
Fix checked in - 3bc7599903dfd8ddc55d9249794efa65b2f527e1

PR #2375, opened 2015-04-13
Fix for #2280 - Cannot do an "add" patch to IList<Anything>

kirthik:
@rynowak - could you please help take a look?


src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -9,6 +9,7 @@
 using Microsoft.AspNet.JsonPatch.Operations;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Serialization;
+using System.Collections.Generic;
```
rynowak: sort

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -638,9 +642,22 @@ public void Copy(Operation<T> operation, T objectToApplyTo)
...
+            if (jsonPropertyType.IsGenericType)
+            {
+                var genericArguments = jsonPropertyType.GetGenericArguments();
+                if (genericArguments.Length != 1)
+                {
+                    return false;
+                }
+                var listType = typeof(IList<>).MakeGenericType(genericArguments);
+                return listType.GetTypeInfo().IsAssignableFrom(jsonPropertyType);
+            }
```
rynowak: The right check here is probably to look at `.ImplementedInterfaces` for a type which is is generic and has `IList<>` as it's generic definition. This code here isn't really accurate.

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -133,9 +134,9 @@ private void Add(string path, object value, T objectToApplyTo, Operation<T> oper
                 // what if it's an array but there's no position??
                 if (IsNonStringArray(patchProperty))
                 {
-                    // now, get the generic type of the enumerable
+                    // now, get the generic type of the enumerable from the runtime type of patchProperty value.
                     var genericTypeOfArray = PropertyHelpers.GetEnumerableType(
-                        patchProperty.Property.PropertyType);
+                        patchProperty.Property.ValueProvider.GetValue(patchProperty.Parent).GetType());
```
rynowak: What specific case is this needed for? What if the value is null.

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -133,9 +134,9 @@ private void Add(string path, object value, T objectToApplyTo, Operation<T> oper
                 // what if it's an array but there's no position??
                 if (IsNonStringArray(patchProperty))
                 {
-                    // now, get the generic type of the enumerable
+                    // now, get the generic type of the enumerable from the runtime type of patchProperty value.
                     var genericTypeOfArray = PropertyHelpers.GetEnumerableType(
```
rynowak: Should we be looking for an `IList<>` type instead? That's really the contract that we need if we're adding/inserting.

test/Microsoft.AspNet.JsonPatch.Test/NestedObjectTests.cs
```diff
@@ -202,6 +202,151 @@ public void AddToListWithSerialization()
         }
 
         [Fact]
+        public void AddToIntListToGeneralList()
```
rynowak: Do all of these tests vary on the declared type of the property? (Didn't you update the code to use the runtime type).

I'd also like to see one where we're modifying a list through a nested property.

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
kirthik: `JsonProperty`'s `PropertyType` returns declared type of the property not the runtime type. So if there is a property of declared type `IList`, this call returns `null`. So I had to change it to take the  runtime type of value.

Its a good point, I have to handle the case where runtime type value is `null`.

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
kirthik: If Property's declared type is `IList<>`, then `.ImplementedInterfaces` would not have `IList<>`. Getting generic type from `IEnumerable` works when Property is of type `'IList` or `IList<>` or `List<>`.

test/Microsoft.AspNet.JsonPatch.Test/NestedObjectTests.cs
kirthik: `JsonProperty`'s `PropertyType` is fetching declared type of the property not the runtime type.

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
kirthik: Forgot to sort. Will do it in the next iteration :)

kirthik:
@rynowak  : Updated code based on comments. Could you please take a look?

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -636,11 +637,14 @@ public void Copy(Operation<T> operation, T objectToApplyTo)
             }
         }
 
-        private bool IsNonStringArray(JsonPatchProperty patchProperty)
+        private bool IsNonStringArray(Type type)
         {
-            return !(patchProperty.Property.PropertyType == typeof(string))
-                    && typeof(IList).GetTypeInfo().IsAssignableFrom(
-                        patchProperty.Property.PropertyType.GetTypeInfo());
+            if (PropertyHelpers.GetIListType(type) != null)
```
rynowak: If this is only used here please just inline the code. We don't want to litter our code with random helpers when a private method does what you want already.

rynowak:
:shipit:

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
kirthik: `GetIListType` is used in multiple places in this class. I just updated this code to re-use existing method.

rynowak: Is it used in multiple classes?

kirthik: No. Just in this class.

kirthik:
Checked in - 3bc7599903dfd8ddc55d9249794efa65b2f527e1

PR #2376, opened 2015-04-13
Add rendering logic for PreElement and PostElement on TagHelperOutput.

NTaylorMullen:
- Added unit tests to validate that the properties were rendered correctly.
- Modified functional tests to utilize PreElement and PostElement.

aspnet/Razor#341


Eilon:
:shipit:

pranavkm:
:shipit:

Issue #1989, opened 2015-02-10
Consider delaying attribute route initialization until first request


rynowak:
Right now attribute route initialization (and error reporting) happens during startup. We considered deferring this until the first request reaches routing (and this is easy to do), but it has an impact on when logging for action discovery takes place (within a request instead of startup).

We should consider whether or not we want to make this change with the logging impact.

rynowak:
d0e51187415fedf666e2f971f6d08cb74a57abae

PR #2377, opened 2015-04-13
Delay Attribute Route initialization to the first request

rynowak:
Fix for #1989

### DIFF
--- a/test/Microsoft.AspNet.Mvc.FunctionalTests/Logging/LoggingStartupTest.cs
+++ b/test/Microsoft.AspNet.Mvc.FunctionalTests/Logging/LoggingStartupTest.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net;
 using System.Threading.Tasks;
 using LoggingWebSite;
 using LoggingWebSite.Controllers;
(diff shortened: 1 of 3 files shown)


test/Microsoft.AspNet.Mvc.FunctionalTests/Logging/LoggingStartupTest.cs
```diff
@@ -92,13 +93,20 @@ public class LoggingStartupTest
...
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
 
-            var logs = activityDtos.FilterByStartup().GetLogsByDataType<T>();
+            response = await client.GetAsync("http://localhost/logs");
 
-            return logs;
+            var body = await response.Content.ReadAsStringAsync();
+            var activityDtos = JsonConvert.DeserializeObject<List<ActivityContextDto>>(body);
+
+            return activityDtos.GetLogsByDataType<T>();
```
rynowak: This change is necessary because this logging now happens on first request, not on running startup.

kichalla:
:shipit: (wondering if this has any affect on apiexplorer being able to get information about attribute routes?)

rynowak:
Nope. ApiExplorer uses the `IActionDescriptorCollectionProvider` to get it's data (as does attribute routing). `IActionDescriptorCollectionProvider` has always been lazy, this just takes advantage of that.

kichalla:
:+1: thanks...

PR #2379, opened 2015-04-14
Logging for content-negotiation

kichalla:
@rynowak @Eilon 

This is a design PR as I would like to add more statements mainly to answer this common question: `I made a request with accept header xml, but why did i get a json response back?` and feel that the current statements would not give much useful information regarding this.

Can we log details of these too?(or would these look more like tracing?)
- User explicitly set a content-type (ex: through Produces or Url path extension mapping) and con-neg is not run.
- No formatter was found using accept header and so finding by request content-type header etc.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ActionResults/ObjectResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ActionResults/ObjectResult.cs
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
@@ -11,6 +10,7 @@ using Microsoft.AspNet.Mvc.Core;
 using Microsoft.AspNet.WebUtilities;
 using Microsoft.Framework.DependencyInjection;
 using Microsoft.Framework.Internal;
+using Microsoft.Framework.Logging;
 using Microsoft.Framework.OptionsModel;
 using Microsoft.Net.Http.Headers;
 
@@ -40,9 +40,12 @@ namespace Microsoft.AspNet.Mvc
 
         public override async Task ExecuteResultAsync(ActionContext context)
         {
+            var logger = context.HttpContext.RequestServices.GetRequiredService<ILogger<ObjectResult>>();
+                            
             // See if the list of content types added to this object result is valid.
             ThrowIfUnsupportedContentType();
             var formatters = GetDefaultFormatters(context);
+
             var formatterContext = new OutputFormatterContext()
             {
                 DeclaredType = DeclaredType,
@@ -55,9 +58,17 @@ namespace Microsoft.AspNet.Mvc
             if (selectedFormatter == null)
             {
                 // No formatter supports this.
+                logger.LogWarning("No output formatter was found to write the response.");
+
                 context.HttpContext.Response.StatusCode = StatusCodes.Status406NotAcceptable;
                 return;
             }
+
+            logger.LogVerbose(
+                "Selected output formatter '{OutputFormatter}'and content type " +
+                "'{ContentType}' to write the response.", 
+                selectedFormatter.GetType().FullName,
+                formatterContext.SelectedContentType);
 
             if (StatusCode.HasValue)
             {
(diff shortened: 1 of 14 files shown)


Eilon:
These logs do seem to say *what*, but not *why*. Would this help someone understand *why* JSON was selected instead of XML?

kichalla:
Updated: Added more log statements. /cc: @rynowak @Eilon

src/Microsoft.AspNet.Mvc.Core/ActionResults/ObjectResult.cs
```diff
@@ -55,10 +63,16 @@ public ObjectResult(object value)
             if (selectedFormatter == null)
             {
                 // No formatter supports this.
+                logger.LogWarning("No output formatter was found to write the response.");
+
                 context.HttpContext.Response.StatusCode = StatusCodes.Status406NotAcceptable;
                 return;
             }
 
+            logger.LogInformation(
```
Eilon: This is probably also verbose.

src/Microsoft.AspNet.Mvc.Core/ActionResults/ObjectResult.cs
```diff
@@ -72,10 +86,18 @@ public ObjectResult(object value)
...
+            var logger = formatterContext.ActionContext.HttpContext.RequestServices
+                .GetRequiredService<ILoggerFactory>()
+                .CreateLogger<ObjectResult>();
+
             // Check if any content-type was explicitly set (for example, via ProducesAttribute 
             // or Url path extension mapping). If yes, then ignore content-negotiation and use this content-type.
             if (ContentTypes.Count == 1)
             {
+                logger.LogVerbose(
+                    "Skipped content-negotiation as content type '{ContentType}' is explicitly set for the response.",
```
Eilon: I don't think `content-negotiation` should by hyphenated.

Eilon:
Looks much better - I think it now has information that would help a developer figure out why something isn't working.

src/Microsoft.AspNet.Mvc.Core/ActionResults/ObjectResult.cs
rynowak: :+1:

Think of it this way. We want logging (at information level) to produce probably 5-6 log statements for a typical request. Assume that MVC's budget is 1 per-request - and we're already using that for the action that was selected. 

This means that unless explicitly approved, everything that's a success case should be verbose by default.

src/Microsoft.AspNet.Mvc.Core/ActionResults/ObjectResult.cs
```diff
@@ -40,9 +40,17 @@ public ObjectResult(object value)
 
         public override async Task ExecuteResultAsync(ActionContext context)
         {
+            var logger = context.HttpContext.RequestServices.GetRequiredService<ILoggerFactory>()
+                .CreateLogger<ObjectResult>();
```
rynowak: You should also be able to do:
```
context.HttpContext.RequestServices.GetRequiredService<ILogger<ObjectResult>>();
```

src/Microsoft.AspNet.Mvc.Core/ActionResults/ObjectResult.cs
```diff
@@ -40,9 +40,17 @@ public ObjectResult(object value)
...
+                            
             // See if the list of content types added to this object result is valid.
             ThrowIfUnsupportedContentType();
             var formatters = GetDefaultFormatters(context);
+
+            logger.LogVerbose(
+                "Available output formatters for writing the response: {OutputFormatters}",
+                formatters.Select(f => f.GetType().FullName));
```
rynowak: I wouldn't include this, it seems like overkill. This is either coming from configuration and is the same for everything, or it's something the user put in their code.

src/Microsoft.AspNet.Mvc.Core/ActionResults/ObjectResult.cs
rynowak: Can we include the content-type as well here?

src/Microsoft.AspNet.Mvc.Core/ActionResults/ObjectResult.cs
rynowak: Should we bother with this? It's pretty much just telling you what code you wrote.

src/Microsoft.AspNet.Mvc.Core/ActionResults/ObjectResult.cs
kichalla: :+1:

kichalla:
updated. /cc: @rynowak @Eilon

src/Microsoft.AspNet.Mvc.Core/ActionResults/ObjectResult.cs
```diff
@@ -55,10 +58,18 @@ public ObjectResult(object value)
...
             {
                 // No formatter supports this.
+                logger.LogWarning("No output formatter was found to write the response.");
+
                 context.HttpContext.Response.StatusCode = StatusCodes.Status406NotAcceptable;
                 return;
             }
 
+            logger.LogVerbose(
+                "Selected output formatter '{OutputFormatter}'and content type " +
```
Eilon: Missing space after `'`.

Eilon:
:shipit: after the one comment I had.

rynowak:
:shipit:

PR #2384, opened 2015-04-14
Log view discovery

kichalla:
@rynowak @pranavkm @dougbu @Eilon 

Please note that I haven't made any test related changes yet as wanted to get some initial feedback (mainly for changes where the logger factory is just being 'passed on')

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ActionResults/PartialViewResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ActionResults/PartialViewResult.cs
@@ -6,6 +6,7 @@ using System.Threading.Tasks;
 using Microsoft.AspNet.Mvc.Rendering;
 using Microsoft.Framework.DependencyInjection;
 using Microsoft.Framework.Internal;
+using Microsoft.Framework.Logging;
 
 namespace Microsoft.AspNet.Mvc
 {
@@ -50,10 +51,21 @@ namespace Microsoft.AspNet.Mvc
             var viewEngine = ViewEngine ??
                              context.HttpContext.RequestServices.GetRequiredService<ICompositeViewEngine>();
 
+            var logger = context.HttpContext.RequestServices.GetRequiredService<ILogger<PartialViewResult>>();
+
             var viewName = ViewName ?? context.ActionDescriptor.Name;
-            var view = viewEngine.FindPartialView(context, viewName)
-                                 .EnsureSuccessful()
-                                 .View;
+            var viewEngineResult = viewEngine.FindPartialView(context, viewName);
+            if (!viewEngineResult.Success)
+            {
+                logger.LogError(
+                    "The partial view '{PartialViewName}' was not found. Searched locations: {SearchedViewLocations}",
+                    viewName,
+                    viewEngineResult.SearchedLocations);
+            }
+
+            var view = viewEngineResult.EnsureSuccessful().View;
+
+            logger.LogVerbose("The partial view '{PartialViewName}' was found.", viewName);
 
             if (StatusCode != null)
             {
--- a/src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewResult.cs
@@ -6,6 +6,7 @@ using System.Threading.Tasks;
 using Microsoft.AspNet.Mvc.Rendering;
 using Microsoft.Framework.DependencyInjection;
 using Microsoft.Framework.Internal;
+using Microsoft.Framework.Logging;
 
 namespace Microsoft.AspNet.Mvc
 {
@@ -50,10 +51,21 @@ namespace Microsoft.AspNet.Mvc
             var viewEngine = ViewEngine ??
                              context.HttpContext.RequestServices.GetRequiredService<ICompositeViewEngine>();
 
+            var logger = context.HttpContext.RequestServices.GetRequiredService<ILogger<ViewResult>>();
+
             var viewName = ViewName ?? context.ActionDescriptor.Name;
-            var view = viewEngine.FindView(context, viewName)
-                                 .EnsureSuccessful()
-                                 .View;
+            var viewEngineResult = viewEngine.FindView(context, viewName);
+            if(!viewEngineResult.Success)
+            {
+                logger.LogError(
+                    "The view '{ViewName}' was not found. Searched locations: {SearchedViewLocations}", 
+                    viewName,
+                    viewEngineResult.SearchedLocations);
+            }
+
+            var view = viewEngineResult.EnsureSuccessful().View;
+
+            logger.LogVerbose("The view '{ViewName}' was found.", viewName);
 
             if (StatusCode != null)
             {
--- a/test/Microsoft.AspNet.Mvc.Core.Test/ActionResults/PartialViewResultTest.cs
+++ b/test/Microsoft.AspNet.Mvc.Core.Test/ActionResults/PartialViewResultTest.cs
@@ -6,6 +6,7 @@ using System.Threading.Tasks;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Mvc.Rendering;
 using Microsoft.AspNet.Routing;
+using Microsoft.Framework.Logging;
 using Moq;
 using Xunit;
 
@@ -21,7 +22,7 @@ namespace Microsoft.AspNet.Mvc
                                        "The view 'MyView' was not found. The following locations were searched:",
                                        "Location1",
                                        "Location2.");
-            var actionContext = new ActionContext(new DefaultHttpContext(),
+            var actionContext = new ActionContext(GetHttpContext(),
                                                   new RouteData(),
                                                   new ActionDescriptor());
             var viewEngine = new Mock<IViewEngine>();
@@ -47,7 +48,7 @@ namespace Microsoft.AspNet.Mvc
         {
             // Arrange
             var viewName = "myview";
-            var context = new ActionContext(new DefaultHttpContext(), new RouteData(), new ActionDescriptor());
+            var context = new ActionContext(GetHttpContext(), new RouteData(), new ActionDescriptor());
             var viewEngine = new Mock<IViewEngine>();
             var view = Mock.Of<IView>();
 
@@ -73,7 +74,7 @@ namespace Microsoft.AspNet.Mvc
         {
             // Arrange
             var viewName = "some-view-name";
-            var context = new ActionContext(new DefaultHttpContext(),
+            var context = new ActionContext(GetHttpContext(),
                                             new RouteData(),
                                             new ActionDescriptor { Name = viewName });
             var viewEngine = new Mock<ICompositeViewEngine>();
@@ -109,6 +110,8 @@ namespace Microsoft.AspNet.Mvc
             var serviceProvider = new Mock<IServiceProvider>();
             serviceProvider.Setup(p => p.GetService(typeof(ICompositeViewEngine)))
                            .Returns(viewEngine.Object);
+            serviceProvider.Setup(p => p.GetService(typeof(ILogger<PartialViewResult>)))
+                           .Returns(new Mock<ILogger<PartialViewResult>>().Object);
             context.HttpContext.RequestServices = serviceProvider.Object;
 
             var viewResult = new PartialViewResult
@@ -121,6 +124,18 @@ namespace Microsoft.AspNet.Mvc
 
             // Assert
             viewEngine.Verify();
+        }
+
+        private HttpContext GetHttpContext()
+        {
+            var serviceProvider = new Mock<IServiceProvider>();
+            serviceProvider.Setup(s => s.GetService(typeof(ILogger<PartialViewResult>)))
+                .Returns(new Mock<ILogger<PartialViewResult>>().Object);
+
+            var httpContext = new DefaultHttpContext();
+            httpContext.RequestServices = serviceProvider.Object;
+
+            return httpContext;
         }
     }
 }
--- a/test/Microsoft.AspNet.Mvc.Core.Test/ActionResults/ViewResultTest.cs
+++ b/test/Microsoft.AspNet.Mvc.Core.Test/ActionResults/ViewResultTest.cs
@@ -6,6 +6,7 @@ using System.Threading.Tasks;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Mvc.Rendering;
 using Microsoft.AspNet.Routing;
+using Microsoft.Framework.Logging;
 using Moq;
 using Xunit;
 
@@ -21,7 +22,8 @@ namespace Microsoft.AspNet.Mvc
                                        "The view 'MyView' was not found. The following locations were searched:",
                                        "Location1",
                                        "Location2.");
-            var actionContext = new ActionContext(new DefaultHttpContext(),
+            
+            var actionContext = new ActionContext(GetHttpContext(),
                                                   new RouteData(),
                                                   new ActionDescriptor());
             var viewEngine = new Mock<IViewEngine>();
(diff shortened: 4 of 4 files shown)


src/Microsoft.AspNet.Mvc.Razor/RazorViewFactory.cs
```diff
@@ -21,11 +23,14 @@ public class RazorViewFactory : IRazorViewFactory
         /// <param name="pageActivator">The <see cref="IRazorPageActivator"/> used to activate pages.</param>
         /// <param name="viewStartProvider">The <see cref="IViewStartProvider"/> used for discovery of _ViewStart
         /// pages</param>
+        /// <param name="loggerFactory">The <see cref="ILoggerFactory"/> to log data.</param>
         public RazorViewFactory(IRazorPageActivator pageActivator,
-                                IViewStartProvider viewStartProvider)
+                                IViewStartProvider viewStartProvider,
+                                ILoggerFactory loggerFactory)
```
kichalla: all these should be decorated with `[NotNull]`?

src/Microsoft.AspNet.Mvc.Core/ActionResults/PartialViewResult.cs
```diff
@@ -50,10 +51,22 @@ public class PartialViewResult : ActionResult
...
+                .CreateLogger<PartialViewResult>();
+
             var viewName = ViewName ?? context.ActionDescriptor.Name;
-            var view = viewEngine.FindPartialView(context, viewName)
-                                 .EnsureSuccessful()
-                                 .View;
+            var viewEngineResult = viewEngine.FindPartialView(context, viewName);
+            if (!viewEngineResult.Success)
+            {
+                logger.LogError(
```
pranavkm: Does logging expand `IEnumerable`s?

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/HtmlHelper.cs
```diff
@@ -32,7 +33,8 @@ public class HtmlHelper : IHtmlHelper, ICanHasViewContext
         private readonly ICompositeViewEngine _viewEngine;
 
         private ViewContext _viewContext;
-
+        private readonly ILogger _logger;
```
pranavkm: List this under `private readonly ICompositeViewEngine _viewEngine;`

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/HtmlHelper.cs
```diff
@@ -50,6 +53,7 @@ public class HtmlHelper : IHtmlHelper, ICanHasViewContext
             HtmlEncoder = htmlEncoder;
             UrlEncoder = urlEncoder;
             JavaScriptStringEncoder = javaScriptStringEncoder;
+            _logger = loggerFactory.CreateLogger<HtmlHelper>();
```
pranavkm: `loggerFactory.CreateLogger(GetType());` if you want to support the derived type scenarios we spoke about.

src/Microsoft.AspNet.Mvc.Core/ActionResults/PartialViewResult.cs
rynowak: Should we just put this logging in the ViewEngine?

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/HtmlHelper.cs
rynowak: Please don't start inventing new patterns, Logging has had a ton of discussion and this isn't something we agreed on doing.

src/Microsoft.AspNet.Mvc.Core/ViewComponents/DefaultViewComponentHelper.cs
```diff
@@ -17,15 +17,18 @@ public class DefaultViewComponentHelper : IViewComponentHelper, ICanHasViewConte
...
             [NotNull] IViewComponentDescriptorCollectionProvider descriptorProvider,
             [NotNull] IViewComponentSelector selector,
-            [NotNull] IViewComponentInvokerFactory invokerFactory)
+            [NotNull] IViewComponentInvokerFactory invokerFactory,
+            [NotNull] ILoggerFactory loggerFactory)
         {
             _descriptorProvider = descriptorProvider;
             _selector = selector;
             _invokerFactory = invokerFactory;
+            _logger = loggerFactory.CreateLogger<DefaultViewComponentHelper>();
```
rynowak: You can also just add `ILogger<DefaultViewComponentHelper>` as a constructor parameter instead of taking the factory

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/TemplateRenderer.cs
```diff
@@ -82,6 +84,8 @@ internal class TemplateRenderer
             _viewData = viewData;
             _templateName = templateName;
             _readOnly = readOnly;
+            _logger = _viewContext.HttpContext.RequestServices.GetRequiredService<ILoggerFactory>()
+                .CreateLogger<TemplateRenderer>();
```
rynowak: I'm not sure we should log inside the TemplateRender-er. All of the places we're doing logger elsewhere in the PR are where the user directly asked for a view.

If we really want to log **all** view discovery, then we can put that in the view engine.

src/Microsoft.AspNet.Mvc.Core/ViewComponents/DefaultViewComponentHelper.cs
```diff
@@ -119,10 +128,18 @@ private ViewComponentDescriptor SelectComponent(Type componentType)
...
+                        componentType.FullName);
+
                     return descriptor;
                 }
             }
 
+            _logger.LogError(
+                "The view component of type '{ViewComponentTypeName}' was not found.", 
+                componentType.FullName);
+
```
rynowak: I'm not convinced that this logging adds value. Can we split this out from the view-search logging (which we absolutely need)?

src/Microsoft.AspNet.Mvc.Core/ViewComponents/DefaultViewComponentInvoker.cs
```diff
@@ -34,6 +38,12 @@ public void Invoke([NotNull] ViewComponentContext context)
                 context.Arguments);
             if (method == null)
             {
+                _logger.LogError(
+                    "Could not find a method with name '{ViewComponentMethodName}' on view " +
+                    "component '{ViewComponentName}'.",
+                    ViewComponentMethodSelector.SyncMethodName,
+                    context.ViewComponentDescriptor.FullName);
```
rynowak: At this point the guidance is not to log where we throw an exception that will be go unhanded by the framework.

src/Microsoft.AspNet.Mvc.Razor/RazorViewFactory.cs
rynowak: Send mail to the eng team, and share the results with the team. This keeps coming up.

rynowak:
:watch:

src/Microsoft.AspNet.Mvc.Core/ViewComponents/DefaultViewComponentInvoker.cs
kichalla: Discussed offline with @rynowak and @lodejard : This is fine but we do not want to log when are re-throwing an exception.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/TemplateRenderer.cs
kichalla: I actually wanted to have template-view related logging for scenarios where users have custom  templates (ex: Customer display/edit template for model Customer) and in case of any failure where the template for the model is not found, we could log the info having the search locations...but the issue is that this method even is called for all the properties on a model when we probe for display/edit templates, which creates lot of entries in the log...
I am not too strong about this...will remove it in next iteration...

src/Microsoft.AspNet.Mvc.Core/ViewComponents/DefaultViewComponentHelper.cs
kichalla: sure, will do...

src/Microsoft.AspNet.Mvc.Core/ViewComponents/DefaultViewComponentHelper.cs
kichalla: I reverted the changes in this file now, so I will not be fixing it.

src/Microsoft.AspNet.Mvc.Core/ActionResults/PartialViewResult.cs
Eilon: @pranavkm it will once Kiran's Logging PR is in :smile:

src/Microsoft.AspNet.Mvc.Core/ActionResults/PartialViewResult.cs
```diff
@@ -50,10 +51,22 @@ public class PartialViewResult : ActionResult
...
+                    "The partial view '{PartialViewName}' was not found. Searched locations: {SearchedViewLocations}",
+                    viewName,
+                    viewEngineResult.SearchedLocations);
+            }
+
+            var view = viewEngineResult.EnsureSuccessful().View;
+
+            logger.LogVerbose("The partial view '{PartialViewName}' was found successfully.", viewName);
```
Eilon: Can nix the word "successfully". Presumably if it was found, then it was successful. Unless you think you can argue that a view can be unsuccessfully found? :smile: :trollface:

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/HtmlHelper.cs
Eilon: Yeah get the logger directly through DI in the ctor, don't get the factory to create it manually.

src/Microsoft.AspNet.Mvc.Core/ViewComponents/DefaultViewComponentInvokerFactory.cs
```diff
@@ -15,11 +17,13 @@ public class DefaultViewComponentInvokerFactory : IViewComponentInvokerFactory
         public DefaultViewComponentInvokerFactory(
             IServiceProvider serviceProvider,
             ITypeActivatorCache typeActivatorCache,
-            IViewComponentActivator viewComponentActivator)
+            IViewComponentActivator viewComponentActivator,
+            [NotNull] ILoggerFactory loggerFactory)
```
Eilon: NotNullNotNeeded?

src/Microsoft.AspNet.Mvc.Razor/RazorViewFactory.cs
Eilon: Asked and answered: No need to check. DI guarantees not null.

src/Microsoft.AspNet.Mvc.Core/ActionResults/PartialViewResult.cs
Eilon: Though now that I think about it - do we even want to log successful things? What value does that add? I'm thinking that if someone is trying to investigate an issue they probably just want to see failures and other suspicious items. By logging all kinds of successful things (parameter was model bound, type conversion succeeded, filter was run) you could end up with an awful lot of data to sift through.

Eilon: BTW I'm not saying that successful things should *never* be logged, but I think we should be doing very few of those where it's a significant event.

rynowak: @Eilon views to me seem like one of those things that's first class enough to be worth it. Especially because there's a search algorithm behind it that's not well understood by many users.

Eilon: Yeah it's an interesting discussion. I just wonder on a real site how many log entries this will generate. I guess here it's just in the view results so it's one per request. But should it be the view engine doing the logging instead?

kichalla: I initially started with logging at view engine, but avoided it because 
- since its at a very low level, i couldn't find a way to provide enough information(ex: in what context is a partial view being executed?..like a partial view can be executed from Html.Partial, PartialViewResult or a ViewComponent too).
- The probing that *we* do for Display/Edit templates is causing lot of log entries to be created which is unnecessary.

We(@rynowak, @pranavkm and myself) discussed about this and felt that logging top level things make sense..ex: ViewResult, PartialViewResult...we can avoid logging for Html.Partial and ViewComponent's views...thoughts?

src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewResult.cs
```diff
@@ -50,10 +51,22 @@ public class ViewResult : ActionResult
...
-                                 .EnsureSuccessful()
-                                 .View;
+            var viewEngineResult = viewEngine.FindView(context, viewName);
+            if(!viewEngineResult.Success)
+            {
+                logger.LogError(
+                    "The view '{ViewName}' was not found. Searched locations: {SearchedViewLocations}", 
+                    viewName,
+                    viewEngineResult.SearchedLocations);
+            }
```
dougbu: why not do this inside `EnsureSuccessful()`?

src/Microsoft.AspNet.Mvc.Core/ViewComponents/DefaultViewComponentInvoker.cs
```diff
@@ -58,6 +68,12 @@ public void Invoke([NotNull] ViewComponentContext context)
                     context.Arguments);
                 if (syncMethod == null)
                 {
+                    _logger.LogError(
+                    "Could not find a method with name '{ViewComponentMethodName}' on view " +
```
pranavkm: Needs indentation.

src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
```diff
@@ -214,11 +220,21 @@ private IRazorPage GetLayoutPage(ViewContext context, string layoutPath)
             var layoutPageResult = _viewEngine.FindPage(context, layoutPath);
             if (layoutPageResult.Page == null)
             {
+                _logger.LogError(
```
pranavkm: When we spoke about this earlier today, I think the plan was to drop this - it's similar to `Html.Partial` \ templates etc which we aren't logging.

src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewResult.cs
kichalla: Doing inside EnsureSuccessful will lose context on what the view engine result is for (like if the view is for partial or not) and also in the latest iteration we have cut down logging in lot of places except top level concepts like ViewResult and PartialResult.

kichalla:
Updated (I now have logging only for ViewResult and PartialViewResult. Please check)

/cc: @rynowak @Eilon @pranavkm @dougbu

pranavkm:
I'm guessing we don't care about unit tests for this (I don't see a great deal of utility in testing logging).

:shipit: if that's the case.

src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewResult.cs
dougbu: :ok:

lodejard:
:shipit:

At some point we should consider adding event id in the calls to log. Not necessary in this PR, cuz we probably need to figure out if we want to standardize on that practice, but having the additional int can make it easier to query for a type of message w/out using the message text.

Eilon:
:shipit:

PR #2386, opened 2015-04-15
Update ModelBinderAttribute not to throw exceptions from BinderType 

ianhong:
@pranavkm 

Please review this. I don't think we need to add more tests since 

BindModel_ReturnsFalseIfNoBinderTypeIsSet()
BindModel_ForNonModelBinder_Throw()

covers those cases where BinderType == null or not an IModelBinder type


pranavkm:
:shipit:

ianhong:
9daf6b48a1246d1e341a9c7a82516a9d09257b97

Eilon:
Yup looks good.

Issue #2373, opened 2015-04-13
When code-gening inherited chunks that replace TModel token, wrap it in line pragmas


pranavkm:
As part of directive inheritance from _GlobalImports, we perform a string replacement on certain kinds of chunks to change the token `TModel` real model type. For e.g., GlobalImport might contain `@inject IFoo<TModel> MyFoo` which gets code-gened as `public IFoo<MyRealModel> MyFoo { get; set; }`. 

We should instead code-gen this with line pragmas as

```C#
public IFoo<
#line 23 something.cshtml
MyRealModel
#line hidden
> { get; set; }
```

This will ensure errors related to `MyRealModel` (misspelt type name etc) map to the view file correctly.

pranavkm:
@NTaylorMullen and I looked at this as part of 89f58aa49f14e1741d12f6c032413d1f944d260a and discovered line mappings need to be above full statements and wrapping partial lines like this generally gets ignored. We made some changes as to how the `@model` gets code gened as part of ff4100e2925579b567be463474ffde255c8a3a6a so design time mappings continue working correctly.

Issue #1968, opened 2015-02-06
Razor CompilationFailedExceptions contain design-time-missing content and incorrect line/column information


yishaigalatzer:
For instance:  I have an exception that fails with:
Additional information: C:\GIT\DataPoint\Source\DataPoint.Web\shell\default\_shelllayout.cshtml(28,36): error CS0115: 'ASPV__shell_default__shelllayout_cshtml.ExecuteAsync()': no suitable method found to override
... but opening that file up with a text editor, I see that line # 28 looks like: 

            </div>

... and furthermore, I don't have any calls to ExecuteAsync()  in the markup.  There are assorted calls to RenderSectionAsync that this is probably related to, but the end user shouldn't have to guess about this.  

** Update:
- I noted that the markup causing the issue is contained in CompiledContent property of the exception.  The exception text still directs me to the original, not compiled file, but at least the needed data is already there.

From @mattgal

yishaigalatzer:
@pranavkm please comment

yishaigalatzer:
We need mostly a doc that shows what a minimal base type should look like. @danroth27

Need to make it show up in helios/and diagnostics middleware

pranavkm:
This should get addressed as part of https://github.com/aspnet/Mvc/issues/2321

PR #2390, opened 2015-04-15
Added SourceLocation to inherited chunks \ tag helpers

### LINKED ISSUE #2321: Give accurate location of syntax error in _GlobalImport.cshtml
Repro steps:
(1) Create a Starter Web App
(2) Turn off precompilation by deleting `Compiler` folder
(3) Create syntax error in `Views/_GlobalImport.cshtml` by `@using` a wrong namespace
(4) F5

The error message shows the error is in `Index.cshtml`, instead of `_GlobalImport.cshtml`:
capture

pranavkm:
Updated CompilationResult to support compilation failures from multiple
files.

Fixes #2321

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor.Host/Directives/ChunkInheritanceUtility.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor.Host/Directives/ChunkInheritanceUtility.cs
@@ -126,22 +126,33 @@ namespace Microsoft.AspNet.Mvc.Razor.Directives
 
         private static CodeTree ParseViewFile(RazorTemplateEngine engine,
                                               IFileInfo fileInfo,
-                                              string viewStartPath)
+                                              string globalImportPath)
         {
             using (var stream = fileInfo.CreateReadStream())
             {
                 using (var streamReader = new StreamReader(stream))
                 {
-                    var parseResults = engine.ParseTemplate(streamReader, viewStartPath);
+                    var parseResults = engine.ParseTemplate(streamReader, globalImportPath);
                     var className = ParserHelpers.SanitizeClassName(fileInfo.Name);
                     var language = engine.Host.CodeLanguage;
                     var codeGenerator = language.CreateCodeGenerator(className,
                                                                      engine.Host.DefaultNamespace,
-                                                                     viewStartPath,
+                                                                     globalImportPath,
                                                                      engine.Host);
                     codeGenerator.Visit(parseResults);
 
-                    return codeGenerator.Context.CodeTreeBuilder.CodeTree;
+                    var codeTree = codeGenerator.Context.CodeTreeBuilder.CodeTree;
+                    // Rewrite the location of inherited chunks so they point to the global import file.
+                    foreach (var chunk in codeTree.Chunks)
+                    {
+                        chunk.Start = new SourceLocation(
+                            globalImportPath,
+                            chunk.Start.AbsoluteIndex,
+                            chunk.Start.LineIndex,
+                            chunk.Start.CharacterIndex);
+                    }
+
+                    return codeTree;
                 }
             }
         }
--- a/src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilationFailedException.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilationFailedException.cs
@@ -17,24 +17,24 @@ namespace Microsoft.AspNet.Mvc.Razor.Compilation
         /// <summary>
         /// Instantiates a new instance of <see cref="CompilationFailedException"/>.
         /// </summary>
-        /// <param name="compilationFailure">The <see cref="ICompilationFailure"/> instance containing
+        /// <param name="compilationFailures"><see cref="ICompilationFailure"/>s containing
         /// details of the compilation failure.</param>
         public CompilationFailedException(
-                [NotNull] ICompilationFailure compilationFailure)
-            : base(FormatMessage(compilationFailure))
+                [NotNull] IEnumerable<ICompilationFailure> compilationFailures)
+            : base(FormatMessage(compilationFailures))
         {
-            CompilationFailures = new[] { compilationFailure };
+            CompilationFailures = compilationFailures;
         }
 
         /// <inheritdoc />
         public IEnumerable<ICompilationFailure> CompilationFailures { get; }
 
-        private static string FormatMessage(ICompilationFailure compilationFailure)
+        private static string FormatMessage(IEnumerable<ICompilationFailure> compilationFailure)
         {
             return Resources.CompilationFailed + Environment.NewLine +
                 string.Join(
                     Environment.NewLine,
-                    compilationFailure.Messages.Select(message => message.FormattedMessage));
+                    compilationFailure.SelectMany(f => f.Messages).Select(message => message.FormattedMessage));
         }
     }
 }
--- a/src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilationResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilationResult.cs
@@ -2,8 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using Microsoft.Framework.Runtime;
+using System.Collections.Generic;
 using Microsoft.Framework.Internal;
+using Microsoft.Framework.Runtime;
 
 namespace Microsoft.AspNet.Mvc.Razor.Compilation
 {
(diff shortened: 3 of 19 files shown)


pranavkm:
Ties into Diagnostics changes - https://github.com/aspnet/Diagnostics/pull/123 and Razor changes - https://github.com/aspnet/Razor/pull/357. The error page from Helios already supports multiple error files.

src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationService.cs
```diff
@@ -131,6 +125,56 @@ public CompilationResult Compile([NotNull] RelativeFileInfo fileInfo, [NotNull]
...
+            var diagnosticGroups = diagnostics
+                .Where(IsError)
+                .GroupBy(diagnostic => GetFilePath(relativePath, diagnostic), StringComparer.Ordinal);
+
+            var failures = new List<ICompilationFailure>();
+            foreach (var group in diagnosticGroups)
+            {
+                var sourceFilePath = group.Key;
+                string sourceFileContent;
+                if (string.Equals(assemblyName, sourceFilePath, StringComparison.Ordinal))
```
pranavkm: This allows us to show errors if they happen to be produced in the generated code (for instance a custom base type that does have a `RenderAsync` method). Mapping it to the cshtml file is not very helpful in diagnosing the issue.

src/Microsoft.AspNet.Mvc.Razor/Compilation/RazorCompilationService.cs
```diff
@@ -30,39 +36,61 @@ public CompilationResult Compile([NotNull] RelativeFileInfo file)
...
+        // Internal for unit testing
+        internal CompilationResult GetCompilationFailedResult(RelativeFileInfo file, IEnumerable<RazorError> errors)
         {
-            try
+            var messageGroups = errors
+                .GroupBy(razorError =>
+                // If a SourceLocation does not specify a file path, assume it is produced
+                // from parsing the current file.
+                razorError.Location.FilePath ?? file.RelativePath,
+                StringComparer.OrdinalIgnoreCase);
```
pranavkm: Needs to be `Ordinal`

src/Microsoft.AspNet.Mvc.Razor.Host/Directives/ChunkInheritanceUtility.cs
```diff
@@ -126,22 +126,33 @@ public virtual IReadOnlyList<CodeTree> GetInheritedCodeTrees([NotNull] string pa
...
                     var codeGenerator = language.CreateCodeGenerator(className,
                                                                      engine.Host.DefaultNamespace,
-                                                                     viewStartPath,
+                                                                     globalImportPath,
                                                                      engine.Host);
                     codeGenerator.Visit(parseResults);
 
-                    return codeGenerator.Context.CodeTreeBuilder.CodeTree;
+                    var codeTree = codeGenerator.Context.CodeTreeBuilder.CodeTree;
+                    // Rewrite the location of inherited chunks so they point to the global import file.
```
dougbu: nit: blank line before mid-block comments.  in this case, might be best to move this comment above previous line.

src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilationFailedException.cs
```diff
@@ -17,24 +17,24 @@ public class CompilationFailedException : Exception, ICompilationException
...
         {
-            CompilationFailures = new[] { compilationFailure };
+            CompilationFailures = compilationFailures;
         }
 
         /// <inheritdoc />
         public IEnumerable<ICompilationFailure> CompilationFailures { get; }
 
-        private static string FormatMessage(ICompilationFailure compilationFailure)
+        private static string FormatMessage(IEnumerable<ICompilationFailure> compilationFailure)
```
dougbu: `compilationFailures` (add an 's')

src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilationResult.cs
```diff
@@ -31,10 +32,10 @@ protected CompilationResult()
         public string CompiledContent { get; protected set; }
 
         /// <summary>
-        /// Gets the <see cref="ICompilationFailure"/> produced from parsing or compiling the Razor file.
+        /// Gets the <see cref="ICompilationFailure"/>s produced from parsing or compiling the Razor file.
         /// </summary>
         /// <remarks>This property is <c>null</c> when compilation succeeded.</remarks>
```
dougbu: what does it mean if the collection is empty?

pranavkm: It's still considered a compilation failure.

dougbu: :ok: this comment should mention that case

src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationService.cs
dougbu: that's a very general statement and doesn't seem correct.  if it were correct, Razor would not generate `#line` pragmas at all.

dougbu: or, are you commenting only on the case where the `#line` mapping is currently `hidden`?

src/Microsoft.AspNet.Mvc.Razor/Resources.resx
```diff
@@ -189,4 +189,7 @@
   <data name="RazorFileInfoCollection_ResourceCouldNotBeFound" xml:space="preserve">
     <value>The resource '{0}' specified by '{1}' could not be found.</value>
   </data>
+  <data name="GeneratedCodeFileName" xml:space="preserve">
+    <value>Generated Code</value>
```
dougbu: :smiley_cat:

test/Microsoft.AspNet.Mvc.Razor.Host.Test/Directives/ChunkInheritanceUtilityTest.cs
```diff
@@ -39,28 +39,38 @@ @inherits MyBaseType
...
             // Assert
             Assert.Equal(2, codeTrees.Count);
-            var viewStartChunks = codeTrees[0].Chunks;
-            Assert.Equal(3, viewStartChunks.Count);
+            var globalImportChunks = codeTrees[0].Chunks;
+            Assert.Equal(3, globalImportChunks.Count);
 
-            Assert.IsType<LiteralChunk>(viewStartChunks[0]);
-            var usingChunk = Assert.IsType<UsingChunk>(viewStartChunks[1]);
+            var globalImportPath = @"Views\home\_GlobalImport.cshtml";
```
dougbu: should be in `// Arrange` or (worse) at least closer to its first use

dougbu:
:shipit:

src/Microsoft.AspNet.Mvc.Razor/Compilation/RazorCompilationService.cs
```diff
@@ -16,12 +19,15 @@ public class RazorCompilationService : IRazorCompilationService
     {
         private readonly ICompilationService _compilationService;
         private readonly IMvcRazorHost _razorHost;
+        private readonly IFileProvider _fileProvider;
 
         public RazorCompilationService(ICompilationService compilationService,
-                                       IMvcRazorHost razorHost)
+                                       IMvcRazorHost razorHost,
+                                       IOptions<RazorViewEngineOptions> viewEngineOptions)
```
NTaylorMullen: curious: Why not just do `IFileProvider`?

pranavkm: The file system that we use for views specifically hangs off of `RazorViewEngineOptions`. It allows doing things like using resources as view sources - https://github.com/aspnet/Mvc/blob/dev/test/WebSites/RazorEmbeddedViewsWebSite/Startup.cs#L20-L23

NTaylorMullen:
:shipit:

src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationService.cs
pranavkm: Yes, this is for generated code that we produce but don't map back to the cshtml - this includes things like default inherited directives (`IHtmlHelper<TModel>`) and any code that relies on a property \ method being present on the base type `DefineSection`, `RenderAsync` etc). 

* I have a tracking bug for the first case - https://github.com/aspnet/Mvc/issues/2373.
* I wasn't able to find the bug the last time I looked - https://github.com/aspnet/Mvc/issues/1968. Mapping to the generated code should give you more more useful information for errors of this sort.

Issue #1919, opened 2015-01-29
RenderBody must be called from a Layout Page" provides no file information


yishaigalatzer:
Repro:
Create new K Web project, comment out RenderBody() call in _Layout.cshtml:
Error message "InvalidOperationException: RenderBody must be called from a layout page."
Imagine a very large project with multiple layouts, perhaps very large files with lots of lines. Determining which one failed to call RenderBody() can be annoying. This error should list the file for which it's throwing the exception.

@Mattgal

https://github.com/aspnet/Razor/issues/280

PR #2391, opened 2015-04-15
Added path info to RazorPage exceptions

ajaybhargavb:
Issue - #1919 
The exception referenced by the issue has already been fixed. 
I've updated few other exceptions in `RazorPage` and `RazorView`.
@Eilon @pranavkm

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
@@ -608,7 +608,7 @@ namespace Microsoft.AspNet.Mvc.Razor
         {
             if (RenderBodyDelegate == null)
             {
-                var message = Resources.FormatRazorPage_MethodCannotBeCalled(nameof(RenderBody));
+                var message = Resources.FormatRazorPage_MethodCannotBeCalled(nameof(RenderBody), Path);
                 throw new InvalidOperationException(message);
             }
 
@@ -704,7 +704,7 @@ namespace Microsoft.AspNet.Mvc.Razor
         {
             if (_renderedSections.Contains(sectionName))
             {
-                var message = Resources.FormatSectionAlreadyRendered(sectionName);
+                var message = Resources.FormatSectionAlreadyRendered(nameof(RenderSectionAsync), Path, sectionName);
                 throw new InvalidOperationException(message);
             }
 
@@ -721,7 +721,7 @@ namespace Microsoft.AspNet.Mvc.Razor
             else if (required)
             {
                 // If the section is not found, and it is not optional, throw an error.
-                throw new InvalidOperationException(Resources.FormatSectionNotDefined(sectionName));
+                throw new InvalidOperationException(Resources.FormatSectionNotDefined(sectionName, Path));
             }
             else
             {
@@ -748,14 +748,15 @@ namespace Microsoft.AspNet.Mvc.Razor
             // change.
             if (_writerScopes.Count > 0)
             {
-                throw new InvalidOperationException(Resources.RazorPage_YouCannotFlushWhileInAWritingScope);
+                throw new InvalidOperationException(
+                    Resources.FormatRazorPage_CannotFlushWhileInAWritingScope(nameof(FlushAsync), Path));
             }
 
             // Calls to Flush are allowed if the page does not specify a Layout or if it is executing a section in the
             // Layout.
             if (!IsLayoutBeingRendered && !string.IsNullOrEmpty(Layout))
             {
-                var message = Resources.FormatLayoutCannotBeRendered(nameof(FlushAsync));
+                var message = Resources.FormatLayoutCannotBeRendered(Path, nameof(FlushAsync));
                 throw new InvalidOperationException(message);
             }
 
@@ -817,7 +818,7 @@ namespace Microsoft.AspNet.Mvc.Razor
         {
             if (PreviousSectionWriters == null)
             {
-                throw new InvalidOperationException(Resources.FormatRazorPage_MethodCannotBeCalled(methodName));
+                throw new InvalidOperationException(Resources.FormatRazorPage_MethodCannotBeCalled(methodName, Path));
             }
         }
     }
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
```diff
@@ -179,7 +179,7 @@ private bool EnableInstrumentation
                     // the body content. Throwing this exception wouldn't return a 500 (since content has already been
                     // written), but a diagnostic component should be able to capture it.
 
-                    var message = Resources.FormatLayoutCannotBeRendered("FlushAsync");
+                    var message = Resources.FormatLayoutCannotBeRendered(Path, "FlushAsync");
```
dougbu: `nameof(RazorPage.FlushAsync)` while you're here

src/Microsoft.AspNet.Mvc.Razor/Resources.resx
```diff
@@ -148,7 +148,7 @@
     <value>There is no active writing scope to end.</value>
   </data>
   <data name="RazorPage_YouCannotFlushWhileInAWritingScope" xml:space="preserve">
-    <value>You cannot flush while inside a writing scope.</value>
+    <value>You cannot flush while inside a writing scope. (Path: '{0}')</value>
```
dougbu: leave out the colon; not needed when you're already quoting the path.  same for `SectionAlreadyRendered`

src/Microsoft.AspNet.Mvc.Razor/Resources.resx
```diff
@@ -133,7 +133,7 @@
     <value>The layout view '{0}' could not be located. The following locations were searched:{1}</value>
   </data>
   <data name="LayoutCannotBeRendered" xml:space="preserve">
-    <value>A layout page cannot be rendered after '{0}' has been invoked.</value>
+    <value>A layout page cannot be rendered at '{0}' after '{1}' has been invoked.</value>
```
dougbu: - updated sentence does not scan.  suggest `Layout page '{1}' cannot be rendered after '{0}' has been invoked.`
- I reversed the parameters to match `SectionAlreadyRendered` but that's a nit

src/Microsoft.AspNet.Mvc.Razor/Resources.resx
dougbu: also suggest `'{0}' invocation in '{1}' is invalid. That method must not be used inside a writing scope.`

src/Microsoft.AspNet.Mvc.Razor/Resources.resx
```diff
@@ -163,7 +163,7 @@
     <value>Section '{0}' is already defined.</value>
   </data>
   <data name="SectionAlreadyRendered" xml:space="preserve">
-    <value>The section named '{0}' has already been rendered.</value>
+    <value>The section named '{0}' has already been rendered. (Path: '{1}')</value>
```
dougbu: suggest `'{0}' invocation in '{2}' is invalid. Section '{1} has already been rendered.` where caller passes `nameof(RenderSectionAsync)` for the 0th argument.

dougbu:
:watch:

pranavkm:
Couple more places where we could print out the path:

* https://github.com/aspnet/Mvc/blob/render-body-fix/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs#L611-L612
> '{method name}' invocation in '{Path}' is invalid.

* https://github.com/aspnet/Mvc/blob/render-body-fix/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs#L820

* https://github.com/aspnet/Mvc/blob/render-body-fix/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs#L724
> Section '{section}' is not defined in '{Path}'.

src/Microsoft.AspNet.Mvc.Razor/Resources.resx
Eilon: :+1:

src/Microsoft.AspNet.Mvc.Razor/Resources.resx
Eilon: `invocation` sounds weird and so does `that method`... Though the original string said `you cannot flush` which while true, is irrelevant: people can never flush anything while inside a writing scope. It's more like `The XYZ operation cannot be performed in ABC mode in 'PATH'.` or something, no?

Eilon: (And change the string name while we're at it - the word `you` is so weird in an error message.)

src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
ajaybhargavb: `FlushAsync` is not part of `IRazorPage`. So can't access it from here.

dougbu: the issue is the `RazorPage` property hiding the class.  but `nameof(Razor.RazorPage.FlushAsync)` works.

ajaybhargavb:
Updated.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -748,14 +748,14 @@ public Task<HtmlString> RenderSectionAsync([NotNull] string name)
             // change.
             if (_writerScopes.Count > 0)
             {
-                throw new InvalidOperationException(Resources.RazorPage_YouCannotFlushWhileInAWritingScope);
+                throw new InvalidOperationException(Resources.FormatRazorPage_CannotFlushWhileInAWritingScope(nameof(FlushAsync), Path));
```
dougbu: nit: long

test/Microsoft.AspNet.Mvc.Razor.Test/RazorPageTest.cs
```diff
@@ -236,7 +239,8 @@ public class RazorPageTest
             await page.ExecuteAsync();
 
             // Assert
-            Assert.Equal("RenderSection can only be called from a layout page.",
+            Assert.Equal("RenderSection invocation in '/Views/TestPath/Test.cshtml' is invalid. " +
+                "RenderSection can only be called from a layout page.",
                          ex.Message);
```
dougbu: nit: odd indentation; left-shift this line

dougbu:
:shipit:

src/Microsoft.AspNet.Mvc.Razor/Resources.resx
pranavkm: Does this apply to the other messages that were changed in this PR?

Eilon:
:shipit: from me too

pranavkm:
:shipit:

PR #2393, opened 2015-04-16
* Use PropertyHelper.SetValue in DefaultControllerActionArgumentBinder

### LINKED ISSUE #2355: [Perf] Consider caching the PropertySetter created in DefaultControllerActionArgumentBinder
We should avoid reflecting on the controller instance and simply cache the properties on it that are of interest. We could leverage the property caches we have in `DefaultControllerFactory` to do this.

https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs#L71-L82
I

pranavkm:
* Update DefaultModelMetadataProvider to use delegates exposed on
  PropertyHelper instance.

Fixes #2355

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
@@ -68,18 +68,23 @@ namespace Microsoft.AspNet.Mvc
 
         private void ActivateProperties(object controller, Type containerType, Dictionary<string, object> properties)
         {
+            if (properties.Count == 0)
+            {
+                return;
+            }
+
             var propertyHelpers = PropertyHelper.GetProperties(controller);
             foreach (var property in properties)
             {
-                var propertyHelper = propertyHelpers.First(helper => helper.Name == property.Key);
+                var propertyHelper = propertyHelpers.First(helper =>
+                    string.Equals(helper.Name, property.Key, StringComparison.Ordinal));
                 if (propertyHelper.Property == null || !propertyHelper.Property.CanWrite)
                 {
                     // nothing to do
                     return;
                 }
 
-                var setter = PropertyHelper.MakeFastPropertySetter(propertyHelper.Property);
-                setter(controller, property.Value);
+                propertyHelper.SetValue(controller, property.Value);
             }
         }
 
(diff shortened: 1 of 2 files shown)


pranavkm:
Associated with https://github.com/aspnet/Common/pull/12.

cc @harshgMSFT @rynowak

harshgMSFT:
:shipit:

harshgMSFT:
:shipit:

src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
```diff
@@ -68,18 +68,23 @@ public class DefaultControllerActionArgumentBinder : IControllerActionArgumentBi
 
         private void ActivateProperties(object controller, Type containerType, Dictionary<string, object> properties)
         {
+            if (properties.Count == 0)
+            {
+                return;
+            }
```
rynowak: just leave this out unless you profiled it and it matters.

src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
```diff
@@ -68,18 +68,23 @@ public class DefaultControllerActionArgumentBinder : IControllerActionArgumentBi
...
+
             var propertyHelpers = PropertyHelper.GetProperties(controller);
             foreach (var property in properties)
             {
-                var propertyHelper = propertyHelpers.First(helper => helper.Name == property.Key);
+                var propertyHelper = propertyHelpers.First(helper =>
+                    string.Equals(helper.Name, property.Key, StringComparison.Ordinal));
```
rynowak: should we throw a nice exception if no setter can be found? This might happen is someone sets up their own ADs.

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadataProvider.cs
```diff
@@ -121,14 +121,14 @@ protected virtual DefaultMetadataDetails[] CreatePropertyDetails([NotNull] Model
...
+                    propertyEntry.PropertyGetter = propertyHelper.ValueGetter;
                 }
 
                 if (propertyHelper.Property.CanWrite &&
                     propertyHelper.Property.SetMethod?.IsPublic == true &&
                     !key.ModelType.IsValueType())
                 {
-                    propertyEntry.PropertySetter = PropertyHelper.MakeFastPropertySetter(propertyHelper.Property);
+                    propertyEntry.PropertySetter = propertyHelper.ValueSetter;
                 }
```
rynowak: :100:

PR #2395, opened 2015-04-16
Add more details to ModelAttributes

rynowak:
This change adds more information to ModelAttributes, so that metadata
providers can look at the attributes on the property and type separately
if so desired


src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DataMemberRequiredBindingMetadataProvider.cs
```diff
@@ -29,7 +29,7 @@ public void GetBindingMetadata([NotNull] BindingMetadataProviderContext context)
             }
 
             var dataMemberAttribute = context
-                .Attributes
+                .PropertyAttributes
```
rynowak: This is slightly more correct than what we were doing here before (which relies on the fact that `[DataMember]` can't be put on a type).

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/ModelAttributesTest.cs
```diff
@@ -112,10 +140,45 @@ public void GetAttributesForType_IncludesMetadataAttributes()
...
+            Assert.IsType<ClassValidator>(attributes.Attributes[2]);
+            Assert.IsType<BindAttribute>(attributes.Attributes[3]);
+
+            Assert.Equal(2, attributes.PropertyAttributes.Count);
+            Assert.IsType<RequiredAttribute>(attributes.PropertyAttributes[0]);
+            Assert.IsType<RangeAttribute>(attributes.PropertyAttributes[1]);
+
+            Assert.Equal(2, attributes.TypeAttributes.Count);
+            Assert.IsType<ClassValidator>(attributes.TypeAttributes[0]);
+            Assert.IsType<BindAttribute>(attributes.TypeAttributes[1]);
```
rynowak: This (critical) test was missing. The wasn't anything verifying ordering.

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultMetadataDetails.cs
```diff
@@ -19,7 +19,7 @@ public class DefaultMetadataDetails
         /// </summary>
         /// <param name="key">The <see cref="ModelMetadataIdentity"/>.</param>
         /// <param name="attributes">The set of model attributes.</param>
-        public DefaultMetadataDetails(ModelMetadataIdentity key, IReadOnlyList<object> attributes)
+        public DefaultMetadataDetails([NotNull] ModelMetadataIdentity key, [NotNull] ModelAttributes attributes)
```
dougbu: so a `struct` can be `null` now?  :smiley_cat:

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadata.cs
```diff
@@ -48,7 +48,7 @@ public IReadOnlyList<object> Attributes
         {
             get
             {
-                return _details.Attributes;
+                return _details.Attributes.Attributes;
```
dougbu: `ModelAttributes` now available throughout the metadata details processing but not after.  should the whole thing also be available in `ModelMetadata`?

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultMetadataDetails.cs
```diff
@@ -28,7 +28,7 @@ public DefaultMetadataDetails(ModelMetadataIdentity key, IReadOnlyList<object> a
         /// <summary>
         /// Gets or sets the set of model attributes.
         /// </summary>
-        public IReadOnlyList<object> Attributes { get; }
+        public ModelAttributes Attributes { get; }
```
dougbu: should this be renamed `ModelAttributes`?

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/ModelAttributes.cs
```diff
@@ -10,60 +10,80 @@
...
         /// </param>
-        /// <returns>An <see cref="IEnumerable{object}"/> containing the attributes on the
-        /// <paramref name="property"/> before the attributes on the <paramref name="property"/> type.</returns>
-        public static IEnumerable<object> GetAttributesForProperty([NotNull] Type type, [NotNull] PropertyInfo property)
+        /// <returns>A <see cref="ModelAttributes"/> instance with the attributes of the property.</returns>
+        public static ModelAttributes GetAttributesForProperty([NotNull] Type type, [NotNull] PropertyInfo property)
         {
-            // Return the property attributes first.
             var propertyAttributes = property.GetCustomAttributes();
             var typeAttributes = property.PropertyType.GetTypeInfo().GetCustomAttributes();
```
dougbu: (might have come up when @kirthik originally added `ModelMetadataTypeAttribute` support but) should this handle `ModelMetadataTypeAttribute` on `property.PropertyType`?  e.g. could call `GetAttributesForType(property.PropertyType)`

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultMetadataDetails.cs
rynowak: heh.

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadata.cs
rynowak: Moot point - `Attributes` is only defined on `DefaultModelMetadata`. So really this isn't exposed unless you downcast.

I can change this to expose `ModelAttributes` instead

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultMetadataDetails.cs
rynowak: sure

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/ModelAttributes.cs
rynowak: Will open a bug. This is not related to this PR.

rynowak: Also, if you look at the test I added, it includes:
1). Attributes on the property
2). Attributes on the property of the metadata class
3). Attributes on the property's type
4). Attributes on the type of the property of the metadata class

rynowak:
@dougbu updated

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/BindingMetadataProviderContext.cs
```diff
@@ -36,6 +39,16 @@ public class BindingMetadataProviderContext
         public ModelMetadataIdentity Key { get; }
 
         /// <summary>
+        /// Gets the property attributes.
+        /// </summary>
+        public IReadOnlyList<object> PropertyAttributes { get; }
```
Eilon: In theory could avoid "extra storage" and stop copying the ref in the ctor and implement this as something like:

```c#
public IReadOnlyList<object> PropertyAttributes { get; } => attributes.PropertyAttributes;
```

rynowak: That would award a lot of #PranavPoints! I can't believe I didn't think of this.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DataAnnotationsMetadataProviderTest.cs
```diff
@@ -551,7 +551,7 @@ public void GetValidationMetadata_RequiredAttribute_SetsIsRequiredToTrue()
 
             var attributes = new Attribute[] { required };
             var key = ModelMetadataIdentity.ForProperty(typeof(int), "Length", typeof(string));
-            var context = new ValidationMetadataProviderContext(key, attributes);
+            var context = new ValidationMetadataProviderContext(key, new ModelAttributes(attributes, new object[0]));
```
dougbu: looks new and different.  am I correct calls to the 2-parameter `ModelAttributes` constructor are all for property metadata?

rynowak: Yes. Where the test is doing something with property metadata, I'm using the constructor that would be called by `GetAttributesForProperty`. It's a bit of a formality and really not needed but I already took the time to make it consistent.

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/ModelAttributes.cs
dougbu: #2403

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DataAnnotationsMetadataProviderTest.cs
dougbu: nice

dougbu:
:shipit:

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/BindingMetadataProviderContext.cs
rynowak: This is on a class that's created and then immediately thrown away so I'm not concerned about the extra two fields. Let me know if you think this is important.

PR #2396, opened 2015-04-17
Creating correct model state keys when body model binding for missing properties or body. Also adds integration tests

### LINKED ISSUE #2416: InputFormatter errors reported by JsonContractResolver should use the entire model prefix as the model state key.
Consider the following two examples : 

**Example 1**
```c#
   public class Person
   {
        [FromBody]
        [Required]
        public Address Address { get; set; }
   }

   public class Address
   {
         public int Zip { get; set; }
   }
```
Request body `""`
In this case the error key would be `""` and not "prefix.Address" (assuming there is a prefix because of additional metadata/positioning for/of the Person model).
Also the model state dictionary incorrectly has two keys
Key 1 : `""`
Key 2 : `Address` -> This gets added because the validator looks for a key `Address` and not `""` to determine if validation has been done for property `Address`.
In this case there should have been only two keys Address and it should be invalid.

**Example 2**
```c#
   public class Person
   {
        [FromBody]
        public Address Address { get; set; }
   }

   public class Address
   {
        public string Street { get; set; }
   
        [Required]
         public int Zip { get; set; }
   }
```

Request body { "Street" : "Some street" }
In this case the error key would be "Street" and not "prefix.Address.Street" (assuming there is a prefix because of additional metadata/positioning for/of the Person model).
Also the model state dictionary incorrectly has three keys
Key 1 : `""`
Key 2 : `Street` 
Key 3: `Zip`  -> This gets added as a valid entry because the validator looks for a key `Zip` and not `""` to determine if validation has been done for property `Zip`.
In this case there should have been only two keys Address.Street and Address.Zip and the second one should be invalid.

The fix for this we would have to make the model state keys consistent with the rest of the model binding system. In all cases it should add the full prefixed version as the error key.

The fix to this might require #2330 to be fixed.

harshgMSFT:
The model state keys for body bound models which are bound at property will use the entire model name with this change for example
Consider
```c#
        public class Person
        {
            [FromBody]
            public Address Address { get; set; }
        }

        public class Address
        {
            [Required]
            public string Street { get; set; }

            public int Zip { get; set; }
        }
```
Request body { "Zip" : 12345 }
In this case the error key would be "prefix.Address.Street" (assuming there is a prefix because of additional metadata/positioning for/of the Person model).

```c#
public class Person
{
       [Required]
}

public void Action([FromBody]Person p)
{
}
```
Request body { }
In this case the prefix gets ignored and the error key is Name.
Please note this is so that we are compatible with MVC 5.0

```c#
public class Person
{
       [Required]
}

public void Action([FromBody][ModelBinder(Name = "prefix")] Person p)
{
}

public void Action2([FromBody][Bind(Name = "prefix")] Person p)
{
}
```
Request body { }
In both these cases (Action and Action2) the prefix gets ignored and the error key is Name.
This is a slight improvement from mvc, as in MVC the action parameter would be null. 

The followup for this would be to fix #2416 - 
This PR ignores the validation assuming that #2416 will address the issues and update the test.

NOTE: previous versions of mvc did not have property binding and hence there is no precedence in this case. For MVC and Web API it was possible to body bind an action parameter which used an empty prefix instead of a parameter name for adding errors to model state (In case of MVC if a custom prefix was provided, it failed binding from body i.e the parameter was null).

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
@@ -63,6 +63,41 @@ namespace Microsoft.AspNet.Mvc
             return actionArguments;
         }
 
+        public async Task<ModelBindingResult> BindModelAsync(
+            ParameterDescriptor parameter,
+            ModelStateDictionary modelState,
+            OperationBindingContext operationContext)
+        {
+            var metadata = _modelMetadataProvider.GetMetadataForType(parameter.ParameterType);
+            var parameterType = parameter.ParameterType;
+            var modelBindingContext = GetModelBindingContext(
+                parameter.Name,
+                metadata,
+                parameter.BindingInfo,
+                modelState,
+                operationContext);
+
+            var modelBindingResult = await operationContext.ModelBinder.BindModelAsync(modelBindingContext);
+            if (modelBindingResult != null && modelBindingResult.IsModelSet)
+            {
+                var key = modelBindingResult.Key;
+                var modelExplorer = new ModelExplorer(
+                    _modelMetadataProvider,
+                    metadata,
+                    modelBindingResult.Model);
+
+                var validationContext = new ModelValidationContext(
+                    key,
+                    modelBindingContext.BindingSource,
+                    operationContext.ValidatorProvider,
+                    modelState,
+                    modelExplorer);
+                _validator.Validate(validationContext);
+            }
+
+            return modelBindingResult;
+        }
+
         private void ActivateProperties(object controller, Type containerType, Dictionary<string, object> properties)
         {
             var propertyHelpers = PropertyHelper.GetProperties(controller);
@@ -88,31 +123,10 @@ namespace Microsoft.AspNet.Mvc
         {
             foreach (var parameter in parameterMetadata)
             {
-                var metadata = _modelMetadataProvider.GetMetadataForType(parameter.ParameterType);
-                var parameterType = parameter.ParameterType;
-                var modelBindingContext = GetModelBindingContext(
-                    parameter.Name,
-                    metadata,
-                    parameter.BindingInfo,
-                    modelState,
-                    operationContext);
-
-                var modelBindingResult = await operationContext.ModelBinder.BindModelAsync(modelBindingContext);
+                var modelBindingResult = await BindModelAsync(parameter, modelState, operationContext);
                 if (modelBindingResult != null && modelBindingResult.IsModelSet)
                 {
-                    var modelExplorer = new ModelExplorer(
-                        _modelMetadataProvider,
-                        metadata,
-                        modelBindingResult.Model);
-
                     arguments[parameter.Name] = modelBindingResult.Model;
-                    var validationContext = new ModelValidationContext(
-                        modelBindingResult.Key,
-                        modelBindingContext.BindingSource,
-                        operationContext.ValidatorProvider,
-                        modelState,
-                        modelExplorer);
-                    _validator.Validate(validationContext);
                 }
             }
         }
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinders/BodyModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinders/BodyModelBinder.cs
@@ -51,8 +51,13 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             {
                 var model = await formatter.ReadAsync(formatterContext);
 
-                // key is empty to ensure that the model name is not used as a prefix for validation.
-                return new ModelBindingResult(model, key: string.Empty, isModelSet: true);
+                var isTopLevelObject = bindingContext.ModelMetadata.ContainerType == null;
+
+                // For compatibility with MVC 5.0 for top level object we want to consider an empty key instead of 
+                // the parameter name/a custom name. In all other cases (like when binding body to a property) we
+                // consider the entire ModelName as a prefix.
+                var modelBindingKey = isTopLevelObject ? string.Empty : bindingContext.ModelName;
+                return new ModelBindingResult(model, key: modelBindingKey, isModelSet: true);
             }
             catch (Exception ex)
             {
--- a/src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DefaultObjectValidator.cs
+++ b/src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DefaultObjectValidator.cs
@@ -187,6 +187,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
 
                 var propertyBindingName = propertyMetadata.BinderModelName ?? propertyMetadata.PropertyName;
                 var childKey = ModelBindingHelper.CreatePropertyModelName(currentModelKey, propertyBindingName);
+
                 if (!ValidateNonVisitedNodeAndChildren(
                     childKey,
                     propertyValidationContext,
--- a/test/Microsoft.AspNet.Mvc.IntegrationTests/BodyValidationIntegrationTests.cs
+++ b/test/Microsoft.AspNet.Mvc.IntegrationTests/BodyValidationIntegrationTests.cs
@@ -1,0 +1,268 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.ComponentModel.DataAnnotations;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Mvc.ModelBinding;
+using Microsoft.Framework.DependencyInjection;
+using Xunit;
+
+namespace Microsoft.AspNet.Mvc.IntegrationTests
+{
+    public class BodyValidationIntegrationTests
+    {
+        private class Person
+        {
+            [FromBody]
+            [Required]
+            public Address Address { get; set; }
+        }
+
+        private class Address
+        {
+            public string Street { get; set; }
+        }
+
+        [Fact]
+        public async Task FromBodyAndRequiredOnProperty_EmptyBody_AddsModelStateError()
+        {
+            // Arrange
+            var argumentBinder = ModelBindingTestHelper.GetArgumentBinder();
+            var parameter = new ParameterDescriptor()
+            {
+                Name = "Parameter1",
+                BindingInfo = new BindingInfo()
+                {
+                    BinderModelName = "CustomParameter",
+                },
+                ParameterType = typeof(Person)
+            };
+
+            var operationContext = ModelBindingTestHelper.GetOperationBindingContext();
+            var httpContext = operationContext.HttpContext;
+
+            ConfigureHttpRequest(httpContext.Request, string.Empty);
+            var modelState = new ModelStateDictionary();
+
+            // Act
+            var modelBindingResult = await argumentBinder.BindModelAsync(parameter, modelState, operationContext);
+
+            // Assert
+            Assert.NotNull(modelBindingResult);
+            Assert.True(modelBindingResult.IsModelSet);
+            var boundPerson = Assert.IsType<Person>(modelBindingResult.Model);
+            Assert.NotNull(boundPerson);
+            var key = Assert.Single(modelState.Keys);
+            Assert.Equal("CustomParameter.Address", key);
+            Assert.False(modelState.IsValid);
+            var error = Assert.Single(modelState[key].Errors);
+            Assert.Equal("The Address field is required.", error.ErrorMessage);
+        }
+
+        [Fact]
+        public async Task FromBodyOnActionParameter_EmptyBody_AddsModelStateError()
+        {
+            // Arrange
+            var argumentBinder = ModelBindingTestHelper.GetArgumentBinder();
+            var parameter = new ParameterDescriptor()
+            {
+                Name = "Parameter1",
+                BindingInfo = new BindingInfo()
+                {
+                    BinderModelName = "CustomParameter",
+                    BindingSource = BindingSource.Body
+                },
+                ParameterType = typeof(Person)
+            };
+
+            var operationContext = ModelBindingTestHelper.GetOperationBindingContext();
+            var httpContext = operationContext.HttpContext;
+
+            ConfigureHttpRequest(httpContext.Request, "{ \"Id\":1234 }");
+            var modelState = new ModelStateDictionary();
+
+            // Act
+            var modelBindingResult = await argumentBinder.BindModelAsync(parameter, modelState, operationContext);
+
+            // Assert
+            Assert.NotNull(modelBindingResult);
+            Assert.True(modelBindingResult.IsModelSet);
+            var boundPerson = Assert.IsType<Person>(modelBindingResult.Model);
+            Assert.NotNull(boundPerson);
+            var key = Assert.Single(modelState.Keys);
+            Assert.Equal("Address", key);
+            Assert.False(modelState.IsValid);
+            var error = Assert.Single(modelState[key].Errors);
+            Assert.Equal("The Address field is required.",error.ErrorMessage);
+        }
+
+        private class Person4
+        {
+            [FromBody]
+            [Required]
+            public int Address { get; set; }
+        }
+
+        [Fact]
+        public async Task FromBodyAndRequiredOnValueTypeProperty_EmptyBody_AddsModelStateError()
+        {
+            // Arrange
+            var argumentBinder = ModelBindingTestHelper.GetArgumentBinder();
+            var parameter = new ParameterDescriptor()
+            {
+                Name = "Parameter1",
+                BindingInfo = new BindingInfo()
+                {
+                    BinderModelName = "CustomParameter",
+                },
+                ParameterType = typeof(Person4)
+            };
+
+            var operationContext = ModelBindingTestHelper.GetOperationBindingContext();
+            var httpContext = operationContext.HttpContext;
+            ConfigureHttpRequest(httpContext.Request, string.Empty);
+            var actionContext = httpContext.RequestServices.GetRequiredService<IScopedInstance<ActionContext>>().Value;
+            var modelState = actionContext.ModelState;
+
+            // Act
+            var modelBindingResult = await argumentBinder.BindModelAsync(parameter, modelState, operationContext);
+
+            // Assert
+            Assert.NotNull(modelBindingResult);
+            Assert.True(modelBindingResult.IsModelSet);
+            var boundPerson = Assert.IsType<Person4>(modelBindingResult.Model);
+            Assert.NotNull(boundPerson);
+            Assert.False(modelState.IsValid);
+
+            // The error with an empty key is a bug(#2416)  in our implementation which does not append the prefix and
+            // use that along with the path. The expected key here would be CustomParameter.Address.
+            var key = Assert.Single(modelState.Keys, k => k == "");
+            var error = Assert.Single(modelState[""].Errors);
+            Assert.StartsWith(
+                "No JSON content found and type 'System.Int32' is not nullable.",
+                error.Exception.Message);
+        }
+
+        private class Person2
+        {
+            [FromBody]
+            public Address2 Address { get; set; }
+        }
+
+        private class Address2
+        {
+            [Required]
+            public string Street { get; set; }
+
+            public int Zip { get; set; }
+        }
+
+        [Theory]
+        [InlineData("{ \"Zip\" : 123 }")]
+        [InlineData("{}")]
+        public async Task FromBodyOnTopLevelProperty_RequiredOnSubProperty_AddsModelStateError(string inputText)
+        {
+            // Arrange
+            var argumentBinder = ModelBindingTestHelper.GetArgumentBinder();
+            var parameter = new ParameterDescriptor()
+            {
+                BindingInfo = new BindingInfo()
+                {
+                    BinderModelName = "CustomParameter",
+                },
+                ParameterType = typeof(Person2)
+            };
+
+            var operationContext = ModelBindingTestHelper.GetOperationBindingContext();
+            var httpContext = operationContext.HttpContext;
+            ConfigureHttpRequest(httpContext.Request, inputText);
+            var modelState = new ModelStateDictionary();
+
+            // Act
+            var modelBindingResult = await argumentBinder.BindModelAsync(parameter, modelState, operationContext);
+
+            // Assert
+            Assert.NotNull(modelBindingResult);
+            Assert.True(modelBindingResult.IsModelSet);
+            var boundPerson = Assert.IsType<Person2>(modelBindingResult.Model);
+            Assert.NotNull(boundPerson);
+            Assert.False(modelState.IsValid);
+            Assert.Equal(2, modelState.Keys.Count);
+            var zip = Assert.Single(modelState.Keys, k => k == "CustomParameter.Address.Zip");
+            Assert.Equal(ModelValidationState.Valid, modelState[zip].ValidationState);
+
+            var street = Assert.Single(modelState.Keys, k => k == "CustomParameter.Address.Street");
+            Assert.Equal(ModelValidationState.Invalid, modelState[street].ValidationState);
+            var error = Assert.Single(modelState[street].Errors);
+            Assert.Equal("The Street field is required.", error.ErrorMessage);
+        }
+
+        private class Person3
+        {
+            [FromBody]
+            public Address3 Address { get; set; }
+        }
+
+        private class Address3
+        {
+            public string Street { get; set; }
+
+            [Required]
+            public int Zip { get; set; }
+        }
+
+        [Theory]
+        [InlineData("{ \"Street\" : \"someStreet\" }")]
+        [InlineData("{}")]
+        public async Task FromBodyOnProperty_RequiredOnValueTypeSubProperty_AddsModelStateError(string inputText)
+        {
+            // Arrange
+            var argumentBinder = ModelBindingTestHelper.GetArgumentBinder();
+            var parameter = new ParameterDescriptor()
+            {
+                BindingInfo = new BindingInfo()
+                {
+                    BinderModelName = "CustomParameter",
+                },
+                ParameterType = typeof(Person3)
+            };
+
+            var operationContext = ModelBindingTestHelper.GetOperationBindingContext();
+            var httpContext = operationContext.HttpContext;
+            ConfigureHttpRequest(httpContext.Request, inputText);
+            var actionContext = httpContext.RequestServices.GetRequiredService<IScopedInstance<ActionContext>>().Value;
+            var modelState = actionContext.ModelState;
+
+            // Act
+            var modelBindingResult = await argumentBinder.BindModelAsync(parameter, modelState, operationContext);
+
+            // Assert
+            Assert.NotNull(modelBindingResult);
+            Assert.True(modelBindingResult.IsModelSet);
+            var boundPerson = Assert.IsType<Person3>(modelBindingResult.Model);
+            Assert.NotNull(boundPerson);
+            Assert.False(modelState.IsValid);
+            var street = Assert.Single(modelState.Keys, k => k == "CustomParameter.Address.Street");
+            Assert.Equal(ModelValidationState.Valid, modelState[street].ValidationState);
+
+            // The error with an empty key is a bug(#2416) in our implementation which does not append the prefix and
+            // use that along with the path. The expected key here would be Address.
+            var zip = Assert.Single(modelState.Keys, k => k == "CustomParameter.Address.Zip");
+            Assert.Equal(ModelValidationState.Valid, modelState[zip].ValidationState);
+            var error = Assert.Single(modelState[""].Errors);
+            Assert.StartsWith(
+                "Required property 'Zip' not found in JSON. Path ''",
+                error.Exception.Message);
+        }
+
+        private static void ConfigureHttpRequest(HttpRequest request, string jsonContent)
+        {
+            request.Body = new MemoryStream(Encoding.UTF8.GetBytes(jsonContent));
+            request.ContentType = "application/json";
+        }
+    }
+}
--- a/test/Microsoft.AspNet.Mvc.IntegrationTests/TestMvcOptions.cs
+++ b/test/Microsoft.AspNet.Mvc.IntegrationTests/TestMvcOptions.cs
@@ -1,0 +1,24 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Framework.OptionsModel;
+
+namespace Microsoft.AspNet.Mvc.IntegrationTests
+{
+    public class TestMvcOptions : IOptions<MvcOptions>
+    {
+        public TestMvcOptions()
+        {
+            Options = new MvcOptions();
+            MvcOptionsSetup.ConfigureMvc(Options);
+        }
+
+        public MvcOptions Options { get; }
+
+        public MvcOptions GetNamedOptions(string name)
+        {
+            throw new NotImplementedException();
+        }
+    }
+}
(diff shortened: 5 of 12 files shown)


test/Microsoft.AspNet.Mvc.FunctionalTests/Microsoft.AspNet.Mvc.FunctionalTests.xproj
```diff
@@ -13,5 +13,8 @@
   <PropertyGroup>
     <SchemaVersion>2.0</SchemaVersion>
   </PropertyGroup>
+  <ItemGroup>
+    <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
+  </ItemGroup>
```
harshgMSFT: not sure why this shows up.

src/Microsoft.AspNet.Mvc.ModelBinding/Validation/DefaultObjectValidator.cs
```diff
@@ -185,8 +185,15 @@ private void MarkPropertiesAsSkipped(string currentModelKey, ModelMetadata metad
                     Visited = validationContext.Visited
                 };
 
-                var propertyBindingName = propertyMetadata.BinderModelName ?? propertyMetadata.PropertyName;
-                var childKey = ModelBindingHelper.CreatePropertyModelName(currentModelKey, propertyBindingName);
+                // For a body bound model, we do not include the entire chain upto the body.
+                // Instead we start with an empty prefix.
+                var childKey = string.Empty;
```
harshgMSFT: This is a temporary fix: The best way of solving these is to reintroduce the validation node concept to pass data between the modelbinding and validation systems.

test/Microsoft.AspNet.Mvc.FunctionalTests/Microsoft.AspNet.Mvc.FunctionalTests.xproj
dougbu: this is fine.  just means it is a test project.  check it in separately, together w/ the rest of this PR, or whatever...

dougbu:
@harshgMSFT it isn't clear what the expected behaviour is in any of the cases you've outlined or why the current behaviour is incorrect.  please describe the exact Asp.Net 4 and MVC 5 behaviours for these cases.

separately what reported issues (if any) will this PR fix?

test/Microsoft.AspNet.Mvc.TestCommon/TestMockMvcOptionsAccessor.cs
```diff
@@ -0,0 +1,26 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNet.Mvc;
+using Microsoft.AspNet.Mvc.Razor;
+using Microsoft.Framework.OptionsModel;
+
+namespace Microsoft.AspNet.Mvc.Core
+{
+    public class TestMvcOptions : IOptions<MvcOptions>
```
dougbu: name of class and containing file must match

test/Microsoft.AspNet.Mvc.ModelBinding.Test/project.json
```diff
@@ -5,6 +5,8 @@
     "dependencies": {
         "Microsoft.AspNet.FileProviders": "1.0.0-*",
         "Microsoft.AspNet.Http.Core": "1.0.0-*",
+        "Microsoft.AspNet.Mvc":"6.0.0-*",
+        "Microsoft.AspNet.Mvc.Core":"6.0.0-*",
```
dougbu: avoid these dependencies.  tests of the lower-level ModelBinding project should not bring in the dependent Mvc or Mvc.Core projects.  if there's something to test here, do it in Mvc.Core.Test or Mvc.Test

side note: Mvc will automatically bring in Mvc.Core

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/ModelMetadataProviderTest.cs
```diff
@@ -898,7 +898,7 @@ public AttributeInjectModelMetadataProvider(object[] attributes)
                 _attributes = attributes;
             }
 
-            protected override DefaultMetadataDetails CreateTypeDetails([NotNull]ModelMetadataIdentity key)
+            protected override DefaultMetadataDetails CreateTypeDetails(ModelMetadataIdentity key)
```
dougbu: why?

test/Microsoft.AspNet.Mvc.ModelBinding.Test/BodyBindingAndValidationIntegrationTest.cs
```diff
@@ -0,0 +1,186 @@
...
+            public Address Address { get; set; }
+        }
+
+        private class Address
+        {
+            public string Street { get; set; }
+        }
+
+        [Fact]
+        public async Task BodyBoundOnProperty_RequiredOnProperty_AddsModelStateError()
```
dougbu: name is hard to parse.  think you mean something like `FromBodyOnProperty_RequiredOnProperty_...` or `PropertyWithFromBodyAndRequired_...`

test/Microsoft.AspNet.Mvc.ModelBinding.Test/BodyBindingAndValidationIntegrationTest.cs
```diff
@@ -0,0 +1,186 @@
...
+            var operationContext = ModelBindingTestHelper.GetOperationBindingContext();
+            var httpContext = operationContext.HttpContext;
+            ConfigureHttpRequest(httpContext.Request, string.Empty);
+            var modelState = new ModelStateDictionary();
+
+            // Act
+            var model = await argumentBinder.BindModelAsync(parameter, modelState, operationContext);
+
+            // Assert
+            Assert.Equal("The Address field is required.", modelState[""].Errors.Single().ErrorMessage);
```
dougbu: behaviour seems surprising.  expect an error message like "The Address field is required." to be present in a key with "Address" in its name.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/BodyBindingAndValidationIntegrationTest.cs
```diff
@@ -0,0 +1,186 @@
...
+
+        private class Person4
+        {
+            [FromBody]
+            [Required]
+            public int Address { get; set; }
+        }
+
+        [Fact]
+        public async Task BodyBoundOnProperty_RequiredOnValueTypeProperty_AddsModelStateError()
```
dougbu: perhaps `IntPropertyWithFromBodyAndRequired_...`, `ValuePropertyWithFromBodyAndRequired_...`, or `ValueTypePropertyWithFromBodyAndRequired_...`

test/Microsoft.AspNet.Mvc.ModelBinding.Test/BodyBindingAndValidationIntegrationTest.cs
```diff
@@ -0,0 +1,186 @@
...
+            var httpContext = operationContext.HttpContext;
+            ConfigureHttpRequest(httpContext.Request, string.Empty);
+            var modelState = new ModelStateDictionary();
+
+            // Act
+            var model = await argumentBinder.BindModelAsync(parameter, modelState, operationContext);
+
+            // Assert
+            Assert.StartsWith("No JSON content found and type 'System.Int32' is not nullable.",
+                modelState[""].Errors.Single().Exception.Message);
```
dougbu: where is this `Exception` thrown and how would we explain to users that the `Exception` relates to the `Address` property of their `Person4` instance?  as-is this information has nothing to associate it with the problematic property.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/BodyBindingAndValidationIntegrationTest.cs
```diff
@@ -0,0 +1,186 @@
...
+            [Required]
+            public string Street { get; set; }
+
+            public int Zip { get; set; }
+        }
+
+        [Theory]
+        [InlineData("{ \"Zip\" : 123 }")]
+        [InlineData("{}")]
+        public async Task BodyBoundOnTopLevelProperty_RequiredOnSubProperty_AddsModelStateError(string inputText)
```
dougbu: just avoid `BodyBound` from here on...

test/Microsoft.AspNet.Mvc.TestCommon/TestMockMvcOptionsAccessor.cs
harshgMSFT: Will fix.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/project.json
harshgMSFT: I am sure you understand the reason for doing this ( there are integration tests which start operating at `DefaultControllerActionArgumentBinder`). I thought about this and there are two ways to avoid it 
1. Move the integration tests to MVC core test project -
This is not a very good solution as the tests are for specific model binder and validation scenarios and it makes much more sense to keep them in model binding test proj.
2. Create a separate integration project. Which at the moment seemed unnecessary. We can do that once we start adding more tests across the board and if we feel there is a need to.

Please let me know if you have a better suggestion, I am all ears :smile:

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/ModelMetadataProviderTest.cs
harshgMSFT: Will revert, however I think our guidance  has been that the interface has `[NotNull]` and the class can skip it. Is that not the case?

test/Microsoft.AspNet.Mvc.ModelBinding.Test/BodyBindingAndValidationIntegrationTest.cs
harshgMSFT: sure `FromBodyOnProperty_RequiredOnProperty_` sounds better.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/BodyBindingAndValidationIntegrationTest.cs
harshgMSFT: Could you suggest what the model state key should be in this case. 
This exception originates from the JsonSerializer and we would need to change the design to get it to a desriable model state key ( if we choose to do something different ). 
This PR I am only ensuring we are consistent across the board.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/BodyBindingAndValidationIntegrationTest.cs
harshgMSFT: why ...

harshgMSFT:
@dougbu  I have updated the comment to provide a better description. And no there is no issue for this one.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/project.json
dougbu: move the integration tests somewhere else, preferably a new test project because existing ones contain unit or functional tests.

(integration tests focused on model binding and validation scenarios are still integration tests, not model binding unit tests.)

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/ModelMetadataProviderTest.cs
dougbu: do cleanup of unrelated files in a separate PR.  btw the reason to remove `[NotNull]` here is `ModelMetadataIdentity` is a `struct`.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/BodyBindingAndValidationIntegrationTest.cs
dougbu: because the attribute is `[FromBody]` and "body bound" is a random term you're inventing

test/Microsoft.AspNet.Mvc.ModelBinding.Test/BodyBindingAndValidationIntegrationTest.cs
dougbu: "Address" or the same key MVC5 would use for `[FromBody]` parameters with a provided prefix.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/ModelMetadataProviderTest.cs
harshgMSFT: yeah thats true i did not notice that :+1:

test/Microsoft.AspNet.Mvc.ModelBinding.Test/project.json
harshgMSFT: That is what i meant with 2. Sure we can do that.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/BodyBindingAndValidationIntegrationTest.cs
harshgMSFT: I am not testing `[FromBody]`, `body bound` is not a term but a way to describe what the test was doing. Anyway I do not have any issue renaming it .. :) will do.

src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
```diff
@@ -66,6 +66,41 @@ public class DefaultControllerActionArgumentBinder : IControllerActionArgumentBi
...
+                    modelBindingResult.Key,
+                    modelBindingContext.BindingSource,
+                    operationContext.ValidatorProvider,
+                    modelState,
+                    modelExplorer);
+                _validator.Validate(validationContext);
+                return modelBindingResult.Model;
+            }
+
+            return null;
```
rynowak: This is a good refactor. I had scoped out something similar thinking about building these tests.

test/Microsoft.AspNet.Mvc.FunctionalTests/Microsoft.AspNet.Mvc.FunctionalTests.xproj
rynowak: In general just let VS do what it's going to do.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/BodyBindingAndValidationIntegrationTest.cs
```diff
@@ -0,0 +1,186 @@
...
+        private class Address
+        {
+            public string Street { get; set; }
+        }
+
+        [Fact]
+        public async Task BodyBoundOnProperty_RequiredOnProperty_AddsModelStateError()
+        {
+            // Arrange
+            var argumentBinder = ModelBindingTestHelper.GetArgumentBinder();
```
rynowak: `ModelBindingTestHelper` gross gross gross

test/Microsoft.AspNet.Mvc.ModelBinding.Test/BodyBindingAndValidationIntegrationTest.cs
```diff
@@ -0,0 +1,186 @@
...
+                ParameterType = typeof(Person)
+            };
+
+            var operationContext = ModelBindingTestHelper.GetOperationBindingContext();
+            var httpContext = operationContext.HttpContext;
+            ConfigureHttpRequest(httpContext.Request, string.Empty);
+            var modelState = new ModelStateDictionary();
+
+            // Act
+            var model = await argumentBinder.BindModelAsync(parameter, modelState, operationContext);
```
rynowak: Need to see validation on the model as well (even it it's null)

test/Microsoft.AspNet.Mvc.ModelBinding.Test/BodyBindingAndValidationIntegrationTest.cs
```diff
@@ -0,0 +1,186 @@
...
+
+            // Assert
+            Assert.Equal("The Address field is required.", modelState[""].Errors.Single().ErrorMessage);
+        }
+
+        private class Person4
```
rynowak: Person4?

test/Microsoft.AspNet.Mvc.ModelBinding.Test/BodyBindingAndValidationIntegrationTest.cs
```diff
@@ -0,0 +1,186 @@
...
+            var operationContext = ModelBindingTestHelper.GetOperationBindingContext();
+            var httpContext = operationContext.HttpContext;
+            ConfigureHttpRequest(httpContext.Request, string.Empty);
+            var modelState = new ModelStateDictionary();
+
+            // Act
+            var model = await argumentBinder.BindModelAsync(parameter, modelState, operationContext);
+
+            // Assert
+            Assert.StartsWith("No JSON content found and type 'System.Int32' is not nullable.",
```
rynowak: Does this error come from JSON.Net? I dunno if we want to verify it if that's the case.

rynowak: I'd also expect to see more verification here than just exception.Message

src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
```diff
@@ -92,31 +127,10 @@ private void ActivateProperties(object controller, Type containerType, Dictionar
...
-                    parameter.Name,
-                    metadata,
-                    parameter.BindingInfo,
-                    modelState,
-                    operationContext);
-
-                var modelBindingResult = await operationContext.ModelBinder.BindModelAsync(modelBindingContext);
-                if (modelBindingResult != null && modelBindingResult.IsModelSet)
+                var model = await BindModelAsync(parameter, modelState, operationContext);
+                if (model != null)
```
rynowak: Note that this is a change in behavior. What happens if the model was bound, but the value was null - now there's no entry in the dictionary. I think we need to return the MBR instead.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/BodyBindingAndValidationIntegrationTest.cs
rynowak: I think the checklist here for these tests to verify (unless N/A):
1). Model Binding Result (null/not-null, success/failure)
2). Model Values
3). ModelState.IsValue
4). Individual model keys

The point of these tests is to lock down these contracts in depth, so verifying just the things we think are significant for the test is a trap.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/BodyBindingAndValidationIntegrationTest.cs
rynowak: Ah, I get it. you're defining a new model class for each test so they are independent.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/ModelBindingTestHelper.cs
```diff
@@ -0,0 +1,88 @@
...
+using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Http.Core;
+using Microsoft.AspNet.Mvc.Core;
+using Microsoft.AspNet.Mvc.ModelBinding.Validation;
+using Microsoft.AspNet.Routing;
+using Microsoft.Framework.DependencyInjection;
+
+namespace Microsoft.AspNet.Mvc.ModelBinding.Test
+{
+    public static class ModelBindingTestHelper
```
rynowak: We might want to just define a base-class and inherit from it. Yes, that's right. I just suggested actually using inheritance. That's how much I hate seeing 'helper' in code.

harshgMSFT: nah, I would hate inherited code here, a static class/methods  is/are much more useful, we can debate about the name.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/BodyBindingAndValidationIntegrationTest.cs
harshgMSFT: Yes and Yes, I should have included those.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/BodyBindingAndValidationIntegrationTest.cs
harshgMSFT: Please suggest alternative, ( I would not like to add inheritance), alternative names are welcomed.

harshgMSFT:
@dougbu @rynowak updated with the behaviors agreed on.

src/Microsoft.AspNet.Mvc.Core/ModelBinders/BodyModelBinder.cs
```diff
@@ -51,8 +51,13 @@ public BodyModelBinder()
             {
                 var model = await formatter.ReadAsync(formatterContext);
 
-                // key is empty to ensure that the model name is not used as a prefix for validation.
-                return new ModelBindingResult(model, key: string.Empty, isModelSet: true);
+                var isTopLevelObject = bindingContext.ModelMetadata.ContainerType == null;
```
rynowak: Didn't we conclude in another discussion that this kind of logic is incorrect, and we can only really know this by adding it back to the context?

src/Microsoft.AspNet.Mvc.Core/ModelBinders/BodyModelBinder.cs
```diff
@@ -51,8 +51,13 @@ public BodyModelBinder()
...
+
+                // For compatibility with MVC 5.0 for top level object we want to consider an empty key instead of 
+                // the parameter name/a custom name. In all other cases (like when binding body to a property) we
+                // consider the entire ModelName as a prefix.
+                var modelBindingKey = isTopLevelObject ? string.Empty : bindingContext.ModelName;
```
rynowak: This seems like to me like the opposite of what we discussed in the meeting.  (Using the empty key for a top-level-model). Is this truly a compat scenario with MVC5?

test/Microsoft.AspNet.Mvc.IntegrationTests/BodyValidationIntegrationTests.cs
```diff
@@ -0,0 +1,263 @@
...
+
+            // Assert
+            Assert.NotNull(modelBindingResult);
+            Assert.True(modelBindingResult.IsModelSet);
+            var boundPerson = Assert.IsType<Person>(modelBindingResult.Model);
+            Assert.NotNull(boundPerson);
+            var key = Assert.Single(modelState.Keys);
+            Assert.Equal("CustomParameter.Address", key);
+            Assert.False(modelState.IsValid);
+            Assert.Equal("The Address field is required.", modelState[key].Errors.Single().ErrorMessage);
```
rynowak: some whitespace would help here.

test/Microsoft.AspNet.Mvc.IntegrationTests/BodyValidationIntegrationTests.cs
```diff
@@ -0,0 +1,263 @@
...
+
+            // Assert
+            Assert.NotNull(modelBindingResult);
+            Assert.True(modelBindingResult.IsModelSet);
+            var boundPerson = Assert.IsType<Person4>(modelBindingResult.Model);
+            Assert.NotNull(boundPerson);
+            Assert.False(modelState.IsValid);
+
+            // The error with an empty key is a bug(#2416)  in our implementation which does not append the prefix and
+            // use that along with the path. The expected key here would be CustomParameter.Address.
```
rynowak: :heart_eyes:

test/Microsoft.AspNet.Mvc.IntegrationTests/TestMvcOptions.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Framework.OptionsModel;
+
+namespace Microsoft.AspNet.Mvc.IntegrationTests
+{
+    public class TestMvcOptions : IOptions<MvcOptions>
```
rynowak: This already exists elsewhere in code - suggest moving the existing one to TestCommon and using it.

harshgMSFT: I had that initially in TestCommon, however that meant a reference to Mvc.Core even in projects which do not refer it.
The case in point is the ModelBinding.Test. Adding this to TestCommon would mean updating ModelBinding.Test 's project.json to include Mvc.Core. Didnt want to get into that debate right now.

src/Microsoft.AspNet.Mvc.Core/ModelBinders/BodyModelBinder.cs
harshgMSFT: This is why this change was added in the first place 27bdec40

src/Microsoft.AspNet.Mvc.Core/ModelBinders/BodyModelBinder.cs
harshgMSFT: Yes, however that is a corner scenario and we should address this when we address that issue.

test/Microsoft.AspNet.Mvc.IntegrationTests/TestMvcOptions.cs
rynowak: :+1:

src/Microsoft.AspNet.Mvc.Core/ModelBinders/BodyModelBinder.cs
rynowak: Discussed in person, I'm on board with this.

rynowak:
:shipit:

src/Microsoft.AspNet.Mvc.Core/ModelBinders/BodyModelBinder.cs
harshgMSFT: #2425

src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
```diff
@@ -63,6 +63,41 @@ public class DefaultControllerActionArgumentBinder : IControllerActionArgumentBi
             return actionArguments;
         }
 
+        public async Task<ModelBindingResult> BindModelAsync(
```
dougbu: this class has a `<summary/>` but no member-level doc comments.  prefer if you add comments for new `public` method but at least open a "task" issue to do the necessary.

src/Microsoft.AspNet.Mvc.Core/ModelBinders/BodyModelBinder.cs
dougbu: won't `bindingContext.ModelName` be empty in all cases where this binder matches at the top level?  what's the exact scenario where `ModelName` will be something else?

test/Microsoft.AspNet.Mvc.IntegrationTests/BodyValidationIntegrationTests.cs
```diff
@@ -0,0 +1,263 @@
...
+            var key = Assert.Single(modelState.Keys, k => k == "");
```
dougbu: inconsistency with previous tests makes this a bit harder to understand.  suggest instead the following where it's the second line which shows the bug.
``` c#
var key = Assert.Single(modelState.Keys);
Assert.Empty(key);
```

test/Microsoft.AspNet.Mvc.IntegrationTests/BodyValidationIntegrationTests.cs
```diff
@@ -0,0 +1,263 @@
...
+            Assert.StartsWith("No JSON content found and type 'System.Int32' is not nullable.",
+                modelState[""].Errors.Single().Exception.Message);
```
dougbu: use `Assert.Single()` for the `Errors` as you've done in other tests

dougbu: same for a couple of other mid-assertion `Single()` calls in this PR

src/Microsoft.AspNet.Mvc.Core/ModelBinders/BodyModelBinder.cs
dougbu: OIC.  had to read the comments a couple more times to get this only kicks in when `MutableObjectModelBinder`'s similar code is not relevant.

test/Microsoft.AspNet.Mvc.IntegrationTests/BodyValidationIntegrationTests.cs
harshgMSFT: sorry i dont get it how is the second line showing the bug, there would be more than one entry in this case and that is the bug, the reason for the bug is the addition of an empty key by the input formatter.

dougbu:
:shipit: with the perfect doc comment addition :smile_cat:

src/Microsoft.AspNet.Mvc.Core/ModelBinders/BodyModelBinder.cs
```diff
@@ -51,8 +51,13 @@ public BodyModelBinder()
             {
                 var model = await formatter.ReadAsync(formatterContext);
 
-                // key is empty to ensure that the model name is not used as a prefix for validation.
-                return new ModelBindingResult(model, key: string.Empty, isModelSet: true);
+                var isTopLevelObject = bindingContext.ModelMetadata.ContainerType == null;
+
+                // For compatibility with MVC 5.0 for top level object we want to consider an empty key instead of 
+                // the parameter name/a custom name. In all other cases (like when binding body to a property) we
```
harshgMSFT: remove extra a in comment.

src/Microsoft.AspNet.Mvc.Core/ModelBinders/BodyModelBinder.cs
harshgMSFT: > this only kicks in when MutableObjectModelBinder's similar code is not relevant.

I didnt get that.

test/Microsoft.AspNet.Mvc.IntegrationTests/BodyValidationIntegrationTests.cs
dougbu: now I see but your code comment is about an incorrect key and says nothing about an extra key.  further your assertions don't check anything about the correct key.  please update the comment and add `Assert.Equal(2, modelState.Keys.Count)` and so on.

src/Microsoft.AspNet.Mvc.Core/ModelBinders/BodyModelBinder.cs
dougbu: sorry, was actually thinking of similar code in `CompositeModelBinder`.  this code is basically doing the same work.

src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
harshgMSFT: Will add doc comments for the public api surface for this class in a separate pr ( not holding this one for furhter review (as you have already signed off)). Not opening individual issues for adding doc comments, there are tons of places which needs fixing and this could be tracked by a singular issue.

test/Microsoft.AspNet.Mvc.IntegrationTests/BodyValidationIntegrationTests.cs
harshgMSFT: not very keen on adding validation which is validating the bug. Not fixing. Please let me know if you think otherwise.

dougbu: my point was about adding validation of what _works_ in this scenario.  the correct key exists and should be checked.

PR #2397, opened 2015-04-17
Making StringCollectionTextWriter use UnicodeEncoding by default.

sornaks:
The `Encoding` property in `StringCollectionTextWriter` is not used anywhere. The only reason it is present is because `TextWriter` forces it to. So introducing a default constructor to avoid passing static encoding in different places in the code.

Removing code which passes `UTF-8` (mostly test) because .net uses `UTF-16` representation for strings (`StringWriter` does the same - https://msdn.microsoft.com/en-us/library/system.io.stringwriter.encoding.aspx).

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/Rendering/StringCollectionTextWriter.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Rendering/StringCollectionTextWriter.cs
@@ -22,6 +22,15 @@ namespace Microsoft.AspNet.Mvc.Rendering
         private readonly Encoding _encoding;
 
         /// <summary>
+        /// Creates a new instance of <see cref="StringCollectionTextWriter"/> with <see cref="Encoding"/>
+        /// defaulting to <see cref="UnicodeEncoding"/>.
+        /// </summary>
+        public StringCollectionTextWriter()
+            : this(new UnicodeEncoding(bigEndian: false, byteOrderMark: false))
+        {
+        }
+
+        /// <summary>
         /// Creates a new instance of <see cref="StringCollectionTextWriter"/>.
         /// </summary>
         /// <param name="encoding">The character <see cref="Encoding"/> in which the output is written.</param>
--- a/src/Microsoft.AspNet.Mvc.Core/Rendering/Html/HtmlHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Rendering/Html/HtmlHelper.cs
@@ -434,7 +434,7 @@ namespace Microsoft.AspNet.Mvc.Rendering
             object model,
             ViewDataDictionary viewData)
         {
-            using (var writer = new StringCollectionTextWriter())
+            using (var writer = new StringCollectionTextWriter(Encoding.UTF8))
             {
                 await RenderPartialCoreAsync(partialViewName, model, viewData, writer);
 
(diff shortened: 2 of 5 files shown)


sornaks:
/cc: @pranavkm @rynowak  (this change will be helpful when I send PR for `TemplateRenderer` because I am using `StringCollectionTextWriter` in many places. Sending it as a separate PR because it is not really a part of that change.)

davidfowl:
Wait what?

dougbu:
:+1 to @davidfowl's comment: what problem does this change solve? avoid one use of a system-provided default encoding?  the downside of making `Encoding` incorrect when specific information is known (from `RazorPage.Output`) seems to outweigh any such upside.

pranavkm:
I think the idea was to use Unicode as the default encoding in places where we don't know what the Encoding is. Plus the property is unused in `StringCollectionTextWriter` - we don't do any string-to-byte conversions. That said, we should leave the code unchanged in places where we were already passing in the right encoding

dougbu:
>  use Unicode as the default encoding in places where we don't know what the Encoding is

why is Unicode more correct than the existing choice for those places?

> we should leave the code unchanged in places where we were already passing in the right encoding

:+1:

pranavkm:
a) Because it would better approximate the behavior of `StringWriter` in places where we drop-in replace it with `StringCollectionTextWriter`
b) The strings we store are Unicode

sornaks:
I changed it in existing places because the passed in encoding was not used anyway.

dougbu:
> the passed in encoding was not used anyway

we discussed this at length when adding `StringCollectionTextWriter` initially and decided to keep all `TextWriter.Encoding` values consistent -- pass it around.  the only inconsistency was a mistake: should pass `ViewContext.Writer.Encoding` in when creating a `StringCollectionTextWriter` in `PartialAsync()`.

src/Microsoft.AspNet.Mvc.Core/Rendering/StringCollectionTextWriter.cs
```diff
@@ -22,6 +22,15 @@ public class StringCollectionTextWriter : TextWriter
         private readonly Encoding _encoding;
 
         /// <summary>
+        /// Creates a new instance of <see cref="StringCollectionTextWriter"/> with <see cref="Encoding"/>
+        /// defaulting to <see cref="UnicodeEncoding"/>.
+        /// </summary>
+        public StringCollectionTextWriter()
```
dougbu: why add something to the public surface that isn't even used in tests (except tests of the new constructor)?

dougbu:
@sornaks please correct `PartialAsync()`

src/Microsoft.AspNet.Mvc.Core/Rendering/StringCollectionTextWriter.cs
sornaks: Because this can be used as a replacement for `StringWriter` wherever needed. (I am going to do this in the next PR.) Do you want me to mark this `internal` now and make this `public` in the next one?

dougbu: sounds like much of this PR starts from the assumption we wanted a `StringWriter`  but we wanted a `TextWriter` with specific semantics, not a `StringWriter` in particular.

pranavkm:
Can we close this PR?

Issue #1586, opened 2014-11-18
Better error when `TagHelperOutputExtensions.CopyHtmlAttribute()` called with unknown `attributeName`


dougbu:
If the provided `TagHelperContext` does not contain `attributeName`, tag helper authors get very little information on how to fix the problem.  Should improve the error message and help these authors out.

Details: An example of the current error:
```
System.InvalidOperationException : Sequence contains no matching element
Stack Trace:
  at System.Linq.Enumerable.First[TSource](IEnumerable`1 source, Func`2 predicate)
  at Microsoft.AspNet.Mvc.TagHelpers.TagHelperOutputExtensions.CopyHtmlAttribute(TagHelperOutput tagHelperOutput, String attributeName, TagHelperContext context) in C:\dd\k\Mvc.0\src\Microsoft.AspNet.Mvc.TagHelpers\TagHelperOutputExtensions.cs:line 33
  at Microsoft.AspNet.Mvc.TagHelpers.InputTagHelper.Process(TagHelperContext context, TagHelperOutput output) in C:\dd\k\Mvc.0\src\Microsoft.AspNet.Mvc.TagHelpers\InputTagHelper.cs:line 114
  at Microsoft.AspNet.Razor.Runtime.TagHelpers.TagHelper.<ProcessAsync>d__1.MoveNext()
  --- End of stack trace from previous location where exception was thrown ---
  ...
```

Could perhaps use `FirstOrDefault()` and `throw` a better `Exception` if that returns `null`.

dougbu:
323ec2f

PR #2400, opened 2015-04-17
Throw exception with unknown attributeName

henkmollema:
Throw a meaningful exception if the attribute to copy does not exists
in the attributes of the TagHelperContext rather than:
```
System.InvalidOperationException : Sequence contains no matching element.
```

Added test too.

#1586

### DIFF
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/Resources.resx
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/Resources.resx
@@ -138,4 +138,7 @@
   <data name="InvalidEnumArgument" xml:space="preserve">
     <value>The value of argument '{0}' ({1}) is invalid for Enum type '{2}'.</value>
   </data>
+  <data name="TagHelperOutput_AttributeDoesNotExist" xml:space="preserve">
+    <value>The attribute '{0}' does not exist in the {1}.</value>
+  </data>
 </root>
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
@@ -35,8 +35,16 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
             {
                 // We look for the original attribute so we can restore the exact attribute name the user typed.
                 // Approach also ignores changes made to tagHelperOutput[attributeName].
-                var entry = context.AllAttributes.First(
+                var entry = context.AllAttributes.FirstOrDefault(
                     attribute => attribute.Key.Equals(attributeName, StringComparison.OrdinalIgnoreCase));
+
+                if (entry.Equals(default(KeyValuePair<string, object>)))
+                {
+                    throw new ArgumentException(
+                        Resources.FormatTagHelperOutput_AttributeDoesNotExist(attributeName, nameof(TagHelperContext)),
+                        nameof(attributeName));
+                }
+                
                 tagHelperOutput.Attributes.Add(entry.Key, entry.Value);
             }
         }
--- a/test/Microsoft.AspNet.Mvc.TagHelpers.Test/TagHelperOutputExtensionsTest.cs
+++ b/test/Microsoft.AspNet.Mvc.TagHelpers.Test/TagHelperOutputExtensionsTest.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Mvc.Rendering;
 using Microsoft.AspNet.Razor.Runtime.TagHelpers;
+using Microsoft.AspNet.Testing;
 using Microsoft.Framework.WebEncoders;
 using Xunit;
 
@@ -77,6 +78,35 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
             // Assert
             var attribute = Assert.Single(tagHelperOutput.Attributes);
             Assert.Equal(expectedAttribute, attribute);
+        }
+
+        [Fact]
+        public void CopyHtmlAttribute_ThrowsWhenUnknownAttribute()
+        {
+            // Arrange
+            var invalidAttributeName = "hello2";
+            var tagHelperOutput = new TagHelperOutput(
+                "p",
+                attributes: new Dictionary<string, object>());
+            var tagHelperContext = new TagHelperContext(
+                allAttributes: new Dictionary<string, object>(StringComparer.Ordinal)
+                {
+                    { "hello", "world" }
+                },
+                items: new Dictionary<object, object>(),
+                uniqueId: "test",
+                getChildContentAsync: () =>
+                {
+                    var tagHelperContent = new DefaultTagHelperContent();
+                    tagHelperContent.Append("Something");
+                    return Task.FromResult<TagHelperContent>(tagHelperContent);
+                });
+
+            // Act & Assert
+            ExceptionAssert.ThrowsArgument(
+                () => tagHelperOutput.CopyHtmlAttribute(invalidAttributeName, tagHelperContext),
+                "attributeName",
+                "The attribute 'hello2' does not exist in the TagHelperContext.");
         }
 
         [Fact]
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/Properties/Resources.Designer.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/Properties/Resources.Designer.cs
@@ -122,6 +122,22 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
             return string.Format(CultureInfo.CurrentCulture, GetString("InvalidEnumArgument"), p0, p1, p2);
         }
 
+        /// <summary>
+        /// The attribute '{0}' does not exist in the {1}.
+        /// </summary>
+        internal static string TagHelperOutput_AttributeDoesNotExist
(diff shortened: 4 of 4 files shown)


test/Microsoft.AspNet.Mvc.TagHelpers.Test/TagHelperOutputExtensionsTest.cs
```diff
@@ -80,6 +80,39 @@ public void CopyHtmlAttribute_DoesNotOverrideAttributes()
...
+                uniqueId: "test",
+                getChildContentAsync: () =>
+                {
+                    var tagHelperContent = new DefaultTagHelperContent();
+                    tagHelperContent.Append("Something");
+                    return Task.FromResult<TagHelperContent>(tagHelperContent);
+                });
+
+            string expectedMessage = 
+                "The specified attribute 'hello2' does not exists in the attributes of the TagHelperContext." + Environment.NewLine +
```
henkmollema: What message is desired here?

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
```diff
@@ -35,8 +35,16 @@ public static class TagHelperOutputExtensions
...
                 // We look for the original attribute so we can restore the exact attribute name the user typed.
                 // Approach also ignores changes made to tagHelperOutput[attributeName].
-                var entry = context.AllAttributes.First(
+                var entry = context.AllAttributes.FirstOrDefault(
                     attribute => attribute.Key.Equals(attributeName, StringComparison.OrdinalIgnoreCase));
+
+                if (entry.Equals(default(KeyValuePair<string, object>)))
+                {
+                    throw new ArgumentException(
+                        string.Format("The specified attribute '{0}' does not exists in the attributes of the TagHelperContext.", attributeName),
```
henkmollema: What message is desired here?

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
```diff
@@ -35,8 +35,16 @@ public static class TagHelperOutputExtensions
             {
                 // We look for the original attribute so we can restore the exact attribute name the user typed.
                 // Approach also ignores changes made to tagHelperOutput[attributeName].
-                var entry = context.AllAttributes.First(
+                var entry = context.AllAttributes.FirstOrDefault(
                     attribute => attribute.Key.Equals(attributeName, StringComparison.OrdinalIgnoreCase));
+
+                if (entry.Equals(default(KeyValuePair<string, object>)))
```
henkmollema: `KeyValuePair<,>` is a `struct`, so can't check for null.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/TagHelperOutputExtensionsTest.cs
```diff
@@ -80,6 +80,39 @@ public void CopyHtmlAttribute_DoesNotOverrideAttributes()
...
+                "Parameter name: attributeName";
```
henkmollema: Hardcoded parameter name seems dirty, can this be done nicer?

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
Eilon: Yeah that's kind of ugly but I suppose it's correct.

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
Eilon: How about:

`The attribute '{0}' does not exist in the {1}.`

And then pass in `nameof(TagHelperContext)` for the `{1}`.

Also, can you put the message in the RESX file (look at the other examples where exception messages are used, e.g. https://github.com/henkmollema/Mvc/blob/fix-1586/src/Microsoft.AspNet.Mvc.TagHelpers/SelectTagHelper.cs#L66-L70

After editing the RESX file, I think you need to run `build resx` from the command line to regenerate the `.designer.cs` file.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/TagHelperOutputExtensionsTest.cs
Eilon: That's pretty much the normal way.

henkmollema:
@Eilon updated, addressed your feedback.

Eilon:
:shipit: thanks!

@dougbu can you do a quick review and merge?

henkmollema:
Gimme a call when you need me to rebase/squash.

Eilon:
@henkmollema could you do that right now?

henkmollema:
@Eilon @dougbu squashed and rebased.

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
dougbu: > Perhaps, the `AllAttributes` dictionary should ignore case, which makes obtaining the value by key easier too.

`AllAttributes` is set up to ignore case.  See https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs#L63

the code here is _not_ just looking up the correct value; it's trying to ensure the attribute is restored with the original attribute name i.e. correct case.  should remain as-is.

and the `entry.Equals()` check here is correct.  thanks for this change.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/TagHelperOutputExtensionsTest.cs
```diff
@@ -80,6 +80,39 @@ public void CopyHtmlAttribute_DoesNotOverrideAttributes()
...
+                    tagHelperContent.Append("Something");
+                    return Task.FromResult<TagHelperContent>(tagHelperContent);
+                });
+
+            string expectedMessage = 
+                "The attribute 'hello2' does not exist in the TagHelperContext." + Environment.NewLine +
+                "Parameter name: attributeName";
+
+            // Act
+            var ex = Assert.Throws<ArgumentException>(() => { tagHelperOutput.CopyHtmlAttribute(invalidAttributeName, tagHelperContext); });
```
dougbu: we're not great about using the more-complete `Assert.Throws<ArgumentException>()` overloads.  but should switch pass the `paramName` argument i.e. `Assert.Throws<ArgumentException>("attributeName", () => ...)`

nit: bit messy. can remove the useless curlies: `() => tagHelperOutput.CopyHtmlAttribute(invalidAttributeName, tagHelperContext)`.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/TagHelperOutputExtensionsTest.cs
```diff
@@ -80,6 +80,39 @@ public void CopyHtmlAttribute_DoesNotOverrideAttributes()
...
+                getChildContentAsync: () =>
+                {
+                    var tagHelperContent = new DefaultTagHelperContent();
+                    tagHelperContent.Append("Something");
+                    return Task.FromResult<TagHelperContent>(tagHelperContent);
+                });
+
+            string expectedMessage = 
+                "The attribute 'hello2' does not exist in the TagHelperContext." + Environment.NewLine +
+                "Parameter name: attributeName";
```
dougbu: if you prefer, can use a helper (from the `Microsoft.AspNet.Testing` namespace) instead:
``` c#
ExceptionAssert.ThrowsArgument(
  () => tagHelperOutput.CopyHtmlAttribute(invalidAttributeName, tagHelperContext),
  "attributeName",
  "The attribute 'hello2' does not exist in the TagHelperContext.");
```

then would need neither `expectedMessage` nor `ex` variables.

src/Microsoft.AspNet.Mvc.TagHelpers/Resources.resx
```diff
@@ -138,4 +138,7 @@
   <data name="InvalidEnumArgument" xml:space="preserve">
     <value>The value of argument '{0}' ({1}) is invalid for Enum type '{2}'.</value>
   </data>
+  <data name="TagHelperOutput_AttributeDoesNotExists" xml:space="preserve">
```
dougbu: remove final "s" from this resource name

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
dougbu: to be clear: is anyone proposing additional changes to this method?

dougbu:
:watch: just for the resource rename. (and wondering why the CIs are still failing)

henkmollema:
@dougbu updated and squashed/rebased.

cc @Eilon

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
henkmollema: Allright, that makes sense.

dougbu:
323ec2f

Issue #830, opened 2014-07-21
Make HTML rendering deferred (IHtmlContent)


pranavkm:
Currently `TemplateRenderer.Render` returns a string which gets converted to HtmlString. This causes it to produce a new StringWriter buffer. We could have it return a HelperResult instead - it's consumed in the same way, but would avoid the extra buffer allocation.

yishaigalatzer:
If a template ever renders a large portion of text > 80K there is definitely a major LOH issue that will cause pauses and lots of CPU spent in GC

yishaigalatzer:
@NTaylorMullen please tie this to the TagHelper Content change when working with @sornaks

sornaks:
Note to self: Move `HelperResult` to `Razor.Runtime`.

sornaks:
https://github.com/aspnet/Mvc/commit/f2540f9ba19937f79c1a780ef12dd9a9d5cd84a6

PR #2402, opened 2015-04-17
Making display and editor templates use HtmlString instead of normal strings to improve performance.

sornaks:
https://github.com/aspnet/Mvc/issues/830


rynowak:
Where are the perf numbers?

rynowak:
Closing per discussion at standup, we're going a different way with this change.

Issue #2294, opened 2015-03-30
Invalid dynamic collection binding and validation


anfomin:
@dougbu asked to create new issue in discussion #2268. Current `CollectionModelBinder` can't properly bind collections with dynamic indexes. It returns invalid results in 2 cases.

**1.** I've class with collection property:
```C#
class RootObject
{
	public IList<AnotherObject> Collection { get; set; }
	// other code goes here
}

class AnotherObject
{
	public string Name { get; set; }
}
```

Binding form data:
```
Collection.index = key1,key2
Collection[key1].Name = name1
Collection[key2].Name = name2
```

If I try to bind `RootObject` then `Collection` property is bound successfully, but ModelState contains unvalidated states for `Collection[key1].Name` and `Collection[key2].Name`:
```C#
var model = new RoomObject();
TryUpdateModel(model);
/// ModelState.IsValid returns false - wrong value.
```

**2.** I've the same classes. If I try to bind collection of `AnotherObject` directly with prefix, result collection is still empty:
```C#
var collection = new List<AnotherObject>();
TryUpdateModel(collection, "Collection");
// collection is still empty, even if there is data
```

I'm using latest **beta5** MVC version.

danroth27:
@anfomin Are you literally sending `Collection.index = key1,key2` or is this just a shorthand for sending `Collection.index = key1` and then `Collection.index = key2`?

danroth27:
@dougbu We should at least be consistent with the MVC 5 behavior.

anfomin:
@danroth27 yes, in previous message I wrote a shorthand. I have 2 inputs: `Collection.index = key1` and `Collection.index = key2`.

dougbu:
@anfomin could you please upload a small sample of what you're trying to do?  Want to avoid testing a slightly different scenario.

A GitHub repository is easiest for me but feel free to send a Zip file offline or whatever.

anfomin:
@dougbu ok, I'll prepare sample tomorrow.

anfomin:
@dougbu sample is ready: https://github.com/anfomin/MvcCollectionBindingSample

dougbu:
thanks @anfomin.  I am able to reproduce the tightly-related issues you reported.  both appear to be regressions compared to MVC 5.

one workaround for the `ModelState.IsValid` issue is to remove the "Collection.index" fields entirely.  instead name the fields using the C# indices e.g.
``` c#
for (var index = 0; index < Model.Collection.Count; index++)
[...]

PR #2405, opened 2015-04-19
Bind to readonly non-`null` collections

dougbu:
- parts 1/3 and 2/3 of #2294
- fill readonly gaps in `CollectionModelBinder` and `MutableObjectModelBinder`
 - `CollectionModelBinder.CopyToModel()` and `MutableObjectModelBinder.AddToProperty()` methods
 - add `CopyToModel()` override in `ArrayModelBinder`
- avoid NREs in `GetModel()` overrides

Test handling of readonly collections
- previous tests barely touched this scenario

nits:
- add missing `[NotNull]` attributes
- add missing doc comments
- consolidate a few `[Fact]`s into `[Theory]`s
- simplify some wrapping; shorten a few lines

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
@@ -4,12 +4,12 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Mvc.Core;
 using Microsoft.AspNet.Mvc.ModelBinding;
 using Microsoft.AspNet.Mvc.ModelBinding.Validation;
 using Microsoft.Framework.Internal;
-using Microsoft.Framework.OptionsModel;
 
 namespace Microsoft.AspNet.Mvc
 {
@@ -19,6 +19,10 @@ namespace Microsoft.AspNet.Mvc
     /// </summary>
     public class DefaultControllerActionArgumentBinder : IControllerActionArgumentBinder
     {
+        private static readonly MethodInfo CallPropertyAddRangeOpenGenericMethod =
+            typeof(DefaultControllerActionArgumentBinder).GetTypeInfo().GetDeclaredMethod(
+                nameof(CallPropertyAddRange));
+
         private readonly IModelMetadataProvider _modelMetadataProvider;
         private readonly IObjectModelValidator _validator;
 
@@ -64,12 +68,11 @@ namespace Microsoft.AspNet.Mvc
         }
 
         public async Task<ModelBindingResult> BindModelAsync(
-            ParameterDescriptor parameter,
-            ModelStateDictionary modelState,
-            OperationBindingContext operationContext)
+            [NotNull] ParameterDescriptor parameter,
+            [NotNull] ModelStateDictionary modelState,
+            [NotNull] OperationBindingContext operationContext)
         {
             var metadata = _modelMetadataProvider.GetMetadataForType(parameter.ParameterType);
-            var parameterType = parameter.ParameterType;
             var modelBindingContext = GetModelBindingContext(
                 parameter.Name,
                 metadata,
@@ -98,6 +101,21 @@ namespace Microsoft.AspNet.Mvc
             return modelBindingResult;
         }
 
+        // Called via reflection.
+        private static void CallPropertyAddRange<TElement>(object target, object source)
+        {
+            var targetCollection = (ICollection<TElement>)target;
+            var sourceCollection = source as IEnumerable<TElement>;
+            if (sourceCollection != null && !targetCollection.IsReadOnly)
+            {
+                targetCollection.Clear();
+                foreach (var item in sourceCollection)
+                {
+                    targetCollection.Add(item);
+                }
+            }
+        }
+
         private void ActivateProperties(object controller, Type containerType, Dictionary<string, object> properties)
         {
             var propertyHelpers = PropertyHelper.GetProperties(controller);
@@ -105,17 +123,47 @@ namespace Microsoft.AspNet.Mvc
             {
                 var propertyHelper = propertyHelpers.First(helper =>
                     string.Equals(helper.Name, property.Key, StringComparison.Ordinal));
-                if (propertyHelper.Property == null || !propertyHelper.Property.CanWrite)
+                var propertyType = propertyHelper.Property.PropertyType;
+                var source = property.Value;
+                if (propertyHelper.Property.CanWrite && propertyHelper.Property.SetMethod?.IsPublic == true)
                 {
-                    // nothing to do
-                    return;
+                    // Handle settable property. Do not set the property if the type is a non-nullable type.
+                    if (source != null || propertyType.AllowsNullValue())
+                    {
+                        propertyHelper.SetValue(controller, source);
+                    }
+
+                    continue;
                 }
 
-                // Do not set the property if the type is a non nullable type.
-                if (property.Value != null || propertyHelper.Property.PropertyType.AllowsNullValue())
+                if (propertyType.IsArray)
                 {
-                    propertyHelper.SetValue(controller, property.Value);
+                    // Do not attempt to copy values into an array because an array's length is immutable. This choice
+                    // is also consistent with MutableObjectModelBinder's handling of a read-only array property.
+                    continue;
                 }
+
+                var target = propertyHelper.GetValue(controller);
+                if (source == null || target == null)
+                {
+                    // Nothing to do when source or target is null.
+                    continue;
+                }
+
+                // Determine T if this is an ICollection<T> property.
+                var collectionTypeArguments = propertyType
+                    .ExtractGenericInterface(typeof(ICollection<>))
+                    ?.GenericTypeArguments;
+                if (collectionTypeArguments == null)
+                {
+                    // Not a collection model.
+                    continue;
+                }
+
+                // Handle a read-only collection property.
+                var propertyAddRange = CallPropertyAddRangeOpenGenericMethod.MakeGenericMethod(
+                    collectionTypeArguments);
+                propertyAddRange.Invoke(obj: null, parameters: new[] { target, source });
             }
         }
 
--- a/src/Microsoft.AspNet.Mvc.ModelBinding/Binders/ArrayModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.ModelBinding/Binders/ArrayModelBinder.cs
@@ -4,12 +4,18 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using Microsoft.Framework.Internal;
 
 namespace Microsoft.AspNet.Mvc.ModelBinding
 {
+    /// <summary>
+    /// <see cref="IModelBinder"/> implementation for binding array values.
+    /// </summary>
+    /// <typeparam name="TElement">Type of elements in the array.</typeparam>
     public class ArrayModelBinder<TElement> : CollectionModelBinder<TElement>
     {
-        public override Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindingContext)
+        /// <inheritdoc />
+        public override Task<ModelBindingResult> BindModelAsync([NotNull] ModelBindingContext bindingContext)
         {
             if (bindingContext.ModelMetadata.IsReadOnly)
             {
@@ -19,9 +25,17 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             return base.BindModelAsync(bindingContext);
         }
 
+        /// <inheritdoc />
         protected override object GetModel(IEnumerable<TElement> newCollection)
         {
-            return newCollection.ToArray();
+            return newCollection?.ToArray();
+        }
+
+        /// <inheritdoc />
+        protected override void CopyToModel([NotNull] object target, IEnumerable<TElement> sourceCollection)
+        {
+            // Do not attempt to copy values into an array because an array's length is immutable. This choice is also
+            // consistent with MutableObjectModelBinder's handling of a read-only array property.
         }
     }
 }
--- a/src/Microsoft.AspNet.Mvc.ModelBinding/Binders/CollectionModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.ModelBinding/Binders/CollectionModelBinder.cs
@@ -4,16 +4,23 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Mvc.ModelBinding.Internal;
+using Microsoft.Framework.Internal;
 
 namespace Microsoft.AspNet.Mvc.ModelBinding
 {
+    /// <summary>
+    /// <see cref="IModelBinder"/> implementation for binding collection values.
+    /// </summary>
+    /// <typeparam name="TElement">Type of elements in the collection.</typeparam>
     public class CollectionModelBinder<TElement> : IModelBinder
     {
-        public virtual async Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindingContext)
+        /// <inheritdoc />
+        public virtual async Task<ModelBindingResult> BindModelAsync([NotNull] ModelBindingContext bindingContext)
         {
             ModelBindingHelper.ValidateBindingContext(bindingContext);
 
@@ -23,19 +30,40 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             }
 
             var valueProviderResult = await bindingContext.ValueProvider.GetValueAsync(bindingContext.ModelName);
-            var bindCollectionTask = valueProviderResult != null ?
-                    BindSimpleCollection(bindingContext, valueProviderResult.RawValue, valueProviderResult.Culture) :
-                    BindComplexCollection(bindingContext);
-            var boundCollection = await bindCollectionTask;
-            var model = GetModel(boundCollection);
+
+            IEnumerable<TElement> boundCollection;
+            if (valueProviderResult == null)
+            {
+                boundCollection = await BindComplexCollection(bindingContext);
+            }
+            else
+            {
+                boundCollection = await BindSimpleCollection(
+                    bindingContext,
+                    valueProviderResult.RawValue,
+                    valueProviderResult.Culture);
+            }
+
+            var model = bindingContext.Model;
+            if (model == null)
+            {
+                model = GetModel(boundCollection);
+            }
+            else
+            {
+                // Special case for TryUpdateModelAsync(collection, ...) scenarios. Model is null in all other cases.
+                CopyToModel(model, boundCollection);
+            }
+
             return new ModelBindingResult(model, bindingContext.ModelName, isModelSet: true);
         }
 
         // Used when the ValueProvider contains the collection to be bound as a single element, e.g. the raw value
         // is [ "1", "2" ] and needs to be converted to an int[].
-        internal async Task<IEnumerable<TElement>> BindSimpleCollection(ModelBindingContext bindingContext,
-                                                                 object rawValue,
-                                                                 CultureInfo culture)
+        internal async Task<IEnumerable<TElement>> BindSimpleCollection(
+            ModelBindingContext bindingContext,
+            object rawValue,
+            CultureInfo culture)
         {
             if (rawValue == null)
             {
@@ -62,7 +90,8 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
                 };
 
                 object boundValue = null;
-                var result = await bindingContext.OperationBindingContext.ModelBinder.BindModelAsync(innerBindingContext);
+                var result =
+                    await bindingContext.OperationBindingContext.ModelBinder.BindModelAsync(innerBindingContext);
                 if (result != null)
                 {
                     boundValue = result.Model;
@@ -79,11 +108,13 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             var indexPropertyName = ModelBindingHelper.CreatePropertyModelName(bindingContext.ModelName, "index");
             var valueProviderResultIndex = await bindingContext.ValueProvider.GetValueAsync(indexPropertyName);
             var indexNames = CollectionModelBinderUtil.GetIndexNamesFromValueProviderResult(valueProviderResultIndex);
+
             return await BindComplexCollectionFromIndexes(bindingContext, indexNames);
         }
 
-        internal async Task<IEnumerable<TElement>> BindComplexCollectionFromIndexes(ModelBindingContext bindingContext,
-                                                                             IEnumerable<string> indexNames)
+        internal async Task<IEnumerable<TElement>> BindComplexCollectionFromIndexes(
+            ModelBindingContext bindingContext,
+            IEnumerable<string> indexNames)
         {
             bool indexNamesIsFinite;
             if (indexNames != null)
@@ -114,7 +145,8 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
 
                 var modelType = bindingContext.ModelType;
 
-                var result = await bindingContext.OperationBindingContext.ModelBinder.BindModelAsync(childBindingContext);
+                var result =
+                    await bindingContext.OperationBindingContext.ModelBinder.BindModelAsync(childBindingContext);
                 if (result != null)
                 {
                     didBind = true;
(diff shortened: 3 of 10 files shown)


src/Microsoft.AspNet.Mvc.ModelBinding/Binders/CollectionModelBinder.cs
```diff
@@ -133,13 +165,49 @@ public class CollectionModelBinder<TElement> : IModelBinder
...
 
+        /// <summary>
+        /// Adds values from <paramref name="sourceCollection"/> to given <paramref name="target"/>.
+        /// </summary>
+        /// <param name="target"><see cref="object"/> into which values are copied.</param>
+        /// <param name="sourceCollection">
+        /// Collection of values retrieved from value providers. Or <c>null</c> if nothing was bound.
+        /// </param>
+        /// <remarks>Called only in TryUpdateModelAsync(collection, ...) scenarios.</remarks>
+        protected virtual void CopyToModel([NotNull] object target, IEnumerable<TElement> sourceCollection)
```
dougbu: this part of the feature was removed (along w/ `ModelValidationNode`) in f19c2e4.  previous `CollectionModelBinder.CreateOrReplaceCollection()` method checked equivalent of `targetCollection.IsReadOnly` called `targetCollection.Clear()` before doing the additions.  restore both in next iteration...

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/ArrayModelBinder.cs
```diff
@@ -19,9 +26,42 @@ public override Task<ModelBindingResult> BindModelAsync(ModelBindingContext bind
...
+                return null;
+            }
+
             return newCollection.ToArray();
         }
+
+        /// <inheritdoc />
+        protected override void CopyToModel([NotNull] object target, IEnumerable<TElement> sourceCollection)
+        {
+            TElement[] targetArray = target as TElement[];
```
harshgMSFT: Use `var` ? is explicit type needed ?

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/ArrayModelBinder.cs
```diff
@@ -19,9 +26,42 @@ public override Task<ModelBindingResult> BindModelAsync(ModelBindingContext bind
...
+            Debug.Assert(targetArray != null); // This binder is instantiated only for array model types.
+
+            if (sourceCollection != null && targetArray != null)
```
harshgMSFT: is `targetArray` ever expected to be null?

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/ArrayModelBinder.cs
```diff
@@ -19,9 +26,42 @@ public override Task<ModelBindingResult> BindModelAsync(ModelBindingContext bind
...
+            {
+                int maxIndex = targetArray.Length - 1;
```
harshgMSFT: var

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/ArrayModelBinder.cs
```diff
@@ -19,9 +26,42 @@ public override Task<ModelBindingResult> BindModelAsync(ModelBindingContext bind
...
+                int index = 0;
```
harshgMSFT: var

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/ArrayModelBinder.cs
```diff
@@ -19,9 +26,42 @@ public override Task<ModelBindingResult> BindModelAsync(ModelBindingContext bind
...
+                foreach (var element in sourceCollection)
+                {
+                    if (index > maxIndex)
+                    {
+                        break;
```
harshgMSFT: what happens to the remaining elements in source array? what is the MVC 5.0 behavior in that case?

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/ArrayModelBinder.cs
```diff
@@ -19,9 +26,42 @@ public override Task<ModelBindingResult> BindModelAsync(ModelBindingContext bind
             return base.BindModelAsync(bindingContext);
         }
 
+        /// <inheritdoc />
         protected override object GetModel(IEnumerable<TElement> newCollection)
         {
+            if (newCollection == null)
```
harshgMSFT: Just a thought should we have this null check in base as well?

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/CollectionModelBinder.cs
```diff
@@ -133,13 +165,49 @@ public class CollectionModelBinder<TElement> : IModelBinder
             return boundCollection;
         }
 
-        // Extensibility point that allows the bound collection to be manipulated or transformed before
-        // being returned from the binder.
+        /// <summary>
+        /// Gets an <see cref="object"/> suitable for the associated property.
```
harshgMSFT: sorry don't understand what makes the object suitable?

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/ArrayModelBinder.cs
dougbu: why?  base just returns `newCollection`

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/ArrayModelBinder.cs
dougbu: `var` is fine.  was just sloppy here and the other places you noted.

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/ArrayModelBinder.cs
dougbu: will be `null` when form data includes a prefix match but value provider returns `null`.  not a common case but can happen.

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/ArrayModelBinder.cs
dougbu: MVC 5 seems to do nothing with arrays passed into `TryUpdateModel`.  may have similar issues with dictionaries as well.  but MVC 5 also doesn't have the equivalent of `ArrayModelBinder` and so can't avoid the problem as easily.

in any case, providing as many elements as we can is an improvement over the silent failures in MVC 5.  an error would be incompatible.

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/CollectionModelBinder.cs
dougbu: will reword to make it clear new value should be assignable to the current `ModelExplorer.ModelType`

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/ArrayModelBinder.cs
harshgMSFT: wouldnt we have a problem with Debug.Assert @ line 44 then?

dougbu: I misread your question and answered as if you were talking about `sourceCollection`.  `targetArray` should never be `null` which is why there's a `Debug.Assert()`.

harshgMSFT: should we drop the check then?

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/ArrayModelBinder.cs
harshgMSFT: on second thoughts yeah not needed.

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/ArrayModelBinder.cs
```diff
@@ -19,9 +26,42 @@ public override Task<ModelBindingResult> BindModelAsync(ModelBindingContext bind
...
+            {
+                return null;
+            }
+
             return newCollection.ToArray();
```
harshgMSFT: could be accomplished in one line 
`newCollection?.ToArray();`

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/ArrayModelBinder.cs
harshgMSFT: :+1:

harshgMSFT:
You probably want to throw in some functional tests which bind collections on controller property. 
To enable this scenario I think you would need to update the line
https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs#L73

dougbu:
@rynowak @harshgMSFT updated

might add a few more unit tests in `ControllerActionArgumentBinderTests` tomorrow morning.  mainly to cover read-only properties that will be ignored, binding more than one controller property, and successful bindings after unsuccessful ones (previous implementation gave up activating on first failure).

src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
```diff
@@ -63,6 +66,21 @@ public class DefaultControllerActionArgumentBinder : IControllerActionArgumentBi
...
+            var sourceCollection = source as IEnumerable<TElement>;
+            if (sourceCollection != null && !targetCollection.IsReadOnly)
+            {
+                targetCollection.Clear();
+                foreach (var item in sourceCollection)
+                {
+                    targetCollection.Add(item);
+                }
+            }
+        }
```
rynowak: One thing that always seems more confusing than necessary when looking through this code is that it just seems excessively generic. Any of the built-in collection types will implement `ICollection` and `IEnumerable` (non-generic), and it's really recommended that user-defined collections do the same.

I'm wondering if we can just simplify our lives by *un-generic-ing* all of the model binding code.

I understand if you want to leave this code as-is for consistency with the legacy, but we should think about simplifying this once we have confidence in our set of tests.

src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
```diff
@@ -19,6 +19,9 @@ namespace Microsoft.AspNet.Mvc
     /// </summary>
     public class DefaultControllerActionArgumentBinder : IControllerActionArgumentBinder
     {
+        private static readonly MethodInfo CallPropertyAddRangeOpenGenericMethod =
+            typeof(MutableObjectModelBinder).GetTypeInfo().GetDeclaredMethod(nameof(CallPropertyAddRange));
```
rynowak: Seems to have some confusion about where this method is defined - did it get copy-pasted?

src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
```diff
@@ -70,13 +88,48 @@ private void ActivateProperties(object controller, Type containerType, Dictionar
...
+                var source = property.Value;
+                var target = propertyHelper.GetValue(controller);
+                if (source == null || target == null)
+                {
+                    // Nothing to do when source or target is null.
+                    continue;
+                }
+
+                // Determine T if this is an ICollection<T> property.
+                var elementTypeArray = propertyType
```
rynowak: nit: `elementTypeArray` gives me the impression that this is the element type **of** an array. Suggestion: `collectionTypeArguments`

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/ArrayModelBinder.cs
```diff
@@ -2,16 +2,23 @@
...
+    /// </summary>
+    /// <typeparam name="TElement">Type of elements in the array.</typeparam>
     public class ArrayModelBinder<TElement> : CollectionModelBinder<TElement>
     {
-        public override Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindingContext)
+        /// <inheritdoc />
+        public override Task<ModelBindingResult> BindModelAsync([NotNull] ModelBindingContext bindingContext)
         {
-            if (bindingContext.ModelMetadata.IsReadOnly)
+            if (bindingContext.Model == null && bindingContext.ModelMetadata.IsReadOnly)
```
rynowak: I kinda figure that you're right, but maybe it deserves a comment.. I read this code and it seems to me that this should be `||` instead of `&&`
- we don't want to try and update a null model
- we don't want to update a readonly model

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/CollectionModelBinder.cs
```diff
@@ -133,13 +165,50 @@ public class CollectionModelBinder<TElement> : IModelBinder
...
+        /// </returns>
+        /// <remarks>
+        /// Extensibility point that allows the bound collection to be manipulated or transformed before being
+        /// returned from the binder.
+        /// </remarks>
         protected virtual object GetModel(IEnumerable<TElement> newCollection)
         {
+            // Depends on fact BindSimpleCollection() and BindComplexCollection() always return a List<TElement>
+            // instance or null. In addition GenericModelBinder confirms a List<TElement> is assignable to the
+            // property prior to instantiating this binder and subclass binders do not call this method.
```
rynowak: :heart_eyes_cat: - these comments and srsgreat

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/MutableObjectModelBinder.cs
```diff
@@ -516,6 +550,82 @@ internal void ProcessDto(ModelBindingContext bindingContext, ComplexModelDto dto
...
+                    targetCollection.Add(item);
+                }
+            }
+        }
+
+        private static void TryAction(ModelBindingContext bindingContext, ModelBindingResult dtoResult, Action action)
+        {
+            try
+            {
+                action();
```
rynowak: I'm really not a big fan of this pattern because it makes the debugging experience pretty poor.  How would you feel about factoring out the code inside the catch block to a private method instead?

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Binders/ArrayModelBinderTest.cs
```diff
@@ -11,57 +11,102 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Test
...
-                { "someName[1]", "84" }
+                { "someName[1]", "84" },
             };
             var bindingContext = GetBindingContext(valueProvider);
+            var modelState = bindingContext.ModelState;
             var binder = new ArrayModelBinder<int>();
 
             // Act
-            var retVal = await binder.BindModelAsync(bindingContext);
+            var result = await binder.BindModelAsync(bindingContext);
```
rynowak: :100:

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Binders/ArrayModelBinderTest.cs
```diff
@@ -11,57 +11,102 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Test
...
 
             // Assert
-            Assert.NotNull(retVal);
+            Assert.NotNull(result);
 
-            int[] array = retVal.Model as int[];
+            int[] array = result.Model as int[];
```
rynowak: should be `var` with `Assert.IsType` or a non-as-cast

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Binders/ArrayModelBinderTest.cs
```diff
@@ -11,57 +11,102 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Test
...
 
             // Act
-            var bound = await binder.BindModelAsync(bindingContext);
+            var result = await binder.BindModelAsync(bindingContext);
+
+            // Assert
+            Assert.Null(result);
+        }
+
+        [Theory]
```
rynowak: Can you add a comment about what 'fails silently' means in this context? I think I get it, but I know I will have forgotten in a month.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Binders/CollectionModelBinderTest.cs
```diff
@@ -67,33 +69,101 @@ public class CollectionModelBinderTest
...
-            var retVal = await binder.BindModelAsync(bindingContext);
+            var result = await binder.BindModelAsync(bindingContext);
 
             // Assert
-            Assert.Equal(new[] { 42, 100, 200 }, ((List<int>)retVal.Model).ToArray());
+            Assert.NotNull(result);
+            Assert.True(result.IsModelSet);
+            var list = Assert.IsAssignableFrom<IList<int>>(result.Model);
+            Assert.Equal(new[] { 42, 100, 200 }, list.ToArray());
+            Assert.True(modelState.IsValid);
```
rynowak: line-spacing here would help

rynowak:
:watch: comments and test improvements are awesome! :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1:

src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
dougbu: sure

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/ArrayModelBinder.cs
dougbu: actually this is no longer correct because I removed the copy-to-array capability in last iteration.  should have undone the change to this line as well.

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/MutableObjectModelBinder.cs
dougbu: can set breakpoints inside actions now but sure

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Binders/ArrayModelBinderTest.cs
dougbu: sure; I just did a search-n-repl and didn't fix other issues in the changed lines

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Binders/CollectionModelBinderTest.cs
dougbu: :ok:

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Binders/ArrayModelBinderTest.cs
dougbu: :ok:

src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
dougbu: > understand if you want to leave this code as-is for consistency with the legacy

yeah, let's leave refactoring the core model binding classes until after we have lots more tests in place

src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
dougbu: I duplicated the `private static` method from `MutableObjectModelBinder` to here.  different projects and it's only a few lines.

rynowak: But your field still refers to `MutableObjectModelBinder.CallPropertyAddRange` :laughing:

dougbu: ah; those Amazing Powers of Reflection :smiley_cat:

dougbu:
@rynowak @harshgMSFT updated: rebased, addressed comments, added tests, and more

src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
```diff
@@ -98,24 +100,74 @@ public class DefaultControllerActionArgumentBinder : IControllerActionArgumentBi
...
             foreach (var property in properties)
             {
-                var propertyHelper = propertyHelpers.First(helper =>
-                    string.Equals(helper.Name, property.Key, StringComparison.Ordinal));
-                if (propertyHelper.Property == null || !propertyHelper.Property.CanWrite)
+                var propertyExplorer = containerExplorer.GetExplorerForProperty(property.Key);
+                var propertyMetadata = propertyExplorer?.Metadata;
+                if (propertyMetadata == null)
+                {
+                    // Nothing to do if property is unknown to the ModelMetadataProvider.
```
harshgMSFT: When would this be the case?

dougbu: removed while going back to `PropertyHelper` because the `PropertyHelper` constructor has `[NotNull]` for the `PropertyInfo`

dougbu:
@harshgMSFT @rynowak updated to _not_ use data annotations or the runtime type of controller properties

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/MutableObjectModelBinder.cs
```diff
@@ -496,6 +535,82 @@ internal void ProcessDto(ModelBindingContext bindingContext, ComplexModelDto dto
             }
         }
 
+        // Neither [DefaultValue] nor [Required] is relevant for a non-settable collection.
+        private void AddToProperty(
+            ModelBindingContext bindingContext,
+            ModelExplorer modelExplorer,
```
rynowak: Note that this code is also using model explorer - that's fine from the POV of using `GetMetadataForProperty`, but it will also use the runtime type if you try to get it's properties. 

From the usage it doesn't seem to cause problems, but we might want to clobber it if we refactor this code.

rynowak:
:shipit:

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/MutableObjectModelBinder.cs
dougbu: suggest filing a separate bug covering the use of `ModelExplorer` and both `GetExplorerForExpression()` and `GetExplorerForProperty()` in this class.  looks like this has existed pretty much forever but is more obvious since the `ModelMetadata` / `ModelExplorer` split.  note we also require a `ModelExplorer` instance to create ` `ModelValidationContext`

rynowak: #2438

test/Microsoft.AspNet.Mvc.Core.Test/ParameterBinding/ControllerActionArgumentBinderTests.cs
```diff
@@ -325,8 +289,8 @@ public void ParameterHasFieldPrefix([Bind(Prefix = "simpleModelPrefix")] string
             actionDescriptor.BoundProperties.Add(
                 new ParameterDescriptor
                 {
-                    Name = "ValueBinderMarkedProperty",
-                        BindingInfo = new BindingInfo(),
+                    Name = nameof(TestController.ValueBinderMarkedProperty),
```
harshgMSFT: :heart_eyes:

test/Microsoft.AspNet.Mvc.Core.Test/ParameterBinding/ControllerActionArgumentBinderTests.cs
```diff
@@ -18,39 +16,6 @@ namespace Microsoft.AspNet.Mvc.Core.Test
 {
     public class ControllerActionArgumentBinderTests
     {
-        public class MySimpleModel
```
harshgMSFT: were these not being used?

test/Microsoft.AspNet.Mvc.Core.Test/ParameterBinding/ControllerActionArgumentBinderTests.cs
harshgMSFT: Can please do a favour and do a similar change for other tests in this class?

test/Microsoft.AspNet.Mvc.Core.Test/ParameterBinding/ControllerActionArgumentBinderTests.cs
dougbu: yes

test/Microsoft.AspNet.Mvc.Core.Test/ParameterBinding/ControllerActionArgumentBinderTests.cs
dougbu: :ok:

test/Microsoft.AspNet.Mvc.Core.Test/ParameterBinding/ControllerActionArgumentBinderTests.cs
```diff
@@ -340,6 +304,35 @@ public void ParameterHasFieldPrefix([Bind(Prefix = "simpleModelPrefix")] string
 
             // Assert
             Assert.Equal("Hello", controller.ValueBinderMarkedProperty);
+            Assert.Equal(new List<string> { "goodbye" }, controller.CollectionValueBinderMarkedProperty);
```
harshgMSFT: Is this just to verify that default values are persisted?

test/Microsoft.AspNet.Mvc.Core.Test/ParameterBinding/ControllerActionArgumentBinderTests.cs
```diff
@@ -340,6 +304,35 @@ public void ParameterHasFieldPrefix([Bind(Prefix = "simpleModelPrefix")] string
...
+            Assert.Null(controller.UnmarkedProperty);
```
harshgMSFT: For all properties where we are not attempting to bind like `CollectionValueBinderMarkedProperty` and `UnmarkedProperty` I am wondering should we just remove these validation to simply and focus the test?

test/Microsoft.AspNet.Mvc.Core.Test/ParameterBinding/ControllerActionArgumentBinderTests.cs
```diff
@@ -470,8 +610,13 @@ private class TestController
         {
             public string UnmarkedProperty { get; set; }
 
-            [NonValueProviderBinderMetadata]
-            public string NonValueBinderMarkedProperty { get; set; }
+            public string[] ArrayProperty { get; } = new string[] { "goodbye" };
+
+            public ICollection<string> CollectionValueBinderMarkedProperty { get; } = new List<string> { "goodbye" };
```
harshgMSFT: These properties without any attribute makes it inconsistent with the way this class was written. 
Now there are few properties with attributes and rest of them without attributes. 
Suggest remove the attributes from the properties of this class as they are not discovered based on attributes but are directly added to the `BoundProperties` in individual tests, and add a comment on `TestController` mentioning the reason why none of the properties need attributes.

test/Microsoft.AspNet.Mvc.Core.Test/ParameterBinding/ControllerActionArgumentBinderTests.cs
dougbu: not overwritten, yes

test/Microsoft.AspNet.Mvc.Core.Test/ParameterBinding/ControllerActionArgumentBinderTests.cs
dougbu: no, leaving as-is

test/Microsoft.AspNet.Mvc.Core.Test/ParameterBinding/ControllerActionArgumentBinderTests.cs
dougbu: don't care much but it's true existing `[ValueProviderMetadata]` isn't relevant in this test class (anymore?)

dougbu:
@harshgMSFT updated that test class...

harshgMSFT:
:shipit:

dougbu:
3fd4991

Issue #2407, opened 2015-04-20
Add back the implicit [Required] for value types


rynowak:
MVC 5 had a behavior where a value-type was implicitly defined with `ModelMetadata.IsRequired == true` - additionally when a property had `ModelMetadata.IsRequired == true` then a `[Required]` attribute was implicitly added as a validator. 

Since this behavior is applied to value types, the only observable effect is on the HTML generation of `data-*` attributes for value type properties. (`ModelMetadata` also contains the setting `ModelMetadata.IsBindingRequired` which effects server-side validation)

We've currently restored the behavior of setting `ModelMetadata.IsRequired == true` by-default for value types, but we need to add back the validator in the `DataAnnotationsModelValidatorProvider` and  add tests to verify the HTML generation.

Doing this will also require removing the behavior from `MutableObjectModelBinder` where a property with a *required validator* will behave the same as a property marked `[BindRequired]`. This will likely have some fallout for tests that rely on the non-MVC5-compatible behavior. See: https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.ModelBinding/Binders/MutableObjectModelBinder.cs#L337

rynowak:
See also #1712 which deals with `BindingBehaviorAttribute`.

rynowak:
Part of this is  fa56df9

rynowak:
Part 2: b64fd7ae39e27637ece40257dfdc5a3873299506

PR #2413, opened 2015-04-20
Fix #1712 - remove reflection from validation code in MutableObjectModel...

### LINKED ISSUE #1712: MutableObjectModelBinder.GetPropertyValidationInfo should not reflect over `ModelType.Properties`

rynowak:
...Binder

This change moves [BindingBehavior(...)] and friends into the model
metadata layer, and removes the reflection code from
MutableObjectModelBinder that was looking for them previously.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.ModelBinding/Binders/MutableObjectModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.ModelBinding/Binders/MutableObjectModelBinder.cs
@@ -8,7 +8,6 @@ using System.Linq;
 using System.Reflection;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Mvc.ModelBinding.Internal;
-using Microsoft.AspNet.Mvc.ModelBinding.Metadata;
 using Microsoft.AspNet.Mvc.ModelBinding.Validation;
 
 namespace Microsoft.AspNet.Mvc.ModelBinding
@@ -313,17 +312,21 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         internal static PropertyValidationInfo GetPropertyValidationInfo(ModelBindingContext bindingContext)
         {
             var validationInfo = new PropertyValidationInfo();
-            var modelTypeInfo = bindingContext.ModelType.GetTypeInfo();
-            var typeAttribute = modelTypeInfo.GetCustomAttribute<BindingBehaviorAttribute>();
-            var properties = bindingContext.ModelType.GetProperties(BindingFlags.Public | BindingFlags.Instance);
-            foreach (var property in properties)
+
+            foreach (var propertyMetadata in bindingContext.ModelMetadata.Properties)
             {
-                var propertyName = property.Name;
-                var propertyMetadata = bindingContext.ModelMetadata.Properties[propertyName];
-                if (propertyMetadata == null)
+                var propertyName = propertyMetadata.PropertyName;
+
+                if (!propertyMetadata.IsBindingAllowed)
                 {
-                    // Skip indexer properties and others ModelMetadata ignores.
+                    // Nothing to do here if binding is not allowed.
+                    validationInfo.SkipProperties.Add(propertyName);
                     continue;
+                }
+
+                if (propertyMetadata.IsBindingRequired)
+                {
+                    validationInfo.RequiredProperties.Add(propertyName);
                 }
 
                 var validatorProviderContext = new ModelValidatorProviderContext(propertyMetadata);
@@ -334,29 +337,6 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
                 if (requiredValidator != null)
                 {
                     validationInfo.RequiredValidators[propertyName] = requiredValidator;
-                }
-
-                var propertyAttribute = property.GetCustomAttribute<BindingBehaviorAttribute>();
-                var bindingBehaviorAttribute = propertyAttribute ?? typeAttribute;
-                if (bindingBehaviorAttribute != null)
-                {
-                    switch (bindingBehaviorAttribute.Behavior)
-                    {
-                        case BindingBehavior.Required:
-                            validationInfo.RequiredProperties.Add(propertyName);
-                            break;
-
-                        case BindingBehavior.Never:
-                            validationInfo.SkipProperties.Add(propertyName);
-                            break;
-                    }
-                }
-                else if (requiredValidator != null)
-                {
-                    validationInfo.RequiredProperties.Add(propertyName);
-                }
-                else if (propertyMetadata.IsBindingRequired)
-                {
                     validationInfo.RequiredProperties.Add(propertyName);
                 }
             }
--- a/src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/BindingMetadata.cs
+++ b/src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/BindingMetadata.cs
@@ -29,6 +29,13 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
         public Type BinderType { get; set; }
 
         /// <summary>
+        /// Gets or sets a value indicating whether or not the property can be model bound.
+        /// Will be ignored if the model metadata being created is not a property.
+        /// See <see cref="ModelMetadata.IsBindingAllowed"/>.
+        /// </summary>
+        public bool IsBindingAllowed { get; set; } = true;
+
+        /// <summary>
         /// Gets or sets a value indicating whether or not the request must contain a value for the model.
         /// Will be ignored if the model metadata being created is not a property.
         /// See <see cref="ModelMetadata.IsBindingRequired"/>.
--- a/src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/BindingMetadata.cs
+++ b/src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/BindingMetadata.cs
@@ -30,14 +30,14 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
 
         /// <summary>
         /// Gets or sets a value indicating whether or not the property can be model bound.
-        /// Will be ignored if the model metadata being created is not a property.
+        /// Will be ignored if the model metadata being created does not represent a property.
         /// See <see cref="ModelMetadata.IsBindingAllowed"/>.
         /// </summary>
         public bool IsBindingAllowed { get; set; } = true;
 
         /// <summary>
         /// Gets or sets a value indicating whether or not the request must contain a value for the model.
-        /// Will be ignored if the model metadata being created is not a property.
+        /// Will be ignored if the model metadata being created does not represent a property.
         /// See <see cref="ModelMetadata.IsBindingRequired"/>.
         /// </summary>
         public bool IsBindingRequired { get; set; }
--- a/src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultBindingMetadataProvider.cs
+++ b/src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultBindingMetadataProvider.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
 using Microsoft.Framework.Internal;
 
 namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
@@ -52,6 +53,21 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
             {
                 context.BindingMetadata.PropertyBindingPredicateProvider = new CompositePredicateProvider(
                     predicateProviders);
+            }
+
+            if (context.Key.MetadataKind == ModelMetadataKind.Property)
+            {
+                // BindingBehavior can fall back to attributes on the Container Type, but we should ignore
+                // attributes on the Property Type.
+                var bindingBehavior =
+                    context.PropertyAttributes.OfType<BindingBehaviorAttribute>().FirstOrDefault() ??
+                    context.Key.ContainerType.GetTypeInfo().GetCustomAttribute<BindingBehaviorAttribute>(inherit: true);
+
+                if (bindingBehavior != null)
+                {
+                    context.BindingMetadata.IsBindingAllowed = bindingBehavior.Behavior != BindingBehavior.Never;
+                    context.BindingMetadata.IsBindingRequired = bindingBehavior.Behavior == BindingBehavior.Required;
+                }
             }
         }
 
--- a/src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadata.cs
+++ b/src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadata.cs
@@ -256,11 +256,34 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
         }
 
         /// <inheritdoc />
+        public override bool IsBindingAllowed
+        {
+            get
+            {
+                if (MetadataKind == ModelMetadataKind.Property)
+                {
+                    return BindingMetadata.IsBindingAllowed;
+                }
+                else
+                {
+                    return true;
+                }
+            }
+        }
+
+        /// <inheritdoc />
         public override bool IsBindingRequired
         {
             get
             {
-                return BindingMetadata.IsBindingRequired;
+                if (MetadataKind == ModelMetadataKind.Property)
+                {
+                    return BindingMetadata.IsBindingRequired;
+                }
+                else
+                {
+                    return false;
+                }
             }
         }
 
--- a/src/Microsoft.AspNet.Mvc.ModelBinding/ModelMetadata.cs
+++ b/src/Microsoft.AspNet.Mvc.ModelBinding/ModelMetadata.cs
@@ -167,7 +167,17 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         public abstract bool HideSurroundingHtml { get; }
 
         /// <summary>
-        /// Gets a value indicating whether or not the model value is required by model binding. This is only
+        /// Gets a value indicating whether or not the model value can be bound by model-binding. This is only
+        /// applicable when the current instance represents a property.
+        /// </summary>
+        /// <remarks>
+        /// If <c>true</c> then the model value is considered supported by model-binding and can be set
+        /// based on provided input in the request.
+        /// </remarks>
+        public abstract bool IsBindingAllowed { get; }
+
+        /// <summary>
+        /// Gets a value indicating whether or not the model value is required by model-binding. This is only
         /// applicable when the current instance represents a property.
         /// </summary>
         /// <remarks>
--- a/src/Microsoft.AspNet.Mvc.ModelBinding/ModelMetadata.cs
+++ b/src/Microsoft.AspNet.Mvc.ModelBinding/ModelMetadata.cs
@@ -167,21 +167,21 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         public abstract bool HideSurroundingHtml { get; }
 
         /// <summary>
-        /// Gets a value indicating whether or not the model value can be bound by model-binding. This is only
+        /// Gets a value indicating whether or not the model value can be bound by model binding. This is only
         /// applicable when the current instance represents a property.
         /// </summary>
         /// <remarks>
-        /// If <c>true</c> then the model value is considered supported by model-binding and can be set
+        /// If <c>true</c> then the model value is considered supported by model binding and can be set
         /// based on provided input in the request.
         /// </remarks>
         public abstract bool IsBindingAllowed { get; }
 
         /// <summary>
-        /// Gets a value indicating whether or not the model value is required by model-binding. This is only
+        /// Gets a value indicating whether or not the model value is required by model binding. This is only
         /// applicable when the current instance represents a property.
         /// </summary>
         /// <remarks>
-        /// If <c>true</c> then the model value is considered required by model-binding and must have a value
+        /// If <c>true</c> then the model value is considered required by model binding and must have a value
         /// supplied in the request to be considered valid.
         /// </remarks>
         public abstract bool IsBindingRequired { get; }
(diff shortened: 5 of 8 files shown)


src/Microsoft.AspNet.Mvc.ModelBinding/Binders/MutableObjectModelBinder.cs
```diff
@@ -334,29 +337,6 @@ internal static PropertyValidationInfo GetPropertyValidationInfo(ModelBindingCon
                 if (requiredValidator != null)
```
dougbu: just to be clear: this change leaves `[Required]` also implicitly meaning `BindingBehavior.Required`?

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/BindingMetadata.cs
```diff
@@ -29,6 +29,13 @@ public class BindingMetadata
         public Type BinderType { get; set; }
 
         /// <summary>
+        /// Gets or sets a value indicating whether or not the property can be model bound.
+        /// Will be ignored if the model metadata being created is not a property.
```
dougbu: nit: sentence doesn't scan.  think you mean `Will be ignored if the current <see cref="ModelMetadata"/> is not for a property.` or perhaps `Will be ignored unless creating a model for a property.`

src/Microsoft.AspNet.Mvc.ModelBinding/ModelMetadata.cs
```diff
@@ -167,7 +167,17 @@ public string PropertyName
...
+        /// applicable when the current instance represents a property.
+        /// </summary>
+        /// <remarks>
+        /// If <c>true</c> then the model value is considered supported by model-binding and can be set
+        /// based on provided input in the request.
+        /// </remarks>
+        public abstract bool IsBindingAllowed { get; }
+
+        /// <summary>
+        /// Gets a value indicating whether or not the model value is required by model-binding. This is only
```
dougbu: the "model binding" to "model-binding" switch doesn't make sense to me.  see a few mentions of "model-binding" in existing doc comments but far more using "model binding".  the majority is correct in this case.

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadata.cs
```diff
@@ -256,11 +256,34 @@ public override bool HtmlEncode
         }
 
         /// <inheritdoc />
+        public override bool IsBindingAllowed
+        {
+            get
+            {
+                if (MetadataKind == ModelMetadataKind.Property)
```
dougbu: these checks look redundant and cause the property to have defaults both here and in the `BindingMetadata` class.  is this done to avoid filling in the `BindingMetadata` property until and unless it's absolutely required?

dougbu:
:watch: and haven't looked in detail at the tests yet

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadata.cs
rynowak: It's defense in depth. I fixed untracked bugs in this exact logic for a similar property in the last PR. It's far easier to just write the correct code in both places.

src/Microsoft.AspNet.Mvc.ModelBinding/ModelMetadata.cs
rynowak: sure. I think I felt like using *model-binding* one day and started doing it :+1: I'll clean it up.

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/MutableObjectModelBinder.cs
rynowak: Yes. I logged a separate bug for that item. https://github.com/aspnet/Mvc/issues/2407 at this point it's getting hard to track what the impact of #1712 is supposed to be, so it felt like another issue was warranted.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/ModelMetadataProviderTest.cs
```diff
@@ -712,7 +712,7 @@ public void PropertySetter_Null_ForPrivateSetProperty()
             Assert.Null(metadata.PropertySetter);
             Assert.NotNull(metadata.PropertyGetter);
         }
-
+
```
harshgMSFT: nit remove.

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/MutableObjectModelBinder.cs
harshgMSFT: to be clear: the issue mentioned in #2407 was there even before this PR.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DefaultBindingMetadataProviderTest.cs
```diff
@@ -147,5 +147,358 @@ public void GetBindingDetails_FindsBindingSource_IfNullFallsBack()
...
+            Assert.True(context.BindingMetadata.IsBindingRequired);
+        }
+
+        [Fact]
+        public void GetBindingDetails_FindsBindRequired_OnProperty()
+        {
+            // Arrange
+            var propertyAttributes = new object[]
+            {
+                new BindingBehaviorAttribute(BindingBehavior.Required),
```
harshgMSFT: Shouldnt this be 
`new BindRequiredAttribute();`

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultBindingMetadataProvider.cs
```diff
@@ -53,6 +54,21 @@ public void GetBindingMetadata([NotNull] BindingMetadataProviderContext context)
...
                     predicateProviders);
             }
+
+            if (context.Key.MetadataKind == ModelMetadataKind.Property)
+            {
+                // BindingBehavior can fall back to attributes on the Container Type, but we should ignore
+                // attributes on the Property Type.
+                var bindingBehavior =
+                    context.PropertyAttributes.OfType<BindingBehaviorAttribute>().FirstOrDefault() ??
+                    context.Key.ContainerType.GetTypeInfo().GetCustomAttribute<BindingBehaviorAttribute>(inherit: true);
```
harshgMSFT: Isn't this an additional behavior we are defining?
Can we do it separately? 
That will help us consider the alternatives like
Shouldn't the attribute on the type containing the property apply to whether a model of that type gets model bound or not?

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DefaultBindingMetadataProviderTest.cs
```diff
@@ -147,5 +147,358 @@ public void GetBindingDetails_FindsBindingSource_IfNullFallsBack()
...
+            // Assert
+            Assert.True(context.BindingMetadata.IsBindingAllowed);
+            Assert.True(context.BindingMetadata.IsBindingRequired);
+        }
+
+        [Fact]
+        public void GetBindingDetails_FindsBindRequired_OnContainerClass()
+        {
+            // Arrange
+            var propertyAttributes = new object[]
```
harshgMSFT: nit: you could inline this as well

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DefaultBindingMetadataProviderTest.cs
```diff
@@ -147,5 +147,358 @@ public void GetBindingDetails_FindsBindingSource_IfNullFallsBack()
...
+            {
+            };
+
+            var context = new BindingMetadataProviderContext(
+                ModelMetadataIdentity.ForProperty(typeof(string), "Property", typeof(BindRequiredOnClass)),
+                new ModelAttributes(propertyAttributes, typeAttributes: new object[0]));
```
harshgMSFT: does this pass? you are passing in an empty typeAttribute array?

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DefaultBindingMetadataProviderTest.cs
```diff
@@ -147,5 +147,358 @@ public void GetBindingDetails_FindsBindingSource_IfNullFallsBack()
...
+        public void GetBindingDetails_FindsBindNever_OnContainerClass()
+        {
+            // Arrange
+            var propertyAttributes = new object[]
+            {
+            };
+
+            var context = new BindingMetadataProviderContext(
+                ModelMetadataIdentity.ForProperty(typeof(string), "Property", typeof(BindNeverOnClass)),
+                new ModelAttributes(propertyAttributes, typeAttributes: new object[0]));
```
harshgMSFT: does this pass?

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultBindingMetadataProvider.cs
rynowak: @harshgMSFT this is the current behavior of this attribute, nothing has changed. I added functional tests for this in a previous PR.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DefaultBindingMetadataProviderTest.cs
rynowak: Yes. Please read the comments in the relevant code, BindingBehavior looks at the container type not the property type.

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultBindingMetadataProvider.cs
harshgMSFT: sorry totally missed https://github.com/aspnet/Mvc/pull/2413/files#diff-62a5557f611407fe0ee3bd9e9e2a1393L340

src/Microsoft.AspNet.Mvc.ModelBinding/Binders/MutableObjectModelBinder.cs
rynowak: Yes.

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/BindingMetadata.cs
rynowak: changed it to `Will be ignored if the model metadata being created does not represent a property`

harshgMSFT:
:shipit: after fixing minor issue in test.

src/Microsoft.AspNet.Mvc.ModelBinding/ModelMetadata.cs
rynowak: changed all occurrences in this class to `model binding`

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DefaultBindingMetadataProviderTest.cs
rynowak: :+1:

rynowak:
@dougbu @harshgMSFT updated

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DefaultBindingMetadataProviderTest.cs
dougbu: isn't `[BindRequired]` tested in the next test?  this one is about `BindingBehavior.Required`

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DefaultBindingMetadataProviderTest.cs
```diff
@@ -147,5 +147,350 @@ public void GetBindingDetails_FindsBindingSource_IfNullFallsBack()
...
+            Assert.True(context.BindingMetadata.IsBindingRequired);
+        }
+
+        [Fact]
+        public void GetBindingDetails_FindsBindRequired_OnContainerClass()
+        {
+            // Arrange
+            var context = new BindingMetadataProviderContext(
+                ModelMetadataIdentity.ForProperty(typeof(string), "Property", typeof(BindRequiredOnClass)),
+                new ModelAttributes(propertyAttributes: new object[0], typeAttributes: new object[0]));
```
dougbu: hmm, should `ModelAttributes` contain the attributes on the container class as well?  this test makes it very clear that there's a third category of attributes and that `ModelAttributes` ignores this category.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DefaultBindingMetadataProviderTest.cs
rynowak: Yep, if you look at the tests **now** they do the right thing. I'm not sure why github didn't hide the original comment, because I did make a change here.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DefaultBindingMetadataProviderTest.cs
```diff
@@ -147,5 +147,350 @@ public void GetBindingDetails_FindsBindingSource_IfNullFallsBack()
...
+            // Act
+            provider.GetBindingMetadata(context);
+
+            // Assert
+            Assert.False(context.BindingMetadata.IsBindingAllowed);
+            Assert.False(context.BindingMetadata.IsBindingRequired);
+        }
+
+        [Fact]
+        public void GetBindingDetails_OverrideBehaviorOnClass_OverrideWithOptional()
```
dougbu: nit `..._BehaviorOnClass_OverriddenWithOptional` and similar for a couple of tests below.  hard to parse name and understand what is overriding what.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DefaultBindingMetadataProviderTest.cs
```diff
@@ -147,5 +147,350 @@ public void GetBindingDetails_FindsBindingSource_IfNullFallsBack()
...
+        public void GetBindingDetails_OverrideInheritedBehaviorOnClass_OverrideWithRequired()
+        {
+            // Arrange
+            var propertyAttributes = new object[]
+            {
+                new BindRequiredAttribute()
+            };
+
+            var context = new BindingMetadataProviderContext(
+                ModelMetadataIdentity.ForProperty(typeof(string), "Property", typeof(InheritedBindNeverOnClass)),
```
dougbu: would like to see another test using `InheritedBindNeverOnClass` but without an override on the property

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DefaultBindingMetadataProviderTest.cs
```diff
@@ -147,5 +147,350 @@ public void GetBindingDetails_FindsBindingSource_IfNullFallsBack()
...
+
+        private class InheritedBindNeverOnClass : BindNeverOnClass
+        {
+        }
+
+        [BindRequired]
+        private class BindRequiredOnClass
+        {
+            public string Property { get; set; }
+        }
```
dougbu: aren't testing precedence of multiple attributes in class hierarchy. would be good to have a test or two using a class like
``` c#
[BindNever]
class BindNeverOnClassTotallyWins : BindRequiredOnClass
{
}
```

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/ModelMetadataProviderTest.cs
```diff
@@ -869,6 +948,21 @@ private class RequiredMember
...
+            [BindRequired]
+            public int BindRequiredOverride { get; set; }
+
+            [BindingBehavior(BindingBehavior.Optional)]
+            public int BindingBehaviorOptionalOverride { get; set; }
+
+            [BindingBehavior(BindingBehavior.Never)]
+            public int BindingBehaviorNever { get; set; }
+
+            public int BindingBehaviorNeverSetOnClass { get; set; }
```
dougbu: `BindingBehaviorInheritedFromClass`

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/ModelMetadataProviderTest.cs
```diff
@@ -869,6 +948,21 @@ private class RequiredMember
...
+        private class BindingBehaviorModel
+        {
+            [BindRequired]
+            public int BindRequiredOverride { get; set; }
+
+            [BindingBehavior(BindingBehavior.Optional)]
+            public int BindingBehaviorOptionalOverride { get; set; }
+
+            [BindingBehavior(BindingBehavior.Never)]
+            public int BindingBehaviorNever { get; set; }
```
dougbu: `BindingBehaviorDuplicatedFromClass`

dougbu: or `BindingBehaviorNeverDuplicatedFromClass`

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/ModelMetadataProviderTest.cs
```diff
@@ -726,6 +726,85 @@ public void Metadata_Null_ForPrivateGetProperty()
...
+            // Act
+            var metadata = metadataProvider.GetMetadataForProperty(type, propertyName);
+
+            // Assert
+            Assert.True(metadata.IsBindingAllowed);
+            Assert.True(metadata.IsBindingRequired);
+        }
+
+        [Fact]
+        public void BindingBehavior_Override_OptionalOnProperty()
```
dougbu: nit: prefer naming scheme used below to what you're using for these two `_Override_` tests.  for this one how 'bout `BindingBehavior_Optional_SetOnPropertyOverridingClass`?

src/Microsoft.AspNet.Mvc.ModelBinding/Metadata/DefaultModelMetadata.cs
dougbu: :ok: though I remain a bit concerned about providing default values twice.

dougbu:
:watch: mostly for test gaps and potential `ModelAttributes` extension

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/ModelMetadataProviderTest.cs
rynowak: sure.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DefaultBindingMetadataProviderTest.cs
rynowak: I don't really like the idea because it's only used in a 2 very specialized places.  If we end up doing https://github.com/aspnet/Mvc/issues/2426 then I think we should do it.

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DefaultBindingMetadataProviderTest.cs
rynowak: this just seems different not better, don't care

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DefaultBindingMetadataProviderTest.cs
rynowak: :+1:

rynowak: Added `GetBindingDetails_FindsBindNever_OnBaseClass`

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DefaultBindingMetadataProviderTest.cs
dougbu: :ok: fine as-is then

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/DefaultBindingMetadataProviderTest.cs
rynowak: Sure, added `GetBindingDetails_OverrideBehaviorOnBaseClass_OverrideWithRequired_OnClass`

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/ModelMetadataProviderTest.cs
rynowak: I'm avoiding using the term inherited for anything other than class -> class inheritance.

dougbu: inferred? FallsBackTo? :smiley_cat:

test/Microsoft.AspNet.Mvc.ModelBinding.Test/Metadata/ModelMetadataProviderTest.cs
rynowak: don't see the improvement sorry

rynowak:
@dougbu updated - added tests

dougbu:
:shipit:

PR #2415, opened 2015-04-20
[JsonPatch]: Removed Test operation code and changed T to TModel

kirthik:
@Eilon, @rynowak - Could you please take a look?

### DIFF
--- a/src/Microsoft.AspNet.JsonPatch/Adapters/IObjectAdapter.cs
+++ b/src/Microsoft.AspNet.JsonPatch/Adapters/IObjectAdapter.cs
@@ -16,6 +16,5 @@ namespace Microsoft.AspNet.JsonPatch.Adapters
         void Move(Operation<TModel> operation, TModel objectToApplyTo);
         void Remove(Operation<TModel> operation, TModel objectToApplyTo);
         void Replace(Operation<TModel> operation, TModel objectToApplyTo);
-        void Test(Operation<TModel> operation, TModel objectToApplyTo);
     }
 }
--- a/src/Microsoft.AspNet.JsonPatch/Adapters/IObjectAdapter.cs
+++ b/src/Microsoft.AspNet.JsonPatch/Adapters/IObjectAdapter.cs
@@ -8,14 +8,14 @@ namespace Microsoft.AspNet.JsonPatch.Adapters
     /// <summary>
     /// Defines the operations that can be performed on a JSON patch document.
     /// </summary>
-    /// <typeparam name="T"></typeparam>
-    public interface IObjectAdapter<T> where T : class
+    /// <typeparam name="TModel">The type of the model.</typeparam>
+    public interface IObjectAdapter<TModel> where TModel : class
     {
-        void Add(Operation<T> operation, T objectToApplyTo);
-        void Copy(Operation<T> operation, T objectToApplyTo);
-        void Move(Operation<T> operation, T objectToApplyTo);
-        void Remove(Operation<T> operation, T objectToApplyTo);
-        void Replace(Operation<T> operation, T objectToApplyTo);
-        void Test(Operation<T> operation, T objectToApplyTo);
+        void Add(Operation<TModel> operation, TModel objectToApplyTo);
+        void Copy(Operation<TModel> operation, TModel objectToApplyTo);
+        void Move(Operation<TModel> operation, TModel objectToApplyTo);
+        void Remove(Operation<TModel> operation, TModel objectToApplyTo);
+        void Replace(Operation<TModel> operation, TModel objectToApplyTo);
+        void Test(Operation<TModel> operation, TModel objectToApplyTo);
     }
 }
--- a/src/Microsoft.AspNet.JsonPatch/Operations/TypedOperation.cs
+++ b/src/Microsoft.AspNet.JsonPatch/Operations/TypedOperation.cs
@@ -5,7 +5,7 @@ using Microsoft.AspNet.JsonPatch.Adapters;
 
 namespace Microsoft.AspNet.JsonPatch.Operations
 {
-    public class Operation<T> : Operation where T : class
+    public class Operation<TModel> : Operation where TModel : class
     {
         public Operation()
         {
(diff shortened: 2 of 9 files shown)


src/Microsoft.AspNet.JsonPatch/Adapters/IObjectAdapter.cs
```diff
@@ -8,14 +8,14 @@ namespace Microsoft.AspNet.JsonPatch.Adapters
...
-        void Move(Operation<T> operation, T objectToApplyTo);
-        void Remove(Operation<T> operation, T objectToApplyTo);
-        void Replace(Operation<T> operation, T objectToApplyTo);
-        void Test(Operation<T> operation, T objectToApplyTo);
+        void Add(Operation<TModel> operation, TModel objectToApplyTo);
+        void Copy(Operation<TModel> operation, TModel objectToApplyTo);
+        void Move(Operation<TModel> operation, TModel objectToApplyTo);
+        void Remove(Operation<TModel> operation, TModel objectToApplyTo);
+        void Replace(Operation<TModel> operation, TModel objectToApplyTo);
+        void Test(Operation<TModel> operation, TModel objectToApplyTo);
```
rynowak: ???

kirthik: I kept `Test` API but just removed the code in `ObjectAdapter` and added an error that says "test operation is not supported"

rynowak: What kind of error will a user see? `ModelState` or 500/Exception?

kirthik: If user calls `ApplyTo(patchDoc, ModelState)` then he will see a `ModelState` error. If he calls `ApplyTo(patchDoc)` then he will see `JsonPatchException`

yishaigalatzer: Should probably see UnsupportedException

kirthik: Sure. I will change `JsonPatchException` to `UnSupportedException`.

src/Microsoft.AspNet.JsonPatch/TypedJsonPatchDocument.cs
```diff
@@ -64,12 +65,18 @@ public JsonPatchDocument<T> Add<TProp>(Expression<Func<T, TProp>> path, TProp va
...
+            IList<TProp>>> path,
+            TProp value,
+            int position)
         {
-            Operations.Add(new Operation<T>(
+            Operations.Add(new Operation<TModel>(
                 "add",
-                ExpressionHelpers.GetPath<T, IList<TProp>>(path).ToLower() + "/" + position,
-                null, value));
+                ExpressionHelpers.GetPath(path).ToLower() + "/" + position,
```
Eilon: `ToLower()` --> `ToLowerInvariant()`. Please check for all occurences.

Eilon: (Can do it in a separate PR if you want.)

Eilon:
:shipit: from me. (Can do my comment in a separate PR so it's not too many changes in this PR.)

src/Microsoft.AspNet.JsonPatch/TypedJsonPatchDocument.cs
kirthik: Yes, I will make this change in a separate PR.

src/Microsoft.AspNet.JsonPatch/Adapters/IObjectAdapter.cs
```diff
@@ -8,14 +8,13 @@ namespace Microsoft.AspNet.JsonPatch.Adapters
     /// <summary>
     /// Defines the operations that can be performed on a JSON patch document.
     /// </summary>
-    /// <typeparam name="T"></typeparam>
-    public interface IObjectAdapter<T> where T : class
+    /// <typeparam name="TModel">The type of the model.</typeparam>
+    public interface IObjectAdapter<TModel> where TModel : class
```
rynowak: This is an improvement, but we should still think about where in code the `TModel` really even makes sense. Just because you have a type doesn't mean you have to be generic (if that makes sense).

rynowak:
Don't undo it now, but my advice in the future would be to send 2 PRs for this. The `T` -> `TModel` change is mechanical and could be reviewed by someone in 30 seconds.

src/Microsoft.AspNet.JsonPatch/Operations/TypedOperation.cs
```diff
@@ -1,11 +1,12 @@
 ﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using Microsoft.AspNet.JsonPatch.Adapters;
 
 namespace Microsoft.AspNet.JsonPatch.Operations
 {
-    public class Operation<T> : Operation where T : class
+    public class Operation<TModel> : Operation where TModel : class
```
rynowak: Another note for cleanup - our guidelines say this file should be called `OperationOfT.cs`

rynowak:
:shipit:

src/Microsoft.AspNet.JsonPatch/Adapters/IObjectAdapter.cs
kirthik: Sure. I will do this exercise in a separate PR.

src/Microsoft.AspNet.JsonPatch/Operations/TypedOperation.cs
kirthik: Will change file names.

kirthik:
Fix checked in - 3042a62ba95c014179dd5604dc10fecfbd4c924a

Issue #2085, opened 2015-02-26
Use custom HtmlEncoder in unit tests rather that default implementation.


ajaybhargavb:
Follow up for #2076 

Consider using a custom Html encoder like this.
```C#
internal class CustomHtmlEncoder : IHtmlEncoder
{
    public string HtmlEncode(string value)
    {
        return "HtmlEncode[[" + value + "]]";
    }

    public void HtmlEncode(string value, int startIndex, int charCount, TextWriter output)
    {
        output.Write("HtmlEncode[[");
        output.Write(value.Substring(startIndex, charCount));
        output.Write("]]");
    }

    public void HtmlEncode(char[] value, int startIndex, int charCount, TextWriter output)
    {
        output.Write("HtmlEncode[[");
        output.Write(value, startIndex, charCount);
        output.Write("]]");
    }
}```

<!---
@huboard:{"order":2277.97265625,"milestone_order":2085,"custom_state":"ready"}
-->

GrabYourPitchforks:
BTW, the HttpAbstractions repo already has this code for you to copy if you wish. See the implementation at https://github.com/aspnet/HttpAbstractions/blob/dev/test/Microsoft.Framework.WebEncoders.Tests/CommonTestEncoder.cs.

PR #2417, opened 2015-04-20
Using custom HtmlEncoder in unit tests - Part 1

ajaybhargavb:
Issue - #2085 
There are changes to rendering tests that are not included in this PR. Will send that out as part 2.
@Eilon @dougbu


Eilon:
:shipit:

test/Microsoft.AspNet.Mvc.Core.Test/Rendering/HtmlHelperFormExtensionsTest.cs
```diff
@@ -276,7 +276,7 @@ public static TheoryData<object> RouteValuesDataSet
         public void BeginFormWithNoParameters_CallsHtmlGeneratorWithExpectedValues()
         {
             // Arrange
-            var tagBuilder = new TagBuilder(tagName: "form", htmlEncoder: new HtmlEncoder());
+            var tagBuilder = new TagBuilder(tagName: "form", htmlEncoder: new NullTestEncoder());
```
dougbu: does anything demonstrate `@Html.BeginForm()` correctly encodes values passed as arguments?

test/Microsoft.AspNet.Mvc.Core.Test/Rendering/TagBuilderTest.cs
```diff
@@ -29,7 +29,7 @@ public class TagBuilderTest
         public void MergeAttribute_IgnoresCase(bool replaceExisting, string expectedKey, string expectedValue)
         {
             // Arrange
-            var tagBuilder = new TagBuilder("p", new HtmlEncoder());
+            var tagBuilder = new TagBuilder("p", new NullTestEncoder());
```
dougbu: `TagBuilder` deals directly with encoding.  is that tested?

test/Microsoft.AspNet.Mvc.TagHelpers.Test/OptionTagHelperTest.cs
```diff
@@ -401,7 +401,14 @@ public static IEnumerable<object[]> DoesNotUseViewContextDataSet
...
                     var tagHelperContent = new DefaultTagHelperContent();
-                    tagHelperContent.SetContent(originalContent);
+                    if (string.IsNullOrEmpty(originalContent))
+                    {
+                        tagHelperContent.SetContent(originalContent);
+                    }
+                    else
+                    {
+                        tagHelperContent.SetContent("HtmlEncode[[" + originalContent + "]]");
+                    }
```
dougbu: why is this doing a fake encoding at this level?  what fails if `originalContent` is assigned without this?

dougbu:
:watch: for better encoding coverage in the test classes that are using just `NullTestEncoder`.

test/Microsoft.AspNet.Mvc.Core.Test/Rendering/HtmlHelperFormExtensionsTest.cs
ajaybhargavb: It is tested here https://github.com/aspnet/Mvc/pull/2420/files#diff-8cd78f293ecf9bb7f3a377fc18b05d39R162

test/Microsoft.AspNet.Mvc.Core.Test/Rendering/TagBuilderTest.cs
ajaybhargavb: Added a test.

ajaybhargavb:
Updated.

src/Microsoft.AspNet.Mvc.TagHelpers/OptionTagHelper.cs
```diff
@@ -78,6 +78,10 @@ public class OptionTagHelper : TagHelper
                     {
                         selected = selectedValues.Contains(Value);
                     }
+                    else if (output.IsContentModified)
+                    {
+                        selected = encodedValues.Contains(output.Content.GetContent());
+                    }
```
dougbu: sorry, this is about where the :dancers: danced

test/Microsoft.AspNet.Mvc.Core.Test/Rendering/TagBuilderTest.cs
dougbu: thx

dougbu:
the following classes do not test using `CommonTestEncoder` at all.  please fill these gaps.
- `AnchorTagHelperTest`
- `FormTagHelperTest`
- `TagHelperOutputExtensionsTest`
- `ValidationMessageTagHelperTest`

test/Microsoft.AspNet.Mvc.Core.Test/Rendering/HtmlHelperFormExtensionsTest.cs
dougbu: thx

dougbu:
:watch:

ajaybhargavb:
@dougbu Not sure what scenario actually uses encoding in these helpers. I couldn't find any part of the code that uses with encoding directly. Could you provide some scenarios? 

> the following classes do not test using CommonTestEncoder at all. please fill these gaps.
>
>AnchorTagHelperTest
>FormTagHelperTest
>TagHelperOutputExtensionsTest
>ValidationMessageTagHelperTest

dougbu:
@ajaybhargavb the named tag helpers all get `TagBuilder` instances from `DefaultHtmlGenerator`.  they then at least call `MergeAttributes.  so the basic thing is to ensure extra encoding does not occur.  for example, use the `CommonTestEncoder` instead in tests that just check `output.Attributes` (and see nothing is encoded).  or, in tests that look at the final `string`, should see HTML encoding of the attribute values.

for `TagHelperOutputExtensionsTest` I recommend creating the `TagBuilder` instances with `CommonTestEncoder` as well.  won't get used but that is still testing the scenario.

dougbu:
:shipit:

PR #2419, opened 2015-04-21
Refactor attribute route

kichalla:
@rynowak

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/Routing/AttributeRoute.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Routing/AttributeRoute.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Mvc.Core;
@@ -103,16 +102,16 @@ namespace Microsoft.AspNet.Mvc.Routing
                 {
                     Order = routeInfo.Order,
                     Precedence = routeInfo.Precedence,
-                    Route = new TemplateRoute(
-                        _target,
-                        routeInfo.RouteTemplate,
-                        defaults: new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase)
+                    Target = _target,
+                    RouteName = routeInfo.Name,
+                    RouteTemplate = routeInfo.RouteTemplate,
+                    TemplateMatcher = new TemplateMatcher(
+                        routeInfo.ParsedTemplate,
+                        new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase)
                         {
-                            { AttributeRouting.RouteGroupKey, routeInfo.RouteGroup },
-                        },
-                        constraints: null,
-                        dataTokens: null,
-                        inlineConstraintResolver: _constraintResolver),
+                            { AttributeRouting.RouteGroupKey, routeInfo.RouteGroup }
+                        }),
+                    Constraints = routeInfo.Constraints
                 });
             }
 
--- a/src/Microsoft.AspNet.Mvc.Core/Routing/AttributeRouteMatchingEntry.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Routing/AttributeRouteMatchingEntry.cs
@@ -1,7 +1,10 @@
 ﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Collections.Generic;
+using Microsoft.AspNet.Routing;
 using Microsoft.AspNet.Routing.Template;
+using Microsoft.Framework.Logging;
 
 namespace Microsoft.AspNet.Mvc.Routing
 {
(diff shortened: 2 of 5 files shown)


src/Microsoft.AspNet.Mvc.Core/Routing/AttributeRoute.cs
```diff
@@ -99,20 +100,23 @@ private InnerAttributeRoute BuildRoute(ActionDescriptorsCollection actions)
...
-                            { AttributeRouting.RouteGroupKey, routeInfo.RouteGroup },
-                        },
-                        constraints: null,
-                        dataTokens: null,
-                        inlineConstraintResolver: _constraintResolver),
+                    Target = _target,
+                    RouteName = routeInfo.Name,
+                    RouteTemplate = routeInfo.RouteTemplate,
+                    TemplateMatcher = new TemplateMatcher(routeInfo.ParsedTemplate, defaults),
+                    Defaults = defaults,
```
rynowak: Is `Defaults` used anywhere or just used by the `TemplateMatcher`?

src/Microsoft.AspNet.Mvc.Core/Routing/AttributeRoute.cs
```diff
@@ -99,20 +100,23 @@ private InnerAttributeRoute BuildRoute(ActionDescriptorsCollection actions)
...
+                    Constraints = routeInfo.Constraints,
+                    DataTokens = EmptyRouteValueDictionary,
```
rynowak: You can just remove `DataTokens` from this class, it's never used in attribute routing.

src/Microsoft.AspNet.Mvc.Core/Routing/AttributeRoute.cs
```diff
@@ -99,20 +100,23 @@ private InnerAttributeRoute BuildRoute(ActionDescriptorsCollection actions)
...
+                    InlineConstraintResolver = _constraintResolver
```
rynowak: This is a service, so we shouldn't pass it around like this. If `InnerAttributeRoute` needs it, just pass it in the constructor.

src/Microsoft.AspNet.Mvc.Core/Routing/AttributeRouteMatchingEntry.cs
```diff
@@ -21,9 +24,20 @@ public class AttributeRouteMatchingEntry
...
+
+        public string RouteTemplate { get; set; }
+
+        public TemplateMatcher TemplateMatcher { get; set; }
+
+        public IReadOnlyDictionary<string, object> Defaults { get; set; }
+
+        public IReadOnlyDictionary<string, IRouteConstraint> Constraints { get; set; }
+
+        public IReadOnlyDictionary<string, object> DataTokens { get; set; }
```
rynowak: No need for data tokens support in attribute routes.

src/Microsoft.AspNet.Mvc.Core/Routing/AttributeRouteMatchingEntry.cs
```diff
@@ -21,9 +24,20 @@ public class AttributeRouteMatchingEntry
...
+
+        public IInlineConstraintResolver InlineConstraintResolver { get; set; }
```
rynowak: This shouldn't be needed, since `Constraints` already has them resolved. By wary of passing around services like this, it usually means something needs a refactor.

src/Microsoft.AspNet.Mvc.Core/Routing/InnerAttributeRoute.cs
```diff
@@ -101,20 +100,54 @@ public class InnerAttributeRoute : IRouter
...
+                {
+                    // If we got back a null value set, that means the URI did not match
+                    continue;
+                }
+
                 var oldRouteData = context.RouteData;
 
                 var newRouteData = new RouteData(oldRouteData);
-                newRouteData.Routers.Add(route);
+                MergeValues(newRouteData.DataTokens, matchingEntry.DataTokens);
```
rynowak: No need for this, we don't expose DataTokens in attribute routing.

rynowak:
:watch: Good start, I think it can be cleaned up a bit more, but this is what I had in mind.

src/Microsoft.AspNet.Mvc.Core/Routing/AttributeRoute.cs
kichalla: Sure..will do...

src/Microsoft.AspNet.Mvc.Core/Routing/AttributeRouteMatchingEntry.cs
kichalla: :+1:

src/Microsoft.AspNet.Mvc.Core/Routing/AttributeRoute.cs
kichalla: yes, I have now removed this property.

kichalla:
Updated. /cc: @rynowak

src/Microsoft.AspNet.Mvc.Core/Routing/InnerAttributeRoute.cs
```diff
@@ -101,20 +100,53 @@ public class InnerAttributeRoute : IRouter
...
+                    RouteDirection.IncomingRequest,
+                    _constraintLogger))
+                {
+                    continue;
+                }
+
+                _logger.LogInformation(
+                "Request successfully matched the route with name '{RouteName}' and template '{RouteTemplate}'.",
+                matchingEntry.RouteName,
+                matchingEntry.RouteTemplate);
```
rynowak: Fix indent here.

rynowak:
:shipit: after fixing spacing

PR #2420, opened 2015-04-21
Using custom HtmlEncoder in unit tests - part 2

ajaybhargavb:
Issue - #2085 
First part - #2417 
@dougbu @Eilon


dougbu:
:shipit: but make sure the CI is green first

Eilon:
:shipit: from me too. I actually like this a lot now. It's so much easier to tell what's going on in the test.

PR #2421, opened 2015-04-21
Utilize `nameof` operator

henkmollema:
Utilized the `nameof` operator with `ArgumentExceptions` in the following projects:
- `Microsoft.AspNet.Mvc.Core`
- `Microsoft.AspNet.Mvc.ModelBinding`
- `Microsoft.AspNet.Mvc.Razor.Host`
- `Microsoft.AspNet.Mvc.WebApiCompatShim`

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/AntiForgery/AntiForgeryTokenSet.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/AntiForgery/AntiForgeryTokenSet.cs
@@ -12,7 +12,7 @@ namespace Microsoft.AspNet.Mvc
         {
             if (string.IsNullOrEmpty(formToken))
             {
-                throw new ArgumentException(Resources.ArgumentCannotBeNullOrEmpty, formToken);
+                throw new ArgumentException(Resources.ArgumentCannotBeNullOrEmpty, nameof(formToken));
             }
 
             FormToken = formToken;
--- a/src/Microsoft.AspNet.Mvc.Core/FormatterMappings.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/FormatterMappings.cs
@@ -68,7 +68,7 @@ namespace Microsoft.AspNet.Mvc
         {
             if (format == "")
             {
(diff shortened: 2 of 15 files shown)


src/Microsoft.AspNet.Mvc.Core/AntiForgery/AntiForgeryTokenSet.cs
```diff
@@ -12,7 +12,7 @@ public AntiForgeryTokenSet(string formToken, string cookieToken)
         {
             if (string.IsNullOrEmpty(formToken))
             {
-                throw new ArgumentException(Resources.ArgumentCannotBeNullOrEmpty, formToken);
+                throw new ArgumentException(Resources.ArgumentCannotBeNullOrEmpty, nameof(formToken));
```
henkmollema: Looks like this was a mistake.

Eilon: Yup, good find.

Eilon:
@kichalla can you take a look and merge?

pranavkm:
:shipit:

src/Microsoft.AspNet.Mvc.Core/FormatterMappings.cs
```diff
@@ -68,7 +68,7 @@ private string RemovePeriodIfPresent(string format)
         {
             if (format == "")
```
pranavkm: This should probably be `string.IsNullOrEmpty`

henkmollema: Should that change be in this commit too? Seems a bit out of scope.

Eilon: Yeah separate change.

kichalla: I will then go ahead and merge this PR.

kichalla:
Merged: https://github.com/aspnet/Mvc/commit/92554fa634b9feea1c3ad6688e83c598f8875aca

Issue #819, opened 2014-07-18
Attribute Routing: Resolve Name + Token Replacement


rynowak:
Consider the following:

```
[Route("Products/[action]", Name = "Products")]
public class ProductsController
{
    public void Add() { }
    public void Buy() { }
}
```

This will fail because it creates two routes with the same name. We need to come up with something for this scenario.

kirthik:
Fix checked in - 9d2b1822d9e93bf74e3f1df8096b6bb2d6d43371

PR #2422, opened 2015-04-21
Fix for #819 : Attribute Routing: Resolve Name + Token Replacement

kirthik:
@rynowak : Could you please help take a look?

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ControllerActionDescriptorBuilder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ControllerActionDescriptorBuilder.cs
@@ -80,6 +80,12 @@ namespace Microsoft.AspNet.Mvc
                             // for this action.
                             ReplaceAttributeRouteTokens(actionDescriptor, routeTemplateErrors);
 
+                            var attributeRouteInfo = actionDescriptor.AttributeRouteInfo;
+                            if (attributeRouteInfo.Name != null)
+                            {
+                                attributeRouteInfo.Name = attributeRouteInfo.Name + "_" + action.ActionName;
+                            }
+
                             // Attribute routed actions will ignore conventional routed constraints. Instead they have
                             // a single route constraint "RouteGroup" associated with it.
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNet.Mvc.Core/ControllerActionDescriptorBuilder.cs
```diff
@@ -80,6 +80,12 @@ public static IList<ControllerActionDescriptor> Build(ApplicationModel applicati
                             // for this action.
                             ReplaceAttributeRouteTokens(actionDescriptor, routeTemplateErrors);
 
+                            var attributeRouteInfo = actionDescriptor.AttributeRouteInfo;
+                            if (attributeRouteInfo.Name != null)
+                            {
+                                attributeRouteInfo.Name = attributeRouteInfo.Name + "_" + action.ActionName;
```
rynowak: I think there's some confusion here and a wire got crossed. I was expecting this fix to run the token replacement code on the route name. I reread your email and I see what you're saying there, I must have missed that you were proposing something different.

So `Products_[action]` -> `Products_Edit`.

The problems with what you're doing here are twofold:
1). This doesn't account for the use of `[controller]` or `[area]` - both of which are legal
2). This is magic - you're taking something the user wrote, and turning it into something she didn't write.

kirthik:
@rynowak - updated code. Could you please take a look?

rynowak:
:shipit: conditionally - can you please try one thing - add a testscase which uses an unknown token in the route name, and **verify that the error message makes sense**.

If the error message doesn't make sense, let's fix that too.

kirthik:
Here is the error message we see when unknown token is used - 
```
While processing template 'Products_[unknown]', a replacement value for the token 'unknown' could not be found. Available tokens: 'action, controller'.
```

Seems ok to me.

rynowak:
Yep. Seems Ok.

kirthik:
checked in - 9d2b1822d9e93bf74e3f1df8096b6bb2d6d43371

Issue #2423, opened 2015-04-21
[Mvc 5.0 Compat] Null values returned by value providers/model binders for value type properties should not create a model state error.


harshgMSFT:
Mvc 5.0 behavior is to use the default value, however in mvc 6.0 we add the following error:
`A value is required.`

For example :
```c#
public class Person
{
     [FromBody]
     public int SSN { get; set; }
}

public void Action(Person p)
{
}
```
Request body is empty.
There would be an error in model state in MVC 6.0 and  no one for a comparable scenario in MVC 5.0 (where a value provider returns a null for the key SSN ).

dougbu:
@harshgMSFT could you please check if de63075 (which removed the `[Required]` special case for JSON inputs) also fixed this compat issue?

Bartmax:
just so you can take into consideration, I hope `[Required]` (and similar) still makes a `ModelState` error on this same scenario, even while on mvc5 it will not. I see this as an enhancement of a bad design on mvc5 that can be fixed (it is already) on mvc6. Just please, don't go backwards with this one.

Specially, 
```
[Required]
public int Amount {get; set;}
```

this won't ever trigger a required error in mvc5, you need to change `int` to `int?` because it's defaulting to 0 (default value). in mvc6 this does triggers and error and it's great!.

I know this is about `[FromBody]` attribute, but just in case.

yishaigalatzer:
MVC never supported `[FromBody]` this should probably be compatible with WebAPI rather than MVC

harshgMSFT:
@Bartmax The reason for not logging an error is that '[Required]' means that the value should be not null. In case of value types it can never be null. So instead of overloading Required to mean not present in the request we are going with what it is supposed to mean (not null).

Bartmax:
@harshgMSFT I understand how it works, but I think it's a bad design (I may be wrong). I created a specific issue for this https://github.com/aspnet/Mvc/issues/2615

yishaigalatzer:
Agree with @bartmax

borgdylan:
I think that even without [Required], the model should be invalid if a value is not given for a valuetype. Nullable<T> should be used to indicate that a value is not by force required to be present.

Eilon:
Closing this. To specify that a value must be on the wire, use `[BindRequired]` on the property. The `[Required]` attribute means only what Data Annotations says it means, which is that the current value of the property is not-null, when applicable. Because value types can't be null, the `[Required]` attribute will by-design not do anything.

Issue #2357, opened 2015-04-08
Value type controller properties with [From*] fail with NullRef exception when no value is available


pranavkm:
Scenario:
```C#
public class TestController
{
   [FromBody]
   public int Value { get; set; }

  [HttpPost]
   public int DoPost()
   {
      return Value;
   }
}
```
When an invalid value is posted (for instance posting a JSON string here), the binder returns a null value. Attempting to set `null` on a value type property fails with a null ref.

PR #2424, opened 2015-04-21
Fix for #2357 : We prevent assigining null values to non nullable contro...

harshgMSFT:
...ller properties.

We do not add a model state error for this case to ensure we have a similar behavior as mvc 5.0 (in terms of no model state errors for value types). see #2423

### DIFF
--- a/test/Microsoft.AspNet.Mvc.Core.Test/ParameterBinding/ControllerActionArgumentBinderTests.cs
+++ b/test/Microsoft.AspNet.Mvc.Core.Test/ParameterBinding/ControllerActionArgumentBinderTests.cs
@@ -318,7 +318,7 @@ namespace Microsoft.AspNet.Mvc.Core.Test
 
 
         [Fact]
-        public async Task BindActionArgumentsAsync_SetsControllerProperties()
+        public async Task BindActionArgumentsAsync_SetsControllerProperties_ForReferenceTypes()
         {
             // Arrange
             var actionDescriptor = GetActionDescriptor();
@@ -351,8 +351,8 @@ namespace Microsoft.AspNet.Mvc.Core.Test
             actionDescriptor.BoundProperties.Add(
                 new ParameterDescriptor
                 {
-                    Name = "ValueBinderMarkedProperty",
-                    BindingInfo = new BindingInfo(),
+                    Name = "NotNullableProperty",
+                    BindingInfo = new BindingInfo() { BindingSource = BindingSource.Custom },
                     ParameterType = typeof(int)
                 });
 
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNet.Mvc.Core.Test/ParameterBinding/ControllerActionArgumentBinderTests.cs
```diff
@@ -400,6 +438,9 @@ private class TestController
             [ValueProviderMetadata]
             public string ValueBinderMarkedProperty { get; set; }
 
+            [CustomBindingSource]
```
rynowak: How does this come into play? It looks like your test doesn't pass it to `BindingInfo`.

test/Microsoft.AspNet.Mvc.Core.Test/ParameterBinding/ControllerActionArgumentBinderTests.cs
```diff
@@ -343,6 +343,44 @@ public void ParameterHasFieldPrefix([Bind(Prefix = "simpleModelPrefix")] string
...
+
+            // Some non default value.
+            controller.NotNullableProperty = -1;
+
+            // Act
+            var result = await argumentBinder.BindActionArgumentsAsync(actionContext, actionBindingContext, controller);
+
+            // Assert
+            Assert.Equal(-1, controller.NotNullableProperty);
+        }
```
rynowak: Do we have a positive test for the nullable-int equivalent?

test/Microsoft.AspNet.Mvc.Core.Test/ParameterBinding/ControllerActionArgumentBinderTests.cs
harshgMSFT: good catch : yeah you are right it doesnt come into play, in a normal (e2e or integartion scenario) this would be required however this is just for completeness ( it will be confusing to see a property which is not marked, being bound). I will update the test to use BindingSource.Custom.

test/Microsoft.AspNet.Mvc.Core.Test/ParameterBinding/ControllerActionArgumentBinderTests.cs
harshgMSFT: We do not have a test for nullable-int, but there are tests for reference type (string). Ill add one for nullable-int as well.

harshgMSFT:
@rynowak  updated

test/Microsoft.AspNet.Mvc.Core.Test/ParameterBinding/ControllerActionArgumentBinderTests.cs
rynowak: Yeah, my slight preference is to just leave stuff like this out if it's not connected to anything. Gives the impression that it's magic :tophat:

rynowak:
:shipit:

Issue #662, opened 2014-06-16
DefaultControllerFactory should throw if the controller is a value type


pranavkm:
Might want to expand it to include more things (static \ abstract) etc and provide a meaningful message.

Eilon:
I'm not sure it's that important to throw for value types, because who would do that. But other cases such as generic types, abstract types, etc. we should look at to ensure we're safe and provide good errors.

rynowak:
Keep in mind that this will only happen if you build your own action descriptors. We already filter out all of these kinds of things during discovery.

PR #2427, opened 2015-04-21
Throw if controller cannot be activated

ajaybhargavb:
Issue - #662 @Eilon @rynowak

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/Resources.resx
+++ b/src/Microsoft.AspNet.Mvc.Core/Resources.resx
@@ -319,9 +319,6 @@
   <data name="ActionResult_ActionReturnValueCannotBeNull" xml:space="preserve">
     <value>Cannot return null from an action method with a return type of '{0}'.</value>
   </data>
-  <data name="ValueTypesCannotBeActivated" xml:space="preserve">
-    <value>Value types cannot be activated by '{0}'.</value>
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNet.Mvc.Core/Resources.resx
```diff
@@ -460,4 +457,7 @@
   <data name="TempData_CannotSerializeDictionary" xml:space="preserve">
     <value>The '{0}' cannot serialize a dictionary with a key of type '{1}' to session state.</value>
   </data>
+  <data name="ValueInterfaceAbstractOrOpenGenericTypesCannotBeActivated" xml:space="preserve">
+    <value>Value, Interface, Abstract or Open Generic types cannot be activated by '{0}'.</value>
```
Eilon: How about: `The type '{0}' cannot be created by '{1}' because it is either a value type, an interface, an abstract class, or an open generic type.`

Eilon:
Other than the string, :shipit:

src/Microsoft.AspNet.Mvc.Core/Resources.resx
rynowak: :+1: was going to suggest something similar. We could even simplify it further and say that 'non-instantiable types and value types' aren't supported.

PR #2442, opened 2015-04-24
Added the default constructor to ViewContext

ianhong:
Added the default constructor for ViewContext. 

1. All properties already have the setter. 
2. I checked three unit test scenarios that this change can be useful
   - ViewComponent : Unit-testable with this default viewcontext 
  -  TagHelper, HtmlHelper : Not really, we still need more for example the default constructor for DefaultHtmlGenerator...

Thanks

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ViewComponents/ViewComponent.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ViewComponents/ViewComponent.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Security.Principal;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Mvc.ModelBinding;
@@ -17,6 +18,8 @@ namespace Microsoft.AspNet.Mvc
     public abstract class ViewComponent
     {
         private dynamic _viewBag;
+        private ViewContext _viewContext;
+        private ViewDataDictionary _viewData;
 
         /// <summary>
         /// Gets the <see cref="HttpContext"/>.
@@ -99,13 +102,49 @@ namespace Microsoft.AspNet.Mvc
         /// Gets or sets the <see cref="ViewContext"/>.
         /// </summary>
         [Activate]
-        public ViewContext ViewContext { get; set; }
+        public ViewContext ViewContext
+        {
+            get
+            {
+                if (_viewContext == null)
+                {
+                    // This should run only for the ViewComponent unit test scenarios
+                    _viewContext = new ViewContext();
+                }
+
+                return _viewContext;
+            }
+
+            [param: NotNull]
+            set
+            {
+                _viewContext = value;
+            }
+        }
 
         /// <summary>
         /// Gets or sets the <see cref="ViewDataDictionary"/>.
         /// </summary>
         [Activate]
-        public ViewDataDictionary ViewData { get; set; }
+        public ViewDataDictionary ViewData
+        {
+            get
+            {
+                if (_viewData == null)
+                {
+                    // This should run only for the ViewComponent unit test scenarios
+                    _viewData = new ViewDataDictionary(ViewContext.ViewData);
+                }
+
+                return _viewData;
+            }
+
+            [param: NotNull]
+            set
+            {
+                _viewData = value;
+            }
+        }
 
         /// <summary>
         /// Gets or sets the <see cref="ICompositeViewEngine"/>.
--- a/src/Microsoft.AspNet.Mvc.Core/ViewContext.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ViewContext.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.IO;
+using Microsoft.AspNet.Mvc.ModelBinding;
 using Microsoft.AspNet.Mvc.Rendering;
 using Microsoft.Framework.Internal;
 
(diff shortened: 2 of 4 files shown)


ianhong:
@Eilon and @rynowak 
This is the PR for MusicStore. It shows how this change can be used to do unit tests for ViewComponents. 

https://github.com/aspnet/MusicStore/pull/474

src/Microsoft.AspNet.Mvc.Core/ViewContext.cs
```diff
@@ -19,6 +19,16 @@ public class ViewContext : ActionContext
         private DynamicViewData _viewBag;
 
         /// <summary>
+        /// Creates a empty <see cref="ViewContext"/>.
```
Eilon: `a` -> `an`

Eilon: Looks like the same typo is here: https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/ActionContext.cs#L17

ianhong:
@rynowak  and @Eilon 

I updated ViewComponent to help its unit test scenario. Please take a look at them. With this change,

https://github.com/aspnet/MusicStore/blob/ViewComponentUnitTest/test/MusicStore.Test/AnnouncementComponentTest.cs#L33-L46

can be like

'''
            var announcementComponent = new AnnouncementComponent()
            {
                DbContext = _serviceProvider.GetRequiredService<MusicStoreContext>(),
                Cache = _serviceProvider.GetRequiredService<IMemoryCache>(),
            };
'''

ianhong:
I'd like to update MusicStore once this is checked in.

src/Microsoft.AspNet.Mvc.Core/ViewComponents/ViewComponent.cs
```diff
@@ -99,13 +102,60 @@ public ModelStateDictionary ModelState
...
+                }
+
+                return _viewContext;
+            }
+
+            set
+            {
+                if (value == null)
+                {
+                    throw new ArgumentNullException(nameof(value));
```
dougbu: use `[param: NotNull]` instead; see guidelines

src/Microsoft.AspNet.Mvc.Core/ViewComponents/ViewComponent.cs
```diff
@@ -99,13 +102,60 @@ public ModelStateDictionary ModelState
...
+        {
+            get
+            {
+                if (_viewContext == null)
+                {
+                    // This should run only for the ViewComponent unit test scenarios
+                    _viewContext = new ViewContext();
+                    _viewContext.ViewData = new ViewDataDictionary(
+                        new EmptyModelMetadataProvider(),
+                        _viewContext.ModelState);
```
dougbu: why doesn't the `ViewContext` constructor do this?

src/Microsoft.AspNet.Mvc.Core/ViewComponents/ViewComponent.cs
```diff
@@ -99,13 +102,60 @@ public ModelStateDictionary ModelState
...
         [Activate]
-        public ViewDataDictionary ViewData { get; set; }
+        public ViewDataDictionary ViewData
+        {
+            get
+            {
+                if (_viewData == null)
+                {
+                    // This should run only for the ViewComponent unit test scenarios
+                    _viewData = new ViewDataDictionary(ViewContext.ViewData);
```
dougbu: avoid inconsistencies between `ViewData` and `ViewContext.ViewData`.  the two properties will refer to the same instance after activation.  could for example change this line to instead be `_viewData = ViewContext.ViewData;`

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.Core/ViewComponents/ViewComponent.cs
ianhong: Thanks @dougbu 

The reason why I created a new instance here because it seems like the default behavior for viewcomponent from the following code: 
https://github.com/aspnet/Mvc/blob/2b82ee0255e223ce2347e0e5b7067c7140779db5/src/Microsoft.AspNet.Mvc.Core/ViewComponents/DefaultViewComponentActivator.cs#L52-L65
And I think we want to took a snap shot for viewdata for viewcomponent. 

@rynowak, could you clarify which the correct behavior is..

src/Microsoft.AspNet.Mvc.Core/ViewComponents/ViewComponent.cs
ianhong: ViewContext is shared everywhere..and I thought we might change that default constructor again when we find new requirements when considering unit testability of like TagHelper, HtmlHelper, etc..So, I think a minimum change should be better at this time..  
@dougbu What do you think??

src/Microsoft.AspNet.Mvc.Core/ViewComponents/ViewComponent.cs
ianhong: Thanks @dougbu I saw this type of null check somewhere else..I think we should clean them up  at some point :-) I'll fix this in next iteration..

src/Microsoft.AspNet.Mvc.Core/ViewComponents/ViewComponent.cs
dougbu: ah, I was slightly incorrect: after activation, `ViewData` will contain a copy-on-write version of what's in `ViewContext.ViewData`.  this is a little strange because other mini-scope scenarios (e.g. `HtmlHelper` recursion and tag helper activation) start that scope with consistent `ViewContext.ViewData` and `ViewData`.  the uniqueness in the method you referenced looks weird.

src/Microsoft.AspNet.Mvc.Core/ViewComponents/ViewComponent.cs
dougbu: as you've seen, view component testing needs this.  HTML helpers definitely use `ViewData` and having an instance available does no harm in tag helpers (few MVC tag helpers even indirectly use `ViewData`). so I think this basic initialization should be done in one test-only location.

ianhong:
@dougbu 
I updated per your comments. Thanks

dougbu:
:shipit:

PR #2450, opened 2015-04-24
Use strongly typed MediaTypeHeaderValue for content type in action resul...

kichalla:
...ts.

@rynowak @pranavkm

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ActionResults/ContentResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ActionResults/ContentResult.cs
@@ -1,6 +1,8 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+using System.IO;
 using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Http;
@@ -11,11 +13,20 @@ namespace Microsoft.AspNet.Mvc
 {
     public class ContentResult : ActionResult
     {
+        private readonly MediaTypeHeaderValue DefaultContentType = new MediaTypeHeaderValue("text/plain")
+        {
+            Encoding = Encodings.UTF8EncodingWithoutBOM
+        };
+
+        /// <summary>
+        /// Gets or set the content representing the body of the response.
+        /// </summary>
         public string Content { get; set; }
 
-        public Encoding ContentEncoding { get; set; }
-
-        public string ContentType { get; set; }
+        /// <summary>
+        /// Gets or sets the <see cref="MediaTypeHeaderValue"/> representing the Content-Type header of the response.
+        /// </summary>
+        public MediaTypeHeaderValue ContentType { get; set; }
 
         /// <summary>
         /// Gets or sets the HTTP status code.
@@ -26,17 +37,30 @@ namespace Microsoft.AspNet.Mvc
         {
             var response = context.HttpContext.Response;
 
-            MediaTypeHeaderValue contentTypeHeader;
-            if (string.IsNullOrEmpty(ContentType))
+            var contentTypeHeader = ContentType;
+            Encoding encoding;
+            if (contentTypeHeader == null)
             {
-                contentTypeHeader = new MediaTypeHeaderValue("text/plain");
+                contentTypeHeader = DefaultContentType;
+                encoding = Encodings.UTF8EncodingWithoutBOM;
             }
             else
             {
-                contentTypeHeader = new MediaTypeHeaderValue(ContentType);
+                if (contentTypeHeader.Encoding == null)
+                {
+                    // 1. Do not modify the user supplied content type
+                    // 2. Parse here to handle parameters apart from charset
+                    contentTypeHeader = MediaTypeHeaderValue.Parse(contentTypeHeader.ToString());
+                    contentTypeHeader.Encoding = Encodings.UTF8EncodingWithoutBOM;
+
+                    encoding = Encodings.UTF8EncodingWithoutBOM;
+                }
+                else
+                {
+                    encoding = contentTypeHeader.Encoding;
+                }
             }
 
-            contentTypeHeader.Encoding = ContentEncoding ?? Encodings.UTF8EncodingWithoutBOM;
             response.ContentType = contentTypeHeader.ToString();
 
             if (StatusCode != null)
@@ -46,7 +70,7 @@ namespace Microsoft.AspNet.Mvc
 
             if (Content != null)
             {
-                await response.WriteAsync(Content, contentTypeHeader.Encoding);
+                await response.WriteAsync(Content, encoding);
             }
         }
     }
--- a/src/Microsoft.AspNet.Mvc.Core/ActionResults/FileContentResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ActionResults/FileContentResult.cs
@@ -6,6 +6,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Http;
 using Microsoft.Framework.Internal;
+using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNet.Mvc
 {
@@ -19,13 +20,14 @@ namespace Microsoft.AspNet.Mvc
 
         /// <summary>
         /// Creates a new <see cref="FileContentResult"/> instance with
-        /// the provided <paramref name="fileContents"/>.
+        /// the provided <paramref name="fileContents"/> and the
+        /// provided <paramref name="contentType"/>.
         /// </summary>
         /// <param name="fileContents">The bytes that represent the file contents.</param>
-        public FileContentResult([NotNull] byte[] fileContents)
-            : base(contentType: null)
+        /// <param name="contentType">The Content-Type header of the response.</param>
+        public FileContentResult([NotNull] byte[] fileContents, [NotNull] string contentType)
+            : this(fileContents, new MediaTypeHeaderValue(contentType))
         {
-            FileContents = fileContents;
         }
 
         /// <summary>
@@ -35,7 +37,7 @@ namespace Microsoft.AspNet.Mvc
         /// </summary>
         /// <param name="fileContents">The bytes that represent the file contents.</param>
         /// <param name="contentType">The Content-Type header of the response.</param>
-        public FileContentResult([NotNull] byte[] fileContents, string contentType)
+        public FileContentResult([NotNull] byte[] fileContents, [NotNull] MediaTypeHeaderValue contentType)
             : base(contentType)
         {
             FileContents = fileContents;
--- a/src/Microsoft.AspNet.Mvc.Core/ActionResults/FilePathResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ActionResults/FilePathResult.cs
@@ -11,6 +11,7 @@ using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Mvc.Core;
 using Microsoft.Framework.DependencyInjection;
 using Microsoft.Framework.Internal;
+using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNet.Mvc
 {
@@ -27,15 +28,15 @@ namespace Microsoft.AspNet.Mvc
 
         /// <summary>
         /// Creates a new <see cref="FilePathResult"/> instance with
-        /// the provided <paramref name="fileName"/>
+        /// the provided <paramref name="fileName"/> and the
+        /// provided <paramref name="contentType"/>.
         /// </summary>
         /// <param name="fileName">The path to the file. The path must be an absolute
         /// path. Relative and virtual paths are not supported.</param>
         /// <param name="contentType">The Content-Type header of the response.</param>
-        public FilePathResult([NotNull] string fileName)
-            : base(contentType: null)
+        public FilePathResult([NotNull] string fileName, [NotNull] string contentType)
+            : this(fileName, new MediaTypeHeaderValue(contentType))
         {
-            FileName = fileName;
         }
 
         /// <summary>
@@ -46,7 +47,7 @@ namespace Microsoft.AspNet.Mvc
         /// <param name="fileName">The path to the file. The path must be an absolute
         /// path. Relative and virtual paths are not supported.</param>
         /// <param name="contentType">The Content-Type header of the response.</param>
-        public FilePathResult([NotNull] string fileName, string contentType)
+        public FilePathResult([NotNull] string fileName, [NotNull] MediaTypeHeaderValue contentType)
             : base(contentType)
         {
             FileName = fileName;
--- a/src/Microsoft.AspNet.Mvc.Core/ActionResults/FileResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ActionResults/FileResult.cs
@@ -24,15 +24,25 @@ namespace Microsoft.AspNet.Mvc
         /// the provided <paramref name="contentType"/>.
         /// </summary>
         /// <param name="contentType">The Content-Type header of the response.</param>
-        protected FileResult(string contentType)
+        protected FileResult([NotNull] string contentType)
+            : this(new MediaTypeHeaderValue(contentType))
+        {
+        }
+
+        /// <summary>
+        /// Creates a new <see cref="FileResult"/> instance with
+        /// the provided <paramref name="contentType"/>.
+        /// </summary>
+        /// <param name="contentType">The Content-Type header of the response.</param>
+        protected FileResult([NotNull] MediaTypeHeaderValue contentType)
         {
             ContentType = contentType;
         }
 
         /// <summary>
-        /// Gets or sets the Content-Type header value that will be written to the response.
+        /// Gets the <see cref="MediaTypeHeaderValue"/> representing the Content-Type header of the response.
         /// </summary>
-        public string ContentType { get; set; }
+        public MediaTypeHeaderValue ContentType { get; }
 
         /// <summary>
         /// Gets the file name that will be used in the Content-Disposition header of the response.
@@ -47,11 +57,7 @@ namespace Microsoft.AspNet.Mvc
         public override Task ExecuteResultAsync([NotNull] ActionContext context)
         {
             var response = context.HttpContext.Response;
-
-            if (ContentType != null)
-            {
-                response.ContentType = ContentType;
-            }
+            response.ContentType = ContentType.ToString();
 
             if (!string.IsNullOrEmpty(FileDownloadName))
             {
--- a/src/Microsoft.AspNet.Mvc.Core/ActionResults/FileStreamResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ActionResults/FileStreamResult.cs
@@ -7,6 +7,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Http;
 using Microsoft.Framework.Internal;
+using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNet.Mvc
 {
@@ -23,13 +24,14 @@ namespace Microsoft.AspNet.Mvc
 
         /// <summary>
         /// Creates a new <see cref="FileStreamResult"/> instance with
-        /// the provided <paramref name="fileStream"/>.
+        /// the provided <paramref name="fileStream"/> and the
+        /// provided <paramref name="contentType"/>.
         /// </summary>
         /// <param name="fileStream">The stream with the file.</param>
-        public FileStreamResult([NotNull] Stream fileStream)
-            : base(contentType: null)
+        /// <param name="contentType">The Content-Type header of the response.</param>
+        public FileStreamResult([NotNull] Stream fileStream, [NotNull] string contentType)
+            : this(fileStream, new MediaTypeHeaderValue(contentType))
         {
-            FileStream = fileStream;
         }
 
         /// <summary>
@@ -39,7 +41,7 @@ namespace Microsoft.AspNet.Mvc
         /// </summary>
         /// <param name="fileStream">The stream with the file.</param>
         /// <param name="contentType">The Content-Type header of the response.</param>
-        public FileStreamResult([NotNull] Stream fileStream, string contentType)
+        public FileStreamResult([NotNull] Stream fileStream, [NotNull] MediaTypeHeaderValue contentType)
             : base(contentType)
         {
             FileStream = fileStream;
--- a/src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewExecutor.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewExecutor.cs
@@ -1,10 +1,13 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.IO;
+using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Mvc.Rendering;
 using Microsoft.Framework.Internal;
+using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNet.Mvc
 {
@@ -14,7 +17,10 @@ namespace Microsoft.AspNet.Mvc
     public static class ViewExecutor
     {
         private const int BufferSize = 1024;
-        private const string ContentType = "text/html; charset=utf-8";
+        private static readonly MediaTypeHeaderValue DefaultContentType = new MediaTypeHeaderValue("text/html")
+        {
+            Encoding = Encodings.UTF8EncodingWithoutBOM
+        };
 
         /// <summary>
         /// Asynchronously renders the specified <paramref name="view"/> to the response body.
@@ -23,21 +29,43 @@ namespace Microsoft.AspNet.Mvc
         /// <param name="actionContext">The <see cref="ActionContext"/> for the current executing action.</param>
         /// <param name="viewData">The <see cref="ViewDataDictionary"/> for the view being rendered.</param>
         /// <param name="tempData">The <see cref="ITempDataDictionary"/> for the view being rendered.</param>
-        /// <returns>A <see cref="Task"/> that represents the asychronous rendering.</returns>
+        /// <returns>A <see cref="Task"/> that represents the asynchronous rendering.</returns>
         public static async Task ExecuteAsync([NotNull] IView view,
                                               [NotNull] ActionContext actionContext,
                                               [NotNull] ViewDataDictionary viewData,
                                               [NotNull] ITempDataDictionary tempData,
-                                              string contentType)
+                                              MediaTypeHeaderValue contentType)
         {
-            if (string.IsNullOrEmpty(contentType))
+            var response = actionContext.HttpContext.Response;
+
+            var contentTypeHeader = contentType;
+            Encoding encoding;
+            if (contentTypeHeader == null)
             {
-                contentType = ContentType;
+                contentTypeHeader = DefaultContentType;
+                encoding = Encodings.UTF8EncodingWithoutBOM;
+            }
+            else
+            {
+                if (contentTypeHeader.Encoding == null)
+                {
+                    // 1. Do not modify the user supplied content type
+                    // 2. Parse here to handle parameters apart from charset
+                    contentTypeHeader = MediaTypeHeaderValue.Parse(contentTypeHeader.ToString());
+                    contentTypeHeader.Encoding = Encodings.UTF8EncodingWithoutBOM;
+
+                    encoding = Encodings.UTF8EncodingWithoutBOM;
+                }
+                else
+                {
+                    encoding = contentTypeHeader.Encoding;
+                }
             }
 
-            actionContext.HttpContext.Response.ContentType = contentType;
-            var wrappedStream = new StreamWrapper(actionContext.HttpContext.Response.Body);
-            var encoding = Encodings.UTF8EncodingWithoutBOM;
+            response.ContentType = contentTypeHeader.ToString();
+
+            var wrappedStream = new StreamWrapper(response.Body);
+
             using (var writer = new StreamWriter(wrappedStream, encoding, BufferSize, leaveOpen: true))
             {
                 try
--- a/src/Microsoft.AspNet.Mvc.Core/Controller.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Controller.cs
@@ -14,6 +14,7 @@ using Microsoft.AspNet.Mvc.ModelBinding.Validation;
 using Microsoft.AspNet.Mvc.Rendering;
 using Microsoft.AspNet.Routing;
 using Microsoft.Framework.Internal;
+using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNet.Mvc
 {
@@ -330,7 +331,7 @@ namespace Microsoft.AspNet.Mvc
         [NonAction]
         public virtual ContentResult Content(string content)
         {
-            return Content(content, contentType: null);
+            return Content(content, (MediaTypeHeaderValue)null);
         }
 
         /// <summary>
@@ -357,10 +358,22 @@ namespace Microsoft.AspNet.Mvc
         [NonAction]
         public virtual ContentResult Content(string content, string contentType, Encoding contentEncoding)
         {
+            return Content(content, new MediaTypeHeaderValue(contentType) { Encoding = contentEncoding });
+        }
+
+        /// <summary>
+        /// Creates a <see cref="ContentResult"/> object by specifying a <paramref name="content"/>
+        /// string and a <paramref name="contentType"/>.
+        /// </summary>
+        /// <param name="content">The content to write to the response.</param>
+        /// <param name="contentType">The content type (MIME type).</param>
+        /// <returns>The created <see cref="ContentResult"/> object for the response.</returns>
+        [NonAction]
+        public virtual ContentResult Content(string content, MediaTypeHeaderValue contentType)
+        {
             var result = new ContentResult
             {
                 Content = content,
-                ContentEncoding = contentEncoding,
                 ContentType = contentType
             };
 
--- a/test/Microsoft.AspNet.Mvc.Core.Test/ActionResults/ContentResultTest.cs
+++ b/test/Microsoft.AspNet.Mvc.Core.Test/ActionResults/ContentResultTest.cs
@@ -1,11 +1,12 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
+using System.IO;
 using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Routing;
+using Microsoft.Net.Http.Headers;
 using Moq;
 using Xunit;
 
@@ -20,8 +21,7 @@ namespace Microsoft.AspNet.Mvc
             var contentResult = new ContentResult
             {
                 Content = "Test Content",
-                ContentType = "application/json",
-                ContentEncoding = null
+                ContentType = new MediaTypeHeaderValue("application/json")
             };
             var httpContext = GetHttpContext();
             var actionContext = GetActionContext(httpContext);
@@ -40,8 +40,10 @@ namespace Microsoft.AspNet.Mvc
             var contentResult = new ContentResult
             {
                 Content = "Test Content",
-                ContentType = "text/plain",
-                ContentEncoding = Encoding.ASCII
+                ContentType = new MediaTypeHeaderValue("text/plain")
+                {
+                    Encoding = Encoding.ASCII
+                }
             };
             var httpContext = GetHttpContext();
             var actionContext = GetActionContext(httpContext);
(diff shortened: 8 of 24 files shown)


src/Microsoft.AspNet.Mvc.Core/ActionResults/FileContentResult.cs
```diff
@@ -19,13 +20,14 @@ public class FileContentResult : FileResult
 
         /// <summary>
         /// Creates a new <see cref="FileContentResult"/> instance with
-        /// the provided <paramref name="fileContents"/>.
+        /// the provided <paramref name="fileContents"/> and the
+        /// provided <paramref name="contentType"/>.
         /// </summary>
         /// <param name="fileContents">The bytes that represent the file contents.</param>
-        public FileContentResult([NotNull] byte[] fileContents)
-            : base(contentType: null)
```
kichalla: If we compare to MVC5, content type was always required a content-type:
http://aspnetwebstack.codeplex.com/SourceControl/latest#src/System.Web.Mvc/FileResult.cs

src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewExecutor.cs
```diff
@@ -23,22 +28,52 @@ public static class ViewExecutor
...
+
+            response.ContentType = contentTypeHeader.ToString();
+
+            var wrappedStream = new StreamWrapper(response.Body);
 
-            actionContext.HttpContext.Response.ContentType = contentType;
-            var wrappedStream = new StreamWrapper(actionContext.HttpContext.Response.Body);
-            var encoding = Encodings.UTF8EncodingWithoutBOM;
-            using (var writer = new StreamWriter(wrappedStream, encoding, BufferSize, leaveOpen: true))
+            using (var writer = new StreamWriter(
```
kichalla: will revert this formatting

src/Microsoft.AspNet.Mvc.Core/Controller.cs
```diff
@@ -357,10 +358,22 @@ public virtual ContentResult Content(string content, string contentType)
...
+
+        /// <summary>
+        /// Creates a <see cref="ContentResult"/> object by specifying a <paramref name="content"/>
+        /// string and a <paramref name="contentType"/>.
+        /// </summary>
+        /// <param name="content">The content to write to the response.</param>
+        /// <param name="contentType">The content type (MIME type).</param>
+        /// <returns>The created <see cref="ContentResult"/> object for the response.</returns>
+        [NonAction]
+        public virtual ContentResult Content(string content, MediaTypeHeaderValue contentType)
```
kichalla: There are lot of helper methods which accept content type as string, should i have corresponding helpers for MediaTypeHeaderValue too? wanted to check first as there are lot of methods to add.

kichalla:
ping

src/Microsoft.AspNet.Mvc.Core/ActionResults/ContentResult.cs
```diff
@@ -26,17 +29,34 @@ public class ContentResult : ActionResult
         {
             var response = context.HttpContext.Response;
 
+            // Encoding property on MediaTypeHeaderValue does not return the exact encoding instance that
```
pranavkm: We shouldn't need to this. If using the encoding set in `ContentType` is incorrect, we have an issue with the way `MediaTypeHeaderValue.Encoding` behaves. cc @Tratcher

src/Microsoft.AspNet.Mvc.Core/ActionResults/FileStreamResult.cs
```diff
@@ -39,7 +41,7 @@ public FileStreamResult([NotNull] Stream fileStream)
         /// </summary>
         /// <param name="fileStream">The stream with the file.</param>
         /// <param name="contentType">The Content-Type header of the response.</param>
-        public FileStreamResult([NotNull] Stream fileStream, string contentType)
+        public FileStreamResult([NotNull] Stream fileStream, MediaTypeHeaderValue contentType)
```
pranavkm: In this case, are we responsible for re-encoding `fileStream` if it doesn't match the `Encoding` on `contentType`?

src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewExecutor.cs
```diff
@@ -23,22 +28,52 @@ public static class ViewExecutor
...
+                                              MediaTypeHeaderValue contentType)
         {
-            if (string.IsNullOrEmpty(contentType))
+            var response = actionContext.HttpContext.Response;
+
+            // Encoding property on MediaTypeHeaderValue does not return the exact encoding instance that
+            // is set, so any settings(for example: BOM) on it will be lost when retrieving the value.
+            // In the scenario where the user does not supply encoding, we want to use UTF8 without BOM and
+            // for this reason do not rely on Encoding property.
+            MediaTypeHeaderValue contentTypeHeader;
```
pranavkm: Could we verify what the behavior was in Mvc 5 - what happened when the `ContentEncoding` was set to `Encoding.Utf8` - did we write the BOM?

src/Microsoft.AspNet.Mvc.Core/Controller.cs
pranavkm: Might be a question for @Eilon \ @danroth27

src/Microsoft.AspNet.Mvc.Core/ActionResults/FileStreamResult.cs
kichalla: I think we are not...this question is applicable to FilePathResult too which uses the following SendFileFeature in some situations.../cc: @Tratcher 

https://github.com/aspnet/Mvc/blob/a9feed4de1eb45bad0750bc868900b20cf842cc7/test/WebSites/FilesWebSite/SendFileMiddleware.cs#L50

src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewExecutor.cs
kichalla: sure..will do...

kichalla: I do not see any effect on the output when ContentEncoding is set to Encoding.Utf8

src/Microsoft.AspNet.Mvc.Core/ActionResults/FileStreamResult.cs
Tratcher: From our perspective files just contain bytes. It's up the application to decide what the best encoding is to use with that file and inform the user appropriately.

src/Microsoft.AspNet.Mvc.Core/ActionResults/ContentResult.cs
Tratcher: I'm planning to remove  MediaTypeHeaderValue.Encoding, it's causing confusion.

src/Microsoft.AspNet.Mvc.Core/Controller.cs
Eilon: Doesn't seem urgent... the `MediaTypeHeaderValue` ctor seems pretty easy to use. Can always add it later as a convenience if anyone asks for it.

src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewExecutor.cs
pranavkm: In which case the right fix for this would be to write a custom `StreamWriter` that doesn't write the BOM and not do whatever we ave here.

src/Microsoft.AspNet.Mvc.Core/ActionResults/ContentResult.cs
pranavkm: :panda_face:

dougbu: @pranavkm is that a happy or sad panda?

pranavkm: Aren't :panda_face: always sad?

kichalla:
Updated. /cc: @pranavkm

davidfowl:
Are we sure we want to do this?

pranavkm:
@davidfowl what's the concern?

kichalla:
Updated /cc: @pranavkm @harshgMSFT @rynowak 

In the newest iteration, I added a `ResponseEncodingWrapper` which ignores the preamble of the original encoding(except the UTF16 little endian which we explicitly set to honor https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Common/Encodings.cs#L20)

src/Microsoft.AspNet.Mvc.Core/ActionResults/ContentResult.cs
```diff
@@ -26,17 +31,47 @@ public class ContentResult : ActionResult
...
+            if(contentTypeHeader == null)
             {
-                contentTypeHeader = new MediaTypeHeaderValue("text/plain");
+                contentTypeHeader = DefaultContentType;
+                encoding = Encodings.UTF8EncodingWithoutBOM;
             }
             else
             {
-                contentTypeHeader = new MediaTypeHeaderValue(ContentType);
+                if(string.IsNullOrEmpty(contentTypeHeader.Charset))
```
pranavkm: Can this not go away?

src/Microsoft.AspNet.Mvc.Core/ActionResults/ContentResult.cs
```diff
@@ -26,17 +31,47 @@ public class ContentResult : ActionResult
...
 
-            MediaTypeHeaderValue contentTypeHeader;
-            if (string.IsNullOrEmpty(ContentType))
+            MediaTypeHeaderValue contentTypeHeader = ContentType;
+            Encoding encoding;
+            if(contentTypeHeader == null)
             {
-                contentTypeHeader = new MediaTypeHeaderValue("text/plain");
+                contentTypeHeader = DefaultContentType;
+                encoding = Encodings.UTF8EncodingWithoutBOM;
```
pranavkm: `Encoding.Utf8` should suffice.

src/Microsoft.AspNet.Mvc.Core/ActionResults/ContentResult.cs
```diff
@@ -11,11 +13,14 @@ namespace Microsoft.AspNet.Mvc
...
+        private readonly MediaTypeHeaderValue DefaultContentType = new MediaTypeHeaderValue("text/plain")
+        {
+            Encoding = Encoding.UTF8
+        };
 
-        public Encoding ContentEncoding { get; set; }
+        public string Content { get; set; }
 
-        public string ContentType { get; set; }
+        public MediaTypeHeaderValue ContentType { get; set; }
```
pranavkm: While you're touching these files, could you add docs?

src/Microsoft.AspNet.Mvc.Core/ActionResults/ContentResult.cs
```diff
@@ -46,7 +60,7 @@ public class ContentResult : ActionResult
 
             if (Content != null)
             {
-                await response.WriteAsync(Content, contentTypeHeader.Encoding);
+                await response.WriteAsync(Content, new ResponseEncodingWrapper(encoding));
```
pranavkm: I think it should be fine to flow the encoding through in this case

src/Microsoft.AspNet.Mvc.Core/ActionResults/ContentResult.cs
```diff
@@ -26,17 +31,26 @@ public class ContentResult : ActionResult
...
+                encoding = DefaultContentType.Encoding;
             }
             else
             {
-                contentTypeHeader = new MediaTypeHeaderValue(ContentType);
+                if (contentTypeHeader.Encoding == null)
+                {
+                    // 1. Do not modify the user supplied content type
+                    // 2. Parse here to handle parameters apart from charset
+                    contentTypeHeader = MediaTypeHeaderValue.Parse(contentTypeHeader.ToString());
```
pranavkm: I don't fully follow - why are we `ToString`ing this?

src/Microsoft.AspNet.Mvc.Core/ActionResults/ContentResult.cs
kichalla: :+1: ...when used with StreamWriter the preamble is being generated...will change it...

src/Microsoft.AspNet.Mvc.Core/ActionResults/FilePathResult.cs
```diff
@@ -27,15 +28,15 @@ public class FilePathResult : FileResult
...
-        /// the provided <paramref name="fileName"/>
+        /// the provided <paramref name="fileName"/> and the
+        /// provided <paramref name="contentType"/>.
         /// </summary>
         /// <param name="fileName">The path to the file. The path must be an absolute
         /// path. Relative and virtual paths are not supported.</param>
         /// <param name="contentType">The Content-Type header of the response.</param>
-        public FilePathResult([NotNull] string fileName)
-            : base(contentType: null)
+        public FilePathResult([NotNull] string fileName, string contentType)
```
pranavkm: Doesn't this remove the overload with `fileName`

src/Microsoft.AspNet.Mvc.Core/ActionResults/FilePathResult.cs
```diff
@@ -27,15 +28,15 @@ public class FilePathResult : FileResult
...
+            : this(fileName, new MediaTypeHeaderValue(contentType))
```
pranavkm: Will this throw if `contentType` is null?

src/Microsoft.AspNet.Mvc.Core/ActionResults/ContentResult.cs
kichalla: we do not want to modify the instance of MediaTypeHeaderValue value that the user supplied to ContentType property and also there could be additional parameters...test: https://github.com/aspnet/Mvc/pull/2450/files#diff-3be60edcf4b71964bc2ec1b25a0984fbR99

src/Microsoft.AspNet.Mvc.Core/ActionResults/FileStreamResult.cs
```diff
@@ -23,13 +24,14 @@ public class FileStreamResult : FileResult
...
-        /// the provided <paramref name="fileStream"/>.
+        /// the provided <paramref name="fileStream"/> and the
+        /// provided <paramref name="contentType"/>.
         /// </summary>
         /// <param name="fileStream">The stream with the file.</param>
-        public FileStreamResult([NotNull] Stream fileStream)
-            : base(contentType: null)
+        /// <param name="contentType">The Content-Type header of the response.</param>
+        public FileStreamResult([NotNull] Stream fileStream, string contentType)
+            : this(fileStream, new MediaTypeHeaderValue(contentType))
```
pranavkm: Again, is this intentional - the single overload ctor disappeared.

kichalla: yes it is intentional https://github.com/aspnet/Mvc/pull/2450/files#r29091219

src/Microsoft.AspNet.Mvc.Core/ResponseEncodingWrapper.cs
```diff
@@ -0,0 +1,55 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Text;
+
+namespace Microsoft.AspNet.Mvc
+{
+    /// <summary>
+    /// Encoding wrapper which makes preamble to be not required.
+    /// </summary>
+    internal class ResponseEncodingWrapper : Encoding
```
pranavkm: If we remove the usage from `ContentResult`, then make this a private type in `ViewExecutor`.

src/Microsoft.AspNet.Mvc.Core/ActionResults/FilePathResult.cs
kichalla: yes, it does

test/Microsoft.AspNet.Mvc.Core.Test/ActionResults/ContentResultTest.cs
```diff
@@ -73,45 +79,60 @@ public class ContentResultTest
...
+            string expectedContentType,
+            byte[] expectedContentData)
         {
             // Arrange
             var contentResult = new ContentResult
             {
-                Content = "Test Content",
-                ContentType = "some-type",
-                ContentEncoding = null
+                Content = "abcd",
```
pranavkm: Maybe have a string with some Utf8 chars so we can differentiate between ASCII and UTF8 in the test

test/Microsoft.AspNet.Mvc.Core.Test/ActionResults/FileContentResultTest.cs
```diff
@@ -45,5 +46,29 @@ public void Constructor_SetsFileContents()
             // Assert
             Assert.Equal(buffer, outStream.ToArray());
         }
+
+        [Fact]
+        public async Task SetsSuppliedContentTypeAndEncoding()
```
pranavkm: `ExecuteResultAsync_SetsSuppliedContentTypeAndEncoding`

test/Microsoft.AspNet.Mvc.Core.Test/ActionResults/FilePathResultTest.cs
```diff
@@ -112,6 +113,37 @@ public void Constructor_SetsFileName()
...
+
+            var context = new ActionContext(httpContext, new RouteData(), new ActionDescriptor());
+
+            // Act
+            await result.ExecuteResultAsync(context);
+            httpContext.Response.Body.Position = 0;
+
+            // Assert
+            Assert.NotNull(httpContext.Response.Body);
+            var contents = await new StreamReader(httpContext.Response.Body).ReadToEndAsync();
```
pranavkm: Maybe do `var contents = Encoding.ASCII.GetString(bodyStream.ToArray());`

test/Microsoft.AspNet.Mvc.Core.Test/ActionResults/FileResultTest.cs
```diff
@@ -22,7 +22,7 @@ public void Constructor_SetsContentType()
             var result = new EmptyFileResult("text/plain");
 
             // Assert
-            Assert.Equal("text/plain", result.ContentType);
+            Assert.Equal("text/plain", result.ContentType?.ToString());
```
pranavkm: `result.ContentType` cannot be null. Drop the `?.`

src/Microsoft.AspNet.Mvc.Core/ActionResults/ContentResult.cs
Tratcher: It's a .Clone hack. See https://github.com/aspnet/HttpAbstractions/issues/176

pranavkm: :+1:

src/Microsoft.AspNet.Mvc.Core/ActionResults/FilePathResult.cs
pranavkm: Maybe throw a `[NotNull]` on the argument instead?

pranavkm:
:watch:

test/Microsoft.AspNet.Mvc.Core.Test/ActionResults/FilePathResultTest.cs
kichalla: :+1:

kichalla:
Updated /cc: @pranavkm

test/Microsoft.AspNet.Mvc.Core.Test/ActionResults/FilePathResultTest.cs
```diff
@@ -12,6 +12,7 @@
 using Moq;
 using Xunit;
 using System.Text;
+using Microsoft.Net.Http.Headers;
```
ajaybhargavb: Sort.

test/Microsoft.AspNet.Mvc.Core.Test/ActionResults/ContentResultTest.cs
```diff
@@ -1,11 +1,14 @@
-﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+﻿
```
ajaybhargavb: Blank line.

test/Microsoft.AspNet.Mvc.Core.Test/ActionResults/FilePathResultTest.cs
```diff
@@ -112,6 +113,36 @@ public void Constructor_SetsFileName()
         }
 
         [Fact]
+        public async Task ExecuteResultAsync_SetsSuppliedContentTypeAndEncoding()
+        {
+            // Arrange
+            var expectedContentType = "text/foo; charset=us-ascii";
+            // path will be C:/.../TestFiles/FilePathResultTestFile.txt
```
pranavkm: Comment's out of date.

test/Microsoft.AspNet.Mvc.Core.Test/ActionResults/FilePathResultTest.cs
pranavkm: :100:

src/Microsoft.AspNet.Mvc.Core/ActionResults/ContentResult.cs
pranavkm: Bump.

src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewExecutor.cs
```diff
@@ -23,22 +29,44 @@ public static class ViewExecutor
...
 
-            actionContext.HttpContext.Response.ContentType = contentType;
-            var wrappedStream = new StreamWrapper(actionContext.HttpContext.Response.Body);
-            var encoding = Encodings.UTF8EncodingWithoutBOM;
-            using (var writer = new StreamWriter(wrappedStream, encoding, BufferSize, leaveOpen: true))
+            response.ContentType = contentTypeHeader.ToString();
+
+            var wrappedStream = new StreamWrapper(response.Body);
+
+            using (var writer = new StreamWriter(
```
pranavkm: Could you add a comment here calling out why we have a wrapper for the encoding?

pranavkm:
:watch: mostly done. There's a couple of un-addressed items. Would also like @rynowak to have a look at this?

kichalla:
Updated. /cc: @pranavkm @rynowak @ajaybhargavb

src/Microsoft.AspNet.Mvc.Core/ActionResults/ContentResult.cs
```diff
@@ -11,11 +13,17 @@ namespace Microsoft.AspNet.Mvc
...
     public class ContentResult : ActionResult
     {
-        public string Content { get; set; }
+        private readonly MediaTypeHeaderValue DefaultContentType = new MediaTypeHeaderValue("text/plain")
+        {
+            Encoding = Encoding.UTF8
+        };
 
-        public Encoding ContentEncoding { get; set; }
+        public string Content { get; set; }
```
pranavkm: Docs for this.

src/Microsoft.AspNet.Mvc.Core/ActionResults/FilePathResult.cs
pranavkm: bump.

src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewExecutor.cs
```diff
@@ -23,22 +29,47 @@ public static class ViewExecutor
...
 
-            actionContext.HttpContext.Response.ContentType = contentType;
-            var wrappedStream = new StreamWrapper(actionContext.HttpContext.Response.Body);
-            var encoding = Encodings.UTF8EncodingWithoutBOM;
-            using (var writer = new StreamWriter(wrappedStream, encoding, BufferSize, leaveOpen: true))
+            response.ContentType = contentTypeHeader.ToString();
+
+            var wrappedStream = new StreamWrapper(response.Body);
+
+            // StreamWriter writes the preamble or BOM bytes to the response if the encoding requires it.
```
pranavkm: :100:

pranavkm:
:shipit:. Try and get @rynowak to look at this before you merge it in.

src/Microsoft.AspNet.Mvc.Core/ActionResults/ContentResult.cs
```diff
@@ -11,11 +13,17 @@ namespace Microsoft.AspNet.Mvc
 {
     public class ContentResult : ActionResult
     {
-        public string Content { get; set; }
+        private readonly MediaTypeHeaderValue DefaultContentType = new MediaTypeHeaderValue("text/plain")
+        {
+            Encoding = Encoding.UTF8
+        };
 
-        public Encoding ContentEncoding { get; set; }
```
harshgMSFT: Are we removing this because ContentEncoding is included in `MediaTypeHeaderValue`?

kichalla: yeah, right...(as a side note...this property name is just confusing...as per Http this would represent the compression encoding like gzip, deflate)..

src/Microsoft.AspNet.Mvc.Core/ActionResults/ContentResult.cs
```diff
@@ -26,17 +34,26 @@ public class ContentResult : ActionResult
...
 
-            MediaTypeHeaderValue contentTypeHeader;
-            if (string.IsNullOrEmpty(ContentType))
+            var contentTypeHeader = ContentType;
+            Encoding encoding;
+            if (contentTypeHeader == null)
             {
-                contentTypeHeader = new MediaTypeHeaderValue("text/plain");
+                contentTypeHeader = DefaultContentType;
+                encoding = DefaultContentType.Encoding;
```
harshgMSFT: Is this same as what existing code does? what is the default encoding for MediaTypeHeaderValue("text/plain") ? Is it UTF8?

kichalla: yeah, right...its UTF8, but we are not specifying the BOM setting as it has no effect in this case, for example when writing content using StreamWriter it writes out the BOM if the encoding requires it, but we are using the following `response.WriteAsync` api which does not use StreamWriter...

https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNet.Http.Abstractions/HttpResponseWritingExtensions.cs#L39

src/Microsoft.AspNet.Mvc.Core/ActionResults/ContentResult.cs
```diff
@@ -11,11 +13,17 @@ namespace Microsoft.AspNet.Mvc
 {
     public class ContentResult : ActionResult
     {
-        public string Content { get; set; }
+        private readonly MediaTypeHeaderValue DefaultContentType = new MediaTypeHeaderValue("text/plain")
+        {
+            Encoding = Encoding.UTF8
```
rynowak: We've defined a static class with these (`Encodings`) - you want to use UTF8 **without BOM**

src/Microsoft.AspNet.Mvc.Core/ActionResults/ContentResult.cs
rynowak: :+1:

src/Microsoft.AspNet.Mvc.Core/ActionResults/ContentResult.cs
rynowak: We should use the correct encoding (without BOM) even if it doesn't have an impact. 

This breaks the cycle of people have to ask why we're using plain UTF8 and you having to explain it again.

src/Microsoft.AspNet.Mvc.Core/ActionResults/ContentResult.cs
```diff
@@ -26,17 +34,26 @@ public class ContentResult : ActionResult
...
             }
             else
             {
-                contentTypeHeader = new MediaTypeHeaderValue(ContentType);
+                if (contentTypeHeader.Encoding == null)
+                {
+                    // 1. Do not modify the user supplied content type
+                    // 2. Parse here to handle parameters apart from charset
+                    contentTypeHeader = MediaTypeHeaderValue.Parse(contentTypeHeader.ToString());
+                    contentTypeHeader.Encoding = Encoding.UTF8;
```
rynowak: If you're creating a `DefaultMediaType` then also create a `DefaultEncoding`

src/Microsoft.AspNet.Mvc.Core/ActionResults/ContentResult.cs
```diff
@@ -26,17 +34,26 @@ public class ContentResult : ActionResult
...
+                }
+
+                encoding = contentTypeHeader.Encoding;
             }
 
-            contentTypeHeader.Encoding = ContentEncoding ?? Encodings.UTF8EncodingWithoutBOM;
```
rynowak: Please don't change the code to be less correct :laughing:

src/Microsoft.AspNet.Mvc.Core/ActionResults/FileContentResult.cs
```diff
@@ -19,13 +20,14 @@ public class FileContentResult : FileResult
...
+        /// <param name="contentType">The Content-Type header of the response.</param>
+        public FileContentResult([NotNull] byte[] fileContents, string contentType)
```
rynowak: `[NotNull]`

src/Microsoft.AspNet.Mvc.Core/ActionResults/FileContentResult.cs
```diff
@@ -35,7 +37,7 @@ public FileContentResult([NotNull] byte[] fileContents)
         /// </summary>
         /// <param name="fileContents">The bytes that represent the file contents.</param>
         /// <param name="contentType">The Content-Type header of the response.</param>
-        public FileContentResult([NotNull] byte[] fileContents, string contentType)
+        public FileContentResult([NotNull] byte[] fileContents, MediaTypeHeaderValue contentType)
```
rynowak: `[NotNull]`

src/Microsoft.AspNet.Mvc.Core/ActionResults/FileContentResult.cs
rynowak: Please review all of these and add `[NotNull]`. The idea is to throw as close as possible to where the mistake occurred.

src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewExecutor.cs
```diff
@@ -14,7 +17,10 @@ namespace Microsoft.AspNet.Mvc
     public static class ViewExecutor
     {
         private const int BufferSize = 1024;
-        private const string ContentType = "text/html; charset=utf-8";
+        private static readonly MediaTypeHeaderValue DefaultContentType = new MediaTypeHeaderValue("text/html")
+        {
+            Encoding = Encoding.UTF8
+        };
```
rynowak: same comments here

src/Microsoft.AspNet.Mvc.Core/ActionResults/ContentResult.cs
harshgMSFT: :+1: with what @rynowak  said.

src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewExecutor.cs
```diff
@@ -23,22 +29,47 @@ public static class ViewExecutor
...
+            var wrappedStream = new StreamWrapper(response.Body);
+
+            // StreamWriter writes the preamble or BOM bytes to the response if the encoding requires it.
+            // Since preamble bytes are unnecessary when generating dynamic content, wrap the original encoding
+            // to avoid writing the preamble bytes.
+            using (var writer = new StreamWriter(
+                wrappedStream,
+                new ResponseEncodingWrapper(encoding),
+                BufferSize,
+                leaveOpen: true))
```
rynowak: I don't understand why we're doing any of this instead of just using the correct encoding.

rynowak:
:watch:

src/Microsoft.AspNet.Mvc.Core/ActionResults/FileContentResult.cs
harshgMSFT: This seems to be like a compatiblity bug fix - should have been dealt with as a separate PR. (Its fine to leave this here for this one, but something to keep in mind in general ).

src/Microsoft.AspNet.Mvc.Core/ActionResults/FilePathResult.cs
harshgMSFT: again use `[NotNull]` if this is required.

src/Microsoft.AspNet.Mvc.Core/ActionResults/FilePathResult.cs
```diff
@@ -46,7 +47,7 @@ public FilePathResult([NotNull] string fileName)
         /// <param name="fileName">The path to the file. The path must be an absolute
         /// path. Relative and virtual paths are not supported.</param>
         /// <param name="contentType">The Content-Type header of the response.</param>
-        public FilePathResult([NotNull] string fileName, string contentType)
+        public FilePathResult([NotNull] string fileName, MediaTypeHeaderValue contentType)
```
harshgMSFT: `[NotNull]`

src/Microsoft.AspNet.Mvc.Core/ActionResults/FileResult.cs
```diff
@@ -24,15 +24,25 @@ public abstract class FileResult : ActionResult
...
         {
             ContentType = contentType;
         }
 
         /// <summary>
-        /// Gets or sets the Content-Type header value that will be written to the response.
+        /// Gets the <see cref="MediaTypeHeaderValue"/> representing the Content-Type header of the response.
         /// </summary>
-        public string ContentType { get; set; }
+        public MediaTypeHeaderValue ContentType { get; }
```
harshgMSFT: Is it intentional to make this readonly property ?

src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewExecutor.cs
```diff
@@ -14,7 +17,10 @@ namespace Microsoft.AspNet.Mvc
     public static class ViewExecutor
     {
         private const int BufferSize = 1024;
-        private const string ContentType = "text/html; charset=utf-8";
+        private static readonly MediaTypeHeaderValue DefaultContentType = new MediaTypeHeaderValue("text/html")
+        {
+            Encoding = Encoding.UTF8
```
harshgMSFT: Same comments as earlier : Use UTF8 Without BOM and use the static helper we have.

test/Microsoft.AspNet.Mvc.Core.Test/ActionResults/ContentResultTest.cs
```diff
@@ -73,45 +77,65 @@ public class ContentResultTest
...
+                    }
+                };
+            }
         }
 
-        [Fact]
-        public async Task ContentResult_Response_BadContentType_ThrowsFormatException()
+        [Theory]
+        [MemberData(nameof(ContentResultContentTypeData))]
+        public async Task ContentResult_SetContentTypeAndEncoding(
```
harshgMSFT: Doesnt tell me what are we testing here?

harshgMSFT:
:watch:

src/Microsoft.AspNet.Mvc.Core/ActionResults/FileResult.cs
Eilon: Yeah I think it's intentional because the underlying property on the `MediaTypeHeaderValue` is get-only. It's passed into the ctor of that type and is read-only. So this is a minor breaking change from MVC 5 (or Web API 2 or something), but we discussed on Friday and it's fine.

src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewExecutor.cs
```diff
@@ -23,21 +29,43 @@ public static class ViewExecutor
...
+            }
+
+            response.ContentType = contentTypeHeader.ToString();
+
+            var wrappedStream = new StreamWrapper(response.Body);
 
-            actionContext.HttpContext.Response.ContentType = contentType;
-            var wrappedStream = new StreamWrapper(actionContext.HttpContext.Response.Body);
-            var encoding = Encodings.UTF8EncodingWithoutBOM;
             using (var writer = new StreamWriter(wrappedStream, encoding, BufferSize, leaveOpen: true))
```
kichalla: As discussed, filed: https://github.com/aspnet/Mvc/issues/2496

kichalla:
Updated as per our discussion. /cc: @pranavkm @rynowak @harshgMSFT

test/Microsoft.AspNet.Mvc.Core.Test/ActionResults/ContentResultTest.cs
kichalla: I think the name is clear enough?

harshgMSFT: `ContentResult_ExecuteResultAsync_SetContentTypeAndEncoding_OnResponse`

src/Microsoft.AspNet.Mvc.Core/ActionResults/FileResult.cs
harshgMSFT: :ok:

rynowak:
:shipit:

harshgMSFT:
Sorry for commenting on the commit rather than on PR. :shipit: after addressing the comments.

PR #2452, opened 2015-04-27
React to aspnet/Razor#182 fix

dougbu:
- use `[HtmlAttributeNotBound]` to make tag helper properties `public`
- properties remain invisible in Razor

### DIFF
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/AnchorTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/AnchorTagHelper.cs
@@ -24,9 +24,8 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
         private const string RouteAttributePrefix = "asp-route-";
         private const string Href = "href";
 
-        // Protected to ensure subclasses are correctly activated. Internal for ease of use when testing.
-        [Activate]
-        protected internal IHtmlGenerator Generator { get; set; }
+        [Activate, HtmlAttributeNotBound]
+        public IHtmlGenerator Generator { get; set; }
 
         /// <summary>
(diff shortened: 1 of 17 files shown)


NTaylorMullen:
:shipit:

dougbu:
c162209

src/Microsoft.AspNet.Mvc.TagHelpers/AnchorTagHelper.cs
```diff
@@ -24,9 +24,8 @@ public class AnchorTagHelper : TagHelper
         private const string RouteAttributePrefix = "asp-route-";
         private const string Href = "href";
 
-        // Protected to ensure subclasses are correctly activated. Internal for ease of use when testing.
-        [Activate]
-        protected internal IHtmlGenerator Generator { get; set; }
+        [Activate, HtmlAttributeNotBound]
```
Eilon: Eh, one attribute per line maybe?

Eilon: (And each in their own brackets.)

dougbu: @Eilon is that a guideline and worth an additional commit / PR?

Eilon: Not a guideline, but it's what we always do. Yes, please clean up. (A simple regex can fix it.)

dougbu: see PR #2457

PR #2453, opened 2015-04-27
Create Mvc.Abstractions

rynowak:
Creates a new package 'Microsoft.AspNet.Mvc.Abstractions' which defines
APIs and contracts for core concepts and extensibility points in MVC.

Includes:
- ModelBinding
- Validation
- Model State
- Model Metadata
- Action Descriptors
- IActionResult
- Filters
- IActionConstraint


Eilon:
Very nice! :shipit:

PR #2454, opened 2015-04-27
Merging ModelBinding into Mvc.Core

rynowak:


src/Microsoft.AspNet.Mvc.Core/ModelBinding/PrefixContainer.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using Microsoft.Framework.Internal;
 
 namespace Microsoft.AspNet.Mvc.ModelBinding.Internal
```
pranavkm: Could you move this to a `Internal` subdirectory \ remove the `Internal` namespace?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/PrefixContainer.cs
```diff
@@ -187,6 +188,36 @@ internal static bool IsPrefixMatch(string prefix, string testString)
             }
         }
 
+        /// <summary>
+        /// Convert an ICollection to an array, removing null values. Fast path for case where
+        /// there are no null values.
+        /// </summary>
+        private static T[] ToArrayWithoutNulls<T>(ICollection<T> collection) where T : class
```
pranavkm: `TVal`

Eilon:
Also very nice...

src/Microsoft.AspNet.Mvc.Core/ModelBinding/PrefixContainer.cs
rynowak: renaming to `TElement` which we use in general in the collection model binders.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/PrefixContainer.cs
rynowak: Making this public and not in the internal namespace. You'd need this to implement a value provider correctly.

Issue #2336, opened 2015-04-04
Support generating `<input/>` elements without a `value` attribute


dougbu:
HTML generation of an `<input/>` element with `type="file"` or `type="image"` should be spec-compliant.  HTML 5 says "The element's `value` attribute must be omitted." for these two types.  but current HTML helpers and the `<input/>` tag helper unconditionally add (or pass through) a `value` attribute.

suggest adding special cases to our MVC helpers for this issue.

TL;DR
while it's possible to add another `<input/>` tag helper that only performs this fix-up, similar compositions with `@Html.TextBox()` aren't available.  probably better to get the HTML correct up front in any case.

Issue #453, opened 2014-05-10
Unit tests for HtmlHelper


NTaylorMullen:
Have added quick confirmations to views in MVC Sample.  Need to add real unit tests, preferably copying over the bulk of HtmlHelper / extensions tests from legacy MVC.  Skip the NRE checks :)

| 170 | 10328 | Reported By: Doug Bunting (AAPT) |

yishaigalatzer:
The idea is to complete porting over the HtmlHelper tests and see if we come up with any new bugs. The code doesn't have to be comitted into the Beta3 branch but the work has to happen before we are shipping Beta3 to find any issues we might have missed.

dougbu:
@pranavkm has greatly improved the coverage of many HTML helpers.  some remaining gaps:
- functional tests of HTML helpers (not exactly matching the bug title but also not covered elsewhere)
- `@Html.DropDownList()` / `@Html.ListBox()`
- `@Html.RadioButton()`
- `@Html.Raw()`
- `@Html.TextArea()`
- `@Html.TextBox()` (yup, no kidding)
- `@Html.ValidationMessage()` / `@Html.ValidationSummary()`

a few other methods are exposed publicly but used in other helpers (and therefore tested indirectly, or will be):
- `@Html.Encode()`
- `@Html.FormatValue()`
- `@Html.GenerateIdFromName()`
- `@Html.GetClientValidationRules()`
- `@Html.GetFormMethodString()`

dougbu:
Crossed out a couple of items in earlier comments.

Some details on remaining `DropDownList` and `ListBox` test gaps.  Test do not currently:
- use non-`null` `optionLabel` and `htmlAttributes` arguments
- confirm value sources and their priorities
- mock an `IHtmlGenerator` and confirm how it is called
- mock an `IHtmlHelper` and confirm how extension methods call that

ajaybhargavb:
Crossed out some items from previous comment. Following are the remaining items.
- `@Html.Raw()`
- Indexers and deeply nested lambdas for the various `*For` e.g `InputFor`, `DisplayFor` etc
- Scenarios gap compared to 5.2
- a few other methods are exposed publicly but used in other helpers (and therefore tested indirectly, or will be). Need to write a few tests to cover usage that isn't already handled indirectly.:
 - `@Html.Encode()`
 - `@Html.FormatValue()`
 - `@Html.GenerateIdFromName()`
 - `@Html.GetClientValidationRules()`
[...]

PR #2458, opened 2015-04-27
Support for generating input elements without 'value' attribute

ajaybhargavb:
Issue - #2336 
@dougbu

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
@@ -1024,7 +1024,23 @@ namespace Microsoft.AspNet.Mvc.Rendering
                         attributeValue = useViewData ? EvalString(viewContext, fullName, format) : valueParameter;
                     }
 
-                    tagBuilder.MergeAttribute("value", attributeValue, replaceExisting: isExplicitValue);
+                    var addValue = true;
+                    object typeAttributeValue;
+                    if (htmlAttributes != null && htmlAttributes.TryGetValue("type", out typeAttributeValue))
+                    {
+                        if (string.Equals(typeAttributeValue.ToString(), "file", StringComparison.OrdinalIgnoreCase) ||
+                            string.Equals(typeAttributeValue.ToString(), "image", StringComparison.OrdinalIgnoreCase))
+                        {
+                            // 'value' attribute is not needed for 'file' and 'image' input types.
+                            addValue = false;
+                        }
+                    }
+
+                    if (addValue)
+                    {
+                        tagBuilder.MergeAttribute("value", attributeValue, replaceExisting: isExplicitValue);
+                    }
+
                     break;
             }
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -1024,6 +1024,16 @@ internal static object GetModelStateValue(ViewContext viewContext, string key, T
                         attributeValue = useViewData ? EvalString(viewContext, fullName, format) : valueParameter;
                     }
 
+                    object typeAttributeValue = null;
```
dougbu: this added block is going to confuse some future developer since multiple breaks from the same `case` is unusual.  suggest reversing the tests (in some way) or adding a `var addValue = true;` to control merging the `value` attribute.

also need a comment explaining why not adding the `value` attribute is correct in these cases.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -1024,6 +1024,16 @@ internal static object GetModelStateValue(ViewContext viewContext, string key, T
...
+                    if (htmlAttributes != null && htmlAttributes.TryGetValue("type", out typeAttributeValue))
+                    {
+                        if (string.Equals(typeAttributeValue.ToString(), "file") ||
+                            string.Equals(typeAttributeValue.ToString(), "image"))
```
dougbu: add `StringComparison.OrdinalIgnoreCase`.  this comparison should not be case-sensitive (and we are always explicit about comparison types in product code).

test/Microsoft.AspNet.Mvc.TagHelpers.Test/InputTagHelperTest.cs
```diff
@@ -697,13 +697,13 @@ public class InputTagHelperTest
 
             var tagBuilder = new TagBuilder("input", new NullTestEncoder());
 
-            Dictionary<string, object> htmlAttributes = null;
+            var htmlAttributes = new Dictionary<string, object>
```
dougbu: add an `image` test or two to the data set for this method

dougbu:
work with @ianhong to track testing this new code when we get to `Html.TextBox()` in #453

dougbu:
:watch:

test/Microsoft.AspNet.Mvc.TagHelpers.Test/InputTagHelperTest.cs
ajaybhargavb: Added.

ajaybhargavb:
Updated.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -1024,7 +1024,23 @@ internal static object GetModelStateValue(ViewContext viewContext, string key, T
                         attributeValue = useViewData ? EvalString(viewContext, fullName, format) : valueParameter;
                     }
 
-                    tagBuilder.MergeAttribute("value", attributeValue, replaceExisting: isExplicitValue);
+                    var addValue = true;
+                    object typeAttributeValue = null;
```
dougbu: hmm, used only if `TryGetValue` succeeds. if the compiler lets you, remove the initialization.

dougbu:
:shipit: after one small change

PR #2463, opened 2015-04-28
Allow returning precompiled views without sources

### LINKED ISSUE #2462: MVC ignores precompiled views where it cannot find the source file
CompilerCache.cs: line 115-124

```csharp
                // For precompiled views, the first time the entry is read, we need to ensure that no changes were made
                // either to the file associated with this entry, or any _GlobalImport associated with it between the time
                // the View was precompiled and the time EnsureInitialized was called. For later iterations, we can
                // rely on expiration triggers ensuring the validity of the entry.

                var fileInfo = _fileProvider.GetFileInfo(relativePath);
                if (!fileInfo.Exists)
                {
                    return null;
                }
```

As can be seen, if the original source file of a precompiled razor view does not exist (or is not found), it ignores it.

Alxandr:
Resolves #2462

### DIFF
--- a/test/Microsoft.AspNet.Mvc.FunctionalTests/PrecompilationTest.cs
+++ b/test/Microsoft.AspNet.Mvc.FunctionalTests/PrecompilationTest.cs
@@ -190,60 +190,20 @@ namespace Microsoft.AspNet.Mvc.FunctionalTests
             
             try
             {
-                // Act - 1
+                // Act
+                await DeleteFile(viewsDirectory, "_ViewStart.cshtml");
+                await DeleteFile(viewsDirectory, "_Layout.cshtml");
+                await DeleteFile(viewsDirectory, "Index.cshtml");
+            
                 var response = await client.GetAsync("http://localhost/Home/Index");
                 var responseContent = await response.Content.ReadAsStringAsync();
-
-                // Assert - 1
+                
+                // Assert
                 Assert.Equal(HttpStatusCode.OK, response.StatusCode);
                 var parsedResponse1 = new ParsedResponse(responseContent);
                 Assert.StartsWith(assemblyNamePrefix, parsedResponse1.ViewStart);
                 Assert.StartsWith(assemblyNamePrefix, parsedResponse1.Layout);
                 Assert.StartsWith(assemblyNamePrefix, parsedResponse1.Index);
-
-                // Act - 2
-                // Delete the Layout file and verify it is still served.
-                await DeleteFile(viewsDirectory, "Layout.cshtml");
-                responseContent = await client.GetStringAsync("http://localhost/Home/Index");
-
-                // Assert - 2
-                var response2 = new ParsedResponse(responseContent);
-                Assert.StartsWith(assemblyNamePrefix, response2.ViewStart);
-                Assert.StartsWith(assemblyNamePrefix, response2.Index);
-                Assert.StartsWith(assemblyNamePrefix, response2.Layout);
-
-                // Act - 3
-                // Delete the _ViewStart file and verify it is still served.
-                await DeleteFile(viewsDirectory, "_ViewStart.cshtml");
-                responseContent = await client.GetStringAsync("http://localhost/Home/Index");
-
-                // Assert - 3
-                var response3 = new ParsedResponse(responseContent);
-                Assert.StartsWith(assemblyNamePrefix, response3.ViewStart);
-                Assert.StartsWith(response2.Index, response3.Index);
-                Assert.StartsWith(response2.Layout, response3.Layout);
-
-                // Act - 4
-                // Delete the _GlobalImport file and verify it is still served.
-                await DeleteFile(viewsDirectory, "_GlobalImport.cshtml");
-                responseContent = await client.GetStringAsync("http://localhost/Home/Index");
-
-                // Assert - 4
-                var response4 = new ParsedResponse(responseContent);
-                Assert.StartsWith(response3.ViewStart, response4.ViewStart);
-                Assert.StartsWith(response3.Index, response4.Index);
-                Assert.StartsWith(response3.Layout, response4.Layout);
-
-                // Act - 5
-                // Delete the Index file and verify it is still served.
-                await DeleteFile(viewsDirectory, "Index.cshtml");
-                responseContent = await client.GetStringAsync("http://localhost/Home/Index");
-
-                // Assert - 5
-                var response5 = new ParsedResponse(responseContent);
-                Assert.StartsWith(response4.Layout, response5.Layout);
-                Assert.StartsWith(response4.ViewStart, response5.ViewStart);
-                Assert.StartsWith(response4.Index, response5.Index);
             }
             finally
             {
--- a/test/Microsoft.AspNet.Mvc.FunctionalTests/PrecompilationTest.cs
+++ b/test/Microsoft.AspNet.Mvc.FunctionalTests/PrecompilationTest.cs
@@ -162,6 +162,97 @@ namespace Microsoft.AspNet.Mvc.FunctionalTests
                 File.WriteAllText(Path.Combine(viewsDirectory, "_GlobalImport.cshtml"), globalContent.TrimEnd(' '));
             }
         }
+        
+        [Fact]
+        public async Task PrecompiledView_RendersWithoutSourceFile()
+        {
+            // Arrange
+            IServiceCollection serviceCollection = null;
+            var server = TestHelper.CreateServer(_app, SiteName, services =>
+            {
+                _configureServices(services);
+                serviceCollection = services;
+            });
+            var client = server.CreateClient();
+
+            var serviceProvider = serviceCollection.BuildServiceProvider();
+            var applicationEnvironment = serviceProvider.GetRequiredService<IApplicationEnvironment>();
+
+            var viewsDirectory = Path.Combine(applicationEnvironment.ApplicationBasePath, "Views", "Home");
+            var layoutContent = File.ReadAllText(Path.Combine(viewsDirectory, "Layout.cshtml"));
+            var indexContent = File.ReadAllText(Path.Combine(viewsDirectory, "Index.cshtml"));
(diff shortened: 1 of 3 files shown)


msotclas:
Hi __@Alxandr__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <span>You've already signed the contribution license agreement. Thanks!</span>
        <p>The agreement was validated by Microsoft Open Technologies, Inc. and real humans are currently evaluating your PR.</p>

TTYL, MSOTBOT;

Eilon:
@pranavkm can you review?

pranavkm:
The change looks good although we should add a functional test to ensure this behaves correctly. See - https://github.com/aspnet/Mvc/blob/dev/test/Microsoft.AspNet.Mvc.FunctionalTests/PrecompilationTest.cs.

The test would do something along the lines of deleting the `chstml` file on disk prior to making the request to the page and see that it uses the precompiled view.

Alxandr:
My usecase was having the views in another project and precompiling them, though deleting the `cshtml` would produce the same effect.

@pranavkm do you want me to add these tests?

pranavkm:
> though deleting the cshtml would produce the same effect.

That should suffice - fewer test webapplications to manage.

Alxandr:
@pranavkm I've added unit tests, but I can't run them, because MVC doesn't build (it didn't build when I started making changes).

pranavkm:
@Alxandr our dev feeds are a bit busted at this point (should be good soon). In the meanwhile, you could try pointing to the volatile feed and retry building.

Edit `~/NuGet.config` 

```
-    <add key="AspNetVNext" value="https://www.myget.org/F/aspnetvnext/api/v2" />
+    <add key="AspNetVNext" value="https://www.myget.org/F/aspnetvolatile/api/v2" />
```

davidfowl:
Ugh, dont point to the volatile feed.

pranavkm:
Well then, ignore my suggestion :joy_cat:

test/Microsoft.AspNet.Mvc.FunctionalTests/PrecompilationTest.cs
```diff
@@ -162,6 +162,97 @@ public class PrecompilationTest
...
+
+                // Assert - 5
+                var response5 = new ParsedResponse(responseContent);
+                Assert.StartsWith(response4.Layout, response5.Layout);
+                Assert.StartsWith(response4.ViewStart, response5.ViewStart);
+                Assert.StartsWith(response4.Index, response5.Index);
+            }
+            finally
+            {
+                File.WriteAllText(Path.Combine(viewsDirectory, "Layout.cshtml"), layoutContent.TrimEnd(' '));
```
Eilon: What's with the trimming at the end?

Eilon:
BTW for the record I think it's weird for a test to delete files that are checked in, and to then attempt to write them back to disk... no?

pranavkm:
@Eilon I agree. We should probably revisit how these tests are done. We have other tests in this test area that edit files on disk and they are problematic to run in parallel (dnx451 + dnxcore50).

test/Microsoft.AspNet.Mvc.FunctionalTests/PrecompilationTest.cs
```diff
@@ -162,6 +162,97 @@ public class PrecompilationTest
...
+                // Assert - 1
+                Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+                var parsedResponse1 = new ParsedResponse(responseContent);
+                Assert.StartsWith(assemblyNamePrefix, parsedResponse1.ViewStart);
+                Assert.StartsWith(assemblyNamePrefix, parsedResponse1.Layout);
+                Assert.StartsWith(assemblyNamePrefix, parsedResponse1.Index);
+
+                // Act - 2
+                // Delete the Layout file and verify it is still served.
+                await DeleteFile(viewsDirectory, "Layout.cshtml");
```
pranavkm: This doesn't actually cover the change you made in the code path. The first time the view is served, `IsValidatedPreCompiled` is set to `true` - so we don't use the code path that you made the change in. What I don't follow though is that we're serving views that have been deleted from disk after we added file watchers on it. That sounds like a bug and might be something we need to address.

What your test might need to do would be closer along the lines of

```C#
try
{
   // Act
   await DeleteFile(viewsDirectory, "_ViewStart.cshtml");
   await DeleteFile(viewsDirectory, "_Layout.cshtml");
   await DeleteFile(viewsDirectory, "Index.cshtml");

   var response = await client.GetAsync("http://localhost/Home/Index");

   // Assert
  
    Assert.Equal(HttpStatusCode.OK, response.StatusCode);
    var parsedResponse1 = new ParsedResponse(responseContent);
    Assert.StartsWith(assemblyNamePrefix, parsedResponse1.ViewStart);
    Assert.StartsWith(assemblyNamePrefix, parsedResponse1.Layout);
    Assert.StartsWith(assemblyNamePrefix, parsedResponse1.Index);
```

pranavkm:
:watch:

Alxandr:
I think it might be possible to inject a `IFileSystem`. That would obviously be mock-able.

pranavkm:
We do have unit tests for this type - https://github.com/aspnet/Mvc/blob/dev/test/Microsoft.AspNet.Mvc.Razor.Test/Compilation/CompilerCacheTest.cs - perhaps we can live with the absence of functional tests until we can sort out a plan for not doing disk IO. @Eilon thoughts?

Eilon:
Doing disk IO might be fine if it's perhaps not editing files that are part of the repo. For example, can the test copy a folder with a sample site to another folder, do whatever it wants (e.g. edit files, delete files), then blow away the whole thing? And even put that whole folder in the `.gitignore` to ensure it never gets checked in even if the test fails to clean up?

Alxandr:
Do I need to sign a new CLA?

Anyways, should I revert the commit that added unit tests?

pranavkm:
I think we can leave the functional test you had as is - it just needs to be fixed. See https://github.com/aspnet/Mvc/pull/2463#discussion_r29489672.

test/Microsoft.AspNet.Mvc.FunctionalTests/PrecompilationTest.cs
Alxandr: With regards to the watchers (if I'm reading this right), the source-files are never on disk to begin with (as far as MVC is concerned). They are precompiled in another project, so the paths MVC looks for them are non-existing. There is no `Views` directory at all.

pranavkm: But they do exist in this case, right? Moving the deletion further up would more closely resemble your scenario?

Alxandr:
Turns out it's quite hard to try to fix this thing when I can never get Mvc to build on my machine. Any good commits I can use as a starting point?

pranavkm:
@Alxandr, what issues are you running into?  The dev feed's been stable for some time now.

Alxandr:
Well, after deleting my `.dnx` folder, and re-downloading MVC, I managed to actually get it to compile.

Yet it crashes when trying to launch the xunit tests.

image

Alxandr:
Ok, so I got it compiling. I found a timezone (or similar) bug in MVC though, but I'll be reporting that in a separate issue. I also rebased off `dev`, so it should (hopefully) build in the CI.

pranavkm:
Looks good. I'll have this merged in.

pranavkm:
@Alxandr looks like this broke one of our existing tests (see AppVeyor) - DeletingPrecompiledGlobalFile_PriorToFirstRequestToAView_CausesViewToBeRecompiled. 

@Eilon, suggestions on what we should do here - the test is trying to verify a scenario that is contrary to the change @Alxandr made here. The scenario:

1. Create an application precomilation that has `_GlobalImport.cshtml` and `Index.cshtml`.
2. Launch the application. 
3. Decide that you don't need `_GlobalImport`.Delete it prior to requesting `Index.cshtml`.

Should the view now recompile? Without the change, it does.

Eilon:
@pranavkm logically that should work, no? If you change a dependency of a CSHTML file (e.g. by changing the global imports), the CSHTML file should recompile...

pranavkm:
@Alxandr we have a design meeting tomorrow afternoon to decide some open questions about precompilation. The other test might no longer be relevant once based on what we land on. I'll update this thread once we have a decision on this.

Eilon:
@Alxandr if you'd like to attend this design discussion meeting I can send you a link to join us. I see you're in Norway so it might be a bit late for you though. The meeting is at 3pm Pacific Time, which is midnight for you.

If you're interested, let me know and I'll send you an invite code.

Alxandr:
Midnight in 20 hrs, or midnight 4hrs ago? (sorry, but terms like "today" and "tomorrow" get especially confusing when operating on different timezones when it's around or just passed midnight ^^).

Eilon:
Oh sorry :smile: It's at 3pm Pacific Time 5/12/2015, and 12am Norway Time 5/13/2015 (that is, very, very early on Wednesday).

Eilon:
I sent you an invite in case you can attend.

panesofglass:
@Eilon, did you happen to reach a decision on this PR?

Eilon:
@panesofglass yeah we had a meeting with several folks from the team, plus @Alxandr and we put down our notes here: https://github.com/aspnet/Mvc/issues/2551

If you'd like, please have a read through that and any extra notes/questions/comments you have.

Right now the team is trying to wrap up the beta5 milestone, plus this weekend in the US is a long weekend (Sat-Sun-Mon off), and several folks are turning it into an extra-extra-long weekend (Fri-Tue off!), so I hope that later next week we can pick this up again and come up with a final design.

pranavkm:
This should be fixed as part of https://github.com/aspnet/Mvc/commit/dfacd2543b56499f220de0f3f205d5e78644241f.

PR #2467, opened 2015-04-28
Simple ModelBinders and Simple ModelBinder Poco - 1

harshgMSFT:
Covers simple scenario for each model binder.
Covers scenarios mixing a POCO model binder -> Simple Model binder.

This contains tests for
HeaderModel
ServicesModelBinder
CancellationTokenModelBinder
ByteArrayModelBinder
FormFileModelBinder

Part 2 Will contain similar tests for
FormCollectionModelBinder
BinderTypeBasedModelBinder
TypeConverterModelBinder
TypeMatchModelBinder
Any leftovers for BodyModelBinder


harshgMSFT:
@rynowak @dougbu

test/Microsoft.AspNet.Mvc.IntegrationTests/BodyValidationIntegrationTests.cs
```diff
@@ -42,11 +42,18 @@ private class Address
...
-            var httpContext = operationContext.HttpContext;
-
-            ConfigureHttpRequest(httpContext.Request, string.Empty);
-            var modelState = new ModelStateDictionary();
+            var operationContext = ModelBindingTestHelper.GetOperationBindingContext(
+              request =>
+              {
+                  request.Body = new MemoryStream(Encoding.UTF8.GetBytes("{ \"Id\":1234 }"));
+                  request.ContentType = "application/json";
+              });
```
rynowak: This piece seems OK. I did something simpler in my tests, and just create/pass-in the `HttpContext` to `GetOperationBindingContext`

test/Microsoft.AspNet.Mvc.IntegrationTests/BodyValidationIntegrationTests.cs
```diff
@@ -42,11 +42,18 @@ private class Address
...
+
+            var actionContext = operationContext
+                .HttpContext
+                .RequestServices
+                .GetRequiredService<IScopedInstance<ActionContext>>().Value;
+            var modelState = actionContext.ModelState;
```
rynowak: Really not a fan of this showing up in tests. What's the need for this? Is this because formatters use the `IScopedInstance<ActionContext>` to access model state? 

This is probably a bug in TryUpdateModel scenarios. See `ModelBindingHelper.TryUpdateModel` - there's an overload which takes a `ModelStateDictionary`

rynowak: As an alternative can we just have the tests initialize things in the order they need to go?

1). Create HttpContext
2). Create ActionContext
3). Create OperationBindingContext (forces service initialization)

test/Microsoft.AspNet.Mvc.IntegrationTests/ByteArrayModelBinderIntegrationTest.cs
```diff
@@ -0,0 +1,166 @@
...
+            // ModelBindingResult
+            Assert.Null(modelBindingResult);
+
+            // ModelState
+            Assert.True(modelState.IsValid);
+            Assert.Empty(modelState.Keys);
+        }
+
+        [Fact(Skip = "ModelState.Value not set due to #2445, #2446")]
+        public async Task BindParameter_WithData_GetsBound()
```
rynowak: Would like to also see one of these at the parameter level with no data.

test/Microsoft.AspNet.Mvc.IntegrationTests/BodyValidationIntegrationTests.cs
harshgMSFT: >Really not a fan of this showing up in tests. What's the need for this? Is this because formatters use the IScopedInstance<ActionContext> to access model state?

Yes- And yes that is a bug that we intend to fix ( Json Formatter would not add errors using model state ). This would be cleaned up when we do that .. 

>As an alternative can we just have the tests initialize things in the order they need to go?

I would like the tests to avoid handling multiple contexts, personally i like the delegate approach.

rynowak:
:shipit: after adding one more test for byte-array

test/Microsoft.AspNet.Mvc.IntegrationTests/ByteArrayModelBinderIntegrationTest.cs
harshgMSFT: `BindParameter_NoData_DoesNotGetBound` is already there.

rynowak: Then there should be one with no data at the property level :laughing:

Issue #1885, opened 2015-01-25
Consider adding a UseMvcWithDefaultRoute extension on BuilderExtensions that adds default routes to the route collection


pranavkm:
Based on the discussion here - https://github.com/aspnet/Mvc/pull/1880/files#r23486914.

This would be a convenience method that is essentially what `UseMvc` did prior to it being removed as part of https://github.com/aspnet/Mvc/pull/1880

pranavkm:
cc @Eilon @dougbu @yishaigalatzer

Eilon:
I'd like to state that it is critical that trivial apps not require a bunch of setup code.

Attribute routing solves this to a good degree. For conventional routing users it's now a bit of a hassle.

If we think attribute routing is the recommended approach for the "getting started" scenario, then I'd concede that the importance of this feature is lower. (And, it can be added at any time.)

yishaigalatzer:
The work involved in building this feature is negligible. Just need a good name.

-----Original Message-----

Eilon:
Some ideas:

1. `.UseMvcWithDefaultRoute()`
2. `.UseMvcConvention()` / `.UseMvcConventions()`
3. `.UseMvcWithConvention()` / `.UseMvcWithConventions()`
4. `.UseMvcAuto()`
5. `.UseMvcDefault()` / `.UseMvcDefaults()`

pranavkm:
We use the term `Conventions` for other things in Mvc - https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/ApplicationModels/ApplicationModelConventions.cs. Might result in confusiing term overload.

Eilon:
@ajaybhargavb let's go with the name `UseMvcWithDefaultRoute` for now.

PR #2468, opened 2015-04-28
Included default route template in docs

henkmollema:
Added the exact route being used in the
`BuilderExtensions.UseMvcWithDefaultRoute()` method.

#1885

### DIFF
--- a/src/Microsoft.AspNet.Mvc/BuilderExtensions.cs
+++ b/src/Microsoft.AspNet.Mvc/BuilderExtensions.cs
@@ -31,7 +31,8 @@ namespace Microsoft.AspNet.Builder
 
         /// <summary>
         /// Adds MVC to the <see cref="IApplicationBuilder"/> request execution pipeline
-        /// with a default route.
+        /// with a default route named 'default' and the following template:
+        /// '{controller=Home}/{action=Index}/{id?}'.
         /// </summary>
         /// <param name="app">The <see cref="IApplicationBuilder"/>.</param>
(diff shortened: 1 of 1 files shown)


henkmollema:
@Eilon @kichalla 

As discussed in 53f3a91f01c740f3cbf5209a9eeeefd614a6cbc2

Eilon:
:shipit: 

@kichalla can you merge?

src/Microsoft.AspNet.Mvc/BuilderExtensions.cs
```diff
@@ -31,7 +31,7 @@ public static IApplicationBuilder UseMvc([NotNull] this IApplicationBuilder app)
 
         /// <summary>
         /// Adds MVC to the <see cref="IApplicationBuilder"/> request execution pipeline
-        /// with a default route.
+        /// with a default route using the following template: '{controller=Home}/{action=Index}/{id?}'.
```
kichalla: Should we include the route name too as it can be used in link generation?

henkmollema: Was thinking about that. Perhaps the most explicit option.

Eilon: Sure. "... with a default route named 'Xyz' and the following template: ..."

henkmollema:
@kichalla @Eilon updated and rebased.

kichalla:
:shipit:

PR #2469, opened 2015-04-28
Part 2 of integration tests for GenericModelBinder

rynowak:
This includes a random sampling of combinations where a collection model
binder is binding another collection as an element type.

### DIFF
--- a/test/Microsoft.AspNet.Mvc.IntegrationTests/GenericModelBinderIntegrationTest.cs
+++ b/test/Microsoft.AspNet.Mvc.IntegrationTests/GenericModelBinderIntegrationTest.cs
@@ -1,0 +1,205 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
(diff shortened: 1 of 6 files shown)


rynowak:
@harshgMSFT will rebase on top of your changes after they are merged.

test/Microsoft.AspNet.Mvc.IntegrationTests/GenericModelBinderIntegrationTest.cs
```diff
@@ -201,5 +201,368 @@ protected override Task<ModelBindingResult> BindModelCoreAsync(ModelBindingConte
...
+            // Arrange
+            var argumentBinder = ModelBindingTestHelper.GetArgumentBinder();
+            var parameter = new ParameterDescriptor()
+            {
+                Name = "parameter",
+                ParameterType = typeof(Dictionary<string, int>[])
+            };
+
+            var httpContext = new DefaultHttpContext();
+            httpContext.Request.QueryString = new QueryString("?parameter[0][key0]=0");
```
harshgMSFT: nit:suggest use a non default value

rynowak: sure, I'll update all of these

harshgMSFT:
:shipit:

PR #2471, opened 2015-04-29
Fix explicit cast of HashSet<string> to IReadOnlyCollection<string> to

pranavkm:
allow tests to succeed in machines with only 4.5.1.


pranavkm:
This might resolve one class of issues that prevents AppVeyor from building.

rynowak:
:shipit: 

Too bad `ReadOnlyCollection<T>` only works with `IList<T>`

PR #2474, opened 2015-04-30
ModelBinding Integration tests part 3

rynowak:


test/Microsoft.AspNet.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
```diff
@@ -0,0 +1,1533 @@
...
+
+        private class Person1
+        {
+            public string Name { get; set; }
+
+            [FromBody]
+            public Address1 Address { get; set; }
+        }
+
+        public class Address1
```
harshgMSFT: nit: private?

test/Microsoft.AspNet.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
```diff
@@ -0,0 +1,1533 @@
...
+        {
+            // Arrange
+            var argumentBinder = ModelBindingTestHelper.GetArgumentBinder();
+            var parameter = new ParameterDescriptor()
+            {
+                Name = "parameter",
+                ParameterType = typeof(Order1)
+            };
+
+            // Need to have a key here so that the MutableObjectModelBinder will recurse to bind elements.
```
harshgMSFT: Wouldnt just having the from body be sufficient for Person1 to be created ( since it would be a poco with only a greedy model binder ).

test/Microsoft.AspNet.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
```diff
@@ -0,0 +1,1533 @@
...
+            Assert.Equal(2, modelState.Count); // This fails due to #2446
+            Assert.Equal(0, modelState.ErrorCount);
+            Assert.True(modelState.IsValid);
+
+            var entry = Assert.Single(modelState, e => e.Key == "parameter.Customer.Name").Value;
+            Assert.Equal("bill", entry.Value.AttemptedValue);
+            Assert.Equal("bill", entry.Value.RawValue);
+
+            // These fail due to #2445
+            entry = Assert.Single(modelState, e => e.Key == "parameter.Customer.Address").Value;
```
harshgMSFT: shouldnt this line pass?

test/Microsoft.AspNet.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
```diff
@@ -0,0 +1,1533 @@
...
+            public int ProductId { get; set; }
+
+            public Person2 Customer { get; set; }
+        }
+
+        private class Person2
+        {
+            public string Name { get; set; }
+
+            [FromBody]
```
harshgMSFT: Should this be `[FromServices]`

test/Microsoft.AspNet.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
```diff
@@ -0,0 +1,1533 @@
...
+            Assert.Null(model.Customer);
+
+            Assert.Equal(0, modelState.Count); // This fails due to #2446
+            Assert.Equal(0, modelState.ErrorCount);
+            Assert.True(modelState.IsValid);
+        }
+
+        public class Order5
+        {
+            public string Name { get; set; }
```
harshgMSFT: nit: In this case this property just seems to be noise.

harshgMSFT:
Fix minor comments then :shipit:

test/Microsoft.AspNet.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
rynowak: :+1:

test/Microsoft.AspNet.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
rynowak: No, not in this case because the `[FromBody]` is two levels deep. It's a property of `Person` not of `Order`

test/Microsoft.AspNet.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
rynowak: You're right. I blew that.

test/Microsoft.AspNet.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
rynowak: It's being used to test the case where some data is provided for the model, but not for the collection.

test/Microsoft.AspNet.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
harshgMSFT: yea but `Order` should be created since it is a top level model, when it comes to deciding whether to create `Person`, it would create `Person` if it is a POCO with all properties from a greedy binder isnt it?

harshgMSFT:
One thing that I missed commenting on : 
Should we add a test case for each category which is 
POCO>POCO, where the nested POCO gets created because all properties are bound from a greedy model binder?

test/Microsoft.AspNet.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
rynowak: It doesn't have all properties from a greedy model binder

rynowak:
> One thing that I missed commenting on : Should we add a test case for each category which is 
POCO>POCO, where the nested POCO gets created because all properties are bound from a greedy model binder?

Sure, I'll add one.

test/Microsoft.AspNet.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
harshgMSFT: I am sorry i was not clear in my comment. What I meant was if we removed `public string Name { get; set; }` then it would be created automatically. I do not see any tests with that scenario covered can we add that as well?

PR #2477, opened 2015-04-30
Integration tests for Collection model binder classes ->Poco modelbinder

harshgMSFT:
Collection Model Binder -> MutableObject(POCO) Model Binder – With-Data/Fallback-to-empty-prefix/Without-Data


rynowak:
:shipit:

PR #2479, opened 2015-05-01
Another batch of ModelBinding integration tests

rynowak:
Includes tests for [Required] on properties with various levels of nesting
inside complex model binders.


harshgMSFT:
:shipit:

PR #2484, opened 2015-05-01
More model binding integration tests

rynowak:
Includes tests for a ValidationAttribute other than RequiredAttribute on
complex models using POCOs and collections

### DIFF
--- a/test/Microsoft.AspNet.Mvc.IntegrationTests/ValidationIntegrationTests.cs
+++ b/test/Microsoft.AspNet.Mvc.IntegrationTests/ValidationIntegrationTests.cs
@@ -97,8 +97,6 @@ namespace Microsoft.AspNet.Mvc.IntegrationTests
             AssertRequiredError("CustomerName", error);
         }
 
-
-
         private class Order2
         {
             [Required]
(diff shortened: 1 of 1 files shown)


test/Microsoft.AspNet.Mvc.IntegrationTests/ValidationIntegrationTests.cs
```diff
@@ -464,6 +462,549 @@ private class Order5
...
+                }
+                else
+                {
+                    return new ValidationResult("Invalid Person.");
+                }
+            }
+        }
+
+        [Fact(Skip = "Extra ModelState key because of #2446")]
+        public async Task Validation_StringLengthAttribute_OnPOCOProperty_Valid()
```
harshgMSFT: `Validation_CustomAttribute_OnPOCOProperty_Valid` ? There is no `StringLengthAttribute` here.

harshgMSFT: Similar comments wherever this `Order8` is used.

harshgMSFT:
Fix test names then :shipit:

PR #2495, opened 2015-05-04
A few quick cleanups

dougbu:
- in .kproj files, let VS mark a couple more test projects as such
- remove duplicate `ModelExpression` class
 - not necessary now that test project references MVC Core project
 - follow up to 9fded74
- avoid `resources` warning in project.json files


dougbu:
FYI only; will push immediately

rynowak:
:shipit: - trusting you that the `resource` -> `resources` change doesn't require any accompanying code changes.

dougbu:
@rynowak yeah, it was just a warning from DNU about changes in the project.json syntax.

Issue #2498, opened 2015-05-04
PrecompilationWebSite has troubles building in VS


dougbu:
- no problems if site built on its own
- 10 errors if Microsoft.AspNet.Mvc.FunctionalTests or solution built
 - 8 messages are "Cannot resolve TagHelper containing assembly 'PrecompilationWebSite'. Error: Could not load file or assembly 'Microsoft.AspNet.Mvc.Core' or one of its dependencies. The system cannot find the file specified."
 - 2 messages are "Encountered an unexpected error when attempting to resolve tag helper directive '@addTagHelper' with value '*, Microsoft.AspNet.Mvc.TagHelpers'. Error: Could not load file or assembly 'Microsoft.AspNet.Mvc.Core' or one of its dependencies. The system cannot find the file specified."

danroth27:
@davidfowl @Eilon What's the plan for precompilation all up?

dougbu:
See also aspnet/RazorTooling#10

dougbu:
Basically there's two ways to go from here:
1. Fix this bug or aspnet/RazorTooling#10 -- either should get Razor precompilation working reliably across the MVC repo.  Then revert commit 3fbead0.
2. Decide Razor precompilation is unstable and will take a significant (unbudgeted) effort to get working reliably.  Then remove our support for this feature, along with the PrecompilationWebSite.

danroth27:
#2551

PR #2499, opened 2015-05-04
Get builds working better within VS (temporary fix?); disable pre-compilation

dougbu:
- remove `RazorPreCompilation.cs` files from all but the PrecompilationWebSite
- related to #2498 but not addressing that issue since pre-compilation is what that site is about


dougbu:
@Eilon @rynowak @pranavkm any thoughts on doing this until and unless we get pre-compilation working reliably when building functional tests within VS?

rynowak:
I'd want to have @eilon 's ok, but to me anything that's blocking developer workflow needs to be addressed.

Are you seeing any issues other than errors in VS?

dougbu:
No other issues.  In fact VS freezes more often when I'm editing in the Razor repo :frowning: 

Note this leaves the #2498 errors in play.  I could remove those 10 as well.  But then we might as well throw away the PrecompilationWebSite entirely.

Eilon:
1. Is this a tooling bug? Runtime bug? Where is the underlying bug logged?
2. I'm ok checking this in, but also log a bug to restore these (including pointing to the changeset that removes these)

dougbu:
@Eilon 
1. I looked at #2498 some without narrowing the problem down much.  The symptoms certainly point toward a component used only when pulling the PrecompilationWebSite into a larger VS build (but I can't think of a component unique to that scenario) or some lack of generality in the site itself, the functional test setup, VS, or the runtime. So, yeah, it's one of the bits you listed :smile: and right now #2498 is filed in MVC.
2. Sure.  Will file after I push.

In the meantime I'll add the additional errors seen without this change to the commit description.  That is, I'll make sure the change still makes a difference a few days later.

Eilon:
@dougbu ok thanks go ahead and merge.

rynowak:
@dougbu should this be closed?

dougbu:
@rynowak I didn't get back to this over the last couple of days.  The change definitely makes debugging Razor code-generation orders of magnitude easier.  I'll try to reproduce the original issue(s) over the weekend and re-confirm the change helps even when not doing low-level debugging.

Will take a more definitive action in the next day or two...

dougbu:
Checked VS behaviour without this change.  Much uglier -- 26 errors instead of the 10 (mentioned in #2498) just for PrecompilationWebSite.  And I can't start the tag helper sample from VS.

So I'll check this change in shortly.

dougbu:
3fbead0

dougbu:
@Eilon w.r.t.
> also log a bug to restore these (including pointing to the changeset that removes these)

I added a link to the commit in the existing #2498 which is generally about Razor precompilation issues.

PR #2502, opened 2015-05-05
Fix #2378 - Fully expand models in ApiExplorer

### LINKED ISSUE #2378: Fully Expand Models in ApiExplorer
This is feedback from @domaindrivendev:

```
    [HttpGet("addresses/validate")
    public bool ValidateAddress([FromQuery] Address address)
```

Today we'd describe `address` as a single parameter, despite the fact that it maps to different query string keys (`address.zipcode`, `address.state`, etc...). The feedback is that Swashbuckle (and other consumers) would always want this data to be broken down as much as possible (today it's as brief as possible).

We should either break this down fully (by property) or break down the first level (by property). 

We'd of course apply cycle-breaking, and terminate on any type with a type-converter from string.

rynowak:
This change dramatically simplifies the parameter discovery logic in
DefaultApiDescriptionProvider. Instead of surfacing POCO objects as
parameters, we now fully expand every model.

The rationale is that we want to show every key/value that can be set by
the user and not force consumers of ApiDescription to do that themselves.

Tests are cleaned up to match the new behavior.


Eilon:
:shipit:

rynowak:
@dougbu should this be closed?

PR #2504, opened 2015-05-05
[Design]Fix #448: Support app-wide defaults for HTML helpers

### LINKED ISSUE #448: Support app-wide defaults for more MVC settings, especially for HTML helpers
Extend use of that infrastructure to support the `ClientValidationEnabled` flag as well. 

Also see Doug's comment below for other properties that needs to be exposed.

kirthik:
@dougbu : I made the change and updated existing tests. Once you are Ok with the change, I will add new tests.


src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewExecutor.cs
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) .NET Foundation. All rights reserved.
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
dougbu: huh?

src/Microsoft.AspNet.Mvc.Core/MvcOptions.cs
```diff
@@ -163,5 +165,28 @@ public int MaxModelValidationErrors
...
+            }
+
+            set
+            {
+                if (value == null)
+                {
+                    throw new ArgumentNullException(
+                        nameof(value),
+                        Resources.FormatPropertyOfTypeCannotBeNull(nameof(HtmlHelperOptions), typeof(MvcOptions)));
+                }
```
dougbu: realize this is a design PR but still don't do this.  use `[param: NotNull]` above line 179.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/HtmlHelper.cs
```diff
@@ -494,7 +494,15 @@ public Task RenderPartialAsync([NotNull] string partialViewName, object model, V
...
-                var viewContext = new ViewContext(ViewContext, view, newViewData, TempData, writer);
+                var htmlHelperOptions = new HtmlHelperOptions
+                {
+                    ClientValidationEnabled = ViewContext.ClientValidationEnabled,
+                    Html5DateRenderingMode = ViewContext.Html5DateRenderingMode,
+                    ValidationMessageElement = ViewContext.ValidationMessageElement,
+                    ValidationSummaryMessageElement = ViewContext.ValidationSummaryMessageElement
+                };
+
+                var viewContext = new ViewContext(ViewContext, view, newViewData, TempData, writer, htmlHelperOptions);
```
dougbu: the `ViewContext` copy constructor should not also require an `HtmlHelperOptions` argument. copy what is needed from the original `ViewContext` in the copy constructor.

src/Microsoft.AspNet.Mvc.Core/ActionResults/PartialViewResult.cs
```diff
@@ -80,7 +81,13 @@ public class PartialViewResult : ActionResult
...
             using (view as IDisposable)
             {
-                await ViewExecutor.ExecuteAsync(view, context, ViewData, TempData, ContentType);
+                await ViewExecutor.ExecuteAsync(
+                    view,
+                    context,
+                    ViewData,
+                    TempData,
+                    new HtmlHelperOptions(),
+                    ContentType);
```
dougbu: this means any `MvcOptions` changes made are irrelevant when executing the view. need to either pass the `HtmlHelperOptions` into this method or include the whole `MvcOptions` in the `ActionContext` when _that_ is created.

src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewResult.cs
```diff
@@ -80,7 +81,13 @@ public class ViewResult : ActionResult
 
             using (view as IDisposable)
             {
-                await ViewExecutor.ExecuteAsync(view, context, ViewData, TempData, ContentType);
+                await ViewExecutor.ExecuteAsync(
+                    view,
+                    context,
+                    ViewData,
+                    TempData,
+                    new HtmlHelperOptions(),
```
dougbu: ditto

src/Microsoft.AspNet.Mvc.Core/Rendering/HtmlHelperOptions.cs
```diff
@@ -0,0 +1,133 @@
...
+        private bool _clientValidationEnabled = true;
+        private string _validationMessageElement = "span";
+        private string _validationSummaryMessageElement = "span";
+
+        public HtmlHelperOptions()
+        {
+        }
+
+        /// <summary>
+        /// Specifies the name of the anti-forgery token field that is used by the anti-forgery system.
```
dougbu: careful of your cut-n-paste

src/Microsoft.AspNet.Mvc.Core/Rendering/HtmlHelperOptions.cs
```diff
@@ -0,0 +1,133 @@
...
+                return _html5DateRenderingMode;
+            }
+            set
+            {
+                _html5DateRenderingMode = value;
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets the character that replaces periods in the ID attribute of an element.
```
dougbu: `character`?  looks like a `<see cref="string"/>` to me.

src/Microsoft.AspNet.Mvc.Core/Rendering/HtmlHelperOptions.cs
```diff
@@ -0,0 +1,133 @@
...
+        public string IdAttributeDotReplacement
+        {
+            get
+            {
+                return _idAttributeDotReplacement;
+            }
+
+            set
+            {
+                if (value == null)
```
dougbu: why is an empty `string` legal here?

src/Microsoft.AspNet.Mvc.Core/Rendering/HtmlHelperOptions.cs
```diff
@@ -0,0 +1,133 @@
...
+        /// </summary>
+        public string ValidationMessageElement
+        {
+            get
+            {
+                return _validationMessageElement;
+            }
+            set
+            {
+                if (value == null)
```
dougbu: same empty `string

dougbu: oops, same empty `string` question here and for the next property

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewExecutor.cs
kirthik: Not sure why this shows up as a change.

src/Microsoft.AspNet.Mvc.Core/MvcOptions.cs
kirthik: Will fix it.

src/Microsoft.AspNet.Mvc.Core/ActionResults/PartialViewResult.cs
kirthik: I will include whole `MvcOptions` in `ActionContext`.

src/Microsoft.AspNet.Mvc.Core/Rendering/HtmlHelperOptions.cs
kirthik: Oops. Will fix it.

src/Microsoft.AspNet.Mvc.Core/Rendering/HtmlHelperOptions.cs
kirthik: Will fix it.

src/Microsoft.AspNet.Mvc.Core/Rendering/HtmlHelperOptions.cs
kirthik: Copy paste error. Will fix it.

src/Microsoft.AspNet.Mvc.Core/Rendering/HtmlHelperOptions.cs
kirthik: Copy paste error. Will fix it.

src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewExecutor.cs
Eilon: Probably a BOM change?

src/Microsoft.AspNet.Mvc.Core/Rendering/HtmlHelperOptions.cs
```diff
@@ -0,0 +1,133 @@
...
+        /// <summary>
+        /// Set this property to <see cref="Html5DateRenderingMode.Rfc3339"/> to have templated helpers such as
+        /// <see cref="Editor"/> and <see cref="IHtmlHelper{TModel}.EditorFor"/> render date and time values as RFC
+        /// 3339 compliant strings. By default these helpers render dates and times using the current culture.
+        /// </summary>
+        public Html5DateRenderingMode Html5DateRenderingMode
+        {
+            get
+            {
+                return _html5DateRenderingMode;
```
Eilon: Use an auto-property for this?

src/Microsoft.AspNet.Mvc.Core/MvcOptions.cs
```diff
@@ -163,5 +165,28 @@ public int MaxModelValidationErrors
         /// </ul>
         /// </remarks>
         public IList<IMetadataDetailsProvider> ModelMetadataDetailsProviders { get; }
+
+        /// <summary>
+        /// Provides programmatic configuration for the html helpers and view context.
+        /// </summary>
+        public HtmlHelperOptions HtmlHelperOptions
```
pranavkm: cc @rynowak Would having this be a separate top level option be easier to deal with when it comes to layering?

src/Microsoft.AspNet.Mvc.Core/ActionResults/PartialViewResult.cs
rynowak: `MvcOptions` isn't and shouldn't be a member of `ActionContext`.

The established pattern for accessing something off of options in an action result is to use `GetRequiredService`

src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewExecutor.cs
```diff
@@ -70,7 +72,14 @@ public static class ViewExecutor
...
                 try
                 {
-                    var viewContext = new ViewContext(actionContext, view, viewData, tempData, writer);
+                    var viewContext = new ViewContext(
+                        actionContext,
+                        view,
+                        viewData,
+                        tempData,
+                        writer,
+                        htmlHelperOptions);
```
rynowak: We should be really careful about adding anything **options** to a context. These contexts are intended to be constructable for unit tests.

src/Microsoft.AspNet.Mvc.Core/MvcOptions.cs
rynowak: It seems like a good idea to me, makes the code more readable.

src/Microsoft.AspNet.Mvc.Core/MvcOptions.cs
rynowak: Also `FormatPropertyOfTypeCannotBeNull` is for when your code needs to access a property of a separate object.

src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewExecutor.cs
rynowak: Also concerned about this because we create ViewContext in a number of places

src/Microsoft.AspNet.Mvc.Core/Rendering/HtmlHelperOptions.cs
```diff
@@ -0,0 +1,133 @@
...
+    {
+        private string _idAttributeDotReplacement = "_";
+        private Html5DateRenderingMode _html5DateRenderingMode;
+        private bool _clientValidationEnabled = true;
+        private string _validationMessageElement = "span";
+        private string _validationSummaryMessageElement = "span";
+
+        public HtmlHelperOptions()
+        {
+        }
```
rynowak: Not needed

src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewExecutor.cs
dougbu: @rynowak the properties we care about are already members of `ViewContext`.  Are you suggesting:
- adding individual constructor parameters for the 4 of them
- passing in just the `HtmlHelperOptions` as @kirthik has done in the current iteration
- requiring _all_ "outermost" (i.e. not for an inner view or similar) `ViewContext` creators to set the individual properties
- something else?

rynowak: Yeah it made more sense the more I read the review. I think I'm mostly ok with it as is.

src/Microsoft.AspNet.Mvc.Core/Rendering/HtmlHelperOptions.cs
kirthik: Will remove.

src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewExecutor.cs
kirthik: Ok. I will keep `HtmlHelperOptions` as is.

src/Microsoft.AspNet.Mvc.Core/Rendering/HtmlHelperOptions.cs
kirthik: Yes. Will change.

src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewExecutor.cs
kirthik: May be. Not sure.

Eilon: Either way I wouldn't worry about this change. Just blame Unicode or something.

dougbu: @Eilon the Consortium or their standard?  Want to be clear here :smiling_imp:

Eilon: I :heart: Unicode.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -55,7 +55,7 @@ public class DefaultHtmlGenerator : IHtmlGenerator
             _htmlEncoder = htmlEncoder;
 
             // Underscores are fine characters in id's.
-            IdAttributeDotReplacement = "_";
+            IdAttributeDotReplacement = optionsAccessor.Options.HtmlHelperOptions.IdAttributeDotReplacement;
```
dougbu: Makes some sense but leaves this setting in a slightly strange place.  `DefaultHtmlGenerator` owns an `IdAttributeDotReplacement` value that newly-created instances will not inherit.  `HtmlHelper` also has a `public` `IdAttributeDotReplacement` that reads from / writes to its `private` `IHtmlGenerator` instance .

MVC 5 equivalent of this property is `static`. It is _not_ possible there to give it a setting unique to any scope of `HtmlHelper` instance.

One possibility: Remove the `IHtmlHelper` and `IHtmlGenerator` setters for this property.  Then `MvcOptionsSetup` would be the _single_ way to change `IdAttributeDotReplacement`.  @Eilon thoughts?

src/Microsoft.AspNet.Mvc.Core/Rendering/HtmlHelperOptions.cs
```diff
@@ -0,0 +1,86 @@
...
+using Microsoft.AspNet.Mvc.Core;
+
+namespace Microsoft.AspNet.Mvc.Rendering
+{
+    public class HtmlHelperOptions
+    {
+        private string _idAttributeDotReplacement = "_";
+        private bool _clientValidationEnabled = true;
+        private string _validationMessageElement = "span";
+        private string _validationSummaryMessageElement = "span";
```
dougbu: why do these backing fields exist?  was it your intent to add `[param: NotNull]` for all of the `string` setters?

dougbu: put another way, auto-properties with default settings in a constructor would be cleaner if these fields aren't required for some other reason.

src/Microsoft.AspNet.Mvc.Core/MvcOptions.cs
```diff
@@ -163,5 +166,22 @@ public int MaxModelValidationErrors
         /// </ul>
         /// </remarks>
         public IList<IMetadataDetailsProvider> ModelMetadataDetailsProviders { get; }
+
+        /// <summary>
+        /// Provides programmatic configuration for the html helpers and view context.
```
dougbu: - `AntiForgeryOptions` is the only property in this class using "Provides". Better to stick with "Gets" e.g. "Gets the configuration information for ..."
- "the html" -> HTML
- "view context" -> `<see cref="ViewContext"/>`

src/Microsoft.AspNet.Mvc.Core/Rendering/HtmlHelperOptions.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        /// <summary>
+        /// Set this property to <see cref="Html5DateRenderingMode.Rfc3339"/> to have templated helpers such as
```
dougbu: Please word all property summaries in this new class consistently -- "Gets or sets ..."

src/Microsoft.AspNet.Mvc.Core/ViewComponents/ViewViewComponentResult.cs
```diff
@@ -92,12 +92,20 @@ public void Execute([NotNull] ViewComponentContext context)
 
             var view = FindView(context.ViewContext, viewEngine, qualifiedViewName);
 
+            var htmlHelperOptions = new HtmlHelperOptions
+            {
+                ClientValidationEnabled = context.ViewContext.ClientValidationEnabled,
+                Html5DateRenderingMode = context.ViewContext.Html5DateRenderingMode,
+                ValidationMessageElement = context.ViewContext.ValidationMessageElement,
+                ValidationSummaryMessageElement = context.ViewContext.ValidationSummaryMessageElement
+            };
```
dougbu: View components are _never_ created at the top level.  The `ViewComponentContext` contains the `ViewContext` for the next-outer scope.  So the updated code below is another case where the `TempData` argument was added in error.  Should be using the `ViewContext` copy constructor.

src/Microsoft.AspNet.Mvc.Core/ViewContext.cs
```diff
@@ -43,18 +43,21 @@ public ViewContext()
...
-            [NotNull] TextWriter writer)
+            [NotNull] TextWriter writer,
+            [NotNull] HtmlHelperOptions htmlHelperOptions)
             : base(actionContext)
         {
             View = view;
             ViewData = viewData;
             TempData = tempData;
             Writer = writer;
+            var config = htmlHelperOptions;
```
dougbu: why does this `var` exist?

dougbu:
:watch:

Would also like to see some tests showing `ViewContext` properties getting values from `HtmlHelperOptions` or a "parent" `ViewContext` and perhaps confirming the equivalent `IHtmlHelper` properties don't get out of sync.  Right now `ViewContextTests` is very light.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
kirthik: I will update the code based on @dougbu's suggestion unless @Eilon objects.

src/Microsoft.AspNet.Mvc.Core/Rendering/HtmlHelperOptions.cs
kirthik: Yes. I will fix that.

src/Microsoft.AspNet.Mvc.Core/ViewContext.cs
kirthik: Will fix it.

kirthik:
@dougbu : Updated the code and add a couple of tests. Could you please take a look?

src/Microsoft.AspNet.Mvc.Core/Rendering/HtmlHelperOptions.cs
```diff
@@ -0,0 +1,52 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
```
dougbu: incorrect copyright

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultHtmlGenerator.cs
Eilon: `IdAttributeDotReplacement` was indeed always static, so giving it an "upgrade" in that it is now a per-MVC instance setting is great.

src/Microsoft.AspNet.Mvc.Core/Rendering/HtmlHelperOptions.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        /// <summary>
+        /// Gets or sets the <see cref="string"/> that replaces periods in the ID attribute of an element.
+        /// </summary>
+        public string IdAttributeDotReplacement { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value that indicates whether client-side validation is enabled.
+        /// </summary>
+        public bool ClientValidationEnabled { get; set; }
```
Eilon: Related to my earlier comment, it's worth noting that in MVC 5.x (and Web Pages 3.x) though many of these properties were exposed as `static`, they in fact used an underlying store called `ScopedStorage` that allowed overriding values on a per-request or even per-page. Having said that, it was not a very well-known feature at all, so even if we don't bring that forward, that's fine.

src/Microsoft.AspNet.Mvc.Core/Rendering/IHtmlHelper.cs
```diff
@@ -26,7 +26,7 @@ public interface IHtmlHelper
         /// <summary>
         /// Gets or sets the character that replaces periods in the ID attribute of an element.
```
dougbu: ~~or sets~~

dougbu: character -> `<see cref="string"/>`

dougbu:
looks like this PR needs a rebase to latest dev.  failing due to a recent HttpAbstractions namespace change.

src/Microsoft.AspNet.Mvc.Core/Rendering/HtmlHelperOptions.cs
dougbu: @Eilon agreed.  The other ones are not the same as `IdAttributeDotReplacement` in either MVC 5 or 6.  `ViewContext` already did something that would feel much like the old `ScopedStorage`-based feature from MVC 5.  So we're providing the old (super secret) feature and the public surface isn't all that different.

test/WebSites/RazorWebSite/Startup.cs
```diff
@@ -25,6 +25,14 @@ public void ConfigureServices(IServiceCollection services)
                 options.ViewLocationExpanders.Add(expander);
                 options.ViewLocationExpanders.Add(new CustomPartialDirectoryViewLocationExpander());
             });
+            services.ConfigureMvc(options =>
+            {
+                options.HtmlHelperOptions.ClientValidationEnabled = false;
+                options.HtmlHelperOptions.Html5DateRenderingMode = Microsoft.AspNet.Mvc.Rendering.Html5DateRenderingMode.Rfc3339;
+                options.HtmlHelperOptions.IdAttributeDotReplacement = "!";
+                options.HtmlHelperOptions.ValidationMessageElement = "validationMsgElement";
```
dougbu: `validationMessageElement` (don't abbrev.)

dougbu:
fix small-ish comments, rebase, and :shipit:

kirthik:
I missed few test updates in the last iteration. I rebased, addressed comments above and included the changes I missed last time.

test/Microsoft.AspNet.Mvc.Core.Test/Rendering/DefaultTemplatesUtilities.cs
```diff
@@ -163,12 +189,18 @@ public static HtmlHelper<TModel> GetHtmlHelper<TModel>(TModel model, ICompositeV
...
             Func<IHtmlHelper, IHtmlHelper> innerHelperWrapper,
-            IHtmlGenerator htmlGenerator)
+            IHtmlGenerator htmlGenerator,
+            string idAttributeDotReplacement = null)
         {
             var httpContext = new DefaultHttpContext();
             var actionContext = new ActionContext(httpContext, new RouteData(), new ActionDescriptor());
 
             var options = new MvcOptions();
+            options.HtmlHelperOptions = new HtmlHelperOptions();
```
dougbu: Remove this line to avoid confusing later developers; it's not necessary.

test/Microsoft.AspNet.Mvc.Core.Test/Rendering/DefaultTemplatesUtilities.cs
```diff
@@ -44,9 +44,9 @@ public class ObjectWithScaffoldColumn
             public string Property3 { get; set; }
         }
 
-        public static HtmlHelper<ObjectTemplateModel> GetHtmlHelper()
+        public static HtmlHelper<ObjectTemplateModel> GetHtmlHelper(string idAttributeDotReplacement = null)
```
dougbu: Not a fan at all of introducing this new ambiguity in the (numerous) `GetHtmlHelper()` overloads.  Callers using `GetHtmlHelper("some string")` should expect the Model to be "some string".

In addition we've successfully avoided using optional parameters throughout this class.

Instead add another (unambiguous) `GetHtmlHelper<TModel>(Something something, string idAttributeDotReplacement)` overload and in the longest overload.  Call the longest overload directly rather than adding optional parameters anywhere along the line.

test/Microsoft.AspNet.Mvc.Core.Test/Rendering/HtmlHelperHiddenTest.cs
```diff
@@ -241,9 +241,10 @@ public void HiddenUsesPrefixName_ToLookupPropertyValueInModelState()
             // Arrange
             var expected = @"<input id=""HtmlEncode[[MyPrefix$Property1]]"" name=""HtmlEncode[[MyPrefix.Property1]]"" type=""HtmlEncode[[hidden]]"" " +
                            @"value=""HtmlEncode[[modelstate-with-prefix]]"" />";
-            var helper = DefaultTemplatesUtilities.GetHtmlHelper(GetViewDataWithModelStateAndModelAndViewDataValues());
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper(
+                GetViewDataWithModelStateAndModelAndViewDataValues(),
+                "$");
```
dougbu: Use named parameters wherever you use this overload.

test/Microsoft.AspNet.Mvc.Core.Test/Rendering/HtmlHelperSelectTest.cs
```diff
@@ -714,7 +714,7 @@ public void ListBoxFor_WithUnreleatedExpression_GeneratesExpectedValue()
             string expectedHtml)
         {
             // Arrange
-            var helper = DefaultTemplatesUtilities.GetHtmlHelper("2");
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper<string>("2");
```
dougbu: Revert changes required only due to new ambiguity.

dougbu:
:watch:

kirthik:
@dougbu : updated.

test/Microsoft.AspNet.Mvc.Core.Test/Rendering/DefaultTemplatesUtilities.cs
```diff
@@ -141,7 +175,12 @@ public static HtmlHelper<TModel> GetHtmlHelper<TModel>(TModel model, ICompositeV
             ICompositeViewEngine viewEngine,
             IModelMetadataProvider provider)
         {
-            return GetHtmlHelper(model, urlHelper, viewEngine, provider, innerHelperWrapper: null);
+            return GetHtmlHelper(
+                model,
+                urlHelper,
+                viewEngine,
+                provider,
+                innerHelperWrapper: null);
```
dougbu: why did this change and the one above happen?  one line is fine when everything fits.

test/Microsoft.AspNet.Mvc.Core.Test/Rendering/HtmlHelperHiddenTest.cs
```diff
@@ -403,8 +405,7 @@ public void HiddenWithComplexExpressions_UsesIdDotSeparator(string expression, s
             viewData["Property3[height]"] = "Prop3Value";
             viewData["Property4.Property5"] = "Prop5Value";
             viewData["Property4.Property6[0]"] = "Prop6Value";
-            var helper = DefaultTemplatesUtilities.GetHtmlHelper(viewData);
-            helper.IdAttributeDotReplacement = "$$";
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper(viewData, "$$");
```
dougbu: name the "$$" argument

test/Microsoft.AspNet.Mvc.Core.Test/Rendering/HtmlHelperHiddenTest.cs
dougbu: name the "$" argument

test/Microsoft.AspNet.Mvc.Core.Test/Rendering/HtmlHelperPasswordTest.cs
```diff
@@ -196,8 +197,7 @@ public void PasswordWithComplexExpressions_UsesIdDotSeparator(string expression,
         {
             // Arrange
             var viewData = GetViewDataWithModelStateAndModelAndViewDataValues();
-            var helper = DefaultTemplatesUtilities.GetHtmlHelper(viewData);
-            helper.IdAttributeDotReplacement = "$$";
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper(viewData, "$$");
```
dougbu: name the "$$" argument and the "$" argument on line 98

test/Microsoft.AspNet.Mvc.Core.Test/ActionResults/PartialViewResultTest.cs
```diff
@@ -167,6 +168,13 @@ public class PartialViewResultTest
                            .Returns(viewEngine.Object);
             serviceProvider.Setup(p => p.GetService(typeof(ILogger<PartialViewResult>)))
                            .Returns(new Mock<ILogger<PartialViewResult>>().Object);
+            serviceProvider.Setup(s => s.GetService(typeof(IOptions<MvcOptions>)))
+                .Returns(() => {
```
dougbu: Not for this PR: In the future please _never_ use trailing curly braces, even in test code.  Curly braces get their own lines except in the special case of single line auto-properties and sometimes close parentheses.  e.g.
``` c#
    serviceProvider.Setup(s => s.GetService(typeof(IOptions<MvcOptions>)))
        .Returns(() =>
        {
            var optionsAccessor = new Mock<IOptions<MvcOptions>>();
            ...
        });
```

dougbu:
:shipit: with fixes for minor comments and after squashing your commits

src/Microsoft.AspNet.Mvc.Core/MvcOptions.cs
```diff
@@ -163,5 +166,22 @@ public int MaxModelValidationErrors
...
         public IList<IMetadataDetailsProvider> ModelMetadataDetailsProviders { get; }
+
+        /// <summary>
+        /// Gets or sets programmatic configuration for the HTML helpers and <see cref="ViewContext"/>.
+        /// </summary>
+        public HtmlHelperOptions HtmlHelperOptions
+        {
+            get
+            {
+                return _htmlHelperOptions;
```
pranavkm: Do we need the backing field if we are using `param: NotNull`? Wouldn't

`public HtmlHelperOptions HtmlHelperOptions { get; [param: NotNull] set; }` work?

test/Microsoft.AspNet.Mvc.Core.Test/ActionResults/ViewResultTest.cs
```diff
@@ -12,6 +12,7 @@
 using Microsoft.Net.Http.Headers;
 using Moq;
 using Xunit;
+using Microsoft.Framework.OptionsModel;
```
pranavkm: Sort usings.

test/Microsoft.AspNet.Mvc.Core.Test/ActionResults/ViewResultTest.cs
```diff
@@ -179,6 +180,13 @@ public class ViewResultTest
                            .Returns(viewEngine.Object);
             serviceProvider.Setup(p => p.GetService(typeof(ILogger<ViewResult>)))
                            .Returns(new Mock<ILogger<ViewResult>>().Object);
+            serviceProvider.Setup(s => s.GetService(typeof(IOptions<MvcOptions>)))
+                .Returns(() => {
+                    var optionsAccessor = new Mock<IOptions<MvcOptions>>();
```
pranavkm: This would create a new instance of `IOptions<MvcOptions>` each time you invoke `GetService(typeof(IOptions<MvcOptions>))`. I realize the test scenario doesn't rely on this behavior, but it would be nicer to have it behave closer to the real scenario and have it return a single value all the time. Could we change this to not use a lambda?

test/WebSites/RazorWebSite/Controllers/HtmlHelperOptionsController.cs
```diff
@@ -0,0 +1,54 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNet.Mvc;
+using System;
```
pranavkm: Sort usings.

src/Microsoft.AspNet.Mvc.Core/Rendering/HtmlHelperOptions.cs
```diff
@@ -0,0 +1,52 @@
...
+namespace Microsoft.AspNet.Mvc.Rendering
+{
+    /// <summary>
+    /// Provides programmatic configuration for the HTML helpers and <see cref="ViewContext"/>.
+    /// </summary>
+    public class HtmlHelperOptions
+    {
+        public HtmlHelperOptions()
+        {
+            IdAttributeDotReplacement = "_";
```
pranavkm: You could initialize these values inline 
`public string IdAttributeDotReplacement { get; set; } = "_";` to avoid having to document this constructor.

pranavkm:

src/Microsoft.AspNet.Mvc.Core/Rendering/HtmlHelperOptions.cs
kirthik: Will do.

test/Microsoft.AspNet.Mvc.Core.Test/ActionResults/PartialViewResultTest.cs
kirthik: Ok.

test/Microsoft.AspNet.Mvc.Core.Test/ActionResults/ViewResultTest.cs
kirthik: yes, I removed lambda and made it return same value.

test/Microsoft.AspNet.Mvc.Core.Test/Rendering/DefaultTemplatesUtilities.cs
kirthik: will change.

kirthik:
Code checked in - caa8ea44fbc59d333dd82f571d756ce6876f0fae

Issue #2507, opened 2015-05-06
MVC is broken on CoreCLR


davidfowl:
![image](https://cloud.githubusercontent.com/assets/95136/7487886/85c23efc-f375-11e4-9ed5-fe32e2aba167.png)

This stems from the fact that we still flow dependencies from shared projects that end up being marked as build dependencies. That's pit of failure so we'll have to make it so that dependencies don't flow through shared projects. 

In the main time, we need to pull the type converter dependency back into the abstractions library.

rynowak:
> In the main time, we need to pull the type converter dependency back into the abstractions library.

We need to take a look at slimming down MSD and VDD - right now there's a lot of code in these classes, but they're really just collections.

rynowak:
@davidfowl Assuming this was fixed by 5cec601abe968c8475bac7f3b6e7f48c154a6d63

PR #2508, opened 2015-05-06
Fix MVC packages being broken on CoreCLR

davidfowl:
- Mvc is currently broken on CoreCLR because it is
inheriting dependencies from Microsoft.AspNet.Mvc.Common but the resulting
dependency to Microsoft.AspNet.Mvc.Common is then erased at pack time.
This change moves the dependencies down and makes the shared package
only depend on System.Runtime.

#2507 

/cc @rynowak

### DIFF
--- a/src/Microsoft.AspNet.JsonPatch/project.json
+++ b/src/Microsoft.AspNet.JsonPatch/project.json
@@ -15,7 +15,10 @@
                 "System.Collections": "4.0.10-beta-*",
                 "System.Collections.Concurrent": "4.0.10-beta-*",
                 "System.Globalization": "4.0.10-beta-*",
-                "System.Runtime.Extensions": "4.0.10-beta-*"
+                "System.Runtime.Extensions": "4.0.10-beta-*",
+                "System.ComponentModel.TypeConverter": "4.0.0-beta-*",
+                "System.Reflection.Extensions": "4.0.0-beta-*",
+                "System.Text.Encoding.Extensions": "4.0.10-beta-*"
             }
         }
     }
--- a/src/Microsoft.AspNet.JsonPatch/project.json
+++ b/src/Microsoft.AspNet.JsonPatch/project.json
@@ -14,10 +14,10 @@
                 "Microsoft.CSharp": "4.0.0-beta-*",
                 "System.Collections": "4.0.10-beta-*",
                 "System.Collections.Concurrent": "4.0.10-beta-*",
-                "System.Globalization": "4.0.10-beta-*",
-                "System.Runtime.Extensions": "4.0.10-beta-*",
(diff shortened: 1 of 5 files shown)


pranavkm:
:shipit:

src/Microsoft.AspNet.Mvc.Common/project.json
```diff
@@ -8,9 +8,7 @@
         "dnx451": { },
         "dnxcore50": {
             "dependencies": {
-		"System.ComponentModel.TypeConverter": "4.0.0-beta-*",
-                "System.Reflection.Extensions": "4.0.0-beta-*",
-                "System.Text.Encoding.Extensions": "4.0.10-beta-*"
+                "System.Runtime": "4.0.20-beta-*"
```
pranavkm: Should we print a warning \ make it an error to have any dependencies in a shared source project if this is the intended behavior?

src/Microsoft.AspNet.JsonPatch/project.json
```diff
@@ -15,7 +15,10 @@
                 "System.Collections": "4.0.10-beta-*",
                 "System.Collections.Concurrent": "4.0.10-beta-*",
                 "System.Globalization": "4.0.10-beta-*",
-                "System.Runtime.Extensions": "4.0.10-beta-*"
+                "System.Runtime.Extensions": "4.0.10-beta-*",
+                "System.ComponentModel.TypeConverter": "4.0.0-beta-*",
+                "System.Reflection.Extensions": "4.0.0-beta-*",
+                "System.Text.Encoding.Extensions": "4.0.10-beta-*"
```
dougbu: aren't we trying to keep dependencies in alphabetic order?

dougbu:
:watch:

src/Microsoft.AspNet.JsonPatch/project.json
Eilon: Yes, we are. Tsk tsk.

davidfowl: That's what you're worried about? MVC doesn't even work right now :stuck_out_tongue_closed_eyes:

src/Microsoft.AspNet.Mvc.Common/project.json
davidfowl: It's not inherently broken. It's because it's marked as a build time dependency so nothing will catch this until you build the package.

src/Microsoft.AspNet.JsonPatch/project.json
davidfowl: On top of that you guys have inconsistent spacing in your project.json files.

PR #2510, opened 2015-05-06
[Design] Wrapper stream writer which prevents writing flushing to avoid chunke…

kichalla:
…d encoding and avoids writing preamble or BOM bytes.

Right, the `HttpResponseStreamWriter` here doesn't do much different than what I have tried earlier before but probably a little improvement is that we wrap the stream and encoding within the writer itself...

I have 'fixed' few tests (json output formatter) which earlier expected BOM bytes to not check them anymore.

Please ignore the location of the `HttpResponseStreamWriter`...we could later move it to some other repo if we are happy with this change...

@pranavkm @rynowak @Tratcher 

Related issue: https://github.com/dotnet/coreclr/issues/933

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewExecutor.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewExecutor.cs
@@ -16,7 +16,6 @@ namespace Microsoft.AspNet.Mvc
     /// </summary>
     public static class ViewExecutor
     {
-        private const int BufferSize = 1024;
         private static readonly MediaTypeHeaderValue DefaultContentType = new MediaTypeHeaderValue("text/html")
         {
             Encoding = Encodings.UTF8EncodingWithoutBOM
@@ -64,23 +63,10 @@ namespace Microsoft.AspNet.Mvc
 
             response.ContentType = contentTypeHeader.ToString();
 
-            var wrappedStream = new StreamWrapper(response.Body);
-
-            using (var writer = new StreamWriter(wrappedStream, encoding, BufferSize, leaveOpen: true))
+            using (var writer = new HttpResponseStreamWriter(response.Body, encoding))
             {
-                try
-                {
-                    var viewContext = new ViewContext(actionContext, view, viewData, tempData, writer);
-                    await view.RenderAsync(viewContext);
-                }
-                catch
-                {
-                    // Need to prevent writes/flushes on dispose because the StreamWriter will flush even if
-                    // nothing got written. This leads to a response going out on the wire prematurely in case an
-                    // exception is being thrown inside the try catch block.
-                    wrappedStream.BlockWrites = true;
-                    throw;
-                }
+                var viewContext = new ViewContext(actionContext, view, viewData, tempData, writer);
+                await view.RenderAsync(viewContext);
             }
         }
 
--- a/src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
@@ -1,0 +1,73 @@
+﻿using System.IO;
+using System.Text;
+using Microsoft.AspNet.Mvc.Internal;
+
+namespace Microsoft.AspNet.Mvc
+{
+    /// <summary>
+    /// Wraps the supplied <see cref="Stream"/> to prevent it from flushing (which can cause response to be sent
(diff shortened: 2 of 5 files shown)


src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewExecutor.cs
```diff
@@ -64,23 +63,10 @@ public static class ViewExecutor
...
+            using (var writer = new HttpResponseStreamWriter(response.Body, encoding))
             {
-                try
-                {
-                    var viewContext = new ViewContext(actionContext, view, viewData, tempData, writer);
-                    await view.RenderAsync(viewContext);
-                }
-                catch
-                {
-                    // Need to prevent writes/flushes on dispose because the StreamWriter will flush even if
```
kichalla: I still have a question regarding this...why is this so much of a concern?...there are couple of unit tests failing which explicitly check this...

src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
```diff
@@ -0,0 +1,73 @@
...
+    {
+        private const int DefaultBufferSize = 1024;
+
+        public HttpResponseStreamWriter(Stream stream, Encoding encoding)
+            : this(stream, encoding, DefaultBufferSize)
+        {
+        }
+
+        public HttpResponseStreamWriter(Stream stream, Encoding encoding, int bufferSize)
+            : base(new NonDisposableStream(stream), new NonBomEncodingWrapper(encoding), bufferSize, leaveOpen: true)
```
kichalla: Right, the `NonDisposable` part of the stream should not be a concern here as we will always leave the stream open and it is the flushing part only that we are interested...just wanted to reuse whatever we already have...

src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
```diff
@@ -0,0 +1,73 @@
...
+using Microsoft.AspNet.Mvc.Internal;
+
+namespace Microsoft.AspNet.Mvc
+{
+    /// <summary>
+    /// Wraps the supplied <see cref="Stream"/> to prevent it from flushing (which can cause response to be sent
+    /// in Chunked encoding) and disposing (as the hosting layers own the stream).
+    /// Also wraps the supplied <see cref="Encoding"/> to prevent writing the preamble or BOM bytes to the response.
+    /// </summary>
+    public class HttpResponseStreamWriter : StreamWriter
```
pranavkm: Why not just write a new `TextWriter` instead of doing this?

kichalla: I wanted to avoid implementing the buffering logic which StreamWriter already provides...

src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewExecutor.cs
rynowak: This is a big concern, and something we should try to address in our writer.

If we flush on dispose then the headers + empty body get sent to the client. Then it means that a diagnostic middleware or exception filter can't write anything.

src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
rynowak: Not sure it matters. I think today we're actually double-buffering everything because of this. Talk to @Tratcher

rynowak: So far this is just the old approach wearing a new :womans_hat:. This PR needs more pizzazz.

kichalla: Right, I discussed with @Tratcher about this...but first we need to identify what are the issues that we are trying to solve...currently we have identified only the Flushing and BOM...like mainly if we are already seeing any issues with buffering and if that's something we need to solve...

src/Microsoft.AspNet.Mvc.Core/ActionResults/ViewExecutor.cs
kichalla: :+1: ...verified it just now...will work on it...

kichalla:
Closing this PR as there is another PR out for this: https://github.com/aspnet/Mvc/pull/2518

PR #2516, opened 2015-05-07
ImageTagHelper

### LINKED ISSUE #2249: Create an ImageTagHelper that supports globbed src paths and cache-busting file version URL hash
This is equivalent to the existing Tag Helpers for `<script>` and `<link>` but for `` elements.

e.g.:
``` HTML
<!-- Globbed images, will render a separate image tag for each matched path -->
<img asp-src-include="images/thumbs/@Model.ProductId/*.jpg" alt="Some alt text" />

<!-- File version/fingerprint, will add a file version to the URL to aid in cache busting  -->
<img src="~/images/site-banner.png" alt="Site banner" asp-file-version="true" />
```

dpaquette:
An ImageTagHelper that supports globbed src paths and cache busting file
version hash
[Resolves #2249]


dpaquette:
What's the deal with @dnfclas ? I signed my cla before submitting this PR

dougbu:
@dpaquette  We just changed the repos from MS Open Tech to .NET Foundation so there's a different CLA. The [new CLA](https://cla2.dotnetfoundation.org/) is simpler than the old one and so should only take a few moments to re-sign. Apologies for the extra bit of work!

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
```diff
@@ -0,0 +1,167 @@
...
+        public string Src { get; set; }
+
+        /// <summary>
+        /// Value indicating if file version should be appended to the src urls.
+        /// </summary>
+        /// <remarks>
+        /// If <c>true</c> then a query string "v" with the encoded content of the file is added.
+        /// </remarks>
+        [HtmlAttributeName(FileVersionAttributeName)]
+        public bool? FileVersion { get; set; }
```
Eilon: `IncludeFileVersion`? Or maybe `AppendFileVersion`? The current name makes it sound like it *is* a file version (e.g. you'd set `asp-file-version="1.23.4"` or something).

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
```diff
@@ -0,0 +1,167 @@
...
+        private const string SrcAttributeName = "src";
+
+        private FileVersionProvider _fileVersionProvider;
+
+        private static readonly ModeAttributes<Mode>[] ModeDetails = new[] {
+            // Regular src with file version
+            ModeAttributes.Create(Mode.FileVersion, new[] { FileVersionAttributeName }),
+        };
+
+        private enum Mode
```
Eilon: A single-value enum is... weird. Is it supposed to be more like `None=0, FileVersion=1` or something? Not sure I quite get how this works.

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
dpaquette: I agree that AppendFileVersion or AddFileVersion would make more sense. This is modeled off the Link and Script tag helpers. If we change the name here then it would need to be changed on both those tag helpers too.

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
dpaquette: This could be a boolean now. It previously had multiple modes back when this supported src globbing.

dpaquette:
@Eilon I simplified the ImageTagHelper class by removing the Mode and Mode Detection code since this was no longer needed.  Not sure what we should do about the naming of the asp-file-version attribute.

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
```diff
@@ -0,0 +1,144 @@
...
+    {
+        private static readonly string Namespace = typeof(ImageTagHelper).Namespace;
+
+        private const string FileVersionAttributeName = "asp-file-version";
+        private const string SrcAttributeName = "src";
+
+        private FileVersionProvider _fileVersionProvider;
+        
+        /// <summary>
+        /// Src of the image.
```
dougbu: Source

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
```diff
@@ -0,0 +1,144 @@
...
+            var builder = new DefaultTagHelperContent();
+
+            if (context.AllAttributes.ContainsName(FileVersionAttributeName))
+            {
+                // Build a <img /> tag to match the original one in the source file.
+                // Or just add file version to the img tag.
+                BuildImageTag(attributes, builder);
+            } else
+            {
+                //No attributes matched so we have nothing to do
```
dougbu: `else` on its own line and always include a space after `//`

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
```diff
@@ -0,0 +1,144 @@
...
+            foreach (var attribute in attributes)
+            {
+                var attributeValue = attribute.Value;
+                if (FileVersion == true &&
+                    string.Equals(attribute.Name, SrcAttributeName, StringComparison.OrdinalIgnoreCase))
+                {
+                    // "src" values come from bound attributes and globbing. So anything but a non-null string is
+                    // unexpected but could happen if another helper targeting the same element does something odd.
+                    // Pass through existing value in that case.
+                    var attributeStringValue = attributeValue as string;
```
dougbu: This isn't necessary since `Src` is bound.

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
```diff
@@ -0,0 +1,144 @@
...
+                    if (attributeStringValue != null)
+                    {
+                        attributeValue = _fileVersionProvider.AddFileVersionToPath(attributeStringValue);
+                    }
+                }
+
+                builder
+                    .Append(attribute.Name)
+                    .Append("=\"")
+                    .Append(HtmlEncoder, ViewContext.Writer.Encoding, attributeValue)
```
dougbu: For every attribute _except_ `Src`, this usually double-encode the attribute value.

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
```diff
@@ -0,0 +1,144 @@
...
+        /// </summary>
+        /// <remarks>
+        /// If <c>true</c> then a query string "v" with the encoded content of the file is added.
+        /// </remarks>
+        [HtmlAttributeName(FileVersionAttributeName)]
+        public bool? FileVersion { get; set; }
+
+        [Activate]
+        [HtmlAttributeNotBound]
+        protected internal ILogger<ImageTagHelper> Logger { get; set; }
```
dougbu: unused

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
```diff
@@ -0,0 +1,144 @@
...
+        [HtmlAttributeNotBound]
+        public IMemoryCache Cache { get; set; }
+
+        [Activate]
+        [HtmlAttributeNotBound]
+        public IHtmlEncoder HtmlEncoder { get; set; }
+
+        [Activate]
+        [HtmlAttributeNotBound]
+        public IJavaScriptStringEncoder JavaScriptEncoder { get; set; }
```
dougbu: unused

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
```diff
@@ -0,0 +1,144 @@
...
+
+         /// <inheritdoc />
+        public override void Process(TagHelperContext context, TagHelperOutput output)
+        {
+            // Pass through attribute that is also a well-known HTML attribute.
```
dougbu: Entire method could be significantly simplified since `<link>` and `<script>` tag helpers only need `SetContent()` workaround due to globbing.  For example:
``` c#
{
    if (Src != null)
    {
        if (FileVersion == true)
        {
            EnsureFileVersionProvider();
        }
        else
        {
            // Pass through attribute that is also a well-known HTML attribute.
            output.CopyHtmlAttribute(SrcAttributeName, context);
        }
    }
}
```

test/WebSites/MvcTagHelpersWebSite/Controllers/MvcTagHelper_HomeController.cs
```diff
@@ -162,6 +162,11 @@ public IActionResult Link()
             return View();
         }
 
+        public IActionResult Image()
```
dougbu: Need another test in [`MvcTagHelpersTest`](https://github.com/aspnet/Mvc/blob/dev/test/Microsoft.AspNet.Mvc.FunctionalTests/MvcTagHelpersTest.cs).  This addition isn't confirmed otherwise.

dougbu:
:watch:

dpaquette:
@dougbu Simplified implementation and updated tests as per your comments

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
```diff
@@ -0,0 +1,265 @@
...
+using Microsoft.AspNet.Mvc.ModelBinding;
+using Microsoft.AspNet.Mvc.Rendering;
+using Microsoft.AspNet.Razor.Runtime.TagHelpers;
+using Microsoft.AspNet.Routing;
+using Microsoft.Framework.Caching;
+using Microsoft.Framework.Caching.Memory;
+using Microsoft.Framework.Runtime;
+
+using Moq;
+using Xunit;
```
dougbu: Please remove and sort `using`s.  That should also clean up the extra blank line.

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
```diff
@@ -0,0 +1,87 @@
...
+        private FileVersionProvider _fileVersionProvider;
+        
+        /// <summary>
+        /// Source of the image.
+        /// </summary>
+        /// <remarks>
+        /// Passed through to the generated HTML in all cases.
+        /// </remarks>
+        [HtmlAttributeName(SrcAttributeName)]
+        public string Src { get; set; }
```
NTaylorMullen: Don't bind this and just add it to the `[TargetElement]`, Aka `[TargetElement(Attributes = FileVersionAttributeName + ",src")]`. That way you don't have to restore it or check for `null`.

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
```diff
@@ -0,0 +1,87 @@
...
+        public ViewContext ViewContext { get; set; }
+
+        [Activate]
+        [HtmlAttributeNotBound]
+        public IMemoryCache Cache { get; set; }        
+
+         /// <inheritdoc />
+        public override void Process(TagHelperContext context, TagHelperOutput output)
+        {
+            if (Src != null)
```
NTaylorMullen: With https://github.com/aspnet/Mvc/pull/2516/files#r30006170 you wont need to check for null here because it'll always be provided. Also wont need to restore the src attribute on line 71.

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
NTaylorMullen: Why nullable? It's in the `TargetElement` so it'll always be provided.

test/Microsoft.AspNet.Mvc.FunctionalTests/MvcTagHelpersTest.cs
```diff
@@ -47,6 +47,8 @@ public class MvcTagHelpersTest
         [InlineData("Link", null)]
         // Testing the ScriptTagHelper
         [InlineData("Script", null)]
+        //Testing the ImageTagHelper
```
NTaylorMullen: `// Testing the ImageTagHelper`

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
```diff
@@ -0,0 +1,265 @@
...
+                    { "asp-file-version", "true" }
+                });
+            var output = MakeImageTagHelperOutput(
+                attributes: new TagHelperAttributeList
+                {
+                    { "alt", new HtmlString("alt text") },
+                    { "data-extra", new HtmlString("something") },                    
+                    { "title", new HtmlString("Image title") },
+                    { "src", "testimage.png" },
+                    { "asp-file-version", "true" }
```
dougbu: Bit unrealistic.  `src` and `asp-file-version` are _not_ in the `TagHelperOutput.Attributes` collection by default.  Bound attributes appear only in the `TagHelperExecutionContext.AllAttributes` collection when tag helper execution begins.  That's why the `CopyHtmlAttribute()` call is necessary in the `` tag helper.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
```diff
@@ -0,0 +1,265 @@
...
+                { "alt", new HtmlString("alt text") },
+            });
+            var hostingEnvironment = MakeHostingEnvironment();
+            var viewContext = MakeViewContext("/bar");
+            var helper = new ImageTagHelper
+            {
+                HostingEnvironment = hostingEnvironment,
+                ViewContext = viewContext,
+                Src = "/bar/images/image.jpg",
+                FileVersion = true,
```
dougbu: Need a test with `FileVersion == false` to confirm `CopyHtmlAttribute()` is called correctly.

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
dougbu: @NTaylorMullen your advice is incorrect.  This is **the** way to ensure the tag helper gets the non-HTML encoded value of the `src` attribute.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
```diff
@@ -0,0 +1,265 @@
...
+                });
+
+            var expectedOutput = MakeImageTagHelperOutput(
+                attributes: new TagHelperAttributeList
+                {
+                    { "alt", new HtmlString("alt text") },
+                    { "data-extra", new HtmlString("something") },
+                    { "title", new HtmlString("Image title") },
+                    { "src", "testimage.png?v=f4OxZX_x_FO5LcGBSKHWXfwtSx-j1ncoSt3SABJtkGk" },
+                    { "asp-file-version", "true" }
```
dougbu: in realistic scenario `asp-file-version` would not appear in the `Attributes` collection after processing

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
```diff
@@ -0,0 +1,265 @@
...
+            };
+
+            // Act
+            helper.Process(context, output);
+
+            // Assert
+            Assert.True(output.Content.IsEmpty);
+            Assert.Equal("img", output.TagName);
+            Assert.Equal(2, output.Attributes.Count);
+            TagHelperAttribute srcAttribute = Assert.Single(output.Attributes, attr => attr.Name.Equals("src"));
```
dougbu: `var`

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
```diff
@@ -0,0 +1,265 @@
...
+                Cache = MakeCache(),
+            };
+
+            // Act
+            helper.Process(context, output);
+            // Assert
+            Assert.True(output.Content.IsEmpty);
+            Assert.Equal("img", output.TagName);
+            Assert.Equal(2, output.Attributes.Count);
+            TagHelperAttribute srcAttribute = Assert.Single(output.Attributes, attr => attr.Name.Equals("src"));
```
dougbu: `var`

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
```diff
@@ -0,0 +1,265 @@
...
+                viewData,
+                Mock.Of<ITempDataDictionary>(),
+                TextWriter.Null);
+
+            return viewContext;
+        }
+
+        private static TagHelperContext MakeTagHelperContext(
+            TagHelperAttributeList attributes = null,
+            string content = null)
```
dougbu: Simplify method signature: No need for `attributes` to be optional.  No need for `content` at all.

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
NTaylorMullen: Ah, fair point since we're modifying it. :+1:

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
```diff
@@ -0,0 +1,265 @@
...
+                uniqueId: Guid.NewGuid().ToString("N"),
+                getChildContentAsync: () =>
+                {
+                    var tagHelperContent = new DefaultTagHelperContent();
+                    tagHelperContent.SetContent(content);
+                    return Task.FromResult<TagHelperContent>(tagHelperContent);
+                });
+        }
+
+        private static TagHelperOutput MakeImageTagHelperOutput(TagHelperAttributeList attributes = null)
```
dougbu: No need for `attributes` to be optional.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
```diff
@@ -0,0 +1,265 @@
...
+                .Returns(emptyDirectoryContents.Object);
+            mockFileProvider.Setup(fp => fp.GetFileInfo(It.IsAny<string>()))
+                .Returns(mockFile.Object);
+            var hostingEnvironment = new Mock<IHostingEnvironment>();
+            hostingEnvironment.Setup(h => h.WebRootFileProvider).Returns(mockFileProvider.Object);
+
+            return hostingEnvironment.Object;
+        }
+
+        private static IApplicationEnvironment MakeApplicationEnvironment(string applicationName = "testApplication")
```
dougbu: Unused; remove.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
```diff
@@ -0,0 +1,265 @@
...
+        {
+            var applicationEnvironment = new Mock<IApplicationEnvironment>();
+            applicationEnvironment.Setup(a => a.ApplicationName).Returns(applicationName);
+            return applicationEnvironment.Object;
+        }
+
+        private static IMemoryCache MakeCache(object result = null)
```
dougbu: Remove `result` parameter.  No caller passes an argument.

test/WebSites/MvcTagHelpersWebSite/Controllers/MvcTagHelper_HomeController.cs
dougbu: thx

dougbu:
:watch: very close

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
NTaylorMullen: Probably still worth adding it to `TargetElement` since it no-ops otherwise. Also, since it's a `string` a `null` value wont be possible if its in `TargetElement`.

dougbu: > since it's a `string` a `null` value wont be possible if its in `TargetElement`.

Side note: Put that way, it sounds like a bug.  Razor doesn't do what author likely meant if author writes `src="@null"` for a bound `string` attribute.

But I agree having both `SrcAttributeName` and `FileVersionAttributeName` in the `TargetElement.Attributes` value makes sense.  Avoids some useless executions of the helper.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
dpaquette: just to confirm...we can use var in tests but shouldn't in other code?

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
dpaquette: I'm wondering what this. It is also unused in the test for the Link tag helper

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
dougbu: No.  Our [Guidelines](https://github.com/aspnet/Home/wiki/Engineering-guidelines#usage-of-the-var-keyword) say "The var keyword is to be used as much as the compiler will allow."  (Everywhere.)

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
dougbu: Might have been needed at some point there.  But you're not using it here.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
dpaquette: My bad. For some reason I read that part of the guideline as 'is not to be used'.

dpaquette:
@dougbu @NTaylorMullen Thanks for the review. I have updated based on your feedback. In retrospect, basing this implementation on the Link and Script Tag Helpers was probably a bad idea :smile:

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
```diff
@@ -0,0 +1,297 @@
...
+using Microsoft.Framework.Caching;
+using Microsoft.Framework.Caching.Memory;
+using Microsoft.Framework.Runtime;
+using Moq;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
```
dougbu: Sorry, should have mentioned that we use the old VS default here: System namespaces first.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
dougbu: better but please also remove the `src` attribute from the original `output.Attributes` collection (line 47)

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
dougbu: thx

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
dougbu: Please remove this method.

dougbu:
:watch:

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
```diff
@@ -0,0 +1,288 @@
...
+        public void PreservesOrderOfSourceAttributesWhenRun()
+        {
+            // Arrange
+            var context = MakeTagHelperContext(
+                attributes: new TagHelperAttributeList
+                {
+                    { "alt", new HtmlString("alt text") },
+                    { "data-extra", new HtmlString("something") },                    
+                    { "title", new HtmlString("Image title") },
+                    { "src", "testimage.png" },
```
dougbu: Please remove this line.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
dougbu: Thanks.  `output` looks good now.

test/Microsoft.AspNet.Mvc.FunctionalTests/MvcTagHelpersTest.cs
```diff
@@ -47,6 +47,8 @@ public class MvcTagHelpersTest
         [InlineData("Link", null)]
         // Testing the ScriptTagHelper
         [InlineData("Script", null)]
+        // Testing the ImageTagHelper
+        [InlineData("Image", null)]
```
dougbu: Odd that MvcTagHelpersWebSite.MvcTagHelper_Home.Image.html shows up here as a binary file.  Anything unusual about it on your end?  (Content looks fine when I click on View.)

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
dougbu: Thanks

dougbu:
:shipit:

Please rebase your branch to latest dev then squash your commits down to one (`git rebase -i head~8` and change `pick` to `s` or `squash` for all commits except the first.  We won't lose attribution when merging if you do that.

test/Microsoft.AspNet.Mvc.FunctionalTests/MvcTagHelpersTest.cs
dpaquette: Looks line an encoding problem. I will try saving as UTF-8

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
```diff
@@ -0,0 +1,84 @@
...
+        public ViewContext ViewContext { get; set; }
+
+        [Activate]
+        [HtmlAttributeNotBound]
+        public IMemoryCache Cache { get; set; }
+
+        /// <inheritdoc />
+        public override void Process(TagHelperContext context, TagHelperOutput output)
+        {
+            if (FileVersion == true)
```
NTaylorMullen: `if (FileVersion)`

NTaylorMullen:
:shipit:

dpaquette:
@dougbu @NTaylorMullen Thanks guys. I did a rebase to latest dev and squashed my commits.

Eilon:
BTW I logged https://github.com/aspnet/Mvc/issues/2540 to track changing the name of the `FileVersion` property.

dougbu:
ab4d2ee Thanks for your contribution!

Issue #2248, opened 2015-03-24
Attribute routing throws ambiguous exception for base controller 


harshgMSFT:
Consider the following 

```c#
[Route("base/[action]")]
    public class BaseController : Controller
    {
        public virtual void Action()
        {
        }
    }

    [Route("derived/[action]")]
    public class DerivedController : BaseController
    {
        public override void Action()
        {
        }
    }
```

When attempting to hit `base/Action` we get an ambiguous route exception.

rynowak:
@harshgMSFT and I spent some time debugging this, the reason is that the entries for `DerivedController::Action` and `BaseController::Action` each end up defining the route `base/Action`.

rynowak:
46db71cfceae350564135ecd2a810660e70a7c42
8a476b5

PR #2517, opened 2015-05-07
Fix for #2248

rynowak:
Split into 2 parts for your convenience and enjoyment!

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ApplicationModels/DefaultControllerModelBuilder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ApplicationModels/DefaultControllerModelBuilder.cs
@@ -76,10 +76,57 @@ namespace Microsoft.AspNet.Mvc.ApplicationModels
         /// <returns>A <see cref="ControllerModel"/> for the given <see cref="TypeInfo"/>.</returns>
         protected virtual ControllerModel CreateControllerModel([NotNull] TypeInfo typeInfo)
         {
+            // For attribute routes on a controller, we want want to support 'overriding' routes on a derived
+            // class. So we need to walk up the hierarchy looking for the first class to define routes.
+            //
+            // Then we want to 'filter' the set of attributes, so that only the effective routes apply.
+            var currentTypeInfo = typeInfo;
+            var objectTypeInfo = typeof(object).GetTypeInfo();
+
+            IRouteTemplateProvider[] routeAttributes = null;
+
+            do
+            {
+                routeAttributes = currentTypeInfo
+                        .GetCustomAttributes(inherit: false)
+                        .OfType<IRouteTemplateProvider>()
+                        .ToArray();
+
+                if (routeAttributes.Length > 0)
+                {
+                    // Found 1 or more route attributes.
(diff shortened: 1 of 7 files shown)


test/WebSites/RoutingWebSite/RoutingWebSite.xproj
```diff
@@ -10,6 +10,12 @@
     <BaseIntermediateOutputPath Condition="'$(BaseIntermediateOutputPath)'=='' ">..\..\..\artifacts\obj\$(MSBuildProjectName)</BaseIntermediateOutputPath>
     <OutputPath Condition="'$(OutputPath)'=='' ">..\..\..\artifacts\bin\$(MSBuildProjectName)\</OutputPath>
   </PropertyGroup>
+  <PropertyGroup Label="Configuration" Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <AssemblyName>RoutingWebSite</AssemblyName>
+  </PropertyGroup>
+  <PropertyGroup Label="Configuration">
+    <RootNamespace>RoutingWebSite</RootNamespace>
+  </PropertyGroup>
```
dougbu: these .xproj properties are useless. don't bother unless VS makes these changes repeatedly.

src/Microsoft.AspNet.Mvc.Core/ApplicationModels/DefaultControllerModelBuilder.cs
```diff
@@ -76,10 +76,57 @@ public ControllerModel BuildControllerModel([NotNull] TypeInfo typeInfo)
...
+                    filteredAttributes.Add(attribute);
+                }
+            }
+            filteredAttributes.AddRange(routeAttributes);
+
+            attributes = filteredAttributes.ToArray();
+
             var controllerModel = new ControllerModel(typeInfo, attributes);
+            AddRange(
+                controllerModel.AttributeRoutes, routeAttributes.Select(a => new AttributeRouteModel(a)));
```
dougbu: ick, YOLO wrapping

test/WebSites/RoutingWebSite/RoutingWebSite.xproj
rynowak: VS did this. I don't think I've ever edited an xproj on purpose.

dougbu:
Change looks good for unnamed routes. But what about uniquely-named routes e.g.
``` c#
  [Route("A", Name="Inherited")]
  public MyController
  {
  }

  [Route("B")]
  public DerivedController : MyController
  {
  }
```
Should the route named "Inherited" be available for the derived controller?

If that decision has already been made, :shipit:

rynowak:
@dougbu - Your intuition is right, we're no longer supporting that on purpose. This feature very quickly becomes a pit of failure when you're try to leverage the 'inherit + add' semantics (removed by this change).

In the example that you give, assume that there's one action in each controller. You'd have a route `A` defined for both of them. You could use the name `Inherited` to generate a link - which would give you a 500 when you visit it....

If you really want to do something like this, you could still easily do it in a convention, which would give much more control.

PR #2518, opened 2015-05-07
Custom stream writer which avoids writing BOM and does not close the stream

kichalla:
@pranavkm @rynowak @Tratcher 

This stream writer does the following:
- Does not write the BOM even if the encoding requires it
- Does not close the provided stream as the response stream is not owned by us
- Does not flush on close/dispose as this can cause chunked encoding.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/Formatters/JsonOutputFormatter.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Formatters/JsonOutputFormatter.cs
@@ -70,8 +70,7 @@ namespace Microsoft.AspNet.Mvc
             var response = context.ActionContext.HttpContext.Response;
             var selectedEncoding = context.SelectedEncoding;
 
-            using (var nonDisposableStream = new NonDisposableStream(response.Body))
-            using (var writer = new StreamWriter(nonDisposableStream, selectedEncoding, 1024, leaveOpen: true))
+            using (var writer = new HttpResponseStreamWriter(response.Body, selectedEncoding))
             {
                 WriteObject(writer, context.Object);
             }
--- a/src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
@@ -1,0 +1,235 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.AspNet.Mvc
+{
+    /// <summary>
+    /// Writes to the <see cref="Stream"/> using the supplied <see cref="Encoding"/>.
+    /// It does not write the BOM and also does not close the stream.
+    /// </summary>
+    public class HttpResponseStreamWriter : TextWriter
+    {
+        private const int DefaultBufferSize = 1024;
+        private readonly Stream _stream;
+        private Encoder _encoder;
+        private byte[] _byteBuffer;
+        private char[] _charBuffer;
+        private int _charBufferSize;
+        private int _charBufferCount;
+
+        public HttpResponseStreamWriter(Stream stream, Encoding encoding)
+            : this(stream, encoding, DefaultBufferSize)
+        {
+        }
+
+        public HttpResponseStreamWriter([NotNull] Stream stream, [NotNull] Encoding encoding, int bufferSize)
+        {
+            _stream = stream;
+            Encoding = encoding;
+            _encoder = encoding.GetEncoder();
+            _charBufferSize = bufferSize;
+            _charBuffer = new char[bufferSize];
+            _byteBuffer = new byte[encoding.GetMaxByteCount(bufferSize)];
+        }
+
+        public override Encoding Encoding { get; }
+
+        public override void Write(char value)
+        {
+            if (_charBufferCount == _charBufferSize)
+            {
+                FlushInternal();
+            }
+
+            _charBuffer[_charBufferCount] = value;
+            _charBufferCount++;
+        }
+
+        public override void Write(char[] values, int index, int count)
+        {
+            if (values == null)
+            {
+                return;
+            }
+
+            while (count > 0)
+            {
+                if (_charBufferCount == _charBufferSize)
+                {
+                    FlushInternal();
+                }
+
+                CopyToCharBuffer(values, ref index, ref count);
+            }
+        }
+
+        public override void Write(string value)
+        {
+            if (value == null)
+            {
+                return;
+            }
+
+            var count = value.Length;
+            var index = 0;
+            while (count > 0)
+            {
+                if (_charBufferCount == _charBufferSize)
+                {
+                    FlushInternal();
+                }
+
+                CopyToCharBuffer(value, ref index, ref count);
+            }
+        }
+
+        public override async Task WriteAsync(char value)
+        {
+            if (_charBufferCount == _charBufferSize)
+            {
+                await FlushInternalAsync();
+            }
+
+            _charBuffer[_charBufferCount] = value;
+            _charBufferCount++;
+        }
+
+        public override async Task WriteAsync(char[] values, int index, int count)
+        {
+            if (values == null)
+            {
+                return;
+            }
+
+            while (count > 0)
+            {
+                if (_charBufferCount == _charBufferSize)
+                {
+                    await FlushInternalAsync();
+                }
+
+                CopyToCharBuffer(values, ref index, ref count);
+            }
+        }
+
+        public override async Task WriteAsync(string value)
+        {
+            if (value == null)
+            {
+                return;
+            }
+
+            var count = value.Length;
+            var index = 0;
+            while (count > 0)
+            {
+                if (_charBufferCount == _charBufferSize)
+                {
+                    await FlushInternalAsync();
+                }
+
+                CopyToCharBuffer(value, ref index, ref count);
+            }
+        }
+
+        // We want to flush the stream when Flush/FlushAsync is explicitly
+        // called by the user (example: from a Razor view).
+
+        public override void Flush()
+        {
+            FlushInternal(true, true);
+        }
+
+        public override async Task FlushAsync()
+        {
+            await FlushInternalAsync(true, true);
+        }
+
+        // Do not flush the stream on Dispose, as this will cause response to be
+        // sent in chunked encoding in case of Helios.
+        protected override void Dispose(bool disposing)
+        {
+            FlushInternal(flushStream: false, flushEncoder: true);
+        }
+
+        private void FlushInternal(bool flushStream = false, bool flushEncoder = false)
+        {
+            if (_charBufferCount == 0)
+            {
+                return;
+            }
+
+            var count = _encoder.GetBytes(_charBuffer, 0, _charBufferCount, _byteBuffer, 0, flushEncoder);
+            if (count > 0)
+            {
+                _stream.Write(_byteBuffer, 0, count);
+            }
+
+            _charBufferCount = 0;
+
+            if (flushStream)
+            {
+                _stream.Flush();
+            }
+        }
+
+        private async Task FlushInternalAsync(bool flushStream = false, bool flushEncoder = false)
+        {
+            if (_charBufferCount == 0)
+            {
+                return;
+            }
+
+            var count = _encoder.GetBytes(_charBuffer, 0, _charBufferCount, _byteBuffer, 0, flushEncoder);
+            if (count > 0)
+            {
+                await _stream.WriteAsync(_byteBuffer, 0, count);
+            }
+
+            _charBufferCount = 0;
+
+            if (flushStream)
+            {
+                await _stream.FlushAsync();
+            }
+        }
+
+        private void CopyToCharBuffer(string value, ref int index, ref int count)
+        {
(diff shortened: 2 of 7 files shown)


src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
```diff
@@ -0,0 +1,179 @@
...
+        {
+            WriteRemainingBytes();
+        }
+#endif
+
+        protected override void Dispose(bool disposing)
+        {
+            //In CoreCLR this is equivalent to Close.
+            WriteRemainingBytes();
+        }
```
rynowak: Should it just call Close then?

src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
```diff
@@ -0,0 +1,179 @@
...
+            get
+            {
+                if (_buffer == null)
+                {
+                    _buffer = new byte[_bufferSize];
+                }
+
+                return _buffer;
+            }
+        }
```
rynowak: properties before methods

src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
```diff
@@ -0,0 +1,179 @@
...
+            _stream = stream;
+            Encoding = encoding;
+            _bufferSize = bufferSize;
+        }
+
+        public override Encoding Encoding { get; }
+
+        public override void Write(char value)
+        {
+            var bytes = Encoding.GetBytes(new[] { value });
```
rynowak: This is going to cause issues. Take a look at how streamwriter does (using a buffer and the `Encoder` class).

kichalla: :+1: ...updated the code now...

src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
kichalla: But `Close` is not available in CoreClr

kichalla:
Updated. Please take a look /cc: @rynowak @pranavkm @Tratcher

src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
```diff
@@ -0,0 +1,280 @@
...
+            if (_charBufferCount == _charBufferSize)
+            {
+                Flush();
+            }
+
+            _charBuffer[_charBufferCount] = value;
+            _charBufferCount++;
+        }
+
+        public override void Write(char[] values)
```
pranavkm: Would it work if we don't override this method? It probably calls into `Write(char[], int, int)`

src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
```diff
@@ -0,0 +1,280 @@
...
+            }
+
+            while (count > 0)
+            {
+                if (_charBufferCount == _charBufferSize)
+                {
+                    Flush();
+                }
+
+                int remaining = _charBufferSize - _charBufferCount;
```
pranavkm: var remaining = `Math.Min(_charBufferSize - _charBufferCount, count);`

src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
```diff
@@ -0,0 +1,280 @@
...
+        private int _charBufferCount;
+
+        public HttpResponseStreamWriter(Stream stream, Encoding encoding)
+            : this(stream, encoding, DefaultBufferSize)
+        {
+        }
+
+        public HttpResponseStreamWriter([NotNull] Stream stream, [NotNull] Encoding encoding, int bufferSize)
+        {
+            this._stream = stream;
```
pranavkm: `this`?

src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
```diff
@@ -0,0 +1,280 @@
...
+                    Flush();
+                }
+
+                int remaining = _charBufferSize - _charBufferCount;
+                if (remaining > count)
+                {
+                    remaining = count;
+                }
+
+                Buffer.BlockCopy(
```
pranavkm: This block of code starting from the `if` could be moved into a separate method so you get to avoid repeating it in the async code block.

src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
```diff
@@ -0,0 +1,280 @@
...
+            int count = value.Length;
+            int index = 0;
+            while (count > 0)
+            {
+                if (_charBufferCount == _charBufferSize)
+                {
+                    await FlushAsync();
+                }
+
+                int charBufferFreeSpace = _charBufferSize - _charBufferCount;
```
pranavkm: `remaining` to be consistent.

src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
```diff
@@ -0,0 +1,280 @@
...
+        }
+#endif
+        protected override void Dispose(bool disposing)
+        {
+            // In CoreClr this is equivalent to calling Close()
+
+            Flush(flushStream: false, flushEncoder: true);
+        }
+
+        private void Flush(bool flushStream = false, bool flushEncoder = false)
```
pranavkm: Call this `FlushInternal` and remove the default arguments.

src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
```diff
@@ -0,0 +1,280 @@
...
+                    destinationIndex: _charBufferCount,
+                    count: charBufferFreeSpace);
+
+                _charBufferCount += charBufferFreeSpace;
+                index += charBufferFreeSpace;
+                count -= charBufferFreeSpace;
+            }
+        }
+
+        // Do not flush the stream on Close/Dispose, as this will cause response to be
```
pranavkm: Is this comment accurate?

pranavkm: Oh nevermind. This makes sense. Maybe put this closer to the `Close` \ `Dispose methods.

test/Microsoft.AspNet.Mvc.Core.Test/HttpResponseStreamWriterTest.cs
```diff
@@ -0,0 +1,337 @@
...
+            // Assert
+            Assert.Equal(0, stream.DisposeCallCount);
+        }
+
+        [Theory]
+        [InlineData(1023)]
+        [InlineData(1024)]
+        [InlineData(1050)]
+        [InlineData(2048)]
+        public async Task FlushesBuffer_OnClose(int byteLength)
```
pranavkm: Isn't this `charLength`?

test/Microsoft.AspNet.Mvc.Core.Test/HttpResponseStreamWriterTest.cs
```diff
@@ -0,0 +1,337 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Microsoft.AspNet.Mvc
+{
+    public class HttpResponseStreamWriterTest
+    {
+        [Fact]
+        public async Task DoesNotWriteBOM()
```
pranavkm: Tests look good to start with - but we should cover other Encodings as part of the writer.

src/Microsoft.AspNet.Mvc.Core/Formatters/JsonOutputFormatter.cs
```diff
@@ -70,8 +70,7 @@ public override Task WriteResponseBodyAsync(OutputFormatterContext context)
             var response = context.ActionContext.HttpContext.Response;
             var selectedEncoding = context.SelectedEncoding;
 
-            using (var nonDisposableStream = new NonDisposableStream(response.Body))
-            using (var writer = new StreamWriter(nonDisposableStream, selectedEncoding, 1024, leaveOpen: true))
+            using (var writer = new HttpResponseStreamWriter(response.Body, selectedEncoding))
```
pranavkm: :100:

pranavkm:
:watch:

src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
kichalla: good catch...text writer is doing exactly this...will remove..

src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
kichalla: Sure...

test/Microsoft.AspNet.Mvc.Core.Test/HttpResponseStreamWriterTest.cs
kichalla: yeah, i thought about it...in many of the tests here we verify the stream length(which is in bytes) against this length...and the char length here matches with the byte length as I am using only single byte characters like 'a'..

src/Microsoft.AspNet.Mvc.Core/Formatters/JsonOutputFormatter.cs
rynowak: :heart_eyes_cat:

src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
```diff
@@ -0,0 +1,244 @@
...
+
+#if DNX451
+        public override void Close()
+        {
+            FlushInternal(flushStream: false, flushEncoder: true);
+        }
+#endif
+        protected override void Dispose(bool disposing)
+        {
+            // In CoreClr this is equivalent to calling Close()
```
rynowak: My comment still stand here, should one of these call the other (full .net)?

test/Microsoft.AspNet.Mvc.Core.Test/Formatters/JsonOutputFormatterTests.cs
```diff
@@ -142,11 +142,7 @@ public void Creates_SerializerSettings_ByDefault()
             var formattedContent = "\"" + content + "\"";
             var mediaType = string.Format("application/json; charset={0}", encodingAsString);
             var encoding = CreateOrGetSupportedEncoding(formatter, encodingAsString, isDefaultEncoding);
-            var preamble = encoding.GetPreamble();
-            var data = encoding.GetBytes(formattedContent);
-            var expectedData = new byte[preamble.Length + data.Length];
-            Buffer.BlockCopy(preamble, 0, expectedData, 0, preamble.Length);
-            Buffer.BlockCopy(data, 0, expectedData, preamble.Length, data.Length);
+            var expectedData = encoding.GetBytes(formattedContent);
```
rynowak: Is this a fix (I assume the test was verifying the BOM, which is incorrect)?

kichalla: yes, exactly...

rynowak:
This is a looking really good. :watch: for a few more tests for a different encoding

src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
kichalla: :+1: ...in full .net call on `Close` in turn calls `Dispose`, so I can remove this...

kichalla:
Updated. /cc: @pranavkm @rynowak

rynowak:
:shipit: once pranav is happy

src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
```diff
@@ -0,0 +1,235 @@
...
+        }
+
+        // Do not flush the stream on Dispose, as this will cause response to be
+        // sent in chunked encoding in case of Helios.
+        protected override void Dispose(bool disposing)
+        {
+            FlushInternal(flushStream: false, flushEncoder: true);
+        }
+
+        private void FlushInternal(bool flushStream = false, bool flushEncoder = false)
```
pranavkm: Can you remove the default args?

test/Microsoft.AspNet.Mvc.Core.Test/ActionResults/ViewExecutorTest.cs
```diff
@@ -84,39 +81,16 @@ public class ViewExecutorTest
...
-        // prior to it filling up will not be written to the underlying stream once an exception is thrown.
-        [Theory]
-        [MemberData(nameof(ExecuteAsync_DoesNotWriteToResponse_OnceExceptionIsThrownData))]
-        public async Task ExecuteAsync_DoesNotWriteToResponse_OnceExceptionIsThrown(int writtenLength, int expectedLength)
+        [Fact]
+        public async Task ExecuteAsync_DoesNotWriteToResponse_OnceExceptionIsThrown()
         {
             // Arrange
-            var longString = new string('a', writtenLength);
+            int expectedLength = 0;
```
pranavkm: `var`

pranavkm:

PR #2519, opened 2015-05-07
Removing superfluous types and methods from Common.

pranavkm:


src/Microsoft.AspNet.Mvc.Core/ParameterBinding/ModelBindingHelper.cs
```diff
@@ -532,5 +532,17 @@ public static object ConvertValuesToCollectionType<T>(Type modelType, IList<T> v
                 return null;
             }
         }
+
+        private static Type[] GetTypeArgumentsIfMatch([NotNull] Type closedType, Type matchingOpenType)
```
dougbu: This has always been a strange method.  Why not use `TypeExtensions.ExtractGenericInterface()`?  Current implementation is much more constrained.

dougbu:
lgtm in general but :watch: for one question

src/Microsoft.AspNet.Mvc.Common/TypeExtensions.cs
```diff
@@ -10,85 +10,23 @@ namespace Microsoft.AspNet.Mvc
 {
     internal static class TypeExtensions
     {
-        // NOTE: Do not make #105 worse! Do not add new extension methods that conflict w/ .NET 4.5 methods. The
-        // exising NETFX_CORE || DNXCORE50 methods should go away (soon).
-#if NETFX_CORE || DNXCORE50
-        private static bool EqualTo([NotNull] this Type[] t1, [NotNull] Type[] t2)
+        public static TypeInfo ExtractGenericInterface([NotNull] this Type queryType, Type interfaceType)
```
rynowak: Can we convert this into a non-extension method and just use it in `TypeHelper`?

src/Microsoft.AspNet.Mvc.Core/BindAttribute.cs
```diff
@@ -40,7 +41,8 @@ public BindAttribute(params string[] include)
         /// </param>
         public BindAttribute([NotNull] Type predicateProviderType)
         {
-            if (!typeof(IPropertyBindingPredicateProvider).IsAssignableFrom(predicateProviderType))
+            if (!typeof(IPropertyBindingPredicateProvider).GetTypeInfo()
+                    .IsAssignableFrom(predicateProviderType.GetTypeInfo()))
```
rynowak: Isn't there a provided extension method for this? http://packagesearch.azurewebsites.net/?q=IsAssignableFrom

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
```diff
@@ -88,7 +89,7 @@ public static class AttributeMatcher
             {
                 if (!context.AllAttributes.ContainsName(attribute) ||
                     context.AllAttributes[attribute] == null ||
-                    (typeof(string).IsAssignableFrom(context.AllAttributes[attribute].Value.GetType()) &&
+                    (typeof(string).GetTypeInfo().IsAssignableFrom(context.AllAttributes[attribute].Value.GetType().GetTypeInfo()) &&
```
rynowak: isn't this just `context.AllAttributes[attribute].Value is string`?

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelMetadata.cs
```diff
@@ -2,7 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
+using System.Reflection;
```
rynowak: Is it possible to remove the reference from `Abstractions` -> `Common`?

pranavkm: Perhaps. I'll have a look at the project.jsons

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
pranavkm: Yup.

src/Microsoft.AspNet.Mvc.Core/BindAttribute.cs
pranavkm: My only issue with using that extension is that you end up with an unused `System.Reflection` using in DNX451. ifdefing that seems like more trouble than just calling `GetTypeInfo` twice over.

src/Microsoft.AspNet.Mvc.Core/ParameterBinding/ModelBindingHelper.cs
pranavkm: The semantics are different in this method - we might be dealing with non-interface types (such as `KeyValuePair<>`).

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelMetadata.cs
pranavkm: Sadly not. We use a couple of methods from `TypeHelper` in `ModelMetadata`, `ValueProviderResult` etc.

rynowak: I think this is fine for a first cut, but we might want to clean it up further by moving those things to core one way or another.

src/Microsoft.AspNet.Mvc.Core/BindAttribute.cs
rynowak: It only seems that way because you made the change already :laughing:

It's pretty easy to paste the same 3 lines in all the files.

src/Microsoft.AspNet.Mvc.Core/ParameterBinding/ModelBindingHelper.cs
rynowak: This and `GetPossibleBinderInstanceType` should just move to `GenericModelBinder` which is where they are used, and probably just simplified to do what they are supposed to do to support `KeyValuePair<>`. That's the only call I can find to either of these.

pranavkm: :+1:

src/Microsoft.AspNet.Mvc.Core/BindAttribute.cs
pranavkm: :dancers:

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelMetadata.cs
pranavkm: Yup - the intent was to trim down Common so it's easier to identify things that are truly common (and perhaps could go end up in the Common repo).

pranavkm:
Updated.

src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
```diff
@@ -151,9 +151,8 @@ private void ActivateProperties(object controller, Type containerType, Dictionar
                 }
 
                 // Determine T if this is an ICollection<T> property.
-                var collectionTypeArguments = propertyType
-                    .ExtractGenericInterface(typeof(ICollection<>))
-                    ?.GenericTypeArguments;
+                var collectionTypeArguments = TypeHelper.ExtractGenericInterface(
+                    propertyType, typeof(ICollection<>))?.GenericTypeArguments;
```
rynowak: #YOLOBREAK

rynowak:
:shipit:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
```diff
@@ -555,9 +555,8 @@ internal void ProcessDto(ModelBindingContext bindingContext, ComplexModelDto dto
             // Determine T if this is an ICollection<T> property. No need for a T[] case because CanUpdateProperty()
             // ensures property is either settable or not an array. Underlying assumption is that CanUpdateProperty()
             // and SetProperty() are overridden together.
-            var collectionTypeArguments = propertyExplorer.ModelType
-                .ExtractGenericInterface(typeof(ICollection<>))
-                ?.GenericTypeArguments;
+            var collectionTypeArguments = TypeHelper.ExtractGenericInterface(
+                propertyExplorer.ModelType, typeof(ICollection<>))?.GenericTypeArguments;
```
dougbu: previous wrapping was much easier to grok and you're now putting multiple arguments on the same line after a break

src/Microsoft.AspNet.Mvc.Common/TypeExtensions.cs
```diff
@@ -1,122 +0,0 @@
...
-                                                            .Select(p => p.ParameterType).ToArray().EqualTo(types));
-        }
-#endif
-
-        public static Type BaseType([NotNull] this Type type)
-        {
-            return type.GetTypeInfo().BaseType;
-        }
-
-        public static Type ExtractGenericInterface([NotNull] this Type queryType, Type interfaceType)
```
dougbu: could you please leave this method as-is for a very short time? the intermediate change will just my move of this method to the Common repo more difficult.

dougbu:
:shipit:

src/Microsoft.AspNet.Mvc.Common/TypeExtensions.cs
dougbu: Why did you check in a duplicate copy of this method?  `TypeHelper.ExtractGenericInterface()` now exists but isn't used in dev.

PR #2520, opened 2015-05-08
[Design] React to aspnet/Razor#89 fix

dougbu:
- use `IDictionary<string, TValue>` support in `<a/>` and `<form/>` tag helpers
- remove `TagHelperOutputExtensions.FindPrefixedAttributes()`
- add error for tag helper dictionary properties where `TValue` is `ModelExpression`

Tests
- `TagHelperAttributeDescriptor` constructor requires 2 more arguments
- arrange `AnchorTagHelper` and `FormTagHelper` correctly
 - also expect `routeValues != null` in calls to `IHtmlGenerator`
- update .cshtml files to use a correct syntax for `object` values
 - fallout of using `RouteValues` in `<a/>` and `<form/>` tag helpers

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor.Host/MvcRazorParser.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor.Host/MvcRazorParser.cs
@@ -62,10 +62,11 @@ namespace Microsoft.AspNet.Mvc.Razor
             {
                 foreach (var attributeDescriptor in descriptor.Attributes)
                 {
-                    if (string.Equals(
-                        attributeDescriptor.PrefixedValueTypeName,
-                        _modelExpressionTypeName,
-                        StringComparison.Ordinal))
+                    if (attributeDescriptor.IsIndexer &&
+                        string.Equals(
+                            attributeDescriptor.TypeName,
+                            _modelExpressionTypeName,
+                            StringComparison.Ordinal))
                     {
                         errorSink.OnError(SourceLocation.Undefined, Resources.FormatMvcRazorParser_InvalidPropertyType(
                             descriptor.TypeName,
--- a/src/Microsoft.AspNet.Mvc.Razor.Host/MvcRazorParser.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor.Host/MvcRazorParser.cs
@@ -1,8 +1,10 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
+using Microsoft.AspNet.Mvc.Razor.Host;
 using Microsoft.AspNet.Razor;
 using Microsoft.AspNet.Razor.Generator.Compiler;
 using Microsoft.AspNet.Razor.Parser;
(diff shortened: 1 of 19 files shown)


test/WebSites/MvcTagHelpersWebSite/Views/MvcTagHelper_Home/Index.cshtml
```diff
@@ -45,9 +45,9 @@
         </a>
     </div>
     <div>
-        <a asp-route="areaRoute" asp-route-area="Customer"
-           asp-route-controller="Customer"
-           asp-route-action="Index"
+        <a asp-route="areaRoute" asp-route-area='"Customer"'
+           asp-route-controller='"Customer"'
+           asp-route-action='"Index"'
```
dougbu: Razor's handling of `string` values for `object` tag helper properties or dictionary entries is not great. Neither `"&quot;Customer&quot;"` nor `"\"Customer\""` work here; the first because this isn't an HTML context and the second because C# `string` parsing is, lets-call-it, minimal.  `"@("Customer")"` is the other way I could write this.

dougbu:
@NTaylorMullen updated

NTaylorMullen:
:shipit:

src/Microsoft.AspNet.Mvc.Razor.Host/MvcRazorParser.cs
```diff
@@ -50,7 +56,28 @@ public class MvcRazorParser : RazorParser
...
+                    if (attributeDescriptor.IsIndexer &&
+                        string.Equals(
+                            attributeDescriptor.TypeName,
+                            _modelExpressionTypeName,
+                            StringComparison.Ordinal))
+                    {
+                        errorSink.OnError(SourceLocation.Undefined, Resources.FormatMvcRazorParser_InvalidPropertyType(
+                            descriptor.TypeName,
+                            attributeDescriptor.Name,
+                            attributeDescriptor.TypeName,
```
dougbu: Will remove this parameter from the resource. This value passed here is no longer the property type and, though it would be easy to find in the `Attributes` collection today, the corresponding dictionary descriptor won't always be available in the future i.e. once we support get-only dictionary properties. Rather than guess `IDictionary<string, ModelExpression>`, I'll reword things now.

src/Microsoft.AspNet.Mvc.TagHelpers/AnchorTagHelper.cs
```diff
@@ -102,15 +108,16 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
...
+                            RouteValuesPrefix,
                             Href));
                 }
             }
             else
             {
-                TagBuilder tagBuilder;
-                var routeValues = GetRouteValues(output, routePrefixedAttributes);
+                // Convert from Dictionary<string, string> to Dictionary<string, object>.
+                var routeValues = RouteValues.ToDictionary(kvp => kvp.Key, kvp => (object)kvp.Value);
```
dougbu: Likely irrelevant because `UrlHelper` creates its own dictionary but I'll add `StringComparer.OrdinalIgnoreCase` here and in the `<form/>` tag helper.

dougbu:
Updated but FYI only. Will merge as soon as I've rebased & rebuilt Razor change and tested this again.

dougbu:
517c013

Issue #2446, opened 2015-04-24
Model state has spurious entries when binding to a collection where element type has no validators.


harshgMSFT:
Consider the following scenario 
```c#
public Action([FromHeader(Name = "Header")]string[] headers)
{
}
```
Input
Header: a, b, c
In this case there should be the following keys in the model state
`Header[0]`
`Header[1]`
`Header[2]`

but there a spurious entry for 
`Header` as well which should not be there. 
This would be the case even for simple scenarios like so

```c#
public Action(string[] parameter)
{
}
```
Query string is 
parameter[0]=foo

in this case there would be two entries
`parameter[0]`
and 
`parameter`

Eilon:
The model state entries should represent the structure of the posted data as opposed to how they were bound.

Eilon:
Also need to check what exactly happens today in MVC5 / MVC6 for the scenarios listed above, plus this one:

```c#
public ActionResult Foo(Person[] people)
{
}

public class Person : IValidatableObject
{
}
```

harshgMSFT:
MVC 5.0

For 
```c#
public ActionResult Foo(Person[] people)
{
}

public class Person : IValidatableObject
{
           public string Name { get; set; }
           public IEnumerable<ValidationResult> Validate(ValidationContext validationContext)
            {
                if (Name ==  "c")
                {
                    yield return new ValidationResult("c is not allowed.");
                }

                yield break;
            }
}
```
if there are three people bound
with query string as  `?people[0].Name=a&people[1].Name=b&people[2].Name=c`
there would be FOUR entries 
`people[0].Name`,`people[1].Name`,`people[2].Name`,`people[2]` - this one has an error and no value.

For the other scenario listed in the bug, MVC 5.0 does not add the key with parameter name. (adds only what is bound).

dougbu:
@harshgMSFT could you please clarify the MVC 5 / 6 differences in these scenarios?

rynowak:
88ac4b94e451441da2a03c9f2f5a3c54b73f45cf

PR #2525, opened 2015-05-08
Restore validation node

### LINKED ISSUE #2294: Invalid dynamic collection binding and validation

harshgMSFT:
Current changes only take care of ensuring that 
a. Tree walk is dependend on ModelValidationNode. In cases like body and pre-existing models ( like TryUpdate/TryValidate), ModelValidationNode tree is built up on demand. 

b. For collections, if there was a custom index it will be handled correctly (i.e fixes #2294).

This PR does not address
a. Extra model state keys #2446.
b. docs / additional tests need to be added. 

This PR does not bring back some aspects of ModelValidationNode like Validated and Validating events. (we can add that if we ever find a user scenario for it). 

@dougbu @rynowak

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingResult.cs
@@ -16,10 +16,25 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         /// <param name="isModelSet">A value that represents if the model has been set by the
         /// <see cref="IModelBinder"/>.</param>
         public ModelBindingResult(object model, string key, bool isModelSet)
+            : this (model, key, isModelSet, validationNode: null)
+        {
+        }
+
+        /// <summary>
+        /// Creates a new <see cref="ModelBindingResult"/>.
+        /// </summary>
+        /// <param name="model">The model which was created by the <see cref="IModelBinder"/>.</param>
+        /// <param name="key">The key using which was used to attempt binding the model.</param>
+        /// <param name="isModelSet">A value that represents if the model has been set by the
+        /// <see cref="IModelBinder"/>.</param>
+        /// <param name="validationNode">A <see cref="ModelValidationNode"/> which captures the validation information.
+        /// </param>
+        public ModelBindingResult(object model, string key, bool isModelSet, ModelValidationNode validationNode)
         {
             Model = model;
             Key = key;
             IsModelSet = isModelSet;
+            ValidationNode = validationNode;
         }
 
         /// <summary>
@@ -47,5 +62,10 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         /// </para>
         /// </summary>
         public bool IsModelSet { get; }
+
+        /// <summary>
+        /// A <see cref="ModelValidationNode"/> associated with the current <see cref="ModelBindingResult"/>.
+        /// </summary>
+        public ModelValidationNode ValidationNode { get; }
     }
 }
--- a/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelValidationNode.cs
+++ b/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelValidationNode.cs
@@ -1,0 +1,76 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.AspNet.Mvc.ModelBinding
+{
+    /// <summary>
+    /// Captures the validation information for a particular model.
+    /// </summary>
+    public class ModelValidationNode
+    {
+        /// <summary>
+        /// Creates a new instance of <see cref="ModelValidationNode"/>.
+        /// </summary>
+        /// <param name="key">The key that will be used by the validation system to find <see cref="ModelState"/>
+        /// entries.</param>
+        /// <param name="modelMetadata">The <see cref="ModelMetadata"/> for the <paramref name="model"/>.</param>
+        /// <param name="model">The model object which is to be validated.</param>
+        public ModelValidationNode([NotNull] string key, [NotNull] ModelMetadata modelMetadata, object model)
+            : this (key, modelMetadata, model, new List<ModelValidationNode>())
+        {
+        }
+
+        /// <summary>
+        /// Creates a new instance of <see cref="ModelValidationNode"/>.
+        /// </summary>
+        /// <param name="key">The key that will be used by the validation system to add
+        /// <see cref="ModelStateDictionary"/> entries.</param>
+        /// <param name="modelMetadata">The <see cref="ModelMetadata"/> for the <paramref name="model"/>.</param>
+        /// <param name="model">The model object which will be validated.</param>
+        /// <param name="childNodes">A collection of child nodes.</param>
+        public ModelValidationNode(
+            [NotNull] string key,
+            [NotNull] ModelMetadata modelMetadata,
+            object model,
+            [NotNull] IList<ModelValidationNode> childNodes)
+        {
+            Key = key;
+            ModelMetadata = modelMetadata;
+            ChildNodes = childNodes;
+            Model = model;
+        }
+
+        /// <summary>
+        /// Gets the key used for adding <see cref="ModelStateDictionary"/> entries.
+        /// </summary>
+        public string Key { get; }
+
+        /// <summary>
+        /// Gets the <see cref="ModelMetadata"/>.
+        /// </summary>
+        public ModelMetadata ModelMetadata { get; }
+
+        /// <summary>
+        /// Gets the model instance which is to be validated.
+        /// </summary>
+        public object Model { get; }
+
+        /// <summary>
+        /// Gets the child nodes.
+        /// </summary>
+        public IList<ModelValidationNode> ChildNodes { get; }
+
+        /// <summary>
+        /// Gets or sets a value that indicates whether all properties of the model should be validated.
+        /// </summary>
+        public bool ValidateAllProperties { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value that indicates whether validation should be suppressed.
+        /// </summary>
+        public bool SuppressValidation { get; set; }
+    }
+}
--- a/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Validation/ModelValidationContext.cs
+++ b/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Validation/ModelValidationContext.cs
@@ -10,23 +10,21 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
         public ModelValidationContext(
             [NotNull] ModelBindingContext bindingContext,
             [NotNull] ModelExplorer modelExplorer)
-            : this(bindingContext.ModelName,
-                   bindingContext.BindingSource,
-                   bindingContext.OperationBindingContext.ValidatorProvider,
-                   bindingContext.ModelState,
-                   modelExplorer)
+            : this(
+                 bindingContext.BindingSource,
+                 bindingContext.OperationBindingContext.ValidatorProvider,
+                 bindingContext.ModelState,
+                 modelExplorer)
         {
         }
 
         public ModelValidationContext(
-            string rootPrefix,
             BindingSource bindingSource,
             [NotNull] IModelValidatorProvider validatorProvider,
             [NotNull] ModelStateDictionary modelState,
             [NotNull] ModelExplorer modelExplorer)
         {
             ModelState = modelState;
-            RootPrefix = rootPrefix;
             ValidatorProvider = validatorProvider;
             ModelExplorer = modelExplorer;
             BindingSource = bindingSource;
@@ -45,7 +43,6 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
             [NotNull] ModelExplorer modelExplorer)
         {
             return new ModelValidationContext(
-                parentContext.RootPrefix,
                 modelExplorer.Metadata.BindingSource,
                 parentContext.ValidatorProvider,
                 parentContext.ModelState,
@@ -55,8 +52,6 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
         public ModelExplorer ModelExplorer { get; }
 
         public ModelStateDictionary ModelState { get; }
-
-        public string RootPrefix { get; set; }
 
         public BindingSource BindingSource { get; set; }
 
--- a/src/Microsoft.AspNet.Mvc.Core/Controller.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Controller.cs
@@ -1336,13 +1336,17 @@ namespace Microsoft.AspNet.Mvc
                 modelName);
 
             var validationContext = new ModelValidationContext(
-                modelName,
                 bindingSource: null,
                 validatorProvider: BindingContext.ValidatorProvider,
                 modelState: ModelState,
                 modelExplorer: modelExplorer);
 
-            ObjectValidator.Validate(validationContext);
+            ObjectValidator.Validate(
+                validationContext,
+                new ModelValidationNode(modelName, modelExplorer.Metadata, model)
+                {
+                    ValidateAllProperties = true
+                });
             return ModelState.IsValid;
         }
 
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/BodyModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/BodyModelBinder.cs
@@ -54,11 +54,21 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
 
                 var isTopLevelObject = bindingContext.ModelMetadata.ContainerType == null;
 
-                // For compatibility with MVC 5.0 for top level object we want to consider an empty key instead of 
+                // For compatibility with MVC 5.0 for top level object we want to consider an empty key instead of
                 // the parameter name/a custom name. In all other cases (like when binding body to a property) we
                 // consider the entire ModelName as a prefix.
                 var modelBindingKey = isTopLevelObject ? string.Empty : bindingContext.ModelName;
-                return new ModelBindingResult(model, key: modelBindingKey, isModelSet: true);
+
+                var validationNode = new ModelValidationNode(modelBindingKey, bindingContext.ModelMetadata, model)
+                {
+                    ValidateAllProperties = true
+                };
+
+                return new ModelBindingResult(
+                    model,
+                    key: modelBindingKey,
+                    isModelSet: true,
+                    validationNode: validationNode);
             }
             catch (Exception ex)
             {
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/ByteArrayModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/ByteArrayModelBinder.cs
@@ -39,7 +39,12 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             try
             {
                 var model = Convert.FromBase64String(value);
-                return new ModelBindingResult(model, bindingContext.ModelName, isModelSet: true);
+
+                // We do not need to set an explict ModelValidationNode since CompositeModelBinder does that automatically.
+                return new ModelBindingResult(
+                    model,
+                    bindingContext.ModelName,
+                    isModelSet: true);
             }
             catch (Exception ex)
             {
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
@@ -31,16 +31,19 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             var valueProviderResult = await bindingContext.ValueProvider.GetValueAsync(bindingContext.ModelName);
 
             IEnumerable<TElement> boundCollection;
+            CollectionResult result;
             if (valueProviderResult == null)
             {
-                boundCollection = await BindComplexCollection(bindingContext);
+                result = await BindComplexCollection(bindingContext);
+                boundCollection = result.Model;
             }
             else
             {
-                boundCollection = await BindSimpleCollection(
+                result = await BindSimpleCollection(
                     bindingContext,
                     valueProviderResult.RawValue,
                     valueProviderResult.Culture);
+                boundCollection = result.Model;
             }
 
             var model = bindingContext.Model;
@@ -54,12 +57,16 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
                 CopyToModel(model, boundCollection);
             }
 
-            return new ModelBindingResult(model, bindingContext.ModelName, isModelSet: true);
+            return new ModelBindingResult(
+                model,
+                bindingContext.ModelName,
+                isModelSet: true,
+                validationNode: result?.ValidationNode);
         }
 
         // Used when the ValueProvider contains the collection to be bound as a single element, e.g. the raw value
         // is [ "1", "2" ] and needs to be converted to an int[].
-        internal async Task<IEnumerable<TElement>> BindSimpleCollection(
+        internal async Task<CollectionResult> BindSimpleCollection(
             ModelBindingContext bindingContext,
             object rawValue,
             CultureInfo culture)
@@ -74,6 +81,10 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             var metadataProvider = bindingContext.OperationBindingContext.MetadataProvider;
             var elementMetadata = metadataProvider.GetMetadataForType(typeof(TElement));
 
+            var validationNode = new ModelValidationNode(
+                bindingContext.ModelName,
+                bindingContext.ModelMetadata,
+                boundCollection);
             var rawValueArray = RawValueToObjectArray(rawValue);
             foreach (var rawValueElement in rawValueArray)
             {
@@ -91,18 +102,26 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
                 object boundValue = null;
                 var result =
                     await bindingContext.OperationBindingContext.ModelBinder.BindModelAsync(innerBindingContext);
-                if (result != null)
+                if (result != null && result.IsModelSet)
                 {
                     boundValue = result.Model;
+                    if (result.ValidationNode != null)
+                    {
+                        validationNode.ChildNodes.Add(result.ValidationNode);
+                    }
                 }
                 boundCollection.Add(ModelBindingHelper.CastOrDefault<TElement>(boundValue));
             }
 
-            return boundCollection;
+            return new CollectionResult
+            {
+                ValidationNode = validationNode,
+                Model = boundCollection
+            };
         }
 
         // Used when the ValueProvider contains the collection to be bound as multiple elements, e.g. foo[0], foo[1].
-        private async Task<IEnumerable<TElement>> BindComplexCollection(ModelBindingContext bindingContext)
+        private async Task<CollectionResult> BindComplexCollection(ModelBindingContext bindingContext)
         {
             var indexPropertyName = ModelNames.CreatePropertyModelName(bindingContext.ModelName, "index");
             var valueProviderResultIndex = await bindingContext.ValueProvider.GetValueAsync(indexPropertyName);
@@ -111,7 +130,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             return await BindComplexCollectionFromIndexes(bindingContext, indexNames);
         }
 
-        internal async Task<IEnumerable<TElement>> BindComplexCollectionFromIndexes(
+        internal async Task<CollectionResult> BindComplexCollectionFromIndexes(
             ModelBindingContext bindingContext,
             IEnumerable<string> indexNames)
         {
@@ -131,6 +150,10 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             var elementMetadata = metadataProvider.GetMetadataForType(typeof(TElement));
 
             var boundCollection = new List<TElement>();
+            var validationNode = new ModelValidationNode(
+                bindingContext.ModelName,
+                bindingContext.ModelMetadata,
+                boundCollection);
             foreach (var indexName in indexNames)
             {
                 var fullChildName = ModelNames.CreateIndexModelName(bindingContext.ModelName, indexName);
@@ -146,10 +169,14 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
 
                 var result =
                     await bindingContext.OperationBindingContext.ModelBinder.BindModelAsync(childBindingContext);
-                if (result != null)
+                if (result != null && result.IsModelSet)
                 {
                     didBind = true;
                     boundValue = result.Model;
+                    if (result.ValidationNode != null)
+                    {
+                        validationNode.ChildNodes.Add(result.ValidationNode);
+                    }
                 }
 
                 // infinite size collection stops on first bind failure
@@ -161,7 +188,18 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
                 boundCollection.Add(ModelBindingHelper.CastOrDefault<TElement>(boundValue));
             }
 
-            return boundCollection;
+            return new CollectionResult
+            {
+                ValidationNode = validationNode,
+                Model = boundCollection
+            };
+        }
+
+        internal class CollectionResult
+        {
+            public ModelValidationNode ValidationNode { get; set; }
+
+            public IEnumerable<TElement> Model { get; set; }
         }
 
         /// <summary>
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/CompositeModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/CompositeModelBinder.cs
@@ -55,6 +55,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             bindingContext.OperationBindingContext.BodyBindingState =
                 newBindingContext.OperationBindingContext.BodyBindingState;
 
+            var bindingKey = bindingContext.ModelName;
             if (modelBindingResult.IsModelSet)
             {
                 // Update the model state key if we are bound using an empty prefix and it is a complex type.
@@ -78,15 +79,28 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
                     // In this case, for the model parameter the key would be SimpleType instead of model.SimpleType.
                     // (i.e here the prefix for the model key is empty).
                     // For the id parameter the key would be id.
-                    return modelBindingResult;
+                    bindingKey = string.Empty;
                 }
             }
 
-            // Fall through to update the ModelBindingResult's key.
+            // Update the model validation node if the model binding result was set but no validation node was provided.
+            // This would typically be the case where leaf level model binders, do not have to add a validation node
+            // for validation to take effect. The composite being the entry point for model binders, takes care or
+            // adding missing validation nodes.
+            var modelValidationNode = modelBindingResult.ValidationNode;
+            if (modelBindingResult.IsModelSet && modelValidationNode == null)
+            {
+                modelValidationNode = new ModelValidationNode(
+                    bindingKey,
+                    bindingContext.ModelMetadata,
+                    modelBindingResult.Model);
+            }
+
             return new ModelBindingResult(
                 modelBindingResult.Model,
-                bindingContext.ModelName,
-                modelBindingResult.IsModelSet);
+                bindingKey,
+                modelBindingResult.IsModelSet,
+                modelValidationNode);
         }
 
         private async Task<ModelBindingResult> TryBind(ModelBindingContext bindingContext)
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/KeyValuePairModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/KeyValuePairModelBinder.cs
@@ -15,8 +15,9 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
                                                       typeof(KeyValuePair<TKey, TValue>),
                                                       allowNullModel: true);
 
-            var keyResult = await TryBindStrongModel<TKey>(bindingContext, "Key");
-            var valueResult = await TryBindStrongModel<TValue>(bindingContext, "Value");
+            var childNodes = new List<ModelValidationNode>();
+            var keyResult = await TryBindStrongModel<TKey>(bindingContext, "Key", childNodes);
+            var valueResult = await TryBindStrongModel<TValue>(bindingContext, "Value", childNodes);
 
             if (keyResult.IsModelSet && valueResult.IsModelSet)
             {
@@ -24,8 +25,20 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
                     ModelBindingHelper.CastOrDefault<TKey>(keyResult.Model),
                     ModelBindingHelper.CastOrDefault<TValue>(valueResult.Model));
 
+                // Update the model for the top level validation node.
+                var modelValidationNode =
+                    new ModelValidationNode(
+                        bindingContext.ModelName,
+                        bindingContext.ModelMetadata,
+                        model,
+                        childNodes);
+
                 // Success
-                return new ModelBindingResult(model, bindingContext.ModelName, isModelSet: true);
+                return new ModelBindingResult(
+                    model,
+                    bindingContext.ModelName,
+                    isModelSet: true,
+                    validationNode: modelValidationNode);
             }
             else if (!keyResult.IsModelSet && valueResult.IsModelSet)
             {
@@ -55,8 +68,10 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             }
         }
 
-        internal async Task<ModelBindingResult> TryBindStrongModel<TModel>(ModelBindingContext parentBindingContext,
-                                                                          string propertyName)
+        internal async Task<ModelBindingResult> TryBindStrongModel<TModel>(
+            ModelBindingContext parentBindingContext,
+            string propertyName,
+            List<ModelValidationNode> childNodes)
         {
             var propertyModelMetadata =
                 parentBindingContext.OperationBindingContext.MetadataProvider.GetMetadataForType(typeof(TModel));
@@ -72,6 +87,11 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
                 propertyBindingContext);
             if (modelBindingResult != null)
             {
+                if (modelBindingResult.ValidationNode != null)
+                {
+                    childNodes.Add(modelBindingResult.ValidationNode);
+                }
+
                 return modelBindingResult;
             }
 
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
@@ -44,12 +44,18 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             EnsureModel(bindingContext);
             var result = await CreateAndPopulateDto(bindingContext, mutableObjectBinderContext.PropertyMetadata);
 
+            var validationNode = new ModelValidationNode(
+                bindingContext.ModelName,
+                bindingContext.ModelMetadata,
+                bindingContext.Model);
+
             // post-processing, e.g. property setters and hooking up validation
-            ProcessDto(bindingContext, (ComplexModelDto)result.Model);
+            ProcessDto(bindingContext, (ComplexModelDto)result.Model, validationNode);
             return new ModelBindingResult(
                 bindingContext.Model,
                 bindingContext.ModelName,
-                isModelSet: true);
+                isModelSet: true,
+                validationNode: validationNode);
         }
 
         /// <summary>
@@ -359,11 +365,14 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             return validationInfo;
         }
 
-        internal void ProcessDto(ModelBindingContext bindingContext, ComplexModelDto dto)
+        // Internal for testing.
+        internal ModelValidationNode ProcessDto(
+            ModelBindingContext bindingContext,
+            ComplexModelDto dto,
+            ModelValidationNode validationNode)
         {
             var metadataProvider = bindingContext.OperationBindingContext.MetadataProvider;
             var modelExplorer = metadataProvider.GetModelExplorerForType(bindingContext.ModelType, bindingContext.Model);
-
             var validationInfo = GetPropertyValidationInfo(bindingContext);
 
             // Eliminate provided properties from requiredProperties; leaving just *missing* required properties.
@@ -415,8 +424,20 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
                         out requiredValidator);
 
                     SetProperty(bindingContext, modelExplorer, propertyMetadata, dtoResult, requiredValidator);
+
+                    var dtoValidationNode = dtoResult.ValidationNode;
+                    if (dtoValidationNode == null)
+                    {
+                        // Make sure that irrespective of if the properties of the model were bound with a value,
+                        // create a validation node so that these get validated.
+                        dtoValidationNode = new ModelValidationNode(dtoResult.Key, entry.Key, dtoResult.Model);
+                    }
+
+                    validationNode.ChildNodes.Add(dtoValidationNode);
                 }
             }
+
+            return validationNode;
         }
 
         /// <summary>
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/ServicesModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/ServicesModelBinder.cs
@@ -26,7 +26,17 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         {
             var requestServices = bindingContext.OperationBindingContext.HttpContext.RequestServices;
             var model = requestServices.GetRequiredService(bindingContext.ModelType);
-            return Task.FromResult(new ModelBindingResult(model, bindingContext.ModelName, isModelSet: true));
+            var validationNode =
+                new ModelValidationNode(bindingContext.ModelName, bindingContext.ModelMetadata, model)
+                {
+                    SuppressValidation = true
+                };
+
+            return Task.FromResult(new ModelBindingResult(
+                    model,
+                    bindingContext.ModelName,
+                    isModelSet: true,
+                    validationNode: validationNode));
         }
     }
 }
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/TypeMatchModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/TypeMatchModelBinder.cs
@@ -19,7 +19,12 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             bindingContext.ModelState.SetModelValue(bindingContext.ModelName, valueProviderResult);
             var model = valueProviderResult.RawValue;
             ModelBindingHelper.ReplaceEmptyStringWithNull(bindingContext.ModelMetadata, ref model);
-            return new ModelBindingResult(model, bindingContext.ModelName, isModelSet: true);
+
+            // We do not need to set an explict ModelValidationNode since CompositeModelBinder does that automatically.
+            return new ModelBindingResult(
+                model,
+                bindingContext.ModelName,
+                isModelSet: true);
         }
 
         internal static async Task<ValueProviderResult> GetCompatibleValueProviderResult(ModelBindingContext context)
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
@@ -35,16 +35,19 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
         }
 
         /// <inheritdoc />
-        public void Validate([NotNull] ModelValidationContext modelValidationContext)
+        public void Validate(
+            [NotNull] ModelValidationContext modelValidationContext,
+            [NotNull] ModelValidationNode validationNode)
         {
             var validationContext = new ValidationContext()
             {
                 ModelValidationContext = modelValidationContext,
                 Visited = new HashSet<object>(ReferenceEqualityComparer.Instance),
+                ValidationNode = validationNode
             };
 
             ValidateNonVisitedNodeAndChildren(
-                modelValidationContext.RootPrefix,
+                validationNode.Key,
                 validationContext,
                 validators: null);
         }
@@ -54,19 +57,16 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
             ValidationContext validationContext,
             IList<IModelValidator> validators)
         {
-            var modelValidationContext = validationContext.ModelValidationContext;
-            var modelExplorer = modelValidationContext.ModelExplorer;
-
             // Recursion guard to avoid stack overflows
             RuntimeHelpers.EnsureSufficientExecutionStack();
 
+            var modelValidationContext = validationContext.ModelValidationContext;
+            var modelExplorer = modelValidationContext.ModelExplorer;
             var modelState = modelValidationContext.ModelState;
-
-            var bindingSource = modelValidationContext.BindingSource;
-            if (bindingSource != null && !bindingSource.IsFromRequest)
+            var currentValidationNode = validationContext.ValidationNode;
+            if (currentValidationNode.SuppressValidation)
             {
-                // Short circuit if the metadata represents something that was not bound using request data.
-                // For example model bound using [FromServices]. Treat such objects as skipped.
+                // Short circuit if the node is marked to be suppressed
                 var validationState = modelState.GetFieldValidationState(modelKey);
                 if (validationState == ModelValidationState.Unvalidated)
                 {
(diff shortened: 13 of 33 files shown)


rynowak:
> This PR does not bring back some aspects of ModelValidationNode like Validated and Validating events. (we can add that if we ever find a user scenario for it).

:+1: for leaving this out. Eventing in a system like this is clumsy. It would be better solved another way

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelValidationNode.cs
```diff
@@ -0,0 +1,38 @@
...
+            Model = model;
+        }
+
+        public string Key { get; set; }
+
+        public ModelMetadata ModelMetadata { get; set; }
+
+        public object Model { get; set; }
+
+        public bool BuildChildNodesUsingModel { get; set; } = false;
```
rynowak: I think this was called `ValidateModelProperties` before? It seemed clearer with the old naming.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelValidationNode.cs
```diff
@@ -0,0 +1,38 @@
...
+            ModelMetadata = modelMetadata;
+            ChildNodes = childNodes;
+            Model = model;
+        }
+
+        public string Key { get; set; }
+
+        public ModelMetadata ModelMetadata { get; set; }
+
+        public object Model { get; set; }
```
rynowak: Why are these mutable but required by the constructor?

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelValidationNode.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        public bool BuildChildNodesUsingModel { get; set; } = false;
+
+        public IList<ModelValidationNode> ChildNodes { get; set; }
```
rynowak: Consider just making this a get-only property. Not sure the 2 constructors are needed

src/Microsoft.AspNet.Mvc.Core/Controller.cs
```diff
@@ -1311,7 +1311,12 @@ public virtual bool TryValidateModel([NotNull] object model, string prefix)
                 modelState: ModelState,
                 modelExplorer: modelExplorer);
 
-            ObjectValidator.Validate(validationContext);
+            ObjectValidator.Validate(
+                validationContext,
+                new ModelValidationNode(modelName, modelExplorer.Metadata, model)
+                {
+                    BuildChildNodesUsingModel = true
+                });
```
rynowak: We shouldn't need a `ModelExplorer` in this code anymore.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
```diff
@@ -161,7 +188,18 @@ public class CollectionModelBinder<TElement> : IModelBinder
...
+                ValidationNode = validationNode,
+                Model = boundCollection
+            };
+        }
+
+        internal class CollectionResult
+        {
+            public ModelValidationNode ValidationNode { get; set; }
+
+            public IEnumerable<TElement> Model { get; set; }
```
rynowak: yay for internal DTOs

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
```diff
@@ -31,16 +31,19 @@ public class CollectionModelBinder<TElement> : IModelBinder
...
+                boundCollection = result.Model;
             }
             else
             {
-                boundCollection = await BindSimpleCollection(
+                result = await BindSimpleCollection(
                     bindingContext,
                     valueProviderResult.RawValue,
                     valueProviderResult.Culture);
+                boundCollection = result.Model;
```
rynowak: Should we just have these methods return `ModelBindingResult` directly?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
```diff
@@ -74,6 +81,10 @@ public class CollectionModelBinder<TElement> : IModelBinder
             var metadataProvider = bindingContext.OperationBindingContext.MetadataProvider;
             var elementMetadata = metadataProvider.GetMetadataForType(typeof(TElement));
 
+            var validationNode = new ModelValidationNode(
+                bindingContext.ModelName,
+                bindingContext.ModelMetadata,
+                boundCollection); // This model should represent the ICollection<T> object.
```
rynowak: > represent the ICollection<T> object

but it **is** the `ICollection<T> object`

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
```diff
@@ -374,11 +375,16 @@ internal static PropertyValidationInfo GetPropertyValidationInfo(ModelBindingCon
             return validationInfo;
         }
 
-        internal void ProcessDto(ModelBindingContext bindingContext, ComplexModelDto dto)
+        internal ModelValidationNode ProcessDto(ModelBindingContext bindingContext, ComplexModelDto dto)
```
rynowak: It's a little strange to have `ProcessDTO` return the validation node. Maybe just return the MBR here?

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelValidationNode.cs
harshgMSFT: This was not there before, this indicates if the ModelValidationNode needs to be lazily expanded using the model - i.e this would be true for all models which are pre-existing/or get created in one go (like from body). Earlier there were separate systems for validating based on validation node and body.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelValidationNode.cs
harshgMSFT: yup you are right all constructor initialized properties should be readonly.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelValidationNode.cs
harshgMSFT: :+1:

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelValidationNode.cs
rynowak: https://msdn.microsoft.com/en-us/library/system.web.http.validation.modelvalidationnode.validateallproperties(v=vs.118).aspx

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
harshgMSFT: yes the comment was supposed to be removed. will remove the comment.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
harshgMSFT: >Maybe just return the MBR here?

Except for the fact that there is no model binding happening here :smile: .
An alternative is I can pass in a Model ValidationNode as a parameter, if that makes it clearer.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
```diff
@@ -111,24 +115,14 @@ public void Validate([NotNull] ModelValidationContext modelValidationContext)
             }
 
             // Check to avoid infinite recursion. This can happen with cycles in an object graph.
+            // Note that this is only applicble in case the model is pre-existing.
```
harshgMSFT: `applicable`

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelValidationNode.cs
harshgMSFT: You are talking about a property which did something different... this does not determine if property validation is skipped or not. Please see DefaultObjectValidator for how this is used.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
```diff
@@ -311,6 +269,55 @@ private bool IsTypeExcludedFromValidation(IList<IExcludeTypeValidationFilter> fi
             return filters.Any(filter => filter.IsTypeExcluded(type));
         }
 
+        private IList<IModelValidator> ExpandValidationNodeAndGetValidators(
+            ValidationContext context, ModelExplorer modelExplorer)
+        {
+            var validationNode = context.ValidationNode;
+            if (validationNode == null || !validationNode.BuildChildNodesUsingModel || validationNode.Model == null)
```
harshgMSFT: there should be another condition here, we should return if there are already child nodes present in the validation node. That would mean that some nodes have been created due to model binding, and we should honor those. (TryUpdateModel etc).

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingResult.cs
```diff
@@ -47,5 +53,10 @@ public ModelBindingResult(object model, string key, bool isModelSet)
         /// </para>
         /// </summary>
         public bool IsModelSet { get; }
+
+        /// <summary>
+        /// A <see cref="ModelValidationNode"/> associated with the current <see cref="ModelBindingResult"/>.
+        /// </summary>
+        public ModelValidationNode ValidationNode { get; set; }
```
dougbu: Could this be get-only?  Set in the constructor already.

src/Microsoft.AspNet.Mvc.Core/Controller.cs
dougbu: @rynowak the `ModelExplorer` instance is used just a couple of lines above when creating the `ModelValidationContext`.  The validation system is all about `ModelExplorer` right now.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/ByteArrayModelBinder.cs
```diff
@@ -39,7 +39,14 @@ public class ByteArrayModelBinder : IModelBinder
...
                 var model = Convert.FromBase64String(value);
-                return new ModelBindingResult(model, bindingContext.ModelName, isModelSet: true);
+                return new ModelBindingResult(
+                    model,
+                    bindingContext.ModelName,
+                    isModelSet: true,
+                    validationNode: new ModelValidationNode(
+                        bindingContext.ModelName,
+                        bindingContext.ModelMetadata,
+                        model));
```
dougbu: Ick.  Create the `ModelValidationNode` separately, not inline.

dougbu: Also why does this model binder need to create its own `ModelValidationNode`?  When we had `ModelValidationNode` in the system before, only `CollectionModelBinder`, `MutableObjectModelBinder` and a few other special cases dealt with `ModelValidationNode` directly.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
```diff
@@ -150,6 +173,10 @@ public class CollectionModelBinder<TElement> : IModelBinder
                 {
                     didBind = true;
                     boundValue = result.Model;
+                    if (result.ValidationNode != null)
+                    {
+                        validationNode.ChildNodes.Add(result.ValidationNode);
+                    }
```
dougbu: - Gap if the lower-level binder doesn't create a `ModelValidationNode` but does return a `ModelBindingResult`.  Earlier code unconditionally created a `ModelValidationNode` for each element that was successfully bound (or perhaps for all elements that seem to exist).
- Loops are also incorrect in that they equate `result != null` with success.  Should check `result.IsModelSet` for each element.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
dougbu: Yay but, as you said above @rynowak, does this particular private DTO add value over using `ModelBindingResult` directly?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
```diff
@@ -430,8 +436,20 @@ internal void ProcessDto(ModelBindingContext bindingContext, ComplexModelDto dto
...
+
+                    var dtoValidationNode = dtoResult.ValidationNode;
+                    if (dtoValidationNode == null)
+                    {
+                        // Make sure that irrespective of if the properties of the model were bound with a value,
+                        // create a validation node so that these get validated.
+                        dtoValidationNode = new ModelValidationNode(dtoResult.Key, entry.Key, dtoResult.Model);
+                    }
+
+                    validationNode.ChildNodes.Add(dtoValidationNode);
```
dougbu: Already have `ModelBindingResult` instances for every bound property. This case should be handled in the same place -- [`ComplexModelDto`](https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/ModelBinding/ComplexModelDto.cs#L26).

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelValidationNode.cs
dougbu: Regardless, use a name that describes what the property means, not the internal details of our validation system.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/TypeMatchModelBinder.cs
```diff
@@ -19,7 +19,14 @@ public sealed class TypeMatchModelBinder : IModelBinder
...
             ModelBindingHelper.ReplaceEmptyStringWithNull(bindingContext.ModelMetadata, ref model);
-            return new ModelBindingResult(model, bindingContext.ModelName, isModelSet: true);
+            return new ModelBindingResult(
+                model,
+                bindingContext.ModelName,
+                isModelSet: true,
+                validationNode: new ModelValidationNode(
+                    bindingContext.ModelName,
+                    bindingContext.ModelMetadata,
+                    model));
```
dougbu: Looks like increased complexity of the simplest model binder is a fallout of this change. Users extending this system to bind individual properties (leaf nodes) should not need to understand `ModelValidationNode` _too_.  Note the concept count here is a regression compared to MVC 6 before `ModelValidationNode` was removed.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
```diff
@@ -34,12 +34,20 @@ public class DefaultObjectValidator : IObjectModelValidator
         }
 
         /// <inheritdoc />
-        public void Validate([NotNull] ModelValidationContext modelValidationContext)
+        public void Validate(
+            [NotNull] ModelValidationContext modelValidationContext,
+            [NotNull] ModelValidationNode validationNode)
         {
+            if (validationNode == null)
```
dougbu: Doesn't make sense together with `[NotNull]`

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
```diff
@@ -55,12 +63,11 @@ public void Validate([NotNull] ModelValidationContext modelValidationContext)
         {
             var modelValidationContext = validationContext.ModelValidationContext;
             var modelExplorer = modelValidationContext.ModelExplorer;
+            var modelState = modelValidationContext.ModelState;
```
dougbu: don't move this further from its first use

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
```diff
@@ -87,22 +94,19 @@ public void Validate([NotNull] ModelValidationContext modelValidationContext)
...
-                var validatorProvider = modelValidationContext.ValidatorProvider;
-                var validatorProviderContext = new ModelValidatorProviderContext(modelExplorer.Metadata);
-                validatorProvider.GetValidators(validatorProviderContext);
-
-                validators = validatorProviderContext.Validators;
+                // instead we just pass them over.
+                validators = GetValidators(modelValidationContext.ValidatorProvider, modelExplorer.Metadata);
             }
 
             // We don't need to recursively traverse the graph for null values
```
dougbu: Where are `null` models now skipped? Comment no longer correct.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
dougbu: huh?  what does a "pre-existing" model mean in this context?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
```diff
@@ -165,32 +159,41 @@ private void MarkPropertiesAsSkipped(string currentModelKey, ModelMetadata metad
             }
         }
 
+        private IList<IModelValidator> GetValidators(IModelValidatorProvider provider, ModelMetadata metadata)
+        {
+            var validatorProvider = provider;
```
dougbu: Parameter name is fine.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
```diff
@@ -165,32 +159,41 @@ private void MarkPropertiesAsSkipped(string currentModelKey, ModelMetadata metad
...
+            var validatorProviderContext = new ModelValidatorProviderContext(metadata);
+            validatorProvider.GetValidators(validatorProviderContext);
+            return validatorProviderContext.Validators;
+        }
+
         private bool ValidateProperties(
```
dougbu: Lousy name since this method now handles all child nodes -- whether elements or properties.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
```diff
@@ -165,32 +159,41 @@ private void MarkPropertiesAsSkipped(string currentModelKey, ModelMetadata metad
...
             string currentModelKey,
             ModelExplorer modelExplorer,
             ValidationContext validationContext)
         {
             var isValid = true;
+            var validators = ExpandValidationNodeAndGetValidators(validationContext, modelExplorer);
 
-            foreach (var property in modelExplorer.Metadata.Properties)
+            foreach (var childNode in validationContext.ValidationNode.ChildNodes.Where(child => child != null))
```
dougbu: When would model binding add a `null` node to `ChildNodes`?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
```diff
@@ -311,6 +269,55 @@ private bool IsTypeExcludedFromValidation(IList<IExcludeTypeValidationFilter> fi
             return filters.Any(filter => filter.IsTypeExcluded(type));
         }
 
+        private IList<IModelValidator> ExpandValidationNodeAndGetValidators(
+            ValidationContext context, ModelExplorer modelExplorer)
```
dougbu: Channeling @rynowak: #YOLO break

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelValidationNode.cs
harshgMSFT: Sure .. do you have some suggestions?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
harshgMSFT: `ModelBindingResult` has more than what is required(viz. properties like `IsModelSet` and the `Key`), moreover the model is loosely typed in MBR, resulting in an unnecessary cast.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
harshgMSFT: ModelBindingResult has more than what is required(viz. properties like IsModelSet and the Key), moreover the model is loosely typed in MBR, resulting in an unnecessary cast.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingResult.cs
harshgMSFT: :+1: can be a get only,

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelValidationNode.cs
dougbu: `ValidateAllProperties` sounds good to me.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
harshgMSFT: >Loops are also incorrect in that they equate result != null with success. Should check result.IsModelSet for each element. 

Yes that seems to be an existing bug, should be fixed. 

> Gap if the lower-level binder doesn't create a ModelValidationNode but does return a ModelBindingResult. Earlier code unconditionally created a ModelValidationNode for each element that was successfully bound (or perhaps for all elements that seem to exist).

There would be a gap if a lower-level model binder does not create a MVN but creates a result (with model set) where it wanted to validate. However with this design PR there are no such model binders, and with this PR the proposal was to have that contract - If you need validation, create the validation node - 

The alternatives to the apporoch are : 
1. Create a MVN at CompositeModelBinder for any model binding result which is missing it(as we discussed):
                The problem with this approach is it breaks where someone directly tries to point to a particular leaf level model binder (for example using TryUpdateModelBinder to directly point to TypeConverterModelBinder).

2. Create a MVN in MBR by default : 
               ModelBindingResult does not capture ModelMetadata today which means we would have to let every model binder pass the `bindingContext.ModelMetadata` when creating the MBR. 

Lets discuss and reach a conclusion.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
harshgMSFT: sorry did not understand your comment.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
harshgMSFT: :+1:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
harshgMSFT: consider the TryUpdateModel scenario. 
```c#
    public class SelfishPerson
    {
        public SelfishPerson MySelf { get { return this; } }
    }
```

If the model is like this and you call TryUpdateModel, we do not want to recurse for the property MySelf.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
harshgMSFT: what is #YOLO ?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
harshgMSFT: how about `ValidateChildNodes`?

harshgMSFT:
@rynowak , @dougbu updated it to not be a design pr anymore.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/IObjectModelValidator.cs
```diff
@@ -13,6 +13,6 @@ public interface IObjectModelValidator
         /// </summary>
         /// <param name="validationContext">The <see cref="ModelValidationContext"/> associated with the current call.
         /// </param>
-        void Validate(ModelValidationContext validationContext);
+        void Validate(ModelValidationContext validationContext, ModelValidationNode validationNode);
```
kirthik: Add param documentation for `validationNode`

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/Validation/DefaultObjectValidatorTests.cs
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-#if DNX451
+//#if DNX451
```
ajaybhargavb: Remove.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/Validation/DefaultObjectValidatorTests.cs
```diff
@@ -751,4 +915,4 @@ private class TestModelValidationContext
         }
     }
 }
-#endif
+//#endif
```
ajaybhargavb: Remove.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingResult.cs
```diff
@@ -16,10 +16,25 @@ public class ModelBindingResult
...
+        }
+
+        /// <summary>
+        /// Creates a new <see cref="ModelBindingResult"/>.
+        /// </summary>
+        /// <param name="model">The model which was created by the <see cref="IModelBinder"/>.</param>
+        /// <param name="key">The key using which was used to attempt binding the model.</param>
+        /// <param name="isModelSet">A value that represents if the model has been set by the
+        /// <see cref="IModelBinder"/>.</param>
+        /// <param name="validationNode">A <see cref="ModelValidationNode"/> which represents the validation
```
rynowak: ugh 'represents' again

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelValidationNode.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        /// <summary>
+        /// Gets or sets a value that indicates whether all properties of the model should be validated.
+        /// </summary>
+        public bool ValidateAllProperties { get; set; } = false;
+
+        /// <summary>
+        /// Gets or sets a value that indicates whether validation should be suppressed.
+        /// </summary>
+        public bool SuppressValidation { get; set; } = false;
```
rynowak: no need for the explicit `false` on either of these since it's the default

src/Microsoft.AspNet.Mvc.Core/ModelBinding/ByteArrayModelBinder.cs
```diff
@@ -39,7 +39,10 @@ public class ByteArrayModelBinder : IModelBinder
             try
             {
                 var model = Convert.FromBase64String(value);
-                return new ModelBindingResult(model, bindingContext.ModelName, isModelSet: true);
+                return new ModelBindingResult(
+                    model,
+                    bindingContext.ModelName,
+                    isModelSet: true);
```
rynowak: Should we be creating a node here? This seems to imply that a `byte[]` property can't have validators.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
rynowak: sure

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
rynowak: FYI Only: (previous API-review feedback) - the complications in this class really come from it have two very different code paths to do the same thing.

I feel like what's here is a good way to build this system based on the existing code, but a thorough refactor here would improve it.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CompositeModelBinder.cs
```diff
@@ -203,5 +209,21 @@ private static BodyBindingState GetBodyBindingState(ModelBindingContext oldBindi
...
+            if (result.IsModelSet && result.ValidationNode == null)
+            {
+                return new ModelValidationNode(
+                    key,
+                    metadata,
+                    result.Model);
+            }
+
+            return result.ValidationNode;
+        }
```
rynowak: IMO it would be much less confusing to have this code **inline** and have a comment explaining under what conditions this might occur. These are kinds of interactions we want to highlight, not hide.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
```diff
@@ -44,12 +44,18 @@ public class MutableObjectModelBinder : IModelBinder
             EnsureModel(bindingContext);
             var result = await CreateAndPopulateDto(bindingContext, mutableObjectBinderContext.PropertyMetadata);
 
+            var validationNode = new ModelValidationNode(
+              bindingContext.ModelName,
+              bindingContext.ModelMetadata,
+              bindingContext.Model);
```
rynowak: indent seems wonky

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
rynowak: I think the suggestion is that because ComplexDTOModelBinder creates the MBR, it should be creating this validation nodes as well. I agree with the suggestion, it's in the spirit of how `ComplexDTOModelBinder` works.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/ServicesModelBinder.cs
```diff
@@ -26,7 +26,17 @@ protected override Task<ModelBindingResult> BindModelCoreAsync([NotNull] ModelBi
         {
             var requestServices = bindingContext.OperationBindingContext.HttpContext.RequestServices;
             var model = requestServices.GetRequiredService(bindingContext.ModelType);
-            return Task.FromResult(new ModelBindingResult(model, bindingContext.ModelName, isModelSet: true));
+            var validationNode =
+                new ModelValidationNode(bindingContext.ModelName, bindingContext.ModelMetadata, model)
+                {
+                    SuppressValidation = true
```
rynowak: One problem with this technique is that it doesn't apply when using TryValidateModel. Do you still check the 'source' in there?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
```diff
@@ -166,32 +152,40 @@ private void MarkPropertiesAsSkipped(string currentModelKey, ModelMetadata metad
...
+            var validators = ExpandValidationNodeAndGetValidators(validationContext, modelExplorer);
 
-            foreach (var property in modelExplorer.Metadata.Properties)
+            foreach (var childNode in validationContext.ValidationNode.ChildNodes)
             {
-                var propertyExplorer = modelExplorer.GetExplorerForProperty(property.PropertyName);
-                var propertyMetadata = propertyExplorer.Metadata;
+                var childModelExplorer = childNode.ModelMetadata.MetadataKind == Metadata.ModelMetadataKind.Type ?
+                    _modelMetadataProvider.GetModelExplorerForType(childNode.ModelMetadata.ModelType, childNode.Model) :
+                    modelExplorer.GetExplorerForProperty(childNode.ModelMetadata.PropertyName);
```
rynowak: This is awkward. We should just delete `ModelExplorer` from this code? It's no longer saving you anything if you're doing this.

this should really just be `new ModelExplorer(...)`

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
```diff
@@ -312,6 +261,57 @@ private bool IsTypeExcludedFromValidation(IList<IExcludeTypeValidationFilter> fi
...
+        {
+            var validationNode = context.ValidationNode;
+            if (validationNode.ChildNodes.Count != 0 ||
+                !validationNode.ValidateAllProperties ||
+                validationNode.Model == null)
+            {
+                return null;
+            }
+
+            var enumerableModel = modelExplorer.Model as IEnumerable;
```
rynowak: This should be using `ModelMetadata.IsCollection`

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
harshgMSFT: IIRC the problem with `new ModelExplorer(...)` was DataAnnotationsValidator needs a container to operate. if we blindly do new here it would break the hierarchy. I agree that we should probably yank out ModelExplorer but that is another PR.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
```diff
@@ -312,6 +261,57 @@ private bool IsTypeExcludedFromValidation(IList<IExcludeTypeValidationFilter> fi
...
+                        ValidateAllProperties = true
+                    };
+
+                    validationNode.ChildNodes.Add(childNode);
+                    index++;
+                }
+
+                return GetValidators(context.ModelValidationContext.ValidatorProvider, elementMetadata);
+            }
+        }
```
rynowak: The contract of this method is really strange, and only exists because of an optimization in WebAPI2 that was done to avoid re-creating the validators for elements of a collection.

My suggestion is to just fork the code path earlier - if a few lines of boilerplate code get duped, that's not the worst thing in the world.

test/Microsoft.AspNet.Mvc.IntegrationTests/CollectionModelBinderIntegrationTest.cs
```diff
@@ -536,5 +539,63 @@ private class Person2
...
+                });
+
+                request.Form = formCollection;
+                request.ContentType = "application/x-www-form-urlencoded";
+            });
+
+            var modelState = new ModelStateDictionary();
+
+            // Act
+            var modelBindingResult = await argumentBinder.BindModelAsync(parameter, modelState, operationContext);
```
rynowak: I'm going to make a suggestion now that you'll hate me for. I deserve it probably.

I think that all of these integration tests should be updated to include verification for the validation nodes. 



Now that you hate me, I suggest you and I split up the work between us and do this after your wave of in-flight PRs is checked in.

rynowak:
:watch: for some small comments, looking pretty close to done from my point of view.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
harshgMSFT: Sure.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/ServicesModelBinder.cs
harshgMSFT: Didnt get you, on the top level model, we would always validate, but lets say a property is marked with FromServies, then yes that property will skip validation.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
harshgMSFT: Well I hate even the existence of a `ComplexDTOModelBinder` now the code to handle a validation node is in 3 places. But :ok: if that makes you two happy.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/ByteArrayModelBinder.cs
harshgMSFT: The CompositeModelBinder adds it. Ill add a comment.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
harshgMSFT: :ok: Not doing anychanges in the current one.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/IObjectModelValidator.cs
harshgMSFT: :+1:

test/Microsoft.AspNet.Mvc.IntegrationTests/CollectionModelBinderIntegrationTest.cs
harshgMSFT: :100: ...

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
harshgMSFT: fixed.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CompositeModelBinder.cs
harshgMSFT: done.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelValidationNode.cs
harshgMSFT: sure.

harshgMSFT:
@rynowak , @kirthik @ajaybhargavb branch updated.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/BodyModelBinder.cs
```diff
@@ -58,7 +58,13 @@ public BodyModelBinder()
...
                 // consider the entire ModelName as a prefix.
                 var modelBindingKey = isTopLevelObject ? string.Empty : bindingContext.ModelName;
-                return new ModelBindingResult(model, key: modelBindingKey, isModelSet: true);
+
+                var validationNode = new ModelValidationNode(modelBindingKey, bindingContext.ModelMetadata, model)
+                {
+                    ValidateAllProperties = true
+                };
+
+                return new ModelBindingResult(model, key: modelBindingKey, isModelSet: true, validationNode: validationNode);
```
rynowak: long line

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
```diff
@@ -312,6 +270,52 @@ private bool IsTypeExcludedFromValidation(IList<IExcludeTypeValidationFilter> fi
             return filters.Any(filter => filter.IsTypeExcluded(type));
         }
 
+        private void ExpandValidationNodeAndGetValidators(
+            ValidationContext context, ModelExplorer modelExplorer)
```
rynowak: #YOLOBreak

rynowak: This should probably just be called `ExpandValidationNode`

harshgMSFT: my bad. Forgot to rename the method.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelValidationNode.cs
```diff
@@ -0,0 +1,75 @@
...
+{
+    /// <summary>
+    /// Captures the validation information for a particular model.
+    /// </summary>
+    public class ModelValidationNode
+    {
+        /// <summary>
+        /// Creates a new instance of <see cref="ModelValidationNode"/>.
+        /// </summary>
+        /// <param name="key">The key that will be used by the validation system to log <see cref="ModelState"/>
```
dougbu: log -> find corresponding (could leave out "corresponding" but "log" is misleading)

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelValidationNode.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// entries.</param>
+        /// <param name="modelMetadata">The <see cref="ModelMetadata"/> for the <paramref name="model"/>.</param>
+        /// <param name="model">The model object which is to be validated.</param>
+        public ModelValidationNode([NotNull] string key, [NotNull] ModelMetadata modelMetadata, object model)
+            : this (key, modelMetadata, model, new List<ModelValidationNode>())
+        {
+        }
+        /// <summary>
```
dougbu: add blank line

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelValidationNode.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// Creates a new instance of <see cref="ModelValidationNode"/>.
+        /// </summary>
+        /// <param name="key">The key that will be used by the validation system to add
+        /// <see cref="ModelStateDictionary"/> entries.</param>
+        /// <param name="modelMetadata">The <see cref="ModelMetadata"/> for the <paramref name="model"/>.</param>
+        /// <param name="model">The model object which will be validated.</param>
+        /// <param name="childNodes">A collection of child nodes.</param>
```
dougbu: Minor because I doubt we have scenarios where callers of this constructor care about the `IList` independently: Should this parameter be an `IEnumerable<ModelValidationNode>` which is _copied_ into a `List` the `ModelValidationNode` always constructs?

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Validation/ModelValidationContext.cs
```diff
@@ -10,8 +10,7 @@ public class ModelValidationContext
         public ModelValidationContext(
             [NotNull] ModelBindingContext bindingContext,
             [NotNull] ModelExplorer modelExplorer)
-            : this(bindingContext.ModelName,
-                   bindingContext.BindingSource,
+            : this(bindingContext.BindingSource,
```
dougbu: nit: first parameter gets its own line too

src/Microsoft.AspNet.Mvc.Core/ModelBinding/ByteArrayModelBinder.cs
```diff
@@ -39,7 +39,12 @@ public class ByteArrayModelBinder : IModelBinder
             try
             {
                 var model = Convert.FromBase64String(value);
-                return new ModelBindingResult(model, bindingContext.ModelName, isModelSet: true);
+
+                // We do not need to set an explict ModelValidationNode since CompositeModelBinder does that automatically.
```
dougbu: Are you going to get rid of this `CompositeModelBinder` behaviour immediately after checking this PR and the two others that depend on it in?  Or is this PR still in flux?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
```diff
@@ -31,16 +31,19 @@ public class CollectionModelBinder<TElement> : IModelBinder
             var valueProviderResult = await bindingContext.ValueProvider.GetValueAsync(bindingContext.ModelName);
 
             IEnumerable<TElement> boundCollection;
+            CollectionResult result = null;
```
dougbu: remove `= null`

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
dougbu: looks good

src/Microsoft.AspNet.Mvc.Core/ModelBinding/KeyValuePairModelBinder.cs
```diff
@@ -15,8 +15,13 @@ public sealed class KeyValuePairModelBinder<TKey, TValue> : IModelBinder
                                                       typeof(KeyValuePair<TKey, TValue>),
                                                       allowNullModel: true);
 
-            var keyResult = await TryBindStrongModel<TKey>(bindingContext, "Key");
-            var valueResult = await TryBindStrongModel<TValue>(bindingContext, "Value");
+            var validationNode = new ModelValidationNode(
+                bindingContext.ModelName,
+                bindingContext.ModelMetadata,
+                model: null);
```
dougbu: node itself is never used; instead create an empty `List<ModelValidationNode>` and pass _that_ to `TryBindStrongModel`

src/Microsoft.AspNet.Mvc.Core/ModelBinding/KeyValuePairModelBinder.cs
```diff
@@ -24,8 +29,20 @@ public sealed class KeyValuePairModelBinder<TKey, TValue> : IModelBinder
                     ModelBindingHelper.CastOrDefault<TKey>(keyResult.Model),
                     ModelBindingHelper.CastOrDefault<TValue>(valueResult.Model));
 
+                // Update the model for the top level validation node.
+                var modelValidationNode =
+                    new ModelValidationNode(
+                        validationNode.Key,
+                        validationNode.ModelMetadata,
```
dougbu: with above change, use `bindingContext` for first two arguments

src/Microsoft.AspNet.Mvc.Core/ModelBinding/KeyValuePairModelBinder.cs
harshgMSFT: sure .. :+1:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/ByteArrayModelBinder.cs
harshgMSFT: >Are you going to get rid of this CompositeModelBinder behaviour immediately after checking this PR and the two others that depend on it.

Yes after whatever is out for PR gets in we can do the refactor..

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
```diff
@@ -374,11 +380,13 @@ internal static PropertyValidationInfo GetPropertyValidationInfo(ModelBindingCon
             return validationInfo;
         }
 
-        internal void ProcessDto(ModelBindingContext bindingContext, ComplexModelDto dto)
+        internal ModelValidationNode ProcessDto(
```
dougbu: while here, could you add an `// Internal for testing` comment for such methods in this class?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
```diff
@@ -430,8 +438,20 @@ internal void ProcessDto(ModelBindingContext bindingContext, ComplexModelDto dto
...
 
                     SetProperty(bindingContext, modelExplorer, propertyMetadata, dtoResult, requiredValidator);
+
+                    var dtoValidationNode = dtoResult.ValidationNode;
+                    if (dtoValidationNode == null)
+                    {
+                        // Make sure that irrespective of if the properties of the model were bound with a value,
+                        // create a validation node so that these get validated.
+                        dtoValidationNode = new ModelValidationNode(dtoResult.Key, entry.Key, dtoResult.Model);
+                    }
```
dougbu: with either the current `CompositeMB` code or the future strategy, why does this block exist?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
```diff
@@ -166,32 +152,49 @@ private void MarkPropertiesAsSkipped(string currentModelKey, ModelMetadata metad
...
 
-            foreach (var property in modelExplorer.Metadata.Properties)
+            IList<IModelValidator> validators = null;
+            if (modelExplorer.Metadata.IsCollectionType)
             {
-                var propertyExplorer = modelExplorer.GetExplorerForProperty(property.PropertyName);
-                var propertyMetadata = propertyExplorer.Metadata;
+                var enumerableModel = (IEnumerable)modelExplorer.Model;
+                var elementType = GetElementType(enumerableModel.GetType());
+                var elementMetadata = _modelMetadataProvider.GetMetadataForType(elementType);
```
dougbu: in what scenario is metadata missing or incorrect?  seems to duplicate `ModelMetadata` determination that should be done when each child node was originally created (in model binder or `ExpandValidationNodeAndGetValidators`)

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
```diff
@@ -312,6 +270,52 @@ private bool IsTypeExcludedFromValidation(IList<IExcludeTypeValidationFilter> fi
...
+                    };
+                    validationNode.ChildNodes.Add(childNode);
+                }
+            }
+            else
+            {
+                var enumerableModel = (IEnumerable)modelExplorer.Model;
+                var elementType = GetElementType(enumerableModel.GetType());
+                var elementMetadata = _modelMetadataProvider.GetMetadataForType(elementType);
+                var index = 0;
```
dougbu: Add a comment that `index` would be incorrect in many scenarios but those scenarios are covered by explicit `ModelValidationNode` creation in model binding.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
```diff
@@ -166,32 +152,49 @@ private void MarkPropertiesAsSkipped(string currentModelKey, ModelMetadata metad
...
-            foreach (var property in modelExplorer.Metadata.Properties)
+            IList<IModelValidator> validators = null;
+            if (modelExplorer.Metadata.IsCollectionType)
             {
-                var propertyExplorer = modelExplorer.GetExplorerForProperty(property.PropertyName);
-                var propertyMetadata = propertyExplorer.Metadata;
+                var enumerableModel = (IEnumerable)modelExplorer.Model;
+                var elementType = GetElementType(enumerableModel.GetType());
+                var elementMetadata = _modelMetadataProvider.GetMetadataForType(elementType);
+                validators = GetValidators(validationContext.ModelValidationContext.ValidatorProvider, elementMetadata);
```
dougbu: why does `ValidateNonVisitedNodeAndChildren` need a separate (explicit) validators collection when child nodes should already have the correct (element) metadata?  do child nodes instead have runtime type-based metadata in this case (and design-time metadata in property case)?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
```diff
@@ -166,32 +152,49 @@ private void MarkPropertiesAsSkipped(string currentModelKey, ModelMetadata metad
...
-                var propertyBindingName = propertyMetadata.BinderModelName ?? propertyMetadata.PropertyName;
-                var childKey = ModelNames.CreatePropertyModelName(currentModelKey, propertyBindingName);
-
                 if (!ValidateNonVisitedNodeAndChildren(
-                    childKey,
-                    propertyValidationContext,
-                    validators: null))
+                        childNode.Key,
+                        propertyValidationContext,
+                        validators: validators))
```
dougbu: don't need to name this parameter anymore

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
harshgMSFT: Its not to do with deign-time vs runtime metadata, the only reason why this exists is you do not want to call the factories to get validators for every element in the collection. That was an optimization existing in web api and i am just carrying it forward.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
dougbu: still creating validation node here.  why?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
harshgMSFT: sure .. isnt it a nit thoug :smiling_imp:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
dougbu: I'm not sure about optimizing here versus caching elsewhere or perhaps finding it just doesn't matter.  suggest we optimize _after_ measuring realistic collection model binding scenarios.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
dougbu: :ok:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
harshgMSFT: NO- I am not removing code which changes the behavior I dont intend to change. At this point if you feel this needs to be measured and simplified, we will track it as a  separate issue/PR. NOT doing as part of this.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
harshgMSFT: because I never moved it to complex dto model binder :) ..

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
harshgMSFT: i think compiler barfed that i was using an uninitialized value . .. this stays as is,

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
harshgMSFT: >in what scenario is metadata missing or incorrect?

It is not missing or incorrect, the propertyMetadata represents IEnumerable<T> we want the metadata for T.

>seems to duplicate ModelMetadata determination that should be done when each child node was originally created (in model binder or ExpandValidationNodeAndGetValidators)

Its there in child node, but that would mean to have this check repeated for every child node. and the code to do that would be slightly complex.
Not fixing it. If we have perf issues .. will fix later.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
harshgMSFT: I think the comment makes it clear. 
-This code block ensures  a validation node for every property of the poco model even where there is no model. The code in `CompositeMB` only adds a validation node if the model is set.
This is required so that we perform validation for all the missing properties of the poco model.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
harshgMSFT: done.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
harshgMSFT: done.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelValidationNode.cs
harshgMSFT: why .. dont see a scenario for it.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
dougbu: try again

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelValidationNode.cs
dougbu: Mostly because our POCO classes tend to initialize and own contained collections.  In other words, just consistency.  But as I said, it's a minor comment.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
harshgMSFT: Tried doing it, but i would prefer not to do it in ComplexModelDTOModelBinder, since the result is immutable, adding a validation node to it is not possible and we would have to create a new result, I dont like it .. makes it ugly. NOT Fixing.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
```diff
@@ -312,6 +270,54 @@ private bool IsTypeExcludedFromValidation(IList<IExcludeTypeValidationFilter> fi
...
+                }
+            }
+            else
+            {
+                var enumerableModel = (IEnumerable)modelExplorer.Model;
+                var elementType = GetElementType(enumerableModel.GetType());
+                var elementMetadata = _modelMetadataProvider.GetMetadataForType(elementType);
+
+                // An integer index is incorrect in scenarios where there is a custom index provided by the user.
+                // However those scenarios are supported by createing a ModelValidationNode with the right keys.
```
dougbu: here's where my :trollface: comment from the commit fits

dougbu:
:shipit:

PR #2532, opened 2015-05-09
Adding relevant unit tests for MvcRazorCodeParser property name trimming

pranavkm:
cc @NTaylorMullen

### DIFF
--- a/test/Microsoft.AspNet.Mvc.Razor.Host.Test/MvcRazorParserTest.cs
+++ b/test/Microsoft.AspNet.Mvc.Razor.Host.Test/MvcRazorParserTest.cs
@@ -177,87 +177,33 @@ namespace Microsoft.AspNet.Mvc.Razor
             }
         }
 
-        public static TheoryData TrimSpacesAndCharsData
-        {
-            get
-            {
-                // input, trimCharacters, expectedOutput
-                return new TheoryData<string, char[], string>
-                {
-                    { "abcd", new char[] {  }, "abcd" },
-                    { "  /.", new char[] { '/', '.' }, string.Empty },
-                    { string.Empty, new char[] {  }, string.Empty },
-                    { " ", new char[] {  }, string.Empty },
-                    { "  ", new char[] {  }, string.Empty },
-                    { "  / ", new char[] { '/' }, string.Empty },
-                    { "  \t ", new char[] { '/' }, string.Empty },
-                    { "   ", new char[] { '/' }, string.Empty },
-                    { "  ", new char[] { '/' }, string.Empty },
-                    { "/", new char[] { '/' }, string.Empty },
-                    { "//", new char[] { '/' }, string.Empty },
-                    { "//  ", new char[] { '/' }, string.Empty },
-                    { "/ ", new char[] { '/' }, string.Empty },
-                    { " a ", new char[] {  }, "a" },
-                    { " a", new char[] {  }, "a" },
-                    { "a ", new char[] {  }, "a" },
-                    { "  a  ", new char[] {  }, "a" },
-                    { " a \n\r", new char[] {  }, "a" },
-                    { "\t\r a ", new char[] {  }, "a" },
-                    { "\ta ", new char[] {  }, "a" },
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNet.Mvc.Razor.Host.Test/MvcRazorParserTest.cs
```diff
@@ -177,87 +177,21 @@ public static TheoryData GlobalImportData
...
+        [InlineData("a bc d", "a bc d")]
+        [InlineData("abc;", "abc")]
+        [InlineData("  abc;", "abc")]
+        [InlineData(";; abc;", ";; abc")]
+        [InlineData("abc;; ;", "abc")]
+        public void RemoveWhitespaceAndTrailingSemicolons_ReturnsExpectedValues(string input, string expectedOutput)
         {
-            // Arrange & Act
-            var output = MvcRazorCodeParser.TrimSpacesAndChars(input, trimCharacters);
+            // Act
```
dougbu: nit: still `// Arrange & Act`

test/Microsoft.AspNet.Mvc.Razor.Host.Test/MvcRazorParserTest.cs
```diff
@@ -177,87 +177,21 @@ public static TheoryData GlobalImportData
...
+        [InlineData("", "")]
+        [InlineData("   ;  ", "")]
+        [InlineData("    ", "")]
+        [InlineData("abcd", "abcd")]
+        [InlineData("abc;d", "abc;d")]
+        [InlineData("a bc d", "a bc d")]
+        [InlineData("abc;", "abc")]
+        [InlineData("  abc;", "abc")]
+        [InlineData(";; abc;", ";; abc")]
+        [InlineData("abc;; ;", "abc")]
```
dougbu: suspect you have an off-by-one error in the product code.  example case would be `[InlineData(";;", "")]`.  method will instead return `";"`.  add case and fix bug.

dougbu: hmm, actually that case will work but `[InlineData("a", "a")]` will not.  method will instead return `string.Empty`.

dougbu:
:watch:

test/Microsoft.AspNet.Mvc.Razor.Host.Test/MvcRazorParserTest.cs
NTaylorMullen: test other whitespace also. tabs, newlines etc.

NTaylorMullen:
:watch:

test/Microsoft.AspNet.Mvc.Razor.Host.Test/MvcRazorParserTest.cs
pranavkm: Tabs sound fine - but we aren't going to have newlines in `@inject` property name

pranavkm:
Updated.

test/Microsoft.AspNet.Mvc.Razor.Host.Test/MvcRazorParserTest.cs
```diff
@@ -177,87 +177,33 @@ public static TheoryData GlobalImportData
...
         [Theory]
-        [MemberData(nameof(TrimSpacesAndCharsData))]
-        public void TrimSpacesAndChars_GeneratesExpectedOutput(
-            string input,
-            char[] trimCharacters,
-            string expectedOutput)
+        [InlineData("", "")]
+        [InlineData("   ;  ", "")]
+        [InlineData("    ", "")]
+        [InlineData("a", "a")]
```
dougbu: Thanks.  Please add the `[InlineData(";;", "")]` case as well.

dougbu:
:shipit: (with one more test case) once @NTaylorMullen is satisfied.

NTaylorMullen:
:shipit:

PR #2543, opened 2015-05-12
Fix #2330 - Reimagine *FormatterContext

### LINKED ISSUE #2330: Can InputFormatterContext be replaced by ModelBindingContext
Today `IInputFormatter` get access to the `ActionContext` and the `ModelType` without getting access to all the properties that `ModelBindingContext` exposes. Could we replace it with `ModelBindingContext` so it's the same programming model when writing binders or formatters? Additionally, if we remove the dependency on `ActionContext`, we could change it so that formatters only rely on the ModelBinding assembly as opposed to Mvc.Core. Might make the layering a bit cleaner.

rynowak:
This change simplifies InputFormatterContext/OutputFormatterContext by
swapping ActionContext for HttpContext.

This change is important especially for InputFormatterContext as it
decouples ModelState and ActionContext - allowing us to fix a
related bug where the _wrong_ ModelState can be passed in for a
TryUpdateModel operation.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Abstractions/InputFormatterContext.cs
+++ b/src/Microsoft.AspNet.Mvc.Abstractions/InputFormatterContext.cs
@@ -2,33 +2,42 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Mvc.ModelBinding;
 using Microsoft.Framework.Internal;
 
 namespace Microsoft.AspNet.Mvc
 {
     /// <summary>
-    /// Represents information used by an input formatter for
-    /// deserializing the request body into an object.
+    /// A context object used by an input formatter for deserializing the request body into an object.
     /// </summary>
     public class InputFormatterContext
     {
         /// <summary>
         /// Creates a new instance of <see cref="InputFormatterContext"/>.
         /// </summary>
-        public InputFormatterContext([NotNull] ActionContext actionContext,
-                                     [NotNull] Type modelType)
+        public InputFormatterContext(
+            [NotNull] HttpContext httpContext,
+            [NotNull] ModelStateDictionary modelState,
+            [NotNull] Type modelType)
         {
-            ActionContext = actionContext;
+            HttpContext = httpContext;
+            ModelState = modelState;
             ModelType = modelType;
         }
 
         /// <summary>
-        /// Action context associated with the current call.
+        /// Gets the <see cref="Http.HttpContext"/> associated with the current operation.
         /// </summary>
-        public ActionContext ActionContext { get; private set; }
+        public HttpContext HttpContext { get; private set; }
 
         /// <summary>
-        /// Represents the expected type of the model represented by the request body.
+        /// Gets the <see cref="ModelStateDictionary"/> associated with the current operation.
+        /// </summary>
+        public ModelStateDictionary ModelState { get; }
+
+        /// <summary>
+        /// Gets the expected <see cref="Type"/> of the model represented by the request body.
         /// </summary>
         public Type ModelType { get; private set; }
     }
--- a/src/Microsoft.AspNet.Mvc.Abstractions/OutputFormatterContext.cs
+++ b/src/Microsoft.AspNet.Mvc.Abstractions/OutputFormatterContext.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Text;
+using Microsoft.AspNet.Http;
 using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNet.Mvc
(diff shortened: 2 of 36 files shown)


src/Microsoft.AspNet.Mvc.Abstractions/OutputFormatterContext.cs
```diff
@@ -24,9 +25,9 @@ public class OutputFormatterContext
         public Type DeclaredType { get; set; }
 
         /// <summary>
-        /// Action context associated with the current call.
+        /// Gets the <see cref="HttpContext"/> context associated with the current operation.
```
pranavkm: > (or sets)

src/Microsoft.AspNet.Mvc.Core/ActionResults/ObjectResult.cs
```diff
@@ -240,17 +239,17 @@ public ObjectResult(object value)
...
 
             // By default we want to ignore considering accept headers for content negotiation when
             // they have a media type like */* in them. Browsers typically have these media types.
             // In these cases we would want the first formatter in the list of output formatters to
             // write the response. This default behavior can be changed through options, so checking here.
-            var options = formatterContext.ActionContext.HttpContext
-                                                        .RequestServices
-                                                        .GetRequiredService<IOptions<MvcOptions>>()
-                                                        .Options;
+            var options = formatterContext.HttpContext
```
pranavkm: Have this use the default VS indentation.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/BodyModelBinder.cs
```diff
@@ -26,14 +25,19 @@ public BodyModelBinder()
...
-        protected async override Task<ModelBindingResult> BindModelCoreAsync([NotNull] ModelBindingContext bindingContext)
+        protected async override Task<ModelBindingResult> BindModelCoreAsync(
+            [NotNull] ModelBindingContext bindingContext)
         {
             var requestServices = bindingContext.OperationBindingContext.HttpContext.RequestServices;
 
-            var actionContext = requestServices.GetRequiredService<IScopedInstance<ActionContext>>().Value;
-            var formatters = requestServices.GetRequiredService<IScopedInstance<ActionBindingContext>>().Value.InputFormatters;
+            var httpContext = bindingContext.OperationBindingContext.HttpContext;
+            var formatters = requestServices
```
pranavkm: Is the next part to figure out how to flow formatters through `ModelBindingContext`?

pranavkm:
:shipit:

src/Microsoft.AspNet.Mvc.Abstractions/InputFormatterContext.cs
```diff
@@ -2,33 +2,42 @@
...
     /// <summary>
-    /// Represents information used by an input formatter for
-    /// deserializing the request body into an object.
+    /// A context object used by an input formatter for deserializing the request body into an object.
     /// </summary>
     public class InputFormatterContext
     {
         /// <summary>
         /// Creates a new instance of <see cref="InputFormatterContext"/>.
         /// </summary>
```
harshgMSFT: missing docs for parameters.

harshgMSFT:
> This change is important especially for InputFormatterContext as it
decouples ModelState and ActionContext - allowing us to fix a
related bug where the wrong ModelState can be passed in for a
TryUpdateModel operation.

Can you point to the bug?
Doesnt this change decouple ModelState and Input/Output FormatterContext and not ModelState and ActionContext?

harshgMSFT:
:shipit:

src/Microsoft.AspNet.Mvc.Abstractions/OutputFormatterContext.cs
rynowak: sure

src/Microsoft.AspNet.Mvc.Core/ActionResults/ObjectResult.cs
rynowak: sure

src/Microsoft.AspNet.Mvc.Core/ModelBinding/BodyModelBinder.cs
rynowak: yes. Then we can rid of most of the usage of `IScopedInstance<ActionBindingContext>`

src/Microsoft.AspNet.Mvc.Abstractions/InputFormatterContext.cs
rynowak: :+1:

rynowak:
> Can you point to the bug? Doesnt this change decouple ModelState and Input/Output FormatterContext and not ModelState and ActionContext?

Opened #2546

Yeah, you're right. I'll update the description.

PR #2544, opened 2015-05-12
Fix #2528 - Revert formatter behavior for [Required]

### LINKED ISSUE #2528: Remove XML input formatter behavior for RequiredAttribute
Was introduced in the fix for https://github.com/aspnet/Mvc/issues/468.

One of the new goals for model binding is to use Data Annotations attribute as they were intended. As such, the RequiredAttribute just means that value can't be null, as opposed to meaning that the value must be present in the serialized data.

rynowak:
This change removes the support in the DCS formatter to issue an error
message when [Required] is used on a value type.


test/Microsoft.AspNet.Mvc.FunctionalTests/XmlDataContractSerializerInputFormatterTest.cs
```diff
@@ -54,10 +54,8 @@ public class XmlDataContractSerializerInputFormatterTest
                 data);
         }
 
-        // Verifies that even though all the required data is posted to an action, the model
-        // state has errors related to value types's Required attribute validation.
         [Fact]
-        public async Task RequiredDataIsProvided_AndModelIsBound_AndHasRequiredAttributeValidationErrors()
+        public async Task RequiredDataIsProvided_AndModelIsBound_NoValidationErrors()
```
rynowak: Kept these tests as-is except the error messages are gone.

test/Microsoft.AspNet.Mvc.Xml.Test/XmlDataContractSerializerInputFormatterTest.cs
```diff
@@ -513,637 +504,6 @@ public void SetMaxDepth_ThrowsWhenMaxDepthIsBelowOne()
             Assert.Equal(expectedInt, dummyModel.SampleInt);
             Assert.Equal(expectedString, dummyModel.SampleString);
         }
-
-        [Fact]
-        public async Task PostingListOfModels_HasRequiredAttributeValidationErrors()
```
rynowak: These tests are very specific to the feature being removed, so I ripped them out too.

kichalla:
:shipit:

harshgMSFT:
Would you be also handling removal of errors for required attributes on value types. 
Basically `JsonInputFormatter`  has code to add an error for those, we should yank that out as well.
https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/Formatters/JsonInputFormatter.cs#L70-L80

rynowak:
James already removed that for us https://github.com/aspnet/Mvc/commit/de630754bfcd4ba4017a64f74f30f144e1977e96

PR #2549, opened 2015-05-12
fix #2535: Test failure in Microsoft.AspNet.Mvc.Core.Tests

### LINKED ISSUE #2535: Test failure in Microsoft.AspNet.Mvc.Core.Tests
When trying to build Mvc on my machine, I get a test-failure in `Microsoft.AspNet.Mvc.Core.Tests`. By the looks of it, my guess it's a time-zone related bug, but I haven't looked much into what's causing it.

The test results are as following:

image

kirthik:
cc @dougbu, @ajaybhargavb

### DIFF
--- a/test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
+++ b/test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
@@ -172,26 +172,13 @@ namespace Microsoft.AspNet.Mvc
         }
 
         [Fact]
-        public void SaveAndLoad_SimpleTypesCanBeStoredAndLoaded()
+        public void SaveAndLoad_StringCanBeStoredAndLoaded()
         {
             // Arrange
             var testProvider = new SessionStateTempDataProvider();
-            var inputGuid = Guid.NewGuid();
-            var inputDatetimeVal = new DateTime(DateTime.Now.Ticks, DateTimeKind.Local);
-            var inputDictionary = new Dictionary<string, string>
-            {
-                { "Hello", "World" },
-            };
             var input = new Dictionary<string, object>
             {
-                { "string", "value" },
-                { "int", 10 },
-                { "bool", false },
-                { "DateTime", inputDatetimeVal },
-                { "Guid", inputGuid },
-                { "List`string", new List<string> { "one", "two" } },
-                { "Dictionary", inputDictionary },
-                { "EmptyDictionary", new Dictionary<string, int>() }
+                { "string", "value" }
             };
             var context = GetHttpContext(new TestSessionCollection(), true);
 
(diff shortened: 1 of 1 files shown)


test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
```diff
@@ -177,6 +177,7 @@ public void SaveAndLoad_SimpleTypesCanBeStoredAndLoaded()
             // Arrange
             var testProvider = new SessionStateTempDataProvider();
             var inputGuid = Guid.NewGuid();
+            var inputDatetimeVal = new DateTime(DateTime.Now.Ticks, DateTimeKind.Local);
```
pranavkm: Use a fixed `DateTime` instead: `new DateTime(2015, 05, 11, 12, 13, 14);`

pranavkm: Also rename this to `inputDateTime` \ `inputDateTimeValue`

test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
```diff
@@ -206,7 +207,7 @@ public void SaveAndLoad_SimpleTypesCanBeStoredAndLoaded()
             var boolVal = Assert.IsType<bool>(TempData["bool"]);
             Assert.Equal(false, boolVal);
             var datetimeVal = Assert.IsType<DateTime>(TempData["DateTime"]);
-            Assert.Equal(new DateTime().ToString(), datetimeVal.ToString());
+            Assert.Equal(inputDatetimeVal.ToString(), datetimeVal.ToString());
```
pranavkm: Change this to `Assert.Equal(inputDateTime, dateTimeValue);`

test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
rynowak: Yeah this is still basically using `DateTime.Now`

test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
```diff
@@ -186,7 +187,7 @@ public void SaveAndLoad_SimpleTypesCanBeStoredAndLoaded()
                 { "string", "value" },
                 { "int", 10 },
                 { "bool", false },
-                { "DateTime", new DateTime() },
+                { "DateTime", inputDatetimeVal },
                 { "Guid", inputGuid },
                 { "List`string", new List<string> { "one", "two" } },
```
rynowak: This test is really 10 tests. Think about converting to a theory or just breaking it up.

kirthik: @pranavkm, @ajaybhargavb and I synced up. I will split this test into 10 tests.

test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
dougbu: nit @rynowak actually, was using `DateTime.MinValue` before.  (well, the default / can't-be-disabled parameterless `DateTime` constructor).  issue was the time zone, not the chosen time.

test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
dougbu: hmm, does the bug go away if that's the _only_ change we make?

kirthik: Earlier `inputDateTimeVal` `DateKind` was `Unspecified` while `datetimeVal`'s `DateKind` was `Local`. With this change both `DateKind` will be `Local`. So that should fix the bug.

kirthik:
cc @rynowak, @dougbu, @pranavkm

test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
```diff
@@ -201,20 +189,150 @@ public void SaveAndLoad_SimpleTypesCanBeStoredAndLoaded()
...
+            testProvider.SaveTempData(context, input);
+            var TempData = testProvider.LoadTempData(context);
+
+            // Assert
             var intVal = Convert.ToInt32(TempData["int"]);
             Assert.Equal(10, intVal);
+        }
+
+        [Fact]
+        public void SaveAndLoad_BoolCanBeStoredAndLoaded()
```
pranavkm: Can we make this a `Theory` so we know both `true` and `false` work (and it's not returning `default(bool)`?

test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
```diff
@@ -201,20 +189,150 @@ public void SaveAndLoad_SimpleTypesCanBeStoredAndLoaded()
...
-            var datetimeVal = Assert.IsType<DateTime>(TempData["DateTime"]);
-            Assert.Equal(new DateTime().ToString(), datetimeVal.ToString());
+        }
+
+        [Fact]
+        public void SaveAndLoad_DateTimeCanBeStoredAndLoaded()
+        {
+            // Arrange
+            var testProvider = new SessionStateTempDataProvider();
+            var inputDatetime = new DateTime(2010, 12, 12, 0, 0, 0, DateTimeKind.Local);
```
pranavkm: Maybe use a non-zero time so we know it round trips that part correctly?

pranavkm:
Looks good to me. :shipit: once @dougbu  \ @rynowak  sign off.

test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
```diff
@@ -201,20 +189,150 @@ public void SaveAndLoad_SimpleTypesCanBeStoredAndLoaded()
...
+            var context = GetHttpContext(new TestSessionCollection(), true);
+
+            // Act
+            testProvider.SaveTempData(context, input);
+            var TempData = testProvider.LoadTempData(context);
+
+            // Assert
             var list = (IList<string>)TempData["List`string"];
             Assert.Equal(2, list.Count);
-            Assert.Equal("one", list[0]);
```
ajaybhargavb: Why was this line removed? (sorry for commenting on the commit)

test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
kirthik: Will do.

test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
kirthik: Sure, I will do that.

test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
kirthik: Will add it. Removed it by accident.

test/Microsoft.AspNet.Mvc.Core.Test/SessionStateTempDataProviderTest.cs
```diff
@@ -201,20 +189,150 @@ public void SaveAndLoad_SimpleTypesCanBeStoredAndLoaded()
...
+            testProvider.SaveTempData(context, input);
+            var TempData = testProvider.LoadTempData(context);
+
+            // Assert
             var dictionary = Assert.IsType<Dictionary<string, string>>(TempData["Dictionary"]);
             Assert.Equal("World", dictionary["Hello"]);
+        }
+
+        [Fact]
+        public void SaveAndLoad_EmptyDictionaryCanBeStoredAndLoaded()
```
dougbu: change name to `SaveAndLoad_EmptyDictionary_RoundTripsAsNull()` since the original value (the empty dictionary) is _not_ loaded.

dougbu:
:watch:

dougbu:
Have you confirmed this works just as well after messing with the timezone on your machine?  If yes, :shipit: from me.

kirthik:
@dougbu : Yes I changed timezone on my machine and ran the test. It worked.

kirthik:
checked in - da740cd6a8e75895e72be28e367196aa3af9a6d4

Issue #2534, opened 2015-05-09
AuthorizeFilter sets null Identity


leastprivilege:
I think you have a bug here:

https://github.com/aspnet/Mvc/blob/64e726d2b26422c5452475627e4afaba307edec3/src/Microsoft.AspNet.Mvc.Core/Filters/AuthorizeFilter.cs#L35-L47

This will result in a null ClaimsPrincipal.Identity when the user is not authenticated.

You rather want to create an Identity that is not authenticated.

danroth27:
@blowdart Can you take a look at this?

blowdart:
I already pinged @HaoK.

HaoK:
1ea1cc43387850606d0799b930de66dd8ae838a2

vcsjones:
Is the milestone for this still beta5?

danroth27:
Yup, the fix is in beta5.

PR #2552, opened 2015-05-12
AuthorizeFilter should always set a default identity

HaoK:
Fixes https://github.com/aspnet/Mvc/issues/2534

cc @blowdart @danroth27

### DIFF
--- a/test/Microsoft.AspNet.Mvc.Core.Test/Filters/AuthorizeFilterTest.cs
+++ b/test/Microsoft.AspNet.Mvc.Core.Test/Filters/AuthorizeFilterTest.cs
@@ -88,6 +88,20 @@ namespace Microsoft.AspNet.Mvc.Test
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNet.Mvc.Core.Test/Filters/AuthorizeFilterTest.cs
```diff
@@ -88,6 +88,20 @@ public void InvalidUser()
         }
 
         [Fact]
+        public async Task Invoke_AuthSchemesFailShouldSetEmptyPrincipalOnContext()
+        {
+            // Arrange
+            var authorizeFilter = new AuthorizeFilter(new AuthorizationPolicyBuilder("Fails").RequireAuthenticatedUser().Build());
```
kirthik: This crosses 120 character limit.

HaoK:
1ea1cc43387850606d0799b930de66dd8ae838a2

PR #2553, opened 2015-05-12
[Fixes #2545] RedirectToRouteResult doesn't use RouteName property wh…

### LINKED ISSUE #2545: RedirectToRouteResult doesn't use RouteName property when calculating destination URL
The `Controller.RedirectToRoute()` method creates a `RedirectToRouteResult` instance and passes in the `routeName` and `routeValues` parameters to it, which are assigned to the `RedirectToRouteResult.RouteName` and `RedirectToRouteResult.RouteValues` properties, respectively.

However, on [line 48](https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/ActionResults/RedirectToRouteResult.cs#L48) in the `RedirectToRouteResult.ExecuteResult()` method, the `RouteName` property is never used to calculate the destination URL; it is essentially ignored.

I would have expected that `Controller.RedirectToRoute()` method to consider my provided `routeName` when calculating the destination URL.

The workaround for now, is pretty simple: just calculate my own destination URL using the controller's UrlHelper, but since `Controller.RedirectToRoute()` didn't behave as I had expected it to, I'm filing an issue for it.

kichalla:
…en calculating destination URL

@rynowak @Eilon

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ActionResults/RedirectToRouteResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ActionResults/RedirectToRouteResult.cs
@@ -45,7 +45,7 @@ namespace Microsoft.AspNet.Mvc
         {
             var urlHelper = GetUrlHelper(context);
 
-            var destinationUrl = urlHelper.RouteUrl(RouteValues);
+            var destinationUrl = urlHelper.RouteUrl(RouteName, RouteValues);
             if (string.IsNullOrEmpty(destinationUrl))
             {
                 throw new InvalidOperationException(Resources.NoRoutesMatched);
--- a/test/Microsoft.AspNet.Mvc.Core.Test/ActionResults/RedirectToRouteResultTest.cs
+++ b/test/Microsoft.AspNet.Mvc.Core.Test/ActionResults/RedirectToRouteResultTest.cs
@@ -3,7 +3,9 @@
 
 using System;
 using System.Collections.Generic;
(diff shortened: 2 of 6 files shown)


src/Microsoft.AspNet.Mvc.Core/ActionResults/RedirectToRouteResult.cs
```diff
@@ -45,7 +45,7 @@ public override void ExecuteResult([NotNull] ActionContext context)
         {
             var urlHelper = GetUrlHelper(context);
 
-            var destinationUrl = urlHelper.RouteUrl(RouteValues);
+            var destinationUrl = urlHelper.RouteUrl(RouteName, RouteValues);
```
rynowak: Are there unit tests for this class? should they be updated/added?

kichalla: Thought the functional tests should be sufficient but yeah there are unit tests for this class...added a new test now...

rynowak: Functional tests are harder to debug and take more time to write. They are great when we want to capture a user-focused scenario or complex integration point.

rynowak:
:shipit:

kichalla:
Commit: 4bed2e6f2b3e775ee58deb82ac584210c3e41ed6

Issue #2466, opened 2015-04-28
For successfully bound models there should be no entry with empty key in the model state dictionary.


harshgMSFT:
For example 
```c#

public class DocumentContainer
{
       public Document Document { get; set; }
}
public class Document
{
     [FromHeader]
     public string Version { get; set; }
}

public void Action(DocumentContainer container);
```

In this example currently we fall back to empty prefix and hence the keys in the model state dictionary are like so
`""`
`Document`
`Document.Version`

The first key (and arguably the second key) should not be there. 

MVC 5.0 behavior in a similar scenario (where Version was bound from query instead of header), was have just one key
`Document.Version`

Eilon:
Also need to check what happens in MVC 5 if `Document` implements `IValidatableObject`.

harshgMSFT:
In Mvc 5.0 if `Document` implements `IValidatableObject` there are two entries 
`Document`  and `Document.Version`.
The first entry is added only if there is an error reported by the `IValidatableObject`.

rynowak:
88ac4b94e451441da2a03c9f2f5a3c54b73f45cf

PR #2554, opened 2015-05-13
Fixing #2466, #2446.

harshgMSFT:
The assumption is ModelState should have entries if
1. An error is explicitly added by a model binder.
2. There is validation error reported while validating the model.
3. There is value bound by the model binder.

With this change there should be no extra entry other than for the cases mentioned above.

Also enabling the integration test cases.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
@@ -257,7 +257,9 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
                 }
             }
 
-            if (isValid)
+            // Add an entry only if there was an entry which was added by a model binder.
+            // This prevents adding spurious entries.
+            if (modelState.ContainsKey(modelKey) && isValid)
             {
                 validationContext.ModelValidationContext.ModelState.MarkFieldValid(modelKey);
             }
--- a/test/Microsoft.AspNet.Mvc.IntegrationTests/ActionParametersIntegrationTest.cs
+++ b/test/Microsoft.AspNet.Mvc.IntegrationTests/ActionParametersIntegrationTest.cs
@@ -28,20 +28,20 @@ namespace Microsoft.AspNet.Mvc.IntegrationTests
             public List<Address> Address { get; }
         }
 
-        [Fact(Skip = "Extra entries in model state dictionary. #2466")]
+        [Fact]
         public async Task ActionParameter_NonSettableCollectionModel_EmptyPrefix_GetsBound()
         {
             // Arrange
             var argumentBinder = ModelBindingTestHelper.GetArgumentBinder();
             var parameter = new ParameterDescriptor()
             {
-                Name = "Address",
+                Name = "prefix",
                 ParameterType = typeof(Person3)
             };
 
             var operationContext = ModelBindingTestHelper.GetOperationBindingContext(request =>
             {
-                request.QueryString = QueryString.Create("[0].Street", "SomeStreet");
+                request.QueryString = QueryString.Create("Address[0].Street", "SomeStreet");
             });
 
             var modelState = new ModelStateDictionary();
@@ -64,7 +64,7 @@ namespace Microsoft.AspNet.Mvc.IntegrationTests
             Assert.True(modelState.IsValid);
 
             Assert.Equal(1, modelState.Keys.Count);
-            var key = Assert.Single(modelState.Keys, k => k == "[0].Street");
+            var key = Assert.Single(modelState.Keys, k => k == "Address[0].Street");
             Assert.NotNull(modelState[key].Value);
             Assert.Equal("SomeStreet", modelState[key].Value.AttemptedValue);
             Assert.Equal("SomeStreet", modelState[key].Value.RawValue);
@@ -77,19 +77,19 @@ namespace Microsoft.AspNet.Mvc.IntegrationTests
             public CustomReadOnlyCollection<Address> Address { get; set; }
         }
 
-        [Fact(Skip = "Extra entries in model state dictionary. #2466")]
+        [Fact]
         public async Task ActionParameter_ReadOnlyCollectionModel_EmptyPrefix_DoesNotGetBound()
         {
             // Arrange
             var argumentBinder = ModelBindingTestHelper.GetArgumentBinder();
             var parameter = new ParameterDescriptor()
             {
-                Name = "Address",
+                Name = "prefix",
                 ParameterType = typeof(Person6)
             };
             var operationContext = ModelBindingTestHelper.GetOperationBindingContext(request =>
             {
-                request.QueryString = QueryString.Create("[0].Street", "SomeStreet");
+                request.QueryString = QueryString.Create("Address[0].Street", "SomeStreet");
             });
 
             var modelState = new ModelStateDictionary();
@@ -119,20 +119,20 @@ namespace Microsoft.AspNet.Mvc.IntegrationTests
             public Address[] Address { get; set; }
         }
 
-        [Fact(Skip = "Extra entries in model state dictionary. #2466")]
+        [Fact]
         public async Task ActionParameter_SettableArrayModel_EmptyPrefix_GetsBound()
         {
             // Arrange
             var argumentBinder = ModelBindingTestHelper.GetArgumentBinder();
             var parameter = new ParameterDescriptor()
             {
-                Name = "Address",
+                Name = "prefix",
                 ParameterType = typeof(Person4)
             };
 
             var operationContext = ModelBindingTestHelper.GetOperationBindingContext(request =>
             {
-                request.QueryString = QueryString.Create("[0].Street", "SomeStreet");
+                request.QueryString = QueryString.Create("Address[0].Street", "SomeStreet");
             });
 
             var modelState = new ModelStateDictionary();
@@ -147,6 +147,7 @@ namespace Microsoft.AspNet.Mvc.IntegrationTests
             // Model
             Assert.NotNull(modelBindingResult.Model);
             var boundModel = Assert.IsType<Person4>(modelBindingResult.Model);
+            Assert.NotNull(boundModel.Address);
             Assert.Equal(1, boundModel.Address.Count());
             Assert.Equal("SomeStreet", boundModel.Address[0].Street);
 
@@ -154,7 +155,7 @@ namespace Microsoft.AspNet.Mvc.IntegrationTests
             Assert.True(modelState.IsValid);
 
             Assert.Equal(1, modelState.Keys.Count);
-            var key = Assert.Single(modelState.Keys, k => k == "[0].Street");
+            var key = Assert.Single(modelState.Keys, k => k == "Address[0].Street");
             Assert.NotNull(modelState[key].Value);
             Assert.Equal("SomeStreet", modelState[key].Value.AttemptedValue);
             Assert.Equal("SomeStreet", modelState[key].Value.RawValue);
@@ -167,20 +168,20 @@ namespace Microsoft.AspNet.Mvc.IntegrationTests
             public Address[] Address { get; } = new Address[] { };
         }
 
-        [Fact(Skip = "Extra entries in model state dictionary. #2466")]
+        [Fact]
         public async Task ActionParameter_NonSettableArrayModel_EmptyPrefix_DoesNotGetBound()
         {
             // Arrange
             var argumentBinder = ModelBindingTestHelper.GetArgumentBinder();
             var parameter = new ParameterDescriptor()
             {
-                Name = "Address",
+                Name = "prefix",
                 ParameterType = typeof(Person5)
             };
 
             var operationContext = ModelBindingTestHelper.GetOperationBindingContext(request =>
             {
-                request.QueryString = QueryString.Create("[0].Street", "SomeStreet");
+                request.QueryString = QueryString.Create("Address[0].Street", "SomeStreet");
             });
 
             var modelState = new ModelStateDictionary();
@@ -204,7 +205,7 @@ namespace Microsoft.AspNet.Mvc.IntegrationTests
             Assert.Empty(modelState.Keys);
         }
 
-        [Fact(Skip = "Extra entries in model state dictionary. #2466")]
+        [Fact]
         public async Task ActionParameter_NonSettableCollectionModel_WithPrefix_GetsBound()
         {
             // Arrange
@@ -251,7 +252,7 @@ namespace Microsoft.AspNet.Mvc.IntegrationTests
             Assert.Equal(ModelValidationState.Valid, modelState[key].ValidationState);
         }
 
-        [Fact(Skip = "Extra entries in model state dictionary. #2466")]
+        [Fact]
         public async Task ActionParameter_ReadOnlyCollectionModel_WithPrefix_DoesNotGetBound()
         {
             // Arrange
@@ -292,7 +293,7 @@ namespace Microsoft.AspNet.Mvc.IntegrationTests
             Assert.Empty(modelState.Keys);
         }
 
-        [Fact(Skip = "Extra entries in model state dictionary. #2466")]
+        [Fact]
         public async Task ActionParameter_SettableArrayModel_WithPrefix_GetsBound()
         {
             // Arrange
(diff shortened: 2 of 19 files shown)


src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
```diff
@@ -248,9 +248,12 @@ private IList<IModelValidator> GetValidators(IModelValidatorProvider provider, M
...
             }
 
-            if (isValid)
+            if (modelState.ContainsKey(modelKey))
             {
-                validationContext.ModelValidationContext.ModelState.MarkFieldValid(modelKey);
+                if (isValid)
+                {
+                    validationContext.ModelValidationContext.ModelState.MarkFieldValid(modelKey);
+                }
```
rynowak: Is the issue here that `MarkFieldValid` creates another entry if it doesn't exist? It would be good to leave a comment to that effect.

nit: also can combine the two conditions to reduce nesting and avoid the lookup if it's not valid.

test/Microsoft.AspNet.Mvc.IntegrationTests/ActionParametersIntegrationTest.cs
```diff
@@ -28,20 +28,20 @@ public Person3()
...
-        [Fact(Skip = "Extra entries in model state dictionary. #2466")]
+        [Fact]
         public async Task ActionParameter_NonSettableCollectionModel_EmptyPrefix_GetsBound()
         {
             // Arrange
             var argumentBinder = ModelBindingTestHelper.GetArgumentBinder();
             var parameter = new ParameterDescriptor()
             {
-                Name = "Address",
+                Name = "prefix",
```
rynowak: Good catch on this.

test/Microsoft.AspNet.Mvc.IntegrationTests/BinderTypeBasedModelBinderIntegrationTest.cs
```diff
@@ -163,7 +163,7 @@ private class Person2
             Assert.True(modelState.IsValid);
             var key = Assert.Single(modelState.Keys);
             Assert.Equal("CustomParameter", key);
-            Assert.Null(modelState[key].Value); // value is only set if the custom model binder sets it.
+            Assert.NotNull(modelState[key].Value); //Do not validate as its the test model binder which sets the value.
```
rynowak: If there is a key set here can we perform any stronger validation about what's stored? Values/Errors/etc?

rynowak: Edit: This comment in another place makes it clear to me what you meant to say here. Suggest copy-pasting this version.  `// Value is set by test model binder, no need to validate it.`

rynowak: Isn't there still something valid to assert about the validation state of this key though?

test/Microsoft.AspNet.Mvc.IntegrationTests/BinderTypeBasedModelBinderIntegrationTest.cs
```diff
@@ -273,7 +264,22 @@ public Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindingContex
 
                 var address = new Address() { Street = "SomeStreet" };
 
-                return Task.FromResult(new ModelBindingResult(address, bindingContext.ModelName, true));
+                bindingContext.ModelState.SetModelValue(
+                  bindingContext.ModelName + ".Street",
```
rynowak: `ModelNames.GetPropertyName(...)`

test/Microsoft.AspNet.Mvc.IntegrationTests/BinderTypeBasedModelBinderIntegrationTest.cs
```diff
@@ -273,7 +264,22 @@ public Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindingContex
...
+                  new ValueProviderResult(
+                      address.Street,
+                      address.Street,
+                      System.Globalization.CultureInfo.CurrentCulture));
```
rynowak: Add using

test/Microsoft.AspNet.Mvc.IntegrationTests/BinderTypeBasedModelBinderIntegrationTest.cs
```diff
@@ -273,7 +264,22 @@ public Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindingContex
...
+
+                var validationNode = new ModelValidationNode(
+                  bindingContext.ModelName,
```
rynowak: Is this right? seems like it should be `ModelName + ".Street"`

test/Microsoft.AspNet.Mvc.IntegrationTests/ValidationIntegrationTests.cs
```diff
@@ -1000,7 +1000,7 @@ private class Order10
             var model = Assert.IsType<List<Order10>>(modelBindingResult.Model);
             Assert.Empty(model);
 
-            //Assert.Equal(0, modelState.Count); // This fails due to #2446
+            //Assert.Equal(0, modelState.Count);
```
rynowak: ???

rynowak:
This is awesome :heart_eyes: :heart_eyes: :heart_eyes: 

:watch: for small comments

test/Microsoft.AspNet.Mvc.IntegrationTests/BinderTypeBasedModelBinderIntegrationTest.cs
```diff
@@ -282,6 +288,10 @@ private class SuccessModelBinder : IModelBinder
             public Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindingContext)
             {
                 var model = "Success";
+                bindingContext.ModelState.SetModelValue(
+                    bindingContext.ModelName,
+                    new ValueProviderResult(model, model, System.Globalization.CultureInfo.CurrentCulture));
```
dougbu: probably need `[ReplaceCulture]` on every test where value provider's culture matters

test/Microsoft.AspNet.Mvc.IntegrationTests/BodyValidationIntegrationTests.cs
```diff
@@ -235,7 +235,7 @@ private class Address3
             public int Zip { get; set; }
         }
 
-        [Theory]
+        [Theory(Skip = "There should be entries for all model properties which are bound. #2445")]
```
dougbu: Why have these scenarios regressed?

test/Microsoft.AspNet.Mvc.IntegrationTests/FormCollectionModelBindingIntegrationTest.cs
```diff
@@ -147,7 +147,6 @@ private class Address
             // No data is passed.
             var operationContext = ModelBindingTestHelper.GetOperationBindingContext(request =>
             {
-                request.ContentType = "multipart/form-data";
```
dougbu: why is this incorrect for a file upload?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
harshgMSFT: >Is the issue here that MarkFieldValid creates another entry if it doesn't exist? It would be good to leave a comment to that effect.

Yes. Specifically we want to only add an entry if 
a. there is an error.
b. there is a value set by a model binder and that needs to be set to valid, invalid or skipped.

> nit: also can combine the two conditions to reduce nesting and avoid the lookup if it's not valid.

yeah, missed that.

test/Microsoft.AspNet.Mvc.IntegrationTests/FormCollectionModelBindingIntegrationTest.cs
harshgMSFT: Because HttpAbstractions threw when there was no data passed. Wasnt interesting for the test case to succeed.

test/Microsoft.AspNet.Mvc.IntegrationTests/BinderTypeBasedModelBinderIntegrationTest.cs
harshgMSFT: sure will do.

test/Microsoft.AspNet.Mvc.IntegrationTests/BodyValidationIntegrationTests.cs
harshgMSFT: because we no longer add missing entries in the model state. Basically now we add only if something was already added to model state by the model binders. Since in case of body, we do not add anything to the model state, this breaks.

harshgMSFT: Note this should get fixed once we fix 2445.

test/Microsoft.AspNet.Mvc.IntegrationTests/ValidationIntegrationTests.cs
harshgMSFT: I did a blind find replace will take a look if this can be enabled. BTW the current test is still skipped.

test/Microsoft.AspNet.Mvc.IntegrationTests/FormCollectionModelBindingIntegrationTest.cs
dougbu: put another way, multipart isn't wrong as long as the form content is in the main part

test/Microsoft.AspNet.Mvc.IntegrationTests/FormFileModelBindingIntegrationTest.cs
```diff
@@ -144,7 +144,6 @@ private class Address
             // No data is passed.
             var operationContext = ModelBindingTestHelper.GetOperationBindingContext(request =>
             {
-                request.ContentType = "multipart/form-data";
```
dougbu: huh?  all tests in this class should be testing multipart submissions

test/Microsoft.AspNet.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
```diff
@@ -550,6 +551,9 @@ private class Person3
             var operationContext = ModelBindingTestHelper.GetOperationBindingContext(request =>
             {
                 request.QueryString = new QueryString("?parameter.Customer.Name=bill");
+
+                // This is set so that the input formatter does not add an error to model state.
```
dougbu: Confusing comment.  Do you mean this avoids an error unrelated to the test scenarios?

test/Microsoft.AspNet.Mvc.IntegrationTests/ValidationIntegrationTests.cs
dougbu: This looks like the one place you touched where code is commented-out.  Remove the commenting.

test/Microsoft.AspNet.Mvc.IntegrationTests/BodyValidationIntegrationTests.cs
dougbu: :ok:

test/Microsoft.AspNet.Mvc.IntegrationTests/FormCollectionModelBindingIntegrationTest.cs
dougbu: :ok:

dougbu:
:watch:

test/Microsoft.AspNet.Mvc.IntegrationTests/BinderTypeBasedModelBinderIntegrationTest.cs
harshgMSFT: Well this is the validation node for `Address` and not `Address.Street`, the validation node for `Address.Street` gets created automatically due to `ValidateAllProperties` flag set to true.

test/Microsoft.AspNet.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
harshgMSFT: yes. Let me know if you want to modify that comment.

test/Microsoft.AspNet.Mvc.IntegrationTests/BinderTypeBasedModelBinderIntegrationTest.cs
harshgMSFT: done.

test/Microsoft.AspNet.Mvc.IntegrationTests/BinderTypeBasedModelBinderIntegrationTest.cs
harshgMSFT: sure ..added.

test/Microsoft.AspNet.Mvc.IntegrationTests/BinderTypeBasedModelBinderIntegrationTest.cs
harshgMSFT: done.

harshgMSFT:
@dougbu @rynowak updated.

test/Microsoft.AspNet.Mvc.IntegrationTests/BinderTypeBasedModelBinderIntegrationTest.cs
rynowak: Please add using for `System.Globalization.CultureInfo.CurrentCulture` - there might be a few other places as well.

test/Microsoft.AspNet.Mvc.IntegrationTests/BodyValidationIntegrationTests.cs
```diff
@@ -214,5 +214,67 @@ private class Address2
...
+
+            var operationContext = ModelBindingTestHelper.GetOperationBindingContext(
+                request =>
+                {
+                    request.Body = new MemoryStream(Encoding.UTF8.GetBytes(inputText));
+                    request.ContentType = "application/json";
+                });
+            var httpContext = operationContext.HttpContext;
+            var actionContext = httpContext.RequestServices.GetRequiredService<IScopedInstance<ActionContext>>().Value;
+            var modelState = actionContext.ModelState;
```
rynowak: This fix was checked in, you don't need to use `IScopedInstance<ActionContext>` here anymore

test/Microsoft.AspNet.Mvc.IntegrationTests/DictionaryModelBinderIntegrationTest.cs
```diff
@@ -266,7 +266,7 @@ private class Person
             Assert.Equal("10", entry.Value.RawValue);
         }
 
-        [Fact(Skip = "Extra ModelState key because of #2446, IsValue == false because of #2470")]
+        [Fact(Skip = "IsValue == false because of #2470")]
```
rynowak: Do you mean `IsValid`?

test/Microsoft.AspNet.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
```diff
@@ -44,7 +44,7 @@ private class Address1
             public string Street { get; set; }
         }
 
-        [Fact(Skip = "Extra entries in model state #2446, ModelState.Value not set due to #2445.")]
+        [Fact(Skip = ", ModelState.Value not set due to #2445.")]
```
rynowak: extra `, `

rynowak:
:shipit: from me after fixing small comments and making @dougbu happy

test/Microsoft.AspNet.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
dougbu: please

dougbu:
:watch: primarily for response on open "huh?" comment

test/Microsoft.AspNet.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
dougbu: and thank-you

dougbu:
:shipit:

PR #2557, opened 2015-05-13
Pass InputFormatters in OBC

rynowak:
This removes the need to use IScopedInstance<ActionBindingContext> to get
access to the formatters.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/BodyModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/BodyModelBinder.cs
@@ -28,11 +28,8 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         protected async override Task<ModelBindingResult> BindModelCoreAsync(
             [NotNull] ModelBindingContext bindingContext)
         {
-            var requestServices = bindingContext.OperationBindingContext.HttpContext.RequestServices;
-
             var httpContext = bindingContext.OperationBindingContext.HttpContext;
-            var formatters = requestServices
-                .GetRequiredService<IScopedInstance<ActionBindingContext>>().Value.InputFormatters;
+            var formatters = bindingContext.OperationBindingContext.InputFormatters;
 
             var formatterContext = new InputFormatterContext(
                 httpContext, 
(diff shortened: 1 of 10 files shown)


test/Microsoft.AspNet.Mvc.IntegrationTests/BodyValidationIntegrationTests.cs
```diff
@@ -49,11 +49,7 @@ private class Address
                   request.ContentType = "application/json";
               });
 
-            var actionContext = operationContext
-                .HttpContext
-                .RequestServices
-                .GetRequiredService<IScopedInstance<ActionContext>>().Value;
-            var modelState = actionContext.ModelState;
+            var modelState = new ModelStateDictionary();
```
rynowak: No longer needed because BodyModelBinder doesn't use `IScopedInstance` anymore

src/Microsoft.AspNet.Mvc.Core/ModelBinding/BodyModelBinder.cs
```diff
@@ -28,11 +28,8 @@ public BodyModelBinder()
         protected async override Task<ModelBindingResult> BindModelCoreAsync(
             [NotNull] ModelBindingContext bindingContext)
         {
-            var requestServices = bindingContext.OperationBindingContext.HttpContext.RequestServices;
-
             var httpContext = bindingContext.OperationBindingContext.HttpContext;
-            var formatters = requestServices
-                .GetRequiredService<IScopedInstance<ActionBindingContext>>().Value.InputFormatters;
```
harshgMSFT: :heart_eyes:

src/Microsoft.AspNet.Mvc.Core/ParameterBinding/ModelBindingHelper.cs
```diff
@@ -35,6 +35,9 @@ public static class ModelBindingHelper
         /// <param name="metadataProvider">The provider used for reading metadata for the model type.</param>
         /// <param name="modelBinder">The <see cref="IModelBinder"/> used for binding.</param>
         /// <param name="valueProvider">The <see cref="IValueProvider"/> used for looking up values.</param>
+        /// <param name="inputFormatters">
```
harshgMSFT: nit: wouldnt it fit in one line?

harshgMSFT:
:shipit:

src/Microsoft.AspNet.Mvc.Core/ParameterBinding/ModelBindingHelper.cs
rynowak: It was close so I broke it up. It's not hurting anything.

PR #2558, opened 2015-05-14
Fixes #2464 - Does not add extra skipped entries for model bound from…

### LINKED ISSUE #2464: Models not bound from request (like services) should not show up in Modelstate.
consider this 
```c#
public void Action([FromService] Service service)
{
}
```
currently there is an entry in model state like so 
key: service, value : null, validation state: Skipped. 

Ideally there should be no entry since the model is not bound using user data.

harshgMSFT:
… services.

Also ensures that when a type is marked as skipped, any sub property which is model bound (and hence a modelstate un validated entry),
is marked as skipped (otherwise it would cause the ModelState to be invalid).
Also fixing a bug in model state dictionary FindKeyWithPrefix was not considering [0] & [0][0] as a valid prefix.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
+++ b/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
@@ -502,7 +502,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             return GetEnumerator();
         }
 
-        private static IEnumerable<KeyValuePair<string, TValue>> FindKeysWithPrefix<TValue>(
+        public static IEnumerable<KeyValuePair<string, TValue>> FindKeysWithPrefix<TValue>(
             [NotNull] IDictionary<string, TValue> dictionary,
             [NotNull] string prefix)
         {
@@ -523,8 +523,8 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
 
                 if (key.StartsWith("[", StringComparison.OrdinalIgnoreCase))
                 {
-                    key = key.Substring(key.IndexOf('.') + 1);
-                    if (string.Equals(prefix, key, StringComparison.Ordinal))
+                    var subKey = key.Substring(key.IndexOf('.') + 1);
+                    if (string.Equals(prefix, subKey, StringComparison.Ordinal))
                     {
                         yield return entry;
                         continue;
--- a/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
+++ b/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
@@ -281,7 +281,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         /// state errors; <see cref="ModelValidationState.Valid"/> otherwise.</returns>
         public ModelValidationState GetFieldValidationState([NotNull] string key)
         {
-            var entries = FindKeysWithPrefix(this, key);
+            var entries = FindKeysWithPrefix(key);
             if (!entries.Any())
             {
                 return ModelValidationState.Unvalidated;
@@ -378,7 +378,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             // If key is null or empty, clear all entries in the dictionary
             // else just clear the ones that have key as prefix
             var entries  = (string.IsNullOrEmpty(key)) ?
-                _innerDictionary : FindKeysWithPrefix(this, key);
+                _innerDictionary : FindKeysWithPrefix(key);
 
             foreach (var entry in entries)
             {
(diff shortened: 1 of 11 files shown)


dougbu:
What fixes are contained in the `SkippedFixesBase` branch?  Is that the same branch as `CleanupMSD`, `RestoreValidationNode`, or something else?

harshgMSFT:
@dougbu  Yes its the same, but i didn't want to update the other branches otherwise the PRs would have been messed up hence created a temp base branch.

dougbu:
Same as which branch?

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
```diff
@@ -502,7 +502,7 @@ IEnumerator IEnumerable.GetEnumerator()
             return GetEnumerator();
         }
 
-        private static IEnumerable<KeyValuePair<string, TValue>> FindKeysWithPrefix<TValue>(
+        public static IEnumerable<KeyValuePair<string, TValue>> FindKeysWithPrefix<TValue>(
```
rynowak: If we're doing this why not just make it an instance method?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
```diff
@@ -66,12 +66,10 @@ public class DefaultObjectValidator : IObjectModelValidator
...
-                // Short circuit if the node is marked to be suppressed
-                var validationState = modelState.GetFieldValidationState(modelKey);
-                if (validationState == ModelValidationState.Unvalidated)
-                {
-                    modelValidationContext.ModelState.MarkFieldSkipped(modelKey);
-                }
+                // Short circuit if the node is marked to be suppressed.
+                // If there are any sub entries which were model bound, they need to be marked as skipped,
+                // Otherwise they will remain as unvalidated and the model state would be Invalid.
+                MarkChildNodesAsSkipped(modelKey, modelExplorer.Metadata, validationContext);
```
rynowak: Would it be possible to implement this logic using the tree of validation nodes? Just curious.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
harshgMSFT: Sure

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/Validation/DefaultObjectValidatorTests.cs
```diff
@@ -653,16 +663,14 @@ public void DictionaryType_ValidationSuccessful()
 
             // Assert
             Assert.True(validationContext.ModelState.IsValid);
-            var modelState = validationContext.ModelState["items"];
-            Assert.Equal(modelState.ValidationState, ModelValidationState.Valid);
-            modelState = validationContext.ModelState["items[0].Key"];
-            Assert.Equal(modelState.ValidationState, ModelValidationState.Skipped);
+            var modelState = validationContext.ModelState["items[0].Key"];
```
dougbu: Should assert the `Count` at least a few places in this class.  Only the tests using `Assert.Single()` seem to do so now.

test/WebSites/ModelBindingWebSite/Controllers/ValidationController.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Collections.Generic;
+using System.Linq;
```
dougbu: why?

dougbu:
:watch:

test/WebSites/ModelBindingWebSite/Controllers/ValidationController.cs
harshgMSFT: will revert.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
harshgMSFT: Yes, but that is error prone. The current implementation is exactly what is needed.

harshgMSFT:
@rynowak updated. @dougbu The base branch `SkippedFixesBase ` is same as `CleanupMSD` branch and was created so that you can only see the changes which pertain to this PR, If you sign off it will be merged with `CleanupMSD` and if you sign off `CleanupMSD` it will be merged to `RestoreValidationNode` branch. If you sign off `RestoreValidationNode` it will be merged to dev.

rynowak:
:shipit: from me

test/Microsoft.AspNet.Mvc.FunctionalTests/ModelBindingTest.cs
```diff
@@ -76,8 +97,8 @@ public class ModelBindingTest
 
             // Assert
             var stringValue = await response.Content.ReadAsStringAsync();
-            var isModelStateValid = JsonConvert.DeserializeObject<bool>(stringValue);
-            Assert.True(isModelStateValid);
+            var json = JsonConvert.DeserializeObject<ModelStateDictionary>(stringValue);
+            Assert.True(json.IsValid);
```
dougbu: nit: this isn't JSON, it's a `ModelStateDictionary` and should thus be named `modelState`

dougbu:
:shipit: with teensy rename

Issue #2414, opened 2015-04-20
Remove DefaultValue support from model binding and do it only in the invoker for action parameters


rynowak:
`MutableObjectModelBinder` performs reflection to look up properties and their default values.

https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.ModelBinding/Binders/MutableObjectModelBinder.cs#L298

We should move this into `ModelMetadata` so it can be cached, and so that the model binding system is more data driven.

rynowak:
We should also consider just removing this code. 


See: https://msdn.microsoft.com/en-us/library/system.componentmodel.defaultvalueattribute%28v=vs.110%29.aspx

```
A DefaultValueAttribute will not cause a member to be automatically initialized with the attribute's value. You must set the initial value in your code.
```

rynowak:
Legacy info: 

MVC5 and WebAPI2 respected `[DefaultValue(...)]` applied to a parameter of an action method.

*MVC futures* and WebAPI2 respected `[DefaultValue(...)]` for binding model properties.

Eilon:
Remove DefaultValue support from model binding and do it only in the invoker for action parameters.

rynowak:
> Remove DefaultValue support from model binding and do it only in the invoker for action parameters.

Does this include bound-controller-properties? This affects the shape of the implementation.

Is the goal to limit the spread of a feature that we need for compat? Or to be consistent?

Eilon:
Bound controller properties are an interesting case. They're more akin to action parameters, so that makes it more interesting.

Having said that, I'm fine *not* supporting DefaultValue on controller properties for now, and we could add that in the future if there's demand.

rynowak:
The interesting thing about doing it on both properties/parameters is that the implementation concerns would probably dictate that it be first-class at that point -- `ParameterModel`, `ParameterDescriptor`, customizing via convention, etc. 

Seems to me like that's a bit overkill, will just do it at a low-level for now and just for parameters :+1:

Eilon:
Yeah that sounds like the best plan for now. Thanks!

rynowak:
Part 1 checked in: 2fc983b23af5515cd1b4a32c47ea8203be92dddf

rynowak:
Part 2 checked in: cc4ee1068de55a0948c35f5a87f6c05907cb8461

PR #2559, opened 2015-05-14
Fix for #2414 - Remove [DefaultValue] support from ModelBinding

rynowak:
This part of the change removes default value support from ModelBinding.

Updated some unit tests to verify that it does nothing in that case.
Deleted a functional test as it was pure duplication of another
(supported) case where the property has a pre-initialized value.


dougbu:
:shipit:

Eilon:
:shipit:

harshgMSFT:
:shipit:

PR #2561, opened 2015-05-14
React to aspnet/Razor#390

dougbu:
- `TagHelperAttributeDescriptor` constructor has one more parameter


dougbu:
No longer needed since the `public` constructor in `TagHelperAttributeDescriptor` will remain unchanged.

PR #2562, opened 2015-05-14
[Fixes #2541] Use custom stream writer for Xml output formatters

### LINKED ISSUE #2541: Use custom stream writer for Xml output formatters
Follow up to https://github.com/aspnet/Mvc/issues/2496

Currently Xml output formatters ([DCS](https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Xml/XmlDataContractSerializerOutputFormatter.cs#L161) and [XmlSerializer](https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Xml/XmlSerializerOutputFormatter.cs#L136)) do not use the custom stream writer.

XmlWriter however provides an [overload](https://msdn.microsoft.com/en-us/library/ms162620%28v=vs.110%29.aspx) which takes in a text writer. We should be able to provide the custom stream writer that we built as part of https://github.com/aspnet/Mvc/issues/2496

kichalla:
@rynowak @pranavkm

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Xml/XmlDataContractSerializerOutputFormatter.cs
+++ b/src/Microsoft.AspNet.Mvc.Xml/XmlDataContractSerializerOutputFormatter.cs
@@ -158,7 +158,9 @@ namespace Microsoft.AspNet.Mvc.Xml
         public virtual XmlWriter CreateXmlWriter([NotNull] Stream writeStream,
                                                  [NotNull] XmlWriterSettings xmlWriterSettings)
         {
-            return XmlWriter.Create(writeStream, xmlWriterSettings);
+            return XmlWriter.Create(
+                new HttpResponseStreamWriter(writeStream, xmlWriterSettings.Encoding),
+                xmlWriterSettings);
         }
 
         /// <inheritdoc />
@@ -167,10 +169,7 @@ namespace Microsoft.AspNet.Mvc.Xml
             var tempWriterSettings = WriterSettings.Clone();
             tempWriterSettings.Encoding = context.SelectedEncoding;
 
-            var innerStream = context.HttpContext.Response.Body;
-
-            using (var outputStream = new NonDisposableStream(innerStream))
-            using (var xmlWriter = CreateXmlWriter(outputStream, tempWriterSettings))
+            using (var xmlWriter = CreateXmlWriter(context.HttpContext.Response.Body, tempWriterSettings))
             {
                 var obj = context.Object;
                 var runtimeType = obj?.GetType();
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNet.Mvc.Xml/XmlDataContractSerializerOutputFormatter.cs
```diff
@@ -167,10 +169,7 @@ public override Task WriteResponseBodyAsync([NotNull] OutputFormatterContext con
             var tempWriterSettings = WriterSettings.Clone();
             tempWriterSettings.Encoding = context.SelectedEncoding;
 
-            var innerStream = context.HttpContext.Response.Body;
-
-            using (var outputStream = new NonDisposableStream(innerStream))
```
kichalla: The type `NonDisposableStream` is used in input formatters and so we cannot remove it entirely (while investigating this noticed that Xml input formatters use this but json does not...we should fix it)

test/Microsoft.AspNet.Mvc.Xml.Test/XmlSerializerOutputFormatterTest.cs
```diff
@@ -228,7 +228,7 @@ public void DefaultConstructor_ExpectedWriterSettings_Created()
             var body = outputFormatterContext.HttpContext.Response.Body;
             body.Position = 0;
 
-            var content = new StreamReader(body).ReadToEnd();
+            var content = new StreamReader(body, Encoding.Unicode).ReadToEnd();
```
rynowak: Assuming this doesn't cause problems without the BOM? Would be a little more reassuring if it was explicit in terms of endian-ness + no-bom

test/Microsoft.AspNet.Mvc.Xml.Test/XmlDataContractSerializerOutputFormatterTest.cs
```diff
@@ -272,7 +272,7 @@ public void DefaultConstructor_ExpectedWriterSettings_Created()
             var body = outputFormatterContext.HttpContext.Response.Body;
             body.Position = 0;
 
-            var content = new StreamReader(body).ReadToEnd();
+            var content = new StreamReader(body, Encoding.Unicode).ReadToEnd();
```
kichalla: This change is fine because before the usage of custom stream writer, we used to [write out the BOM](https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Common/Encodings.cs#L20) for utf-16 encoding and because of this i believe StreamReader was able to figure out that the content was utf-16 without the requirement of explicitly supplying the encoding...but since the custom stream writer avoids writing totally, the StreamReaders requires supplying the encoding...

test/Microsoft.AspNet.Mvc.Xml.Test/XmlSerializerOutputFormatterTest.cs
kichalla: sure..

kichalla: I have update the PR with the change..thanks... @rynowak

rynowak:
:shipit:

kichalla:
commit 6ef8be92ff7fbae610e3b03a93c399a0c1abae3d

PR #2563, opened 2015-05-14
Fix #2414 Part 2 - Add [DefaultValue(...)] support to the invoker for action parameters

### LINKED ISSUE #2414: Remove DefaultValue support from model binding and do it only in the invoker for action parameters

rynowak:


test/Microsoft.AspNet.Mvc.FunctionalTests/DefaultValuesTest.cs
```diff
@@ -0,0 +1,90 @@
...
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Builder;
+using Microsoft.Framework.DependencyInjection;
+using Xunit;
+
+namespace Microsoft.AspNet.Mvc.FunctionalTests
+{
+    public class DefaultValuesTest
```
rynowak: I tried also writing a test for `ViewComponents` to verify that they **don't** support `[DefaultValue(...)]` but it was impossible. VCs use the runtime types of the arguments to dispatch, and you can't call them with a *missing* argument so it's moot.

src/Microsoft.AspNet.Mvc.Core/ControllerActionExecutor.cs
```diff
@@ -125,9 +126,19 @@ private static Task<object> Convert<T>(object taskAsObject)
                     }
                     else
                     {
-                        value = parameterInfo.ParameterType.GetTypeInfo().IsValueType
-                            ? Activator.CreateInstance(parameterInfo.ParameterType)
-                            : null;
+                        var defaultValueAttribute = 
+                            parameterInfo.GetCustomAttribute<DefaultValueAttribute>(inherit: false);
```
harshgMSFT: mm.. do we want to not see the inherited attributes? Is it because of compat or are we making a call here?

rynowak: From MVC5:
```
 DefaultValueAttribute[] attrs = (DefaultValueAttribute[])parameterInfo.GetCustomAttributes(typeof(DefaultValueAttribute), false);
```

Also, interesting note from MSDN:
> This method ignores the inherit parameter. To search the inheritance chain for attributes on parameters, use the appropriate overloads of the Attribute.GetCustomAttributes method.

test/Microsoft.AspNet.Mvc.Core.Test/ActionExecutorTests.cs
```diff
@@ -194,6 +194,70 @@ public class ActionExecutorTests
...
+                syncMethod.GetMethodInfo(),
+                _controller,
+                new Dictionary<string, object>());
+
+            // Assert
+            Assert.Equal("world", result);
+        }
+
+        [Fact]
+        public async Task ExecuteAsync_WithArgumentDictionary_AnyValue_HasPrecedenceOverDefaults()
```
harshgMSFT: similar test for argument array?

rynowak: That overload **intentionally** doesn't have any of these features.

rynowak:
@harshgMSFT any more feedback on this?

src/Microsoft.AspNet.Mvc.Core/ControllerActionExecutor.cs
harshgMSFT: :+1:

test/Microsoft.AspNet.Mvc.Core.Test/ActionExecutorTests.cs
harshgMSFT: :+1: I was lazy in looking up the implementation my bad.

test/Microsoft.AspNet.Mvc.FunctionalTests/DefaultValuesTest.cs
harshgMSFT: :+1:

harshgMSFT:
:shipit:

PR #2565, opened 2015-05-14
Adding some deserialization to a test for SerializableError

rynowak:


test/Microsoft.AspNet.Mvc.FunctionalTests/JsonOutputFormatterTests.cs
```diff
@@ -72,7 +73,21 @@ public class JsonOutputFormatterTests
 
             // Assert
             Assert.Equal(HttpStatusCode.BadRequest, response.StatusCode);
-            Assert.Equal(expectedOutput, await response.Content.ReadAsStringAsync());
+
+            var actualContent = await response.Content.ReadAsStringAsync();
+            Assert.Equal(expectedOutput, actualContent);
+
+
+            var modelStateErrors = JsonConvert.DeserializeObject<Dictionary<string, string[]>>(actualContent);
```
rynowak: If you try to deserialize into a `SerializableError` directly, you end up with `JArray[]` which seems consistent with what we used to do.

test/Microsoft.AspNet.Mvc.FunctionalTests/JsonOutputFormatterTests.cs
```diff
@@ -72,7 +73,21 @@ public class JsonOutputFormatterTests
 
             // Assert
             Assert.Equal(HttpStatusCode.BadRequest, response.StatusCode);
-            Assert.Equal(expectedOutput, await response.Content.ReadAsStringAsync());
+
+            var actualContent = await response.Content.ReadAsStringAsync();
+            Assert.Equal(expectedOutput, actualContent);
+
+
```
harshgMSFT: extra line.

rynowak:
@harshgMSFT any more hard-hitting feedback?

test/Microsoft.AspNet.Mvc.FunctionalTests/JsonOutputFormatterTests.cs
```diff
@@ -72,7 +73,21 @@ public class JsonOutputFormatterTests
...
+            var modelStateErrors = JsonConvert.DeserializeObject<Dictionary<string, string[]>>(actualContent);
+            Assert.Equal(2, modelStateErrors.Count);
+
+            var errors = Assert.Single(modelStateErrors, kvp => kvp.Key == "Id").Value;
+            Assert.Single(errors);
+            Assert.Equal("The field Id must be between 10 and 100.", errors[0]);
+
+            errors = Assert.Single(modelStateErrors, kvp => kvp.Key == "Name").Value;
+            Assert.Single(errors);
```
harshgMSFT: you could use var `error = Assert.Single`... and then use error instead of `error[0]`.

harshgMSFT:
:shipit:

PR #2566, opened 2015-05-14
Fix 2224: [Jsonpatch]Added tests for custom converters on properties

kirthik:
@rynowak : Could you please take a look?

### DIFF
--- a/test/Microsoft.AspNet.Mvc.FunctionalTests/JsonPatchTest.cs
+++ b/test/Microsoft.AspNet.Mvc.FunctionalTests/JsonPatchTest.cs
@@ -11,6 +11,7 @@ using JsonPatchWebSite.Models;
 using Microsoft.AspNet.Builder;
 using Microsoft.Framework.DependencyInjection;
 using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 using Xunit;
 
 namespace Microsoft.AspNet.Mvc.FunctionalTests
@@ -313,7 +314,7 @@ namespace Microsoft.AspNet.Mvc.FunctionalTests
 
             var input = "[{ \"op\": \"add\", " +
                 "\"path\": \"Customer/Orders/2\", " +
-               "\"value\": { \"OrderName\": \"Name2\" }}]";
+               "\"value\": { \"OrderType\": \"Type2\" }}]";
             var request = new HttpRequestMessage
             {
                 Content = new StringContent(input, Encoding.UTF8, "application/json-patch+json"),
@@ -326,7 +327,8 @@ namespace Microsoft.AspNet.Mvc.FunctionalTests
 
             // Assert
             var body = await response.Content.ReadAsStringAsync();
-            Assert.Contains("OrderTypeSetInConverter", body);
+            dynamic d = JObject.Parse(body);
+            Assert.Equal("OrderTypeSetInConverter", (string)d.Orders[2].OrderType);
         }
 
         [Fact]
@@ -351,7 +353,9 @@ namespace Microsoft.AspNet.Mvc.FunctionalTests
 
             // Assert
             var body = await response.Content.ReadAsStringAsync();
-            Assert.Contains("CategorySetInConverter", body);
+            dynamic d = JObject.Parse(body);
+            Assert.Equal("CategorySetInConverter", (string)d.ProductCategory.CategoryName);
+
         }
     }
 }
--- a/test/Microsoft.AspNet.Mvc.FunctionalTests/JsonPatchTest.cs
+++ b/test/Microsoft.AspNet.Mvc.FunctionalTests/JsonPatchTest.cs
@@ -303,5 +303,55 @@ namespace Microsoft.AspNet.Mvc.FunctionalTests
             var body = await response.Content.ReadAsStringAsync();
             Assert.Equal("{\"\":[\"The input was not valid.\"]}", body);
         }
+
+        [Fact]
+        public async Task JsonPatch_JsonConverterOnProperty_Success()
+        {
+            // Arrange
+            var server = TestHelper.CreateServer(_app, SiteName, _configureServices);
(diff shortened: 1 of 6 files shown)


test/Microsoft.AspNet.Mvc.FunctionalTests/JsonPatchTest.cs
```diff
@@ -303,5 +303,55 @@ public static IEnumerable<object[]> InvalidJsonPatchData
...
+                Method = new HttpMethod("PATCH"),
+                RequestUri = new Uri("http://localhost/jsonpatch/JsonPatchForProduct")
+            };
+
+            // Act
+            var response = await client.SendAsync(request);
+
+            // Assert
+            var body = await response.Content.ReadAsStringAsync();
+            Assert.Contains("CategorySetInConverter", body);
```
rynowak: There are much better ways to do this than to just use a `.Contains` on a string - can we deserialize the value and verify it that way? (same comment as above)

test/WebSites/JsonPatchWebSite/ProductCategoryConverter.cs
```diff
@@ -0,0 +1,47 @@
...
+    public class ProductCategoryConverter : JsonConverter
+    {
+        public override bool CanConvert(Type objectType)
+        {
+            return true;
+        }
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            var obj = JObject.Load(reader);
```
rynowak: Not used right?

test/WebSites/JsonPatchWebSite/ProductCategoryConverter.cs
```diff
@@ -0,0 +1,47 @@
...
+            var category = new Category();
+            category.CategoryName = "CategorySetInConverter";
+
+            return category;
+        }
+
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            JObject jo = new JObject();
```
rynowak: var

test/WebSites/JsonPatchWebSite/ProductCategoryConverter.cs
```diff
@@ -0,0 +1,47 @@
...
+
+            foreach (var prop in value.GetType().GetTypeInfo().GetProperties())
+            {
+                if (prop.CanRead)
+                {
+                    object propValue = prop.GetValue(value);
```
rynowak: var, don't abbreviate

rynowak:
:watch:

test/WebSites/JsonPatchWebSite/ProductCategoryConverter.cs
kirthik: Yes but since reader has value and if I don't load it, then I was getting an error "Additional text found in JSON string after finishing deserializing object."

test/Microsoft.AspNet.Mvc.FunctionalTests/JsonPatchTest.cs
kirthik: The reason I din't use `JsonConvert.Deserialize()` is because that calls `ReadJson()` on custom converter again and now we can't be sure if `Category` has values from the deserialize call in `ObjectAdapter` code or this one.

test/WebSites/JsonPatchWebSite/ProductCategoryConverter.cs
rynowak: Then please add a comment that it must be called for its side-effect.

test/Microsoft.AspNet.Mvc.FunctionalTests/JsonPatchTest.cs
rynowak: You can deserizalize it into a dynamic, anonymous object, or dictionary. All of those are better options than just doing a `.Contains`

kirthik:
@rynowak : updated

test/WebSites/JsonPatchWebSite/ReplaceOrderTypeConverter.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        public override bool CanConvert(Type objectType)
+        {
+            return true;
+        }
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            return "OrderTypeSetInConverter";
+
```
rynowak: extra line

test/WebSites/JsonPatchWebSite/ReplaceOrderTypeConverter.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace JsonPatchWebSite
+{
+    public class ReplaceOrderTypeConverter : JsonConverter
+    {
+        public override bool CanConvert(Type objectType)
+        {
+            return true;
+        }
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
```
rynowak: long line

test/WebSites/JsonPatchWebSite/ProductCategoryConverter.cs
```diff
@@ -0,0 +1,50 @@
...
+            category.CategoryName = "CategorySetInConverter";
+
+            return category;
+        }
+
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            var jo = new JObject();
+
+            foreach (var prop in value.GetType().GetTypeInfo().GetProperties())
```
rynowak: don't abbreviate

rynowak:

kirthik:
merged - 8b5223518f3fb1d4b90744370ef4a08d9a2a31c9

PR #2567, opened 2015-05-15
React to directive error squiggly fix 

ajaybhargavb:
Related PR - aspnet/Razor#392
@NTaylorMullen @dougbu

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor.Host/MvcRazorCodeParser.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor.Host/MvcRazorCodeParser.cs
@@ -50,17 +50,18 @@ namespace Microsoft.AspNet.Mvc.Razor
         {
             // Verify we're on the right keyword and accept
             AssertDirective(ModelKeyword);
+            var startModelLocation = CurrentLocation;
             AcceptAndMoveNext();
 
-            var endModelLocation = CurrentLocation;
 
             BaseTypeDirective(Resources.FormatMvcRazorCodeParser_KeywordMustBeFollowedByTypeName(ModelKeyword),
                               CreateModelCodeGenerator);
 
             if (_modelStatementFound)
             {
-                Context.OnError(endModelLocation,
-                                Resources.FormatMvcRazorCodeParser_OnlyOneModelStatementIsAllowed(ModelKeyword));
+                Context.OnError(startModelLocation,
+                                Resources.FormatMvcRazorCodeParser_OnlyOneModelStatementIsAllowed(ModelKeyword),
+                                ModelKeyword.Length);
             }
 
             _modelStatementFound = true;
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.Mvc.Razor.Host/MvcRazorCodeParser.cs
```diff
@@ -50,17 +50,19 @@ protected virtual void ModelDirective()
         {
             // Verify we're on the right keyword and accept
             AssertDirective(ModelKeyword);
+            var startModelLocation = CurrentLocation;
             AcceptAndMoveNext();
+            var keywordLength = Span.GetContent().Value.Length;
```
NTaylorMullen: `ModelKeyword.Length`

dougbu: And move down just before it's used.  No reason to calculate at all unless adding an error.

NTaylorMullen: :+1:

NTaylorMullen:
:shipit:

src/Microsoft.AspNet.Mvc.Razor.Host/MvcRazorCodeParser.cs
```diff
@@ -95,8 +100,10 @@ protected virtual void InjectDirective()
             var hasTypeError = !At(CSharpSymbolType.Identifier);
             if (hasTypeError)
             {
-                Context.OnError(CurrentLocation,
-                                Resources.FormatMvcRazorCodeParser_KeywordMustBeFollowedByTypeName(InjectKeyword));
+                Context.OnError(
+                    startLocation,
+                    Resources.FormatMvcRazorCodeParser_KeywordMustBeFollowedByTypeName(InjectKeyword),
+                    keywordwithSingleWhitespaceLength + remainingWhitespaceLength);
```
dougbu: Be consistent.  Not including any whitespace in the `model` error.

src/Microsoft.AspNet.Mvc.Razor.Host/MvcRazorCodeParser.cs
```diff
@@ -105,14 +112,14 @@ protected virtual void InjectDirective()
...
             AcceptWhile(IsSpacingToken(includeNewLines: false, includeComments: true));
 
             if (!hasTypeError && (EndOfFile || At(CSharpSymbolType.NewLine)))
             {
                 // Add an error for the property name only if we successfully read the type name
-                Context.OnError(propertyStartLocation,
-                                Resources.FormatMvcRazorCodeParser_InjectDirectivePropertyNameRequired(InjectKeyword));
+                Context.OnError(startLocation,
+                                Resources.FormatMvcRazorCodeParser_InjectDirectivePropertyNameRequired(InjectKeyword),
+                                keywordwithSingleWhitespaceLength + remainingWhitespaceLength);
```
dougbu: This is going to look strange.  Suggest the error include all of the directive that was present -- the type name as well as the keyword?

dougbu:
General thought: In an erroneous line such as `@model  `, errors will highlight `model` with this change.  Would including the entire line or at least the leading `@` make more sense?  (Highlighting either `@model  ` or `@model` here.)

Could / should we do that for all errors within directives?

dougbu:
:watch:

NTaylorMullen:
Whatever we do for `@model` we should do for `@inherits`

ajaybhargavb:
@dougbu, I don't think highlighting the leading `@` is a good idea because it won't be consistent with other errors. Eg: When you type `@abc` without creating the variable `abc` it is going to highlight `abc` and not `@abc`.

NTaylorMullen:
oo I agree with @ajaybhargavb here, we don't highlight the `@` for any other directive and never have

src/Microsoft.AspNet.Mvc.Razor.Host/MvcRazorCodeParser.cs
ajaybhargavb: :+1: I am removing whitespace from this error. Will also change `@inherits` to not include the trailing whitespace.

ajaybhargavb:
Updated.

dougbu:
:shipit:

PR #2574, opened 2015-05-18
Fix #2407 - Part 1 - Make model binding behavior for [Required] compatible with MVC5

### LINKED ISSUE #2407: Add back the implicit [Required] for value types

rynowak:
This change removes the behavior in model binding to validate values 'on
the wire' for requiredness instead of the looking at the model. This
restores the behavior of [Required] for model binding to the MVC5
semantics.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
@@ -349,17 +349,6 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
                 {
                     validationInfo.RequiredProperties.Add(propertyName);
                 }
-
-                var validatorProviderContext = new ModelValidatorProviderContext(propertyMetadata);
-                bindingContext.OperationBindingContext.ValidatorProvider.GetValidators(validatorProviderContext);
-
-                var requiredValidator = validatorProviderContext.Validators
-                    .FirstOrDefault(v => v != null && v.IsRequired);
-                if (requiredValidator != null)
-                {
-                    validationInfo.RequiredValidators[propertyName] = requiredValidator;
-                    validationInfo.RequiredProperties.Add(propertyName);
-                }
             }
 
             return validationInfo;
@@ -393,15 +382,18 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
                     bindingContext.ModelName,
                     propertyName);
 
-                // Execute validator (if any) to get custom error message.
-                IModelValidator validator;
-                if (validationInfo.RequiredValidators.TryGetValue(missingRequiredProperty, out validator))
+                // Get the first 'required' validator (if any) to get custom error message.
+                var validatorProviderContext = new ModelValidatorProviderContext(propertyExplorer.Metadata);
+                bindingContext.OperationBindingContext.ValidatorProvider.GetValidators(validatorProviderContext);
+                var validator = validatorProviderContext.Validators.FirstOrDefault(v => v.IsRequired);
+
+                if (validator != null)
                 {
                     addedError = RunValidator(validator, bindingContext, propertyExplorer, modelStateKey);
                 }
 
-                // Fall back to default message if BindingBehaviorAttribute required this property or validator
-                // (oddly) succeeded.
+                // Fall back to default message if BindingBehaviorAttribute required this property and we have no
+                // actual validator for it.
                 if (!addedError)
                 {
                     bindingContext.ModelState.TryAddModelError(
@@ -418,12 +410,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
                 if (dtoResult != null)
                 {
                     var propertyMetadata = entry.Key;
-                    IModelValidator requiredValidator;
-                    validationInfo.RequiredValidators.TryGetValue(
-                        propertyMetadata.PropertyName,
-                        out requiredValidator);
-
-                    SetProperty(bindingContext, modelExplorer, propertyMetadata, dtoResult, requiredValidator);
+                    SetProperty(bindingContext, modelExplorer, propertyMetadata, dtoResult);
 
                     var dtoValidationNode = dtoResult.ValidationNode;
                     if (dtoValidationNode == null)
@@ -458,8 +445,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             [NotNull] ModelBindingContext bindingContext,
             [NotNull] ModelExplorer modelExplorer,
             [NotNull] ModelMetadata propertyMetadata,
-            [NotNull] ModelBindingResult dtoResult,
-            IModelValidator requiredValidator)
+            [NotNull] ModelBindingResult dtoResult)
         {
             var bindingFlags = BindingFlags.Instance | BindingFlags.Public | BindingFlags.IgnoreCase;
             var property = bindingContext.ModelType.GetProperty(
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
```diff
@@ -485,26 +471,6 @@ internal static PropertyValidationInfo GetPropertyValidationInfo(ModelBindingCon
                 value = dtoResult.Model;
             }
 
-            // 'Required' validators need to run first so that we can provide useful error messages if
-            // the property setters throw, e.g. if we're setting entity keys to null.
```
rynowak: Added some code to the `DefaultObjectValidator` to sort them.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
```diff
@@ -512,31 +478,16 @@ internal static PropertyValidationInfo GetPropertyValidationInfo(ModelBindingCon
...
+            catch (Exception exception)
             {
-                // trying to set a non-nullable value type to null, need to make sure there's a message
-                var modelStateKey = dtoResult.Key;
-                var validationState = bindingContext.ModelState.GetFieldValidationState(modelStateKey);
-                if (validationState == ModelValidationState.Unvalidated)
-                {
-                    var errorMessage = Resources.ModelBinding_ValueRequired;
-                    bindingContext.ModelState.TryAddModelError(modelStateKey, errorMessage);
-                }
```
rynowak: Had some discussion - we think that this case means you've got a badly-behaving model binder. We're not going to special case it. I reduced the number of tests hitting this because it was overkill.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/MutableObjectModelBinderTest.cs
```diff
@@ -905,14 +905,13 @@ public void ProcessDto_BindRequiredFieldNull_RaisesModelError()
...
-            Assert.NotNull(modelError.ErrorMessage);
-            Assert.Equal("A value is required.", modelError.ErrorMessage);
+            Assert.Equal(string.Empty, modelError.ErrorMessage);
+            Assert.IsType<NullReferenceException>(modelError.Exception);
         }
 
         [Fact]
         [ReplaceCulture]
-        public void ProcessDto_RequiredFieldMissing_RaisesModelError()
+        public void ProcessDto_MissingDataForRequiredFields_NoErrors()
```
rynowak: A number of these tests have been simplified because we expect the validation system to produce these errors now.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/MutableObjectModelBinderTest.cs
```diff
@@ -1594,50 +1553,20 @@ public void SetProperty_SettingNonNullableValueTypeToNull_RequiredValidatorNotPr
...
-                dtoResult,
-                requiredValidator);
+                dtoResult);
 
             // Assert
             Assert.False(bindingContext.ModelState.IsValid);
-        }
 
-        [Fact]
-        public void SetProperty_SettingNonNullableValueTypeToNull_RequiredValidatorPresent_AddsModelError()
```
rynowak: This test becomes a dupe of some others after fixing the issues with it.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/MutableObjectModelBinderTest.cs
```diff
@@ -1673,41 +1599,6 @@ public void SetProperty_SettingNullableTypeToNull_RequiredValidatorNotPresent_Pr
                          bindingContext.ModelState["foo.NameNoAttribute"].Errors[0].Exception.Message);
         }
 
-        [Fact]
-        public void SetProperty_SettingNullableTypeToNull_RequiredValidatorPresent_PropertySetterThrows_AddsRequiredMessageString()
```
rynowak: Same comment, this becomes a dupe.

test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/MvcTagHelpersWebSite.MvcTagHelper_Customer.Index.html
```diff
@@ -7,7 +7,7 @@
         <div>
             <label class="order" for="Number">Number</label>
             <input class="form-control input-validation-error" type="number" data-val="true" data-val-range="The field Number must be between 1 and 100." data-val-range-max="100" data-val-range-min="1" id="Number" name="Number" value="" />
-            <span class="field-validation-error" data-valmsg-for="Number" data-valmsg-replace="true">A value is required.</span>
+            <span class="field-validation-error" data-valmsg-for="Number" data-valmsg-replace="true">The value &#x27;&#x27; is invalid.</span>
```
rynowak: This is an intentional behavior change here. This code was relying on 'on the wire' semantics for a value type.

Eilon:
:shipit:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
dougbu: Code was previously here to execute required validators before attempting to set the property, not before other validators. `DefaultObjectValidator` execution comes much later, after all the setting has settled.

With this change, the first error message will be from the `Exception` and not the required validator.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
```diff
@@ -394,14 +383,17 @@ internal static PropertyValidationInfo GetPropertyValidationInfo(ModelBindingCon
                     propertyName);
 
                 // Execute validator (if any) to get custom error message.
-                IModelValidator validator;
-                if (validationInfo.RequiredValidators.TryGetValue(missingRequiredProperty, out validator))
+                var validatorProviderContext = new ModelValidatorProviderContext(propertyExplorer.Metadata);
+                bindingContext.OperationBindingContext.ValidatorProvider.GetValidators(validatorProviderContext);
+                var validator = validatorProviderContext.Validators.FirstOrDefault();
+
+                if (validator != null)
```
dougbu: Why run the first validator if it does not have `IsRequired == true`?

Further, the error (if any) will likely be incorrect and the "fall back" message better. Until and unless we add a new always-first validator for `[BindRequired]` cases, suggest removing lines 386 through 393 (new line numbers) entirely.

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
rynowak: > With this change, the first error message could be from an Exception and not the required validator.

That's exactly the point. The model property is only set when `isModelSet == true`. An exception can only occur in a case where the model binder has returned a value that can't be set **and** indicated that it should be set on the model.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
rynowak: Yep, this is a bug. It should be `FirstOrDefault(v => v.IsRequired)`

rynowak:
Updated

rynowak:
Also added 2 tests for the case where `[BindRequired]` and `[Required]` play together.

dougbu:
Agree with the separation of bugs we discussed offline. This PR is good to go.

dougbu:
:shipit:

Issue #2540, opened 2015-05-11
Change the tag helpers `FileVersion` property to something clearer


Eilon:
We have a few tag helpers (`Link`, `Script`, and the upcoming `Image`) that have a `FileVersion` property that indicates whether to add a "version" to the URL to help with cache busting.

The property name is a bit confusing: it looks like *it is* a file version, as opposed to saying whether to *append* a file version.

I recommend changing the property to be `AppendVersion` or similar because that'll be a lot clearer about what it does.

danroth27:
@DamianEdwards thoughts?

DamianEdwards:
Sure

dpaquette:
Any objection to me tackling this one on the weekend?

Eilon:
Yeah that would be great. Let's settle on a name. Here are a few ideas:

* `AppendFileVersion` / `AddFileVersion`
* `AppendVersion` / `AddVersion`
* `AppendFileHash` / `AddFileHash`
* `AppendHash` / `AddHash`
* `CacheBust`

I do kind of like my original suggestion of `AppendVersion` because it's concise and reasonably clear about what it does.

Thoughts?

danroth27:
I can live with AppendVersion. I also like AddFileVersion.

dpaquette:
I like AppendVersion

dougbu:
Is this about just the property name?  Note that's irrelevant in every scenario except unit testing.  The _attribute_ name (what users see in the Razor source) is currently `asp-file-version`.

If we're going to change the attribute as well, would the name there be `asp-append-version`?

danroth27:
Yup.

Eilon:
@dougbu yes it's just the property name and yes we'd rename the tag helper attribute name.

Sounds like everyone wants to go with `AppendVersion`!

DamianEdwards:
:+1:

dougbu:
3fe0490

Daniel15:
The web app template in Visual Studio still uses `asp-file-version`, looks like it wasn't updated.

danroth27:
https://github.com/aspnet/Templates/issues/142

Daniel15:
I get a 404 on that link; is it a private repo?

danroth27:
Yeah, sorry about that. Removing the link to avoid further confusion.

PR #2580, opened 2015-05-19
Renamed FileVersion to AppendVersion in tag helpers that support cache busting

dpaquette:
FileVersion property renamed to AppendVersion in ImageTagHelper, LinkTagHelper and ScriptTagHelper.

asp-file-version attribute renamed to asp-append-version.

Resolves issue #2540

### DIFF
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
@@ -29,7 +29,7 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
     [TargetElement("link", Attributes = FallbackTestClassAttributeName)]
     [TargetElement("link", Attributes = FallbackTestPropertyAttributeName)]
     [TargetElement("link", Attributes = FallbackTestValueAttributeName)]
-    [TargetElement("link", Attributes = FileVersionAttributeName)]
+    [TargetElement("link", Attributes = AppendVersionAttributeName)]
     public class LinkTagHelper : TagHelper
     {
         private static readonly string Namespace = typeof(LinkTagHelper).Namespace;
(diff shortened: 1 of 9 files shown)


src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -88,7 +88,7 @@ private enum Mode
             /// <summary>
             /// Just adding a file version for the generated urls.
             /// </summary>
-            FileVersion = 0,
+            AppendVersion = 0,
             /// <summary>
```
dougbu: While here please add blank lines within this `enum` to separate the members. Same in `ScriptTagHelper`.

dougbu:
:shipit: let me know if you can't address my minor comment. If so, I'll merge as-is and we'll do that cleanup some other time.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
dpaquette: Will do

dpaquette:
Added blank lines as requested

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -136,8 +138,8 @@ private enum Mode
         /// <remarks>
         /// If <c>true</c> then a query string "v" with the encoded content of the file is added.
         /// </remarks>
-        [HtmlAttributeName(FileVersionAttributeName)]
-        public bool? FileVersion { get; set; }
+        [HtmlAttributeName(AppendVersionAttributeName)]
+        public bool? AppendVersion { get; set; }
```
Eilon: I need a history lesson on this: why is this a nullable bool? Why not just a bool? Does it ever matter whether it was "not set" or "set to false"? I ask because in WebForms no controls ever needed this behavior (ok fine, and nullable didn't exist in .NET 1.0 either, but it was still possible to implement manually).

Eilon: @DamianEdwards

Eilon:
But anyway, the code here looks great. BTW we're working on some stabilization right now so we will likely wait to merge this until next week if that's ok!

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
dougbu: @Eilon IIRC this might be historical / hysterical. Currently determine mode using `TagHelperContext` (its `AllAttributes` collection) and not the properties themselves. So nullable doesn't look mandatory anymore.

But I suggest we clean this up separately if @DamianEdwards agrees.

dpaquette:
No problem waiting. Let me know if there is anything else I need to do.

dougbu:
@dpaquette sorry to say, there is something else you can do. Not urgent (due to the long weekend) but please rebase your fork on our dev branch and fix the conflicts in your commit. No need to create a new PR; just `git push -f` to your server branch. Let me know when it's updated.

tl;dr; Our recent stabilization work and large number of breaking changes means I can no longer automatically rebase your branch. If I were to address the conflicts, it wouldn't be your change anymore. There are ways around this but you fixing the conflicts is simpler.

dpaquette:
No problem. I will do that this weekend

dougbu:
3fe0490

PR #2583, opened 2015-05-19
Fix #1579 - Bind top-level collections as an empty collection

### LINKED ISSUE #1579: Action argument of IList<T> will be bound as Null instead of an empty List<T>
Today CollectionModelBinder returns false if a value provider cannot provide a value, it is the mutable object binder which creates a model for it. 
Since it is the ```CollectionModelBinder``` which is responsible for binding collections, it should be the one which creates empty object and not let it slip for other model binders.

rynowak:
This change treats 'top-level' collection-type models similarly to
top-level POCO model - namely that they will always be instantiated even
if there's no data to put inside.


src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
```diff
@@ -13,25 +13,54 @@ public class GenericModelBinder : IModelBinder
...
+                    // Success - propegate the values returned by the model binder.
+                    return new ModelBindingResult(result.Model, result.Key, result.IsModelSet, result.ValidationNode);
+                }
+
+                // If this is the fallback case, and we didn't bind as a top-level model, then generate a
+                // default 'empty' model and return it.
+                var isTopLevelObject = bindingContext.ModelMetadata.ContainerType == null;
+                var hasExplicitAlias = bindingContext.BinderModelName != null;
+
+                if (isTopLevelObject && (hasExplicitAlias || bindingContext.ModelName == string.Empty))
```
rynowak: This is a simplified version of the logic used by `MutableObjectModelBinder` - simplified because we don't have to handle as many cases.

test/Microsoft.AspNet.Mvc.IntegrationTests/ArrayModelBinderIntegrationTest.cs
```diff
@@ -160,8 +160,8 @@ public class ArrayModelBinderIntegrationTest
             var modelBindingResult = await argumentBinder.BindModelAsync(parameter, modelState, operationContext);
 
             // Assert
-            Assert.NotNull(modelBindingResult);  // This fails due to #1579
-            Assert.False(modelBindingResult.IsModelSet);
+            Assert.NotNull(modelBindingResult);
+            Assert.True(modelBindingResult.IsModelSet);
```
rynowak: This was a mistake in writing these tests originally. `modelBindingResult.IsModelSet == true` is consistent with the result when an empty POCO is bound.

I did a CTRL+F for #1579 in the code and was able to enable all of them.

test/Microsoft.AspNet.Mvc.IntegrationTests/ByteArrayModelBinderIntegrationTest.cs
```diff
@@ -76,7 +76,7 @@ private class Person
             Assert.Equal(ModelValidationState.Valid, modelState[key].ValidationState); // Should be skipped. bug#2447
         }
 
-        [Fact]
+        [Fact(Skip = "ByteArrayModelBinder should return a non-null result #2456")]
         public async Task BindParameter_NoData_DoesNotGetBound()
```
rynowak: I marked this one as failing... because it's passing for the wrong reason. The verification steps in the test code are wrong, and after correcting them it will pass because `GenericModelBinder` results in the same result.

I still put in the 'skip' here to have a reminder that we have work to do.

test/Microsoft.AspNet.Mvc.IntegrationTests/ValidationIntegrationTests.cs
```diff
@@ -997,14 +997,14 @@ private class Order10
...
-            Assert.False(modelBindingResult.IsModelSet);
+            Assert.True(modelBindingResult.IsModelSet);
 
             var model = Assert.IsType<List<Order10>>(modelBindingResult.Model);
             Assert.Empty(model);
 
             Assert.Equal(0, modelState.Count);
             Assert.Equal(0, modelState.ErrorCount);
-            Assert.False(modelState.IsValid);
+            Assert.True(modelState.IsValid);
```
rynowak: This was just a bad copy-paste I think. Because there's no data, there's nothing to be invalid.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
rynowak: Working on adding a few unit tests for this logic.

rynowak:
@dougbu updated with some unit tests

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
```diff
@@ -39,57 +68,65 @@ private static Type ResolveBinderType(Type modelType)
...
+                    ModelBinderType = modelBinderType,
+                    UnderlyingModelType = modelType,
+                };
             }
+
             return null;
         }
 
-        private static Type GetCollectionBinder(Type modelType)
+        private static GenericModelBindingInfo GetCollectionBinder(Type modelType)
```
ajaybhargavb: Suggest renaming this method and similar ones below.
GetCollectionBinder => GetCollectionBinderInfo

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
```diff
@@ -13,25 +13,54 @@ public class GenericModelBinder : IModelBinder
...
                 var result = await binder.BindModelAsync(bindingContext);
 
-                var modelBindingResult = result != null ?
-                    new ModelBindingResult(result.Model, result.Key, result.IsModelSet, result.ValidationNode) :
-                    new ModelBindingResult(model: null, key: bindingContext.ModelName, isModelSet: false);
-
-                // Were able to resolve a binder type.
+                if (result != null && result.IsModelSet)
+                {
+                    // Success - propegate the values returned by the model binder.
```
dougbu: typo: propagate

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
```diff
@@ -13,25 +13,54 @@ public class GenericModelBinder : IModelBinder
...
+                    return new ModelBindingResult(result.Model, result.Key, result.IsModelSet, result.ValidationNode);
```
dougbu: why not `return result;`?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
```diff
@@ -13,25 +13,54 @@ public class GenericModelBinder : IModelBinder
...
+                }
+
+                // If this is the fallback case, and we didn't bind as a top-level model, then generate a
+                // default 'empty' model and return it.
+                var isTopLevelObject = bindingContext.ModelMetadata.ContainerType == null;
+                var hasExplicitAlias = bindingContext.BinderModelName != null;
+
+                if (isTopLevelObject && (hasExplicitAlias || bindingContext.ModelName == string.Empty))
+                {
+
```
dougbu: remove x-tra blank line

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
```diff
@@ -13,25 +13,54 @@ public class GenericModelBinder : IModelBinder
...
+                    var model = result?.Model;
```
dougbu: How could `result?.Model` ever be non-`null` if we reach here?  Have already confirmed `!IsModelSet`.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
```diff
@@ -13,25 +13,54 @@ public class GenericModelBinder : IModelBinder
...
+
+                    var validationNode = new ModelValidationNode(
+                        bindingContext.ModelName,
+                        bindingContext.ModelMetadata,
+                        model);
+
+                    return new ModelBindingResult(model, bindingContext.ModelName, true, validationNode);
+                }
+
+                // We always want to return a result for model-types that we handle.
```
dougbu: don't hyphenate "model-types"; confusing because it appears to introduce a new term.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
```diff
@@ -13,25 +13,54 @@ public class GenericModelBinder : IModelBinder
...
+                //
                 // Always tell the model binding system to skip other model binders i.e. return non-null.
```
dougbu: The existing comment and what you added two lines above look much more closely-related than the separation implies. Suggest merging into one paragraph. Something like "We always want ... This tells the model binding system ..." (and remove the "i.e." bit).

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
```diff
@@ -39,57 +68,65 @@ private static Type ResolveBinderType(Type modelType)
...
             return null;
         }
 
-        /// <remarks>
-        /// Example: <c>GetGenericBinderType(typeof(IList&lt;T&gt;), typeof(List&lt;T&gt;),
-        /// typeof(ListBinder&lt;T&gt;), ...)</c> means that the <c>ListBinder&lt;T&gt;</c> type can update models that
-        /// implement <see cref="IList{T}"/>, and if for some reason the existing model instance is not updatable the
-        /// binder will create a <see cref="List{T}"/> object and bind to that instead. This method will return
-        /// <c>ListBinder&lt;T&gt;</c> or <c>null</c>, depending on whether the type and updatability checks succeed.
-        /// </remarks>
```
dougbu: Suggest preserving the example as a regular `//` comment.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
```diff
@@ -13,25 +13,54 @@ public class GenericModelBinder : IModelBinder
...
+
+                    var model = result?.Model;
+                    if (model == null && bindingInfo.UnderlyingModelType.IsArray)
+                    {
+                        model = Array.CreateInstance(bindingInfo.UnderlyingModelType.GetElementType(), 0);
+                    }
+                    else if (model == null)
+                    {
+                        model = Activator.CreateInstance(bindingInfo.UnderlyingModelType);
+                    }
```
dougbu: `Activator` feels like overkill here. Still have the `IModelBinder` instance and it's always either a `KeyValuePairModelBinder` or a `CollectionModelBinder`. That instance also already knows how to create a new instance (they're generic so it's a simple `new`). Given that, suggest creating a new `interface IGenericModelBinder` those two classes would implement. `IGenericModelBinder` would include an `object CreateEmptyModel()` method and inherit from `IModelBinder`.

test/Microsoft.AspNet.Mvc.IntegrationTests/ByteArrayModelBinderIntegrationTest.cs
dougbu: Could you make it a bit more clear what needs to change in the verification steps? That is, add code comments or more to the `Skip` message.

dougbu: Or did you already fix the verification?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
dougbu: You're done "working on" those tests, correct?

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
rynowak: yep.

test/Microsoft.AspNet.Mvc.IntegrationTests/ByteArrayModelBinderIntegrationTest.cs
rynowak: I already fixed the verification.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
rynowak: :+1:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
rynowak: sure

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
rynowak: Fixed

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
rynowak: fixed

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
rynowak: No idea what I was thinking.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
rynowak: sure

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
rynowak: yep

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
rynowak: sure

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
rynowak: I don't agree at all. Creating a new interface so we can call `new` instead of `Activator.CreateInstance` is overkill.

I agree with what you're saying from a separation of concerns point of view, but I'd rather us finish the quality push and do a real refactor than keep creating things.

rynowak: What would you think about moving this concern down into the those model-binders entirely - IE, have ArrayModelBinder decide when to create the model?

That might be simpler, and would address this design-point.

rynowak:
@dougbu Addressed most feedback, have one idea we should discuss.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
```diff
@@ -13,83 +13,129 @@ public class GenericModelBinder : IModelBinder
...
+
+                    var validationNode = new ModelValidationNode(
+                        bindingContext.ModelName,
+                        bindingContext.ModelMetadata,
+                        model);
+
+                    return new ModelBindingResult(model, bindingContext.ModelName, true, validationNode);
```
dougbu: pls name the `true` argument

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
```diff
@@ -13,83 +13,129 @@ public class GenericModelBinder : IModelBinder
...
+        // Example: 
+        //      GetGenericBinderType(typeof(IList<T>), typeof(List<T>), typeof(ListBinder<T&>), ...)
+        //
+        // This means that the ListBinder<T> type can work with models that implement IList<T>, and if there is no 
+        // existing model instance, the binder will create a List{T}. 
+        //
+        // This method will return null if the given model type isn't compatible with the combination of
+        // supportedInterfaceType and modelType. If supportedInterfaceType and modelType are compatible, then
+        // it will return the closed-generic newInstanceType and closed-generic openBinderType.
+        // </remarks>
```
dougbu: remove line

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
```diff
@@ -13,83 +13,129 @@ public class GenericModelBinder : IModelBinder
...
-        /// typeof(ListBinder&lt;T&gt;), ...)</c> means that the <c>ListBinder&lt;T&gt;</c> type can update models that
-        /// implement <see cref="IList{T}"/>, and if for some reason the existing model instance is not updatable the
-        /// binder will create a <see cref="List{T}"/> object and bind to that instead. This method will return
-        /// <c>ListBinder&lt;T&gt;</c> or <c>null</c>, depending on whether the type and updatability checks succeed.
-        /// </remarks>
-        private static Type GetGenericBinderType(Type supportedInterfaceType,
-                                                 Type newInstanceType,
-                                                 Type openBinderType,
-                                                 Type modelType)
+        //
```
dougbu: remove line

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
```diff
@@ -13,83 +13,129 @@ public class GenericModelBinder : IModelBinder
...
+        // Example: 
+        //      GetGenericBinderType(typeof(IList<T>), typeof(List<T>), typeof(ListBinder<T&>), ...)
+        //
+        // This means that the ListBinder<T> type can work with models that implement IList<T>, and if there is no 
+        // existing model instance, the binder will create a List{T}. 
+        //
+        // This method will return null if the given model type isn't compatible with the combination of
+        // supportedInterfaceType and modelType. If supportedInterfaceType and modelType are compatible, then
+        // it will return the closed-generic newInstanceType and closed-generic openBinderType.
```
dougbu: "closed generic" (In addition to deleting things, you also like hyphenating things that aren't hyphenated terms.)

rynowak:
Updated. Moved the model-creation down into the CollectionModelBinder classes.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/KeyValuePairModelBinder.cs
```diff
@@ -62,8 +62,27 @@ public sealed class KeyValuePairModelBinder<TKey, TValue> : IModelBinder
             }
             else
             {
-                // Caller (GenericModelBinder) was able to resolve a binder type and will create a ModelBindingResult
-                // that exits current ModelBinding loop.
+                // If this is the fallback case, and we failed to find data as a top-level model, then generate a
+                // default 'empty' model and return it.
```
dougbu: Odd case. Curious whether Web API 2 or MVC 5 pass `default(KeyValuePair<TKey, TValue>)` into an action. Of course it's fine to support here.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/GenericModelBinderTest.cs
```diff
@@ -0,0 +1,17 @@
...
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Http.Internal;
+using Microsoft.AspNet.Mvc.ModelBinding;
+using Xunit;
+
+namespace Microsoft.AspNet.Mvc
+{
+    public class GenericModelBinderTest
```
dougbu: Remove empty file from the commit.

dougbu:
:shipit:

PR #2591, opened 2015-05-21
Fix #2582 - Remove @inject dependency on [Activate]

### LINKED ISSUE #2582: Remove dependency of @inject on [Activate]
Right now `@inject` creates a public property that's decorated with `[Activate]`. Given the recent push back on how `[Activate]` isn't the best feature we need to bullet proof the Razor code generation to no longer depend on it.

Assigning to beta5 and to myself. Talked to @DamianEdwards

rynowak:
Removes usage of [Activate] in razor code generation. Razor will now use
it's own special attribute recognized by the Razor activator.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor/Internal/RazorInjectAttribute.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/Internal/RazorInjectAttribute.cs
@@ -1,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNet.Mvc.Razor.Internal
+{
+    [AttributeUsage(AttributeTargets.Property, AllowMultiple = false, Inherited = true)]
+    public class RazorInjectAttribute : Attribute
+    {
+    }
+}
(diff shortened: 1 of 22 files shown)


src/Microsoft.AspNet.Mvc.Razor/Internal/RazorInjectAttribute.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNet.Mvc.Razor.Internal
+{
+    public class RazorInjectAttribute : Attribute
```
rynowak: Will add `[AttributeUsage(AttributeTargets.Property, AllowMultiple = false, Inherited = true)]`

dougbu:
:shipit: after the MVC problems clear. Does what's on the tin.

NTaylorMullen:
:shipit:

pranavkm:
:shipit:

NTaylorMullen:
Actually @rynowak just had an epiphany tonight on how to remove the "activatish" attribute entirely. Only reason why the private backing field was bad was because it'd show up in IntelliSense... On `CodeBuilderContext` there's a `DesignTime` member indicating if it's generating code in `DesignTime`. You can actually conditionally generate the backing field/the getter content based on if its `DesignTime` or not. This way you have the best of both worlds:
- No `[Activate]`
- No IntelliSense badness.

rynowak:
yay. :heart_eyes: 

I'm going to go ahead and check in this fix when the build is clean, and time given tomorrow do your suggestion.

Issue #2151, opened 2015-03-10
Naming inconsistency between [Activate] and @inject


justinvp:
Is there a reason why these are named differently?

 - In Razor it's currently `@inject`
 - In Mvc it's currently `[Activate]`

Why not `[Inject]` and `@inject`, or `[Activate]` and `@activate`?

eamodio:
I wondered the same thing. And inject makes a lot of sense, but activate? Seems very unclear what that does

danroth27:
@rynowak @DamianEdwards @Eilon Do we still plan to move from [Activate] to [FromServices]?

justinvp:
Personally, I prefer `[Inject]` over `[Activate]` and `[FromServices]`. I agree with @eamodio on the ambiguity of "activate".

Whatever name is chosen, I really hope it is consistent between Mvc and Razor. Using different names for the same thing increases the concept count (more cognitive load) unnecessarily.

I'm not sure `[FromServices]` will be as nice in Razor... `@fromServices`?

eamodio:
I will grant you that [FromServices] is better than [Activate], but I would still strongly argue for [Inject] or [Injected]. IMO, using "inject" aligns well with Dependency Injection while also expressing more with less confusion than [FromServices].

Both basically tell me that the property will be set by someone/thing else, but [FromServices] feels oddly specific about who/what will inject it - while at the same time "Services" is so ambiguous that it feels confusing.

While [Inject]/[Injected] seems much less about how the property will be set and more about the fact that it will be.

Just my 2c.

ctolkien:
:+1:  on `[Inject]`

bangoker:
:+1: on [Inject] as well

atrauzzi:
I agree, `[Inject]` sounds a lot better than `[FromServices]`.  Although most of the time what you're going to be injecting will be a service, the container shouldn't make assumptions about what it's containing and what their intended uses are.

davidroth:
:+1:  on `[Inject]` as it is clear from the name what`s happening. 
`[FromServices]` sounds very ambiguous (What kind of services, is it only for services?, etc.)

nil4:
:+1:  for `[Inject]` rather than `[FromServices]`, it relates clearly to the [DependencyInjection](https://github.com/aspnet/DependencyInjection) package and namespace.

dougbu:
[...]

PR #2592, opened 2015-05-21
Part of #2151 - Remove [Activate] from ViewComponents

rynowak:
This change removes [Activate] from ViewComponents. Accessing context
should be done through [ViewComponentContext]. Accessing services should
be done though constructor injection.


src/Microsoft.AspNet.Mvc.Core/ViewComponents/DefaultViewComponentActivator.cs
```diff
@@ -3,40 +3,38 @@
...
 using System.Reflection;
-using Microsoft.AspNet.Mvc.Rendering;
-using Microsoft.Framework.DependencyInjection;
 using Microsoft.Framework.Internal;
 
 namespace Microsoft.AspNet.Mvc.ViewComponents
 {
     /// <summary>
-    /// Represents the <see cref="IViewComponentActivator"/> that is registered by default.
+    /// A default implemented of <see cref="IViewComponentActivator"/>.
```
dougbu: implemented -> implementation

src/Microsoft.AspNet.Mvc.Core/ViewComponents/ViewComponent.cs
```diff
@@ -96,64 +99,95 @@ public ModelStateDictionary ModelState
...
+
+        [ViewComponentContext]
+        public ViewComponentContext ViewComponentContext
         {
             get
             {
-                if (_viewContext == null)
+                if (_viewComponentContext == null)
                 {
-                    // This should run only for the ViewComponent unit test scenarios
```
dougbu: Why is this comment now incorrect?

src/Microsoft.AspNet.Mvc.Core/ViewComponents/ViewComponent.cs
```diff
@@ -96,64 +99,95 @@ public ModelStateDictionary ModelState
...
-                _viewContext = value;
+                _viewComponentContext = value;
             }
         }
 
         /// <summary>
-        /// Gets or sets the <see cref="ViewDataDictionary"/>.
+        /// Gets the <see cref="ViewContext"/>.
         /// </summary>
         [Activate]
```
dougbu: Remove this line

src/Microsoft.AspNet.Mvc.Core/ViewComponents/ViewComponent.cs
```diff
@@ -96,64 +99,95 @@ public ModelStateDictionary ModelState
...
+        public ICompositeViewEngine ViewEngine
+        {
+            get
+            {
+                if (_viewEngine == null)
                 {
-                    // This should run only for the ViewComponent unit test scenarios
-                    _viewData = new ViewDataDictionary(ViewContext.ViewData);
+                    // May be null in unit-testing scenarios.
+                    var services = ViewComponentContext.ViewContext?.HttpContext?.RequestServices;
```
dougbu: Isn't it `ViewComponentContext` (not its `ViewContext` property) that needs the `null` check?  Could use `ViewContext` property instead.

src/Microsoft.AspNet.Mvc.Core/ViewComponents/DefaultViewComponentActivator.cs
dougbu: Suggest mentioning `[ViewComponentContext]` and fact activator handles only `ViewComponentContext` properties with that attribute.

kirthik: implemented -> implementation

test/Microsoft.AspNet.Mvc.Core.Test/ViewComponentTests.cs
```diff
@@ -180,7 +164,7 @@ public void ViewComponent_ViewContext_ViewData_ReturnsDefaultInstanceIfNull()
             Assert.Empty(viewComponent.ViewContext.ViewData);
             Assert.NotNull(viewComponent.ViewData);
             Assert.Empty(viewComponent.ViewData);
-            Assert.NotSame(viewComponent.ViewData, viewComponent.ViewContext.ViewData);
+            Assert.Same(viewComponent.ViewData, viewComponent.ViewContext.ViewData);
```
dougbu: Glad to here this inconsistency go bye-bye

dougbu:
:watch: mostly because I feel I'm missing something

src/Microsoft.AspNet.Mvc.Core/ViewComponents/DefaultViewComponentActivator.cs
rynowak: :+1: to all

src/Microsoft.AspNet.Mvc.Core/ViewComponents/ViewComponent.cs
rynowak: No idea. Put it back.

src/Microsoft.AspNet.Mvc.Core/ViewComponents/ViewComponent.cs
rynowak: fixt

src/Microsoft.AspNet.Mvc.Core/ViewComponents/ViewComponent.cs
rynowak: ViewComponentContext is guaranteed to not be null. Removed some extra ?s from other places in code.

test/Microsoft.AspNet.Mvc.Core.Test/ViewComponentTests.cs
rynowak: It's back!

It's not an inconsistency, it's a defensive copy

dougbu: We need to discuss this. The "defensive copy" causes a user experience different from that in other contexts e.g. within a controller action.

src/Microsoft.AspNet.Mvc.Core/ViewComponents/ViewComponentContextAttribute.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNet.Mvc.ViewComponents
+{
+    [AttributeUsage(AttributeTargets.Property, AllowMultiple = false, Inherited = true)]
+    public class ViewComponentContextAttribute : Attribute
```
Eilon: Doc this one.

rynowak:
Updated - changed to defensive copy of `ViewContext` instead of a copy of just `ViewData`

src/Microsoft.AspNet.Mvc.Core/ViewComponents/ViewComponentContext.cs
```diff
@@ -29,32 +42,54 @@ public class ViewComponentContext
         {
             ViewComponentDescriptor = viewComponentDescriptor;
             Arguments = arguments;
-            ViewContext = viewContext;
-            Writer = writer;
+
+            // We want to create a defensive copy of the VDD here so that changes done in the VC
+            // aren't visible in the calling view.
```
dougbu: Nice. This behaviour is also consistent with tag helper activations, HTML helper recursions, and so on.

test/Microsoft.AspNet.Mvc.Core.Test/ViewComponentTests.cs
dougbu: It's back to being changed :dancers:

test/WebSites/ActivatorWebSite/Components/ValueComponent.cs
```diff
@@ -1,19 +0,0 @@
...
-
-using Microsoft.AspNet.Mvc;
-
-namespace ActivatorWebSite
-{
-    [ViewComponent(Name = "Value")]
-    public class ValueComponent : ViewComponent
-    {
-        [Activate]
-        public ViewService MyViewService { get; set; }
```
dougbu: Do we have tests elsewhere that test constructor injection into a view component. Fine to file a follow-up bug if not. Idea is to cover the general scenario of using some service or other in a view component.

rynowak: Yes. There's a preponderance of tests for this aspect of activation

dougbu: Nevermind. Just a couple of files further down...

dougbu:
:shipit:

PR #2593, opened 2015-05-21
Fix #2407 - Add back the implicit [Required] for value types

### LINKED ISSUE #2407: Add back the implicit [Required] for value types

rynowak:
This change adds a [Required] client validator when
ModelMetadata.IsRequired == true. The bulk of the changes here are
mechanical updates to test files.


test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/Validation/DataAnnotationsClientModelValidatorProviderTest.cs
```diff
@@ -14,12 +14,53 @@ public class DataAnnotationsClientModelValidatorProviderTest
...
+                nameof(DummyRequiredAttributeHelperClass.ValueTypeWithoutAttribute));
+
+            var providerContext = new ClientValidatorProviderContext(metadata);
+
+            // Act
+            provider.GetValidators(providerContext);
+
+            // Assert
+            var validator = Assert.Single(providerContext.Validators);
+            var adapter = Assert.IsType<RequiredAttributeAdapter>(validator);
```
ajaybhargavb: remove `var adapter = `

ajaybhargavb:
:shipit: after addressing one minor comment

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DataAnnotationsClientModelValidatorProvider.cs
```diff
@@ -32,14 +32,24 @@ internal delegate IClientModelValidator
...
                 if (_attributeFactories.TryGetValue(attribute.GetType(), out factory))
                 {
                     context.Validators.Add(factory(attribute));
                 }
             }
+
+            if (!hasRequiredAttribute && context.ModelMetadata.IsRequired)
+            {
+                // Add a default '[Required]' validator for generating HTML if necessary.
+                context.Validators.Add(new RequiredAttributeAdapter(new RequiredAttribute()));
```
dougbu: Agree with offline discussion. Should also add a `RequiredAttribute` to the validations that will be run on the server.

Issue #2581, opened 2015-05-19
Anchor and Form TagHelpers apply to all form and anchor elements.


NTaylorMullen:
Once https://github.com/aspnet/Razor/issues/361 is completed we will be able to restrict `<form>` and `<a>` `TagHelper`s to only target elements that happen to have server side attributes. This will enable the editor scenario of seeing all anchor tags being purple => only ones with a specified `TagHelper` bound property being purple.

NTaylorMullen:
e68986346176105db379da023eac245e0ad912bb

PR #2598, opened 2015-05-21
Add required attributes to AnchorTagHelper and FormTagHelper.

NTaylorMullen:
- This involved also adding required attributes with wildcards.
- With this change AnchorTagHelpers and FormTagHelpers should no longer light up on every `<form>` or `<a>` tag.

#2581


dougbu:
:shipit:

PR #2600, opened 2015-05-21
Fix #2151 - Part 4 remove [Activate] support from controllers.

### LINKED ISSUE #2151: Naming inconsistency between [Activate] and @inject

rynowak:
This change completely removes [Activate]. In a controller, you should
constructor injection or [FromServices] to access services.

To access context items (ActionContext, ActionBindingContext, root
ViewDataDictionary) you should use the respective attribute class.

We'd like to consider streamlining this further in the future by getting
down to a single injectable context for controllers, but for now this will
have to do.


src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
```diff
@@ -49,7 +49,6 @@ public ImageTagHelper(IHostingEnvironment hostingEnvironment, IMemoryCache cache
         [HtmlAttributeName(FileVersionAttributeName)]
         public bool FileVersion { get; set; }
 
-        [Activate]
```
dougbu: Isn't this part of the other PR?

test/Microsoft.AspNet.Mvc.Core.Test/DefaultControllerFactoryTest.cs
```diff
@@ -185,8 +184,10 @@ public void CreateController_ThrowsIfPropertyCannotBeActivated()
 
             // Act and Assert
             var exception = Assert.Throws<InvalidOperationException>(() => factory.CreateController(context));
-            Assert.Equal("The property 'Service' on controller '" + typeof(ControllerThatCannotBeActivated) +
-                        "' cannot be activated.", exception.Message);
+            Assert.Equal(
+                $"Unable to resolve service for type '{typeof(TestService)}' while attempting to activate " +
+                $"'{typeof(ControllerThatCannotBeActivated)}'.",
```
dougbu: Minor icky-ness. Could you change to be explicit and use `.FullName` while here please?

dougbu:
:shipit:

pranavkm:
:shipit:

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
rynowak: bingo.

test/Microsoft.AspNet.Mvc.Core.Test/DefaultControllerFactoryTest.cs
rynowak: Fixed locally.

PR #2613, opened 2015-05-27
Fix: Unable to use an incomplete `CacheProfile`

### LINKED ISSUE #2310: Unable to use an incomplete `CacheProfile`
Though `ResponseCacheFilter` has settable properties intended (apparently) to override and supplement the `CacheProfile` passed to its constructor, the constructor contains checks that prevent obvious scenarios.  For example:
``` c#
var partialCacheProfile = new CacheProfile
{
  Location = ResponseCacheLocation.Client,
};
options.Filters.Add(new ResponseCacheFilter(partialCacheProfile) { Duration = 24 });
```
will `throw` on the last line because the intentionally-incomplete `CacheProfile` is checked prior to the `Duration` setting.

The `ResponseCacheAttribute` works around this issue; it creates a complete `CacheProfile` and passes that to the `ResponseCacheFilter` constructor. So issues are specific to scenarios where users set up a default `ResponseCacheFilter` globally.

Suggest postponing the check until `OnActionExecuting()` is called (my preference), removing most `ResponseCacheFilter` setters, or correcting the following comment.
``` c#
// Duration MUST be set (either in the cache profile or in the attribute) unless NoStore is true.
```

CodingGorilla:
Updated the `ResponseCacheFilter` to defer checking of cache parameters until `OnActionExecuting`

Resolves #2310

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/Filters/ResponseCacheFilter.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Filters/ResponseCacheFilter.cs
@@ -14,6 +14,12 @@ namespace Microsoft.AspNet.Mvc
     /// </summary>
     public class ResponseCacheFilter : IActionFilter, IResponseCacheFilter
     {
+        private readonly CacheProfile _cacheProfile;
+        private int? _cacheDuration;
+        private ResponseCacheLocation? _cacheLocation;
+        private bool? _cacheNoStore;
+        private string _cacheVaryByHeader;
+
         /// <summary>
         /// Creates a new instance of <see cref="ResponseCacheFilter"/>
         /// </summary>
@@ -21,20 +27,7 @@ namespace Microsoft.AspNet.Mvc
         /// <see cref="ResponseCacheFilter"/>.</param>
         public ResponseCacheFilter(CacheProfile cacheProfile)
         {
-            if (!(cacheProfile.NoStore ?? false))
-            {
-                // Duration MUST be set (either in the cache profile or in the attribute) unless NoStore is true.
-                if (cacheProfile.Duration == null)
-                {
-                    throw new InvalidOperationException(
-                            Resources.FormatResponseCache_SpecifyDuration(nameof(NoStore), nameof(Duration)));
-                }
-            }
-
-            Duration = cacheProfile.Duration ?? 0;
-            Location = cacheProfile.Location ?? ResponseCacheLocation.Any;
-            NoStore = cacheProfile.NoStore ?? false;
-            VaryByHeader = cacheProfile.VaryByHeader;
+            _cacheProfile = cacheProfile;
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


rynowak:
Hi @CodingGorilla - product changes look good to me, but it would be great to see some tests that validate the new scenario supported by this change. 

Can you add a unit test to `ResponseCacheFilterTest` that will call `OnActionExecuting` after "completing" the cache profile? It would also be nice to see a functional test that does the same (`ResponseCacheTest` + `ResponseCacheWebSite`).

CodingGorilla:
@rynowak Added some unit tests (`ResponseCacheFilterTest`) that I think cover what you were asking - I was a little unclear what you were asking because the existing tests  would be verifying that the headers were being properly generated based on the `CacheProfile` that was being used.  So I wrote tests that verify that changing the individual properties directly on the `ResponseCacheFilter` itself indeed overrides the settings provided by the `CacheProfile`.  I hope that was what you were looking for.  

I'm working on the functional tests presently - again I believe here what you are looking for is a functional test that sets up a `ResponseCacheAttribute` and provides it a `CacheProfileName` and then overrides one of the cache profile settings with a property on the attribute.  Please let me know if I'm misunderstanding your request.

rynowak:
> I believe here what you are looking for is a functional test that sets up a ResponseCacheAttribute and provides it a CacheProfileName and then overrides one of the cache profile settings with a property on the attribute. Please let me know if I'm misunderstanding your request.

Exactly

CodingGorilla:
@rynowak Functional tests are complete now as well.

rynowak:
Wow! Super thorough. :shipit: - I'll merge this tomorrow.

src/Microsoft.AspNet.Mvc.Core/Filters/ResponseCacheFilter.cs
```diff
@@ -21,20 +27,7 @@ public class ResponseCacheFilter : IActionFilter, IResponseCacheFilter
         /// <see cref="ResponseCacheFilter"/>.</param>
         public ResponseCacheFilter(CacheProfile cacheProfile)
```
pranavkm: Unrelated to this PR, but should this be marked `[NotNull]`?

test/Microsoft.AspNet.Mvc.Core.Test/Filters/ResponseCacheAttributeTest.cs
```diff
@@ -172,12 +172,11 @@ public void CreateInstance_ThrowsWhenTheDurationIsNotSet_WithNoStoreFalse()
             var cacheProfiles = new Dictionary<string, CacheProfile>();
             cacheProfiles.Add("Test", new CacheProfile { NoStore = false });
 
-            // Act & Assert
-            var ex = Assert.Throws<InvalidOperationException>(
-                () => responseCache.CreateInstance(GetServiceProvider(cacheProfiles)));
-            Assert.Equal(
-                "If the 'NoStore' property is not set to true, 'Duration' property must be specified.",
-                ex.Message);
+            // Act
```
pranavkm: Indentation

rynowak: Fixed before merging

src/Microsoft.AspNet.Mvc.Core/Filters/ResponseCacheFilter.cs
rynowak: Fixed before merging.

rynowak:
70b56f157cc225b0c3668263189404068ad60eb9 merged thanks!

PR #2623, opened 2015-05-28
Removing creating ModelValidationNode by default in CompositeModelBin…

harshgMSFT:
…der.

This moves the responsibility of of MVN creation to individual model binders if they want the individual models to be validated.

Follow up to this would be cleaning up MVN class to have better (and less error prone constructors).

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/CancellationTokenModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/CancellationTokenModelBinder.cs
@@ -17,7 +17,13 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             if (bindingContext.ModelType == typeof(CancellationToken))
             {
                 var model = bindingContext.OperationBindingContext.HttpContext.RequestAborted;
-                return Task.FromResult(new ModelBindingResult(model, bindingContext.ModelName, isModelSet: true));
+                var validationNode =
+                    new ModelValidationNode(bindingContext.ModelName, bindingContext.ModelMetadata, model);
+                return Task.FromResult(new ModelBindingResult(
+                    model,
+                    bindingContext.ModelName,
+                    isModelSet: true,
+                    validationNode: validationNode));
             }
 
             return Task.FromResult<ModelBindingResult>(null);
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/FormCollectionModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/FormCollectionModelBinder.cs
@@ -46,7 +46,13 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
                 model = new FormCollection(new Dictionary<string, string[]>());
             }
 
-            return new ModelBindingResult(model, bindingContext.ModelName, isModelSet: true);
+            var validationNode =
+                 new ModelValidationNode(bindingContext.ModelName, bindingContext.ModelMetadata, model);
+            return new ModelBindingResult(
+                model,
+                bindingContext.ModelName,
+                isModelSet: true,
+                validationNode: validationNode);
         }
     }
 }
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
@@ -247,8 +247,8 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
                                 modelState.GetFieldValidationState(errorKey) == ModelValidationState.Unvalidated)
                             {
 
-                                // If we are not able to add a model error 
-                                // for instance when the max error count is reached, mark the model as skipped. 
+                                // If we are not able to add a model error
+                                // for instance when the max error count is reached, mark the model as skipped.
                                 modelState.MarkFieldSkipped(errorKey);
                             }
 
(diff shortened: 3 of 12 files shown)


harshgMSFT:
Fix line breaks all across.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CancellationTokenModelBinder.cs
```diff
@@ -17,7 +17,10 @@ public Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindingContex
             if (bindingContext.ModelType == typeof(CancellationToken))
             {
                 var model = bindingContext.OperationBindingContext.HttpContext.RequestAborted;
-                return Task.FromResult(new ModelBindingResult(model, bindingContext.ModelName, isModelSet: true));
+                var validationNode =
+                    new ModelValidationNode(bindingContext.ModelName, bindingContext.ModelMetadata, model);
+                return Task.FromResult(
+                    new ModelBindingResult(model, bindingContext.ModelName, isModelSet: true, validationNode: validationNode));
```
rynowak: still long line

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
```diff
@@ -37,8 +37,15 @@ public class DefaultObjectValidator : IObjectModelValidator
         /// <inheritdoc />
         public void Validate(
             [NotNull] ModelValidationContext modelValidationContext,
-            [NotNull] ModelValidationNode validationNode)
+            ModelValidationNode validationNode)
```
rynowak: Why not just skip the call to this API when there's nothing to validate.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CancellationTokenModelBinder.cs
harshgMSFT: ill clean up long lines, I somehow missed cleaning up before sending.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
harshgMSFT: sure

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CancellationTokenModelBinder.cs
```diff
@@ -17,7 +17,13 @@ public Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindingContex
             if (bindingContext.ModelType == typeof(CancellationToken))
             {
                 var model = bindingContext.OperationBindingContext.HttpContext.RequestAborted;
-                return Task.FromResult(new ModelBindingResult(model, bindingContext.ModelName, isModelSet: true));
+                var validationNode =
+                    new ModelValidationNode(bindingContext.ModelName, bindingContext.ModelMetadata, model);
```
rynowak: Why is there a validation node here? We don't want to validate a `CancellationToken`

src/Microsoft.AspNet.Mvc.Core/ModelBinding/FormCollectionModelBinder.cs
```diff
@@ -46,7 +46,13 @@ public class FormCollectionModelBinder : IModelBinder
                 model = new FormCollection(new Dictionary<string, string[]>());
             }
 
-            return new ModelBindingResult(model, bindingContext.ModelName, isModelSet: true);
+            var validationNode =
+                 new ModelValidationNode(bindingContext.ModelName, bindingContext.ModelMetadata, model);
```
rynowak: Same question here, I don't think we want to validate the `IFormCollection`

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CancellationTokenModelBinder.cs
harshgMSFT: Thats y It doesnt has a child node, this is basically for the case when you have a property for cancellation token and there is a [Required] on the property you want to add an error. 
Note this is consistent with what was before.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/FormCollectionModelBinder.cs
harshgMSFT: same as Cancellation token.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CancellationTokenModelBinder.cs
rynowak: Discussed in person, makes sense.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/KeyValuePairModelBinderTest.cs
```diff
@@ -309,7 +309,9 @@ private static IModelBinder CreateIntBinder()
                 {
                     if (mbc.ModelType == typeof(int))
                     {
-                        return Task.FromResult(new ModelBindingResult(42, mbc.ModelName, true));
+                        var model = 42;
+                        var validationNode = new ModelValidationNode(mbc.ModelName, mbc.ModelMetadata, model);
+                        return Task.FromResult(new ModelBindingResult(model, mbc.ModelName, true, validationNode));
```
dougbu: (it's test code but ...) probably worth sticking in the `isModelSet` argument name here and a couple of places below

test/Microsoft.AspNet.Mvc.FunctionalTests/ModelBindingTest.cs
```diff
@@ -1462,7 +1462,7 @@ public class ModelBindingTest
 
             };
             var url = "http://localhost/dealers/43/update-vehicle?dealer.name=TestCarDealer&dealer.location=NE";
-            
+
```
dougbu: :100: :dancers: :dango: !!

dougbu:
:shipit:

rynowak:
:shipit:

dougbu:
@harshgMSFT 4f419ee closes this, correct?

harshgMSFT:
Yes. Closed.

Issue #2470, opened 2015-04-28
Dictionary ModelBinding with an empty prefix has unvalidated keys


rynowak:
Trying to model bind the following:

```
"?[0].Key=key0&[0].Value=10"

public void DoThing(IDictionary<string, int> p)
{
}
```

Will result in `ModelState.IsValid == false`. The keys `[0].Key` and `[0].Value` are marked as unvalidated.

If you use a prefix in the query string, this issue doesn't occur. `"?p.[0].Key=key0&p.[0].Value=10"`

rynowak:
Very likely fixed as part of 22f1881cc641c390773783ba5a982e63a3b91464 - this no longer repros.

Enabled tests in 6e1226bfe46f75c894dfc162dc77435f0dbef208

PR #2624, opened 2015-05-28
Enable tests for #2470

rynowak:
This was very likely fixed as part of
22f1881cc641c390773783ba5a982e63a3b91464


rynowak:
FYI only - merging once it builds clean

Issue #2473, opened 2015-04-30
Composite Model Binder should not return a null result if a model binder returns a non-null result.


harshgMSFT:
Currently if any model binder returns a MBR like so
`new ModelBindingResult(model: null, isModelSet: false, key:"Key);`
then the composite model binder returns a null result 
https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/ModelBinding/CompositeModelBinder.cs#L120

The original intent of this code was to ensure we fallback to the empty model in
https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/ModelBinding/CompositeModelBinder.cs#L40

however there are better ways to achieve it ( for eg. rewriting the condition :smile: ).

dougbu:
@harshgMSFT why is this a bug?  what visible impact do the current choices have?  if there's an impact, suggest filing a bug about _that_.

we discussed the approach at length when making the changes you've highlighted and this was the best idea anyone came up with.  in particular changing the condition leads to a ripple of additional complexity in other parts of the system and unnecessary churn, ending up with no measurable improvement.

harshgMSFT:
>we discussed the approach at length when making the changes you've highlighted and this was the best idea anyone came up with. in particular changing the condition leads to a ripple of additional complexity in other parts of the system and unnecessary churn, ending up with no measurable improvement.

Well the only reason we (or at least I did) agreed that time was we wanted to move forward with the release in mind, having that line in the code is a hack and not a solution. And yes you are right I should have included the user impact. Sorry for the confusion and here is the detailed scenario : 

```c# 

public class Document
{
       [FromHeader]
       [Required]
       public string Version { get; set; }
}

public void Action(Document doc)
{
}
```

In this case if there is no header called `Version` we end up returning 
`new ModelBindingResult(model: null, isModelSet: false, key:"Key");`
which means a composite model binder chooses to return null from https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/ModelBinding/CompositeModelBinder.cs#L120

Now in this particular scenario the user visible problem is
We return null from https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/ModelBinding/CompositeModelBinder.cs#L52

[...]

Issue #2456, opened 2015-04-27
`byte[]` model binder should prevent falling back to other binders when the type matches.


harshgMSFT:
If the Model Type is `IFormFile` or `IEnumerable<IFormFile>`, even though we return a non null 
`ModelBindingResult`, in cases where the model is not set ( for example when there is no form body ), 
it will result into a ModelBindingResult of null returned to the caller. This is because of 
https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/ModelBinding/CompositeModelBinder.cs#L107

Ideally FormFileModelBinder should behave like a greedy model binder.

rynowak:
Can you include some sample code and an example of what the user-visible effect of this would be?

Eilon:
The main concern is that this model binder currently allows other model binders to run after it. The correct behavior of this model binder should be to always produce a result, even if it didn't bind anything (e.g. because there was no file or body).

rynowak:
This is also true of `byte[]` model binder. There are numerous cases where it doesn't handle a model of `byte[]`.

dougbu:
Updated title. `ByteArrayModelBinder` is the only type-matching binder that _ever_ returns `null` when the tested property is of the matched type. The rest of the binders work as expected. In particular all other tests skipped supposedly due to this bug had other issues.

dougbu:
e31eab0

PR #2625, opened 2015-05-29
Correct test expectations

dougbu:
### Correct test expectations
- use valid `multipart/form-data` content type; include a `boundary`
- correct expectations of `FormCollection` model binder's operation
 - restore tests actually skipped for either of the above reasons
- add more tests with `ModelBindingResult.Model==null` and both `IsModelSet` values

#### Remove test references to Won't Fix bug #2473
- restore #2473 tests; update test expectations
- expect `null` composite results whenever binding fails

#### Restore test skipped due to "#2646"
- that issue does not exist; was likely #2466 or similar fixed bug

#### Minor src changes
- remove unused variable and unnecessary nesting in `DefaultControllerActionArgumentBinder`
- remove dangling mention of `[DefaultValue]` in `ComplexModelDtoModelBinder`

#### nits:
- remove empty delegates from some `GetOperationBindingContext` calls; `null` fine
- do some `using` cleanup
- combine two test methods in `KeyValuePairModelBinderTest`
- name a few more arguments

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
@@ -81,24 +81,21 @@ namespace Microsoft.AspNet.Mvc
                 operationContext);
 
             var modelBindingResult = await operationContext.ModelBinder.BindModelAsync(modelBindingContext);
-            if (modelBindingResult != null && modelBindingResult.IsModelSet)
+            if (modelBindingResult != null &&
+                modelBindingResult.IsModelSet &&
+                modelBindingResult.ValidationNode != null)
             {
-                var key = modelBindingResult.Key;
                 var modelExplorer = new ModelExplorer(
                     _modelMetadataProvider,
                     metadata,
                     modelBindingResult.Model);
+                var validationContext = new ModelValidationContext(
+                    modelBindingContext.BindingSource,
+                    operationContext.ValidatorProvider,
+                    modelState,
+                    modelExplorer);
 
-                if (modelBindingResult.ValidationNode != null)
-                {
-                    var validationContext = new ModelValidationContext(
-                        modelBindingContext.BindingSource,
-                        operationContext.ValidatorProvider,
-                        modelState,
-                        modelExplorer);
-
-                    _validator.Validate(validationContext, modelBindingResult.ValidationNode);
-                }
+                _validator.Validate(validationContext, modelBindingResult.ValidationNode);
             }
 
             return modelBindingResult;
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/ComplexModelDtoModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/ComplexModelDtoModelBinder.cs
@@ -32,7 +32,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
                     await bindingContext.OperationBindingContext.ModelBinder.BindModelAsync(propertyBindingContext);
                 if (modelBindingResult == null)
                 {
-                    // Could not bind. Add a result so MutableObjectModelBinder will check for [DefaultValue].
+                    // Could not bind. Let MutableObjectModelBinder know explicitly.
                     dto.Results[propertyMetadata] =
                         new ModelBindingResult(model: null, key: propertyModelName, isModelSet: false);
                 }
--- a/test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/BindingSourceModelBinderTest.cs
+++ b/test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/BindingSourceModelBinderTest.cs
@@ -27,18 +27,18 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
 
             // Act & Assert
             var exception = Assert.Throws<ArgumentException>(
-                () => new TestableBindingSourceModelBinder(bindingSource));
+                () => new TestableBindingSourceModelBinder(bindingSource, isModelSet: false));
             Assert.Equal(expected, exception.Message);
         }
 
         [Fact]
-        public async Task BindingSourceModelBinder_ReturnsFalse_WithNoSource()
+        public async Task BindingSourceModelBinder_ReturnsNull_WithNoSource()
         {
             // Arrange
             var context = new ModelBindingContext();
             context.ModelMetadata = new EmptyModelMetadataProvider().GetMetadataForType(typeof(string));
 
-            var binder = new TestableBindingSourceModelBinder(BindingSource.Body);
+            var binder = new TestableBindingSourceModelBinder(BindingSource.Body, isModelSet: false);
 
             // Act
             var result = await binder.BindModelAsync(context);
@@ -49,7 +49,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         }
 
         [Fact]
-        public async Task BindingSourceModelBinder_ReturnsFalse_NonMatchingSource()
+        public async Task BindingSourceModelBinder_ReturnsNull_NonMatchingSource()
         {
             // Arrange
             var provider = new TestModelMetadataProvider();
@@ -58,7 +58,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             var context = new ModelBindingContext();
             context.ModelMetadata = provider.GetMetadataForType(typeof(string));
 
-            var binder = new TestableBindingSourceModelBinder(BindingSource.Body);
+            var binder = new TestableBindingSourceModelBinder(BindingSource.Body, isModelSet: false);
 
             // Act
             var result = await binder.BindModelAsync(context);
@@ -68,8 +68,10 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             Assert.False(binder.WasBindModelCoreCalled);
         }
 
-        [Fact]
-        public async Task BindingSourceModelBinder_ReturnsTrue_MatchingSource()
+        [Theory]
+        [InlineData(false)]
+        [InlineData(true)]
+        public async Task BindingSourceModelBinder_ReturnsNonNull_MatchingSource(bool isModelSet)
         {
             // Arrange
             var provider = new TestModelMetadataProvider();
@@ -82,30 +84,35 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
                 BinderModelName = modelMetadata.BinderModelName
             };
 
-            var binder = new TestableBindingSourceModelBinder(BindingSource.Body);
+            var binder = new TestableBindingSourceModelBinder(BindingSource.Body, isModelSet);
 
             // Act
             var result = await binder.BindModelAsync(context);
 
             // Assert
             Assert.NotNull(result);
-            Assert.False(result.IsModelSet);
+            Assert.Equal(isModelSet, result.IsModelSet);
+            Assert.Null(result.Model);
             Assert.True(binder.WasBindModelCoreCalled);
         }
 
         private class TestableBindingSourceModelBinder : BindingSourceModelBinder
         {
-            public bool WasBindModelCoreCalled { get; private set; }
+            bool _isModelSet;
 
-            public TestableBindingSourceModelBinder(BindingSource source)
+            public TestableBindingSourceModelBinder(BindingSource source, bool isModelSet)
                 : base(source)
             {
+                _isModelSet = isModelSet;
             }
+
+            public bool WasBindModelCoreCalled { get; private set; }
 
             protected override Task<ModelBindingResult> BindModelCoreAsync([NotNull] ModelBindingContext bindingContext)
             {
                 WasBindModelCoreCalled = true;
-                return Task.FromResult(new ModelBindingResult(model: null, key: bindingContext.ModelName, isModelSet: false));
+                return Task.FromResult(
+                    new ModelBindingResult(model: null, key: bindingContext.ModelName, isModelSet: _isModelSet));
             }
         }
     }
--- a/test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/BindingSourceModelBinderTest.cs
+++ b/test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/BindingSourceModelBinderTest.cs
@@ -40,7 +40,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
 
             var binder = new TestableBindingSourceModelBinder(BindingSource.Body);
 
-            // Act 
+            // Act
             var result = await binder.BindModelAsync(context);
 
             // Assert
@@ -60,7 +60,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
 
             var binder = new TestableBindingSourceModelBinder(BindingSource.Body);
 
-            // Act 
+            // Act
             var result = await binder.BindModelAsync(context);
 
             // Assert
@@ -84,12 +84,12 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
 
             var binder = new TestableBindingSourceModelBinder(BindingSource.Body);
 
-            // Act 
+            // Act
             var result = await binder.BindModelAsync(context);
 
             // Assert
             Assert.NotNull(result);
-            Assert.True(result.IsModelSet);
+            Assert.False(result.IsModelSet);
             Assert.True(binder.WasBindModelCoreCalled);
         }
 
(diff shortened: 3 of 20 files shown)


test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/BindingSourceModelBinderTest.cs
```diff
@@ -84,12 +84,12 @@ public void BindingSourceModelBinder_ThrowsOnNonGreedySource()
...
 
-            // Act 
+            // Act
             var result = await binder.BindModelAsync(context);
 
             // Assert
             Assert.NotNull(result);
-            Assert.True(result.IsModelSet);
+            Assert.False(result.IsModelSet);
             Assert.True(binder.WasBindModelCoreCalled);
```
rynowak: While you're doing cleanup here can you rename these tests to be more accurate? `BindingSourceModelBinder_ReturnsTrue_MatchingSource` `BindingSourceModelBinder_ReturnsFalse_NonMatchingSource`

They don't really return true or false

dougbu: :ok:

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/CompositeModelBinderTest.cs
```diff
@@ -445,6 +445,7 @@ public class CompositeModelBinderTest
                 .Returns(
                     delegate (ModelBindingContext context)
                     {
+                        // Note this ModelBindingResult is not valid.
```
rynowak: It's not really invalid, the model value will just be ignored.

test/Microsoft.AspNet.Mvc.Core.Test/ParameterBinding/ControllerActionArgumentBinderTests.cs
```diff
@@ -361,7 +361,7 @@ public class ControllerActionArgumentBinderTests
             binder
                 .Setup(b => b.BindModelAsync(It.IsAny<ModelBindingContext>()))
                 .Returns(Task.FromResult(
-                    result: new ModelBindingResult(model: null, key: string.Empty, isModelSet: true)));
+                    result: new ModelBindingResult(model: null, key: string.Empty, isModelSet: false)));
```
rynowak: This really changes the meaning of this test.

test/Microsoft.AspNet.Mvc.IntegrationTests/BinderTypeBasedModelBinderIntegrationTest.cs
```diff
@@ -12,89 +12,11 @@ namespace Microsoft.AspNet.Mvc.IntegrationTests
 {
     public class BinderTypeBasedModelBinderIntegrationTest
     {
-        // The NullModelBinder and NullModelNotSetModelBinder return a non null ModelBindingResult but a null model.
-        [Theory(Skip = "ModelBindingResult should be non null if a model binder returns a non null resul #2473.")]
-        [InlineData(typeof(NullModelBinder), true)]
-        [InlineData(typeof(NullModelNotSetModelBinder), false)]
-        public async Task BindParameter_WithModelBinderType_NoData(Type modelBinderType, bool isModelSet)
```
rynowak: Are these two tests dupes now?

test/Microsoft.AspNet.Mvc.Core.Test/ParameterBinding/ControllerActionArgumentBinderTests.cs
dougbu: Not really much at all. Note I didn't change the assertions at all. My thought was to avoid testing a corner case in this general test. Few if any model binders return a result with `Model == null && IsModelSet`. But I can revert this one-liner change or rename the test if you prefer.

test/Microsoft.AspNet.Mvc.IntegrationTests/ModelBindingTestHelper.cs
```diff
@@ -67,6 +69,21 @@ public static IObjectModelValidator GetObjectValidator(MvcOptions options = null
...
+            {
+                if (typeof(TModelBinder) == options.ModelBinders[index].GetType())
+                {
+                    options.ModelBinders.Insert(index + 1, new AssertIfCalledModelBinder());
+                    return;
+                }
+            }
+
+            Assert.False(true, $"Unable to find { typeof(TModelBinder).FullName } in { nameof(MvcOptions) }.");
+        }
```
rynowak: I'm really not a fan of this in the integration tests. Why is this needed? If these tests are failing then let them fail.

test/Microsoft.AspNet.Mvc.IntegrationTests/BinderTypeBasedModelBinderIntegrationTest.cs
dougbu: Mostly irrelevant since the `CompositeModelBinder` should return `null` result in this case. I could change the assertions if we care about the scenario. Looked unimportant to me but I'm fine either way.

test/Microsoft.AspNet.Mvc.Core.Test/ParameterBinding/ControllerActionArgumentBinderTests.cs
rynowak: See `DefaultControllerActionArgumentBinder`@130 - there's code in place to specifically handle this scenario, and this is a test written specifically to target that code.

test/Microsoft.AspNet.Mvc.IntegrationTests/ModelBindingTestHelper.cs
dougbu: This meaning the assertion? It's a guard to ensure the setup does what the test expects.

test/Microsoft.AspNet.Mvc.IntegrationTests/BinderTypeBasedModelBinderIntegrationTest.cs
rynowak: I think it would be preferrable to change the expectations to match the behavior then. It sounds like removing these tests would mean that we just don't have coverage for the scenario.

test/Microsoft.AspNet.Mvc.IntegrationTests/ModelBindingTestHelper.cs
rynowak: I mean the whole thing `UpdateOptionsToEnsureNothingFollows` does. Why is it needed? 

If the goal is to verify the behavior of an individual model binder (returning null/not-null), we should have unit tests for that.

test/Microsoft.AspNet.Mvc.Core.Test/ParameterBinding/ControllerActionArgumentBinderTests.cs
dougbu: Then the test should confirm correct behaviour when `isModelSet` is _either_ `true` or `false`.

test/Microsoft.AspNet.Mvc.IntegrationTests/BinderTypeBasedModelBinderIntegrationTest.cs
dougbu: :ok:

test/Microsoft.AspNet.Mvc.IntegrationTests/ModelBindingTestHelper.cs
dougbu: The goal is to verify the system pays attention to what the individual model binders return.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/CompositeModelBinderTest.cs
```diff
@@ -291,7 +291,7 @@ public class CompositeModelBinderTest
             var modelBinder = new Mock<IModelBinder>();
             modelBinder
                 .Setup(mb => mb.BindModelAsync(It.IsAny<ModelBindingContext>()))
-                .Returns(Task.FromResult(new ModelBindingResult(null, "someName", true)));
+                .Returns(Task.FromResult(new ModelBindingResult(model: "some value", key: "someName", isModelSet: true)));
```
harshgMSFT: The test says, sets null value and sets the model state key.. why are we chaing it to a non null value?

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/CompositeModelBinderTest.cs
harshgMSFT: With my latest changes, this test does not make sense. 
IMO lets replace it with `BindModel_DoesNotAddAValidationNode_IfModelBindingResultDoesNotContainAValidationNode`
PS: shorten the name if you want to.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/FormValueProviderFactoryTests.cs
```diff
@@ -33,8 +33,8 @@ public void GetValueProvider_ReturnsNull_WhenContentTypeIsNotFormUrlEncoded()
         [Theory]
         [InlineData("application/x-www-form-urlencoded")]
         [InlineData("application/x-www-form-urlencoded;charset=utf-8")]
-        [InlineData("multipart/form-data")]
-        [InlineData("multipart/form-data;charset=utf-8")]
+        [InlineData("multipart/form-data; boundary=----WebKitFormBoundarymx2fSWqWSd0OxQqq")]
+        [InlineData("multipart/form-data; boundary=----WebKitFormBoundarymx2fSWqWSd0OxQqq; charset=utf-8")]
```
harshgMSFT: :+1:

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/KeyValuePairModelBinderTest.cs
```diff
@@ -139,14 +139,13 @@ public class KeyValuePairModelBinderTest
             var innerBinder = new Mock<IModelBinder>();
             innerBinder
                 .Setup(o => o.BindModelAsync(It.IsAny<ModelBindingContext>()))
-                .Returns((ModelBindingContext mbc) =>
+                .Returns((ModelBindingContext context) =>
                 {
-                    Assert.Equal("someName.key", mbc.ModelName);
-                    return Task.FromResult(new ModelBindingResult(null, string.Empty, true));
+                    Assert.Equal("someName.key", context.ModelName);
+                    return Task.FromResult(new ModelBindingResult(model: "not int", key: string.Empty, isModelSet: true));
```
harshgMSFT: line break here.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/BindingSourceModelBinderTest.cs
```diff
@@ -105,7 +105,7 @@ public TestableBindingSourceModelBinder(BindingSource source)
             protected override Task<ModelBindingResult> BindModelCoreAsync([NotNull] ModelBindingContext bindingContext)
             {
                 WasBindModelCoreCalled = true;
-                return Task.FromResult(new ModelBindingResult(null, bindingContext.ModelName, true));
+                return Task.FromResult(new ModelBindingResult(model: null, key: bindingContext.ModelName, isModelSet: false));
```
harshgMSFT: All the changes in this class seem to be cosmetic .. have we made sure that none of the other tests depend on a behavior like model is null and IsModelSet is true?

test/Microsoft.AspNet.Mvc.IntegrationTests/BinderTypeBasedModelBinderIntegrationTest.cs
harshgMSFT: :+1: with @rynowak I would rather have tests that ensure the behavior.

test/Microsoft.AspNet.Mvc.IntegrationTests/FormFileModelBindingIntegrationTest.cs
```diff
@@ -154,9 +154,8 @@ private class Address
 
             // Assert
 
-            // ModelBindingResult
-            Assert.NotNull(modelBindingResult); // Fails due to bug #2456
-            Assert.Null(modelBindingResult.Model);
+            // ModelBindingResult; expect null because binding failed.
+            Assert.Null(modelBindingResult);
```
harshgMSFT: This is wrong, the FormFileModelBinder should return non null MBR, because the type is IFormFile. That is what bug #2456 was about. Because of the quirkiness of the CompositeModelBinder, the result is null. Lets discuss this.

dougbu: Will make sure the test name matches the expected behaviour.

harshgMSFT: Thanks, discussed in person what is here makes sense.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/BindingSourceModelBinderTest.cs
dougbu: Not exactly cosmetic. I should have tested `isModelSet: false` with `model: null` as well as `isModelSet: true`. Instead I just switched `isModelSet` here. Will make sure both are tested.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/CompositeModelBinderTest.cs
dougbu: Test makes sense since no `ModelValidationNode` with `isModelSet: false` can definitely happen. Or am I missing something?

dougbu:
Rebased and updated
- also changed PR description to match latest overall changes
- note additional fixes e.g. restoration of test skipped due to non-existent "#2646" issue and src nits

test/Microsoft.AspNet.Mvc.IntegrationTests/BinderTypeBasedModelBinderIntegrationTest.cs
```diff
@@ -87,14 +110,14 @@ private class Person2
...
+            Assert.Null(modelBindingResult);
 
             // ModelState
             Assert.True(modelState.IsValid);
             Assert.Empty(modelState.Keys);
         }
-        
-        // ModelBinderAttribute can be used without specifing the binder type. 
+
+        // ModelBinderAttribute can be used without specifying the binder type.
```
harshgMSFT: :dancers:

src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
```diff
@@ -81,24 +81,21 @@ public class DefaultControllerActionArgumentBinder : IControllerActionArgumentBi
...
                     _modelMetadataProvider,
                     metadata,
                     modelBindingResult.Model);
+                var validationContext = new ModelValidationContext(
+                    modelBindingContext.BindingSource,
+                    operationContext.ValidatorProvider,
+                    modelState,
+                    modelExplorer);
 
-                if (modelBindingResult.ValidationNode != null)
```
harshgMSFT: why is this removed? ValidationNode can be null if a model binder chooses not to use a MVN simply because it does not want to validate the model. I expected some tests to break, are they breaking?

harshgMSFT:
:watch:

src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
rynowak: It was moved up into the surrounding `if`

src/Microsoft.AspNet.Mvc.Core/ModelBinding/ComplexModelDtoModelBinder.cs
```diff
@@ -32,7 +32,7 @@ public sealed class ComplexModelDtoModelBinder : IModelBinder
                     await bindingContext.OperationBindingContext.ModelBinder.BindModelAsync(propertyBindingContext);
                 if (modelBindingResult == null)
                 {
-                    // Could not bind. Add a result so MutableObjectModelBinder will check for [DefaultValue].
+                    // Could not bind. Let MutableObjectModelBinder know explicitly.
```
rynowak: Not suggesting to add more to this PR, but the comment suggests this block is unnecessary. IMO correcting the comment is the right thing to do for now though :100:

src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
harshgMSFT: my bad. Please ignore my comment.

harshgMSFT:
:shipit:

rynowak:
:shipit:

dougbu:
26795bd

PR #2626, opened 2015-05-29
`ByteArrayModelBinder` should return `null` only when type is not matched

dougbu:
- #2456
- visible behaviours now match MVC 5
- prevent `CollectionModelBinder` from converting no value to `byte[0]`
- prevent `TypeConverterModelBinder` from converting empty value to `byte[] { '\0' }`

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/ByteArrayModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/ByteArrayModelBinder.cs
@@ -15,6 +15,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         /// <inheritdoc />
         public async Task<ModelBindingResult> BindModelAsync([NotNull] ModelBindingContext bindingContext)
         {
+            // Check if this binder applies.
             if (bindingContext.ModelType != typeof(byte[]))
             {
                 return null;
@@ -22,25 +23,26 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
 
             var valueProviderResult = await bindingContext.ValueProvider.GetValueAsync(bindingContext.ModelName);
 
-            // case 1: there was no <input ... /> element containing this data
+            // Check for missing data case 1: There was no <input ... /> element containing this data.
             if (valueProviderResult == null)
             {
-                return null;
+                return new ModelBindingResult(model: null, key: bindingContext.ModelName, isModelSet: false);
             }
 
             var value = valueProviderResult.AttemptedValue;
 
-            // case 2: there was an <input ... /> element but it was left blank
+            // Check for missing data case 2: There was an <input ... /> element but it was left blank.
             if (string.IsNullOrEmpty(value))
             {
-                return null;
+                return new ModelBindingResult(model: null, key: bindingContext.ModelName, isModelSet: false);
             }
 
             try
             {
                 var model = Convert.FromBase64String(value);
 
-                // We do not need to set an explict ModelValidationNode since CompositeModelBinder does that automatically.
+                // We do not need to set an explicit ModelValidationNode since CompositeModelBinder does that
+                // automatically.
                 return new ModelBindingResult(
                     model,
                     bindingContext.ModelName,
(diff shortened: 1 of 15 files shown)


src/Microsoft.AspNet.Mvc.Core/ModelBinding/ByteArrayModelBinder.cs
```diff
@@ -22,25 +23,26 @@ public class ByteArrayModelBinder : IModelBinder
...
+                return new ModelBindingResult(model: null, key: bindingContext.ModelName, isModelSet: false);
             }
 
             try
             {
                 var model = Convert.FromBase64String(value);
 
-                // We do not need to set an explict ModelValidationNode since CompositeModelBinder does that automatically.
+                // We do not need to set an explicit ModelValidationNode since CompositeModelBinder does that
+                // automatically.
```
rynowak: Why not just do the right thing here? Harsh's PR is already in flight.

test/Microsoft.AspNet.Mvc.FunctionalTests/ModelBindingTest.cs
```diff
@@ -778,7 +778,7 @@ public class ModelBindingTest
 
             //Assert
             Assert.Equal(HttpStatusCode.OK, response.StatusCode);
-            Assert.Equal("\0", await response.Content.ReadAsStringAsync());
+            Assert.Equal(string.Empty, await response.Content.ReadAsStringAsync());
```
rynowak: :clap:

test/Microsoft.AspNet.Mvc.IntegrationTests/ByteArrayModelBinderIntegrationTest.cs
```diff
@@ -92,8 +92,9 @@ private class Person
                 ParameterType = typeof(byte[])
             };
 
-            // No data is passed.
-            var operationContext = ModelBindingTestHelper.GetOperationBindingContext(httpContext => { });
+            // No data is passed. The ByteArrayModelBinder should ensure no later binders run.
+            var operationContext = ModelBindingTestHelper.GetOperationBindingContext(
+                updateOptions: ModelBindingTestHelper.UpdateOptionsToEnsureNothingFollows<ByteArrayModelBinder>);
             var modelState = new ModelStateDictionary();
```
rynowak: I don't see this method as part of the PR. Also, the purpose of these tests is to test what model binding does with the default/typical configuration. If we need to do esoteric things to these tests it's probably just better suited for a unit test.

rynowak:
:watch:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/ByteArrayModelBinder.cs
dougbu: _because_ @harshgMSFT's PR is in flight, I'm leaving cleaning this up to him.

test/Microsoft.AspNet.Mvc.IntegrationTests/ByteArrayModelBinderIntegrationTest.cs
dougbu: The method is in the PR this is based on -- #2625. The purpose of the method is to confirm everything works as-expected, not to change the overall operation. Have a look at the other PR.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/ByteArrayModelBinder.cs
harshgMSFT: this should already be removed ..

src/Microsoft.AspNet.Mvc.Core/ModelBinding/ByteArrayModelBinder.cs
```diff
@@ -22,25 +23,26 @@ public class ByteArrayModelBinder : IModelBinder
...
             }
 
             var value = valueProviderResult.AttemptedValue;
 
-            // case 2: there was an <input ... /> element but it was left blank
+            // Check for missing data case 2: There was an <input ... /> element but it was left blank.
             if (string.IsNullOrEmpty(value))
             {
-                return null;
+                return new ModelBindingResult(model: null, key: bindingContext.ModelName, isModelSet: false);
```
harshgMSFT: This would mean that we would never be validating a controller property of type byte[] when there is no data. 
This makes it inconsistent with things like FromFileModelBinder. 
https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/ModelBinding/FormFileModelBinder.cs

@dougbu can we open a tracking issue, we can triage and come to a conclusion on this.

harshgMSFT:
open an issue and then :shipit:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/ByteArrayModelBinder.cs
dougbu: #2633

dougbu:
Closing since this PR can be based on `dev` since we aren't doing the `ModelBindingTestHelper.UpdateOptionsToEnsureNothingFollows<TModelBinder>()` thing. Will open a new PR in a bit...

PR #2644, opened 2015-06-02
Event Notification for MVC Prototype

rynowak:
Here's a first take on the pattern for publishing notifications from MVC.


rynowak:
@Eilon reviewed in person. Will check in once coherence is clean.

PR #2645, opened 2015-06-02
Added LanguageViewLocationExpander into framework 

kirthik:
cc @rynowak


src/Microsoft.AspNet.Mvc.Razor/LanguageViewLocationExpander.cs
```diff
@@ -62,7 +59,6 @@ public void PopulateValues(ViewLocationExpanderContext context)
             foreach (var location in viewLocations)
             {
                 yield return location.Replace("{0}", value + "/{0}");
-                yield return location;
```
pranavkm: Removing the original is a bit problematic though - you'll never fall back to Home/Index.cshtml if Home/en-us/Index.cshtml is available.

samples/MvcSample.Web/Startup.cs
```diff
@@ -63,12 +63,6 @@ public IServiceProvider ConfigureServices(IServiceCollection services)
                 diSystem.Equals("AutoFac", StringComparison.OrdinalIgnoreCase))
             {
                 _autoFac = true;
-                services.ConfigureRazorViewEngine(options =>
-                {
-                    var expander = new LanguageViewLocationExpander(
-                        context => context.HttpContext.Request.Query["language"]);
-                    options.ViewLocationExpanders.Insert(0, expander);
-                });
```
rynowak: There should still be some kind of equivalent gesture for enabling this feature or at least for enabling MVC localization features.

src/Microsoft.AspNet.Mvc.Razor/LanguageViewLocationExpander.cs
```diff
@@ -3,13 +3,14 @@
...
+using System.Globalization;
+using System.Linq;
+using System.Threading;
 
-namespace RazorWebSite
+namespace Microsoft.AspNet.Mvc.Razor
 {
     /// <summary>
-    /// A <see cref="IViewLocationExpander"/> that replaces adds the language as an extension prefix to view names.
+    /// A <see cref="IViewLocationExpander"/> that adds the language as an extension prefix to view names.
```
rynowak: Might want to also mention that the 'language' in practice is coming from the `HttpContext`

src/Microsoft.AspNet.Mvc.Razor/LanguageViewLocationExpander.cs
```diff
@@ -21,36 +22,32 @@ namespace RazorWebSite
...
+            var cultureInfo = Thread.CurrentThread.CurrentUICulture;
+#else
+            var cultureInfo = CultureInfo.CurrentUICulture;
+#endif
+
+            var updatedViewLocations = new List<string>();
+            while (cultureInfo != cultureInfo.Parent)
+            {
+                updatedViewLocations.AddRange(ExpandViewLocationsCore(viewLocations, cultureInfo.Name));
+                cultureInfo = cultureInfo.Parent;
```
rynowak: From MSDN:
> The cultures have a hierarchy in which the parent of a specific culture is a neutral culture, the parent of a neutral culture is the InvariantCulture, and the parent of the InvariantCulture is the invariant culture itself. The parent culture encompasses only the set of information that is common among its children.

Should we be checking `cultureInfo.IsNeutralCulture`? as our stop condition?

test/Microsoft.AspNet.Mvc.FunctionalTests/InlineConstraintTests.cs
```diff
@@ -49,7 +49,10 @@ public class InlineConstraintTests
             // Assert
             var exception = response.GetServerException();
             Assert.Equal("The view 'Index' was not found." +
-                         " The following locations were searched:__/Areas/Users/Views/Home/Index.cshtml__" +
+                         " The following locations were searched:__/Areas/Users/Views/Home/en-GB/Index.cshtml__" +
+                         "/Areas/Users/Views/Shared/en-GB/Index.cshtml__/Views/Shared/en-GB/Index.cshtml__" +
+                         "/Areas/Users/Views/Home/en/Index.cshtml__/Areas/Users/Views/Shared/en/Index.cshtml__" +
+                         "/Views/Shared/en/Index.cshtml__/Areas/Users/Views/Home/Index.cshtml__" +
                          "/Areas/Users/Views/Shared/Index.cshtml__/Views/Shared/Index.cshtml.",
```
rynowak: We shouldn't make this the default, you should have to opt-in to it.

src/Microsoft.AspNet.Mvc.Razor/LanguageViewLocationExpander.cs
```diff
@@ -21,36 +22,32 @@ namespace RazorWebSite
...
         }
 
         /// <inheritdoc />
         public virtual IEnumerable<string> ExpandViewLocations(ViewLocationExpanderContext context,
                                                                IEnumerable<string> viewLocations)
         {
-            string value;
-            if (context.Values.TryGetValue(ValueKey, out value))
+#if DNX451
+            var cultureInfo = Thread.CurrentThread.CurrentUICulture;
```
rynowak: Has there been much discussion yet about whether we want to use `CurrentUICulture` or `CurrentCulture` for this? This feature really blurs the lines here between the two - but IMO is closer to `CurrentUICulture`. I wanted to ask the question in case it hadn't been asked.

If there's a decision here, it's worth documenting for posterity IMO.

test/Microsoft.AspNet.Mvc.FunctionalTests/ViewEngineTests.cs
```diff
@@ -288,7 +292,7 @@ public static IEnumerable<object[]> RazorViewEngine_UsesExpandersForLayoutsData
 View With Layout
 </language-layout>";
 
-                yield return new[] { "gb", expected1 };
+                yield return new[] { "en-GB", expected1 };
```
rynowak: Is this intended?

src/Microsoft.AspNet.Mvc.Razor/LanguageViewLocationExpander.cs
```diff
@@ -21,36 +22,32 @@ namespace RazorWebSite
     /// </example>
     public class LanguageViewLocationExpander : IViewLocationExpander
```
rynowak: Needs unit tests

rynowak:
:watch:

src/Microsoft.AspNet.Mvc.Razor/LanguageViewLocationExpander.cs
kirthik: Yes. @DamianEdwards suggested we use CurrentUICulture. I will document it.

src/Microsoft.AspNet.Mvc.Razor/LanguageViewLocationExpander.cs
kirthik: will do.

test/Microsoft.AspNet.Mvc.FunctionalTests/InlineConstraintTests.cs
kirthik: Makes sense. Will change it.

test/Microsoft.AspNet.Mvc.FunctionalTests/ViewEngineTests.cs
kirthik: Yes. `CultureInfo` does not recognize `gb`. So changed it to valid culture name.

src/Microsoft.AspNet.Mvc.Razor/LanguageViewLocationExpander.cs
kirthik: yes, it will result in an entry for neutral culture. If I check `cultureInfo.IsNeutralCulture` as stopping condition, we won't add the neutral culture which we want to, as a fall back in case specific culture does not exist.

src/Microsoft.AspNet.Mvc.Razor/LanguageViewLocationExpander.cs
kirthik: I am adding it in [Line 49](https://github.com/aspnet/Mvc/pull/2645/files#diff-13c0456c912173b92459976512739c7eR49). I had to remove it from here since it is getting added multiple times.

pranavkm: One of the reasons this expander uses a `yield return` is because you could chain multiple expanders together and the number of strings that you generate will grow very quickly. In general, the expander should avoid creating strings until they're actually iterated upon.

Additionally appending the original list at the end might not be the correct behavior - for instance we should prefer `Views/Home/_Partial.cshtml` to `Views/Shared/{locale}/_Partial.cshtml`

src/Microsoft.AspNet.Mvc.Razor/LanguageViewLocationExpander.cs
```diff
@@ -21,36 +22,32 @@ namespace RazorWebSite
...
-        /// <param name="valueFactory">A factory that provides tbe language to use for expansion.</param>
-        public LanguageViewLocationExpander(Func<ActionContext, string> valueFactory)
-        {
-            _valueFactory = valueFactory;
-        }
-
         /// <inheritdoc />
         public void PopulateValues(ViewLocationExpanderContext context)
         {
-            var value = _valueFactory(context.ActionContext);
```
pranavkm: This needs to put the `CurrentUICulture` in context.Values under some well known key.

samples/MvcSample.Web/LanguageViewLocationExpander.cs
```diff
@@ -1,69 +0,0 @@
-// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-using System;
```
pranavkm: There's a copy of this expander in one of the functional tests too. Might want to remove that.

test/Microsoft.AspNet.Mvc.FunctionalTests/ViewEngineTests.cs
rynowak: Well should it be `CultureInfo.TwoLetterISOLanguageName` vs `CultureInfo.Name`? `.Name` generally refers to a locale, while `.TwoLetterISOLanguageName` refers to a language

kirthik: We want it to be locale specific right? For instance date format will be different in `en-US` and `en-GB` but `CultureInfo.TwoLetterISOLanguageName` for both will be `en`

samples/MvcSample.Web/LanguageViewLocationExpander.cs
kirthik: I removed that too. Its in `RazorWebsite`

src/Microsoft.AspNet.Mvc.Razor/LanguageViewLocationExpander.cs
kirthik: Talked offline. Understood why its needed. I will add it.

test/Microsoft.AspNet.Mvc.FunctionalTests/ViewEngineTests.cs
rynowak: Yep. Sounds right, just asking for confirmation

src/Microsoft.AspNet.Mvc/MvcServiceCollectionExtensions.cs
```diff
@@ -254,6 +254,16 @@ public static IServiceCollection AddMvc([NotNull] this IServiceCollection servic
             return WithControllersAsServices(services, controllerTypes.Select(type => type.AsType()));
         }
 
+        public static IServiceCollection AddMvcLocalization([NotNull] this IServiceCollection services)
```
kirthik: This is temporary. I will move this method to Localization package once I create one.

samples/MvcSample.Web/Startup.cs
rynowak: yay!

test/Microsoft.AspNet.Mvc.Razor.Test/LanguageViewLocationExpanderTest.cs
```diff
@@ -0,0 +1,140 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Xunit;
+
+namespace Microsoft.AspNet.Mvc.Razor.Test
```
rynowak: `Microsoft.AspNet.Mvc.Razor`

rynowak:
:shipit: from me. Please get signoff from @pranavkm as well

src/Microsoft.AspNet.Mvc.Razor/LanguageViewLocationExpander.cs
```diff
@@ -0,0 +1,79 @@
...
+            // Using CurrentUICulture so it loads the locale specific resources for the views.
+#if DNX451
+            context.Values[ValueKey] = Thread.CurrentThread.CurrentUICulture.Name;
+#else
+            context.Values[ValueKey] = CultureInfo.CurrentUICulture.Name;
+#endif
+        }
+
+        /// <inheritdoc />
+        public virtual IEnumerable<string> ExpandViewLocations([NotNull] ViewLocationExpanderContext context,
```
pranavkm: Formatting: Can you have `context` on the next line using VS's default indentation?

src/Microsoft.AspNet.Mvc.Razor/LanguageViewLocationExpander.cs
```diff
@@ -0,0 +1,79 @@
...
+                                                               [NotNull] IEnumerable<string> viewLocations)
+        {
+            string value;
+            context.Values.TryGetValue(ValueKey, out value);
+
+            try
+            {
+                if (!string.IsNullOrEmpty(value))
```
pranavkm: I'll leave this up to you, but it's nicer to limit the scope of try-catch be as specific as possible. In this case, this would be:

```C#
CultureInfo culture;
try
{
   culture = new CultureInfo(value);
}
catch (CultureNotFoundException)
{
   return viewLocations;
}

return ExpandViewLocationsCore(viewLocations, culture);
```

src/Microsoft.AspNet.Mvc.Razor/LanguageViewLocationExpander.cs
```diff
@@ -0,0 +1,79 @@
...
+                {
+                    return ExpandViewLocationsCore(viewLocations, new CultureInfo(value));
+                }
+            }
+            catch(CultureNotFoundException)
```
pranavkm: Ctrl K - D

src/Microsoft.AspNet.Mvc/MvcServiceCollectionExtensions.cs
```diff
@@ -254,6 +254,16 @@ public static IServiceCollection AddMvc([NotNull] this IServiceCollection servic
...
+        {
+            services.ConfigureRazorViewEngine(options =>
+            {
+                options.ViewLocationExpanders.Insert(0, new LanguageViewLocationExpander());
```
pranavkm: Should this just `Add` \ `Append` instead of being first? This way if the user adds a expander, and calls this, their ordering is preserved.

pranavkm:

PR #2647, opened 2015-06-03
React to Session api review changes

kichalla:
@Tratcher @ajaybhargavb

### DIFF
--- a/samples/MvcSample.Web/Startup.cs
+++ b/samples/MvcSample.Web/Startup.cs
@@ -100,7 +100,7 @@ namespace MvcSample.Web
 #endif
             app.UseRequestLocalization();
 
-            app.UseInMemorySession();
+            app.UseSession();
             app.UseMvc(routes =>
             {
                 routes.MapRoute("areaRoute", "{area:exists}/{controller}/{action}");
--- a/src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
@@ -61,11 +61,16 @@ namespace Microsoft.AspNet.Mvc
                 return null;
             }
 
-            var tempDataDictionary = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
             var session = context.Session;
+            if (session == null)
+            {
+                return null;
+            }
+
+            var tempDataDictionary = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
             byte[] value;
 
-            if (session != null && session.TryGetValue(TempDataSessionStateKey, out value))
+            if (session.TryGetValue(TempDataSessionStateKey, out value))
             {
                 using (var memoryStream = new MemoryStream(value))
                 using (var writer = new BsonReader(memoryStream))
@@ -139,7 +144,7 @@ namespace Microsoft.AspNet.Mvc
             {
                 // Since we call Save() after the response has been sent, we need to initialize an empty session
                 // so that it is established before the headers are sent.
-                session[TempDataSessionStateKey] = new byte[] { };
+                session.Set(TempDataSessionStateKey, new byte[] { });
             }
 
             return tempDataDictionary;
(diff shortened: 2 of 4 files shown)


samples/MvcSample.Web/Startup.cs
```diff
@@ -103,7 +103,7 @@ public void Configure(IApplicationBuilder app)
             }
 #endif
 
-            app.UseInMemorySession();
+            app.UseSession();
```
kichalla: `UseInMemorySession` extension is going away and also the default is in-memory session.

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
```diff
@@ -61,11 +61,16 @@ public class SessionStateTempDataProvider : ITempDataProvider
                 return null;
             }
 
-            var tempDataDictionary = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
             var session = context.Session;
+            if (session == null)
```
kichalla: This change because we were not checking for null in the `else` case. This was uncovered when I tried simplifying the test case httpcontext setup.

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
```diff
@@ -61,11 +61,16 @@ public class SessionStateTempDataProvider : ITempDataProvider
...
+            {
+                return null;
```
ajaybhargavb: Why return `null` here instead of an empty dictionary?

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
Tratcher: Are you ok with context.Session throwing when the session middleware is not present?

ajaybhargavb: There is already a check for that here. https://github.com/aspnet/Mvc/pull/2647/files#diff-a87ef667e5f25cf11336f7ebc2be8530L58

pranavkm: Should we additionally have a check for this as part of setting the temp data?
So

`TempData["foo"] = value;` would throw if the backing TempDataProvider isn't "enabled". It avoids user exceptions being lost to ether \ logging.

kichalla: @Tratcher : Discussed this with @pranavkm and @ajaybhargavb ...we will let context.Session throw instead...will update the PR...

kichalla: Discussed offline: we should not throw if Session is not available as it would impact users who do not even care about TempData(example: purely Web API users). Even logging the details might not help as we potentially might need to log it as a Warning and it would also appear in logs when the user doesn't care about TempData to begin with.

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
kichalla: TempDataDictionary already checks for null and creates an empty dictionary...also i think we should throw an exception in this case.

kichalla:
Any other comments? /cc: @Tratcher @pranavkm @ajaybhargavb

src/Microsoft.AspNet.Mvc.Core/SessionStateTempDataProvider.cs
ajaybhargavb: Returning `null` here changes the behavior https://github.com/aspnet/Mvc/pull/2647/files#diff-66225f3019f3fc28323c68c616ec48b9R29. Is this intentional?

kichalla: yes, right.

ajaybhargavb:
:shipit:

PR #2653, opened 2015-06-04
Fix #2278 - Only activate public properties

### LINKED ISSUE #2278: `Activate` activates non-public Controller properties, but [From*] do not
We have a mismatch between what we'd be willing to activate using `[ActivateAttribute]` as opposed to `[From*]` attributes. Consider

```C#
public class TestController
{
    [Activate]
    protected HttpContext HttpContext2 { get; set; }

    [FromServices]
    protected IMyService SomeService { get; set; }
}
```

We'll activate the `HttpContext2` property, but not model bind the `SomeService` property. We should consider changing activation to only look at public properties although this might have impact on other things that also get activated.

rynowak:


test/Microsoft.AspNet.Mvc.Core.Test/ViewComponents/DefaultViewComponentActivatorTests.cs
```diff
@@ -39,6 +56,12 @@ public Task ExecuteAsync()
                 throw new NotImplementedException();
             }
         }
+
+        private class VisibilityViewComponent : ViewComponent
+        {
+            [ViewComponentContext]
+            protected internal ViewComponentContext C { get; set; }
```
pranavkm: Let's get a better name. :8ball:

pranavkm:
:shipit:

src/Microsoft.AspNet.Mvc.Razor/RazorPageActivator.cs
```diff
@@ -87,7 +87,8 @@ private PageActivationInfo CreateViewActivationInfo(Type type)
                 PropertyActivators = PropertyActivator<ViewContext>.GetPropertiesToActivate(
                     type,
                     typeof(RazorInjectAttribute),
-                    CreateActivateInfo)
+                    CreateActivateInfo,
+                    includeNonPublic: true)
```
dougbu: should we change the `InjectChunkVisitor` to spit out a `public` setter for the generated properties?

PR #2654, opened 2015-06-04
Fixing 2340: ModelMetadata should recompute localizable properties.

harshgMSFT:


src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DataAnnotationsMetadataProvider.cs
```diff
@@ -84,7 +84,7 @@ public void GetDisplayMetadata([NotNull] DisplayMetadataProviderContext context)
             // DisplayName
             if (displayAttribute != null)
             {
-                displayMetadata.DisplayName = displayAttribute.GetName();
+                displayMetadata.DisplayName = (propertyName) => displayAttribute.GetName();
```
harshgMSFT: might impact perf as its capturing a lambda. Opinions ?

rynowak: Typically when you want to be careful with a capturing lambda is on a known hot-path or in a tight loop. The issue that is a capturing lambda is an allocation.

Three reasons not to care:
- Measure first, then optimize
- This is definitely not a hot path (it's already cached)
- What non-allocating option would you choose?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/ILocalizableDisplayMetadataProvider.cs
```diff
@@ -0,0 +1,16 @@
...
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Linq;
+using System.Reflection;
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
+{
+    public interface ILocalizableDisplayMetadataProvider : IDisplayMetadataProvider
```
rynowak: What's this for?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DisplayMetadata.cs
```diff
@@ -40,10 +41,10 @@ public class DisplayMetadata
         public string DisplayFormatString { get; set; }
 
         /// <summary>
-        /// Gets or sets a display name for the model.
+        /// Gets or sets a delegate used for getting the display name for the model.
         /// See <see cref="ModelMetadata.DisplayName"/>
         /// </summary>
-        public string DisplayName { get; set; }
+        public Func<string, string> DisplayName { get; set; }
```
rynowak: Why not just make this a `Func<string>`?

IDisposable: Because it's a mapper that needs the property name (or else you have to capture that up higher in the delegate creation.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DataAnnotationsMetadataProvider.cs
```diff
@@ -72,7 +72,7 @@ public void GetDisplayMetadata([NotNull] DisplayMetadataProviderContext context)
             // Description
             if (displayAttribute != null)
             {
-                displayMetadata.Description = displayAttribute.GetDescription();
+                displayMetadata.Description = (propertyName) => displayAttribute.GetDescription();
```
dougbu: don't parenthesize a single lambda argument

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DataAnnotationsMetadataProvider.cs
dougbu: minor (and unrelated to perf :smiley_cat:) but don't we use `_` for unused lambda arguments?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DisplayMetadata.cs
```diff
@@ -28,10 +29,10 @@ public class DisplayMetadata
         public string DataTypeName { get; set; }
 
         /// <summary>
-        /// Gets or sets a model description.
+        /// Gets or sets a delegate used for getting the model description.
```
dougbu: Describe the `Func`'s input and output.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/ILocalizableDisplayMetadataProvider.cs
dougbu: yeah, not referenced elsewhere in this PR

dougbu:
Would like to see a test or two showing different descriptions or display names as the current locale changes. For example, might be a `[Theory]` unit test with a `[Display]` attribute in the model that uses a (faked) locale-sensitive resource.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DisplayMetadata.cs
dougbu: @IDisposable and @harshgMSFT I understand this makes it possible to use the same lambda for all properties of a `Type`. But I agree with @rynowak that this is an unlikely implementation choice. `Description` and `DisplayName` are assigned in a property-specific context. If the caller wants to deter use of the property name, they can do so without the API forcing this.

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DisplayMetadata.cs
harshgMSFT: Well, I agree that it is in property specific context ( when the metadata represents a property), Having a string key as an input enables cases when someone wants to use it as a key to lookup a resource and this key might not be a property name ( for example when the metadata represents a type). Unless I hear a strong push against it this stays as is.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DataAnnotationsMetadataProvider.cs
harshgMSFT: @rynowak :+1:  ... @dougbu  sure.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DisplayMetadata.cs
harshgMSFT: done.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/ILocalizableDisplayMetadataProvider.cs
harshgMSFT: my bad was doing something and forgot to remove it.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DisplayMetadata.cs
dougbu: If a user wishes to use a string key for a resource lookup directly, they can do that without enabling it at this level. They just need to create a lambda with a closure over the property name. This is important because the current approach forces _every_ user to create such a closure -- and most (as shown in your test changes) don't need the property name.

``` c#
  public class MyDisplayMetadataProvider
  {
    public void GetDisplayMetadata(DisplayMetadataProviderContext context)
    {
      if (context.Key.MetadataKind == ModelMetadataKind.Property)
      {
        context.DisplayMetadata.DisplayName = () => GetDisplayName(context.Key.Name);
      }
    }

    private static string GetDisplayName(string propertyName)
    {
      // Do something magical with propertyName.
      return propertyName;
    }
  }
```

dougbu: BTW this isn't about performance. It's about DRY, providing one way to do something.

harshgMSFT: hmm.. :OK...

harshgMSFT:
@dougbu there are similar tests in DataAnnotationsMetadataProviderTest already(`GetDisplayMetadata_DisplayAttribute_NameFromResources` `GetDisplayMetadata_DisplayAttribute_DescriptionFromResources`). let me know if you meant something else

harshgMSFT:
updated.

test/WebSites/ModelBindingWebSite/Controllers/ModelMetadataController.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
```
harshgMSFT: revert

dougbu:
> there are similar tests in `DataAnnotationsMetadataProviderTest`

Those tests get a (single) value from a resource but don't demonstrate getting _different_ values as the culture changes.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
```diff
@@ -179,7 +179,12 @@ public override string Description
         {
             get
             {
-                return DisplayMetadata.Description;
+                if (DisplayMetadata.Description == null)
+                {
+                    return null;
+                }
+
+                return DisplayMetadata.Description(PropertyName);
```
rynowak: This looks like it hasn't been updated yet.

rynowak:
:shipit: from me once doug is happy

harshgMSFT:
@dougbu updated.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/TestResources.cs
```diff
@@ -12,5 +13,10 @@ public static class TestResources
         public static string DisplayAttribute_Description { get; } = Resources.DisplayAttribute_Description;
 
         public static string DisplayAttribute_Name { get; } = Resources.DisplayAttribute_Name;
+
+        public static string DisplayAttribute_Name2 =>
+            Resources.DisplayAttribute_Name + Thread.CurrentThread.CurrentUICulture;
+        public static string DisplayAttribute_Description2 =>
```
harshgMSFT: add a newline.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/Metadata/ModelMetadataProviderTest.cs
```diff
@@ -942,6 +998,16 @@ private class User
             public int NotIncludedOrExcluded { get; set; }
         }
 
+        private class ResourceType
```
harshgMSFT: remove this class and the one below,

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DisplayMetadata.cs
```diff
@@ -28,10 +29,10 @@ public class DisplayMetadata
         public string DataTypeName { get; set; }
 
         /// <summary>
-        /// Gets or sets a model description.
-        /// See <see cref="ModelMetadata.Description"/>
+        /// Gets or sets a delegate which is used to get a value for the
+        /// model description see <see cref="ModelMetadata.Description"/>.
```
dougbu: change to "description. See"

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/TestResources.cs
```diff
@@ -12,5 +13,10 @@ public static class TestResources
         public static string DisplayAttribute_Description { get; } = Resources.DisplayAttribute_Description;
 
         public static string DisplayAttribute_Name { get; } = Resources.DisplayAttribute_Name;
+
+        public static string DisplayAttribute_Name2 =>
```
dougbu: Give this and `DisplayAttribute_Description2` more meaningful names. Suggest `DisplayAttribute_NamePlusCulture` for this one. Or maybe `DisplayAttribute_CultureSensitiveName`.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/Metadata/ModelMetadataProviderTest.cs
```diff
@@ -421,6 +423,60 @@ public void DisplayAttribute_Description()
...
+                ResourceType = typeof(TestResources),
+            };
+
+            var provider = CreateProvider(new[] { display });
+            var metadata = provider.GetMetadataForType(typeof(string));
+            var currentCulture = Thread.CurrentThread.CurrentUICulture;
+
+            // Act & Assert
+            var result = metadata.DisplayName;
+            Assert.Equal("name from resources" + currentCulture.Name, result);
```
dougbu: This is not at all reliable. Never depend on the ambient locale in a test unless within a context protected by `[ReplaceCulture]`, `CultureReplacer`, or `CultureReplacingMiddleware`.

Suggest changing this `// Act & Assert` block to loop through two or three UI culture names and consistently use `CultureReplacer` in that loop. Same for the new test below.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/Metadata/ModelMetadataProviderTest.cs
```diff
@@ -421,6 +423,60 @@ public void DisplayAttribute_Description()
...
+
+            var cultureString = "fr-FR";
+            using (new CultureReplacer(uiCulture: cultureString))
+            {
+                // Reading it again ensures that we recompute.
```
dougbu: With change above, update this comment to `// Later iteration ensures value is recomputed` or (if you decide on 3 UI cultures) `// Later iterations ensure value is recomputed.`

dougbu:
:watch:  (close)

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/Metadata/ModelMetadataProviderTest.cs
harshgMSFT: will surround a culturereplacer block to the first call as well.

harshgMSFT:
67d0bf880a2c584c08aa5fd82c94a0408cf8647c

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
```diff
@@ -179,7 +179,12 @@ public override string Description
         {
             get
             {
-                return DisplayMetadata.Description;
+                if (DisplayMetadata.Description == null)
+                {
+                    return null;
+                }
+
+                return DisplayMetadata.Description();
```
dougbu: FYI only. This can be done in one line w/ C# 6.0: `return DisplayMetadata.Description?.Invoke();`

harshgMSFT: you are right .. dont know why i overlooked while writing it .. thanks for pointing out.

PR #2655, opened 2015-06-05
Use stable locale when testing date/time formats

dougbu:
- `CultureReplacerMiddleware` no longer helping `RazorWebSite`
 - likely regressed with @30d11a8

### DIFF
--- a/test/Microsoft.AspNet.Mvc.FunctionalTests/HtmlHelperOptionsTest.cs
+++ b/test/Microsoft.AspNet.Mvc.FunctionalTests/HtmlHelperOptionsTest.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Builder;
+using Microsoft.AspNet.Testing;
 using Microsoft.Framework.DependencyInjection;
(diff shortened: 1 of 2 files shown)


rynowak:
Isn't this change just going to fight with the localisation middleware? 

Would it be better to have functional tests that depend on locales play *with* the localization middleware and abandon `[ReplaceCulture]` for functional test? Not sure I'm convinced either way, but I wanted to pose the question.

dougbu:
> Isn't this change just going to fight with the localisation middleware?

`RazorWebSite` configures both `CultureReplacerMiddleware` (our test middleware) and `RequestLocalizationMiddleware`. `RequestLocalizationMiddleware` is winning which means either the ambient culture must be given a consistent value (which is exactly what `[ReplaceCulture]` is for) or the test client would need to send an `Accept-Language` header (making it a different test). `CultureReplacerMiddleware` should probably be removed from this particular site since it is irrelevant.

dougbu:
Updated per offline conversations...

kirthik:
:shipit:

dougbu:
cda2c67

Issue #2564, opened 2015-05-14
BindAttribute is easy to get wrong, leading to a model with no bindable properties


rynowak:
It's easy to do the wrong thing with `[Bind(...)]` and end up excluding all properties from ModelBinding. The MVC5 `BindAttribute` supported a comma-separated list of included properties:
https://msdn.microsoft.com/en-us/library/system.web.mvc.bindattribute.include(v=vs.118).aspx

*whereas* MVC6 `BindAttribute` uses `params string[]`. Anyone copy-pasting code is likely to make this mistake.

Compare:
```
[Bind("Title", "Year", "Price",  "Genre")] // correct in MVC 6
public class Book
{
        [Required]
        public string Title { get; set; }
}

[Bind(Include = "Title, Year, Price, Genre")] // correct in MVC 5
public class Book
{
        [Required]
        public string Title { get; set; }
}

[Bind("Title, Year, Price, Genre")] // incorrect in MVC 6
public class Book
{
        [Required]
        public string Title { get; set; }
}
```

The MutableObjectModelBinder will attempt to handle this model, as it has `ModelMetadata.IsComplexType == true`, but then won't actually bind anything, because the predicate excludes all of the properties.

The result is a *success* but with a default-initialized model object. If you have validators on the properties, they will of course run and generally find the model to be invalid.

Muchiachio:
Little question, will you add "Exlude" option later on, or was it design decision to only leave "Include"?

rynowak:
/cc @blowdart

It was a design decision to only have `Include` this time, it's too easy to get it wrong with `Exclude` when you add properties to the model.

If you look at the implementation, it's now a customizable extensibility point:
https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/BindAttribute.cs

https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/IPropertyBindingPredicateProvider.cs

You could implement your own policy that does something creative if `Include` by itself isn't sufficient. Here's an example that does `Include` using expression trees: https://github.com/aspnet/Mvc/blob/64e726d2b26422c5452475627e4afaba307edec3/test/WebSites/ModelBindingWebSite/Controllers/BindAttributeController.cs#L73


We'd still recommend as the most secure option, only using view models that expose **just** what you want to bind. If you can't use viewmodels or are porting a legacy then whitelisting (`Include`) is the next best thing.

[...]

PR #2656, opened 2015-06-05
Support comma separated string include in BindAttribute

ajaybhargavb:
Issue - #2564 
@rynowak @Eilon

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/BindAttribute.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/BindAttribute.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using Microsoft.AspNet.Mvc.Core;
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Mvc.Core/BindAttribute.cs
```diff
@@ -136,5 +140,17 @@ private ObjectFactory GetFactory()
...
         }
+
+        private static string[] SplitString(string original)
+        {
+            if (string.IsNullOrEmpty(original))
+            {
+                return new string[0];
+            }
+
+            var split = original.Split(',').Select(piece => piece.Trim()).Where(piece => !string.IsNullOrEmpty(piece));
```
rynowak: Look at the overload of `Split` with `StringSplitOptions` - then you don't need any of this ceremony

src/Microsoft.AspNet.Mvc.Core/BindAttribute.cs
```diff
@@ -136,5 +140,17 @@ private ObjectFactory GetFactory()
...
+
+            return split.ToArray();
```
rynowak: A bit of a nitpick, but think about removing some of the ToArrays here by just adding to a list or using `SelectMany` in your top level loop.

Eilon:
:shipit:

rynowak:
:shipit:

Issue #2526, opened 2015-05-08
Remove use of RequiredValidator from Mutable Model Binder


Eilon:
We should just ignore this case and let whatever happens, happen.

dougbu:
Had three cases when we opened this bug:
1. Property in model type has an associated `[Required]` attribute that strangely doesn't actually validate. Submission leaves the property `null`.
2. Property in model type has associated `[BindRequired]` and `[Required]` attributes. `[Required]` attribute strangely does not actually validate. Submission does not include data for the property.
3. Property in model type has associated `[BindRequired]` attribute but no `[Required]` attribute. Submission does not include data for the property.

In all cases, we previously provided a fallback error message.

However case 1 no longer exists because the fallback code existed only in model binding. @rynowak removed the `[Required]` special cases from model binding. Validation code has no equivalent fallback so :dancers: 

We have a number of other bugs and design change requests out related to cases 2 and 3. For now i.e. for this bug, I will just remove the fallback message from case 2. If there is no `[Required]` attribute (case 3), model binding will continue to provide its own error for the missing property data.

tl;dr; wherever I said "`[Required]` attribute" above, I meant "validator with `IsRequired == true`".

dougbu:
Postponing based on offline conversations. Will discuss and decide what to do about the mentioned cluster of bugs touching on how `[BindRequired]` works.

Eilon:
To be clear, in the Mutable Model Binder where it does some fancy stuff to create an error message for a "not provided" value on the wire, just put in our own custom error message always.

In the far future, we can look at some way to customize this.

Eilon:
The error message should be: `A value for the '{0}' property was not provided.`

Copied from https://github.com/aspnet/Mvc/issues/2493

Eilon:
The future work is captured in https://github.com/aspnet/Mvc/issues/2303.

dougbu:
eefa582

Issue #2493, opened 2015-05-04
Model Validation Errors text should be more inconsistent.


harshgMSFT:
For example error for the following says : 
`The 'OrderIds' property is required.`
```c#
private class Order
{
         [BindRequired]
         public List<int> OrderIds { get; set; }
}
```

While the error for the following scenario says :
`The OrderIds field is required.`

```c#
private class Order
{
         [Required]
         public List<int> OrderIds { get; set; }
}
``` 
This is because error for BindRequired is added by MutableObjectBinder and not the validation code.

Eilon:
The case of `[BindRequired]` is one where the app developer has a programming error because they didn't post a value at all for the field, and so the displayed error could be improved to indicate something like "A value for the '{0}' property was not provided."

The Data Annotations one is fine because that represents a user error (user didn't type a value into a required field).

Eilon:
Dup of https://github.com/aspnet/Mvc/issues/2526.

PR #2658, opened 2015-06-05
Address #2526; remove use of a required validator when validating `[BindRequired]`

dougbu:
- only use MVC error message when `[BindRequired]` is violated
- update that error message to more clearly describe the problem
- enable all tests skipped due to dupe bug #2493
- update expectations of a few tests using the old messages

nits:
- rename `ModelBinding_MissingRequiredMember` to `ModelBinding_MissingBindRequiredMember`
- remove `<param>` description of removed `requiredValidator` parameter
- remove unused `MutableObjectModelBinderTest.GetRequiredValidator()`

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.ComponentModel;
 using System.Linq;
 using System.Reflection;
 using System.Threading.Tasks;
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
```diff
@@ -364,42 +363,19 @@ internal static PropertyValidationInfo GetPropertyValidationInfo(ModelBindingCon
...
             foreach (var missingRequiredProperty in validationInfo.RequiredProperties)
             {
-                var addedError = false;
-
-                // We want to provide the 'null' value, not the value of model.Property,
-                // so avoiding modelExplorer.GetProperty here which would call the actual getter on the
-                // model. This avoids issues with value types, or properties with pre-initialized values.
-                var propertyExplorer = modelExplorer.GetExplorerForProperty(missingRequiredProperty, model: null);
-
+                var propertyExplorer = modelExplorer.GetExplorerForProperty(missingRequiredProperty);
```
rynowak: This shouldn't be needed anymore right?

rynowak:
:shipit:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
dougbu: Need the `BinderModelName`. So this is needed still.

dougbu:
eefa582

PR #2659, opened 2015-06-06
Rename `MvcTagHelpersWebSite` to `HtmlGenerationsWebSite`

dougbu:
- name was too specific and I am about to add another HTML helper scenario


dougbu:
FYI only. Will submit once one of the Travis builds succeeds. Discussed new name offline and this is too big to leave sitting around.

dougbu:
261975b

Issue #2609, opened 2015-05-26
Integers from JSON don't bind to enums


Tragetaschen:
I have the following setup:
- A Razor `<select>` populated by `GetEnumSelectList`, so the options's values are numbers
- Some client code that creates JSON from the currently selected entry
- A controller, that parses the JSON (now the values are `Int64`!)
- Model binding trying to fill an enum field

This setup fails, because enum binding [is only supported for `Int32`](https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs#L145).

I didn't check, but a failing test looks much like [this `ValueProviderResult` test](https://github.com/aspnet/Mvc/blob/dev/test/Microsoft.AspNet.Mvc.Abstractions.Test/ModelBinding/ValueProviderResultTest.cs#L163) with the `1` replaced by `1L`.

danroth27:
@Tragetaschen Thanks for reporting this issue! Do you have a simple project you could send us that reproduces this issue so that we can investigate?

Eilon:
We should ideally support value provider results containing all kinds of numeric treasures.

@ajaybhargavb , you can sync with @dougbu on the background for this bug.

PR #2671, opened 2015-06-08
Support for binding all integer types to enums

ajaybhargavb:
Issue - #2609 
@dougbu @Eilon

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
@@ -142,9 +142,17 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             if (!(canConvertFrom || converter.CanConvertTo(destinationType)))
             {
                 // EnumConverter cannot convert integer, so we verify manually
-                if (destinationType.GetTypeInfo().IsEnum && (value is int))
+                if (destinationType.GetTypeInfo().IsEnum &&
+                    (value is int ||
+                    value is uint ||
+                    value is long ||
+                    value is ulong ||
+                    value is short ||
+                    value is ushort ||
+                    value is byte ||
+                    value is sbyte))
                 {
-                    return Enum.ToObject(destinationType, (int)value);
+                    return Enum.ToObject(destinationType, value);
                 }
 
                 throw new InvalidOperationException(
--- a/test/Microsoft.AspNet.Mvc.Abstractions.Test/ModelBinding/ValueProviderResultTest.cs
+++ b/test/Microsoft.AspNet.Mvc.Abstractions.Test/ModelBinding/ValueProviderResultTest.cs
@@ -160,13 +160,47 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         public void ConvertToReturnsValueIfArrayElementIsIntegerAndDestinationTypeIsEnum()
         {
             // Arrange
-            var vpr = new ValueProviderResult(new object[] { 1 }, null, CultureInfo.InvariantCulture);
+            var result = new ValueProviderResult(new object[] { 1 }, null, CultureInfo.InvariantCulture);
 
             // Act
-            var outValue = vpr.ConvertTo(typeof(MyEnum));
+            var outValue = result.ConvertTo(typeof(IntEnum));
 
             // Assert
-            Assert.Equal(outValue, MyEnum.Value1);
+            Assert.Equal(outValue, IntEnum.Value1);
+        }
+
+        [Theory]
+        [InlineData(1, typeof(IntEnum), IntEnum.Value1)]
+        [InlineData(1L, typeof(LongEnum), LongEnum.Value1)]
+        [InlineData(long.MaxValue, typeof(LongEnum), LongEnum.MaxValue)]
+        [InlineData(1U, typeof(UnsignedIntEnum), UnsignedIntEnum.Value1)]
+        [InlineData(1UL, typeof(IntEnum), IntEnum.Value1)]
+        [InlineData((byte)1, typeof(ByteEnum), ByteEnum.Value1)]
+        [InlineData(byte.MaxValue, typeof(ByteEnum), ByteEnum.MaxValue)]
+        [InlineData((sbyte)1, typeof(ByteEnum), ByteEnum.Value1)]
+        [InlineData((short)1, typeof(IntEnum), IntEnum.Value1)]
+        [InlineData((ushort)1, typeof(IntEnum), IntEnum.Value1)]
+        [InlineData(int.MaxValue, typeof(IntEnum?), IntEnum.MaxValue)]
+        [InlineData(null, typeof(IntEnum?), null)]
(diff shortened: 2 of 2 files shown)


test/Microsoft.AspNet.Mvc.Abstractions.Test/ModelBinding/ValueProviderResultTest.cs
```diff
@@ -169,6 +169,27 @@ public void ConvertToReturnsValueIfArrayElementIsIntegerAndDestinationTypeIsEnum
...
 
+        [Theory]
+        [InlineData(1)]
+        [InlineData(1L)]
+        [InlineData(1U)]
+        [InlineData(1UL)]
+        [InlineData((byte)1)]
+        [InlineData((sbyte)1)]
+        [InlineData((short)1)]
+        [InlineData((ushort)1)]
```
rynowak: What about nullables?

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
```diff
@@ -142,9 +142,17 @@ private object ConvertSimpleType(CultureInfo culture, object value, Type destina
...
+                    value is uint ||
+                    value is long ||
+                    value is ulong ||
+                    value is short ||
+                    value is ushort ||
+                    value is byte ||
+                    value is sbyte))
                 {
-                    return Enum.ToObject(destinationType, (int)value);
+                    return Enum.ToObject(destinationType, Convert.ToInt32(value));
```
rynowak: Should we instead convert to `Enum.GetUnderlyingType`? You'll have loss of precision with a `long`-based enum for example.

rynowak: You could also use https://msdn.microsoft.com/en-us/library/k6x6b517(v=vs.110).aspx - looks like this is documented to do what you want

rynowak: And it also looks like we should be checking `Enum.IsDefined`.

test/Microsoft.AspNet.Mvc.Abstractions.Test/ModelBinding/ValueProviderResultTest.cs
rynowak: Should also test some varying enum-bases.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
dougbu: W.r.t. using `Enum.ToObject()`, note the rest of this line :smiley_cat:  The cast or conversion of this argument is not needed at all.

dougbu: @rynowak why should this code check `Enum.IsDefined`?  Conversion from `int`s to `enum`s as well as `enum` handling in model binding has no such requirement.

test/Microsoft.AspNet.Mvc.Abstractions.Test/ModelBinding/ValueProviderResultTest.cs
dougbu: yeah, agree we should change `m` to 16 then square `m`.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
IDisposable: :+1: @dougbu there's no call for IsDefined here. If someone wants that later it would be a different validation.

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
rynowak: @dougbu - surprising to me that we don't have it. It seems really surprising to me from a security POV because it's untrusted input. I'll open a separate bug. 

I trust that HtmlHelpers do the right thing with non-defined values?

dougbu: > I trust that HtmlHelpers do the right thing with non-defined values?

Yes they do the right thing. They ignore whether or not the value is defined. Same as it ever was because a defined `enum` value is just an `int` (or similar) with a special name.

Eilon: The loss of precision here concerns me. Can we add a unit test for really, really, *really* big long values? I.e. `long.MaxValue`?

The check for `IsDefined` on the other hand does not concern me. A validator should do that work, not type conversion.

ajaybhargavb:
Updated.

test/Microsoft.AspNet.Mvc.Abstractions.Test/ModelBinding/ValueProviderResultTest.cs
```diff
@@ -160,13 +160,44 @@ public void ConvertToReturnsNullIfTrimmedValueIsEmptyString()
...
+        [InlineData((ushort)1, typeof(IntEnum), IntEnum.Value1)]
+        [InlineData(int.MaxValue, typeof(IntEnum?), IntEnum.Value2)]
+        [InlineData(null, typeof(IntEnum?), null)]
+        [InlineData(1L, typeof(LongEnum?), LongEnum.Value1)]
+        [InlineData(null, typeof(LongEnum?), null)]
+        [InlineData(uint.MaxValue, typeof(UnsignedIntEnum?), UnsignedIntEnum.Value2)]
+        [InlineData((byte)1, typeof(ByteEnum?), ByteEnum.Value1)]
+        [InlineData(null, typeof(ByteEnum?), null)]
+        [InlineData((ushort)1, typeof(LongEnum?), LongEnum.Value1)]
+        public void ConvertToReturnsValueIfArrayElementIsAnyIntegerTypeAndDestinationTypeIsEnum(object input, Type enumType, object expected)
```
rynowak: long line

rynowak:
:shipit: from me

test/Microsoft.AspNet.Mvc.Abstractions.Test/ModelBinding/ValueProviderResultTest.cs
```diff
@@ -492,10 +523,32 @@ private class MyClassWithoutConverter
         {
         }
 
-        private enum MyEnum
+        private enum IntEnum
+        {
+            Value0 = 0,
+            Value1 = 1,
+            Value2 = int.MaxValue
```
dougbu: suggest `MaxValue` for these. not mandatory but better than `Value2`.

dougbu:
:shipit: after quick `enum` value renames

Eilon:
:shipit:

Issue #2664, opened 2015-06-08
`ArgumentNullException` thrown in `DefaultHtmlGenerator.GetCurrentValues()`


dougbu:
When the model contains a collection of `enum` values and the view uses `@Html.ListBox()` for the property, `GetCurrentValues()` will throw an `ArgumentNullException`. Collections of `enum` values may be unusual but should be correctly supported.

The root cause is twofold: `GetCurrentValues()` determines the `ModelExplorer` for the property but uses that directly rather than the `ModelMetadata` for the element type. And the method assumes `enumNames` (from that `ModelMetadata` / `ModelExplorer`) will be non-`null` when a raw value is an `enum`.

The problem is a bit more general because it's possible to place `enum` values in `ViewDataDictionary` though the model is (say) an `int`. But that's much less likely than the collection case.

dougbu:
Self-assigning because this is also mixed into the #1485 / #1487 / #2662 mess.

dougbu:
3f6ab3b

Issue #1487, opened 2014-10-30
Most string-based HTML helpers use unprefixed `ViewDataDictionary` entries if called from templates


dougbu:
The string-based helpers _not_ mentioned in  #1485 (`@Html.TextArea()` etc.) also do not behave as expected in template views.  But this problem occurs only if an entry exists in the `ViewDataDictionary` matching the (un-prefixed) expression name.  So unexpected values will be used in the generated HTML -- luckily, in a somewhat narrow case.

E.g. action contains
``` c#
ViewData["Age"] = 39;
return View(new User()
{
  Age = 26,
  Dependent = new User()
  {
    Age = 13,
  },
});
```
Index.cshtml contains
``` c#
@Html.DisplayFor(model => model.Dependent, templateName: "Dependent")
```
DisplayTemplates/Dependent.cshtml contains:
``` html
<td>
    @Html.Label(expression: "Age", labelText: null, htmlAttributes: new { @class = "control-label", })
</td>
<td>
    '@Html.Value(name: "Age")'
</td>
```

Result is:
![capture](https://cloud.githubusercontent.com/assets/6431421/4853017/0d52e31c-6082-11e4-9680-5530bc380302.png)

Expect to see "13" displayed between the quotes.

dougbu:
Problem also affects how client validation attributes are found. For example with the following in an `Int32.cshtml` template,
``` c#
@{
  // Don't do this; causes template to ignore `ViewData.Model` and overwrite existing `ViewData`.
  ViewData[string.Empty] = "two";
  ViewData[Html.ViewData.TemplateInfo.HtmlFieldPrefix] = 23;
}

@Html.TextBox(expression: string.Empty);
```

The generated HTML will be
``` html
<input id="z0__Number" name="[0].Number" type="text" value="23" />
```

The HTML helper finds the correct value but the incorrect `ModelMetadata`. (Client-side validation should require the `int` value but no `data-val-required` attribute is generated.)

dougbu:
27283ec

PR #2676, opened 2015-06-09
#2664: Add `ModelMetadata.ElementMetadata`

dougbu:
- builds on dougbu/testing.1485.1487.2662.2664 branch, unblocking some #1487 tests
- use new property to correctly determine `isTargetEnum` in `GetCurrentValues()`
 - avoid `ArgumentNullException` in all cases where raw values are `enum` but target is not
- also use new property instead of private `GetElementType()` methods where possible

nits:
- move properties above methods in `ModelMetadata`
- avoid accidentally-incorrect "Remove Unnecessary Usings"

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelMetadata.cs
+++ b/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelMetadata.cs
@@ -5,7 +5,9 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.ComponentModel;
+#if DNXCORE50
 using System.Reflection;
+#endif
 using Microsoft.AspNet.Mvc.ModelBinding.Metadata;
 using Microsoft.Framework.Internal;
 
@@ -79,7 +81,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         public abstract string BinderModelName { get; }
 
         /// <summary>
-        /// Gets the <see cref="Type"/> of an <see cref="IModelBinder"/> of a model if specified explicitly using 
+        /// Gets the <see cref="Type"/> of an <see cref="IModelBinder"/> of a model if specified explicitly using
         /// <see cref="IBinderTypeProviderMetadata"/>.
         /// </summary>
         public abstract Type BinderType { get; }
@@ -123,6 +125,18 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         /// http://msdn.microsoft.com/en-us/library/txafckwd.aspx) used to edit the model.
         /// </summary>
         public abstract string EditFormatString { get; }
+
+        /// <summary>
+        /// Gets the <see cref="ModelMetadata"/> for elements of <see cref="ModelType"/> if that <see cref="Type"/>
+        /// implements <see cref="IEnumerable"/>.
+        /// </summary>
+        /// <value>
+        /// <see cref="ModelMetadata"/> for <c>T</c> if <see cref="ModelType"/> implements
+        /// <see cref="IEnumerable{T}"/>. <see cref="ModelMetadata"/> for <c>object</c> if <see cref="ModelType"/>
+        /// implements <see cref="IEnumerable"/> but not <see cref="IEnumerable{T}"/>. <c>null</c> otherwise i.e. when
+        /// <see cref="IsCollectionType"/> is <c>false</c>.
+        /// </value>
+        public abstract ModelMetadata ElementMetadata { get; }
 
         /// <summary>
         /// Gets the ordered display names and values of all <see cref="Enum"/> values in <see cref="ModelType"/> or
@@ -330,6 +344,16 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         }
 
         /// <summary>
+        /// Gets or sets a property getter delegate to get the property value from a model object.
+        /// </summary>
+        public abstract Func<object, object> PropertyGetter { get; }
+
+        /// <summary>
+        /// Gets or sets a property setter delegate to set the property value on a model object.
+        /// </summary>
+        public abstract Action<object, object> PropertySetter { get; }
+
+        /// <summary>
         /// Gets a display name for the model.
         /// </summary>
         /// <remarks>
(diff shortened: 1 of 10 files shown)


src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelMetadata.cs
```diff
@@ -330,6 +344,16 @@ public bool IsCollectionType
         }
 
         /// <summary>
+        /// Gets or sets a property getter delegate to get the property value from a model object.
+        /// </summary>
+        public abstract Func<object, object> PropertyGetter { get; }
+
+        /// <summary>
+        /// Gets or sets a property setter delegate to set the property value on a model object.
```
rynowak: These shouldn't say `or sets`

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelMetadata.cs
```diff
@@ -5,7 +5,9 @@
 using System.Collections;
 using System.Collections.Generic;
 using System.ComponentModel;
+#if DNXCORE50
 using System.Reflection;
+#endif
```
rynowak: ???

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
```diff
@@ -221,6 +227,48 @@ public override string EditFormatString
...
         /// <inheritdoc />
+        public override ModelMetadata ElementMetadata
+        {
+            get
+            {
+                if (!_haveCalculatedElementMetadata)
+                {
+                    // Short-circuit checks below. As in IsCollectionType, do not consider string a collection.
+                    _haveCalculatedElementMetadata = ModelType == typeof(string);
+                }
```
rynowak: Why not check `IsCollectionType` here?

rynowak: If this is a short circuit, then just return

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
```diff
@@ -221,6 +227,48 @@ public override string EditFormatString
...
+
+                if (!_haveCalculatedElementMetadata)
```
rynowak: Wierd to see the same condition checked twice in a row

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelMetadata.cs
```diff
@@ -125,6 +127,18 @@ public string PropertyName
...
+        /// Gets the <see cref="ModelMetadata"/> for elements of <see cref="ModelType"/> if that <see cref="Type"/>
+        /// implements <see cref="IEnumerable"/>.
+        /// </summary>
+        /// <value>
+        /// <see cref="ModelMetadata"/> for <c>T</c> if <see cref="ModelType"/> implements
+        /// <see cref="IEnumerable{T}"/>. <see cref="ModelMetadata"/> for <c>object</c> if <see cref="ModelType"/>
+        /// implements <see cref="IEnumerable"/> but not <see cref="IEnumerable{T}"/>. <c>null</c> otherwise i.e. when
+        /// <see cref="IsCollectionType"/> is <c>false</c>.
+        /// </value>
+        public abstract ModelMetadata ElementMetadata { get; }
```
rynowak: :heart_eyes_cat:

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultDisplayTemplates.cs
```diff
@@ -106,40 +107,42 @@ public static string CollectionTemplate(IHtmlHelper htmlHelper)
...
-                        itemType = item.GetType();
+                        itemMetadata = metadataProvider.GetMetadataForType(item.GetType());
                     }
 
-                    var modelExplorer = metadataProvider.GetModelExplorerForType(itemType, item);
+                    var modelExplorer = new ModelExplorer(
+                        metadataProvider,
+                        container: htmlHelper.ViewData.ModelExplorer,
+                        metadata: itemMetadata,
+                        model: item);
```
rynowak: Should this be another OL on ModelExplorer?

rynowak:
:shipit: from me after addressing comments in the MM part. I'm not really qualified to speak about the HtmlHelper part of this change.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelMetadata.cs
dougbu: Only used when `Type` itself doesn't contain `IsAssignableFrom()`.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelMetadata.cs
dougbu: Sure.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
dougbu: OK, I'll move the `string` check after `_haveCalculatedElementMetadata` is set to `true` and `return` in that case.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
dougbu: > Why not check IsCollectionType here?

Because that repeats many of the checks needed below and only tells us `ElementMetadata` should be non-`null`. IOW its repetitive and not that helpful.

src/Microsoft.AspNet.Mvc.Core/Rendering/Html/DefaultDisplayTemplates.cs
dougbu: what's "OL" ?

rynowak: overload

dougbu:
Rebased and updated

dougbu:
FYI @rynowak delivered and offline :shipit: for this one.

dougbu:
3f6ab3b

PR #2681, opened 2015-06-09
A new pattern for adding multi-registration services

rynowak:
This is some cleanup of how we add multi-registration services to avoid
duplication when calling AddMvcServices multiple times. Also improved
tests to be more clear, and to verify all of our special cases
explicitly.

### DIFF
--- a/src/Microsoft.AspNet.Mvc/MvcServiceCollectionExtensions.cs
+++ b/src/Microsoft.AspNet.Mvc/MvcServiceCollectionExtensions.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Reflection;
 using Microsoft.AspNet.Mvc;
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Mvc/MvcServiceCollectionExtensions.cs
```diff
@@ -163,13 +164,22 @@ public static IServiceCollection AddMvc([NotNull] this IServiceCollection servic
...
+            TryAddMultiRegistrationService(
+                services,
+                ServiceDescriptor.Transient<IConfigureOptions<MvcFormatterMappingOptions>, JsonMvcFormatterMappingOptionsSetup>());
+            TryAddMultiRegistrationService(
+                services,
+                ServiceDescriptor.Transient<IConfigureOptions<MvcViewOptions>, MvcViewOptionsSetup>());
+            TryAddMultiRegistrationService(
+                services,
+                ServiceDescriptor
+                .Transient<IConfigureOptions<RazorViewEngineOptions>, RazorViewEngineOptionsSetup>());
```
dougbu: Odd to wrap this line and not the one for `JsonMvcFormatterMappingOptionsSetup`.

src/Microsoft.AspNet.Mvc/MvcServiceCollectionExtensions.cs
```diff
@@ -359,6 +382,27 @@ public static IServiceCollection AddMvcLocalization([NotNull] this IServiceColle
...
+        // services like IConfigureOptions<MvcOptions> or IApplicationModelProvider where you need the ability
+        // to register multiple implementation types for the same service type.
+        private static bool TryAddMultiRegistrationService(IServiceCollection services, ServiceDescriptor descriptor)
+        {
+            // This can't work when registering a factory or instance, you have to register a type.
+            // Additionally, if any existing registrations use a factory or instance, we can't check those, but we don't
+            // assert that because it might be added by user-code.
+            Debug.Assert(descriptor.ImplementationType != null);
+
+            if (services.Any(d =>
```
dougbu: `d` for "dude" bro?

src/Microsoft.AspNet.Mvc/MvcServiceCollectionExtensions.cs
```diff
@@ -359,6 +382,27 @@ public static IServiceCollection AddMvcLocalization([NotNull] this IServiceColle
             return services;
         }
 
+        // Adds a service if the service type and implementation type hasn't been added yet. This is needed for
+        // services like IConfigureOptions<MvcOptions> or IApplicationModelProvider where you need the ability
+        // to register multiple implementation types for the same service type.
+        private static bool TryAddMultiRegistrationService(IServiceCollection services, ServiceDescriptor descriptor)
```
dougbu: Nothing uses the return value. Suggest returning `void`.

dougbu:
Realize current implementation will correctly handle user services registered for multiple types (a single implementation of multiple interfaces). Might be worth a `[Fact]` to spot check that continues to work in the future for both single and multi-registration services.

dougbu:
:shipit:

src/Microsoft.AspNet.Mvc/MvcServiceCollectionExtensions.cs
IDisposable: That would break from the pattern of TryXXX returning a success indicator, should the function be renamed to AddIfNeededXXX

src/Microsoft.AspNet.Mvc/MvcServiceCollectionExtensions.cs
davidfowl: descriptor, that's fine right. It's a common convention to use a single letter for linq queries like this.

dougbu: @davidfowl if I actually follow @rynowak and say "bro", the smiley should be assumed :smiley_cat:

src/Microsoft.AspNet.Mvc/MvcServiceCollectionExtensions.cs
dougbu: @IDisposable I agree the method name should change if @rynowak follows my suggestion.

rynowak: Leaving this as-is for consistency for DI apis. We might be taking this function over to that repo anyway.

Issue #2409, opened 2015-04-20
Compatibility with MVC 5 auth code


rynowak:
Reported by @rustd 

This is MVC5 template code - `SignIn` and `SignOut` don't work in MVC6 due to no-content formatter turning a challenge/redirect into a 204
```
    public class AccountController : Controller
    {
        public void SignIn()
        {
            // Send an OpenID Connect sign-in request.
            if (!Request.IsAuthenticated)
            {
                HttpContext.GetOwinContext().Authentication.Challenge(new AuthenticationProperties { RedirectUri = "/" }, 
                    OpenIdConnectAuthenticationDefaults.AuthenticationType);
            }
        }

        public void SignOut()
        {
            string callbackUrl = Url.Action("SignOutCallback", "Account", routeValues: null, protocol: Request.Url.Scheme);

            HttpContext.GetOwinContext().Authentication.SignOut(
                new AuthenticationProperties { RedirectUri = callbackUrl },
                OpenIdConnectAuthenticationDefaults.AuthenticationType, CookieAuthenticationDefaults.AuthenticationType);
        }

        public ActionResult SignOutCallback()
        {
            if (Request.IsAuthenticated)
            {
                // Redirect to home page if the user is authenticated.
                return RedirectToAction("Index", "Home");
            }

            return View();
        }
    }
```

The fix for this would be to rewrite the code to use action results instead of returning void.

danroth27:
Should we just remove the behavior of turning void into a 204? How much do we care about the scenario where I write an action method that manipulates the HttpContext directly and I really want the method return type to be meaningless?

rynowak:
The initial hope with this feature was that the HTTP Context would give us a way to tell if the status code had been set, and that was never added.

danroth27:
Detecting if stuff has been set on `HttpContext` seems like a slippery slope. What about headers? Content?

danroth27:
We decided to remove the 204 behavior when returning void.

PR #2685, opened 2015-06-10
Removed HttpNoContentFormatter from MvcOptionsSetup

ajaybhargavb:
Issue - #2409 
@rynowak

### DIFF
--- a/test/Microsoft.AspNet.Mvc.FunctionalTests/ActionResultTests.cs
+++ b/test/Microsoft.AspNet.Mvc.FunctionalTests/ActionResultTests.cs
@@ -301,7 +301,7 @@ namespace Microsoft.AspNet.Mvc.FunctionalTests
 
             // Assert
             Assert.Equal(HttpStatusCode.NotFound, response.StatusCode);
-            Assert.Equal("", await response.Content.ReadAsStringAsync());
+            Assert.Equal("null", await response.Content.ReadAsStringAsync());
         }
 
         [Fact]
--- a/test/Microsoft.AspNet.Mvc.FunctionalTests/ModelBindingTest.cs
+++ b/test/Microsoft.AspNet.Mvc.FunctionalTests/ModelBindingTest.cs
@@ -189,7 +189,7 @@ namespace Microsoft.AspNet.Mvc.FunctionalTests
             var response = await client.GetAsync("http://localhost/FromBodyControllerProperty/GetSiteUser");
 
             // Assert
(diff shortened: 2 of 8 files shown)


test/Microsoft.AspNet.Mvc.FunctionalTests/ActionResultTests.cs
```diff
@@ -301,7 +301,7 @@ public class ActionResultTests
 
             // Assert
             Assert.Equal(HttpStatusCode.NotFound, response.StatusCode);
-            Assert.Equal("", await response.Content.ReadAsStringAsync());
+            Assert.Equal("null", await response.Content.ReadAsStringAsync());
```
rynowak: I'm not sure this is right - how is "null" getting in there?

test/Microsoft.AspNet.Mvc.FunctionalTests/ModelBindingTest.cs
```diff
@@ -2159,7 +2159,7 @@ public async Task
             // Assert
             Assert.Equal(HttpStatusCode.NoContent, response.StatusCode);
             var body = await response.Content.ReadAsStringAsync();
-            Assert.Equal(string.Empty, body);
+            Assert.Equal("null", body);
```
rynowak: seems suspect.

rynowak: Discussed in person, the action code is specifically creating an `ObjectResult() { Value = null }`

test/Microsoft.AspNet.Mvc.FunctionalTests/ActionResultTests.cs
rynowak: Discussed in person, the action code is specifically creating an `ObjectResult() { Value = null }`

rynowak:
:shipit:

test/Microsoft.AspNet.Mvc.FunctionalTests/ActionResultTests.cs
rynowak: Discussed in person, the action code is specifically creating an `ObjectResult() { Value = null }`

rynowak:
I think I jumped the gun on this, we want to change the behavior for `void` but not for returning a `null`. Let's make the change in the invoker

ajaybhargavb:
Created a new PR with the invoker changes. https://github.com/aspnet/Mvc/pull/2687
Closing this one.

Issue #2662, opened 2015-06-07
`RadioButton(string.Empty)` ignores `Model` value


dougbu:
Found while testing some scenarios for #1485 and #1487. With a template such as
```
@model Gender
@Html.RadioButton(expression: string.Empty, value: "Female") Female
@Html.RadioButton(expression: string.Empty, value: "Male") Male
```
neither value will be `checked` because the HTML helper has an incorrect `!string.IsNullOrEmpty(expression)` guard.

The guard does not prevent `ModelState` look-ups; so values will display correctly if model bound.

Work-around is to use `@Html.RadioButtonFor(m => m)`.

dougbu:
Self-assigning because, though symptoms are different, fix overlaps what I'm doing for #1485 and #1497.

dougbu:
27283ec

Issue #1485, opened 2014-10-30
Some string-based HTML helpers can't get `Model` values if called from templates


dougbu:
`@Html.Checkbox()`, `@Html.Hidden()`, and `@Html.TextBox()` do not behave as expected in template views.  They resolve values using the full expression names (which won't find properties of the inner model), unlike all other HTML helpers.  This leads to unexpectedly empty values in the generated HTML.

dougbu:
27283ec

PR #2686, opened 2015-06-10
Make `string`-based expression evaluations behave consistently

dougbu:
- #1485, #1487
 - handle `TemplateInfo.HtmlFieldPrefix` in `ViewDataEvaluator.Eval()`
  - attempt lookup in the `ViewDataDictionary` using full name then evaluate
    relative `expression` against `viewData.Model`
  - handle `null` or empty `expression` special case in this method (remove `throw`s)
 - always pass relative `expression` name into `Eval()`
  - remove `null` or empty `expression` handling from higher-level code
  - in a couple of cases, special-case returned `ViewDataInfo`
- #2662
 - remove incorrect guard from `DefaultHtmlGenerator.GenerateRadioButtion()`
- add doc comments for the core methods that have changed
- enable unit tests skipped due to one of above bugs
 - fix one (yeah, just one) other test with incorrect expectations
- remove functional test comments about the above bugs and update expectations

nits:
- move some comments describing `ViewDataEvaluator` methods above the methods

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Extensions/Rendering/Html/DefaultHtmlGenerator.cs
+++ b/src/Microsoft.AspNet.Mvc.Extensions/Rendering/Html/DefaultHtmlGenerator.cs
@@ -322,11 +322,10 @@ namespace Microsoft.AspNet.Mvc.Rendering
 
                     // isChecked not provided nor found in the given attributes; fall back to view data.
                     var valueString = Convert.ToString(value, CultureInfo.CurrentCulture);
-                    isChecked = !string.IsNullOrEmpty(expression) &&
-                        string.Equals(
-                            EvalString(viewContext, expression),
-                            valueString,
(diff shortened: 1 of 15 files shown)


dougbu:
ping @rynowak

src/Microsoft.AspNet.Mvc.Extensions/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -1194,7 +1185,11 @@ private static string GetInputTypeString(InputType inputType)
             string expression)
         {
             var value = viewContext.ViewData.Eval(expression);
```
dougbu: Add comment about expecting user to provide select list items in the dictionary and (definitely) not as the `Model`.

dougbu:
Offline :shipit: from @rynowak with new comment

dougbu:
27283ec

Issue #2684, opened 2015-06-10
Remove Microsoft.AspNet.Mvc.Common


Eilon:
I believe we might have just missed this when we decided to rename the shared source projects.

FYI @rynowak @muratg @danroth27

rynowak:
We should just remove this instead :laughing: We have better ways of doing shared packages for anything important. Other stuff in here is either trivial or used only in 1-2 places.

PR #2693, opened 2015-06-12
Removed Mvc.Common package

ajaybhargavb:
Related issue - #2684 
- Removed Mvc.Common package
- Deleted TaskHelper
- Deleted Encodings and moved custom encodings to InputFormatter base type.
- Refactored TypeHelper and moved it to Mvc.Extensions

@rynowak


rynowak:
:shipit:

dougbu:
Curious: Why not remove `TaskHelperTest` as well? Currently just testing BCL.

ajaybhargavb:
Updated.

src/Microsoft.AspNet.Mvc.Core/OutputFormatter.cs
```diff
@@ -23,6 +23,19 @@ public abstract class OutputFormatter : IOutputFormatter, IApiResponseFormatMeta
         private readonly List<MediaTypeHeaderValue> _supportedMediaTypes;
 
         /// <summary>
+        /// Returns UTF8 Encoding without BOM and throws on invalid bytes.
+        /// </summary>
+        public static readonly Encoding UTF8EncodingWithoutBOM
```
pranavkm: cc @kichalla 

Do we care about being particular about the encoding for output formatters? The BOM specification shouldn't apply since you made your change and we wouldn't be producing invalid bytes to throw, would we?

test/Microsoft.AspNet.Mvc.Common.Test/TaskHelperTest.cs
```diff
@@ -1,58 +0,0 @@
-// Copyright (c) .NET Foundation. All rights reserved.
```
pranavkm: Could you verify there were tests for the places that were using these methods (and if not add some to ensure we work correctly \ bubble the exception?)

src/Microsoft.AspNet.Mvc.Core/OutputFormatter.cs
ajaybhargavb: @pranavkm, you are right but we still need these because we explicitly set `throwOnInvalidBytes` to `true`. It is `false` by default.

pranavkm: My question was is there any utility in setting this for an output formatter? Seems like it's useful specifically for decoding, not encoding.

src/Microsoft.AspNet.Mvc.Core/InputFormatter.cs
```diff
@@ -18,6 +18,19 @@ namespace Microsoft.AspNet.Mvc
...
+        /// </summary>
+        public static readonly Encoding UTF8EncodingWithoutBOM
+            = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false, throwOnInvalidBytes: true);
+
+        /// <summary>
+        /// Returns UTF16 Encoding which uses littleEndian byte order with BOM and throws on invalid bytes.
+        /// </summary>
+        public static readonly Encoding UTF16EncodingLittleEndian = new UnicodeEncoding(bigEndian: false,
+                                                                                        byteOrderMark: true,
+                                                                                        throwOnInvalidBytes: true);
```
dougbu: Wrap same way you wrap the previous added field

dougbu: Also why not just use the new fields from `OutputFormatter`? Or at least initialize these fields using those?

src/Microsoft.AspNet.Mvc.Core/OutputFormatter.cs
ajaybhargavb: Will remove these from OutputFormatter.

test/Microsoft.AspNet.Mvc.Common.Test/TaskHelperTest.cs
ajaybhargavb: Verified. There are other tests verifying these scenarios.

ajaybhargavb:
Updated. @rynowak @dougbu @pranavkm

test/Microsoft.AspNet.Mvc.Xml.Test/XmlSerializerInputFormatterTest.cs
```diff
@@ -307,10 +307,10 @@ public void SetMaxDepth_ThrowsWhenMaxDepthIsBelowOne()
                 "There is an error in XML document." :
                 "The expected encoding 'utf-8' does not match the actual encoding 'utf-16LE'.";
 
-            var inpStart = Encodings.UTF16EncodingLittleEndian.GetBytes("<?xml version=\"1.0\" encoding=\"UTF-16\"?>" +
+            var inpStart = InputFormatter.UTF16EncodingLittleEndian.GetBytes("<?xml version=\"1.0\" encoding=\"UTF-16\"?>" +
```
pranavkm: You should be fine using `System.Text.Encoding.` types here

pranavkm: Also the other test

src/Microsoft.AspNet.Mvc.ApiExplorer/DefaultApiDescriptionProvider.cs
```diff
@@ -647,6 +664,8 @@ public int GetHashCode(PropertyKey obj)
                     return obj.ContainerType.GetHashCode() ^ obj.PropertyName.GetHashCode() ^ obj.Source.GetHashCode();
                 }
             }
+
```
pranavkm: Remove

Mvc.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.22823.1
+VisualStudioVersion = 14.0.22801.0
```
pranavkm: Undo? You're going back in time

pranavkm:
:shipit:

src/Microsoft.AspNet.Mvc.Core/InputFormatter.cs
```diff
@@ -18,6 +18,18 @@ namespace Microsoft.AspNet.Mvc
     public abstract class InputFormatter : IInputFormatter
     {
         /// <summary>
+        /// Returns UTF8 Encoding without BOM and throws on invalid bytes.
+        /// </summary>
+        public static readonly Encoding UTF8EncodingWithoutBOM
```
davidfowl: public?

ajaybhargavb: Any harm in having them as public?

src/Microsoft.AspNet.Mvc.Extensions/ViewDataDictionary.cs
```diff
@@ -426,6 +427,12 @@ private Type GetModelType(object value)
             return (value == null) ? _declaredModelType : value.GetType();
         }
 
+        private static bool IsCompatibleWith([NotNull] Type type, object value)
+        {
+            return (value == null && (!type.GetTypeInfo().IsValueType || Nullable.GetUnderlyingType(type) != null)) ||
+                (value != null && type.GetTypeInfo().IsAssignableFrom(value.GetType().GetTypeInfo()));
```
rynowak: This is a bit of an unreadable mess, can you format this better? 

Using `if` is fine.

src/Microsoft.AspNet.Mvc.Extensions/TypeHelper.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// associated value to a dictionary.
+        ///
+        /// If the object is already an <see cref="IDictionary{string, object}"/> instance, then a copy
+        /// is returned.
+        /// </summary>
+        /// <remarks>
+        /// The implementation of PropertyHelper will cache the property accessors per-type. This is
+        /// faster when the the same type is used multiple times with ObjectToDictionary.
+        /// </remarks>
+        public static IDictionary<string, object> ObjectToDictionary(object value)
```
rynowak: We should just move this to the `PropertyHelper` code since that's what it relies on.

This is a significant enough amount of code that it makes sense to have it be truly shared.

src/Microsoft.AspNet.Mvc.Core/InputFormatter.cs
rynowak: I'd go for protected unless you have a strong reason to make them public. The point is to make it easy for subclasses to do the right thing right?

src/Microsoft.AspNet.Mvc.Core/ControllerActionInvoker.cs
```diff
@@ -125,5 +125,22 @@ internal static IActionResult CreateActionResult([NotNull] Type declaredReturnTy
                 DeclaredType = actualReturnType
             };
         }
+
+        private static Type GetTaskInnerTypeOrNull([NotNull] Type type)
```
rynowak: Can this be refactored/merged with generic-type-matcher? Talk to @dougbu if you have questions

rynowak:
:watch:

src/Microsoft.AspNet.Mvc.Extensions/TypeHelper.cs
ajaybhargavb: :+1: Makes it much cleaner. Will move it.

ajaybhargavb: https://github.com/aspnet/Common/pull/25

src/Microsoft.AspNet.Mvc.Core/ControllerActionInvoker.cs
ajaybhargavb: Refactored.

ajaybhargavb:
Updated.

src/Microsoft.AspNet.Mvc.Core/ControllerActionInvoker.cs
rynowak: Much better

src/Microsoft.AspNet.Mvc.ApiExplorer/DefaultApiDescriptionProvider.cs
```diff
@@ -372,6 +372,13 @@ private Type GetDeclaredReturnType(ControllerActionDescriptor action)
             }
         }
 
+        private static Type GetTaskInnerTypeOrNull([NotNull] Type type)
```
rynowak: nitpick: no need for `[NotNull]` on private methods. If you're concerned about a high likelihood of this happening, use `Debug.Assert`

rynowak:
:shipit:

Issue #2445, opened 2015-04-24
Model State should have values for greedy model binders as well.


harshgMSFT:
Currently HeaderModelBinder and BodyModelBinder do not populate the model state by calling `SetModelValue`, for success cases. This means we cannot do a ajax post which uses the populated value.

dougbu:
Based on the tests that are disabled with this bug as the reason, problem is much more broad. All values bound from user input should have `ModelState` entries. This affects `ByteArrayModelBinder`, `FormCollectionModelBinder`, and `FormFileModelBinder` as well as the two mentioned in the original description (`BodyModelBinder` and `HeaderModelBinder`). So, "type-matching" as well as "greedy" binders.

dougbu:
- c4fa402 and 715a0b6
- further cleanup in a170a4e

PR #2695, opened 2015-06-14
Add `ModelBindingResult.IsFatalError` and make body binding more consistent

dougbu:
- part I of II for #2445 (with a duplicate code PR to follow)
- needed for #2445 because new `ModelState` entries for values will make inconsisteny worse
- change `BodyModelBinder` to use same keys for all `ModelBindingResult`s and `ModelState` entries
 - return fatal error result if formatter adds an error to `ModelState`
 - update potential callers to avoid avoid ignoring `IsFatalError`
- fix test attempting to serialize all of `ModelState`
 - will be borked with additional `RawValue`s in state
- fix two other tests that serialized `ModelState` but checked only `IsValid`

nits:
- address minor inconsistencies in `ModelBindingContext`
- use `System.Reflection.Extensions` package a bit more, where it's already referenced
- remove some unused resources


src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingResult.cs
```diff
@@ -9,6 +9,16 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
     public class ModelBindingResult
     {
         /// <summary>
+        /// Creates a new <see cref="ModelBindingResult"/> indicating a fatal error.
+        /// </summary>
+        /// <param name="key">The key using which was used to attempt binding the model.</param>
+        public ModelBindingResult(string key)
```
dougbu: In future PRs remove all `public` constructors and use explicit `static` factory methods instead.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/BodyModelBinder.cs
```diff
@@ -28,36 +27,45 @@ public BodyModelBinder()
         protected async override Task<ModelBindingResult> BindModelCoreAsync(
             [NotNull] ModelBindingContext bindingContext)
         {
+            // For compatibility with MVC 5.0 for top level object we want to consider an empty key instead of
+            // the parameter name/a custom name. In all other cases (like when binding body to a property) we
```
dougbu: Might need to file a bug for the explicit name (not the parameter name) case. Does it make sense to ignore the user's explicit name request?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/BodyModelBinder.cs
```diff
@@ -28,36 +27,45 @@ public BodyModelBinder()
...
+                // This model binder is the only handler for the Body binding source and it cannot run twice. Always
+                // tell the model binding system to skip other model binders and never to fall back i.e. indicate a
+                // fatal error.
+                return new ModelBindingResult(modelBindingKey);
             }
 
             try
             {
+                var previousCount = bindingContext.ModelState.ErrorCount;
                 var model = await formatter.ReadAsync(formatterContext);
```
dougbu: Log an issue to expand `ReadAsync()` or `FormatterContext` to handle
- adding errors at prefixed location in `ModelState`
- lack of a clear indication of an error since at least JSON formatter will never `throw`
- apparent lack of tests

Note `ErrorCount` checks won't handle the corner case where we've already got too many errors in the `ModelState`.

dougbu:
:shipit: from @rynowak

dougbu:
c4fa402

src/Microsoft.AspNet.Mvc.Core/ModelBinding/BodyModelBinder.cs
dougbu: #2722 Have also started an internal discussion on the first bullet.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/BodyModelBinder.cs
dougbu: Have started an internal discussion on this point.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingResult.cs
dougbu: @rynowak if you feel strongly about the factory method approach, please file a bug. I hesitate because this would affect all `IModelBinder` implementations.

Issue #2610, opened 2015-05-26
`ModelMetadataType` chain followed further in MVC 6 than `MetadataType` chain in MVC 5


dougbu:
For compatibility, should not check property types within a type found using `[ModelMetadataType]`.

While confirming the behaviour for #2403, found MVC 6 actually finds more metadata (checks type chains further) than MVC 5. MVC 5 finds no validators with the following model:
``` c#
    [MetadataType(typeof(MergedAttributesMetadata))]
    public class MergedAttributes
    {
        public PropertyType Property { get; set; }
    }

    public class MergedAttributesMetadata
    {
        public MetadataPropertyType Property { get; set; }
    }

    public class PropertyType
    {
    }

    [MetadataPropertyTypeValidator]
    public class MetadataPropertyType
    {
    }
```

In contrast, MVC 6 will find the `MetadataPropertyTypeValidator`. It should not.

dougbu:
296ec77

PR #2699, opened 2015-06-15
Go one less step when resolving `[ModelMetadataType]`

dougbu:
- #2610
- make MVC 6's attribute consistent with data annotation's `[MetadataType]`, used in MVC 5


rynowak:
:shipit:

dougbu:
296ec77

Issue #2700, opened 2015-06-15
Microsoft.AspNet.Mvc.HttpHeadAttribute missing?


codeclash:
I see there are 

* HttpGetAttribute
* HttpPutAttribute
* HttpPostAttribute
* HttpPatchAttribute
* HttpDeleteAttribute

... but HttpHeadAttribute seems to be missing...?

PR #2702, opened 2015-06-15
Added HttpHeadAttribute

ajaybhargavb:
Issue - #2700

### DIFF
--- a/test/Microsoft.AspNet.Mvc.Core.Test/DefaultActionSelectorTests.cs
+++ b/test/Microsoft.AspNet.Mvc.Core.Test/DefaultActionSelectorTests.cs
@@ -530,6 +530,7 @@ namespace Microsoft.AspNet.Mvc
         [InlineData("POST")]
         [InlineData("DELETE")]
         [InlineData("PATCH")]
+        [InlineData("HEAD")]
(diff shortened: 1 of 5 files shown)


dougbu:
Good so far but these attributes are also tested in `DefaultActionSelectorTests`, `HttpMethodProviderAttributesTests`, and `RouteTemplateProviderAttributesTest`. Should add `[HttpHead]` there as well.

In addition the `DefaultApplicationModelProviderTest` cases should include use of the `template` constructor parameter.

dougbu:
:watch:

test/Microsoft.AspNet.Mvc.Core.Test/DefaultActionSelectorTests.cs
```diff
@@ -873,22 +874,6 @@ public void RPCMethodWithHttpGet()
             }
         }
 
-        private class HttpMethodAttributeTests_DefaultMethodValidationController
```
ajaybhargavb: Removed because it was unused.

ajaybhargavb:
Updated.

dougbu:
:shipit:

Issue #2392, opened 2015-04-15
Is @Html.Raw(...) working? ... or am I just a nOOb?


GuardRex:
Trying to use `@Html.Raw(@ViewBag.xxx)` in a view to inject some raw HTML into the view, but I'm receiving: `RuntimeBinderException: 'Microsoft.AspNet.Mvc.Rendering.IHtmlHelper<object>' does not contain a definition for 'Raw'`

I'm new to MVC, so I'm not surprised if I'm just doing something wrong.

dougbu:
should work fine.  suspect the issue is the second `@`.  try `@Html.Raw(ViewBag.xxx)` instead.

GuardRex:
Thanks. I changed it, but I found out that ...
```
@Html.Raw("<meta name=\"xx\" content=\"xx\">")
```
works, but
```
@Html.Raw(ViewBag.Description)
```
throws the `RuntimeBinderException` even with that modification.

The string for that is simply in the top of the rendered body view:
```
@{
ViewBag.Description = "<meta name=\"description\" content=\"Description here\">";
}
```
Clearly, I'm learning MVC and not doing this right; however, `IHtmlHelper<object>` does contain a def for `Raw`, so getting that exception here seems incorrect regardless of how bad my coding is.

GuardRex:
Ok, so I get it now that
```
<meta name="description" content="@ViewBag.Description">
```
in the shared view with 
```
@{
ViewBag.Description = "Description here";
}
```
works. It still seems strange that I can't just inject the entire HTML tag into the view. I'll leave this open while my concern about that `RuntimeBinderException` being thrown for what I was doing is the correct exception ... and even perhaps to find out why `@Html.Raw(ViewBag.Description)` fails.

henkmollema:
Does casting `ViewBag.Description` to a string works when passing it to the `@Html.Raw()` method?
Something like:
```chsarp
@Html.Raw((string)ViewBag.Description))
```

Bartmax:
This is likely a bug:

```
@{ 
    ViewBag.Description = "Hey!";
}

<p>@ViewBag.Description</p> // works
<p>Raw: @Html.Raw(ViewBag.Description)</p> //throws modelbinder exception
<p>Raw: @Html.Raw(ViewBag.Description as string)</p> // works
```

henkmollema:
This does work as expected with MVC 5.

dougbu:
[...]

PR #2703, opened 2015-06-16
Add `Encode()` and `Raw()` overrides to `IHtmlHelper<TModel>`

dougbu:
- #2392
- `dynamic` does not work correctly when inherited from a base `interface`

### DIFF
--- a/test/Microsoft.AspNet.Mvc.Extensions.Test/Rendering/HtmlHelperTest.cs
+++ b/test/Microsoft.AspNet.Mvc.Extensions.Test/Rendering/HtmlHelperTest.cs
@@ -42,6 +42,98 @@ namespace Microsoft.AspNet.Mvc.Rendering
             }
         }
 
+        // value, expectedString
+        public static TheoryData<object, string> EncodeDynamicTestData
+        {
+            get
+            {
+                var data = new TheoryData<object, string>
+                {
+                    { null, string.Empty },
+                    // Dynamic implementation calls the string overload when possible.
+                    { string.Empty, string.Empty },
+                    { "<\">", "HtmlEncode[[<\">]]" },
+                    { "<br />", "HtmlEncode[[<br />]]" },
+                    { "<b>bold</b>", "HtmlEncode[[<b>bold</b>]]" },
+                    { new ObjectWithToStringOverride(), "HtmlEncode[[<b>boldFromObject</b>]]" },
+                };
+
+                return data;
+            }
+        }
+
+        // value, expectedString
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNet.Mvc.Extensions.Test/Rendering/HtmlHelperTest.cs
```diff
@@ -54,5 +146,135 @@ public class HtmlHelperTest
             var entry = Assert.Single(result);
             Assert.Equal(expectedEntry, entry);
         }
+
+        [Theory]
+        [MemberData(nameof(EncodeDynamicTestData))]
+        public void EncodeDynamic_ReturnsExpectedString(object value, string expectedString)
```
dougbu: Adding lots of new tests. This method and `RawDynamic_ReturnsExpectedString()` are the ones that fail without `IHtmlHelper<TModel>` change.

test/Microsoft.AspNet.Mvc.Extensions.Test/Rendering/HtmlHelperTest.cs
```diff
@@ -54,5 +146,135 @@ public class HtmlHelperTest
...
+        {
+            // Arrange
+            dynamic dynamicValue = value;
+            IHtmlHelper<DefaultTemplatesUtilities.ObjectTemplateModel> helper =
```
ajaybhargavb: `var` here and similar places below.

dougbu: No that would break the test or at least render it useless. Intent here is to make sure things work when the compiler treats the `HtmlHelper` instance as a specific `interface`.

ajaybhargavb: Ah. Got it.

ajaybhargavb: Maybe add a comment so that some one later wouldn't blindly `var` it.

Eilon: Yeah a comment to explain why both lines have explicit types would be good.

dougbu: NP

ajaybhargavb:
:shipit:

dougbu:
8b5931d

PR #2696, opened 2015-06-14
Add `ModelState` entries for greedy and type-matching model binders

dougbu:
- part II of II for #2445
- enable tests that #2445 was blocking
 - fix these and other tests expecting different `ModelState` entries

PR #2723, opened 2015-06-22
Add `ModelState` entries for greedy and type-matching model binders

dougbu:
- part II of II for #2445
- enable tests that #2445 was blocking
 - fix these and other tests expecting different `ModelState` entries

Supersedes #2696 and addresses comments on that PR


dougbu:
@rynowak please double-check this PR, commit f0ae34d and the new doc comments in [`ValueProviderResult`](https://github.com/aspnet/Mvc/pull/2723/files#diff-c980cada1fba5fea5f395f7773b4ced7L12) in particular. I've made all changes we discussed in #2696.

rynowak:
:shipit:

dougbu:
715a0b6

PR #2734, opened 2015-06-24
Use latest Mono (v4.0.1) for Travis CI

henkmollema:
Addresses the MyGet 'unknown header' error.

https://github.com/aspnet/dnx/pull/2113

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,4 @@
 language: csharp
 sudo: false
-mono:
-  - 3.12.0
 script:
   - ./build.sh --quiet verify


Eilon:
@sornaks any thoughts on this? Also @BrennanConroy because you were also looking at some Mono stuff.

BrennanConroy:
4.0.1 is fairly good, and its what people get by default so it would be good to run on it. Someone should look at the failures on travis before we decide to use this.

sornaks:
As discussed with @NTaylorMullen the other day, I was going to make travis run against their alpha feed. (4.0.2) That way we have more of their fixes. Once there is an official mono release with fixes to all the bugs which we have filed (most of the fixes are checked in their branch but we do not have an official release) we could move to their official release (from alpha). @Eilon what do you think?

Eilon:
@sornaks yeah I think we might as well try that. More fixes == better as far as I'm concerned.

PR #2736, opened 2015-06-24
Jsonpatch dynamics support

KevinDockx:
Support for dynamics for JsonPatch (after merging dev into jsonpatch branch).  

Quite a big change - split up in separate commits, comments on each commit & lots of comments in code.


rynowak:
Can you rebase your changes on top of dev instead? 

Can also send us a cliffnotes of the design changes here?

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -15,14 +15,15 @@
 namespace Microsoft.AspNet.JsonPatch.Adapters
 {
     /// <inheritdoc />
-    public class ObjectAdapter<TModel> : IObjectAdapter<TModel> where TModel : class
+    internal class ObjectAdapter : IObjectAdapter
```
kirthik: what did you change this to `internal`?

src/Microsoft.AspNet.JsonPatch/Adapters/IObjectAdapter.cs
```diff
@@ -7,14 +7,13 @@ namespace Microsoft.AspNet.JsonPatch.Adapters
 {
     /// <summary>
     /// Defines the operations that can be performed on a JSON patch document.
-    /// </summary>
-    /// <typeparam name="TModel">The type of the model.</typeparam>
-    public interface IObjectAdapter<TModel> where TModel : class
+    /// </summary>  
+    public interface IObjectAdapter
```
davidfowl: Why dynamic? Did you compare the before and after performance impact of this change?

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -107,15 +108,13 @@ public void Add(Operation<TModel> operation, TModel objectToApplyTo)
         /// Add is used by various operations (eg: add, copy, ...), yet through different operations;
         /// This method allows code reuse yet reporting the correct operation on error
         /// </summary>
-        private void Add(string path, object value, TModel objectToApplyTo, Operation<TModel> operationToReport)
+        private void Add(string path, object value, object objectToApplyTo, Operation operationToReport)
```
kirthik: why is `objectToApplyTo` `dynamic` in the above `Add` method and `object` in this one?

src/Microsoft.AspNet.JsonPatch/Converters/JsonPatchDocumentConverter.cs
```diff
@@ -0,0 +1,74 @@
...
+    {
+        public override bool CanConvert(Type objectType)
+        {
+            return true;
+        }
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, 
+            JsonSerializer serializer)
+        {
+
```
kirthik: remove blank line.

src/Microsoft.AspNet.JsonPatch/Converters/JsonPatchDocumentConverter.cs
```diff
@@ -0,0 +1,74 @@
+﻿using System;
```
kirthik: Need the license.

src/Microsoft.AspNet.JsonPatch/Adapters/IObjectAdapter.cs
KevinDockx: You're right - it doesn't have to be; will change.

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
KevinDockx: That's a mistake, it doesn't have to be - will change.

src/Microsoft.AspNet.JsonPatch/Converters/JsonPatchDocumentConverter.cs
KevinDockx: Ok, will do

src/Microsoft.AspNet.JsonPatch/Converters/JsonPatchDocumentConverter.cs
KevinDockx: Ok, will add license on top of file

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
KevinDockx: Hi Kirthi, I don't think there's a real use case for a consumer of the component to be able to use the IObjectAdapter implementation we provide outside of this component.  Either the consumer uses ApplyTo, which instantiates the ObjectAdapter, or he can provide his own implementation of an IObjectAdapter if he chooses to.

KevinDockx:
Incorporated changes as per comments

KevinDockx:
Pulled latest dev branch and rebased my changes onto it

KevinDockx:
Summary of the PR: the intent of this PR is to add support for dynamically typed objects to JsonPath.  Up until now, applying patch would only work correctly if the parent containing the property that's to be patched wasn't dynamic, eg: if the parent wasn't an IDictionary<string,object>.  With this change, that's supported as well.  

The main change to allow for this is the addition of of the ObjectTreeAnalysisResult.  This analyzes the passed-in object and checks whether or not we should use "dynamic" or "regular" logic to apply the operation.  Applying an operation when the parent is an IDictionary<string,object> effectively allows for adding new properties/removing existing properties from the object tree - it's closer in line with the spec where possible.

To summarize, this change enables the use case where patches can be applied to objects without forcing the user to create DTO's to apply them to (same for the creation of a JsonPatchDocument).

rynowak:
> Pulled latest dev branch and rebased my changes onto it

The base branch here should be `aspnet:dev`. Now that jsonpatch has made it into the mainline, we don't need to do development in a parallel branch.

rynowak:
As discussed, waiting on a smaller PR based on `dev`. Thanks! :+1:

PR #2737, opened 2015-06-24
React to Security/Http changes

HaoK:
https://github.com/aspnet/HttpAbstractions/pull/323
https://github.com/aspnet/Security/pull/283

cc @rynowak

### DIFF
--- a/test/Microsoft.AspNet.Mvc.FunctionalTests/FiltersTest.cs
+++ b/test/Microsoft.AspNet.Mvc.FunctionalTests/FiltersTest.cs
@@ -155,7 +155,7 @@ namespace Microsoft.AspNet.Mvc.FunctionalTests
                 "http://localhost/AuthorizeUser/Impossible");
 
             // Assert
-            Assert.Equal(HttpStatusCode.Unauthorized, response.StatusCode);
+            Assert.Equal(HttpStatusCode.Forbidden, response.StatusCode);
         }
 
         [Fact]
--- a/test/WebSites/FiltersWebSite/BasicAuthenticationHandler.cs
+++ b/test/WebSites/FiltersWebSite/BasicAuthenticationHandler.cs
@@ -24,7 +24,8 @@ namespace FiltersWebSite
                     new Claim(ClaimTypes.NameIdentifier, "John")
                 },
                 Options.AuthenticationScheme));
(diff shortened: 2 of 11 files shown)


test/WebSites/FiltersWebSite/BasicAuthenticationHandler.cs
```diff
@@ -32,7 +24,7 @@ protected override AuthenticationTicket AuthenticateCore()
                     new Claim(ClaimTypes.NameIdentifier, "John")
                 },
                 Options.AuthenticationScheme));
-            return new AuthenticationTicket(principal, new AuthenticationProperties(), Options.AuthenticationScheme);
+            return Task.FromResult(new AuthenticationTicket(principal, new AuthenticationProperties(), Options.AuthenticationScheme));
```
rynowak: Break up long line

test/Microsoft.AspNet.Mvc.FunctionalTests/FiltersTest.cs
```diff
@@ -155,7 +155,7 @@ public class FiltersTest
                 "http://localhost/AuthorizeUser/Impossible");
 
             // Assert
-            Assert.Equal(HttpStatusCode.Unauthorized, response.StatusCode);
+            Assert.Equal(HttpStatusCode.Forbidden, response.StatusCode);
```
rynowak: How intended is this behavior change?

HaoK: This is actually finally the correct behavior, when a policy fails with a missing claim, you want 403 forbidden (not unauthorized).  This finally fixes it (for one path).  There's still an issue which currently isn't shown in the mvc tests)

rynowak: Ok, cool. I'll sign off on this change, but can you send a PR to add the missing test with `[Fact(Skip = "blah blah")]`. I don't think anyone else is aware of the bug.

rynowak:
:shipit:

test/Microsoft.AspNet.Mvc.FunctionalTests/FiltersTest.cs
HaoK: Yeah this is tracking the case that fails https://github.com/aspnet/Security/issues/300

HaoK:
7c426f7835731e620f7dd0d03f58bf63ab33d301

Issue #2715, opened 2015-06-18
Revisit IJsonOutputFormatter 


danroth27:
`IJsonOutputFormatter` is currently just used as a marker interface to indicate that a formatter outputs JSON. We should revisit this approach and see if we can come up with something cleaner and more general.

danroth27:
We decided to remove the dependency on formatting in `JsonResult` and `JsonViewComponentResult` and instead just use JSON.NET directly. Remove `IJsonOutputFormatter` as part of this work.

ajaybhargavb:
Part 1 - f206cf6759dd09920c8a30cd98e2800e70e878cf

JamesNK:
Are the serializer settings from `MvcJsonOptions` still used by default?

rynowak:
I look a look here and @JamesNK is right. This should be respecting `MvcJsonOptions`

PR #2738, opened 2015-06-24
Removed formatter dependency from JsonResult and JsonViewComponentResult

ajaybhargavb:
Issue - #2715 
@pranavkm @rynowak


src/Microsoft.AspNet.Mvc.Extensions/JsonResult.cs
```diff
@@ -81,80 +59,44 @@ public JsonResult(object value, IOutputFormatter formatter)
...
-                    // If the available user-configured formatters can't write this type, then fall back to the
-                    // 'global' one.
-                    formatter = formatterContext
-                        .HttpContext
-                        .RequestServices
-                        .GetRequiredService<JsonOutputFormatter>();
-
-                    // Run SelectFormatter again to try to choose a content type that this formatter can do.
-                    objectResult.SelectFormatter(formatterContext, new[] { formatter });
+                    jsonWriter.CloseOutput = false;
```
rynowak: Does this even matter?

src/Microsoft.AspNet.Mvc.Extensions/JsonResult.cs
```diff
@@ -18,13 +15,11 @@ namespace Microsoft.AspNet.Mvc
     /// </summary>
     public class JsonResult : ActionResult
```
rynowak: So much simpler

src/Microsoft.AspNet.Mvc.Extensions/ViewComponents/JsonViewComponentResult.cs
```diff
@@ -74,11 +65,5 @@ public Task ExecuteAsync([NotNull] ViewComponentContext context)
             Execute(context);
             return Task.FromResult(true);
         }
-
-        private static JsonOutputFormatter ResolveFormatter(ViewComponentContext context)
-        {
-            var services = context.ViewContext.HttpContext.RequestServices;
-            return services.GetRequiredService<JsonOutputFormatter>();
```
rynowak: You should be able to remove this from the service registration now right?

rynowak:
:watch:

src/Microsoft.AspNet.Mvc.Extensions/JsonResult.cs
```diff
@@ -81,80 +59,44 @@ public JsonResult(object value, IOutputFormatter formatter)
...
             }
             else
             {
-                objectResult.ContentTypes = ContentTypes;
+                if (contentTypeHeader.Encoding == null)
+                {
+                    // 1. Do not modify the user supplied content type
+                    // 2. Parse here to handle parameters apart from charset
+                    contentTypeHeader = MediaTypeHeaderValue.Parse(contentTypeHeader.ToString());
+                    contentTypeHeader.Encoding = Encoding.UTF8;
```
IDisposable: What if the content type header contains the encoding?

darrelmiller: Technically it shouldn't.  Despite its widespread usage, the charset parameter is not defined for application/json.  https://tools.ietf.org/html/rfc7159#section-11

ajaybhargavb: @IDisposable, I don't understand your concern here. This line won't be hit if the content type header already contains the encoding.

src/Microsoft.AspNet.Mvc.Extensions/JsonResult.cs
```diff
@@ -18,13 +15,11 @@ namespace Microsoft.AspNet.Mvc
     /// </summary>
     public class JsonResult : ActionResult
     {
-        /// <summary>
-        /// The list of content-types used for formatting when <see cref="ContentTypes"/> is null or empty.
-        /// </summary>
-        public static readonly IReadOnlyList<MediaTypeHeaderValue> DefaultContentTypes = new MediaTypeHeaderValue[]
+        private JsonSerializerSettings _serializerSettings;
```
pranavkm: `readonly`?

src/Microsoft.AspNet.Mvc.Extensions/JsonResult.cs
```diff
@@ -18,13 +15,11 @@ namespace Microsoft.AspNet.Mvc
...
+
+        private readonly MediaTypeHeaderValue DefaultContentType = new MediaTypeHeaderValue("application/json")
```
pranavkm: `static readonly`

test/Microsoft.AspNet.Mvc.Extensions.Test/ViewComponents/JsonViewComponentResultTest.cs
```diff
@@ -18,29 +18,12 @@ namespace Microsoft.AspNet.Mvc
...
-            // Assert
-            Assert.Equal(expectedFormatter, result.Formatter);
-            Assert.Equal("1", new StreamReader(buffer).ReadToEnd());
-        }
-
-        [Fact]
         public void Execute_UsesFormatter_WithSpecifiedSerializerSettings()
         {
             // Arrange
+            var abcdIndentedUTF8Bytes
```
pranavkm: Could you use strings instead? `Assert.Equal(abcdIndented, Encoding.UTF8.GetString(body.ToArray())` - easier to understand what's happening

pranavkm:
:shipit:

src/Microsoft.AspNet.Mvc.Extensions/ViewComponents/JsonViewComponentResult.cs
ajaybhargavb: No. JsonHelper fails without this.

src/Microsoft.AspNet.Mvc.Extensions/JsonResult.cs
ajaybhargavb: spoke to @kichalla, it is needed so we don't close the underlying response stream. MVC 5 has the same behavior.

ajaybhargavb:
Updated.

rynowak:
:shipit:

src/Microsoft.AspNet.Mvc.Extensions/JsonResult.cs
IDisposable: Sorry, didn't read up higher. Reading @darrelmiller link it looks like it should be assumed to be UTF-8 (and the only other valid encodings would be UTF16 or UTF32, which should have been specified) https://tools.ietf.org/html/rfc7159#section-8.1

PR #2744, opened 2015-06-25
Tweak lifetimes for a few commonly resolved services

rynowak:
This is some low hanging fruit for reducing the number of resolves we have
per request.

DefaultHtmlGenerator: Lots of these are created by RazorPage. It needs
IUrlHelper, so scoped is the best we can do for now. For an example, on
the front page of our sample, 48 of these are created for each request.
48! This takes it down to 1-per-request.

JsonResult: Again, multiple created per request (12 for the sample). This
class is totally stateless, so we can get down to 0-per-request.

DefaultViewComponentInvokerFactory: Same story as JsonResult.

DefaultObjectValidator/MvcMarkerService these are stateless and pretty
much guaranteed to be used by every request. Getting them off the table.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/MvcCoreServiceCollectionExtensions.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/MvcCoreServiceCollectionExtensions.cs
@@ -78,24 +78,28 @@ namespace Microsoft.Framework.DependencyInjection
                 services,
                 ServiceDescriptor.Transient<IActionConstraintProvider, DefaultActionConstraintProvider>());
 
-            // Action Invoker
+            // Controller Factory
             //
             // This has a cache, so it needs to be a singleton
             services.TryAdd(ServiceDescriptor.Singleton<IControllerFactory, DefaultControllerFactory>());
+            // These will be cached by the controller factory
             services.TryAdd(ServiceDescriptor.Transient<IControllerActivator, DefaultControllerActivator>());
-            // This accesses per-request services
-            services.TryAdd(ServiceDescriptor.Transient<IActionInvokerFactory, ActionInvokerFactory>());
-            services.TryAdd(ServiceDescriptor
-                .Transient<IControllerActionArgumentBinder, DefaultControllerActionArgumentBinder>());
-            TryAddMultiRegistrationService(
-                services,
-                ServiceDescriptor.Transient<IActionInvokerProvider, ControllerActionInvokerProvider>());
-            TryAddMultiRegistrationService(
-                services,
-                ServiceDescriptor.Transient<IFilterProvider, DefaultFilterProvider>());
             TryAddMultiRegistrationService(
                 services,
                 ServiceDescriptor.Transient<IControllerPropertyActivator, DefaultControllerPropertyActivator>());
+
+            // Action Invoker
+            //
+            // These two access per-request services
+            services.TryAdd(ServiceDescriptor.Transient<IActionInvokerFactory, ActionInvokerFactory>());
+            TryAddMultiRegistrationService(
+                services,
+                ServiceDescriptor.Transient<IActionInvokerProvider, ControllerActionInvokerProvider>());
+            // Stateless
+            services.TryAddSingleton<IControllerActionArgumentBinder, DefaultControllerActionArgumentBinder>();
+            TryAddMultiRegistrationService(
+                services,
+                ServiceDescriptor.Singleton<IFilterProvider, DefaultFilterProvider>());
 
             // ModelBinding, Validation and Formatting
             //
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNet.Mvc/MvcServiceCollectionExtensions.cs
```diff
@@ -261,7 +261,7 @@ internal static void AddMvcServices(IServiceCollection services)
             services.TryAdd(ServiceDescriptor
                 .Singleton<IAntiForgeryAdditionalDataProvider, DefaultAntiForgeryAdditionalDataProvider>());
 
-            // Api Description
+            // Api Description
```
IDisposable: remove trailing blank the :shipit:

rynowak:
For the genuinely curious, here's my raw data from 5 requests on the sample. I excluded startup.

Before:
```
**DefaultHtmlGenerator 240**
HtmlHelper 180
AutofacServiceProvider 97
**DefaultViewComponentInvokerFactory 60**
**JsonHelper 60**
DefaultViewComponentHelper 60
HtmlHelper<Object> 40
HtmlHelper<User> 15
AutofacServiceScopeFactory 12
**IObjectModelValidator 10**
RazorViewFactory 5
IFilterProvider[] 5
ControllerActionInvokerProvider 5
UrlHelper 5
MvcRazorHost 5
FormatFilter 5
ScopedInstance<ActionContext> 5
ScopedInstance<ActionBindingContext> 5
DefaultControllerActionArgumentBinder 5
ViewStartProvider 5
CompositeViewEngine 5
VirtualPathRazorPageFactory 5
DefaultFilterProvider 5
**MvcMarkerService 5**
RazorCompilationService 5
HtmlHelper<String[]> 5
ActionInvokerFactory 5
TempDataDictionary 5
IActionInvokerProvider[] 5
```

After:
```
HtmlHelper 180
DefaultViewComponentHelper 60
HtmlHelper<Object> 40
AutofacServiceProvider 26
HtmlHelper<User> 15
AutofacServiceScopeFactory 11
FormatFilter 5
ActionInvokerFactory 5
TempDataDictionary 5
ScopedInstance<ActionBindingContext> 5
HtmlHelper<String[]> 5
DefaultFilterProvider 5
IFilterProvider[] 5
ScopedInstance<ActionContext> 5
ControllerActionInvokerProvider 5
IActionInvokerProvider[] 5
CompositeViewEngine 5
DefaultControllerActionArgumentBinder 5
ViewStartProvider 5
VirtualPathRazorPageFactory 5
MvcRazorHost 5
**DefaultHtmlGenerator 5**
RazorCompilationService 5
RazorViewFactory 5
UrlHelper 5
```

Since there are 5 requests in the data, anything with a 5 by it is *effectively scoped* meaning that it's resolved/created once per request. I haven't looked at every item on the list, but this generally means that they touch `IScopedInstance<ActionContext>` transitively.

Anything with a higher count (HtmlHelper) is likely tied to each view that's invoked. Improving this would require some **substantial** change.

I haven't yet looked into the higher number of resolves for `AutofacServiceProvider` and `AutofacServiceScopeFactory`

rynowak:
By the way, the vast majority of the non-generic `HtmlHelper` instances come from `TemplateRenderer.MakeHtmlHelper`. This will be called once-per-display/editor for simple models and once-per-property/item for complex models (hence 180 instances).

dougbu:
> By the way, the vast majority ...

That makes sense but could probably be fixed with some API changes in template rendering. Today each property corresponds to `ViewDataDictionary`, `ViewContext`, and `HtmlHelper` instances; `HtmlHelper.Contextualize()` is never called twice.

Theoretically only the `ViewDataDictionary` needs to be unique (to avoid `ViewData` changes in a template for `Name` affecting a template for `Zip`). Even the `ViewDataDictionary` instances might be reused (per type) as long as the system doesn't call out to user code -- but that's getting complex.

src/Microsoft.AspNet.Mvc.Core/MvcCoreServiceCollectionExtensions.cs
```diff
@@ -78,24 +78,28 @@ internal static void AddMvcCoreServices(IServiceCollection services)
                 services,
                 ServiceDescriptor.Transient<IActionConstraintProvider, DefaultActionConstraintProvider>());
 
-            // Action Invoker
+            // Controller Factory
             //
             // This has a cache, so it needs to be a singleton
             services.TryAdd(ServiceDescriptor.Singleton<IControllerFactory, DefaultControllerFactory>());
+            // These will be cached by the controller factory
```
dougbu: Does this comment apply to both `DefaultControllerActivator` and `DefaultControllerPropertyActivator`? Break up with a blank line if not. Add a blank line above either way.

src/Microsoft.AspNet.Mvc/MvcServiceCollectionExtensions.cs
```diff
@@ -238,9 +238,9 @@ internal static void AddMvcServices(IServiceCollection services)
             // Consumed by the Cache tag helper to cache results across the lifetime of the application.
             services.TryAdd(ServiceDescriptor.Singleton<IMemoryCache, MemoryCache>());
 
-            // DefaultHtmlGenerator is pretty much stateless but depends on Scoped services such as IUrlHelper and
-            // IActionBindingContextProvider. Therefore it too is scoped.
-            services.TryAdd(ServiceDescriptor.Transient<IHtmlGenerator, DefaultHtmlGenerator>());
+            // DefaultHtmlGenerator is pretty much stateless but depends on IUrlHelper, which is scoped.
+            // Therefore it too is scoped.
+            services.TryAdd(ServiceDescriptor.Scoped<IHtmlGenerator, DefaultHtmlGenerator>());
```
dougbu: Nice. Looks like the intent was `Scoped` all along.

dougbu:
:shipit:

src/Microsoft.AspNet.Mvc.Core/MvcCoreServiceCollectionExtensions.cs
rynowak: Cleared this up in the comments

PR #2751, opened 2015-06-26
React to newly added media type header value Copy and CopyReadOnly apis

kichalla:
@rynowak @Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Extensions/FormatterMappings.cs
+++ b/src/Microsoft.AspNet.Mvc.Extensions/FormatterMappings.cs
@@ -27,7 +27,7 @@ namespace Microsoft.AspNet.Mvc
         {
             ValidateContentType(contentType);
             format = RemovePeriodIfPresent(format);
-            _map[format] = MediaTypeHeaderValue.Parse(contentType.ToString());
+            _map[format] = contentType.CopyAsReadOnly();
         }
 
         /// <summary>
--- a/src/Microsoft.AspNet.Mvc.Extensions/Formatters/MediaTypeHeaderValues.cs
+++ b/src/Microsoft.AspNet.Mvc.Extensions/Formatters/MediaTypeHeaderValues.cs
@@ -1,0 +1,19 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNet.Mvc
+{
+    internal class MediaTypeHeaderValues
+    {
+        public static readonly MediaTypeHeaderValue ApplicationJson
+            = MediaTypeHeaderValue.Parse("application/json").CopyAsReadOnly();
+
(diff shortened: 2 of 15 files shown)


src/Microsoft.AspNet.Mvc.Extensions/FormatterMappings.cs
```diff
@@ -27,7 +27,7 @@ public void SetMediaTypeMappingForFormat([NotNull] string format, [NotNull] Medi
         {
             ValidateContentType(contentType);
             format = RemovePeriodIfPresent(format);
-            _map[format] = MediaTypeHeaderValue.Parse(contentType.ToString());
+            _map[format] = contentType.CopyAsReadOnly();
```
kichalla: FYI...this looks different as here we have a chance to intercept the mediatype and convert to readonly copy before putting in the dictionary...but in case of formatters, the media type header value is directly aded to the list...

To keep it consistent with formatters, i could just leave this line of code to be `_map[format] = contentType;` as we are anyway creating a copy in output formatter, thoughts?

Tratcher: Making it read-only when you can is good.

src/Microsoft.AspNet.Mvc.Extensions/JsonResult.cs
```diff
@@ -72,9 +72,8 @@ public override Task ExecuteResultAsync([NotNull] ActionContext context)
             {
                 if (contentTypeHeader.Encoding == null)
                 {
-                    // 1. Do not modify the user supplied content type
-                    // 2. Parse here to handle parameters apart from charset
-                    contentTypeHeader = MediaTypeHeaderValue.Parse(contentTypeHeader.ToString());
+                    // Do not modify the user supplied content type, so clone it instead
```
Tratcher: Copy, not clone.

kichalla: Sure..changed it now.

Tratcher:
:shipit:

rynowak:
would be good to see our formatters insert readonly mediatypes

kichalla:
@rynowak I update this PR with the change you mentioned.
/cc: @Tratcher

src/Microsoft.AspNet.Mvc.Extensions/Formatters/MediaTypeHeaderValues.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    internal class MediaTypeHeaderValues
+    {
+        public static readonly MediaTypeHeaderValue ApplicationJson
+            = MediaTypeHeaderValue.Parse("application/json").CopyAsReadOnly();
+
+        public static readonly MediaTypeHeaderValue TextJson
+            = MediaTypeHeaderValue.Parse("text/json").CopyAsReadOnly();
+
+        public static readonly MediaTypeHeaderValue ApplicationJsonPatch
```
rynowak: Don't these already exist somewhere? /cc @Tratcher

Issue #2750, opened 2015-06-26
ImageTagHelper append file version does not work for dynamic or external src paths


ajaybhargavb:
Example:
```html
<img asp-append-version="true" src="/Home/GetImage/" />
<img asp-append-version="true" src="http://www.some.com/thing.png" />
```
Need to figure out what the experience should be for these scenarios.

ajaybhargavb:
cc @Eilon @DamianEdwards

DamianEdwards:
We can either ignore it, log a warning, or throw an exception. In this case given that the path could be a variable value, throwing seems extreme, so I'd lean to ignoring it or logging a warning.

ajaybhargavb:
Currently the first example ignores it but the second one throws an exception when the file provider tries to get FileInfo. We need to change that behavior.

DamianEdwards:
The helper should be ensuring that the string is a valid file path before attempting to load it

ajaybhargavb:
6213354b85ee4b0e7c42d0eb64fc76447e28e9fb

Issue #2719, opened 2015-06-18
Append file version does not work when href/src contains a url with query string


ajaybhargavb:
Example:

This works
`<img src="images/sample.png" asp-file-version="true" alt="some text" />`
Output:
`<img src="images/sample.png?v=M9cy9a-9lV1-MFfOg2JDzIzBsWiFUVl17uHAR0cgpTE" alt="some text" />`

This produces 500
`<img src="images/sample.png?some=value" asp-file-version="true" alt="some text" />`

Expected output:
`<img src="images/sample.png?some=value&v=M9cy9a-9lV1-MFfOg2JDzIzBsWiFUVl17uHAR0cgpTE" alt="some text" />`

Same applies for Link and Script tag helper

ajaybhargavb:
6213354b85ee4b0e7c42d0eb64fc76447e28e9fb

PR #2760, opened 2015-06-29
Fixes for FileVersionProvider

ajaybhargavb:
Issues #2719 and #2750 
 - Path with query string works
 - No exception is thrown for absolute path

@Eilon @DamianEdwards

### DIFF
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
@@ -48,6 +48,20 @@ namespace Microsoft.AspNet.Mvc.TagHelpers.Internal
         public string AddFileVersionToPath([NotNull] string path)
         {
             var resolvedPath = path;
+
+            var queryStringStartIndex = path.IndexOf('?');
+            if (queryStringStartIndex != -1)
+            {
+                resolvedPath = path.Substring(0, queryStringStartIndex);
+            }
+
+            Uri uri;
+            if (Uri.TryCreate(resolvedPath, UriKind.Absolute, out uri))
+            {
+                // Don't append version if the path is absolute.
+                return path;
+            }
+
             var fileInfo = _fileProvider.GetFileInfo(resolvedPath);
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
```diff
@@ -47,7 +47,15 @@ public class FileVersionProvider
         /// </remarks>
         public string AddFileVersionToPath([NotNull] string path)
         {
-            var resolvedPath = path;
+            var resolvedPath = path.Split('?')[0];
```
pranavkm: Use `IndexOf` \ `SubString` instead - no point allocating an array if you don't care about the result.

pranavkm:
:shipit: once somebody else signs off

test/WebSites/HtmlGenerationWebSite/Views/HtmlGeneration_Home/Image.cshtml
```diff
@@ -13,10 +13,19 @@
...
     <img src="~/images/red.png" alt="Red explicitly not versioned" title="Red versioned" asp-append-version="false" />
+
+    <!-- Plain image tag with absolute path and file version -->
+    <img src="http://contoso.com/hello/world" alt="Absolute path versioned" asp-append-version="true" />
+
+    <!-- Plain image tag with file version and path to file that does not exist -->
+    <img src="~/images/fake.png" alt="Path to non existing file" asp-append-version="true" />
+
+    <!-- Plain image tag with file version and path containing query string -->
+    <img src="~/images/red.png?abc=def" alt="Path with query string" asp-append-version="true" />
```
Eilon: What happens with `src="/foo/bar"` for the same types of cases?

ajaybhargavb: It will just return it without adding file version. I've added that case to the test.

ajaybhargavb:
Updated.

Eilon:
:shipit: thanks!

Issue #2512, opened 2015-05-06
TagBuilder.CreateSanitizedId must return without allocating objects when no change is needed.


sornaks:
Found this when doing perf changes. CreateSanitizedId call seems to be allocating few megabytes of data (tested for 1000 requests). This function can be modified to check the input `id` and return without allocating objects if no modifications are needed to the `id`.

sornaks:
https://github.com/aspnet/Mvc/commit/57f1ae755f1cf4afbcc5d7dd5436b33befdbdfa3

PR #2761, opened 2015-06-29
Optimizing TagBuilder.CreateSanitizedId to create buffer only when needed.

sornaks:
https://github.com/aspnet/Mvc/issues/2512

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Extensions/Rendering/Html/TagBuilder.cs
+++ b/src/Microsoft.AspNet.Mvc.Extensions/Rendering/Html/TagBuilder.cs
@@ -77,16 +77,46 @@ namespace Microsoft.AspNet.Mvc.Rendering
                 return string.Empty;
             }
 
+            // If there are no invalid characters in the string, then we don't have to create the buffer.
+            var i = 0;
+            for (i = 1; i < name.Length; i++)
+            {
+                if (!Html401IdUtil.IsValidIdCharacter(name[i]))
+                {
+                    break;
+                }
+            }
+
             var firstChar = name[0];
-            if (!Html401IdUtil.IsAsciiLetter(firstChar))
+            var startsWithAsciiLetter = Html401IdUtil.IsAsciiLetter(firstChar);
+            if (!startsWithAsciiLetter)
             {
                 // The first character must be a letter according to the HTML 4.01 specification.
                 firstChar = 'z';
             }
 
+            if (i == name.Length)
+            {
+                if (startsWithAsciiLetter)
+                {
+                    return name;
+                }
+                else
+                {
+                    return firstChar + name.Substring(1);
+                }
+            }
+
             var stringBuffer = new StringBuilder(name.Length);
             stringBuffer.Append(firstChar);
-            for (var index = 1; index < name.Length; index++)
+
+            // Characters until 'i' have already been checked for validity. Hence just copying them.
+            for (var index = 1; index < i; index ++)
+            {
+                stringBuffer.Append(name[index]);
+            }
+
+            for (var index = i; index < name.Length; index++)
             {
                 var thisChar = name[index];
                 if (Html401IdUtil.IsValidIdCharacter(thisChar))
--- a/src/Microsoft.AspNet.Mvc.Extensions/Rendering/Html/TagBuilder.cs
+++ b/src/Microsoft.AspNet.Mvc.Extensions/Rendering/Html/TagBuilder.cs
@@ -78,10 +78,10 @@ namespace Microsoft.AspNet.Mvc.Rendering
             }
 
             // If there are no invalid characters in the string, then we don't have to create the buffer.
-            var i = 0;
-            for (i = 1; i < name.Length; i++)
+            var firstIndexOfInvalidCharacter = 1;
+            for (; firstIndexOfInvalidCharacter < name.Length; firstIndexOfInvalidCharacter++)
             {
-                if (!Html401IdUtil.IsValidIdCharacter(name[i]))
+                if (!Html401IdUtil.IsValidIdCharacter(name[firstIndexOfInvalidCharacter]))
                 {
                     break;
                 }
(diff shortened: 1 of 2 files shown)


dougbu:
#2512 mentions "a few megabytes". But I see nothing about how much of that memory is saved with this fix. Since the `id` will need to be sanitized a large percentage of the time, my expectation is the added complexity does not buy us much.

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.Extensions/Rendering/Html/TagBuilder.cs
```diff
@@ -77,16 +77,46 @@ public static string CreateSanitizedId(string name, [NotNull] string invalidChar
                 return string.Empty;
             }
 
+            // If there are no invalid characters in the string, then we don't have to create the buffer.
+            var i = 0;
+            for (i = 1; i < name.Length; i++)
```
rynowak: why initialize it twice?

src/Microsoft.AspNet.Mvc.Extensions/Rendering/Html/TagBuilder.cs
```diff
@@ -77,16 +77,46 @@ public static string CreateSanitizedId(string name, [NotNull] string invalidChar
...
+
             var stringBuffer = new StringBuilder(name.Length);
             stringBuffer.Append(firstChar);
-            for (var index = 1; index < name.Length; index++)
+
+            // Characters until 'i' have already been checked for validity. Hence just copying them.
+            for (var index = 1; index < i; index ++)
+            {
+                stringBuffer.Append(name[index]);
+            }
```
rynowak: My guess is that handling this extra case here doesn't have any measurable benefit. if you find that it's a meaningful optimization then document it.

src/Microsoft.AspNet.Mvc.Extensions/Rendering/Html/TagBuilder.cs
```diff
@@ -77,16 +77,46 @@ public static string CreateSanitizedId(string name, [NotNull] string invalidChar
...
 
+            if (i == name.Length)
+            {
+                if (startsWithAsciiLetter)
+                {
+                    return name;
+                }
+                else
+                {
+                    return firstChar + name.Substring(1);
```
rynowak: Profiling helps. Need to see data. For instance, I assume you haven't measured, because this *optimization* would allocate more than letting it fall through to the string builder case.

src/Microsoft.AspNet.Mvc.Extensions/Rendering/Html/TagBuilder.cs
```diff
@@ -77,16 +77,46 @@ public static string CreateSanitizedId(string name, [NotNull] string invalidChar
...
+                }
+            }
+
             var stringBuffer = new StringBuilder(name.Length);
             stringBuffer.Append(firstChar);
-            for (var index = 1; index < name.Length; index++)
+
+            // Characters until 'i' have already been checked for validity. Hence just copying them.
+            for (var index = 1; index < i; index ++)
```
rynowak: really tricky to see `i` and `index` in the same piece of code. The right thing to do would be to choose a more semantically meaningful name like `firstIndexOfInvalidCharacter` and then use `i` where you need a loop counter.

src/Microsoft.AspNet.Mvc.Extensions/Rendering/Html/TagBuilder.cs
rynowak: And by `extra case` I mean splitting the building process into 'clean' and 'unclean' parts. What we're really after here was to remove the allocation when the string was already safe.

sornaks:
<b>Perf measurements:</b>
Used the following View:
```html
@model CreateSanitizedId.User

<form asp-action="Create" asp-controller="Home">
    @Html.Label("Hello¡World", "Hello¡World", null)
    @Html.Label("¡HelloWorld", "Hello¡World", null)
    @Html.EditorForModel()
    <input type="submit" value="ok" />
</form>
```
Used the following model:
````c#
public class User
{
    [Required]

    [Required]
    public string Occupation { get; set; }
    }
````
Around 33% of the page has invalid Ids which go through TagBuilder.CreateSanitizedId. I ran 10,000 requests and measured perf.
<b>Data without the change:</b>
System.String allocated by CreateSanitizedId: 1.84 mb
System.Text.StringBuilder allocated by CreateSanitizedId: 1.68 mb

<b>Data after the change:</b>
System.String allocated by CreateSanitizedId: 0.720 mb
System.Text.StringBuilder allocated by CreateSanitizedId: 0.560 mb

This is around 60% improvement from the original case.

File: src/Microsoft.AspNet.Mvc.Extensions/Rendering/Html/TagBuilder.cs (hunk 2 above)
Comment (sornaks) [inline src/Microsoft.AspNet.Mvc.Extensions/Rendering/Html/TagBuilder.cs]: Yes this does not save space. But it avoids checking `Html401IdUtil.IsValidIdCharacter` again for the 'clean' part. Explained it in the comment..

Comment (rynowak) [inline src/Microsoft.AspNet.Mvc.Extensions/Rendering/Html/TagBuilder.cs]: > But it avoids checking Html401IdUtil.IsValidIdCharacter again for the 'clean' part. 

What's the **measured** impact of that?

File: src/Microsoft.AspNet.Mvc.Extensions/Rendering/Html/TagBuilder.cs [hunk 5]
```diff
@@ -77,16 +77,40 @@ public static string CreateSanitizedId(string name, [NotNull] string invalidChar
...
+                return name;
+            }
+
             var stringBuffer = new StringBuilder(name.Length);
             stringBuffer.Append(firstChar);
-            for (var index = 1; index < name.Length; index++)
+
+            // Characters until 'firstIndexOfInvalidCharacter' have already been checked for validity.
+            // So just copying them. This avoids running them through Html401IdUtil.IsValidIdCharacter again.
+            for (var index = 1; index < firstIndexOfInvalidCharacter; index ++)
```
Eilon: Fix spacing before the `++`

Eilon:
Looks good to me :shipit:

src/Microsoft.AspNet.Mvc.Extensions/Rendering/Html/TagBuilder.cs
sornaks: It saves around 0.95s for 100,000 requests (with 200 in parallel) :)

rynowak:
:shipit:

sornaks:
Thanks! https://github.com/aspnet/Mvc/commit/57f1ae755f1cf4afbcc5d7dd5436b33befdbdfa3

PR #2765, opened 2015-06-30
Using default SerializerSettings from MvcJsonOptions

ajaybhargavb:
Issue - #2715 
@rynowak

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Extensions/JsonResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Extensions/JsonResult.cs
@@ -3,8 +3,9 @@
 
 using System.Text;
 using System.Threading.Tasks;
-using Microsoft.AspNet.Mvc.Internal;
+using Microsoft.Framework.DependencyInjection;
 using Microsoft.Framework.Internal;
+using Microsoft.Framework.OptionsModel;
 using Microsoft.Net.Http.Headers;
 using Newtonsoft.Json;
 
@@ -27,8 +28,8 @@ namespace Microsoft.AspNet.Mvc
         /// </summary>
         /// <param name="value">The value to format as JSON.</param>
         public JsonResult(object value)
-            : this(value, serializerSettings: SerializerSettingsProvider.CreateSerializerSettings())
         {
+            Value = value;
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNet.Mvc.Extensions/JsonResult.cs
```diff
@@ -86,6 +87,16 @@ public override Task ExecuteResultAsync([NotNull] ActionContext context)
                 response.StatusCode = StatusCode.Value;
             }
 
+            if (_serializerSettings == null)
```
rynowak: I don't really mind too much because these result types are single-use objects, but it's cleaner to avoid mutating the state of the object and just use a local variable. 

As a general principle, it makes debugger harder if you have to think about *what's changed* after calling a method.

test/Microsoft.AspNet.Mvc.Extensions.Test/JsonResultTest.cs
```diff
@@ -109,15 +112,23 @@ public class JsonResultTest
...
             httpContext.Response.Body = new MemoryStream();
 
+            var mockJsonOptions = new Mock<IOptions<MvcJsonOptions>>();
+            mockJsonOptions.Setup(m => m.Options).Returns(new MvcJsonOptions());
+
+            var serviceProvider = new Mock<IServiceProvider>();
+            serviceProvider.Setup(s => s.GetService(typeof(IOptions<MvcJsonOptions>)))
+                .Returns(mockJsonOptions.Object);
+            httpContext.RequestServices = serviceProvider.Object;
+
```
rynowak: This is also a case where it's pretty easy to avoid using Moq if you're so inclined. Not really important, but food for thought. You can create an `IServiceCollection` and call `AddOptions` and that would do what you need.

test/Microsoft.AspNet.Mvc.Extensions.Test/ViewComponents/JsonViewComponentResultTest.cs
```diff
@@ -37,9 +40,27 @@ public void Execute_UsesFormatter_WithSpecifiedSerializerSettings()
             Assert.Equal("{\r\n  \"foo\": \"abcd\"\r\n}", Encoding.UTF8.GetString(buffer.ToArray()));
         }
 
+        [Fact]
+        public void Execute_UsesDefaultSerializerSettings_IfNotProvided()
```
rynowak: It might be better to way `..._UsesSerializerSettingsFromOptions_...` - `Default` kinda fits the old semantic.

ajaybhargavb:
Updated.

src/Microsoft.AspNet.Mvc.Extensions/JsonResult.cs
```diff
@@ -15,7 +16,7 @@ namespace Microsoft.AspNet.Mvc
     /// </summary>
     public class JsonResult : ActionResult
     {
-        private readonly JsonSerializerSettings _serializerSettings;
+        private JsonSerializerSettings _serializerSettings;
```
rynowak: You should be able to leave this as readonly now.

rynowak:
:shipit:

Issue #2720, opened 2015-06-19
When submitting empty value types, ModelState has a NullRef exception


sornaks:
The view looks like this:
```html
<form asp-action="Create" asp-controller="Home">
@Html.EditorForModel()
<input type="submit" value="ok" />
</form>
```
The Model is this:
```C#
public class Person
{
    [Required]
    public string Name { get; set; }
    public decimal Height { get; set; }
    public double Weight { get; set; }
    public float BMI { get; set; }	
    [Range(1, 40)]
    public short Age { get; set; }
    public ushort IsWorking { get; set; }
}
```
NullRef is thrown for - `Height`, `Weight` etc. This is the stack trace in mono:
```
"  at Microsoft.Framework.Internal.PropertyHelper.CallPropertySetter[Person,Decimal] (System.Action`2 setter, System.Object target, System.Object value) [0x00000] in <filename unknown>:0 \n  
at Microsoft.AspNet.Mvc.ModelBinding.MutableObjectModelBinder.SetProperty (Microsoft.AspNet.Mvc.ModelBinding.ModelBindingContext bindingContext, Microsoft.AspNet.Mvc.ModelBinding.ModelExplorer modelExplorer, Microsoft.AspNet.Mvc.ModelBinding.ModelMetadata propertyMetadata, Microsoft.AspNet.Mvc.ModelBinding.ModelBindingResult dtoResult) [0x00000] in <filename unknown>:0 "
```

sornaks:
This is the exception in CLR (Windows):
```
at Microsoft.Framework.Internal.PropertyHelper.CallPropertySetter[TDeclaringType,TValue](Action`2 setter, Object target, Object value)\r\n   
at Microsoft.AspNet.Mvc.ModelBinding.MutableObjectModelBinder.SetProperty(ModelBindingContext bindingContext, ModelExplorer modelExplorer, ModelMetadata propertyMetadata, ModelBindingResult dtoResult)
```

dougbu:
Thanks very much for filing this bug @sornaks! This bug uncovers at least 3 intertwined issues, most of which are regressions from MVC 5:
[...]

PR #2777, opened 2015-07-02
Modify `TypeConverterModelBinder`'s `ModelBindingResult.IsModelSet` to be false when model value is `null`.

NTaylorMullen:
- Previously `ModelBindingResult.IsModelSet` would be set to true if type conversions resulted in empty => `null` values. This resulted in improper usage of `ModelBindingResult`s creating null ref exceptions in certain cases.
- Updated existing functional test to account for new behavior. Previously it was handling the null ref exception by stating that errors were simply invalid; now we can provide a more distinct error.
- Added unit test to validate `TypeConverterModelBinder` does what it's supposed to when conversions result in null values.

https://github.com/aspnet/Mvc/issues/2720

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/TypeConverterModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/TypeConverterModelBinder.cs
@@ -2,7 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Reflection;
 using System.Threading.Tasks;
+using Microsoft.AspNet.Mvc.Core;
 
 namespace Microsoft.AspNet.Mvc.ModelBinding
 {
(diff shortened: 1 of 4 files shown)


NTaylorMullen:
@rynowak added integration tests commit.

NTaylorMullen:
@rynowak Added errors commits.

NTaylorMullen:
@rynowak Added nullability checking to `TypeConverter`. This means `string`s can now be set to `null` or empty string without error.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/TypeConverterModelBinder.cs
```diff
@@ -34,12 +36,18 @@ public sealed class TypeConverterModelBinder : IModelBinder
...
                    bindingContext.ModelMetadata,
                    newModel);
+                var isModelSet = true;
 
-                return new ModelBindingResult(
-                    newModel,
-                    bindingContext.ModelName,
-                    isModelSet: true,
-                    validationNode: validationNode);
+                if (newModel == null && !AllowsNullValue(bindingContext.ModelType))
```
rynowak: This deserves a comment about what case it handles (type converter returns null instead of throwing).

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/TypeConverterModelBinderTest.cs
```diff
@@ -67,6 +67,35 @@ public class TypeConverterModelBinderTest
...
+        [InlineData(typeof(Guid))]
+        [InlineData(typeof(double))]
+        [InlineData(typeof(DayOfWeek))]
+        public async Task BindModel_CreatesError_WhenTypeConversionIsNull(Type destinationType)
+        {
+            // Arrange
+            var bindingContext = GetBindingContext(destinationType);
+            bindingContext.ValueProvider = new SimpleHttpValueProvider
+            {
+                { "theModelName", "" }
```
rynowak: `string.Empty`

rynowak:
:shipit:

PR #2783, opened 2015-07-06
Rename builder extension types to have more specific names

Eilon:


src/Microsoft.AspNet.Mvc.Core/MvcBuilderExtensions.cs
```diff
@@ -13,7 +13,7 @@ namespace Microsoft.AspNet.Builder
     /// <summary>
     /// Extension methods for <see cref="IApplicationBuilder"/> to add MVC to the request execution pipeline.
     /// </summary>
-    public static class BuilderExtensions
+    public static class MvcBuilderExtensions
```
rynowak: We should call this `MvcCoreApplicationBuilderExtensions` or something even longer and more horrid. We're about to add an `IMvcBuilder` so this name is a little ambiguous. Also, there's 'MVC' extension methods for `IApplicationBuilder` coming from a few packages (hence the 'Core').

src/Microsoft.AspNet.Mvc.WebApiCompatShim/Routing/WebApiCompatShimRouteBuilderExtensions.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace Microsoft.AspNet.Builder
 {
-    public static class RouteBuilderExtensions
+    public static class WebApiCompatShimRouteBuilderExtensions
```
rynowak: :+1:

src/Microsoft.AspNet.Mvc.Core/MvcBuilderExtensions.cs
Eilon: I didn't find any other builder extensions in this repo outside of these two types... But I'm happy to change this to `MvcApplicationBuilderExtensions` so that it's clearer about the target an less likely to conflict.

rynowak: Yeah, I double checked and you're right. We'll only need one version of `AddMvc`

rynowak:
:shipit:

Eilon:
Thanks Broyan.

PR #2785, opened 2015-07-07
Rename IFilter -> IFilterMetadata

rynowak:
Simple rename and cleanup of a few tests that had a hardcoded string.


Eilon:
:shipit:

PR #2789, opened 2015-07-07
Refactoring to get ready for JsonPatch (for dynamics) support.  Relay…

KevinDockx:
This is the first in a series of PR's to allow supporting dynamics in JsonPatch.  It's a split-up of the commits I initially added to a PR (https://github.com/aspnet/Mvc/pull/2736#issuecomment-118213531), as discussed with Eilon, Daniel and Kirthi on Skype.  

This may look like it contains a lot of changes, but it's always just a few lines per file :)

The reason for this refactoring is that we need to allow users to create a JsonPatchDocument without having to know the type the patchdoc will be executed on, and later on: executing that patchdoc on a dynamically typed object - a very typical case that will be possible when the full implementation of dynamics support is done.


src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -15,16 +15,16 @@
 namespace Microsoft.AspNet.JsonPatch.Adapters
 {
     /// <inheritdoc />
-    public class ObjectAdapter<TModel> : IObjectAdapter<TModel> where TModel : class
+    internal class ObjectAdapter : IObjectAdapter
```
rynowak: We don't use the `internal` keyword in our repos unless there's a very good reason to keep someone from using something. The general philosophy is to make it possible to use if a user wants to mock it for tests or new it up.

If you want to put it in a more obscure namespace like `Microsoft.AspNet.JsonPatch.Internal` that would be ok too, but I think it's fine as-is.

src/Microsoft.AspNet.JsonPatch/Exceptions/JsonPatchException.cs
```diff
@@ -3,14 +3,13 @@
 
 using System;
 using Microsoft.AspNet.JsonPatch.Operations;
-using Microsoft.Framework.Internal;
 
 namespace Microsoft.AspNet.JsonPatch.Exceptions
 {
-    public class JsonPatchException<TModel> : JsonPatchException where TModel : class
+    public class JsonPatchException : JsonPatchExceptionBase //where TModel : class
```
rynowak: Delete instead of commenting out.

Do we really need the base-class at all here?

src/Microsoft.AspNet.JsonPatch/Exceptions/JsonPatchExceptionBase.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace Microsoft.AspNet.JsonPatch.Exceptions
 {
-    public class JsonPatchException : Exception
+    public abstract class JsonPatchExceptionBase : Exception
```
rynowak: This really just do what `System.Exception` does doesn't it?

src/Microsoft.AspNet.JsonPatch/JsonPatchDocumentOfT.cs
```diff
@@ -32,10 +31,8 @@ public JsonPatchDocument()
             ContractResolver = new DefaultContractResolver();
         }
 
-        // Create from list of operations.
-        public JsonPatchDocument(
-            [NotNull] List<Operation<TModel>> operations,
-            [NotNull] IContractResolver contractResolver)
+        // Create from list of operations
+        public JsonPatchDocument(List<Operation<TModel>> operations, IContractResolver contractResolver)
```
rynowak: Where `[NotNull]` has been removed... is that because `null` is allowed now?

src/Microsoft.AspNet.JsonPatch/JsonPatchDocumentOfT.cs
```diff
@@ -61,15 +58,16 @@ public JsonPatchDocument<TModel> Add<TProp>([NotNull] Expression<Func<TModel, TP
...
+        /// <typeparam name="TProp">value type</typeparam>
+        /// <param name="path">path</param>
+        /// <param name="value">value</param>
         /// <param name="position">position</param>
-        /// <returns>The <see cref="JsonPatchDocument{TModel}"/>.</returns>
+        /// <returns></returns>
         public JsonPatchDocument<TModel> Add<TProp>(
-            [NotNull] Expression<Func<TModel, IList<TProp>>> path,
+            Expression<Func<TModel,
+            IList<TProp>>> path,
```
rynowak: This linebreak makes it hard to grok

rynowak:
:watch: simple change, just a few minor comments.

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
KevinDockx: Ok, I'll return it to public.

src/Microsoft.AspNet.JsonPatch/Exceptions/JsonPatchException.cs
KevinDockx: Used it to be able to set the InnerException - but it's better without the custom base class by calling : base(message, innerException).  So you're right, we don't need that base class.  Will change.

src/Microsoft.AspNet.JsonPatch/JsonPatchDocumentOfT.cs
KevinDockx: No - that's a mistake.  Re-added NotNull where applicable.

KevinDockx:
Committed changes cfr PR comments.

rynowak:
:shipit: - Looks good!

I'll merge this for you.

rynowak:
Merged! 580914d87321f73651ac20fc657ab1aca3e90062

PR #2792, opened 2015-07-08
Switch to using shared security helper

HaoK:
Remove dependency on AuthN and switch to use shared sources for security helper instead

Depends on: https://github.com/aspnet/Common/pull/27

cc @rynowak


rynowak:
:shipit:

HaoK:
46bc7dd219a73b5a08766e025ffb9db51fad23d3

Issue #2787, opened 2015-07-07
FileStreamResult should be renamed to StreamResult


davidfowl:
There's nothing File* about it /cc @rynowak

rynowak:
We need to (briefly) consider legacy. I believe @Eilon was saying the other day that names of action result classes rarely show up in user code.

Eilon:
Yeah the type names we don't super care about. It's the helper factory methods on Controller that matter a lot.

Eilon:
That is to say: we can rename the types all we want. But breaking the helper methods on Controller is bad.

ajaybhargavb:
@davidfowl, spoke to @Eilon and decided not to do this. We can return the stream directly from an action and `StreamFormatter` will take care of it.

PR #2794, opened 2015-07-08
Renamed FileStreamResult to StreamResult

ajaybhargavb:
Issue - #2787 
@Eilon

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/FileStreamResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/FileStreamResult.cs
@@ -13,48 +13,48 @@ namespace Microsoft.AspNet.Mvc
 {
     /// <summary>
     /// Represents an <see cref="ActionResult"/> that when executed will
-    /// write a file from a stream to the response.
+    /// write a stream to the response.
     /// </summary>
-    public class FileStreamResult : FileResult
+    public class StreamResult : FileResult
     {
         // default buffer size as defined in BufferedStream type
         private const int BufferSize = 0x1000;
 
-        private Stream _fileStream;
+        private Stream _stream;
 
         /// <summary>
-        /// Creates a new <see cref="FileStreamResult"/> instance with
+        /// Creates a new <see cref="StreamResult"/> instance with
         /// the provided <paramref name="fileStream"/> and the
         /// provided <paramref name="contentType"/>.
         /// </summary>
-        /// <param name="fileStream">The stream with the file.</param>
+        /// <param name="stream">The stream with the content.</param>
         /// <param name="contentType">The Content-Type header of the response.</param>
-        public FileStreamResult([NotNull] Stream fileStream, [NotNull] string contentType)
-            : this(fileStream, new MediaTypeHeaderValue(contentType))
+        public StreamResult([NotNull] Stream stream, [NotNull] string contentType)
+            : this(stream, new MediaTypeHeaderValue(contentType))
         {
         }
 
         /// <summary>
-        /// Creates a new <see cref="FileStreamResult"/> instance with
+        /// Creates a new <see cref="StreamResult"/> instance with
         /// the provided <paramref name="fileStream"/> and the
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNet.Mvc.Core/FileStreamResult.cs
```diff
@@ -13,48 +13,48 @@ namespace Microsoft.AspNet.Mvc
 {
     /// <summary>
     /// Represents an <see cref="ActionResult"/> that when executed will
-    /// write a file from a stream to the response.
+    /// write a stream to the response.
     /// </summary>
-    public class FileStreamResult : FileResult
+    public class StreamResult : FileResult
```
kichalla: I think this name is confusing...for users of Web API they might translate `StreamContent` -> `StreamResult` just like how `ObjectContent` became `ObjectResult`

davidfowl:
Why is the build failing?

src/Microsoft.AspNet.Mvc.Core/FileStreamResult.cs
davidfowl: Why does it derive from FileResult?

kichalla: Instead should we have `StreamResult` and also `FileStreamResult` which would derive from `StreamResult`?
@rynowak

ajaybhargavb:
I don't think the build failures are related to this change.

Eilon:
@davidfowl The build failures are do to aspnetvnext being behind volatile so the new packages from Hao aren't there yet.

Eilon:
Today we have:


Base class:
```c#
abstract FileResult
	protected abstract Task WriteFileAsync(HttpResponse response, CancellationToken cancellation);
```

Three concrete classes:
```c#
FileContentResult : FileResult
FilePathResult : FileResult
FileStreamResult : FileResult
```

The idea of the "file" base results is that they are meant for rendering a "file" and I think optionally prompting a "Save As" dialog in the browser.

Is there a better set of names we can use?

davidfowl:
So these things automatically set the content disposition as well right? We should have a pure `StreamResult` and have a different `FileStreamResult` that actually takes a file stream since they are different things.

Eilon:
If you set the `FileDownloadName` property on the result it'll set the `Content-Disposition` header. Otherwise it just dumps the file as-is (e.g. maybe you're rendering an image from a database).

We could certainly just add a new `StreamResult` that derives from nothing special.

Would there be any reason to specifically support `FileStream` though? The `FileStreamResult` action result takes in `Stream`, which is what `FileStream` derives from. What would we do differently for `FileStream`? Would we use it to grab the filename or something?

rynowak:
Streams 'just work' even without an action result.

Eilon:
Yes, but action result == happy unit testing :smile:

rynowak:
> Yes, but action result == happy unit testing :smile:

`ObjectResult` then. 

I'm not opposed to adding `StreamResult` to make it more obvious. You could say that `Stream` + `ObjectResult` isn't as discoverable, but adding it doesn't address any scenarios that aren't *easy* already.

danroth27:
Moved this to beta7. Please commit to dev, not release.

darrelmiller:
Anyone know what happened to the plan to add StreamResult?  I just spent the last half an hour trying to figure out if FileStreamResult did anything funky as I just want to return a stream and a content type.

@rynowak Because of the crazy stuff ObjectContent does, I would never have thought of using ObjectResult.

rynowak:
FileStreamResult should do what you want. It can do ranges by default, and it can optionally do ETag/If-ModifiedSince and Content-Disposition

darrelmiller:
@rynowak Thanks.  Feels a bit weird but I'll get used to it.

davidfowl:
It's pretty weird. We *could* add `StreamResult` that did the same thing as `FileStreamResult` to improve discoverability.

Issue #2712, opened 2015-06-17
In chrome generating type=number without step attribute does not allow decimal values.


sornaks:
When using `Html.EditorFor()` if the model is of type `float` or `double` we generate a `input` tag with `type=number`. Chrome does not allow decimal values in `input` tags of type `number` unless a `step` attribute is specified - https://code.google.com/p/chromium/issues/detail?id=74120. From the link:
>According to the standard, a number with a fractional part such as 97.2 is invalid if step= attribute is not specified.

We do not generate that `step` attribute. We should add `step=any` to get it working.

However, IE allows decimal values without adding `step` attribute.

jasonwilliams200OK:
That is an old issue (2011). Not sure if it still holds the candle as on latest Chrome (v43.0.2357.130), it accepts the decimal without step, as default value of step in IE, Chrome and Firefox is null.

However, the behavior is reproducible in IE11.

```html
<!doctype html>
<html>

    <body>
        <form type="post" action="https://github.com/">
            <input type="number" value="97.2" />
            <input type="submit" />
        </form>
        <script>
            var element = document.querySelector("[type=number]")

            alert
            (
                "default value of step: " + element.getAttribute("step") +
                "\nvalue: " + element.value +
                "\nFails validation on submit in IE11"
            )
        </script>
    </body>

</html>
```

At this point I am not sure which browser is actually conforming with the standard here, but the step (with decimal like 0.3) is required in IE11.

Try entering an alphabet to trigger the real validation error on all Firefox, IE and Chrome.

ctolkien:
In EN-AU culture:

    <input type="number" value="1.5" />

Will not submit in either IE11 or Chrome 43.

http://codepen.io/anon/pen/mJLNvO

ajaybhargavb:
Spoke to @dougbu and @NTaylorMullen, we decided to use `type=text` for float and double types. This also maintains consistency with MVC 5 behavior. Closing this issue as this is no longer relevant.

Issue #2780, opened 2015-07-03
Do not use type=number for double and single


rubenprins:
Currently, the tag helpers and default editor templates use type=number for double and single, but type=text for decimal.

There are a very good reason *not* to use type=number, which are browser support (IE as usual) and internationalization (model binding and validation).

Take 
```html
<input type="number" name="num" step="0.001" min="0" max="9999">
```

* In several browsers, input type=number will always send "." as a decimal separator over the wire (as specified by the HTML5 specs), and this cannot be changed. So a Dutch input "1.234" (1234) will be sent as "1234", and "1,234" (one and almost a quarter) will be sent as "1.234". (That is, if they accept localized separators at all, which is browser-specific.)
* Other browsers (like IE) will treat type=number mostly like type=text, and send what the user typed. So a Dutch input "1.234" (1234) will be sent as "1.234", and "1,234" (one and a bit) will be sent as "1,234". IE does do some minor validation (treating "." and "," mostly like a number separator) and range checking.
* Older browsers will treat type=number like type=text, and send what the user typed. (Like IE, but without any validation.)

In other words, you cannot see from the posted data whether "1.234" and "1,234" come from a formatted 1234 or a decimal 1 and a bit. 

This means that it is impossible for model binding/validation to know which is which, especially when you also get decimal (type=text) and explicit type=text into the mix (e.g., for full control over formatting, which HTML5 does not provide, and supporting exponential notation). 

And I highly doubt model binding/validation has been tested properly in these cases under cultures different from en-US. I'm quite sure model binding/validation will cause errors, or blindly parse HTML5-compliant decimal values like "1.125" as "1125" under several European cultures, like nl-NL. The nl-NL culture is ideal for testing here, as it has the separators swapped compared to en-US.

For users knowing they will only need to handle English formatting, a property like Html5DateRenderingMode (say, Html5FloatRenderingMode) could be added to the HtmlHelper options. But for users outside the English-speaking world, the current default is a problem, and MVC should err on the safe side like it used to.

Yes, MVC5 uses type=text for double and single, so fixing this would not be a breaking change from MVC5. It would in fact be a breaking change not to revert this change.

(The HTML5 spec is way to anglocentric here; the WG had been warned of these ambiguities, but they chose to ignore the warnings, and deal with the "most common" case -- as decided by English speakers, obviously.)

PR #2795, opened 2015-07-09
Added Step attribute to number input type

ajaybhargavb:
Issue - #2712 
@dougbu


src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultEditorTemplates.cs
```diff
@@ -378,7 +378,11 @@ public static string TimeInputTemplate(IHtmlHelper htmlHelper)
 
         public static string NumberInputTemplate(IHtmlHelper htmlHelper)
         {
-            return GenerateTextBox(htmlHelper, inputType: "number");
+            var htmlAttributes =
+                CreateHtmlAttributes(htmlHelper, className: "text-box single-line", inputType: "number");
+            htmlAttributes["step"] = "any";
```
Eilon: I thought this was needed for only decimal-enabled types: `float`, `double`, and `decimal`. For integral types there's no step needed, right?

Eilon: Plus, the code here now looks more "complicated" because it seems to slightly duplicate some existing logic...

ajaybhargavb: You are right. I was investigating #2780 and it is possible that we might not even need this change based on what we decide for that scenario.

ajaybhargavb:
Closing this PR since we decided to use `type=text` for float and double types.
Created https://github.com/aspnet/Mvc/pull/2812 with correct changes.

PR #2797, opened 2015-07-09
Introducing IHtmlContent in Mvc.Razor.

sornaks:
- Changing HtmlHelper and Helper result to implement IHtmlContent.
- Introducing BufferedHtmlContent.
- Making RazorPage handle only IHtmlContent and clearing out other types.
- Making StringCollectionTextWriter use BufferedHtmlContent so that it can be returned where necessary.
- Updating places which involve Write/Copy to pass in encoders.
- The encoders are currently not being used during write. But when HtmlString is modified to carry encode metadata, the encoder can be used for writing. This is a perf optimization and hence not a part of this change.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor/HelperResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/HelperResult.cs
@@ -3,30 +3,33 @@
 
 using System;
 using System.IO;
+using Microsoft.AspNet.Html.Abstractions;
 using Microsoft.Framework.Internal;
+using Microsoft.Framework.WebEncoders;
 
 namespace Microsoft.AspNet.Mvc.Razor
 {
     /// <summary>
     /// Represents a deferred write operation in a <see cref="RazorPage"/>.
     /// </summary>
-    public class HelperResult
+    public class HelperResult : IHtmlContent
     {
-        private readonly Action<TextWriter> _action;
+        private readonly Action<TextWriter, IHtmlEncoder> _action;
 
         /// <summary>
         /// Creates a new instance of <see cref="HelperResult"/>.
         /// </summary>
-        /// <param name="action">The delegate to invoke when <see cref="WriteTo(TextWriter)"/> is called.</param>
-        public HelperResult([NotNull] Action<TextWriter> action)
+        /// <param name="action">The delegate to invoke when
+        /// <see cref="WriteTo(TextWriter, IHtmlEncoder)"/> is called.</param>
+        public HelperResult([NotNull] Action<TextWriter, IHtmlEncoder> action)
         {
             _action = action;
         }
 
         /// <summary>
-        /// Gets the delegate to invoke when <see cref="WriteTo(TextWriter)"/> is called.
+        /// Gets the delegate to invoke when <see cref="WriteTo(TextWriter, IHtmlEncoder)"/> is called.
         /// </summary>
-        public Action<TextWriter> WriteAction
+        public Action<TextWriter, IHtmlEncoder> WriteAction
         {
             get { return _action; }
         }
@@ -35,9 +38,10 @@ namespace Microsoft.AspNet.Mvc.Razor
         /// Method invoked to produce content from the <see cref="HelperResult"/>.
         /// </summary>
         /// <param name="writer">The <see cref="TextWriter"/> instance to write to.</param>
-        public virtual void WriteTo([NotNull] TextWriter writer)
+        /// <param name="encoder">The <see cref="IHtmlEncoder"/> to encode the content.</param>
+        public virtual void WriteTo([NotNull] TextWriter writer, [NotNull] IHtmlEncoder encoder)
         {
-            _action(writer);
+            _action(writer, encoder);
         }
     }
 }
--- a/src/Microsoft.AspNet.Mvc.Razor/HelperResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/HelperResult.cs
@@ -14,14 +14,14 @@ namespace Microsoft.AspNet.Mvc.Razor
     /// </summary>
     public class HelperResult : IHtmlContent
     {
-        private readonly Action<TextWriter, IHtmlEncoder> _action;
+        private readonly Action<TextWriter> _action;
 
         /// <summary>
         /// Creates a new instance of <see cref="HelperResult"/>.
         /// </summary>
         /// <param name="action">The delegate to invoke when
         /// <see cref="WriteTo(TextWriter, IHtmlEncoder)"/> is called.</param>
-        public HelperResult([NotNull] Action<TextWriter, IHtmlEncoder> action)
+        public HelperResult([NotNull] Action<TextWriter> action)
         {
             _action = action;
         }
@@ -29,7 +29,7 @@ namespace Microsoft.AspNet.Mvc.Razor
         /// <summary>
         /// Gets the delegate to invoke when <see cref="WriteTo(TextWriter, IHtmlEncoder)"/> is called.
         /// </summary>
-        public Action<TextWriter, IHtmlEncoder> WriteAction
+        public Action<TextWriter> WriteAction
         {
             get { return _action; }
         }
@@ -41,7 +41,7 @@ namespace Microsoft.AspNet.Mvc.Razor
         /// <param name="encoder">The <see cref="IHtmlEncoder"/> to encode the content.</param>
         public virtual void WriteTo([NotNull] TextWriter writer, [NotNull] IHtmlEncoder encoder)
         {
-            _action(writer, encoder);
+            _action(writer);
         }
     }
 }
--- a/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
@@ -9,6 +9,7 @@ using System.Linq;
 using System.Security.Claims;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Antiforgery;
+using Microsoft.AspNet.Html.Abstractions;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Mvc.Razor.Internal;
 using Microsoft.AspNet.Mvc.Rendering;
@@ -125,7 +126,7 @@ namespace Microsoft.AspNet.Mvc.Razor
         }
 
         /// <inheritdoc />
-        public Action<TextWriter> RenderBodyDelegate { get; set; }
+        public Action<TextWriter, IHtmlEncoder> RenderBodyDelegate { get; set; }
 
         /// <inheritdoc />
         public bool IsLayoutBeingRendered { get; set; }
@@ -268,14 +269,14 @@ namespace Microsoft.AspNet.Mvc.Razor
             var razorWriter = writer as RazorTextWriter;
             if (razorWriter != null)
             {
-                razorWriter.CopyTo(tagHelperContentWrapperTextWriter);
+                razorWriter.CopyTo(tagHelperContentWrapperTextWriter, HtmlEncoder);
             }
             else
             {
                 var stringCollectionTextWriter = writer as StringCollectionTextWriter;
                 if (stringCollectionTextWriter != null)
                 {
-                    stringCollectionTextWriter.CopyTo(tagHelperContentWrapperTextWriter);
+                    stringCollectionTextWriter.CopyTo(tagHelperContentWrapperTextWriter, HtmlEncoder);
                 }
                 else
                 {
@@ -395,7 +396,8 @@ namespace Microsoft.AspNet.Mvc.Razor
         /// <param name="value">The <see cref="object"/> to write.</param>
         /// <remarks>
         /// <paramref name="value"/>s of type <see cref="HtmlString"/> are written without encoding and the
-        /// <see cref="HelperResult.WriteTo(TextWriter)"/> is invoked for <see cref="HelperResult"/> types.
+        /// <see cref="HelperResult.WriteTo(TextWriter, IHtmlEncoder)"/> is invoked for
+        /// <see cref="HelperResult"/> types.
         /// For all other types, the encoded result of <see cref="object.ToString"/> is written to the
         /// <paramref name="writer"/>.
         /// </remarks>
@@ -416,7 +418,8 @@ namespace Microsoft.AspNet.Mvc.Razor
         /// </param>
         /// <remarks>
         /// <paramref name="value"/>s of type <see cref="HtmlString"/> are written without encoding and the
-        /// <see cref="HelperResult.WriteTo(TextWriter)"/> is invoked for <see cref="HelperResult"/> types.
+        /// <see cref="HelperResult.WriteTo(TextWriter, IHtmlEncoder)"/> is invoked for
+        /// <see cref="HelperResult"/> types.
         /// For all other types, the encoded result of <see cref="object.ToString"/> is written to the
         /// <paramref name="writer"/>.
         /// </remarks>
@@ -431,15 +434,8 @@ namespace Microsoft.AspNet.Mvc.Razor
                 return;
             }
 
-            var helperResult = value as HelperResult;
-            if (helperResult != null)
-            {
-                helperResult.WriteTo(writer);
-                return;
-            }
-
-            var htmlString = value as HtmlString;
-            if (htmlString != null)
+            var htmlContent = value as IHtmlContent;
+            if (htmlContent != null)
             {
                 if (escapeQuotes)
                 {
@@ -449,7 +445,7 @@ namespace Microsoft.AspNet.Mvc.Razor
                     //
                     // Do not combine following condition with check of escapeQuotes; htmlString.ToString() can be
                     // expensive when the HtmlString is created with a StringCollectionTextWriter.
-                    var stringValue = htmlString.ToString();
+                    var stringValue = htmlContent.ToString();
                     if (stringValue.Contains("\""))
                     {
                         writer.Write(stringValue.Replace("\"", "&quot;"));
@@ -457,7 +453,7 @@ namespace Microsoft.AspNet.Mvc.Razor
                     }
                 }
 
-                htmlString.WriteTo(writer);
+                htmlContent.WriteTo(writer, encoder);
                 return;
             }
 
--- a/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
@@ -387,9 +387,8 @@ namespace Microsoft.AspNet.Mvc.Razor
         /// <param name="writer">The <see cref="TextWriter"/> instance to write to.</param>
         /// <param name="value">The <see cref="object"/> to write.</param>
         /// <remarks>
-        /// <paramref name="value"/>s of type <see cref="HtmlString"/> are written without encoding and the
-        /// <see cref="HelperResult.WriteTo(TextWriter, IHtmlEncoder)"/> is invoked for
-        /// <see cref="HelperResult"/> types.
+        /// <paramref name="value"/>s of type <see cref="IHtmlContent"/> are written using 
+        /// <see cref="IHtmlContent.WriteTo(TextWriter, IHtmlEncoder)"/>.
         /// For all other types, the encoded result of <see cref="object.ToString"/> is written to the
         /// <paramref name="writer"/>.
         /// </remarks>
@@ -409,9 +408,8 @@ namespace Microsoft.AspNet.Mvc.Razor
         /// Otherwise writes <see cref="HtmlString"/> values as-is.
         /// </param>
         /// <remarks>
-        /// <paramref name="value"/>s of type <see cref="HtmlString"/> are written without encoding and the
-        /// <see cref="HelperResult.WriteTo(TextWriter, IHtmlEncoder)"/> is invoked for
-        /// <see cref="HelperResult"/> types.
+        /// <paramref name="value"/>s of type <see cref="IHtmlContent"/> are written using 
+        /// <see cref="IHtmlContent.WriteTo(TextWriter, IHtmlEncoder)"/>.
         /// For all other types, the encoded result of <see cref="object.ToString"/> is written to the
         /// <paramref name="writer"/>.
         /// </remarks>
@@ -435,8 +433,8 @@ namespace Microsoft.AspNet.Mvc.Razor
                     // an attribute value that may have been quoted with single quotes, must handle any double quotes
                     // in the value. Writing the value out surrounded by double quotes.
                     //
-                    // Do not combine following condition with check of escapeQuotes; htmlString.ToString() can be
-                    // expensive when the HtmlString is created with a StringCollectionTextWriter.
+                    // Do not combine following condition with check of escapeQuotes; htmlContent.ToString() can be
+                    // expensive when the IHtmlContent is created with a BufferedHtmlContent.
                     var stringValue = htmlContent.ToString();
                     if (stringValue.Contains("\""))
                     {
--- a/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
@@ -126,7 +126,7 @@ namespace Microsoft.AspNet.Mvc.Razor
         }
 
         /// <inheritdoc />
-        public Action<TextWriter, IHtmlEncoder> RenderBodyDelegate { get; set; }
+        public Action<TextWriter> RenderBodyDelegate { get; set; }
 
         /// <inheritdoc />
         public bool IsLayoutBeingRendered { get; set; }
@@ -269,7 +269,7 @@ namespace Microsoft.AspNet.Mvc.Razor
             var razorWriter = writer as RazorTextWriter;
             if (razorWriter != null)
             {
-                razorWriter.CopyTo(tagHelperContentWrapperTextWriter, HtmlEncoder);
+                razorWriter.CopyTo(tagHelperContentWrapperTextWriter);
             }
             else
             {
--- a/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
@@ -316,7 +316,7 @@ namespace Microsoft.AspNet.Mvc.Razor
             var tagHelperOutput = tagHelperExecutionContext.Output;
             var isTagNameNullOrWhitespace = string.IsNullOrWhiteSpace(tagHelperOutput.TagName);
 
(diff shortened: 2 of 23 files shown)


sornaks:
Haven't added changes which include updating `TagHelperContent` to implement `IHtmlContent`. I'll do that tomorrow. /cc: @rynowak

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/BufferedHtmlContent.cs
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        /// <summary>
+        /// Appends a character array to the collection.
+        /// </summary>
+        /// <param name="value">The character array to be appended.</param>
+        /// <param name="index">The index from which the character array must be read.</param>
+        /// <param name="count">The count till which the character array must be read.</param>
+        /// <remarks>
+        /// Splits the character array into strings of 1KB length and appends them.
```
IDisposable: Should we be specifying 1KB here when it's actually MaxCharToStringLength. Seems like an implementation detail that is being leaked.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/BufferedHtmlContent.cs
```diff
@@ -0,0 +1,131 @@
...
+                    writer.Write(entryAsString);
+                }
+                else
+                {
+                    // Only string, IHtmlContent values can be added to the buffer.
+                    var entryAsIHtmlContent = (IHtmlContent)entry;
+                    if (entryAsIHtmlContent != null)
+                    {
+                        entryAsIHtmlContent.WriteTo(writer, encoder);
+                    }
```
IDisposable: else Debug.Assert?

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/BufferedHtmlContent.cs
```diff
@@ -0,0 +1,131 @@
...
+                if (entryAsString != null)
+                {
+                    // encoder.HtmlEncode(entryAsString, writer);
+                    writer.Write(entryAsString);
+                }
+                else
+                {
+                    // Only string, IHtmlContent values can be added to the buffer.
+                    var entryAsIHtmlContent = (IHtmlContent)entry;
+                    if (entryAsIHtmlContent != null)
```
IDisposable: We could throw here with for an entry that is a ```string``` but is null. Since we're exposing the ```List<object>``` as an internal property we don't know going forward that nobody is going to do that... perhaps the whole thing should be based on ```is string``` and ```is IHtmlContent``` rather that throwing an ```InvalidCastException```

src/Microsoft.AspNet.Mvc.Razor/RazorTextWriter.cs
```diff
@@ -51,6 +52,8 @@ public override Encoding Encoding
 
         private TextWriter TargetWriter { get; set; }
 
+        private IHtmlEncoder HtmlEncoder { get; set; } = new HtmlEncoder();
```
IDisposable: Are ```HtmlEncoders``` both cheap to new and stateful? If not **both**, perhaps we should cache a static Instance for defaulting? Even if they are stateful, but still expensive to new, perhaps a lazy creation?

src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
```diff
@@ -205,7 +206,7 @@ private bool EnableInstrumentation
             if (bodyWriter.IsBuffering)
             {
                 // Only copy buffered content to the Output if we're currently buffering.
-                await bodyWriter.CopyToAsync(context.Writer);
+                await bodyWriter.CopyToAsync(context.Writer, new HtmlEncoder());
```
davidfowl: Why are we newing this up everywhere? Don't we get it from DI?

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/BufferedHtmlContent.cs
sornaks: We block 'null' 'string's and 'IHtmlContent' in 'Append'.

src/Microsoft.AspNet.Mvc.Razor/HelperResult.cs
```diff
@@ -3,30 +3,33 @@
...
 namespace Microsoft.AspNet.Mvc.Razor
 {
     /// <summary>
     /// Represents a deferred write operation in a <see cref="RazorPage"/>.
     /// </summary>
-    public class HelperResult
+    public class HelperResult : IHtmlContent
     {
-        private readonly Action<TextWriter> _action;
+        private readonly Action<TextWriter, IHtmlEncoder> _action;
```
rynowak: Would it be simpler to just leave this as `Func<TextWriter>`? We're planning to delete this class anyway. How big of a change would that be to include in this PR?

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -449,7 +445,7 @@ public virtual void WriteTo([NotNull] TextWriter writer, object value)
                     //
                     // Do not combine following condition with check of escapeQuotes; htmlString.ToString() can be
                     // expensive when the HtmlString is created with a StringCollectionTextWriter.
-                    var stringValue = htmlString.ToString();
+                    var stringValue = htmlContent.ToString();
```
rynowak: I'm not happy to see a `ToString()` - but I get why it's showing up. Let's just make a mental note to revist as part of the perf part of this change.

src/Microsoft.AspNet.Mvc.Razor/RazorTextWriter.cs
rynowak: They are pretty cheap, but this is wrong for a different reason - this is a service that we get from DI.

src/Microsoft.AspNet.Mvc.Razor/RazorTextWriter.cs
```diff
@@ -186,24 +189,24 @@ public override void Flush()
...
         }
 
         /// <inheritdoc />
-        public Task CopyToAsync(TextWriter writer)
+        public Task CopyToAsync(TextWriter writer, IHtmlEncoder encoder)
         {
             writer = UnWrapRazorTextWriter(writer);
-            return BufferedWriter.CopyToAsync(writer);
+            return BufferedWriter.CopyToAsync(writer, encoder);
         }
```
rynowak: Can you revert the changes to this class? What breaks? No-one should be calling `TextWriter.Write(...)` and passing in an `IHtmlContent`

src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
rynowak: This should be available already somehow.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/BufferedHtmlContent.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        /// <inheritdoc />
+        public void WriteTo([NotNull] TextWriter writer, [NotNull] IHtmlEncoder encoder)
+        {
+            foreach (var entry in Entries)
+            {
+                var entryAsString = entry as string;
+                if (entryAsString != null)
+                {
+                    // encoder.HtmlEncode(entryAsString, writer);
```
rynowak: bleh

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/BufferedHtmlContent.cs
rynowak: So then don't have the null-check since it can't happen.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/BufferedHtmlContent.cs
```diff
@@ -0,0 +1,131 @@
...
+using Microsoft.AspNet.Html.Abstractions;
+using Microsoft.Framework.Internal;
+using Microsoft.Framework.WebEncoders;
+
+namespace Microsoft.AspNet.Mvc.Rendering
+{
+    /// <summary>
+    /// Enumerable object collection which knows how to write itself.
+    /// </summary>
+    public class BufferedHtmlContent : IHtmlContent, IEnumerable<object>
```
rynowak: Do we need `IEnumerable<object>`? Who's looking at it?

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/StringCollectionTextWriter.cs
```diff
@@ -28,7 +29,7 @@ public class StringCollectionTextWriter : TextWriter
         public StringCollectionTextWriter(Encoding encoding)
         {
             _encoding = encoding;
-            Buffer = new BufferEntryCollection();
+            Content = new BufferedHtmlContent();
```
rynowak: Not sure about the need for changes to this class. It's a `TextWriter` and as such it can safely just operate on strings.

src/Microsoft.AspNet.PageExecutionInstrumentation.Interfaces/IBufferedTextWriter.cs
```diff
@@ -19,14 +20,16 @@ public interface IBufferedTextWriter
...
 
         /// <summary>
         /// Asynchronously copies the buffered content to the <paramref name="writer"/>.
         /// </summary>
-        /// <param name="writer">The writer to copy the contents to.</param>
+        /// <param name="writer">The <see cref="TextWriter"/> to copy the contents to.</param>
+        /// <param name="encoder">The <see cref="IHtmlEncoder"/> which encodes the content to be written.</param>
         /// <returns>A <see cref="Task"/> representing the copy operation.</returns>
-        Task CopyToAsync(TextWriter writer);
+        Task CopyToAsync(TextWriter writer, IHtmlEncoder encoder);
```
rynowak: We shouldn't change this api

rynowak:
:watch:

src/Microsoft.AspNet.PageExecutionInstrumentation.Interfaces/IBufferedTextWriter.cs
pranavkm: We'll need to let the BrowserLink folks know if any API changes are made in this project.

marcosph: I have an MVC6 intranet application that uses roles. The admin users can
manage all the user roles. However any changes made to a users' roles are
only effective when the user logs on (this makes sense as the roles are
stored as claims in a cookie). The requirement is for role changes to take
effect immediately, like they did in the older asp.net membership and role
managers. How can one achieve this with the new identity models? The only
way that comes to mind is to set some flag that is checked on each request
and if set then update the cookie to reflect the new roles. Do you have any
other suggestions or examples of how to ensure the role changes take effect
immediatley?

sornaks: @marcosph can you file an issue for this? We can discuss it there.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/StringCollectionTextWriter.cs
sornaks: This is changing because we removed the `StringCollectionTextWriter` constructor in `HtmlString` which is used here - https://github.com/aspnet/Mvc/pull/2797/files#diff-0e5ef27109c7145789f12182064f84acL423

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/BufferedHtmlContent.cs
sornaks: We dont. Let me remove it.

src/Microsoft.AspNet.Mvc.Razor/RazorTextWriter.cs
sornaks: This is the place it is used - https://github.com/aspnet/Mvc/pull/2797/files#diff-21281ad55491937a33b7eee9b50bdc81L278

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/BufferedHtmlContent.cs
sornaks: The comment was a mistake. But this code will be present with the perf change (when encoding is delayed).

src/Microsoft.AspNet.Mvc.Razor/HelperResult.cs
sornaks: It shouldn't be a big change. Let me try getting it done.

sornaks:
Updated..

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -395,7 +388,8 @@ public virtual void Write(object value)
         /// <param name="value">The <see cref="object"/> to write.</param>
         /// <remarks>
         /// <paramref name="value"/>s of type <see cref="HtmlString"/> are written without encoding and the
-        /// <see cref="HelperResult.WriteTo(TextWriter)"/> is invoked for <see cref="HelperResult"/> types.
+        /// <see cref="HelperResult.WriteTo(TextWriter, IHtmlEncoder)"/> is invoked for
+        /// <see cref="HelperResult"/> types.
         /// For all other types, the encoded result of <see cref="object.ToString"/> is written to the
         /// <paramref name="writer"/>.
         /// </remarks>
```
rynowak: These docs are out of date, these should just refer to `IHtmlContent`

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/HtmlString.cs
```diff
@@ -39,29 +37,26 @@ public static HtmlString Empty
...
+            var writerAsStringCollectionTextWriter = writer as StringCollectionTextWriter;
+            if (writerAsStringCollectionTextWriter != null)
             {
-                _writer.CopyTo(targetWriter);
+                writerAsStringCollectionTextWriter.Content.Append(_input);
             }
             else
             {
-                targetWriter.Write(_input);
+                writer.Write(_input);
```
rynowak: What does this actually accomplish? I would expect `writerAsStringCollectionTextWriter.Content.Append(_input)` and `writer.Write(_input)` to do the same thing

src/Microsoft.AspNet.PageExecutionInstrumentation.Interfaces/project.json
```diff
@@ -15,5 +15,8 @@
                 "System.IO": "4.0.10-beta-*"
             }
         }
+    },
+    "dependencies": {
+        "Microsoft.Framework.WebEncoders.Core": "1.0.0-*"
```
rynowak: Remove

test/Microsoft.AspNet.Mvc.Razor.Test/RazorTextWriterTest.cs
```diff
@@ -228,9 +231,9 @@ public void Copy_CopiesContent_IfTargetTextWriterIsARazorTextWriterAndBuffering(
 
             // Assert
             // Make sure content was written to the source.
-            Assert.Equal(2, source.BufferedWriter.Buffer.BufferEntries.Count);
-            Assert.Equal(1, target.BufferedWriter.Buffer.BufferEntries.Count);
-            Assert.Same(source.BufferedWriter.Buffer.BufferEntries, target.BufferedWriter.Buffer.BufferEntries[0]);
+            Assert.Equal(2, source.BufferedWriter.Content.Entries.Count());
+            Assert.Equal(1, target.BufferedWriter.Content.Entries.Count());
```
rynowak: Isn't this `IList`?

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/BufferedHtmlContentTest.cs
```diff
@@ -0,0 +1,127 @@
...
+namespace Microsoft.AspNet.Mvc.Rendering
+{
+    public class BufferedHtmlContentTest
+    {
+        [Fact]
+        public void AppendString_AppendsAString()
+        {
+            // Arrange
+            var content = new BufferedHtmlContent();
+            var writer = new StringWriter();
```
rynowak: unused

rynowak: Looks like there are more of these, clean it up!

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/BufferedHtmlContentTest.cs
```diff
@@ -0,0 +1,127 @@
...
+
+            // Act
+            content.Append(new TestHtmlContent("Hello"));
+
+            // Assert
+            var result = Assert.Single(content.Entries);
+            var testHtmlContent = Assert.IsType<TestHtmlContent>(result);
+            testHtmlContent.WriteTo(writer, new CommonTestEncoder());
+            Assert.Equal("Written from TestHtmlContent: Hello", writer.ToString());
+            writer.Dispose();
```
rynowak: No need to dispose a `StringWriter`

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/BufferedHtmlContentTest.cs
```diff
@@ -0,0 +1,127 @@
...
+            var writer = new StringWriter();
+            content.Append(new TestHtmlContent("Hello"));
+            content.Append("Test");
+
+            // Act
+            content.WriteTo(writer, new CommonTestEncoder());
+
+            // Assert
+            Assert.Equal(2, content.Entries.Count());
+            Assert.Equal("Written from TestHtmlContent: HelloTest", writer.ToString());
```
rynowak: How is this different from `CanAppendMultipleItems`?

rynowak:
Looking better. :watch:

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/HtmlString.cs
sornaks: Yes you are right. It does the same thing.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/BufferedHtmlContentTest.cs
sornaks: I'll make the other test not use a writer.. They are similar but with different purposes.

sornaks:
Updated..

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -449,7 +435,7 @@ public virtual void WriteTo([NotNull] TextWriter writer, object value)
                     //
                     // Do not combine following condition with check of escapeQuotes; htmlString.ToString() can be
                     // expensive when the HtmlString is created with a StringCollectionTextWriter.
```
kirthik: comment should be changed. It mentions `htmlString.ToString()` which does not exist anymore. Also change `HtmlString` to `IHtmlContent`

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/BufferedHtmlContent.cs
```diff
@@ -0,0 +1,113 @@
...
+using Microsoft.Framework.WebEncoders;
+
+namespace Microsoft.AspNet.Mvc.Rendering
+{
+    /// <summary>
+    /// Enumerable object collection which knows how to write itself.
+    /// </summary>
+    public class BufferedHtmlContent : IHtmlContent
+    {
+        private const int MaxCharToStringLength = 1024;
```
kirthik: Add a new line.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/HtmlString.cs
```diff
@@ -39,29 +37,18 @@ public static HtmlString Empty
...
         /// <summary>
         /// Writes the value in this instance of <see cref="HtmlString"/> to the target
-        /// <paramref name="targetWriter"/>.
+        /// <paramref name="writer"/>.
         /// </summary>
-        /// <param name="targetWriter">The <see cref="TextWriter"/> to write contents to.</param>
-        public void WriteTo(TextWriter targetWriter)
+        /// <param name="writer">The <see cref="TextWriter"/> to write contents to.</param>
+        /// <param name="encoder">The <see cref="IHtmlEncoder"/> with which the output must be encoded.</param>
+        public void WriteTo(TextWriter writer, IHtmlEncoder encoder)
```
kirthik: Is encoder needed here as a paramater? Looks like we are not using it in the method?

sornaks: Yes. It is coming from the interface `IHtmlContent`. We currently encode the text beforehand. Hence its not being used here. I have a change coming, which delays encoding until it is written.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/StringCollectionTextWriter.cs
```diff
@@ -41,12 +42,12 @@ public override Encoding Encoding
...
         /// </summary>
         // internal for testing purposes.
-        internal BufferEntryCollection Buffer { get; }
+        internal BufferedHtmlContent Content { get; }
 
         /// <inheritdoc />
         public override void Write(char value)
         {
-            Buffer.Add(value.ToString());
+            Content.Append(value.ToString());
```
KrzysztofCwalina: Why do you call ToString here? BufferedHtmlContent has an overload to add a char.

sornaks: It is a character array which in turn is converted to a string. Instead of char -> char[] -> string it is just char -> string here.

KrzysztofCwalina: Ah, for some reason I thought you had a char overload. Anyway, did you consider storing more types (char[], char) "natively" in the Entries array so that you can write them to the textwriter without creating string first?

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/BufferedHtmlContent.cs
```diff
@@ -0,0 +1,113 @@
...
+        // This is not List<IHtmlContent> because that would lead to boxing all strings to IHtmlContent
+        // which is not space performant.
+        // internal for testing.
+        internal List<object> Entries { get; set; } = new List<object>();
```
Eilon: Can remove the `set;` right? I didn't see anything calling it.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/HtmlString.cs
```diff
@@ -2,33 +2,31 @@
...
+        /// </summary>
+        /// <param name="input"><c>string</c>to initialize <see cref="HtmlString"/>.</param>
         public HtmlString(string input)
         {
             _input = input;
         }
 
         /// <summary>
-        /// Initializes a new instance of <see cref="HtmlString"/> that is backed by <paramref name="writer"/>.
+        /// Returns a <see cref="HtmlString"/> with empty content.
```
Eilon: `a` -> `an` :smile:

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/StringCollectionTextWriter.cs
sornaks: Yes, the reason why `char` is not present is because it is not a widely used overload in razor (`TagBuilder` uses it for adding some consts. I'm changing this in an upcoming PR). And the reason why `char[]` is not present is to avoid calling `TextWriter.Write()`s on each character and instead have it call `Write()` on the entire string once.
In general I'd prefer keeping `BufferedHtmlContent` simple; otherwise we'll have to worry about handling cases for each type we support. The only reason why it has a `List<object>` and not a `List<IHtmlContent>` is because we have to box strings everytime we store them (and strings are widely used in Razor).

KrzysztofCwalina: But TextWriter has Write overloads that write char[] (not char by char)

rynowak:
:shipit: from me. Take a note for @KrzysztofCwalina 's suggestion so we can follow up on it when we do the perf part of this change.

sornaks:
Sure will do.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/BufferedHtmlContent.cs
```diff
@@ -0,0 +1,113 @@
...
+
+namespace Microsoft.AspNet.Mvc.Rendering
+{
+    /// <summary>
+    /// Enumerable object collection which knows how to write itself.
+    /// </summary>
+    public class BufferedHtmlContent : IHtmlContent
+    {
+        private const int MaxCharToStringLength = 1024;
+        // This is not List<IHtmlContent> because that would lead to boxing all strings to IHtmlContent
```
dougbu: `string` is not a value type. Do you mean "wrap"?

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/BufferedHtmlContent.cs
sornaks: Yes. Changed it.

sornaks:
Checked in. Thanks! https://github.com/aspnet/Mvc/commit/a02082397a4e012b925a0f9ffff86ee54259812a

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/BufferedHtmlContent.cs
sornaks: Did not see this comment before check in. Yes. Wrap. I'll update this in the next PR.

Issue #2793, opened 2015-07-08
Concrete Collection types don't work with GenericModelBinder


kulmugdha:
If the parameter is Concrete Collection type, the `GetCollectionBinder` in `GenericModelBinder` doesn't get CollectionBinder for this type.

dougbu:
d45e2ee

PR #2800, opened 2015-07-09
Adding fix for 2756 and test cases. Skipping the test cases for 2793.

kulmugdha:


src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
```diff
@@ -202,6 +215,11 @@ private static bool CanBindType(ModelMetadata modelMetadata)
                 return false;
             }
 
+            if(modelMetadata.IsCollectionType)
```
rynowak: spacing

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
```diff
@@ -109,6 +115,13 @@ protected virtual bool CanUpdateProperty([NotNull] ModelMetadata propertyMetadat
 
         private async Task<bool> CanValueBindAnyModelProperties(MutableObjectBinderContext context)
         {
+            // If there are no properties on the model, there is nothing to bind. We are here means this is not a top
```
pranavkm: This statement seems incorrect - wouldn't we be here if we are a top level object with some properties? Also, could we rewrite the earlier condition as

```C#
if (context.PropertyMetadata != null && context.PropertyMetadata.Count == 0)
{
   return isTopLevelObject;
}
```

We wouldn't require this check here.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
```diff
@@ -98,6 +98,12 @@ protected virtual bool CanUpdateProperty([NotNull] ModelMetadata propertyMetadat
                 return true;
             }
 
+            // Create the object if it is top level object and there are no properties to bind
```
pranavkm: Could we change the comment to follow the numbering scheme that the previous comment starts?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinderContext.cs
```diff
@@ -9,6 +9,6 @@ public class MutableObjectBinderContext
     {
         public ModelBindingContext ModelBindingContext { get; set; }
 
-        public IEnumerable<ModelMetadata> PropertyMetadata { get; set; }
+        public IReadOnlyList<ModelMetadata> PropertyMetadata { get; set; }
```
pranavkm: Could we get away with doing `Any()` on the sequence and not changing this?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
pranavkm: formatting.

test/Microsoft.AspNet.Mvc.FunctionalTests/ModelBindingTest.cs
```diff
@@ -693,6 +693,47 @@ public class ModelBindingTest
...
+            // Arrange
+            var server = TestHelper.CreateServer(_app, SiteName, _configureServices);
+            var client = server.CreateClient();
+
+            // Act
+            var response = await
+                     client.GetAsync("http://localhost/Properties" +
+                     "/GetPerson");
+
+            //Assert
```
kirthik: space after //

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
kulmugdha: The line
`if (isTopLevelObject && context.PropertyMetadata != null && context.PropertyMetadata.Count == 0)
            {
                return true;
            }`

from `CanCreateModel` will return the object and we will not come here. We will come here only if we are internal object.

test/Microsoft.AspNet.Mvc.FunctionalTests/ModelBindingTest.cs
```diff
@@ -693,6 +693,47 @@ public class ModelBindingTest
...
+            // Arrange
+            var server = TestHelper.CreateServer(_app, SiteName, _configureServices);
+            var client = server.CreateClient();
+
+            // Act
+            var response = await
+                     client.GetAsync("http://localhost/Properties" +
+                     "/GetPeople?people[0].Name=asdf");
+
+            //Assert
```
kirthik: space after //

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
pranavkm: Would this never be hit once #2793 gets resolved?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
```diff
@@ -109,6 +115,13 @@ protected virtual bool CanUpdateProperty([NotNull] ModelMetadata propertyMetadat
 
         private async Task<bool> CanValueBindAnyModelProperties(MutableObjectBinderContext context)
         {
+            // If there are no properties on the model, there is nothing to bind. We are here means this is not a top 
+            // level object. So we return false.
+            if (context.PropertyMetadata == null || context.PropertyMetadata.Count == 0)
```
IDisposable: .Any() instead of .Count seems cheaper and allows PropertyMetadata to stay IEnumerable<ModelMetadata>.

rynowak: `GetMetadataForProperties` runs user-code and we need to be careful to evaluate the sequence at-most once. This code should have *already* been using something stronger than `IEnumerable`.

pranavkm: Should we change the method signature to return a `IReadOnlyList` in that case?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
rynowak: This gets hit when you have a collection type that can't be handled by `GenericModelBinder`. In theory there would be more repro cases that behave like #2793

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
IDisposable: :+1:  on the IReadOnlyList then

Should we make the copy happen on the property settor?

rynowak: I considered it and didn't feel strongly. Tradeoffs here are extra copy vs making it like 5% easier for implementors :laughing: 

We're taking a hard look at all of this extensibility for beta7, so any changes we make will last until then.

kulmugdha:
@rynowak -- are you done looking at this? Should I create another iteration with the minor fixes that are suggested?

rynowak:
There's a lot of feedback here without replies, can you create another iteration or at least reply and explain what you're doing to address the comments?

kulmugdha:
@rynowak -- Yes, I want to send another iteration addressing the feedback. That's why I asked you if you were done looking at it. WIll send another iteration out soon.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
kulmugdha: Done

test/Microsoft.AspNet.Mvc.FunctionalTests/ModelBindingTest.cs
kulmugdha: Done

test/Microsoft.AspNet.Mvc.FunctionalTests/ModelBindingTest.cs
kulmugdha: Done

kulmugdha:
@rynowak -- I have incorporated the feedback. Please have a look.

rynowak:
:shipit: from me

kulmugdha:
Checked in 933a13608f885b284ced937da53caf31d910fb1c

Issue #2739, opened 2015-06-24
OutputFormatter provides no way to disable text encoding


timothy-shields:
### Title
OutputFormatter provides no way to disable text encoding

### Functional impact
It seems impossible to write a class derived from `OutputFormatter` for handling a custom binary media type that uses no text encoding.

### Minimal repro steps

1. Create new Web API app (using `"Microsoft.AspNet.Mvc": "6.0.0-beta4"`)

2. Add a `MyBinaryOutputFormatter` class:

        public class MyBinaryOutputFormatter : OutputFormatter
        {
            public MyBinaryOutputFormatter()
            {
                SupportedMediaTypes.Add(new MediaTypeHeaderValue("application/x-mybinary"));
                SupportedEncodings.Add(Encoding.UTF8);
            }

            public override bool CanWriteResult(OutputFormatterContext context, MediaTypeHeaderValue contentType)
            {
                return context.Object is byte[];
            }

            public override async Task WriteResponseBodyAsync(OutputFormatterContext context)
            {
                using (var stream = new System.IO.MemoryStream())
                {
                    byte[] data = (byte[])context.Object;
                    await context.ActionContext.HttpContext.Response.Body.WriteAsync(data, 0, data.Length);
                }
            }
        }

3. Add the output formatter in `Startup.cs`:

        services.AddMvc().Configure<MvcOptions>(options =>
        {
            options.OutputFormatters.Add(new MyBinaryOutputFormatter());
        });

4. Add a controller and action like the following:

        [HttpGet]
        public byte[] Test()
        {
            return new byte[] { 0xFF };
        }

5. Start the service and do a GET request to the action method using a HTTP client that allows you to inspect the response body contents as hex.

### Expected result
Since I write to the response body stream directly, I would expect the bytes I write to be exactly those seen in the response body client-side.  That is the response body seen client side should be the single byte FF (hex).

### Actual result
The response body seen client side is the single byte FD (hex).

### Further technical details
Using `Encoding.ASCII` instead in the `MyBinaryOutputFormatter` constructor seems to be a hacky workaround to this problem, but then all responses have a Content-Type header of `application/x-mybinary; charset=us-ascii`, which doesn't look great to clients.

If the single-byte response is not convincing that the problem is the encoding, try returning a larger `byte[]` from the `Test` action method.  It quickly becomes evident that the encoding is corrupting the data.

danroth27:
@Tratcher Why would what got written to the body change in this case?

[...]

PR #2801, opened 2015-07-09
Fix for #2739

rynowak:
This change removes the validation that forces an OutputFormatter to set
an encoding, so that you can use the OutputFormatter base class for
non-text.

The check that we had would pretty much only be hit when you
didn't have any SupportedEncodings. If you have anything in
SupportedEncodings, then one of them will be picked as a default. So
removing the check is to do, because for a text-based formatter you'd
never run into this issue in the first place.

### DIFF
--- a/test/Microsoft.AspNet.Mvc.Core.Test/OutputFormatterTests.cs
+++ b/test/Microsoft.AspNet.Mvc.Core.Test/OutputFormatterTests.cs
@@ -76,24 +76,31 @@ namespace Microsoft.AspNet.Mvc.Test
         }
 
         [Fact]
-        public void WriteResponseContentHeaders_FormatterWithNoEncoding_Throws()
+        public void WriteResponseContentHeaders_NoSupportedEncodings_NoEncodingIsSet()
         {
             // Arrange
-            var testFormatter = new TestOutputFormatter();
-            var testContentType = MediaTypeHeaderValue.Parse("text/invalid");
-            var formatterContext = new OutputFormatterContext();
-            var mockHttpContext = new Mock<HttpContext>();
-            var httpRequest = new DefaultHttpContext().Request;
-            mockHttpContext.SetupGet(o => o.Request).Returns(httpRequest);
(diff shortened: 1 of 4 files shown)


kirthik:
I am not an expert here but looks good to me :shipit:

test/Microsoft.AspNet.Mvc.Core.Test/OutputFormatterTests.cs
```diff
@@ -76,24 +76,31 @@ public static IEnumerable<object[]> SelectResponseCharacterEncodingData
         }
 
         [Fact]
-        public void WriteResponseContentHeaders_FormatterWithNoEncoding_Throws()
+        public void WriteResponseContentHeaders_NoSupportedEncodings_NoEncodingIsSet()
```
dougbu: Does this one test cover the original poster's scenario and related scenarios well enough? E.g. seems to cover only the "we assume this is a non-text format" case and not the "custom implementation of SelectCharacterEncoding" case.

dougbu:
:watch: Fix looks perfect but I'd like to see related scenarios tested a bit more.

test/Microsoft.AspNet.Mvc.Core.Test/OutputFormatterTests.cs
rynowak: Both of these scenarios are reasons why `SelectEncoding` might return `null`, which is what's covered here. 

I could add another test using mocks that does the same thing this one does. I don't really seem much value add.

dougbu: :ok:

dougbu:
:shipit:

dougbu:
Well, ship it if AppVeyor and Travis say everything is fine.

rynowak:
> Well, ship it if AppVeyor and Travis say everything is fine.

This was checked in like 2 weeks ago :laughing:

PR #2803, opened 2015-07-09
[Design] Refactor `WriteAttributeTo` to enable Razor to utilize value removal for `TagHelper`s.

NTaylorMullen:
- Refactored 3 methods out of the existing `WriteAttributeTo`. The refactoring was based on removing the `Begin`/`EndContext` for instrumentation bits which isn't valid in a `TagHelper` attribute scenario.

aspnet/Razor#247

Razor PR: https://github.com/aspnet/Razor/pull/446

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
@@ -547,15 +547,6 @@ namespace Microsoft.AspNet.Mvc.Razor
                 {
                     var attrVal = values[i];
                     var val = attrVal.Value;
-                    var next = i == values.Length - 1 ?
-                        suffix : // End of the list, grab the suffix
-                        values[i + 1].Prefix; // Still in the list, grab the next prefix
-
-                    if (val.Value == null)
-                    {
-                        // Nothing to write
-                        continue;
-                    }
 
                     // The special cases here are that the value we're writing might already be a string, or that the
                     // value might be a bool. If the value is the bool 'true' we want to write the attribute name
@@ -563,23 +554,12 @@ namespace Microsoft.AspNet.Mvc.Razor
                     // Otherwise the value is another object (perhaps an HtmlString) and we'll ask it to format itself.
                     string stringValue;
 
-                    // Intentionally using is+cast here for performance reasons. This is more performant than as+bool?
-                    // because of boxing.
-                    if (val.Value is bool)
+                    if (!ShouldRenderAttributeValue(val))
                     {
-                        if ((bool)val.Value)
-                        {
-                            stringValue = name;
-                        }
-                        else
-                        {
-                            continue;
-                        }
+                        continue;
                     }
-                    else
-                    {
-                        stringValue = val.Value as string;
-                    }
+
+                    stringValue = GetStringAttributeValue(name, val);
 
                     if (first)
                     {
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -547,15 +547,6 @@ public virtual void WriteLiteralTo([NotNull] TextWriter writer, string value)
                 {
                     var attrVal = values[i];
                     var val = attrVal.Value;
```
dougbu: Remove these abbreviations while updating this method; perhaps `attributeValue` and `value`.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -857,6 +849,34 @@ public virtual HtmlString SetAntiforgeryCookieAndHeader()
...
+        public static bool ShouldRenderAttributeValue(object rawAttributeValue)
+        {
+            if (rawAttributeValue == null)
+            {
+                // Nothing to write
+                return false;
+            }
+
+            // Intentionally using is+cast here for performance reasons. This is more performant than as+bool?
+            // because of boxing.
```
dougbu: Comment applies in `GetStringAttributeValue()` as well.

More importantly, now doing the same casts and checks twice. Suggest instead extending `GetStringAttributeValue()` to return `null` when `rawAttributeValue` is `null` or a `false` Boolean value.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -563,24 +554,13 @@ public virtual void WriteLiteralTo([NotNull] TextWriter writer, string value)
                     // Otherwise the value is another object (perhaps an HtmlString) and we'll ask it to format itself.
```
dougbu: Comment relates to details of `ShouldRenderAttributeValue()` and `GetStringAttributeValue()`. Mentioned special cases are not visible here.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -563,24 +554,13 @@ public virtual void WriteLiteralTo([NotNull] TextWriter writer, string value)
...
                     string stringValue;
```
dougbu: Combine with line 559 i.e. `var stringValue = GetStringAttributeValue(name, value);`

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -857,6 +849,34 @@ public virtual HtmlString SetAntiforgeryCookieAndHeader()
             return HtmlString.Empty;
         }
 
+        public static bool ShouldRenderAttributeValue(object rawAttributeValue)
```
dougbu: `static` methods normally come before instance methods. At least move these two above `WriteUnprefixedAttributeValueTo()` to match the order the three are used.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
dougbu: (A special case for `rawAttributeValue==null` isn't necessary in `GetStringAttributeValue()`; will already return `null` in that case. And `ShouldRenderAttributeValue()` won't be needed once `GetStringAttributeValue()` handles `rawAttributeValue==false`.)

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -563,24 +554,13 @@ public virtual void WriteLiteralTo([NotNull] TextWriter writer, string value)
                     // Otherwise the value is another object (perhaps an HtmlString) and we'll ask it to format itself.
                     string stringValue;
 
-                    // Intentionally using is+cast here for performance reasons. This is more performant than as+bool?
-                    // because of boxing.
-                    if (val.Value is bool)
+                    if (!ShouldRenderAttributeValue(val))
```
dougbu: Incorrect since `value` is of type `PositionTagged<object>` and `ShouldRenderAttributeValue()` never checks its `Value`. Either change the `value` initialization to be `attributeValue.Value.Value` or restore use of `value.Value` in the rest of this method.

dougbu: If existing tests are passing, it's to the implicit conversion from `PositionTagged<object>` to `object` or because existing tests don't cover `null` and `false` cases at all. Be explicit and avoid the repeated / buried conversion regardless.

Separately extend `RazorPageTest` to exercise `WriteAttribute[To]()` more thoroughly.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -625,6 +591,32 @@ public virtual void WriteLiteralTo([NotNull] TextWriter writer, string value)
             }
         }
 
+        public void WriteUnprefixedAttributeValueTo(
+            TextWriter writer,
+            string stringValue,
+            object rawValue,
```
dougbu: Name this `rawAttributeValue` to be consistent.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -592,29 +572,15 @@ public virtual void WriteLiteralTo([NotNull] TextWriter writer, string value)
...
-                    else if (stringValue != null)
-                    {
-                        WriteTo(writer, stringValue);
-                    }
-                    else
-                    {
-                        WriteTo(writer, val.Value);
-                    }
-
+                    WriteUnprefixedAttributeValueTo(writer, stringValue, val.Value, attrVal.Literal);
```
dougbu: Strange this remains `value.Value`

dougbu:
:watch:

NTaylorMullen:
After the new WriteAttribute changes need to re-work how this is done. Closing for now, will re-open fresh.

PR #2805, opened 2015-07-10
JsonPatch dynamics support - part 2

KevinDockx:
Next in a series of PR's that'll lead to dynamics support for JsonPatch - follow-up of #2789 :-)

In this PR, non-generic JsonPatchDocument was added, including other classes that were needed for this to work: 

- Add non-generic JsonPatchDocument - allows creating a JsonPatchDocument without having a DTO/class to create it on.  Useful when you don't work with a shared DTO layer, but also when your backend (API) isn't built in .NET. 

- Add PathHelper (used by JsonPatchDocument to execute preliminary checks on path/from)

- Add Translations to Resource file

- Add "Apply" to Operation (next to the already existing OperationOfT) (as these will be used by the non-generic JsonPatchDocument)


src/Microsoft.AspNet.JsonPatch/JsonPatchDocument.cs
```diff
@@ -0,0 +1,210 @@
...
+            foreach (var op in Operations)
+            {
+                op.Apply(objectToApplyTo, adapter);
+            }
+        }
+
+        // LogError method is required on untyped JsonPatchDocument, as errors may 
+        // be thrown in case of invalid paths.
+
+        private void LogError([NotNull] JsonPatchError jsonPatchError)
```
kirthik: This code exists in `ObjectAdapter` too. We should put it at a separate class and use it this and the other class to avoid repetition?

src/Microsoft.AspNet.JsonPatch/Converters/JsonPatchDocumentConverter.cs
```diff
@@ -0,0 +1,71 @@
...
+            return true;
+        }
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, 
+            JsonSerializer serializer)
+        {
+            try
+            {
+                if (reader.TokenType == JsonToken.Null)
+                    return null;
```
rynowak: braces

src/Microsoft.AspNet.JsonPatch/Converters/JsonPatchDocumentConverter.cs
```diff
@@ -0,0 +1,71 @@
...
+using System.Collections.Generic;
+using Microsoft.AspNet.JsonPatch.Exceptions;
+using Microsoft.AspNet.JsonPatch.Operations;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using Newtonsoft.Json.Serialization;
+
+namespace Microsoft.AspNet.JsonPatch.Converters
+{
+    public class JsonPatchDocumentConverter : JsonConverter
```
rynowak: Should we have the existing class inherit from this?

src/Microsoft.AspNet.JsonPatch/JsonPatchDocument.cs
```diff
@@ -0,0 +1,210 @@
...
+        {
+            if (LogErrorAction != null)
+            {
+                LogErrorAction(jsonPatchError);
+            }
+ 
+            // should throw error, even when logging, according to spec.
+            throw new JsonPatchException(jsonPatchError);
```
kirthik: @rynowak : Is it the right thing to do - to always throw? I remember we talked earlier that framework should not throw an exception if possible. That was the whole point of introducing `ModelState`

src/Microsoft.AspNet.JsonPatch/Converters/JsonPatchDocumentConverter.cs
```diff
@@ -0,0 +1,71 @@
...
+            {
+                if (reader.TokenType == JsonToken.Null)
+                    return null;
+
+                // load jObject
+                var jObject = JArray.Load(reader);
+
+                // Create target object for Json => list of operations
+                var concreteList = typeof(List<Operation>);
+                var targetOperations = Activator.CreateInstance(concreteList);
```
rynowak: No need for this, you can just `new` the list.

src/Microsoft.AspNet.JsonPatch/Converters/JsonPatchDocumentConverter.cs
```diff
@@ -0,0 +1,71 @@
...
+                // container target: the JsonPatchDocument. 
+                var container = Activator.CreateInstance(objectType, targetOperations, new DefaultContractResolver());
+
+                return container;
+            }
+            catch (Exception ex)
+            {
+                throw new JsonPatchException(Resources.FormatInvalidJsonPatchDocument(objectType.Name), ex);
+            }
+
```
rynowak: extra line

src/Microsoft.AspNet.JsonPatch/Operations/OperationOfT.cs
```diff
@@ -52,11 +52,5 @@ public void Apply([NotNull] TModel objectToApplyTo, [NotNull] IObjectAdapter ada
             }
```
kirthik: This class is not needed anymore right? Same case with `JsonPatchDocumentOfT`. Are you planning on removing it in one of the next PRs?

src/Microsoft.AspNet.JsonPatch/Converters/JsonPatchDocumentConverter.cs
```diff
@@ -0,0 +1,71 @@
...
+        {
+            return true;
+        }
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, 
+            JsonSerializer serializer)
+        {
+            try
+            {
+                if (reader.TokenType == JsonToken.Null)
```
rynowak: Should we validate here that `objectType` is always `typeof(JsonPatchDocument)`. Then you can just call the constructor later

src/Microsoft.AspNet.JsonPatch/Converters/JsonPatchDocumentConverter.cs
```diff
@@ -0,0 +1,71 @@
...
+                jObjectReader.Culture = reader.Culture;
+                jObjectReader.DateParseHandling = reader.DateParseHandling;
+                jObjectReader.DateTimeZoneHandling = reader.DateTimeZoneHandling;
+                jObjectReader.FloatParseHandling = reader.FloatParseHandling;
+
+                // Populate the object properties
+                serializer.Populate(jObjectReader, targetOperations);
+
+                // container target: the JsonPatchDocument. 
+                var container = Activator.CreateInstance(objectType, targetOperations, new DefaultContractResolver());
```
rynowak: `new DefaultContractResolver()` isn't good, it means that we're starting with an empty cache.

src/Microsoft.AspNet.JsonPatch/Helpers/PathHelpers.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            string adjustedPath = pathToCheck;
+
+            // check for most common path errors on create.  This is not
+            // absolutely necessary, but it allows us to already catch mistakes
+            // on creation of the patch document rather than on execute.
+
+            if (pathToCheck.Contains(".") || pathToCheck.Contains("//")
+                || pathToCheck.Contains(" ") || pathToCheck.Contains("\\")
+              )
```
kirthik: formatting is not right here

src/Microsoft.AspNet.JsonPatch/Helpers/PathHelpers.cs
```diff
@@ -0,0 +1,31 @@
...
+    internal static class PathHelpers
+    {
+        internal static CheckPathResult CheckPath(string pathToCheck)
+        {
+            string adjustedPath = pathToCheck;
+
+            // check for most common path errors on create.  This is not
+            // absolutely necessary, but it allows us to already catch mistakes
+            // on creation of the patch document rather than on execute.
+
```
kirthik: extra line

src/Microsoft.AspNet.JsonPatch/Helpers/PathHelpers.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.JsonPatch.Helpers
+{
+    internal static class PathHelpers
+    {
+        internal static CheckPathResult CheckPath(string pathToCheck)
```
rynowak: Would we ever do anything other than throw if the path is invalid? Wondering if this would be simpler as something like `string PathHelpers.NormalizePath(string)`

src/Microsoft.AspNet.JsonPatch/JsonPatchDocument.cs
```diff
@@ -0,0 +1,210 @@
...
+using Microsoft.Framework.Internal;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
+
+namespace Microsoft.AspNet.JsonPatch
+{
+    [JsonConverter(typeof(JsonPatchDocumentConverter))]
+    public class JsonPatchDocument : IJsonPatchDocument
+    {
+
```
rynowak: extra line

rynowak: A few more places too

src/Microsoft.AspNet.JsonPatch/JsonPatchDocument.cs
```diff
@@ -0,0 +1,210 @@
...
+using Microsoft.AspNet.JsonPatch.Converters;
+using Microsoft.AspNet.JsonPatch.Exceptions;
+using Microsoft.AspNet.JsonPatch.Helpers;
+using Microsoft.AspNet.JsonPatch.Operations;
+using Microsoft.Framework.Internal;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
+
+namespace Microsoft.AspNet.JsonPatch
+{
```
kirthik: Missing documentation. Can you take it relevant parts from `JsonPatchDocumentOfT` class and add the rest?

src/Microsoft.AspNet.JsonPatch/JsonPatchDocument.cs
```diff
@@ -0,0 +1,210 @@
...
+            ContractResolver = contractResolver;
+        }
+
+
+        public JsonPatchDocument Add([NotNull] string path, object value)
+        {
+            var checkPathResult = PathHelpers.CheckPath(path);
+            if (!checkPathResult.IsCorrectlyFormedPath)
+            {
+                LogError(new JsonPatchError(
```
rynowak: The error logging feature is capturing errors during deserialization. It seems to make a lot of more sense to throw if the user passes an invalid path to `Add` right? 

Am I missing something, is this used for something else?

src/Microsoft.AspNet.JsonPatch/JsonPatchDocument.cs
```diff
@@ -0,0 +1,210 @@
...
+
+        public List<Operation> Operations { get; private set; }
+
+        [JsonIgnore]
+        public IContractResolver ContractResolver { get; set; }
+
+        /// <summary>
+        /// Action for logging <see cref="JsonPatchError"/>.
+        /// </summary>
+        public Action<JsonPatchError> LogErrorAction { get; }
```
rynowak: should this also have `[JsonIgnore]`?

src/Microsoft.AspNet.JsonPatch/JsonPatchDocument.cs
```diff
@@ -0,0 +1,210 @@
...
+                LogErrorAction(jsonPatchError);
+            }
+ 
+            // should throw error, even when logging, according to spec.
+            throw new JsonPatchException(jsonPatchError);
+        }
+
+
+
+        public List<Operation> GetOperations()
```
rynowak: Should this be private

src/Microsoft.AspNet.JsonPatch/JsonPatchDocument.cs
rynowak: The framework shouldn't throw on **deserialization**. If user code does something invalid - like use an invalid path - we need to throw. 

I don't think that logging the error here is really appropriate though, that's for deserialization.

src/Microsoft.AspNet.JsonPatch/Converters/JsonPatchDocumentConverter.cs
KevinDockx: Ok, will change

src/Microsoft.AspNet.JsonPatch/Converters/JsonPatchDocumentConverter.cs
KevinDockx: Good idea :)

src/Microsoft.AspNet.JsonPatch/Converters/JsonPatchDocumentConverter.cs
KevinDockx: Ok

src/Microsoft.AspNet.JsonPatch/Converters/JsonPatchDocumentConverter.cs
KevinDockx: ok

src/Microsoft.AspNet.JsonPatch/Helpers/PathHelpers.cs
KevinDockx: Ok, I'll check

src/Microsoft.AspNet.JsonPatch/JsonPatchDocument.cs
KevinDockx: Will do!

src/Microsoft.AspNet.JsonPatch/Operations/OperationOfT.cs
KevinDockx: I think both should exists... if the user is developing using the .NET stack with a shared DTO layer between a client (say an MVC app) and the server (RESTful API/Web API), then the preferred way of working with JsonPatchDocument (or with a list of operations to create the doc from) is the generic version. This allows better checking on typo's etc - those can be avoid through creating operations using expressions.  

But there's another use case: one where there's no shared DTO layer, or one where the client might be an MVC app & API is built using another tech: sometimes, users won't create a DTO layer for those cases either => then, the user would typically use the non-generic JsonPatchDocument.  

Both result in the exact same JSON when serialized - but the added convenience (& reliability) of a generic version (as we have already) should be a good reason to keep both.

src/Microsoft.AspNet.JsonPatch/JsonPatchDocument.cs
KevinDockx: Not used for anything else - didn't realize it should only be used during deserialization; I'll remove the logError function from JsonPatchDocument

src/Microsoft.AspNet.JsonPatch/JsonPatchDocument.cs
KevinDockx: Code will be removed :)

src/Microsoft.AspNet.JsonPatch/JsonPatchDocument.cs
KevinDockx: Indeed - my mistake, will remove.

src/Microsoft.AspNet.JsonPatch/JsonPatchDocument.cs
KevinDockx: It's used by the JsonConverter implementations to get the list of operations - but maybe an explicit interface implementation for that method could be used to hide the method?

src/Microsoft.AspNet.JsonPatch/Converters/JsonPatchDocumentConverter.cs
KevinDockx: That was code from the generic converter - what should it be if not DefaultContractResolver?

src/Microsoft.AspNet.JsonPatch/Converters/JsonPatchDocumentConverter.cs
KevinDockx: Yes, makes sense - will do

src/Microsoft.AspNet.JsonPatch/JsonPatchDocument.cs
KevinDockx: LogErrorAction will be removed

src/Microsoft.AspNet.JsonPatch/Helpers/PathHelpers.cs
KevinDockx: Was used to allow logging in the calling class (return CheckPathResult to JsonPatchDoc, where LogErrorAction is) - but as logging shouldn't be done there, this can indeed be changes to PathHelpers.NormalizePath

src/Microsoft.AspNet.JsonPatch/Converters/JsonPatchDocumentConverter.cs
rynowak: OK, I see why it's this way, it's just initalizing the default value.

src/Microsoft.AspNet.JsonPatch/Converters/JsonPatchDocumentConverter.cs
```diff
@@ -0,0 +1,76 @@
...
+            {
+                if (reader.TokenType == JsonToken.Null)
+                {
+                    return null;
+                }
+
+                if (objectType != typeof(JsonPatchDocument))
+                {
+                    throw new ArgumentException(Resources.ObjectTypeMustBeJsonPatchDocument, "objectType");
+                }
```
rynowak: This should be outside the `try`. As it is you'll catch and rethrow which isn't really helping.

src/Microsoft.AspNet.JsonPatch/JsonPatchDocumentOfT.cs
```diff
@@ -466,7 +481,7 @@ public void ApplyTo(TModel objectToApplyTo, IObjectAdapter adapter)
             }
         }
 
-        public List<Operation> GetOperations()
+        List<Operation> IJsonPatchDocument.GetOperations()
```
rynowak: This should probably be IList<...> if it's going in a public contract.

src/Microsoft.AspNet.JsonPatch/Resources.resx
```diff
@@ -129,9 +132,21 @@
...
     <value>For operation '{0}', the provided path is invalid for array property at path '{1}'.</value>
   </data>
+  <data name="InvalidValueForPath" xml:space="preserve">
+    <value>The provided string '{0}' is an invalid path.</value>
+  </data>
   <data name="InvalidValueForProperty" xml:space="preserve">
     <value>The value '{0}' is invalid for property at path '{1}'.</value>
   </data>
+  <data name="ObjectTypeMustBeJsonPatchDocument" xml:space="preserve">
+    <value>objectType must be of type JsonPatchDocument.</value>
```
rynowak: Any code-names like `objectType` or `JsonPatchDocument` should be passed in as parameter as well.

rynowak:
:watch: looking close to complete :+1:

src/Microsoft.AspNet.JsonPatch/JsonPatchDocumentOfT.cs
KevinDockx: Indeed - will also fix in non-generic jsonpatchdoc

src/Microsoft.AspNet.JsonPatch/Converters/JsonPatchDocumentConverter.cs
KevinDockx: Wow, can't believe I missed that :)  will fix.

src/Microsoft.AspNet.JsonPatch/Resources.resx
KevinDockx: Ok - will also rename to something more generic

rynowak:
:shipit: looks good! Merging.

rynowak:
bbc059a6897793880ed7fa4922cf9c3016b02f83

Issue #1966, opened 2015-02-05
Url.RouteUrl can be simpler for common patterns.


shanselman:
This is a common pattern:

    dinner = await _repository.CreateDinnerAsync(dinner);
    var url = Url.RouteUrl("GetDinnerById", new { id = dinner.DinnerId }, Request.Scheme, Request.Host.ToUriComponent());

    Context.Response.StatusCode = (int)HttpStatusCode.Created;
    Context.Response.Headers["Location"] = url;
    return new JsonResult(dinner);

I find 

    var url = Url.RouteUrl("GetDinnerById", new { id = dinner.DinnerId }, Request.Scheme, Request.Host.ToUriComponent())

mixing two things and a little tedious. The last part 

    Request.Scheme, Request.Host.ToUriComponent())

is kind of administrivia. Can we make it more like?

    var Url.RouteUrlFull("GetDinnerById", new { id = dinner.DinnerId });

or

    var Url.RouteUrlAbsolute("GetDinnerById", new { id = dinner.DinnerId });

davidfowl:
What about if that method returned a builder kinda like UriBuilder that you could then use to get the full url?

```
var urlBuilder = Url.RouteUrlAsBuilder("GetDinnerById", new { id = dinner.DinnerId });
var full = urlBuilder.ToFull();
```

shanselman:
So like this?

     var myString = Url.RouteUrlAsBuilder("GetDinnerById", new { id = dinner.DinnerId }).ToFull();

Or 

     var myString = Url.RouteUrlAsBuilder("GetDinnerById", new { id = dinner.DinnerId }).ToAbsolute();

Still seems like a lot of moving parts to do something I'm gonna do ALL DAY LONG.

rustd:
var myString = Url.RouteUrlAsBuilder("GetDinnerById", new { id = dinner.DinnerId }).ToFull(); is much better

shanselman:
@rustd But why do I need to RouteUrlAsBuilder? What's the Builder? What's wrong with RouteUrlAsAbsolute and don't even bother with the ToFull. The ToString is implied.

csharpfritz:
This does feel a bit tedious...  I'd almost want the scheme and host to default to the values the current request came in on.  Most of the time, I think these values would stay the same from one request to the next

yishaigalatzer:
Add an extension method in your code to remove repetitive tediousness?

yishaigalatzer:
[...]

PR #2810, opened 2015-07-11
Absolute URL Methods Added to UrlHelper, Canonical URL Support and RequireHttpsAttribute Fix

RehanSaeed:
# UrlHelper

Following on from https://github.com/aspnet/Mvc/issues/1966 where a Link method was added to ```UrlHelper``` which requires a route name and route values. Having similar methods for 'Action' (controller and action name) and 'Content' (images and static content) would be really useful too. Here are some places where I find having 'Absolute Content' useful:

- Atom/RSS feeds require absolute URL's to image/audio/video (There is a way to specify a relative base URL but it's buggy in FireFox). See my ASP.NET MVC 6 Boilerplate [example](https://github.com/RehanSaeed/ASP.NET-MVC-Boilerplate/blob/master/Source/MVC6/Boilerplate.Web.Mvc6.Sample/Services/Feed/FeedService.cs).
- Creating a dynamic OpenSearch.xml file which contains absolute URL links to images. See my ASP.NET MVC 6 Boilerplate [example](https://github.com/RehanSaeed/ASP.NET-MVC-Boilerplate/blob/master/Source/MVC6/Boilerplate.Web.Mvc6.Sample/Services/OpenSearch/OpenSearchService.cs).
- Open Graph image/audio/video URL's are also required to be absolute.

I have these methods in my [ASP.NET MVC 6 Boilerplate Framework](https://www.nuget.org/packages/Boilerplate.Web.Mvc6/) NuGet package (Note that I add the word 'Absolute' at the beginning of the function. The current method is named 'Link' but it uses route names, so it could be named better at the very least).

# Canonical URL's for Better Search Engine Optimization (SEO)

Support for canonical URL's with the RedirectToCanonicalUrlAttribute and NoTrailingSlashAttribute filters. Performing 301 permanent redirects to the canonical URL of a resource is the correct approach according to both [Google](http://googlewebmastercentral.blogspot.co.uk/2010/04/to-slash-or-not-to-slash.html) and [Bing](http://blogs.bing.com/webmaster/2012/01/26/moving-content-think-301-not-relcanonical). The RedirectToCanonicalUrlAttribute filter helps achieve this in conjunction with the RouteOptions (AppendTrailingSlash and LowercaseUrls). See my [blog](http://rehansaeed.com/canonical-urls-for-asp-net-mvc/) for full details.


davidfowl:
@RehanSaeed The team will evaluate these changes but here's some initial feedback:
- Split these unrelated changes into different PRs. 
- Check out the coding guidelines, the style doesn't match what we expect.
- Add tests for each of the features implemented.

If you're just looking for initial feedback because this idea hasn't been cleared with the team then that should be reflected in either the title or description.

RehanSaeed:
Will split this pull request. Thank you!

PR #2812, opened 2015-07-13
Using type=text for float and double types

ajaybhargavb:
Issue - #2780 
Partly related issue - #2712 

@dougbu


dougbu:
Not quiet complete.

The commit to partially revert for this fix is 73e8fc1. Note that commit also updated the `<input>` tag helper and associated tests. (Easy to miss because the relevant dictionary uses `nameof(type)` constants instead of `typeof(alias).Name` aliases.)

dougbu:
:watch:

ajaybhargavb:
Updated.

src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
```diff
@@ -43,8 +43,6 @@ public class InputTagHelper : TagHelper
                 { nameof(UInt32), "number" },
                 { nameof(Int64), "number" },
                 { nameof(UInt64), "number" },
-                { nameof(Single), "number" },
-                { nameof(Double), "number" },
                 { nameof(Boolean), InputType.CheckBox.ToString().ToLowerInvariant() },
                 { nameof(Decimal), InputType.Text.ToString().ToLowerInvariant() },
```
dougbu: Please also remove this line because the template is not `decimal`-specific and is now inconsistent with how `float` and `double` reach `type="text"`.

Alternatively restore the `float` and `double` lines with value `InputType.Text.ToString().ToLowerInvariant()`. This alternative will behave no differently (well, it'll save one lookup iteration) but will make the table appear even-more different from that in `TemplateRenderer`. Of course that table _is_ different because we have a `decimal`-specific template there.

dougbu:
:shipit:

PR #2815, opened 2015-07-14
2721: Fixing the error message for empty html field name

kulmugdha:


kulmugdha:
@dougbu can you please have a look?

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -413,7 +413,7 @@ public virtual HtmlString GenerateAntiforgery([NotNull] ViewContext viewContext)
             var fullName = GetFullHtmlFieldName(viewContext, expression);
             if (string.IsNullOrEmpty(fullName))
             {
-                throw new ArgumentException(Resources.ArgumentCannotBeNullOrEmpty, nameof(expression));
+                throw new ArgumentException(Resources.HtmlGenerator_FieldNameCannotBeNullOrEmpty);
```
dougbu: Do not remove the `nameof(expression)` argument from any of these `ArgumentException` constructor calls. `expression` remains the parameter with the problem.

src/Microsoft.AspNet.Mvc.ViewFeatures/Resources.resx
```diff
@@ -259,4 +259,7 @@
   <data name="RemoteAttribute_RemoteValidationFailed" xml:space="preserve">
     <value>'{0}' is invalid.</value>
   </data>
+  <data name="HtmlGenerator_FieldNameCannotBeNullOrEmpty" xml:space="preserve">
+    <value>The field name for Html field cannot be Null or Empty. Please provide a non-Empty field name OR use 'EditorFor' as a workaround with parameter name=null.</value>
```
dougbu: Suggest `The name of an HTML field cannot be null or empty. Instead use methods {1}.{2} or {3}.{4} with a non-empty {5} argument value.` and pass `typeof(IHtmlHelper).FullName`, `nameof(IHtmlHelper.Editor)`, `typeof(IHtmlHelper<>).FullName`, `nameof(IHtmlHelper<>.EditorFor)`, `"htmlFieldName"`.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -969,7 +969,7 @@ internal static object GetModelStateValue(ViewContext viewContext, string key, T
...
```
dougbu: This is the only change you're actually testing. Update `DefaultHtmlGeneratorTest` (which tests `GetCurrentValues()`), `HtmlHelperCheckboxTest`, `HtmlHelperSelectTest`, the relevant tag helper tests e.g. `TextAreaTagHelperTest`, and perhaps `DefaultEditorTemplatesTest` to increase coverage. (Is :ok: that the HTML helper tests don't exercise `TextArea` and `ValidationMessage` since the corresponding `IHtmlGenerator` methods will be tested through the tag helpers.)

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultHtmlGenerator.cs
kulmugdha: Okay

src/Microsoft.AspNet.Mvc.ViewFeatures/Resources.resx
kulmugdha: @dougbu Thanks! 
There is no method `EditorFor` in `IHtmlHelper`. There is `EditorForModel`. Did you mean that? Also `EditorForModel` is an extension method so, `nameof` doesn't work with it. So I will have to hard code the name if that's the method you intend here.

dougbu: I meant the method that exists on the generic interface.

Looks like `nameof` doesn't like the open generic type; use `nameof(IHtmlHelper<object>.EditorFor)`.

kulmugdha:
@dougbu I have incorporated the feedback.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -413,7 +413,15 @@ public virtual HtmlString GenerateAntiforgery([NotNull] ViewContext viewContext)
             var fullName = GetFullHtmlFieldName(viewContext, expression);
             if (string.IsNullOrEmpty(fullName))
             {
-                throw new ArgumentException(Resources.ArgumentCannotBeNullOrEmpty, nameof(expression));
+                throw new ArgumentException(
+                    string.Format(
+                        Resources.HtmlGenerator_FieldNameCannotBeNullOrEmpty,
```
dougbu: Use `Resources.FormatHtmlGenerator_FieldNameCannotBeNullOrEmpty(...)` and get rid of the explicit `string.Format()` calls.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/DefaultHtmlGeneratorTest.cs
```diff
@@ -60,6 +60,122 @@ public void GetCurrentValues_WithNullExpressionResult_ReturnsNull(bool allowMult
...
+
+            var expected = string.Format(
+                            "The name of an HTML field cannot be null or empty. Instead " +
+                            "use methods {0}.{1} or {2}.{3} with a non-empty {4} " +
+                            "argument value.\r\nParameter name: expression",
+                            typeof(IHtmlHelper).FullName,
+                            nameof(IHtmlHelper.Editor),
+                            typeof(IHtmlHelper<>).FullName,
+                            nameof(IHtmlHelper<object>.EditorFor),
+                            "htmlFieldName");
```
dougbu: 1. Helpful in tests to expand everything so it's possible to see the final message. But at least use string interpolation so readers don't need to figure out which argument goes where.
2. Don't indent more than Visual Studio does by default plus, in a few cases, 4 spaces to offset something nested. This applies to many new tests -- some lines almost seem right-justified :smiley_cat:

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/DefaultHtmlGeneratorTest.cs
```diff
@@ -60,6 +60,122 @@ public void GetCurrentValues_WithNullExpressionResult_ReturnsNull(bool allowMult
...
+
+            // Act and assert
+            var ex = Assert.Throws<ArgumentException>(() => htmlGenerator.GetCurrentValues(
+                                            viewContext,
+                                            modelExplorer,
+                                            expression: null,
+                                            allowMultiple: true));
```
dougbu: Use the `Throws<T>` overload that has a `paramName` parameter and pass "expression".

dougbu:
:watch: very close

kulmugdha:
@dougbu Please have a look.

dougbu:
Travis build is failing due to "\r\n" in expected error messages. Use `Environment.NewLine` instead.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperHiddenTest.cs
```diff
@@ -403,10 +403,14 @@ public void HiddenWithEmptyNameAndPrefixThrows()
...
+            var expected = "The name of an HTML field cannot be null or empty. Instead use methods " +
+                "Microsoft.AspNet.Mvc.Rendering.IHtmlHelper.Editor or Microsoft.AspNet.Mvc.Rendering." +
+                "IHtmlHelper`1.EditorFor with a non-empty htmlFieldName argument value." +
+                "\r\nParameter name: expression";
 
             // Act and Assert
-            ExceptionAssert.ThrowsArgumentNullOrEmpty(() => helper.Hidden(string.Empty, string.Empty, attributes),
-                                                      "expression");
+            ExceptionAssert.Throws<ArgumentException>(
+                () => helper.Hidden(string.Empty, string.Empty, attributes), expected);
```
dougbu: Use `ExceptionAssert.ThrowsArgument()` instead. That will check the parameter as well.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -413,7 +413,14 @@ public virtual HtmlString GenerateAntiforgery([NotNull] ViewContext viewContext)
...
             {
-                throw new ArgumentException(Resources.ArgumentCannotBeNullOrEmpty, nameof(expression));
+                throw new ArgumentException(
+                        Resources.FormatHtmlGenerator_FieldNameCannotBeNullOrEmpty(
+                            typeof(IHtmlHelper).FullName,
+                            nameof(IHtmlHelper.Editor),
+                            typeof(IHtmlHelper<>).FullName,
+                            nameof(IHtmlHelper<object>.EditorFor),
+                            "htmlFieldName"),
+                        nameof(expression));
```
dougbu: Remove 4 spaces on the argument lines for all changes in this class.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperPasswordTest.cs
```diff
@@ -128,10 +128,15 @@ public void PasswordWithEmptyNameAndPrefixThrows()
...
             var value = string.Empty;
+            var expected = "The name of an HTML field cannot be null or empty. Instead use methods " +
+                "Microsoft.AspNet.Mvc.Rendering.IHtmlHelper.Editor or Microsoft.AspNet.Mvc.Rendering." +
+                "IHtmlHelper`1.EditorFor with a non-empty htmlFieldName argument value." +
+                "\r\nParameter name: expression";
 
             // Act and Assert
-            ExceptionAssert.ThrowsArgumentNullOrEmpty(() => helper.Password(name, value, htmlAttributes: null),
-                                                      "expression");
+            ExceptionAssert.Throws<ArgumentException>(
```
dougbu: Use ExceptionAssert.ThrowsArgument() here too.

dougbu:
:shipit: after addressing small comments and a successful Travis build.

dougbu:
@kulmugdha please go through the CLA process. Takes about 30 seconds...

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultHtmlGenerator.cs
kulmugdha: the spaces are for arguments to the Resources.FormatHtmlGenerator_FieldNameCannotBeNullOrEmpty method as it is a level deeper.

kulmugdha:
Checked in a6a7903b846eadd62f43866beedea826e75e2b3f to dev.

Issue #2778, opened 2015-07-02
Label Inside EditorTemplate Disappears During Validation


Reafidy:
I am using ASP.net MVC 6 and the following EditorTemplate:

    <div class="form-group">        
        @Html.LabelFor(m => m, new { @class = "control-label col-md-2" })
        <div class="col-md-10">
            @Html.TextBox("", ViewData.TemplateInfo.FormattedModelValue,  htmlAttributes)
            @Html.ValidationMessageFor(m => m, null, new { @class = "help-block" })
        </div>
    </div>

When attempting to save the form with a deliberate error, the page refreshes and correctly shows the validation error but the label completely disappears.   It is not hidden it is completely gone from the html source.

Here is my EditorFor:

     @Html.EditorFor(model => model.Description, new { htmlAttributes = new { @class = "form-control" } })

A bug?

danroth27:
@ajaybhargavb Please try to reproduce this

Reafidy:
New to Github and At the risk of sounding completely dumb, what does the '2- working' label mean?

danroth27:
@Reafidy It just means that the person assigned to the issue is now working on the issue.

Reafidy:
Thanks.

ajaybhargavb:
92f3e21fe62ea40342f2eb462ee8a92361ad591b

PR #2820, opened 2015-07-15
Fix: Metadata ignored for non model-specific EditorTemplate

ajaybhargavb:
Issue - #2778 

@dougbu

### DIFF
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Internal/TemplateBuilder.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Internal/TemplateBuilder.cs
@@ -72,7 +72,8 @@ namespace Microsoft.AspNet.Mvc.Rendering.Internal
             // We need to copy the ModelExplorer to copy the model metadata. Otherwise we might
             // lose track of the model type/property. Passing null here explicitly, because
             // this might be a typed VDD, and the model value might not be compatible.
-            var viewData = new ViewDataDictionary(_viewData, model: null);
+            // Create VDD of type object so it retains the correct metadata when the model type is not known.
+            var viewData = new ViewDataDictionary<object>(_viewData, model: null);
 
             // We're setting ModelExplorer in order to preserve the model metadata of the original
             // _viewData even though _model may be null.
--- a/test/Microsoft.AspNet.Mvc.FunctionalTests/HtmlGenerationTest.cs
+++ b/test/Microsoft.AspNet.Mvc.FunctionalTests/HtmlGenerationTest.cs
@@ -539,5 +539,41 @@ Products: Laptops (3)";
             Assert.Equal(expectedContent.Trim(), responseContent, ignoreLineEndingDifferences: true);
 #endif
         }
+
+        [Fact]
+        public async Task EditorTemplateWithNoModel_RendersWithCorrectMetadata()
+        {
+            // Arrange
+            var server = TestHelper.CreateServer(_app, SiteName, _configureServices);
+            var client = server.CreateClient();
+            var expected = PlatformNormalizer.NormalizeContent(
+                "<label class=\"control-label col-md-2\" for=\"Name\">ItemName</label>" + Environment.NewLine +
+                "<input id=\"Name\" name=\"Name\" type=\"text\" value=\"\" />" + Environment.NewLine + Environment.NewLine +
+                "<label class=\"control-label col-md-2\" for=\"Id\">ItemNo</label>" + Environment.NewLine +
+                "<input data-val=\"true\" data-val-required=\"The ItemNo field is required.\" id=\"Id\" name=\"Id\" type=\"text\" value=\"\" />" +
+                Environment.NewLine + Environment.NewLine);
+
+            // Act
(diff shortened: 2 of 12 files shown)


test/Microsoft.AspNet.Mvc.FunctionalTests/HtmlGenerationTest.cs
```diff
@@ -539,5 +539,39 @@ public class HtmlGenerationTest
...
+            var expected = "<label class=\"control-label col-md-2\" for=\"Name\">ItemName</label>\r\n" +
+                "<input id=\"Name\" name=\"Name\" type=\"text\" value=\"\" />\r\n\r\n" +
+                "<label class=\"control-label col-md-2\" for=\"Id\">ItemNo</label>\r\n" +
+                "<input data-val=\"true\" data-val-required=\"The ItemNo field is required.\" id=\"Id\" name=\"Id\" type=\"text\" value=\"\" />\r\n\r\n";
+
+            // Act
+            var response = await client.GetStringAsync("http://localhost/HtmlGeneration_Home/ItemUsingSharedEditorTemplate");
+
+            // Assert
+            Assert.Equal(expected, response);
```
dougbu: Need `ignoreLineEndingDifferences: true` here and in next test. This is the reason Travis is currently failing.

test/Microsoft.AspNet.Mvc.FunctionalTests/HtmlGenerationTest.cs
```diff
@@ -539,5 +539,39 @@ public class HtmlGenerationTest
...
+        [Fact]
+        public async Task EditorTemplateWithNoModel_RendersWithCorrectMetadata()
+        {
+            // Arrange
+            var server = TestHelper.CreateServer(_app, SiteName, _configureServices);
+            var client = server.CreateClient();
+            var expected = "<label class=\"control-label col-md-2\" for=\"Name\">ItemName</label>\r\n" +
+                "<input id=\"Name\" name=\"Name\" type=\"text\" value=\"\" />\r\n\r\n" +
+                "<label class=\"control-label col-md-2\" for=\"Id\">ItemNo</label>\r\n" +
+                "<input data-val=\"true\" data-val-required=\"The ItemNo field is required.\" id=\"Id\" name=\"Id\" type=\"text\" value=\"\" />\r\n\r\n";
```
dougbu: Don't write "\r\n". Instead use a verbatim `string` literal; that will be easier to read, much more consistent with other tests, and platform-neutral.

test/WebSites/HtmlGenerationWebSite/Views/Shared/EditorTemplates/Item.cshtml
```diff
@@ -0,0 +1,4 @@
+﻿@using HtmlGenerationWebSite.Models
+@model Item
+@Html.Label(nameof(Model.Description))
+@Html.TextBoxFor(m => m.Description)
```
dougbu: To match the shared case more closely, add a `String.cshtml` template containing `@model string` plus the lines in `Common.cshtml`. The expectations shouldn't need to change...

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Internal/TemplateBuilder.cs
```diff
@@ -72,7 +72,7 @@ public string Build()
             // We need to copy the ModelExplorer to copy the model metadata. Otherwise we might
             // lose track of the model type/property. Passing null here explicitly, because
             // this might be a typed VDD, and the model value might not be compatible.
```
dougbu: Add a bit to this comment about why the next line creates a `ViewDataDictionary<object>`.

dougbu:
:watch:

dougbu:
Also please add a test or two to `RazorPageActivatorTest` covering expected `RazorPage.ViewData` values when activation context contains a `ViewDataDictionary<object>` instance.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Internal/TemplateBuilder.cs
ajaybhargavb: Added.

ajaybhargavb:
Updated.

test/Microsoft.AspNet.Mvc.FunctionalTests/HtmlGenerationTest.cs
```diff
@@ -539,5 +539,42 @@ public class HtmlGenerationTest
...
+            Assert.Equal(expected, response);
+        }
+
+        [Fact]
+        public async Task EditorTemplateWithSpecificModel_RendersWithCorrectMetadata()
+        {
+            // Arrange
+            var server = TestHelper.CreateServer(_app, SiteName, _configureServices);
+            var client = server.CreateClient();
+            var expected = PlatformNormalizer.NormalizeContent(
```
dougbu: Should not need `PlatformNormalizer.NormalizeContent()` here.

dougbu:
:shipit: after removing extra `NormalizeContent` and squashing

rynowak:
:shipit: from me

PR #2827, opened 2015-07-17
Remove a VS build warning

dougbu:
- was "CS0649: Field is never assigned to, and will always have its default value `null'"
 - visible only in VS builds due to aspnet/dnx#2284

nit: let VS do its thing with Microsoft.AspNet.Mvc.Localization.xproj


dougbu:
FYI only. Will merge immediately.

dougbu:
3056c8e

Issue #2382, opened 2015-04-14
Transition Script/Link TagHelper to utilize Pre/Post Element properties on TagHelperOutput.


NTaylorMullen:
With the completion of aspnet/Razor#341 there is now `PreElement` and `PostElement` properties on `TagHelperOutput`. Today we rebuild our `ScriptTagHelper` and `LinkTagHelper` contents to give the appearance of `PostElement` behavior; we should refactor this to use the new `TagHelperOutput` properties.

NTaylorMullen:
Why is this a bug?

dougbu:
aspnet/Razor#225 includes a [comment](https://github.com/aspnet/Razor/issues/225#issuecomment-95979090) that mentions complete support for maintaining the original attribute order involves that fix (done), #2639, completing this task (#2382), as well as
- use `PostElement` for `type="checkbox"` case in `<input/>` tag helper and merge added attributes for original element

My comment is incorrect because attribute order will be maintained correctly as soon as #2639 is done. But we have agreed `PostContent` avoids lots of special-case code in the `<link>` and `<script>` tag helpers. So why not the `<input>` tag helper as well?

Question for @Triage: Should we do the `checkbox` task as part of #2382 (this issue) or in another issue?

Issue #2639, opened 2015-06-01
Modify TagHelperOutput.CopyHtmlAttribute to maintain attribute order.


NTaylorMullen:
Once aspnet/Razor#225 is completed attributes will be in-order rather than in a semi-random order. With this change we should ensure that `TagHelperOutput.CopyHtmlAttribute` re-adds attributes to `TagHelperOutput` to their "most correct" location.

NTaylorMullen:
f1eefdb650c5f7bd07ccc4499867ccaaf6dc6382

PR #2829, opened 2015-07-17
Use PostElement in Script and Link tag helpers

ajaybhargavb:
Issue - #2382 
@dougbu @NTaylorMullen 

This includes changes for script and link tag helper. Will update the PR if we decide to refactor `InputTagHelper` as part of this issue. Ref: https://github.com/aspnet/Mvc/issues/2382#issuecomment-122384250

### DIFF
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
@@ -221,23 +221,35 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
                 return;
             }
 
-            // Get the highest matched mode
-            var mode = modeResult.FullMatches.Select(match => match.Mode).Max();
-
             // NOTE: Values in TagHelperOutput.Attributes may already be HTML-encoded.
             var attributes = new TagHelperAttributeList(output.Attributes);
 
+            if (AppendVersion == true)
+            {
+                EnsureFileVersionProvider();
+
+                var attributeStringValue = output.Attributes[HrefAttributeName]?.Value as string;
+                if (attributeStringValue != null)
+                {
+                    output.Attributes[HrefAttributeName].Value =
+                        _fileVersionProvider.AddFileVersionToPath(attributeStringValue);
+                }
+            }
+
             var builder = new DefaultTagHelperContent();
 
-            if (mode == Mode.Fallback && string.IsNullOrEmpty(HrefInclude) || mode == Mode.AppendVersion)
-            {
-                // No globbing to do, just build a <link /> tag to match the original one in the source file.
-                // Or just add file version to the link tag.
-                BuildLinkTag(attributes, builder);
-            }
-            else
+            // Get the highest matched mode
+            var mode = modeResult.FullMatches.Select(match => match.Mode).Max();
+
+            if (mode == Mode.GlobbedHref || mode == Mode.Fallback && !string.IsNullOrEmpty(HrefInclude))
             {
                 BuildGlobbedLinkTags(attributes, builder);
+                if (string.IsNullOrEmpty(Href))
+                {
+                    // Only HrefInclude is specified. Don't render the original tag.
+                    output.TagName = null;
+                    output.Content.SetContent(string.Empty);
+                }
             }
 
             if (mode == Mode.Fallback)
@@ -245,21 +257,25 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
                 BuildFallbackBlock(builder);
             }
 
-            // We've taken over tag rendering, so prevent rendering the outer tag
-            output.TagName = null;
-            output.Content.SetContent(builder);
+            output.PostElement.SetContent(builder);
         }
 
         private void BuildGlobbedLinkTags(TagHelperAttributeList attributes, TagHelperContent builder)
         {
             EnsureGlobbingUrlBuilder();
 
-            // Build a <link /> tag for each matched href as well as the original one in the source file
-            var urls = GlobbingUrlBuilder.BuildUrlList(Href, HrefInclude, HrefExclude);
+            // Build a <link /> tag for each matched href.
+            var urls = GlobbingUrlBuilder.BuildUrlList(null, HrefInclude, HrefExclude);
             foreach (var url in urls)
             {
                 // "url" values come from bound attributes and globbing. Must always be non-null.
                 Debug.Assert(url != null);
+
+                if (string.Equals(Href, url, StringComparison.OrdinalIgnoreCase))
+                {
+                    // Don't build duplicate link tag for the original href url.
+                    continue;
+                }
 
                 attributes[HrefAttributeName] = url;
                 BuildLinkTag(attributes, builder);
@@ -332,7 +348,6 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
 
         private void BuildLinkTag(TagHelperAttributeList attributes, TagHelperContent builder)
         {
-            EnsureFileVersionProvider();
             builder.Append("<link ");
 
             foreach (var attribute in attributes)
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
@@ -172,7 +172,7 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
         protected internal GlobbingUrlBuilder GlobbingUrlBuilder { get; set; }
 
         /// <inheritdoc />
-        public override async Task ProcessAsync(TagHelperContext context, TagHelperOutput output)
+        public override void Process(TagHelperContext context, TagHelperOutput output)
         {
             // Pass through attribute that is also a well-known HTML attribute.
             if (Src != null)
@@ -190,24 +190,35 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
                 return;
             }
 
-            // Get the highest matched mode
-            var mode = modeResult.FullMatches.Select(match => match.Mode).Max();
-
             // NOTE: Values in TagHelperOutput.Attributes may already be HTML-encoded.
             var attributes = new TagHelperAttributeList(output.Attributes);
 
-            var builder = new DefaultTagHelperContent();
-            var originalContent = await context.GetChildContentAsync();
+            if (AppendVersion == true)
+            {
+                EnsureFileVersionProvider();
 
-            if (mode == Mode.Fallback && string.IsNullOrEmpty(SrcInclude) || mode == Mode.AppendVersion)
-            {
-                // No globbing to do, just build a <script /> tag to match the original one in the source file
-                // Or just add file version to the script tag.
-                BuildScriptTag(originalContent, attributes, builder);
+                var attributeStringValue = output.Attributes[SrcAttributeName]?.Value as string;
+                if (attributeStringValue != null)
+                {
+                    output.Attributes[SrcAttributeName].Value =
+                        _fileVersionProvider.AddFileVersionToPath(attributeStringValue);
+                }
             }
-            else
+
+            var builder = new DefaultTagHelperContent();
+
+            // Get the highest matched mode
+            var mode = modeResult.FullMatches.Select(match => match.Mode).Max();
+
+            if (mode == Mode.GlobbedSrc || mode == Mode.Fallback && !string.IsNullOrEmpty(SrcInclude))
             {
-                BuildGlobbedScriptTags(originalContent, attributes, builder);
+                BuildGlobbedScriptTags(attributes, builder);
+                if (string.IsNullOrEmpty(Src))
+                {
+                    // Only SrcInclude is specified. Don't render the original tag.
+                    output.TagName = null;
+                    output.Content.SetContent(string.Empty);
+                }
             }
 
             if (mode == Mode.Fallback)
@@ -215,41 +226,36 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
                 BuildFallbackBlock(attributes, builder);
             }
 
-            // We've taken over tag rendering, so prevent rendering the outer tag
-            output.TagName = null;
-            output.Content.SetContent(builder);
+            output.PostElement.SetContent(builder);
         }
 
         private void BuildGlobbedScriptTags(
-            TagHelperContent originalContent,
             TagHelperAttributeList attributes,
             TagHelperContent builder)
         {
             EnsureGlobbingUrlBuilder();
 
             // Build a <script> tag for each matched src as well as the original one in the source file
-            var urls = GlobbingUrlBuilder.BuildUrlList(Src, SrcInclude, SrcExclude);
+            var urls = GlobbingUrlBuilder.BuildUrlList(null, SrcInclude, SrcExclude);
             foreach (var url in urls)
             {
                 // "url" values come from bound attributes and globbing. Must always be non-null.
                 Debug.Assert(url != null);
 
-                var content = originalContent;
-                if (!string.Equals(url, Src, StringComparison.OrdinalIgnoreCase))
+                if (string.Equals(url, Src, StringComparison.OrdinalIgnoreCase))
                 {
-                    // Do not copy content into added <script/> elements.
-                    content = null;
+                    // Don't build duplicate script tag for the original source url.
+                    continue;
                 }
 
                 attributes[SrcAttributeName] = url;
-                BuildScriptTag(content, attributes, builder);
+                BuildScriptTag(attributes, builder);
             }
         }
 
         private void BuildFallbackBlock(TagHelperAttributeList attributes, DefaultTagHelperContent builder)
         {
             EnsureGlobbingUrlBuilder();
-            EnsureFileVersionProvider();
 
             var fallbackSrcs = GlobbingUrlBuilder.BuildUrlList(FallbackSrc, FallbackSrcInclude, FallbackSrcExclude);
             if (fallbackSrcs.Any())
@@ -330,11 +336,9 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
         }
 
         private void BuildScriptTag(
-            TagHelperContent content,
             TagHelperAttributeList attributes,
             TagHelperContent builder)
         {
-            EnsureFileVersionProvider();
             builder.Append("<script");
 
             foreach (var attribute in attributes)
@@ -356,9 +360,7 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
                 AppendAttribute(builder, attribute.Name, attributeValue, escapeQuotes: false);
             }
 
-            builder.Append(">")
-                   .Append(content)
-                   .Append("</script>");
+            builder.Append("></script>");
         }
 
         private void AppendAttribute(TagHelperContent content, string key, object value, bool escapeQuotes)
--- a/test/Microsoft.AspNet.Mvc.TagHelpers.Test/LinkTagHelperTest.cs
+++ b/test/Microsoft.AspNet.Mvc.TagHelpers.Test/LinkTagHelperTest.cs
@@ -32,16 +32,15 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
         {
             get
             {
-                // outputAttributes, expectedAttributeString
-                return new TheoryData<TagHelperAttributeList, string>
+                // outputAttributes
+                return new TheoryData<TagHelperAttributeList>
                 {
                     {
                         new TagHelperAttributeList
                         {
                             { "hello", "world" },
                             { "hello", "world2" }
-                        },
-                        "hello=\"HtmlEncode[[world]]\" hello=\"HtmlEncode[[world2]]\""
+                        }
                     },
                     {
                         new TagHelperAttributeList
@@ -49,16 +48,14 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
                             { "hello", "world" },
                             { "hello", "world2" },
                             { "hello", "world3" }
-                        },
-                        "hello=\"HtmlEncode[[world]]\" hello=\"HtmlEncode[[world2]]\" hello=\"HtmlEncode[[world3]]\""
+                        }
                     },
                     {
                         new TagHelperAttributeList
                         {
                             { "HelLO", "world" },
                             { "HELLO", "world2" }
-                        },
-                        "HelLO=\"HtmlEncode[[world]]\" HELLO=\"HtmlEncode[[world2]]\""
+                        }
                     },
                     {
                         new TagHelperAttributeList
@@ -66,16 +63,14 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
                             { "Hello", "world" },
                             { "HELLO", "world2" },
                             { "hello", "world3" }
-                        },
-                        "Hello=\"HtmlEncode[[world]]\" HELLO=\"HtmlEncode[[world2]]\" hello=\"HtmlEncode[[world3]]\""
+                        }
                     },
                     {
                         new TagHelperAttributeList
                         {
                             { "HeLlO", "world" },
                             { "hello", "world2" }
-                        },
-                        "HeLlO=\"HtmlEncode[[world]]\" hello=\"HtmlEncode[[world2]]\""
+                        }
                     },
                 };
             }
@@ -83,9 +78,7 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
 
         [Theory]
         [MemberData(nameof(MultiAttributeSameNameData))]
-        public void HandlesMultipleAttributesSameNameCorrectly(
-            TagHelperAttributeList outputAttributes,
-            string expectedAttributeString)
+        public void HandlesMultipleAttributesSameNameCorrectly(TagHelperAttributeList outputAttributes)
         {
             // Arrange
             var allAttributes = new TagHelperAttributeList(
@@ -125,12 +118,14 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
                 FallbackTestValue = "hidden",
                 Href = "test.css",
             };
+            var expectedAttributes = new TagHelperAttributeList(output.Attributes);
+            expectedAttributes.Add(new TagHelperAttribute("href", "test.css"));
 
             // Act
             helper.Process(context, output);
 
             // Assert
-            Assert.StartsWith("<link " + expectedAttributeString + " rel=\"stylesheet\"", output.Content.GetContent());
+            Assert.Equal(expectedAttributes, output.Attributes);
         }
 
         public static TheoryData RunsWhenRequiredAttributesArePresent_Data
@@ -139,28 +134,6 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
             {
                 return new TheoryData<TagHelperAttributeList, Action<LinkTagHelper>>
                 {
-                    {
-                        new TagHelperAttributeList
-                        {
-                            ["asp-href-include"] = "*.css"
-                        },
-                        tagHelper =>
-                        {
-                            tagHelper.HrefInclude = "*.css";
-                        }
-                    },
-                    {
-                        new TagHelperAttributeList
-                        {
-                            ["asp-href-include"] = "*.css",
-                            ["asp-href-exclude"] = "*.min.css"
-                        },
-                        tagHelper =>
-                        {
-                            tagHelper.HrefInclude = "*.css";
-                            tagHelper.HrefExclude = "*.min.css";
-                        }
-                    },
                     {
                         new TagHelperAttributeList
                         {
@@ -194,42 +167,6 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
                         }
                     },
                     // File Version
-                    {
-                        new TagHelperAttributeList
-                        {
-                            ["asp-append-version"] = "true"
-                        },
-                        tagHelper =>
-                        {
-                            tagHelper.AppendVersion = true;
-                        }
-                    },
-                    {
-                        new TagHelperAttributeList
-                        {
-                            ["asp-href-include"] = "*.css",
-                            ["asp-append-version"] = "true"
-                        },
-                        tagHelper =>
-                        {
-                            tagHelper.HrefInclude = "*.css";
-                            tagHelper.AppendVersion = true;
-                        }
-                    },
-                    {
-                        new TagHelperAttributeList
-                        {
-                            ["asp-href-include"] = "*.css",
-                            ["asp-href-exclude"] = "*.min.css",
-                            ["asp-append-version"] = "true"
-                        },
-                        tagHelper =>
-                        {
-                            tagHelper.HrefInclude = "*.css";
-                            tagHelper.HrefExclude = "*.min.css";
-                            tagHelper.AppendVersion = true;
-                        }
-                    },
                     {
                         new TagHelperAttributeList
                         {
@@ -282,6 +219,9 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
             var logger = new Mock<ILogger<LinkTagHelper>>();
             var hostingEnvironment = MakeHostingEnvironment();
             var viewContext = MakeViewContext();
+            var globbingUrlBuilder = new Mock<GlobbingUrlBuilder>();
+            globbingUrlBuilder.Setup(g => g.BuildUrlList(It.IsAny<string>(), It.IsAny<string>(), It.IsAny<string>()))
+                .Returns(new[] { "/common.css" });
 
             var helper = new LinkTagHelper(
                 logger.Object,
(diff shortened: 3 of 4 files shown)


src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -229,15 +229,19 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
...
-                // Or just add file version to the link tag.
-                BuildLinkTag(attributes, builder);
+                output.Attributes[HrefAttributeName].Value = AppendFileVersionIfApplicable(Href);
             }
-            else
+
+            if (mode == Mode.GlobbedHref || mode == Mode.Fallback && !string.IsNullOrEmpty(HrefInclude))
             {
                 BuildGlobbedLinkTags(attributes, builder);
+                if (string.IsNullOrEmpty(Href))
```
dougbu: This block needs a comment explaining _why_ the lack of `href` attribute means the original element should not be output.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -245,22 +249,26 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
...
+            // Build a <link /> tag for each matched href.
+            var urls = GlobbingUrlBuilder.BuildUrlList(null, HrefInclude, HrefExclude);
             foreach (var url in urls)
             {
                 // "url" values come from bound attributes and globbing. Must always be non-null.
                 Debug.Assert(url != null);
 
+                if (string.Equals(Href, url, StringComparison.OrdinalIgnoreCase))
+                {
+                    // Don't build script tag for the original source url.
```
dougbu: Since the original `href` attribute is no longer passed into `BuildUrlList()`, this comment seems incorrect. Is this actually about protecting against duplicates of the `href` value in the globbed results? Suspect this block isn't needed.

dougbu: At least correct the comment.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -361,6 +366,21 @@ private void BuildLinkTag(TagHelperAttributeList attributes, TagHelperContent bu
             builder.Append("/>");
         }
 
+        private object AppendFileVersionIfApplicable(object attributeValue)
```
dougbu: This refactoring turns things means fallback handling is inconsistent and results in duplicate checks of `AppendVersion` in `BuildLinkTag()`. Also doesn't seem related to fixing #2382. Feel free to follow up later with a separate cleanup PR if you feel strongly about the small amount of duplicate code.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -229,15 +229,19 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
 
             var builder = new DefaultTagHelperContent();
 
-            if (mode == Mode.Fallback && string.IsNullOrEmpty(HrefInclude) || mode == Mode.AppendVersion)
+            if (!string.IsNullOrEmpty(Href))
```
dougbu: This condition seems incorrect since the previous code would have added the file version to any `href` attribute in the `output` as long as `AppendVersion` was `true`. (Another negative aspect of the `AppendFileVersionIfApplicable()` refactoring is separating this condition from those in the new method.)

In any case handle the `href` attribute in the original element immediately after restoring it and bring the conditions back together, removing `AppendFileVersionIfApplicable()`.

Separately, the previous version of this class did not rely on `Href` in this fashion, allowing earlier tag helpers to add an `href` attribute to the `output` without strange side-effects here.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -229,15 +229,19 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
...
             {
-                // No globbing to do, just build a <link /> tag to match the original one in the source file.
-                // Or just add file version to the link tag.
-                BuildLinkTag(attributes, builder);
+                output.Attributes[HrefAttributeName].Value = AppendFileVersionIfApplicable(Href);
             }
-            else
+
+            if (mode == Mode.GlobbedHref || mode == Mode.Fallback && !string.IsNullOrEmpty(HrefInclude))
```
dougbu: FYI it would have been easier to understand your changes if you had left the condition in the original order. No change needed now but it was harder than necessary to see you'd simply remove the positive part of the earlier `if` / `else`.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -269,6 +277,8 @@ private void BuildGlobbedLinkTags(TagHelperAttributeList attributes, TagHelperCo
         private void BuildFallbackBlock(TagHelperContent builder)
         {
             EnsureGlobbingUrlBuilder();
+            EnsureFileVersionProvider();
```
dougbu: Why did you move this call?

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -215,34 +218,30 @@ public class ScriptTagHelper : TagHelper
...
-            var urls = GlobbingUrlBuilder.BuildUrlList(Src, SrcInclude, SrcExclude);
+            var urls = GlobbingUrlBuilder.BuildUrlList(null, SrcInclude, SrcExclude);
             foreach (var url in urls)
             {
                 // "url" values come from bound attributes and globbing. Must always be non-null.
                 Debug.Assert(url != null);
 
-                var content = originalContent;
-                if (!string.Equals(url, Src, StringComparison.OrdinalIgnoreCase))
+                if (string.Equals(Src, url, StringComparison.OrdinalIgnoreCase))
```
dougbu: Why did you reverse the arguments here?

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -307,6 +306,44 @@ private void BuildFallbackBlock(TagHelperAttributeList attributes, DefaultTagHel
             }
         }
 
+        private void BuildScriptTag(
```
dougbu: Move this back down where the `BuildScriptTag()` method was before. That way I'll be able to see what changed here.

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -197,17 +197,20 @@ public class ScriptTagHelper : TagHelper
             var attributes = new TagHelperAttributeList(output.Attributes);
 
             var builder = new DefaultTagHelperContent();
-            var originalContent = await context.GetChildContentAsync();
```
dougbu: :dancers:

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -197,17 +197,20 @@ public class ScriptTagHelper : TagHelper
...
 
-            if (mode == Mode.Fallback && string.IsNullOrEmpty(SrcInclude) || mode == Mode.AppendVersion)
+            if (!string.IsNullOrEmpty(Src))
```
dougbu: All comments in `LinkTagHelper` apply here.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/LinkTagHelperTest.cs
```diff
@@ -125,12 +118,14 @@ public static TheoryData MultiAttributeSameNameData
                 FallbackTestValue = "hidden",
                 Href = "test.css",
             };
+            var expectedAttributes = output.Attributes;
+            expectedAttributes.Add(new TagHelperAttribute("href", "test.css"));
```
dougbu: Do not change the `output.Attributes` collection. Instead use `expectedAttributes = new TagHelperAttributeList(output.Attributes)`.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/LinkTagHelperTest.cs
```diff
@@ -298,12 +287,11 @@ public static TheoryData RunsWhenRequiredAttributesArePresent_Data
             helper.Process(context, output);
 
             // Assert
-            Assert.Null(output.TagName);
-            Assert.NotNull(output.Content);
-            Assert.True(output.IsContentModified);
+            Assert.True(output.IsContentModified || !string.IsNullOrEmpty(output.PostElement.GetContent()));
```
dougbu: Or'd conditions are clear indications you've got more than one test. Also the second condition isn't the best way to check `TagHelperContent`.

Split this test and `RunsWhenRequiredAttributesArePresent_Data` in two. Then `Assert.True(output.IsContentModified); Assert.False(output.PostElement.IsModified)` in one test and the opposite, plus `Assert.False(output.PostElement.IsModified)`, in the other.

If there isn't a clear split or the `IsContentModified` cases overlap the `PostElement.IsModified` ones, divide things up differently but always check both `bool` properties.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/LinkTagHelperTest.cs
```diff
@@ -298,12 +287,11 @@ public static TheoryData RunsWhenRequiredAttributesArePresent_Data
...
         }
 
         [Fact]
+        // Does not work as expected - https://github.com/aspnet/Mvc/issues/2639
```
dougbu: Remove this comment since the test executes successfully. We know fixing #2639 will involve changing tests. But may be worth changing the test name to something like `PreservesOrderOfNonHrefAttributes`.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/LinkTagHelperTest.cs
```diff
@@ -451,8 +440,9 @@ public static TheoryData DoesNotRunWhenARequiredAttributeIsMissing_Data
             helper.Process(context, output);
 
             // Assert
-            Assert.NotNull(output.TagName);
             Assert.False(output.IsContentModified);
+            Assert.Empty(output.Attributes);
+            Assert.Empty(output.PostElement.GetContent());
```
dougbu: `Assert.True(output.PostElement.IsEmpty)` instead throughout your changes. Only need to call `GetContent()` when looking for specific content.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/LinkTagHelperTest.cs
```diff
@@ -197,16 +192,6 @@ public static TheoryData RunsWhenRequiredAttributesArePresent_Data
                     {
                         new TagHelperAttributeList
                         {
-                            ["asp-append-version"] = "true"
-                        },
-                        tagHelper =>
-                        {
-                            tagHelper.AppendVersion = true;
-                        }
```
dougbu: Where did this case move to? It's a valid scenario and should be tested, perhaps in a new test.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/LinkTagHelperTest.cs
```diff
@@ -298,12 +287,11 @@ public static TheoryData RunsWhenRequiredAttributesArePresent_Data
             helper.Process(context, output);
 
             // Assert
-            Assert.Null(output.TagName);
```
dougbu: Can't this be `Assert.NotNull(output.TagName)`?

test/Microsoft.AspNet.Mvc.TagHelpers.Test/LinkTagHelperTest.cs
```diff
@@ -451,8 +440,9 @@ public static TheoryData DoesNotRunWhenARequiredAttributeIsMissing_Data
             helper.Process(context, output);
 
             // Assert
-            Assert.NotNull(output.TagName);
```
dougbu: Why remove this?

test/Microsoft.AspNet.Mvc.TagHelpers.Test/LinkTagHelperTest.cs
```diff
@@ -479,8 +469,9 @@ public void DoesNotRunWhenAllRequiredAttributesAreMissing()
...
```
dougbu: Why remove this?

test/Microsoft.AspNet.Mvc.TagHelpers.Test/LinkTagHelperTest.cs
```diff
@@ -311,8 +299,8 @@ public void PreservesOrderOfSourceAttributesWhenRun()
                 attributes: new TagHelperAttributeList
                 {
                     { "rel", new HtmlString("stylesheet") },
-                    { "data-extra", new HtmlString("something") },
                     { "href", "test.css" },
+                    { "data-extra", new HtmlString("something") },
```
dougbu: What required a change here?

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ScriptTagHelperTest.cs
```diff
@@ -68,15 +66,14 @@ public class ScriptTagHelperTest
                 FallbackTestExpression = "http://www.example.com/blank.js",
                 Src = "/blank.js",
             };
+            var expectedAttributes = output.Attributes;
+            expectedAttributes.Add(new TagHelperAttribute("src", "/blank.js"));
```
dougbu: See comments in `LinkTagHelperTest`

dougbu: Looks like most of those comments apply here.

dougbu:
:watch:

NTaylorMullen:
Waiting till Doug's comments are addressed to review.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
ajaybhargavb: Yes it is protecting against duplicates. Corrected the comment.

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
ajaybhargavb: Unintentional. Will undo.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/LinkTagHelperTest.cs
ajaybhargavb: This is covered in `RendersLinkTags_AddsFileVersion` test and a similar one in `SciptTagHelperTest`.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/LinkTagHelperTest.cs
ajaybhargavb: No. This will be `null` in the case when `Href` is not specified but `HrefInclude` is specified.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/LinkTagHelperTest.cs
ajaybhargavb: Because `output.TagName` is no longer set to `null` for all cases. It is set to null only for a special case as mentioned here https://github.com/aspnet/Mvc/pull/2829#discussion_r35039322.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/LinkTagHelperTest.cs
ajaybhargavb: https://github.com/aspnet/Mvc/pull/2829#discussion_r35039634

test/Microsoft.AspNet.Mvc.TagHelpers.Test/LinkTagHelperTest.cs
ajaybhargavb: Just to show that it doesn't maintain the order of `href` attribute.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
ajaybhargavb: I've removed `AppendFileVersionIfApplicable()` but I don't understand what side effects this condition might introduce.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
ajaybhargavb: Added.

ajaybhargavb:
Updated.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -229,15 +229,21 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
 
             var builder = new DefaultTagHelperContent();
 
-            if (mode == Mode.Fallback && string.IsNullOrEmpty(HrefInclude) || mode == Mode.AppendVersion)
+            if (!string.IsNullOrEmpty(Href) && AppendVersion == true)
```
NTaylorMullen: Move this `if` above the 3 variables prior to this (so the variables are next to their first uses)

NTaylorMullen: Actually the `var attributes` needs to stay, just move the `mode` and `builder before their first uses

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -245,22 +251,26 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
...
         }
 
         private void BuildGlobbedLinkTags(TagHelperAttributeList attributes, TagHelperContent builder)
         {
             EnsureGlobbingUrlBuilder();
 
-            // Build a <link /> tag for each matched href as well as the original one in the source file
-            var urls = GlobbingUrlBuilder.BuildUrlList(Href, HrefInclude, HrefExclude);
+            // Build a <link /> tag for each matched href.
+            var urls = GlobbingUrlBuilder.BuildUrlList(null, HrefInclude, HrefExclude);
```
NTaylorMullen: `BuildUrlList` isn't used anywhere else except for link and script tag helpers (at least that I can see). Might as well re-write the method to only accept an include and exclude.

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
NTaylorMullen: Curious: Why are we deciding to remove this capability in this PR?

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
dougbu: @NTaylorMullen why can't the `attributes` variable move?

NTaylorMullen: Because it'd add the `href` attribute to the copied list of attributes which would then run through `BuildGlobbedLinkTags` => `BuildLinkTag` resulting in an `href` attribute that'd have two file version values added to it.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -229,15 +229,21 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
...
             var builder = new DefaultTagHelperContent();
 
-            if (mode == Mode.Fallback && string.IsNullOrEmpty(HrefInclude) || mode == Mode.AppendVersion)
+            if (!string.IsNullOrEmpty(Href) && AppendVersion == true)
             {
-                // No globbing to do, just build a <link /> tag to match the original one in the source file.
-                // Or just add file version to the link tag.
-                BuildLinkTag(attributes, builder);
+                EnsureFileVersionProvider();
+                output.Attributes[HrefAttributeName].Value = _fileVersionProvider.AddFileVersionToPath(Href);
```
dougbu: Do not use `Href` here. That causes this helper to ignore changes earlier tag helpers may have made. Must instead use `output.Attributes[HrefAttributeName].Value as string`. Check that expression for `null` (not null or empty) rather than checking `string.IsNullOrEmpty(Href)` at all.

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
dougbu: @NTaylorMullen :-1: the method is also used in these classes with the `FallbackHref` value.

NTaylorMullen: Ah I see :+1:

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
dougbu: Huh? The `attributes` collection is not relevant here. This block is about updating the original tag's attributes. Globbing results in new tags added to the `PostElement`.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
dougbu: Also, after removing the `string.IsNullOrEmpty(Href)` check, we won't need `EnsureFileVersionProvider()` calls anywhere else e.g.
```
  if (AppendVersion == true)
  {
    EnsureFileVersionProvider();
    var attributeStringValue = output.Attributes[HrefAttributeName].Value as string;
    if (attributeStringValue != null)
    {
      output.Attributes[HrefAttributeName].Value = _fileVersionProvider.AddFileVersionToPath(attributeStringValue);
    }
  }
```

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
ajaybhargavb: No. Actually if we move it the Href might contain appended file version and the condition to prevent against duplicate link tags with same href will not work.

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
dougbu: @NTaylorMullen this change doesn't remove a capability. When using `PostElement`, there's no reason to read the original element's content.

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -197,17 +197,22 @@ public class ScriptTagHelper : TagHelper
...
-            var originalContent = await context.GetChildContentAsync();
 
-            if (mode == Mode.Fallback && string.IsNullOrEmpty(SrcInclude) || mode == Mode.AppendVersion)
+            if (!string.IsNullOrEmpty(Src) && AppendVersion == true)
             {
-                // No globbing to do, just build a <script /> tag to match the original one in the source file
-                // Or just add file version to the script tag.
-                BuildScriptTag(originalContent, attributes, builder);
+                EnsureFileVersionProvider();
+                output.Attributes[SrcAttributeName].Value = _fileVersionProvider.AddFileVersionToPath(Src);
```
dougbu: As in `LinkTagHelper`:
``` c#
if (AppendVersion == true)
  {
    EnsureFileVersionProvider();
    var attributeStringValue = output.Attributes[SrcAttributeName].Value as string;
    if (attributeStringValue != null)
    {
      output.Attributes[SrcAttributeName].Value = _fileVersionProvider.AddFileVersionToPath(attributeStringValue);
    }
  }
```

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -341,6 +351,8 @@ private void BuildLinkTag(TagHelperAttributeList attributes, TagHelperContent bu
                 if (AppendVersion == true &&
                     string.Equals(attribute.Name, HrefAttributeName, StringComparison.OrdinalIgnoreCase))
                 {
+                    EnsureFileVersionProvider();
```
dougbu: I'm hoping this can be removed entirely. But definitely should not be moved inside a loop.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/LinkTagHelperTest.cs
dougbu: Than make that case a separate test. Concerned we're not consistently checking what the tag helper does and does not do in all cases.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/LinkTagHelperTest.cs
dougbu: Again, need more tests then.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
ajaybhargavb: In this case a slight change is needed since Href can be null. 
`output.Attributes[HrefAttributeName].Value as string` =>
`output.Attributes[HrefAttributeName]?.Value as string`.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
dougbu: I followed up in comments that are still visible. But FYI what you want to avoid is trouncing changes `OtherLinkTagHelper` has made already to `output.Attributes[HrefAttributeName]`. The problematic "side effect" would be that tag helper not working or not doing everything it's supposed to do.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
NTaylorMullen: Ah, it seems the `BuildGlobbedLinkTags` replaces the value of `href`, therefore the functional impact wont exist. Only benefit of not moving the `attributes` then is so that it reduces the number of attribute copies that occur when we duplicate the `output.Attributes` list.

src/Microsoft.AspNet.Mvc.TagHelpers/ScriptTagHelper.cs
NTaylorMullen: :+1:

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
dougbu: But @ajaybhargavb and @NTaylorMullen you were correct and I was wrong. Don't move the creation of that `attributes` collection below where `output.Attributes[HrefAttributeName]` may be changed.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
dougbu: @ajaybhargavb good point.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
dougbu: Note we're talking about a logical issue -- we want the original collection -- and not a major functional difference. We will confuse ourselves but not immediately introduce bugs by moving the assignment to `attributes` down.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/LinkTagHelperTest.cs
ajaybhargavb: I might have misunderstood but I thought that the whole point of the test was to check if the tag helper runs beyond the check for required attributes and not to test the entire behavior including all special cases.

dougbu: Suspect "Runs" in the names of these tag helpers isn't helping us here. Instead think of this set of tests as verifying the tag helper behaviour in positive scenarios. So we want to check what the tag helper did, not just that it did something.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/LinkTagHelperTest.cs
```diff
@@ -191,78 +200,100 @@ public static TheoryData RunsWhenRequiredAttributesArePresent_Data
...
-                    // File Version
-                    {
-                        new TagHelperAttributeList
-                        {
-                            ["asp-append-version"] = "true"
-                        },
-                        tagHelper =>
-                        {
                             tagHelper.AppendVersion = true;
                         }
```
dougbu: Add a new test that just sets `AppendVersion`. `HrefInclude` is now set in every case that includes `AppendVersion`.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/LinkTagHelperTest.cs
dougbu: :ok_hand:

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ScriptTagHelperTest.cs
```diff
@@ -288,8 +326,8 @@ public static TheoryData RunsWhenRequiredAttributesArePresent_Data
             await helper.ProcessAsync(context, output);
 
             // Assert
-            Assert.Null(output.TagName);
```
dougbu: As in the `<link>` tag helper tests, please split the cases up enough that `output.TagName` can be tested.

One option would be to pass in the expected tag name i.e. add it to the `TheoryData`. That might make things easier.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ScriptTagHelperTest.cs
```diff
@@ -88,28 +85,6 @@ public static TheoryData RunsWhenRequiredAttributesArePresent_Data
                     {
                         new TagHelperAttributeList
                         {
-                            ["asp-src-include"] = "*.js"
-                        },
-                        tagHelper =>
-                        {
-                            tagHelper.SrcInclude = "*.js";
-                        }
```
dougbu: This removal is fine because `RenderScriptTags_WithFileVersion()` covers the case.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ScriptTagHelperTest.cs
ajaybhargavb: Actually the split for `NoSrc` is same as the split for `output.TagName`. I am adding back these asserts.

dougbu:
:watch:

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ScriptTagHelperTest.cs
dougbu: Big party: :balloon: :bouquet: :dancers: :bouquet: :balloon:

test/Microsoft.AspNet.Mvc.TagHelpers.Test/LinkTagHelperTest.cs
ajaybhargavb: That case is covered in this test `RendersLinkTags_AddsFileVersion`.

ajaybhargavb:
Updated.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/LinkTagHelperTest.cs
dougbu: How? The tag helper in that test method also has `HrefInclude` set.

ajaybhargavb: It has `href` set. I don't see `HrefInclude`. Could you please point me to the line where it is being set?

dougbu: https://github.com/aspnet/Mvc/blob/post-element/test/Microsoft.AspNet.Mvc.TagHelpers.Test/LinkTagHelperTest.cs#L645

ajaybhargavb: Ah. Actually this shouldn't be there. It makes no difference because there is no corresponding mock for `BuildUrlList`. That case is tested here `RendersLinkTags_GlobbedHref_AddsFileVersion`. Will remove this line.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/LinkTagHelperTest.cs
dougbu: Shouldn't this line have come back in this iteration?

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ScriptTagHelperTest.cs
```diff
@@ -413,8 +455,9 @@ public static TheoryData DoesNotRunWhenARequiredAttributeIsMissing_Data
             await helper.ProcessAsync(tagHelperContext, output);
 
             // Assert
-            Assert.Equal("script", output.TagName);
```
dougbu: Restore this line

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ScriptTagHelperTest.cs
```diff
@@ -453,8 +496,9 @@ public static TheoryData DoesNotRunWhenARequiredAttributeIsMissing_Data
...
```
dougbu: Restore this line

dougbu:
:shipit: after restoring a few `TagName` assertions

NTaylorMullen:
:shipit:

PR #2832, opened 2015-07-19
packExclue > publishEclude and removed xproj and kproj exludes

tugberkugurlu:
- removed `packExclude` from unnecessary project where the publish is not necessary.
 - renamed `packExclude` > `publishExclude`.
 - removed xproj and kproj excludes as xproj is being excluded by default and kproj has been renamed to xproj.

### DIFF
--- a/samples/MvcSample.Web/project.json
+++ b/samples/MvcSample.Web/project.json
@@ -36,10 +36,7 @@
         "node_modules",
         "bower_components"
     ],
-    "packExclude": [
-        "node_modules",
-        "bower_components",
-        "**.kproj",
+    "publishExclude": [
         "**.user",
         "**.vspscc"
     ],
--- a/samples/TagHelperSample.Web/project.json
+++ b/samples/TagHelperSample.Web/project.json
@@ -28,10 +28,7 @@
     "node_modules",
     "bower_components"
   ],
-  "packExclude": [
-    "node_modules",
-    "bower_components",
-    "**.kproj",
+  "publishExclude": [
     "**.user",
     "**.vspscc"
   ],
--- a/src/Microsoft.AspNet.Mvc.Localization/project.json
+++ b/src/Microsoft.AspNet.Mvc.Localization/project.json
@@ -20,12 +20,5 @@
     "wwwroot",
     "node_modules",
     "bower_components"
-  ],
-  "packExclude": [
-    "node_modules",
-    "bower_components",
-    "**.kproj",
-    "**.user",
-    "**.vspscc"
   ]
 }
--- a/test/WebSites/ValidationWebSite/project.json
+++ b/test/WebSites/ValidationWebSite/project.json
@@ -20,10 +20,7 @@
         "node_modules",
         "bower_components"
     ],
-    "packExclude": [
-        "node_modules",
-        "bower_components",
-        "**.kproj",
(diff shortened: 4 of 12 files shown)


dougbu:
@tugberkugurlu 

1. What affect does this change have? `packExclude` is no longer supported and `.xproj` files are excluded automatically (unless a project overrides `excludeBuiltin`, which isn't done in the MVC repo).
  - May make sense to use the new `publishExclude` patterns group and remove the `**.kproj` values completely.
  - Similarly nothing in the `exclude` patterns group need be repeated in `publishExclude`. Should therefore probably remove `node_modules` and `bower_components` as well.
2. Why is this PR limited to sample projects? 13 other projects exclude `**.kproj`.
  - It's true sample projects are more likely to be published than source projects but the other projects are also incorrect. Suggest instead removing the `packExclude` patterns group from all projects under `src` and most under `test`.
  - Projects under `test\WebSites` are also likely to be published and should probably be fixed as described in 1.

tugberkugurlu:
Makes sense. Yes, fogot about publishExclude change :smilie:

The point was that kproj is xproj now. I didn't look at other places as I was looking at samples only. I will alter the PR to accommodate those changes.

tugberkugurlu:
@dougbu updated.

samples/MvcSample.Web/project.json
```diff
@@ -36,10 +36,9 @@
         "node_modules",
         "bower_components"
     ],
-    "packExclude": [
+    "publishExclude": [
         "node_modules",
         "bower_components",
```
dougbu: Remove these two lines since they duplicate patterns in `exclude`.

samples/TagHelperSample.Web/project.json
```diff
@@ -28,10 +28,9 @@
     "node_modules",
     "bower_components"
   ],
-  "packExclude": [
+  "publishExclude": [
     "node_modules",
     "bower_components",
```
dougbu: Remove these two lines since they duplicate patterns in `exclude`.

src/Microsoft.AspNet.Mvc.Localization/project.json
```diff
@@ -15,17 +15,5 @@
   "frameworks": {
     "dnx451": {},
     "dnxcore50": {}
-  },
-  "exclude": [
-    "wwwroot",
-    "node_modules",
-    "bower_components"
-  ],
```
dougbu: Do not remove `exclude` from any project. That pattern group affects many other `dnu` commands and helps us avoid `PathTooLongException`s.

test/WebSites/ValidationWebSite/project.json
```diff
@@ -20,10 +20,9 @@
         "node_modules",
         "bower_components"
     ],
-    "packExclude": [
+    "publishExclude": [
         "node_modules",
         "bower_components",
```
dougbu: Remove these two lines since they duplicate patterns in `exclude`.

dougbu:
:watch: already much more useful!

tugberkugurlu:
What sure what the last comment means but will do the changes.

dougbu:
:shipit: I'll get this in shortly. Thanks @tugberkugurlu!

dougbu:
81f8b3e

Issue #1418, opened 2014-10-21
Dictionaries do not get model bound


pranavkm:
Consider the model
```C#
public class Foo
{
    public Dictionary<string, string> Values { get; set; }
}
```

The name generated by `Html.NameFor(v => v.Values["height"])` is `Values[height]` however this expression does not get model bound.

There's a gap in the collection model binder where it only attempts to bind numeric keys. The right fix for this would be to call into `IEnumerableValueProvider` and then bind each key and value independently and add the result to the dictionary.

yishaigalatzer:
@pranavkm did you do the comparison to MVC5?

kichalla:
Just FYI...I have tried in MVC 5 and it also didn't support this..

dougbu:
what about Web API 3?

kichalla:
Pranav and I investigated earlier(please take a look at the earlier comment in this bug) about this for 2.3.2 and it was not supported there too..

rynowak:
@kichalla can you please update the original bug to include the supported and not supported formats - include a comparison with MVC5/WebAPI2.

kichalla:
**Summary of findings:**

**For the following model and action:**
**Model:**
```c#
public class BugInfo
{
    public string Title { get; set; }

    public Dictionary<string, string> Properties { get; set; }
}
```
**Action:**
```c#
public ActionResult CreateBug(BugInfo bug)
```

(Note that in case of Web API, we decorate it with FromBody attribute)

**FormUrlEncoded formats support in different versions:**

**MVC4** supported following formats:
- bug.Properties[0].key=Height&bug.Properties[0].value=160
- bug.Properties[Height]=160 **(So this seems to have regressed in MVC5 and later)**

**MVC5** supported following format:
- bug.Properties[0].key=Height&bug.Properties[0].value=160

**Web API 2** supported following format(notice there is no prefix 'bug' here):
- Properties[0].key=Height&Properties[0].value=160

**ASP.NET 5** supports following format:
- bug.Properties[0].key=Height&bug.Properties[0].value=160

rynowak:
This is being moved out to beta5. We have parity with MVC5/WebAPI2, but we feel like any change here would need to be accompanied by HtmlHelper template support to actually generate fields that we can model bind.

danroth27:
We can revisit this based on customer demand.

rynowak:
[...]

PR #2833, opened 2015-07-20
Add support for model binding dictionaries from `prefix[name]=value` entries

dougbu:
- #1418
- add new fallback binding in `DictionaryModelBinder`
 - similar to MVC 5 approach but more explicit and with better key conversion support
- fix bugs in `PrefixContainer` encountered while adding new tests of #1418 scenarios
 - did not handle entries like "[key]" or "prefix.key[index]" correctly
 - refactor part of `GetKeyFromEmptyPrefix()` into `IndexOfDelimiter()`; share with `GetKeyFromNonEmptyPrefix()`
 - extend `ReadableStringCollectionValueProviderTest` to cover bracketed key segments

nits:
- remove use of "foo", "bar", and "baz" in affected test classes
- `""` -> `string.Empty`
- `vpResult` -> `result`

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
@@ -2,7 +2,10 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Framework.Internal;
 
 namespace Microsoft.AspNet.Mvc.ModelBinding
 {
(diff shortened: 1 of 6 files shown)


rynowak:
`vpResult` -> `result`

:dancers:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
```diff
@@ -14,14 +17,86 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
...
+        {
+            var result = await base.BindModelAsync(bindingContext);
+            if (result == null || !result.IsModelSet)
+            {
+                // No match for the prefix at all.
+                return result;
+            }
+
+            var model = result.Model as Dictionary<TKey, TValue>;
+            Debug.Assert(model != null);
```
rynowak: This should be a c-style cast not an `as`.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
```diff
@@ -14,14 +17,86 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
...
+            if (model.Count != 0)
+            {
+                // ICollection<KeyValuePair<TKey, TValue>> approach was successful.
+                return result;
+            }
+
+            var enumerableValueProvider = bindingContext.ValueProvider as IEnumerableValueProvider;
```
rynowak: Will this work with a filtered value provider?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
```diff
@@ -14,14 +17,86 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
...
+                var dictionaryKey = ConvertFromString(key.Key);
+                valueBindingContext.ModelName = key.Value;
+
+                var valueResult = await modelBinder.BindModelAsync(valueBindingContext);
+
+                // Always add an entry to the dictionary but validate only if binding was successful.
+                model[dictionaryKey] = ModelBindingHelper.CastOrDefault<TValue>(valueResult?.Model);
+                if (valueResult != null && valueResult.IsModelSet)
+                {
+                    validationNode.ChildNodes.Add(valueResult.ValidationNode);
```
rynowak: Should we also be validating the key?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
```diff
@@ -14,14 +17,86 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
...
 
+        /// <inheritdoc />
         protected override object CreateEmptyCollection()
         {
             return new Dictionary<TKey, TValue>();
         }
+
+        private static TKey ConvertFromString(string keyString)
+        {
+            // Use InvariantCulture to convert string since ExpressionHelper.GetExpressionText() used that culture.
```
rynowak: Not sure this is right - are you saying that when we generate a form-field with this format the key is always `ToString()`ed with invariant?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/PrefixContainer.cs
```diff
@@ -81,31 +81,34 @@ public bool ContainsPrefix([NotNull] string prefix)
...
-            var delimiterPosition = -1;
+            string key;
+            string fullName;
+            var delimiterPosition = IndexOfDelimiter(entry, 0);
 
-            if (dotPosition == -1)
+            if (delimiterPosition == 0 && entry[0] == '[')
             {
-                if (bracketPosition != -1)
+                // Handle an entry such as "[key]".
```
rynowak: :heart_eyes_cat: these comments

rynowak:
:watch: test coverage looks good, overall design seems right

src/Microsoft.AspNet.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
dougbu: By "filtered value provider" do you mean the special cases around `IBindingSourceValueProvider`? If so, that's a orthogonal concept related to choosing one or more value providers that support a particular `BindingSource`. `IEnumerableValueProvider` is about just `GetKeysFromPrefixAsync()` support, once you've got your value provider(s).

This check fails with any value provider that does not implement `IEnumerableValueProvider`. `ReadableStringCollectionValueProvider` (for form and query values) works but nothing else does. The `CompositeValueProvider` defers to the first `IEnumerableValueProvider` that returns keys but does not supplement inner providers that lack `IEnumerableValueProvider` support.

As it happens `ReadableStringCollectionValueProvider` also implements `IBindingSourceValueProvider`.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
dougbu: You mean for the extreme special case of a key type that implements `IValidatableObject`? Wasn't done in MVC 5 and I'm not sure it's worthwhile in MVC 6.

Further, in the collection of `KeyValuePair` case, MVC 5 simply skips `...Key` values that aren't instances of `TKey`. MVC 5 special-cases assignment of `null` to the key or value if either is a `ValueType` but MVC 6 does that at a lower level, in `TypeConverterModelBinder`.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
dougbu: Yup, that's what I'm saying.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
dougbu: Done

src/Microsoft.AspNet.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
rynowak: So, practically speaking then - you could hit this case if you put `[FromRoute]` on your model? It would be good to document what reasonable cases lead to this for clarity, that's my only concern.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
rynowak: I suppose, any key needs to be something that can be converted from a string - and you don't have a place to put validators other than the type, so it seems ok then.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
rynowak: :ok:

rynowak:
:shipit: from me after adding some comments about what scenarios lead to the non-IEnumerable-ValueProvider

dougbu:
79a2982

PR #2835, opened 2015-07-20
converted warningsAsErrors values to Boolean

tugberkugurlu:
Some code editors and IDEs give schema validation over project.json (e.g. VS Code) and gives warnings on some `warningsAsErrors` fileds under `compilationOptions` object inside project.json files as they had the values as `string` rather than `Boolean`. This PR converts them to Boolean values.

### DIFF
--- a/test/Microsoft.AspNet.JsonPatch.Test/project.json
+++ b/test/Microsoft.AspNet.JsonPatch.Test/project.json
@@ -1,6 +1,6 @@
 {
     "compilationOptions": {
-        "warningsAsErrors": "true"
+        "warningsAsErrors": true
     },
     "dependencies": {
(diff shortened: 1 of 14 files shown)


pranavkm:
@tugberkugurlu, we parse these files using the DataContractSerializer in our build scripts. Parsing a json file with an unquoted boolean was failing on Mono which would cause test runs to not kick off.

pranavkm:
See https://github.com/aspnet/Mvc/commit/2f00fc61214c0ddb6dfeb2c47deb4ba9fe4cbce2

tugberkugurlu:
@pranavkm ah, OK. thanks for pointing that out.

Issue #2799, opened 2015-07-09
TryUpdateModelAsync Out of Memory error


tom-cognify:
when calling TryUpdateModelAsync on most(not all) of the entitys in my domain i get an out of memory error. IIS Express seems to loop endlessly until it runs out of memory. This problem started happening in Beta 4, I just updated to Beta 5 and I'm still seeing the same behavior.  Domain is about 50 entities.

Here is an example of controller Action.

    public async Task<ActionResult> Edit(EnrollmentService updatedES)
		{
			var es = db.EnrollmentService.Find(updatedES.EnrollmentServiceId);
			if (ModelState.IsValid)
			{
				var updateResult = await TryUpdateModelAsync(es);

here is the Error:

```
2015-07-09 10:54:11.437 -07:00 Error: [Microsoft.AspNet.Diagnostics.ErrorHandlerMiddleware] An unhandled exception has occurred: Exception of type 'System.OutOfMemoryException' was thrown.
System.OutOfMemoryException: Exception of type 'System.OutOfMemoryException' was thrown.
   at System.String.Concat(String str0, String str1, String str2)
   at Microsoft.AspNet.Mvc.ModelBinding.ModelNames.CreatePropertyModelName(String prefix, String propertyName)
   at Microsoft.AspNet.Mvc.ModelBinding.Validation.DefaultObjectValidator.ExpandValidationNode(ValidationContext context, ModelExplorer modelExplorer)
   at Microsoft.AspNet.Mvc.ModelBinding.Validation.DefaultObjectValidator.ValidateChildNodes(String currentModelKey, ModelExplorer modelExplorer, ValidationContext validationContext)
   at Microsoft.AspNet.Mvc.ModelBinding.Validation.DefaultObjectValidator.ValidateNonVisitedNodeAndChildren(String modelKey, ValidationContext validationContext, IList`1 validators)
   at Microsoft.AspNet.Mvc.ModelBinding.Validation.DefaultObjectValidator.ValidateChildNodes(String currentModelKey, ModelExplorer modelExplorer, ValidationContext validationContext)
   at Microsoft.AspNet.Mvc.ModelBinding.Validation.DefaultObjectValidator.ValidateNonVisitedNodeAndChildren(String modelKey, ValidationContext validationContext, IList`1 validators)
[...]

PR #2838, opened 2015-07-21
Fix for #2799 - OOM during TryUpdateModelAsync

rynowak:
The change here is that when we create the ModelValidationNodes to just
return them rather than adding them to the tree. For a very large model,
having these extra nodes in the tree would eventually cause an OOM.

We're going to be taking a more thorough look at this code separately,
hence the quick fix.


dougbu:
:shipit:

dougbu:
@rynowak does bae442c allow you to close this PR?

rynowak:
Totes m'goats! :goat: :goat: :goat:

PR #2839, opened 2015-07-22
React to change in `HttpInfo`

dougbu:
- aspnet/Diagnostics@fea15eb


dougbu:
FYI only; will push immediately
Note this is separate from the reaction work @anurse is doing.

dougbu:
bb618ac

Issue #2825, opened 2015-07-17
Missing `Ok()` in `Controller`


Bartmax:
I think there's a missing helper in controller to return an object. or I can't find it :)

```
public virtual ObjectResult Ok(object value);
```

so
```
[HttpGet("{id}")]
public IActionResult Detail(int id)
{
    var ret = xxx.GetById(id);
    if (ret == null) return HttpNotFound($"id {id} does not exist");
    return Ok(ret);
}
```
instead of:
```
return new ObjectResult(ret);
````

dougbu:
@Bartmax this feature is available in the `Microsoft.AspNet.Mvc.WebApiCompatShim` package. Have a look at the `ApiController` class.

One alternative is the `HttpStatusCodeResult` class, which is similar to Web API's `OkResult` and available in both MVC 5 and 6.

Bartmax:
I'm not doing any compatibility stuff here, while Ok is there, I'm not doing any webapi2, so I don't think importing that would actually fit.

I found that ObjectResult (instead of HttpStatusCodeResult) being more accurate, also there's no method on base controller that returns either ObjectResult or HttpStatusCodeResult.

I think there should be a `public virtual ObjectResult Ok(object value);` in the Controller class on the framework or HttpOk like the other methods for the common scenarios.

It feels weird to have HttpBadRequest, HttpNotFound, HttpUnauthorized, Created, but not an HttpOk on base abstract class controller.

This was kind of a question/suggestion.

dougbu:
Updated subject because this is no longer a question.

Bartmax:
:shipit: hehe nice work @dougbu

dougbu:
6106375

PR #2840, opened 2015-07-22
Add `HttpOkResult`, `HttpOkObjectResult`, and `Ok()` methods in `Controller`

dougbu:
- #2825
- new class names align with existing types such as `HttpNotFoundResult` and `HttpNotFoundObjectResult`
- remove similar types from WebApiCompatShim and use replacements in `ApiController`
 - `NegotiatedContentResult<T>` remains because Core doesn't have an exact replacement

nits:
- add missing periods in some `Controller` doc comments


src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
```diff
@@ -424,6 +424,33 @@ public virtual JsonResult Json(object data, [NotNull] JsonSerializerSettings ser
         }
 
         /// <summary>
+        /// Creates a <see cref="HttpOkResult"/> object that produces an empty OK (200) response.
+        /// </summary>
+        /// <returns>The created <see cref="HttpOkResult"/> for the response.</returns>
+        [NonAction]
+        public virtual HttpOkResult Ok()
```
rynowak: This should be HttpOk to be like the others.

dougbu: Others? You mean like `Created()`, `Redirect()` and `RedirectPermanent()`? I agree we've also got `HttpBadRequest()`, `HttpNotFound()` and `HttpUnauthorized()` but why is that method naming convention preferred?

Eilon: What was the name in Web API 2.x? Can we keep that name?

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
```diff
@@ -424,6 +424,33 @@ public virtual JsonResult Json(object data, [NotNull] JsonSerializerSettings ser
...
+        /// </summary>
+        /// <param name="value">The content value to format in the entity body.</param>
+        /// <returns>The created <see cref="HttpOkObjectResult"/> for the response.</returns>
+        [NonAction]
+        public virtual HttpOkObjectResult Ok(object value)
+        {
+            var disposableValue = value as IDisposable;
+            if (disposableValue != null)
+            {
+                Response.RegisterForDispose(disposableValue);
```
Eilon: Is this extra logic here because the previous WebApiCompatShim code had something similar?

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
rynowak: That would be `Ok`. And also :ok: with me. I realize that we've got a hodge-podge here and can't really fix it for compat reasons.

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
rynowak: All of our `ObjectResult`-returning methods do this. There's no real other good place to put it because a filter could present your result from running.

rynowak:
:shipit: from me

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
dougbu: No real _MVC_ compat issue for this particular method; it existed only in Web API 2.x. There, the method was named `Ok()` and the results `OkResult` and `OkNegotiatedContentResult<T>`. Doesn't seem worthwhile to maintain the generic nature of the second class nor the class names.

So I also lean toward sticking with the names I've got here.

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
dougbu: @Eilon what @rynowak said :smiley_cat:

dougbu:
@Eilon are you fine with the names as-is?

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
Eilon: Yeah don't care about class names. Do care about compat mostly with MVC, but secondarily about Web API. So because MVC didn't have this one way or another, let's keep it with the Web API name, so `Ok` is :ok: per Ryan.

Eilon:
Yup :shipit:

dougbu:
6106375

PR #2848, opened 2015-07-22
Added repository information to Localization project

kichalla:
To enable `dnu sources` work this package

@kirthik

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Localization/project.json
+++ b/src/Microsoft.AspNet.Mvc.Localization/project.json
@@ -1,24 +1,34 @@
 ﻿{
-  "description": "Features that enable globalization & localization of MVC applications.",
-  "version": "6.0.0-*",
-  "compilationOptions": {
-    "warningsAsErrors": false
-  },
-  "dependencies": {
-    "Microsoft.AspNet.Localization": "1.0.0-*",
-    "Microsoft.AspNet.Mvc.Razor": "6.0.0-*",
-    "Microsoft.Framework.DependencyInjection": "1.0.0-*",
-    "Microsoft.Framework.Localization": "1.0.0-*",
-    "Microsoft.Framework.NotNullAttribute.Sources": { "version": "1.0.0-*", "type": "build" },
-    "Microsoft.Framework.PropertyHelper.Sources": { "version": "1.0.0-*", "type": "build" }
-  },
-  "frameworks": {
-    "dnx451": {},
-    "dnxcore50": {}
-  },
-  "exclude": [
-    "wwwroot",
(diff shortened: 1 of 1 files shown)


kirthik:
:shipit:

src/Microsoft.AspNet.Mvc.Localization/project.json
```diff
@@ -16,9 +16,13 @@
...
-  ]
+    "exclude": [
+        "wwwroot",
+        "node_modules",
+        "bower_components"
+    ],
+    "repository": {
+        "type": "git",
+        "url": "https://github.com/aspnet/mvc"
+    }
```
dougbu: Inconsistent formatting. Stick to 2 spaces per logical tab.

kichalla: @dougbu updated (I used the default settings of my VS RTM which is set to use 4 spaces per tab)

src/Microsoft.AspNet.Mvc.Localization/project.json
```diff
@@ -1,24 +1,34 @@
...
+    "version": "6.0.0-*",
+    "compilationOptions": {
+        "warningsAsErrors": false
+    },
+    "dependencies": {
+        "Microsoft.AspNet.Localization": "1.0.0-*",
+        "Microsoft.AspNet.Mvc.Razor": "6.0.0-*",
+        "Microsoft.Framework.DependencyInjection": "1.0.0-*",
+        "Microsoft.Framework.Localization": "1.0.0-*",
+        "Microsoft.Framework.NotNullAttribute.Sources": {
```
Eilon: Don't care about 4 vs. 2 spaces, but we do but the dependency sub-settings on the same line... make it easier to read.

src/Microsoft.AspNet.Mvc.Localization/project.json
dougbu: That's fine and aligns w/ @Eilon's comment. My concern was the inconsistent indentation seen in a previous iteration.

src/Microsoft.AspNet.Mvc.Localization/project.json
dougbu: @Eilon so basically we need to remember to add `?w=1` to the URL when reviewing changes done by developers with different VS settings? Or will we eventually standardize on one setting or the other e.g. to match the templates VS ships with?

Eilon: Discussed in person with a few people and we should just use the VS formatting instead of fighting it.

PR #2854, opened 2015-07-23
RequireHttpsAttribute - Returning a 405 Method Not Allowed Response

RehanSaeed:
Returning a 405 Method Not Allowed response is the more appropriate status code when you only want to allow GET requests.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/RequireHttpsAttribute.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/RequireHttpsAttribute.cs
@@ -27,7 +27,7 @@ namespace Microsoft.AspNet.Mvc
             // body correctly.
             if (!string.Equals(filterContext.HttpContext.Request.Method, "GET", StringComparison.OrdinalIgnoreCase))
             {
-                filterContext.Result = new HttpStatusCodeResult(StatusCodes.Status403Forbidden);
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Mvc.Core/RequireHttpsAttribute.cs
```diff
@@ -27,7 +27,7 @@ protected virtual void HandleNonHttpsRequest(AuthorizationContext filterContext)
             // body correctly.
             if (!string.Equals(filterContext.HttpContext.Request.Method, "GET", StringComparison.OrdinalIgnoreCase))
             {
-                filterContext.Result = new HttpStatusCodeResult(StatusCodes.Status403Forbidden);
+                filterContext.Result = new HttpStatusCodeResult(StatusCodes.Status405MethodNotAllowed);
```
Eilon: Choosing between HTTP 4xx status codes is tricky. Is the problem the protocol, the method, the URL, the body, a header?

In this particular case my impression is that the issue isn't about using the wrong HTTP *method*. It's that the wrong *protocol* is being used to access the resource. That is, for someone seeing this error the fix isn't to change from POST to GET, it's to change from HTTP to HTTPS.

RehanSaeed: I see your point. The correct scheme is required before selecting the correct method.

PR #2864, opened 2015-07-27
Implementation of the new "Add" operation for JsonPatch

KevinDockx:
Most important changes:

- Added ObjectTreeAnalysisResult: this class analyzes the object tree and results in an object that's used by the Add method to know if it should use "dynamic logic" (ie: add to the Dictionary) or "regular logic" (staticly typed).  JsonContract checks moved to this class instead of in the Add method
- Changes in the Add method: split-up between regular & dynamic
- Addition of GetNumericEndResult: used in Add to act differently when a numeric end is found vs when it's found but is invalid

Note that this is just the Add operation.  Once this is approved & suggestions/comments are incorporated, the other methods will be changed as well - and that's when I can clean up the code a bit more (for example: GetNumericEnd can be removed, GetNumericEndResult can be renamed, helpers to find properties with JsonContract can be removed as they now are in ObjectTreeAnalyisResult, ...)

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Localization/project.json
+++ b/src/Microsoft.AspNet.Mvc.Localization/project.json
@@ -1,24 +1,34 @@
 ﻿{
-  "description": "Features that enable globalization & localization of MVC applications.",
-  "version": "6.0.0-*",
-  "compilationOptions": {
-    "warningsAsErrors": false
-  },
-  "dependencies": {
-    "Microsoft.AspNet.Localization": "1.0.0-*",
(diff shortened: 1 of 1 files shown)


KevinDockx:
Just noticed the previous commits are still there - only the last one ("Implementation of new Add operation") contains changes cfr aspnet\dev, all the previous ones are already in aspnet\dev.  But: if you want me to rebase, let me know :)

kirthik:
@KevinDockx : Can you please rebase :) ?

Issue #2769, opened 2015-07-01
Change Razor attribute behavior to not special case boolean values in middle of complex attribute values.


NTaylorMullen:
Today if a user writes: 

```
@{
    var isActive = true;
}
<input type="button" onclick="doSomething(@isActive);" />
```

It generates `<input type="button" onclick="doSomething(onclick);" />`. If `isActive` is `false` it removes the content entirely. This feature had good intentions in MVC5 but should be scoped to non-complex attribute values; aka values that only have a single value.

Intended output:

```
<input checked="@true" />                 => <input checked="checked" />
<input checked="@false" />                => <input />
<input checked="   @true         " />     => <input checked="   True         " />
<input checked="   @false        " />     => <input checked="   False        " />
<input checked="value: @true" />          => <input checked="value: True" />
<input checked="value: @false" />         => <input checked="value: False" />
```

PR #2868, opened 2015-07-27
Razor boolean attribute special case handled correctly

ajaybhargavb:
Issue - #2769 
@NTaylorMullen @dougbu

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
@@ -519,96 +519,62 @@ namespace Microsoft.AspNet.Mvc.Razor
             [NotNull] PositionTagged<string> suffix,
             params AttributeValue[] values)
         {
-            var first = true;
-            var wroteSomething = false;
             if (values.Length == 0)
             {
-                // Explicitly empty attribute, so write the prefix and suffix
+                // Explicitly empty attribute, so write the prefix
                 WritePositionTaggedLiteral(writer, prefix);
-                WritePositionTaggedLiteral(writer, suffix);
+            }
+            else if (values.Length == 1 && values[0].Prefix == "" &&
+                (values[0].Value.Value is bool || values[0].Value.Value == null))
+            {
+                // Value is either null or a bool with no prefix.
+                var attributeValue = values[0];
+                var positionTaggedAttributeValue = attributeValue.Value;
+
+                if (positionTaggedAttributeValue.Value == null || !(bool)positionTaggedAttributeValue.Value)
+                {
+                    // The value is null or just the bool 'false', don't write anything.
+                    return;
+                }
+
+                WritePositionTaggedLiteral(writer, prefix);
+
+                var sourceLength = suffix.Position - positionTaggedAttributeValue.Position;
+
+                // The value is just the bool 'true', write the attribute name instead of the string 'True'.
+                WriteAttributeValue(writer, attributeValue, name, sourceLength);
             }
             else
             {
+                // This block handles two cases.
+                // 1. Single value with prefix.
+                // 2. Multiple values with or without prefix.
+                WritePositionTaggedLiteral(writer, prefix);
                 for (var i = 0; i < values.Length; i++)
                 {
-                    var attrVal = values[i];
-                    var val = attrVal.Value;
-                    var next = i == values.Length - 1 ?
-                        suffix : // End of the list, grab the suffix
-                        values[i + 1].Prefix; // Still in the list, grab the next prefix
+                    var attributeValue = values[i];
+                    var positionTaggedAttributeValue = attributeValue.Value;
 
-                    if (val.Value == null)
+                    if (positionTaggedAttributeValue.Value == null)
                     {
                         // Nothing to write
                         continue;
                     }
 
-                    // The special cases here are that the value we're writing might already be a string, or that the
-                    // value might be a bool. If the value is the bool 'true' we want to write the attribute name
-                    // instead of the string 'true'. If the value is the bool 'false' we don't want to write anything.
-                    // Otherwise the value is another object (perhaps an HtmlString) and we'll ask it to format itself.
-                    string stringValue;
-
-                    // Intentionally using is+cast here for performance reasons. This is more performant than as+bool?
-                    // because of boxing.
-                    if (val.Value is bool)
-                    {
-                        if ((bool)val.Value)
-                        {
-                            stringValue = name;
-                        }
-                        else
-                        {
-                            continue;
-                        }
-                    }
-                    else
-                    {
-                        stringValue = val.Value as string;
-                    }
-
-                    if (first)
-                    {
-                        WritePositionTaggedLiteral(writer, prefix);
-                        first = false;
-                    }
-
-                    if (!string.IsNullOrEmpty(attrVal.Prefix))
-                    {
-                        WritePositionTaggedLiteral(writer, attrVal.Prefix);
-                    }
+                    var next = i == values.Length - 1 ?
+                        suffix : // End of the list, grab the suffix
+                        values[i + 1].Prefix; // Still in the list, grab the next prefix
 
                     // Calculate length of the source span by the position of the next value (or suffix)
-                    var sourceLength = next.Position - attrVal.Value.Position;
+                    var sourceLength = next.Position - attributeValue.Value.Position;
 
-                    BeginContext(attrVal.Value.Position, sourceLength, isLiteral: attrVal.Literal);
-                    // The extra branching here is to ensure that we call the Write*To(string) overload where
-                    // possible.
-                    if (attrVal.Literal && stringValue != null)
-                    {
-                        WriteLiteralTo(writer, stringValue);
-                    }
-                    else if (attrVal.Literal)
-                    {
-                        WriteLiteralTo(writer, val.Value);
-                    }
-                    else if (stringValue != null)
-                    {
-                        WriteTo(writer, stringValue);
-                    }
-                    else
-                    {
-                        WriteTo(writer, val.Value);
-                    }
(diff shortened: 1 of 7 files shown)


dougbu:
@ajaybhargavb is this the issue or PR you and @NTaylorMullen discussed a lot today? If yes, does it currently match any decisions from those discussions? Basically is it too early to review?

ajaybhargavb:
@dougbu, yes. Don't review now. I'll send an update.

ajaybhargavb:
Updated. @dougbu @NTaylorMullen

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -529,85 +527,52 @@ public virtual void WriteLiteralTo([NotNull] TextWriter writer, string value)
...
-                        WritePositionTaggedLiteral(writer, prefix);
-                        first = false;
-                    }
+                        string stringValue = null;
 
-                    if (!string.IsNullOrEmpty(attrVal.Prefix))
-                    {
-                        WritePositionTaggedLiteral(writer, attrVal.Prefix);
-                    }
+                        stringValue = positionTaggedAttributeValue.Value as string;
```
dougbu: combine w/ previous line: `var stringValue = positionTaggedAttributeValue.Value as string;`

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -529,85 +527,52 @@ public virtual void WriteLiteralTo([NotNull] TextWriter writer, string value)
...
 
-                    // Intentionally using is+cast here for performance reasons. This is more performant than as+bool?
-                    // because of boxing.
-                    if (val.Value is bool)
+                    // The value is just the bool 'true', write the attribute name instead of the string 'true'.
+                    WriteAttributeValue(writer, attributeValue, name, suffix.Position);
+                }
+                else
+                {
+                    WritePositionTaggedLiteral(writer, prefix);
```
dougbu: Shouldn't `prefix` be written out only if the first value is not `@null` or equivalent?

ajaybhargavb: The decision was to not write prefix only if the value contains `@false` or `@null` with no surrounding whitespace.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -621,6 +586,42 @@ public virtual string Href([NotNull] string contentPath)
...
+        {
+            var positionTaggedAttributeValue = attributeValue.Value;
+
+            if (!string.IsNullOrEmpty(attributeValue.Prefix))
+            {
+                WritePositionTaggedLiteral(writer, attributeValue.Prefix);
+            }
+
+            BeginContext(attributeValue.Value.Position, sourceLength, isLiteral: attributeValue.Literal);
+            // The extra branching here is to ensure that we call the Write*To(string) overload where
```
dougbu: Add a blank line above to make this grouping even more clear.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -529,85 +527,52 @@ public virtual void WriteLiteralTo([NotNull] TextWriter writer, string value)
             }
             else
             {
-                for (var i = 0; i < values.Length; i++)
```
dougbu: Suggest a comment here and in the commit description that informs reader of the two rules now enforced / implemented in this large block i.e. `null` / `bool` handling when there's just one value and `null` plus prefix handling in all other cases.

dougbu:
:watch: mostly for description wording

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -529,85 +527,52 @@ public virtual void WriteLiteralTo([NotNull] TextWriter writer, string value)
...
+                if (values.Length == 1 && values[0].Prefix == "" &&
```
NTaylorMullen:
```C#
var attributeValue = values.SingleOrDefault();
if (attributeValue?.Prefix == "" &&
    (attributeValue.Value.Value is bool || attributeValue.Value.Value == null))
{
    var positionTaggedAttributeValue = attributeValue.Value;
```

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -529,85 +527,52 @@ public virtual void WriteLiteralTo([NotNull] TextWriter writer, string value)
...
+                    for (var i = 0; i < values.Length; i++)
                     {
-                        if ((bool)val.Value)
-                        {
-                            stringValue = name;
-                        }
-                        else
+                        var attributeValue = values[i];
+                        var positionTaggedAttributeValue = attributeValue.Value;
+                        var next = i == values.Length - 1 ?
```
NTaylorMullen: Move down to where it's used

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -529,85 +527,52 @@ public virtual void WriteLiteralTo([NotNull] TextWriter writer, string value)
...
-                    // value might be a bool. If the value is the bool 'true' we want to write the attribute name
-                    // instead of the string 'true'. If the value is the bool 'false' we don't want to write anything.
-                    // Otherwise the value is another object (perhaps an HtmlString) and we'll ask it to format itself.
-                    string stringValue;
+                    WritePositionTaggedLiteral(writer, prefix);
 
-                    // Intentionally using is+cast here for performance reasons. This is more performant than as+bool?
-                    // because of boxing.
-                    if (val.Value is bool)
+                    // The value is just the bool 'true', write the attribute name instead of the string 'true'.
```
NTaylorMullen: `Instead of the string 'True'`

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -529,85 +527,52 @@ public virtual void WriteLiteralTo([NotNull] TextWriter writer, string value)
...
+                    WriteAttributeValue(writer, attributeValue, name, suffix.Position);
```
NTaylorMullen: The `SourceLength` is wrong here. It should be `suffix.Position - attributeValue.Value.Position`.

Should add unit tests to validate we're beginning contexts with the correct info.

test/Microsoft.AspNet.Mvc.FunctionalTests/HtmlGenerationTest.cs
```diff
@@ -575,5 +575,22 @@ public class HtmlGenerationTest
...
+            var expectedContent =
+                await ResourceFile.ReadResourceAsync(_resourcesAssembly, outputFile, sourceFile: false);
+
+            // Act
+            var response = await client.GetStringAsync("http://localhost/HtmlGeneration_Home/AttributesWithBooleanValues");
+
+            // Assert
+            Assert.Equal(expectedContent, response);
+        }
     }
```
NTaylorMullen: Should add a test variation that encodes the same file with the `TestHtmlEncoder`. Will make sure the new `WriteAttributeValue` method encodes correctly in the new cases.

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
dougbu: @NTaylorMullen `SingleOrDefault()` will `throw` if there are more than one elements. Definitely not what we want here. (Further @ajaybhargavb's code is pretty clear and we've got an array so it's already efficient.)

test/Microsoft.AspNet.Mvc.FunctionalTests/HtmlGenerationTest.cs
dougbu: +1 it's rather scary that no existing tests covered the complex rules we're removing here. Should have some / all of the variations added to this functional test covered in the `RazorPageTest` class.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
NTaylorMullen: Dur dur dur. I always mess up singleordefault :+1:

NTaylorMullen: Can combine this if as an else if above then

ajaybhargavb:
Updated.

test/Microsoft.AspNet.Mvc.FunctionalTests/HtmlGenerationTest.cs
```diff
@@ -110,13 +112,14 @@ public class HtmlGenerationTest
...
-        [InlineData("OrderUsingHtmlHelpers", "/HtmlGeneration_Order/Submit")]
-        [InlineData("Product", null)]
-        [InlineData("Script", null)]
+        //[InlineData("EditWarehouse", null)]
+        //[InlineData("Index", null)]
+        //[InlineData("Link", null)]
+        //[InlineData("Order", "/HtmlGeneration_Order/Submit")]
+        //[InlineData("OrderUsingHtmlHelpers", "/HtmlGeneration_Order/Submit")]
+        //[InlineData("Product", null)]
+        //[InlineData("Script", null)]
```
dougbu: Remove, don't comment out code

ajaybhargavb: Oops. Forgot to uncomment them.

NTaylorMullen:
:shipit: after uncomments.

dougbu:
:shipit:

PR #2870, opened 2015-07-28
Add operation JsonPatch => dynamic support

KevinDockx:
(after rebase & including fix for NotNull on path as per Kirthi's comments)

Most important changes:
- Added ObjectTreeAnalysisResult: this class analyzes the object tree and results in an object that's used by the Add method to know if it should use "dynamic logic" (ie: add to the Dictionary) or "regular logic" (staticly typed). JsonContract checks moved to this class instead of in the Add method
- Changes in the Add method: split-up between regular & dynamic
- Addition of GetNumericEndResult: used in Add to act differently when a numeric end is found vs when it's found but is invalid

Note that this is just the Add operation. Once this is approved & suggestions/comments are incorporated, the other methods will be changed as well - and that's when I can clean up the code a bit more (for example: GetNumericEnd can be removed, GetNumericEndResult can be renamed, helpers to find properties with JsonContract can be removed as they now are in ObjectTreeAnalyisResult, ...)


src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -109,19 +109,12 @@ public void Add([NotNull] Operation operation, [NotNull] object objectToApplyTo)
         /// Add is used by various operations (eg: add, copy, ...), yet through different operations;
         /// This method allows code reuse yet reporting the correct operation on error
         /// </summary>
-        private void Add(
-            [NotNull] string path,
-            object value,
-            [NotNull] object objectToApplyTo,
-            [NotNull] Operation operationToReport)
+        private void Add([NotNull] string path, object value, [NotNull] object objectToApplyTo, [NotNull] Operation operationToReport)
```
rynowak: formatting change makes this harder to review :frowning: - we use a 120 char line length because that's what works in github's UI.

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -133,91 +126,242 @@ public void Add([NotNull] Operation operation, [NotNull] object objectToApplyTo)
...
-                if (positionAsInteger > -1)
+                if (checkNumericEndResult.HasNumericEnd)
                 {
-                    actualPathToProperty = path.Substring(0,
-                        path.LastIndexOf('/' + positionAsInteger.ToString()));
+                    positionAsInteger = checkNumericEndResult.NumericEnd;
+                    if (positionAsInteger > -1)
+                    {
+                        actualPathToProperty = path.Substring(0,
+                       path.LastIndexOf('/' + positionAsInteger.ToString()));
```
rynowak: indent is wonky

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -133,91 +126,242 @@ public void Add([NotNull] Operation operation, [NotNull] object objectToApplyTo)
             }
             else
             {
-                positionAsInteger = GetNumericEnd(path);
+                var checkNumericEndResult = GetNumericEndResult(path);
```
rynowak: My thought about this... this code could be simplified by doing one of two things:
1. Return an `int?` here - the extra type you created doesn't really do anything that `int?` doesn't.
2. Instead of 1 - do the 'split' and return that if it's a numeric ending-path. Basically roll all of the code in the 'if' into the `GetNumericEndResult` method

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
rynowak: Put the `0` and the `path.LastIndexOf(...)` each on their own line so we can see that they are parameters.

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -133,91 +126,242 @@ public void Add([NotNull] Operation operation, [NotNull] object objectToApplyTo)
...
+                        actualPathToProperty = path.Substring(0,
+                       path.LastIndexOf('/' + positionAsInteger.ToString()));
+                    }
+                    else
+                    {
+                        // negative position - invalid path
+                        LogError(new JsonPatchError(
+                                    objectToApplyTo,
+                                    operationToReport,
+                                    Resources.FormatPropertyDoesNotExist(path)));
```
rynowak: Is this really the right error?

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -133,91 +126,242 @@ public void Add([NotNull] Operation operation, [NotNull] object objectToApplyTo)
...
+                                    else
+                                    {
+                                        LogError(new JsonPatchError(
+                                            objectToApplyTo,
+                                            operationToReport,
+                                            Resources.FormatInvalidIndexForArrayProperty(operationToReport.op, path)));
+                                    }
+                                }
+                            }
+                            else
```
rynowak: Some of this is hard to read because there's so much of a gap between the `if` and the `else`. It might benefit from being rewritten in the style of:

```
if (error_condition)
{
    <report error>
    return;
}

// Do more
```

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -133,91 +126,242 @@ public void Add([NotNull] Operation operation, [NotNull] object objectToApplyTo)
...
-                var conversionResultTuple = ConvertToActualType(
-                    patchProperty.Property.PropertyType,
-                    value);
+                if (!treeAnalysisResult.IsValidPathForAdd)
+                {
+                    LogError(new JsonPatchError(
+                                     objectToApplyTo,
+                                     operationToReport,
+                                     Resources.FormatPropertyDoesNotExist(path)));
+                }
```
rynowak: fix indent

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -133,91 +126,242 @@ public void Add([NotNull] Operation operation, [NotNull] object objectToApplyTo)
...
+
+                // If it' an array, add to that array.  If it's not, we replace.
+
+                // is the path an array (but not a string (= char[]))?  In this case,
```
rynowak: what's the issue of `string` <-> `char[]`? It might be helpful to mention what CLR types will cause this case as well as what JSON.

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -722,5 +866,17 @@ private int GetNumericEnd(string path)
 
             return -1;
         }
+
+        private CheckNumericEndResult GetNumericEndResult(string path)
+        {
+            var possibleIndex = path.Substring(path.LastIndexOf("/") + 1);
```
rynowak: Yeah, this is what I expected when I looked at the other code... You'll later use the same logic to split a path like `/foo/bar/5` into `/foo/bar` and `5`. You may as well do it here and make it part of the result.

src/Microsoft.AspNet.JsonPatch/Helpers/ExpandoObjectDictionaryExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNet.JsonPatch.Helpers
+{
+    // Helper methods to allow case-insensitive key search
+    internal static class ExpandoObjectDictionaryExtensions
+    {
+        internal static void SetValueForCaseInsensitiveKey(this IDictionary<string, object> propertyDictionary,
```
rynowak: Who creates these dictionaries? Wondering if it's possible to just use a case-insensitive comparer or a thing wrapper.

src/Microsoft.AspNet.JsonPatch/Helpers/ExpandoObjectDictionaryExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+                    realKey = kvp.Key;
+                    break;
+                }
+            }
+
+            if (realKey != "")
+            {
+                propertyDictionary.Remove(realKey);
+            }
+        }
```
rynowak: What should we do if the dictionary has the same key with multiple casings? Would this ever happen?

src/Microsoft.AspNet.JsonPatch/Helpers/ExpandoObjectDictionaryExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+                    propertyDictionary[kvp.Key] = value;
+                    break;
+                }
+            }
+        }
+
+        internal static void RemoveValueForCaseInsensitiveKey(this IDictionary<string, object> propertyDictionary,
+        string key)
+        {
+            string realKey = "";
```
rynowak: Use `null` if you want a sentinel value. `null` isn't legal as a key in `Dictionary` and `""` is.

src/Microsoft.AspNet.JsonPatch/Helpers/ObjectTreeAnalyisResult.cs
```diff
@@ -0,0 +1,211 @@
...
+            // empty item (that's the case when it starts with a "/")
+
+            var propertyPathTree = propertyPath.Split('/').ToList();
+            object targetObject = objectToSearch;
+
+            if (string.IsNullOrWhiteSpace(propertyPathTree[0]))
+            {
+                // remove it
+                propertyPathTree.RemoveAt(0);
+            }
```
rynowak: use `string.Split(..., StringSplitOptions.IgnoreEmptyEntries)` and you don't have to do this :+1:

src/Microsoft.AspNet.JsonPatch/Helpers/ObjectTreeAnalyisResult.cs
```diff
@@ -0,0 +1,211 @@
...
+                    // find the value in the dictionary                   
+                    if ((targetObject as IDictionary<string, object>)
+                        .ContainsCaseInsensitiveKey(propertyPathTree[i]))
+                    {
+                        var possibleNewTargetObject = (targetObject as IDictionary<string, object>)
+                       .GetValueForCaseInsensitiveKey(propertyPathTree[i]);
+
+                        // unless we're at the last item, we should set the targetobject
+                        // to the new object.  If we're at the last item, we need to stop
+                        if (!(i == propertyPathTree.Count - 1))
```
rynowak: use `!=`

src/Microsoft.AspNet.JsonPatch/Helpers/ObjectTreeAnalyisResult.cs
```diff
@@ -0,0 +1,211 @@
...
+                // remove it
+                propertyPathTree.RemoveAt(0);
+            }
+
+            // we've now got a split up property tree "base/property/otherproperty/..."
+            int lastPosition = 0;
+            for (int i = 0; i < propertyPathTree.Count; i++)
+            {
+                // if the current target object is an ExpandoObject (IDictionary<string, object>),
+                // we cannot use the ContractResolver.
```
rynowak: I would think that it would return `JsonDynamicContract` here, which should do the right thing wrt property name lookups, and allow you to add a property if it doesn't exist. Is that not the case?

rynowak:
I think I get what's going on here, still have some questions about why `JsonDynamicContract` hasn't already solved this for us.

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
KevinDockx: Ok, will change

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
KevinDockx: Ok, will change

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
KevinDockx: Well, in my original code it's "invalid path" - this came from what was already in the source code :)  Both could work - "property at path array/-1 does not exist" => element at array index can be regarded as a property when you continue working with it down the line, ie array/1/name.  But "invalid path" is a better fit imo.  I'll change it.

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
KevinDockx: Ok, will change

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
KevinDockx: Treating string as an array of char would make path "person/name" invalid - will clarify it in the comments

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
KevinDockx: Good idea, will do

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
KevinDockx: Will go for option 2 cfr comments below

src/Microsoft.AspNet.JsonPatch/Helpers/ExpandoObjectDictionaryExtensions.cs
KevinDockx: That is an issue.  Thing is: a jsonpatch doc is case insensitive by default, .NET properties aren't.  We could thrown an error if two are found?

src/Microsoft.AspNet.JsonPatch/Helpers/ExpandoObjectDictionaryExtensions.cs
KevinDockx: ;-)  tbh, when I started working on this I figured I'd be able to create it with a case-insensitive comp, but: no such luck.  Yet looking deeper into it, it makes sense - keys should be case-sensitive for ExpObj, as they relate to .NET properties - "name != Name".  

I wrote these extensions as they're used throughout the analysisresult/objectadapter; these methods keep the code in one place.  If you prefer, I could use helper methods rather than extension methods?

src/Microsoft.AspNet.JsonPatch/Helpers/ExpandoObjectDictionaryExtensions.cs
KevinDockx: Tnx for the tip, will do!

src/Microsoft.AspNet.JsonPatch/Helpers/ObjectTreeAnalyisResult.cs
KevinDockx: Great tip, tnx :)

src/Microsoft.AspNet.JsonPatch/Helpers/ObjectTreeAnalyisResult.cs
KevinDockx: I will look into this - could be a more elegant solution than what I've got now.

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
rynowak: If it says "invalid path" - that's better than "property doesn't exist" when it's a really an index out of range. maybe rename the resource so it matches?

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
rynowak: Is that because the code is looking for `IEnumerable<T>`? Just wanted clarification.

src/Microsoft.AspNet.JsonPatch/Helpers/ExpandoObjectDictionaryExtensions.cs
rynowak: I don't really feel strongly about that distinction... I checked into and there really aren't a ton of options for improving this :cry:

src/Microsoft.AspNet.JsonPatch/Helpers/ExpandoObjectDictionaryExtensions.cs
rynowak: Yeah I'm not sure I feel too strongly about it either way - CLS guidelines strongly recommend against coding that wayt.. What would happen right now in the code when a non-dynamic object has properties of different casings?

If we have the same issue for non-dynamic, I'd say log an issue and move on.

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
KevinDockx: Ok - I created a new, more specific error message for this (invalid path is used in other places)

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
KevinDockx: Yep, that's the IsNonStringArray code - if the path according to the analysis should be an array (ie: ends on /- or /number), the code checks if it indeed is; but if it's a string (char[]), an error should be thrown (not valid to add to / remove from / replace in at a specific position).

src/Microsoft.AspNet.JsonPatch/Helpers/ExpandoObjectDictionaryExtensions.cs
KevinDockx: Yup, same issue would arise for non-dynamic :( Don't really have an elegant solution for this... I'll log an issue.

src/Microsoft.AspNet.JsonPatch/Helpers/ObjectTreeAnalyisResult.cs
KevinDockx: @rynowak I looked into this - I can resolve to a JsonDictionaryContract (JsonDynamicContract will throw an exception - cannot convert JsonDictionaryContract to JsonDynamicContract), but after that, we're stuck.  Don't seem to be able to access the underlying dictionary items etc; and after all, the purpose is to simply check if the object is a dic<string,obj), and check for items in it / add to it.

But maybe I'm overlooking something obvious here, tends to happen when staring at the same piece of code for some time :) I'm always happy to improve this code, but I don't really see how?

rynowak: Ok, seems like we should leave it as-is then.

KevinDockx: Ok - I think that's it for this PR then?

src/Microsoft.AspNet.JsonPatch/Helpers/ExpandoObjectDictionaryExtensions.cs
KevinDockx: Logged issue: https://github.com/aspnet/Mvc/issues/2911

src/Microsoft.AspNet.JsonPatch/Helpers/ObjectTreeAnalyisResult.cs
rynowak: ya. Sorry, I've been a bit in and out of the office lately. Let me do one more quick pass.

rynowak:
:shipit: - looking pretty good. I'm not in the office today, but I'm going to do a bit of style cleanup later and then merge it.

src/Microsoft.AspNet.JsonPatch/Helpers/ExpandoObjectDictionaryExtensions.cs
Eilon: In .NET a CLS-compliant type cannot have members that very only by casing. That is to support languages such as VB, which are case-insensitive. As such, I think it's totally acceptable to have odd, unusual, or bizarre behavior if members vary only by case.

Eilon: BTW I closed the bug about dealing with this because I don't think there's any value in supporting it.

Eilon:
@KevinDockx sorry to bother you with CLAs, but I think that maybe you never signed the updated CLA that we started using a few months ago. You can find it here: https://cla2.dotnetfoundation.org/

This new one is even simpler than the old one. Sorry to trouble you with this!

KevinDockx:
@Eilon No problem, just signed it :-)

Eilon:
Thanks!

rynowak:
Thanks Kevin! Now I'll take care of merging this.

rynowak:
@KevinDockx - I've merged this to dev.

I ended up tweaking a bunch of the error handling to make it more readable. Please take a look and try to do the next parts in a more similar style.

Issue #2807, opened 2015-07-10
Add URL resolving TagHelper to replace Razor ~ functionality.


NTaylorMullen:
This is in reaction to the conversation: https://github.com/aspnet/Razor/issues/427#issuecomment-120494046

Today Razor special cases `~/` at the beginning of attributes and treats them as "resolve an application relative URL"; however, it does this for every attribute which is overkill/incorrect. After investigating the attributes that have URL capability in Razor the list warrants something less hard coded into the core of Razor to resolve application relative URLs.

The list of HTML5 url based attributes I was able to find are:
- href : `<a>`, `<area>`, `<link>`, `<base>`
- src : `<audio>`, `<embed>`, `<iframe>`, `<img>`, `<input>`, `<script>`, `<source>`, `<track>`, `<video>`
- action : `<form>`
- cite : `<blockquote>`, `<del>`, `<ins>`, `<q>`
- data : `<object>`
- formaction : `<button>`, `<input>`
- icon : `<menuitem>`
- itemid : everything
- manifest : `<html>`
- poster : `<video>`
- archive : `<object>`, `<applet>`

The `TagHelper` should exist in `Microsoft.AspNet.Mvc.Razor` and should be auto-added to Razor pages (much like how we auto add `@inject` directives). It should inspect `TagHelperOutput.Attributes` and `TagHelperOutput.TagName` for the various combinations listed above and resolve URLs if they begin with `~/`. Lastly, it should have an `[EditorBrowsable(EditorBrowsableState.Never)]` attribute to prevent excess highlighting in the Razor editor (will work once https://github.com/aspnet/Razor/issues/447 is completed).

Bartmax:
One question, about data attributes? I may want to do `data-bigImageSource="~/Images/big/"` In this case I should use `data-bigImageSource=@Url.Content("~/images/big/")` ?

NTaylorMullen:
Correct.

Bartmax:
:+1: thank you.

NTaylorMullen:
0ef68eefc898247fe6a3c6d6a8ab8820ea2b34e4

Issue #2841, opened 2015-07-22
Link tag helper does not expand comma-separated list


natemcmaster:
https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs#L118 says "asp-href-include" can include a comma-separated list of files and globs. Only the first file/pattern included is actually produced.

Setup:
```html
<link asp-href-include="~/test1.css,~/test2.css" rel="stylesheet" />
```
Expected:
```html
<link href="/test1.css" rel="stylesheet" />
<link href="/test2.css" rel="stylesheet" />
```
Actual:
```html
<link href="/test1.css" rel="stylesheet" />
```

Repro:
https://github.com/natemcmaster/mvc-issue-2841

ajaybhargavb:
@natemcmaster, Looks like the `~` is not being handled properly in comma separated inputs. Your example should work if you remove `~` from the second url.
```HTML
<link asp-href-include="~/test1.css,/test2.css" rel="stylesheet" />
```
I am working on fixing this.

PR #2875, opened 2015-07-28
Added default `UrlResolutionTagHelper` to resolve app relative URLs.

NTaylorMullen:
- Razor removed the ability to automatically resolve URLs prefixed with `~/`; therefore `ScriptTagHelper`, `LinkTagHelper` and `ImageTagHelper` have changed to take in `IUrlHelper`s and auto-resolve their URL based properties if they start with `~/`.
- Added a catch-all `~/` resolver for non `TagHelper` based HTML elements. Razor used to resolve any attribute value that started with `~/` now the behavior is restricted to attributes that can contain URLs.
- Updated `IUrlHelper` to accept `null` values.
- Added functional tests to validate that URLs resolve correctly.
- Updated `TagHelper` tests to ensure that URLs are resolved via an `IUrlHelper`.

#2807

Razor PR to remove existing app-relative URL resolution: https://github.com/aspnet/Razor/pull/466

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/UrlHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/UrlHelper.cs
@@ -120,8 +120,13 @@ namespace Microsoft.AspNet.Mvc
         }
 
         /// <inheritdoc />
-        public virtual string Content([NotNull] string contentPath)
+        public virtual string Content(string contentPath)
         {
+            if (contentPath == null)
+            {
+                return null;
+            }
+
             return GenerateClientUrl(_httpContext.Request.PathBase, contentPath);
         }
 
--- a/src/Microsoft.AspNet.Mvc.Razor.Host/MvcRazorHost.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor.Host/MvcRazorHost.cs
@@ -31,12 +31,16 @@ namespace Microsoft.AspNet.Mvc.Razor
             "Microsoft.AspNet.Mvc",
             "Microsoft.AspNet.Mvc.Rendering",
         };
-        private static readonly Chunk[] _defaultInheritedChunks = new[]
+        private static readonly Chunk[] _defaultInheritedChunks = new Chunk[]
         {
             new InjectChunk("Microsoft.AspNet.Mvc.Rendering.IHtmlHelper<TModel>", HtmlHelperPropertyName),
             new InjectChunk("Microsoft.AspNet.Mvc.Rendering.IJsonHelper", "Json"),
             new InjectChunk("Microsoft.AspNet.Mvc.IViewComponentHelper", "Component"),
             new InjectChunk("Microsoft.AspNet.Mvc.IUrlHelper", "Url"),
+            new AddTagHelperChunk
+            {
+                LookupText = "Microsoft.AspNet.Mvc.Razor.TagHelpers.UrlResolutionTagHelper, Microsoft.AspNet.Mvc.Razor"
+            },
         };
 
         // CodeGenerationContext.DefaultBaseClass is set to MyBaseType<dynamic>.
@@ -99,7 +103,6 @@ namespace Microsoft.AspNet.Mvc.Razor
                     MarkAsHtmlEncodedMethodName = HtmlHelperPropertyName + ".Raw",
                 })
             {
-                ResolveUrlMethodName = "Href",
                 BeginContextMethodName = "BeginContext",
                 EndContextMethodName = "EndContext"
             };
--- a/src/Microsoft.AspNet.Mvc.Razor/Resources.resx
+++ b/src/Microsoft.AspNet.Mvc.Razor/Resources.resx
@@ -1,17 +1,17 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <root>
-  <!--
-    Microsoft ResX Schema
-
+  <!-- 
+    Microsoft ResX Schema 
+    
     Version 2.0
-
-    The primary goals of this format is to allow a simple XML format
-    that is mostly human readable. The generation and parsing of the
-    various data types are done through the TypeConverter classes
+    
+    The primary goals of this format is to allow a simple XML format 
+    that is mostly human readable. The generation and parsing of the 
+    various data types are done through the TypeConverter classes 
     associated with the data types.
-
+    
     Example:
-
+    
     ... ado.net/XML headers & schema ...
     <resheader name="resmimetype">text/microsoft-resx</resheader>
     <resheader name="version">2.0</resheader>
@@ -26,36 +26,36 @@
         <value>[base64 mime encoded string representing a byte array form of the .NET Framework object]</value>
         <comment>This is a comment</comment>
     </data>
-
-    There are any number of "resheader" rows that contain simple
+                
+    There are any number of "resheader" rows that contain simple 
     name/value pairs.
-
-    Each data row contains a name, and value. The row also contains a
-    type or mimetype. Type corresponds to a .NET class that support
-    text/value conversion through the TypeConverter architecture.
-    Classes that don't support this are serialized and stored with the
+    
+    Each data row contains a name, and value. The row also contains a 
+    type or mimetype. Type corresponds to a .NET class that support 
+    text/value conversion through the TypeConverter architecture. 
+    Classes that don't support this are serialized and stored with the 
     mimetype set.
-
-    The mimetype is used for serialized objects, and tells the
-    ResXResourceReader how to depersist the object. This is currently not
+    
+    The mimetype is used for serialized objects, and tells the 
+    ResXResourceReader how to depersist the object. This is currently not 
     extensible. For a given mimetype the value must be set accordingly:
-
-    Note - application/x-microsoft.net.object.binary.base64 is the format
-    that the ResXResourceWriter will generate, however the reader can
+    
+    Note - application/x-microsoft.net.object.binary.base64 is the format 
+    that the ResXResourceWriter will generate, however the reader can 
     read any of the formats listed below.
-
+    
     mimetype: application/x-microsoft.net.object.binary.base64
-    value   : The object must be serialized with
+    value   : The object must be serialized with 
             : System.Runtime.Serialization.Formatters.Binary.BinaryFormatter
             : and then encoded with base64 encoding.
-
+    
     mimetype: application/x-microsoft.net.object.soap.base64
-    value   : The object must be serialized with
+    value   : The object must be serialized with 
             : System.Runtime.Serialization.Formatters.Soap.SoapFormatter
             : and then encoded with base64 encoding.
 
     mimetype: application/x-microsoft.net.object.bytearray.base64
-    value   : The object must be serialized into a byte array
+    value   : The object must be serialized into a byte array 
             : using a System.ComponentModel.TypeConverter
             : and then encoded with base64 encoding.
     -->
@@ -194,5 +194,10 @@
   </data>
   <data name="RazorPage_InvalidTagHelperIndexerAssignment" xml:space="preserve">
     <value>Unable to perform '{0}' assignment. Tag helper property '{1}.{2}' must not be null.</value>
+  </data>
+  <data name="CouldNotResolveApplicationRelativeUrl_TagHelper" xml:space="preserve">
+    <value>Unexpected return value from '{1}.{2}' for URL '{0}'. If the '{1}' service has been overridden, change '{2}' to replace only the '~/' prefix. Otherwise, add the following directive to the Razor page to disable URL resolution relative to the application's 'webroot' setting:
+
+@{3} "{4}, {5}"</value>
   </data>
 </root>
--- a/src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
@@ -1,0 +1,218 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Linq;
+using System.Reflection;
+using Microsoft.AspNet.Mvc.Rendering;
+using Microsoft.AspNet.Razor.Runtime.TagHelpers;
+using Microsoft.Framework.WebEncoders;
+
+namespace Microsoft.AspNet.Mvc.Razor.TagHelpers
+{
+    /// <summary>
+    /// <see cref="ITagHelper"/> implementation targeting elements containing attributes with URL expected values.
+    /// </summary>
+    /// <remarks>Resolves URLs starting with '~/' (relative to the application's 'webroot' setting) that are not
+    /// targeted by other <see cref="ITagHelper"/>s. Runs prior to other <see cref="ITagHelper"/>s to ensure
+    /// application-relative URLs are resolved.</remarks>
+    [TargetElement("*", Attributes = "itemid")]
+    [TargetElement("a", Attributes = "href")]
+    [TargetElement("applet", Attributes = "archive")]
+    [TargetElement("area", Attributes = "href")]
+    [TargetElement("audio", Attributes = "src")]
+    [TargetElement("base", Attributes = "href")]
+    [TargetElement("blockquote", Attributes = "cite")]
+    [TargetElement("button", Attributes = "formaction")]
+    [TargetElement("del", Attributes = "cite")]
+    [TargetElement("embed", Attributes = "src")]
+    [TargetElement("form", Attributes = "action")]
+    [TargetElement("html", Attributes = "manifest")]
+    [TargetElement("iframe", Attributes = "src")]
+    [TargetElement("img", Attributes = "src")]
+    [TargetElement("input", Attributes = "src")]
+    [TargetElement("input", Attributes = "formaction")]
+    [TargetElement("ins", Attributes = "cite")]
+    [TargetElement("link", Attributes = "href")]
+    [TargetElement("menuitem", Attributes = "icon")]
+    [TargetElement("object", Attributes = "archive")]
+    [TargetElement("object", Attributes = "data")]
+    [TargetElement("q", Attributes = "cite")]
+    [TargetElement("script", Attributes = "src")]
+    [TargetElement("source", Attributes = "src")]
+    [TargetElement("track", Attributes = "src")]
+    [TargetElement("video", Attributes = "src")]
+    [TargetElement("video", Attributes = "poster")]
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public class UrlResolutionTagHelper : TagHelper
+    {
+        // Valid whitespace characters defined by the HTML5 spec.
+        private static readonly char[] ValidAttributeWhitespaceChars =
+            new[] { '\t', '\n', '\u000C', '\r', ' ' };
+        private static readonly IReadOnlyDictionary<string, IEnumerable<string>> ElementAttributeLookups =
+            new Dictionary<string, IEnumerable<string>>(StringComparer.OrdinalIgnoreCase)
+            {
+                { "a", new[] { "href" } },
+                { "applet", new[] { "archive" } },
+                { "area", new[] { "href" } },
+                { "audio", new[] { "src" } },
+                { "base", new[] { "href" } },
+                { "blockquote", new[] { "cite" } },
+                { "button", new[] { "formaction" } },
+                { "del", new[] { "cite" } },
+                { "embed", new[] { "src" } },
+                { "form", new[] { "action" } },
+                { "html", new[] { "manifest" } },
+                { "iframe", new[] { "src" } },
+                { "img", new[] { "src" } },
+                { "input", new[] { "src", "formaction" } },
+                { "ins", new[] { "cite" } },
+                { "link", new[] { "href" } },
+                { "menuitem", new[] { "icon" } },
+                { "object", new[] { "archive", "data" } },
+                { "q", new[] { "cite" } },
+                { "script", new[] { "src" } },
+                { "source", new[] { "src" } },
+                { "track", new[] { "src" } },
+                { "video", new[] { "poster", "src" } },
+            };
+
+        /// <summary>
+        /// Creates a new <see cref="UrlResolutionTagHelper"/>.
+        /// </summary>
+        /// <param name="urlHelper">The <see cref="IUrlHelper"/>.</param>
+        /// <param name="htmlEncoder">The <see cref="IHtmlEncoder"/>.</param>
+        public UrlResolutionTagHelper(IUrlHelper urlHelper, IHtmlEncoder htmlEncoder)
+        {
+            UrlHelper = urlHelper;
+            HtmlEncoder = htmlEncoder;
+        }
+
+        /// <inheritdoc />
+        public override int Order
+        {
+            get
+            {
+                return DefaultOrder.DefaultFrameworkSortOrder - 999;
+            }
+        }
+
+        protected IUrlHelper UrlHelper { get; }
+
+        protected IHtmlEncoder HtmlEncoder { get; }
+
+        /// <inheritdoc />
+        public override void Process(TagHelperContext context, TagHelperOutput output)
+        {
+            IEnumerable<string> attributeNames;
+            if (ElementAttributeLookups.TryGetValue(output.TagName, out attributeNames))
+            {
+                foreach (var attributeName in attributeNames)
+                {
+                    ProcessUrlAttribute(attributeName, output);
+                }
+            }
+
+            // itemid can be present on any HTML element.
+            ProcessUrlAttribute("itemid", output);
+        }
+
+        /// <summary>
+        /// Resolves and updates URL values starting with '~/' (relative to the application's 'webroot' setting) for
+        /// <paramref name="output"/>'s <see cref="TagHelperOutput.Attributes"/> whose
+        /// <see cref="TagHelperAttribute.Name"/> is <paramref name="attributeName"/>.
+        /// </summary>
+        /// <param name="attributeName">The attribute name used to lookup values to resolve.</param>
+        /// <param name="output">The <see cref="TagHelperOutput"/>.</param>
+        protected void ProcessUrlAttribute(string attributeName, TagHelperOutput output)
+        {
+            IEnumerable<TagHelperAttribute> attributes;
+            if (output.Attributes.TryGetAttributes(attributeName, out attributes))
+            {
+                foreach (var attribute in attributes)
+                {
+                    string resolvedUrl;
+
+                    var stringValue = attribute.Value as string;
+                    if (stringValue != null)
+                    {
+                        if (TryResolveUrl(stringValue, encodeWebRoot: false, resolvedUrl: out resolvedUrl))
+                        {
+                            attribute.Value = resolvedUrl;
+                        }
+                    }
+                    else
+                    {
+                        var htmlStringValue = attribute.Value as HtmlString;
+                        if (htmlStringValue != null &&
+                            TryResolveUrl(
+                                htmlStringValue.ToString(),
+                                encodeWebRoot: true,
+                                resolvedUrl: out resolvedUrl))
+                        {
+                            attribute.Value = new HtmlString(resolvedUrl);
+                        }
+                    }
+                }
+            }
+        }
+
+        /// <summary>
+        /// Tries to resolve the given <paramref name="url"/> value relative to the application's 'webroot' setting.
+        /// </summary>
+        /// <param name="url">The URL to resolve.</param>
+        /// <param name="encodeWebRoot">If <c>true</c>, will HTML encode the expansion of '~/'.</param>
+        /// <param name="resolvedUrl">Absolute URL beginning with the application's virtual root. <c>null</c> if
+        /// <paramref name="url"/> could not be resolved.</param>
+        /// <returns><c>true</c> if the <paramref name="url"/> could be resolved; <c>false</c> otherwise.</returns>
+        protected bool TryResolveUrl(string url, bool encodeWebRoot, out string resolvedUrl)
+        {
+            resolvedUrl = null;
+
+            if (url == null)
+            {
+                return false;
+            }
+
+            var trimmedUrl = url.Trim(ValidAttributeWhitespaceChars);
+
+            // Before doing more work, ensure that the URL we're looking at is app relative.
+            if (trimmedUrl.Length >= 2 && trimmedUrl[0] == '~' && trimmedUrl[1] == '/')
+            {
+                var appRelativeUrl = UrlHelper.Content(trimmedUrl);
+
+                if (encodeWebRoot)
+                {
+                    var postTildeSlashUrlValue = trimmedUrl.Substring(2);
+
+                    if (!appRelativeUrl.EndsWith(postTildeSlashUrlValue, StringComparison.Ordinal))
+                    {
+                        throw new InvalidOperationException(
+                            Resources.FormatCouldNotResolveApplicationRelativeUrl_TagHelper(
+                                url,
+                                nameof(IUrlHelper),
+                                nameof(IUrlHelper.Content),
+                                "removeTagHelper",
+                                typeof(UrlResolutionTagHelper).FullName,
+                                typeof(UrlResolutionTagHelper).GetTypeInfo().Assembly.GetName().Name));
+                    }
+
+                    var applicationPath = appRelativeUrl.Substring(0, appRelativeUrl.Length - postTildeSlashUrlValue.Length);
+                    var encodedApplicationPath = HtmlEncoder.HtmlEncode(applicationPath);
+
+                    resolvedUrl = string.Concat(encodedApplicationPath, postTildeSlashUrlValue);
+                }
+                else
+                {
+                    resolvedUrl = appRelativeUrl;
+                }
+
+                return true;
+            }
+
+            return false;
+        }
+    }
+}
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
@@ -1,11 +1,12 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
 using Microsoft.AspNet.Hosting;
+using Microsoft.AspNet.Mvc.Razor.TagHelpers;
 using Microsoft.AspNet.Mvc.TagHelpers.Internal;
 using Microsoft.AspNet.Razor.Runtime.TagHelpers;
 using Microsoft.Framework.Caching.Memory;
+using Microsoft.Framework.WebEncoders;
 
 namespace Microsoft.AspNet.Mvc.TagHelpers
 {
@@ -16,7 +17,7 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
     /// The tag helper won't process for cases with just the 'src' attribute.
     /// </remarks>
     [TargetElement("img", Attributes = AppendVersionAttributeName + "," + SrcAttributeName)]
-    public class ImageTagHelper : TagHelper
+    public class ImageTagHelper : UrlResolutionTagHelper
     {
         private static readonly string Namespace = typeof(ImageTagHelper).Namespace;
 
@@ -30,7 +31,13 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
         /// </summary>
         /// <param name="hostingEnvironment">The <see cref="IHostingEnvironment"/>.</param>
         /// <param name="cache">The <see cref="IMemoryCache"/>.</param>
-        public ImageTagHelper(IHostingEnvironment hostingEnvironment, IMemoryCache cache)
+        /// <param name="urlHelper">The <see cref="IUrlHelper"/>.</param>
+        public ImageTagHelper(
+            IHostingEnvironment hostingEnvironment,
+            IMemoryCache cache,
+            IHtmlEncoder htmlEncoder,
+            IUrlHelper urlHelper)
+            : base(urlHelper, htmlEncoder)
         {
             HostingEnvironment = hostingEnvironment;
             Cache = cache;
@@ -68,12 +75,19 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
             if (AppendVersion)
             {
                 EnsureFileVersionProvider();
+
+                string resolvedUrl;
+                if (TryResolveUrl(Src, encodeWebRoot: false, resolvedUrl: out resolvedUrl))
+                {
+                    Src = resolvedUrl;
+                }
                 output.Attributes[SrcAttributeName] = _fileVersionProvider.AddFileVersionToPath(Src);
             }
             else
             {
                 // Pass through attribute that is also a well-known HTML attribute.
                 output.CopyHtmlAttribute(SrcAttributeName, context);
+                ProcessUrlAttribute(SrcAttributeName, output);
             }
         }
 
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
@@ -2,11 +2,11 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
 using Microsoft.AspNet.Hosting;
+using Microsoft.AspNet.Mvc.Razor.TagHelpers;
 using Microsoft.AspNet.Mvc.TagHelpers.Internal;
 using Microsoft.AspNet.Razor.Runtime.TagHelpers;
 using Microsoft.Framework.Caching.Memory;
@@ -30,7 +30,7 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
     [TargetElement("link", Attributes = FallbackTestPropertyAttributeName)]
     [TargetElement("link", Attributes = FallbackTestValueAttributeName)]
     [TargetElement("link", Attributes = AppendVersionAttributeName)]
-    public class LinkTagHelper : TagHelper
+    public class LinkTagHelper : UrlResolutionTagHelper
     {
         private static readonly string Namespace = typeof(LinkTagHelper).Namespace;
 
@@ -91,17 +91,19 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
         /// <param name="cache">The <see cref="IMemoryCache"/>.</param>
         /// <param name="htmlEncoder">The <see cref="IHtmlEncoder"/>.</param>
         /// <param name="javaScriptEncoder">The <see cref="IJavaScriptStringEncoder"/>.</param>
+        /// <param name="urlHelper">The <see cref="IUrlHelper"/>.</param>
         public LinkTagHelper(
             ILogger<LinkTagHelper> logger,
             IHostingEnvironment hostingEnvironment,
             IMemoryCache cache,
             IHtmlEncoder htmlEncoder,
-            IJavaScriptStringEncoder javaScriptEncoder)
+            IJavaScriptStringEncoder javaScriptEncoder,
+            IUrlHelper urlHelper)
+            : base(urlHelper, htmlEncoder)
         {
             Logger = logger;
             HostingEnvironment = hostingEnvironment;
             Cache = cache;
-            HtmlEncoder = htmlEncoder;
             JavaScriptEncoder = javaScriptEncoder;
         }
 
@@ -195,8 +197,6 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
 
         protected IMemoryCache Cache { get; }
 
-        protected IHtmlEncoder HtmlEncoder { get; }
-
         protected IJavaScriptStringEncoder JavaScriptEncoder { get; }
 
         // Internal for ease of use when testing.
@@ -205,10 +205,20 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
         /// <inheritdoc />
         public override void Process(TagHelperContext context, TagHelperOutput output)
         {
+            string resolvedUrl;
+
             // Pass through attribute that is also a well-known HTML attribute.
             if (Href != null)
             {
                 output.CopyHtmlAttribute(HrefAttributeName, context);
+
+                // Resolve any application relative URLs (~/) now so they can be used in comparisons later.
+                if (TryResolveUrl(Href, encodeWebRoot: false, resolvedUrl: out resolvedUrl))
+                {
+                    Href = resolvedUrl;
+                }
+
+                ProcessUrlAttribute(HrefAttributeName, output);
             }
 
             var modeResult = AttributeMatcher.DetermineMode(context, ModeDetails);
@@ -243,6 +253,15 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
 
             if (mode == Mode.GlobbedHref || mode == Mode.Fallback && !string.IsNullOrEmpty(HrefInclude))
             {
+                if (TryResolveUrl(HrefInclude, encodeWebRoot: false, resolvedUrl: out resolvedUrl))
+                {
+                    HrefInclude = resolvedUrl;
+                }
+                if (TryResolveUrl(HrefExclude, encodeWebRoot: false, resolvedUrl: out resolvedUrl))
+                {
+                    HrefExclude = resolvedUrl;
+                }
+
                 BuildGlobbedLinkTags(attributes, builder);
                 if (string.IsNullOrEmpty(Href))
                 {
@@ -254,6 +273,19 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
 
             if (mode == Mode.Fallback)
             {
+                if (TryResolveUrl(FallbackHref, encodeWebRoot: false, resolvedUrl: out resolvedUrl))
+                {
+                    FallbackHref = resolvedUrl;
+                }
+                if (TryResolveUrl(FallbackHrefInclude, encodeWebRoot: false, resolvedUrl: out resolvedUrl))
+                {
+                    FallbackHrefInclude = resolvedUrl;
+                }
+                if (TryResolveUrl(FallbackHrefExclude, encodeWebRoot: false, resolvedUrl: out resolvedUrl))
+                {
+                    FallbackHrefExclude = resolvedUrl;
+                }
+
                 BuildFallbackBlock(builder);
             }
 
@@ -292,7 +324,7 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
             {
                 if (AppendVersion == true)
                 {
-                    for (var i=0; i < fallbackHrefs.Length; i++)
+                    for (var i = 0; i < fallbackHrefs.Length; i++)
                     {
                         // fallbackHrefs come from bound attributes and globbing. Must always be non-null.
                         Debug.Assert(fallbackHrefs[i] != null);
--- a/test/Microsoft.AspNet.Mvc.FunctionalTests/UrlResolutionTest.cs
+++ b/test/Microsoft.AspNet.Mvc.FunctionalTests/UrlResolutionTest.cs
@@ -1,0 +1,71 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Reflection;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Builder;
+using Microsoft.Framework.DependencyInjection;
+using Microsoft.Framework.WebEncoders;
+using RazorWebSite;
+using Xunit;
+
+namespace Microsoft.AspNet.Mvc.FunctionalTests
+{
+    public class UrlResolutionTest
+    {
+        private const string SiteName = nameof(RazorWebSite);
+        private readonly Action<IApplicationBuilder> _app = new Startup().Configure;
+        private readonly Action<IServiceCollection> _configureServices = new Startup().ConfigureServices;
+        private static readonly Assembly _resourcesAssembly = typeof(UrlResolutionTest).GetTypeInfo().Assembly;
+
+        [Fact]
+        public async Task AppRelativeUrlsAreResolvedCorrectly()
+        {
+            var server = TestHelper.CreateServer(_app, SiteName, _configureServices);
+
+            var client = server.CreateClient();
+            var outputFile = "compiler/resources/RazorWebSite.UrlResolution.Index.html";
+            var expectedContent =
+                await ResourceFile.ReadResourceAsync(_resourcesAssembly, outputFile, sourceFile: false);
+
+            // Act
+            var response = await client.GetAsync("http://localhost/UrlResolution/Index");
+            var responseContent = await response.Content.ReadAsStringAsync();
+
+            // Assert
+            responseContent = responseContent.Trim();
+#if GENERATE_BASELINES
+            ResourceFile.UpdateFile(_resourcesAssembly, outputFile, expectedContent, responseContent);
+#else
+            Assert.Equal(expectedContent.Trim(), responseContent, ignoreLineEndingDifferences: true);
+#endif
+        }
+
+        [Fact]
+        public async Task AppRelativeUrlsAreResolvedAndEncodedCorrectly()
+        {
+            var server = TestHelper.CreateServer(_app, SiteName, services =>
+            {
+                _configureServices(services);
+                services.AddTransient<IHtmlEncoder, TestHtmlEncoder>();
+            });
+            var client = server.CreateClient();
+            var outputFile = "compiler/resources/RazorWebSite.UrlResolution.Index.Encoded.html";
+            var expectedContent =
+                await ResourceFile.ReadResourceAsync(_resourcesAssembly, outputFile, sourceFile: false);
+
+            // Act
+            var response = await client.GetAsync("http://localhost/UrlResolution/Index");
+            var responseContent = await response.Content.ReadAsStringAsync();
+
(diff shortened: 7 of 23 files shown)


src/Microsoft.AspNet.Mvc.Core/UrlHelper.cs
```diff
@@ -120,8 +120,13 @@ protected virtual string GeneratePathFromRoute(string routeName, IDictionary<str
...
         /// <inheritdoc />
-        public virtual string Content([NotNull] string contentPath)
+        public virtual string Content(string contentPath)
         {
+            if (contentPath == null)
+            {
+                return null;
+            }
+
             return GenerateClientUrl(_httpContext.Request.PathBase, contentPath);
```
ajaybhargavb: Not sure if we need to fix this here, but one of the reasons for #2841 is that `GenerateClientUrl` isn't trimming the `contentPath` so a leading space will result in `~` not being removed.
Do we want to fix that here or leaving the trimming up to caller?

NTaylorMullen: Hmm, part of me feels like it'd be too protective over users; but then again I can't imagine a user purposefully calling into this method with an untrimmed value.

I'm not sure change this method directly would be the right approach; however, it brings up some gray area in the `UrlResolutionTagHelper` where it is smarter about parsing URLs with whitespace where our `ScriptTagHelper`, `LinkTagHelper` and `ImageTagHelper` take a users value verbatim and hand it over to the `IUrlHelper`.

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
```diff
@@ -61,9 +63,13 @@ public ImageTagHelper(IHostingEnvironment hostingEnvironment, IMemoryCache cache
 
         protected IMemoryCache Cache { get; }
 
+        protected IUrlHelper UrlHelper { get; }
+
         /// <inheritdoc />
         public override void Process(TagHelperContext context, TagHelperOutput output)
         {
+            Src = UrlHelper.Content(Src);
```
NTaylorMullen: Do we want to be smarter about these values? Aka handle whitespace much like the `UrlResolutionTagHelper` does? Question spurred from convo: https://github.com/aspnet/Mvc/pull/2875/files#r35715590

I'm torn between doing the extra computation and keeping the users value as they provided it (since executing a `TagHelper` is more like executing a method).

/cc @DamianEdwards

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
```diff
@@ -0,0 +1,206 @@
...
+
+            // Find the start of the potential application relative URL value.
+            var prefixEndIndex = -1;
+            while (++prefixEndIndex < urlLength && char.IsWhiteSpace(url[prefixEndIndex])) ;
+
+            // Before doing more work, ensure that the URL we're looking at is app relative.
+            if (prefixEndIndex < urlLength - 2 && url[prefixEndIndex] == '~' && url[prefixEndIndex + 1] == '/')
+            {
+                var valueEndIndex = prefixEndIndex - 1;
+                while (++valueEndIndex < urlLength && !char.IsWhiteSpace(url[valueEndIndex])) ;
```
ajaybhargavb: Remove space before `;`

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
```diff
@@ -0,0 +1,206 @@
...
+
+                var prefix = url.Substring(0, prefixEndIndex);
+                var urlValue = url.Substring(prefixEndIndex, valueEndIndex - prefixEndIndex);
+                var appRelativeUrl = UrlHelper.Content(urlValue);
+                var suffix = url.Substring(valueEndIndex);
+
+                if (tryEncodeApplicationPath)
+                {
+                    var postTildaSlashUrlValue = urlValue.Substring(2);
```
ajaybhargavb: `postTildeSlashValue`

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
dougbu: Better-yet replace the `;` with
``` c#
  {
  }
```

IDisposable: :+1: { @dougbu }

src/Microsoft.AspNet.Mvc.Razor.Host/MvcRazorHost.cs
```diff
@@ -31,12 +31,16 @@ public class MvcRazorHost : RazorEngineHost, IMvcRazorHost
...
+        private static readonly Chunk[] _defaultInheritedChunks = new Chunk[]
         {
             new InjectChunk("Microsoft.AspNet.Mvc.Rendering.IHtmlHelper<TModel>", HtmlHelperPropertyName),
             new InjectChunk("Microsoft.AspNet.Mvc.Rendering.IJsonHelper", "Json"),
             new InjectChunk("Microsoft.AspNet.Mvc.IViewComponentHelper", "Component"),
             new InjectChunk("Microsoft.AspNet.Mvc.IUrlHelper", "Url"),
+            new AddTagHelperChunk
+            {
+                LookupText = "Microsoft.AspNet.Mvc.Razor.TagHelpers.UrlResolutionTagHelper, Microsoft.AspNet.Mvc.Razor"
+            }
```
dougbu: please stick with the trailing comma

src/Microsoft.AspNet.Mvc.Razor/Resources.resx
```diff
@@ -195,4 +195,9 @@
   <data name="RazorPage_InvalidTagHelperIndexerAssignment" xml:space="preserve">
     <value>Unable to perform '{0}' assignment. Tag helper property '{1}.{2}' must not be null.</value>
   </data>
+  <data name="CouldNotResolveApplicationRelativeUrl_TagHelper" xml:space="preserve">
+    <value>Could not resolve application relative URL '{0}' by calling '{1}.{2}'. To disable application relative URL resolution you can add the following directive to your Razor page:
+
+@removeTagHelper "{3}, {4}".</value>
```
dougbu: Given the current focus of the message, the `IUrlHelper.Content()` call is an implementation detail of the `UrlResolutionTagHelper`. Should remove `by calling '{1}.{2}'` if the message otherwise remains unchanged.

The period in the code sample is incorrect and should also be removed.

But resolution was probably fine and removing the `UrlResolutionTagHelper` is probably incorrect for this case. It's more likely the user has overridden the `IUrlHelper` in DI, replacing it with something that does more than change the prefix.

All up
```
Unexpected return value from '{1}.{2}' for URL '{0}'. If you have overridden the '{1}' service, change '{2}' to replace only the '~' prefix. Otherwise, add the following directive to your Razor page to disable application-relative URL resolution:

@removeTagHelper "{3}, {4}"
```

dougbu: Separately it may be worth passing `"removeTagHelper"` in.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
```diff
@@ -0,0 +1,206 @@
...
+using System.ComponentModel;
+using System.Reflection;
+using Microsoft.AspNet.Mvc.Rendering;
+using Microsoft.AspNet.Razor.Runtime.TagHelpers;
+using Microsoft.Framework.WebEncoders;
+
+namespace Microsoft.AspNet.Mvc.Razor.TagHelpers
+{
+    /// <summary>
+    /// <see cref="ITagHelper"/> implementation targeting elements containing attributes with url expected values.
```
dougbu: Stick with uppercase "URL".

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
```diff
@@ -0,0 +1,206 @@
...
+                { "button", new[] { "formaction" } },
+                { "form", new[] { "action" } },
+                { "blockquote", new[] { "cite" } },
+                { "del", new[] { "cite" } },
+                { "ins", new[] { "cite" } },
+                { "q", new[] { "cite" } },
+                { "menuitem", new[] { "icon" } },
+                { "html", new[] { "manifest" } },
+                { "object", new[] { "data", "archive" } },
+                { "applet", new[] { "archive" } },
```
dougbu: Hard to make sense of this collection and all of the `[TargetElement]` attributes. Suggest sorting both alphabetically by element then attribute names.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
```diff
@@ -0,0 +1,206 @@
...
+            UrlHelper = urlHelper;
+            HtmlEncoder = htmlEncoder;
+        }
+
+        /// <inheritdoc />
+        public override int Order
+        {
+            get
+            {
+                return int.MinValue;
```
dougbu: Understand we want this one to run early but earliest possible means the only thing anyone can do to override this helper is to remove it. Doesn't cover for example changing a specific attribute this helper targets. Further comments in `DefaultOrder` imply we shouldn't be using `0` as we do today in other tag helpers or `MinValue`: `// ... framework code that depends on order should be ordered between -1 to -1999.`.

Suggest using MVC-traditional `DefaultOrder.DefaultFrameworkSortOrder` (`-1000`). If instead we decide the rest of our tag helpers should have `DefaultOrder.DefaultFrameworkSortOrder`, then this one should have `DefaultOrder.DefaultFrameworkSortOrder - 999`.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
```diff
@@ -0,0 +1,206 @@
...
+using Microsoft.AspNet.Mvc.Rendering;
+using Microsoft.AspNet.Razor.Runtime.TagHelpers;
+using Microsoft.Framework.WebEncoders;
+
+namespace Microsoft.AspNet.Mvc.Razor.TagHelpers
+{
+    /// <summary>
+    /// <see cref="ITagHelper"/> implementation targeting elements containing attributes with url expected values.
+    /// </summary>
+    /// <remarks>Resolves application relative URLs.</remarks>
```
dougbu: Also mention that this runs prior to other tag helpers and thus targets elements and attributes based on the Razor file content. Plus that this tag helper only targets attributes no other tag helper binds.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
dougbu: Pity the `[TargetElement]` attributes can't be changed via configuration. Nonetheless should this come from configuration or via an `ITagHelperInitializer`? That would at least allow users to subset the actions of this tag helper without turning it off completely or running another tag helper before it.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
```diff
@@ -0,0 +1,206 @@
...
+                            attribute.Value = resolvedUrl;
+                        }
+                    }
+                    else if (attribute.Value is HtmlString)
+                    {
+                        var htmlStringValue = ((HtmlString)attribute.Value).ToString();
+                        if (TryResolveUrl(
+                                htmlStringValue,
+                                tryEncodeApplicationPath: true,
+                                resolvedUrl: out resolvedUrl))
```
dougbu: Formatting is off here and in the `TryResolveUrl()` call above.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
```diff
@@ -0,0 +1,206 @@
...
+            if (url == null)
+            {
+                return false;
+            }
+
+            var urlLength = url.Length;
+
+            // Find the start of the potential application relative URL value.
+            var prefixEndIndex = -1;
+            while (++prefixEndIndex < urlLength && char.IsWhiteSpace(url[prefixEndIndex])) ;
```
dougbu: Statement block here too.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
```diff
@@ -0,0 +1,206 @@
...
+            resolvedUrl = null;
+
+            if (url == null)
+            {
+                return false;
+            }
+
+            var urlLength = url.Length;
+
+            // Find the start of the potential application relative URL value.
```
dougbu: Using "application relative" a few times in this PR. Where is that defined? How would a developer interpret this wording a few years down the road?

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
dougbu: `char.IsWhiteSpace()` covers far more characters than the expected Latin1 set. Since newlines, tabs, and so on aren't supported in a URL, suggest just `' ' == url[prefixEndIndex]`.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
```diff
@@ -0,0 +1,206 @@
...
+                                nameof(IUrlHelper),
+                                nameof(IUrlHelper.Content),
+                                typeof(UrlResolutionTagHelper).FullName,
+                                typeof(UrlResolutionTagHelper).GetTypeInfo().Assembly.GetName().Name));
+                    }
+
+                    var applicationPath = appRelativeUrl.Substring(0, appRelativeUrl.Length - postTildaSlashUrlValue.Length);
+                    var encodedApplicationPath = HtmlEncoder.HtmlEncode(applicationPath);
+
+                    resolvedUrl = string.Join(string.Empty, prefix, encodedApplicationPath, postTildaSlashUrlValue, suffix);
```
dougbu: - Why restore the whitespace (the `prefix`)? Unless you're going to support attributes like `html/manifest` completely, whitespace isn't relevant.
- Method seems to focus on efficiency e.g. not using `string,Trim()` or `StartsWith()`. If that's important, use string interpolation or `string.Format()` and not `string.Join()`.
- Wrap long lines.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
```diff
@@ -0,0 +1,206 @@
...
+
+            // Before doing more work, ensure that the URL we're looking at is app relative.
+            if (prefixEndIndex < urlLength - 2 && url[prefixEndIndex] == '~' && url[prefixEndIndex + 1] == '/')
+            {
+                var valueEndIndex = prefixEndIndex - 1;
+                while (++valueEndIndex < urlLength && !char.IsWhiteSpace(url[valueEndIndex])) ;
+
+                var prefix = url.Substring(0, prefixEndIndex);
+                var urlValue = url.Substring(prefixEndIndex, valueEndIndex - prefixEndIndex);
+                var appRelativeUrl = UrlHelper.Content(urlValue);
```
dougbu: Don't abbreviate, `applicationRelativeUrl`

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
```diff
@@ -73,6 +79,9 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
             {
                 // Pass through attribute that is also a well-known HTML attribute.
                 output.CopyHtmlAttribute(SrcAttributeName, context);
+
+                // Need to update the TagHelperOutput's attribute if the URL is application relative.
+                output.Attributes[SrcAttributeName].Value = Src;
```
dougbu: Doesn't play nicely with other tag helpers targeting this element and makes much of the work done in `CopyHtmlAttribute()` irrelevant. Should either be conditional e.g. done only if `output.Attributes[SrcAttributeName].Value` matches the original `Src` value or instead `UrlHelper.Content(output.Attributes[SrcAttributeName].Value)`.

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
dougbu: Since "valid non-empty URL potentially surrounded by spaces" appears fairly frequently in the HTML 5.0 spec and is applied to all attributes updated in existing tag helpers, should at least `Trim()` the value before calling `Content()`.

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
```diff
@@ -73,6 +79,9 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
             {
                 // Pass through attribute that is also a well-known HTML attribute.
                 output.CopyHtmlAttribute(SrcAttributeName, context);
+
+                // Need to update the TagHelperOutput's attribute if the URL is application relative.
```
dougbu: `Content()` is called about here in the other two existing tag helpers. Why is it done above this block in this tag helper?

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
dougbu: The HTML 5.0 specification says the relevant spaces here are "U+0020 SPACE, "tab" (U+0009), "LF" (U+000A), "FF" (U+000C), and "CR" (U+000D)". For example
- [`script/src`](http://www.w3.org/TR/html5/scripting-1.html#attr-script-src) says
> The value of the attribute must be a valid non-empty URL potentially surrounded by spaces ...

- That links to [text stating](http://www.w3.org/TR/html5/infrastructure.html#valid-non-empty-url-potentially-surrounded-by-spaces)
> A string is a valid non-empty URL potentially surrounded by spaces if, after stripping leading and trailing whitespace from it, it is a valid non-empty URL.

- The "stripping" part links to [text stating](http://www.w3.org/TR/html5/infrastructure.html#strip-leading-and-trailing-whitespace)
> ... the user agent must remove all space characters that are at the start or end of the string.

> "space characters links to the [definition above](http://www.w3.org/TR/html5/infrastructure.html#space-character)

Basically the "whitespace" wording in the middle there is a misdirection.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
NTaylorMullen: Second comment: Nah, if we could configure the `TargetElement` like you mentioned then I'd be be on board; however, this list is directly correlated to what the `TagHelper` targets. If a user wants to subset the `TagHelper` a better choice would be to sub-class.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
NTaylorMullen: 1. User typed whitespace, no need to nuke it.
1. Wasn't aware `string.Format` was quicker than `string.Join`.
1. :+1:

IDisposable: I don't agree that string.Format is faster than joining strings via simple concatenation. Did anything MAJOR change since http://www.yoda.arachsys.com/csharp/stringbuilder.html because String.Concat is what REALLY should be here. (and for reference, that's what the C# compiler will emit for simple concatenations.

NTaylorMullen: Ahh, very nice :+1: @IDisposable

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
NTaylorMullen: Will just use `UrlHelper.Content` on both, no need to compare the strings.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
NTaylorMullen: It's what we've always used

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
NTaylorMullen: Inconsistent, moved to be next to where it's used.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
dougbu: Where?

NTaylorMullen: Search "application relative" on the MVC repo. It's used all over.

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
dougbu: Ah, :ok: there are a few mentions of "application relative" but all are using the obvious meaning: Relative to the application (or project) root e.g .finding views. That is _not_ what you're doing here. These paths are relative to the web root. Hence my confusion.

Not many comments close to our `IHostingEnvironment.WebRootFileProvider` uses. Closest I see is a few comments about how paths are "relative to the application's 'webroot' setting".

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
```diff
@@ -0,0 +1,212 @@
...
+            {
+                foreach (var attribute in attributes)
+                {
+                    string resolvedUrl;
+                    if (attribute.Value is string)
+                    {
+                        if (TryResolveUrl(
+                                (string)attribute.Value,
+                                tryEncodeApplicationPath: false,
+                                resolvedUrl: out resolvedUrl))
```
dougbu: Fixed one bit of "bit off" formatting but not the other...

NTaylorMullen:
Looks like some tests broke after rebasing, fixed them and the formatting issue.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
```diff
@@ -0,0 +1,212 @@
...
+using Microsoft.AspNet.Razor.Runtime.TagHelpers;
+using Microsoft.Framework.WebEncoders;
+
+namespace Microsoft.AspNet.Mvc.Razor.TagHelpers
+{
+    /// <summary>
+    /// <see cref="ITagHelper"/> implementation targeting elements containing attributes with URL expected values.
+    /// </summary>
+    /// <remarks>Resolves application relative URLs that are not targeted by other <see cref="ITagHelper"/>s. Runs
+    /// prior to other <see cref="ITagHelper"/>s to ensure application-relative URLs are resolved.</remarks>
```
dougbu: As discussed elsewhere, "application relative" is not consistent with how that term is used elsewhere. But also have no existing doc comments hyphenating "application-relative".

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
```diff
@@ -0,0 +1,212 @@
...
+    [TargetElement("source", Attributes = "src")]
+    [TargetElement("track", Attributes = "src")]
+    [TargetElement("video", Attributes = "src")]
+    [TargetElement("video", Attributes = "poster")]
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public class UrlResolutionTagHelper : TagHelper
+    {
+        // Valid whitespace characters defined by the HTML5 spec.
+        private static readonly char[] ValidAttributeWhitespaceChars =
+            new[] { '\u0009', '\u000A', '\u000C', '\u000D', ' ' };
```
dougbu: use better-known character escapes i.e. `{ '\t', '\n, '\u000C', '\r', ' ' }`

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
```diff
@@ -0,0 +1,212 @@
...
+
+        private void ProcessUrlAttribute(string attributeName, TagHelperOutput output)
+        {
+            IEnumerable<TagHelperAttribute> attributes;
+            if (output.Attributes.TryGetAttributes(attributeName, out attributes))
+            {
+                foreach (var attribute in attributes)
+                {
+                    string resolvedUrl;
+                    if (attribute.Value is string)
```
dougbu: Why cast twice? `as T` -> `!= null` is more obvious and direct.

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
```diff
@@ -61,18 +63,23 @@ public ImageTagHelper(IHostingEnvironment hostingEnvironment, IMemoryCache cache
...
 
+        protected IUrlHelper UrlHelper { get; }
+
         /// <inheritdoc />
         public override void Process(TagHelperContext context, TagHelperOutput output)
         {
             if (AppendVersion)
             {
                 EnsureFileVersionProvider();
+                Src = UrlHelper.Content(Src.Trim());
```
dougbu: Why restore surrounding whitespace in the `UrlResolutionTagHelper` tag helper but drop it in the others?

Also `Trim()` has the same Unicode vs. HTML issue as `IsWhiteSpace()`. Should call it as `Src.Trim(ValidAttributeWhitespaceChars)` after moving `ValidAttributeWhitespaceChars` some place central.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -209,6 +213,10 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
             if (Href != null)
             {
                 output.CopyHtmlAttribute(HrefAttributeName, context);
+
+                Href = UrlHelper.Content(Href.Trim());
```
dougbu: Add a comment about later `Href` comparison to avoid someone "cleaning this up" later.

test/Microsoft.AspNet.Mvc.FunctionalTests/UrlResolutionTest.cs
```diff
@@ -0,0 +1,71 @@
...
+{
+    public class UrlResolutionTest
+    {
+        private const string SiteName = nameof(RazorWebSite);
+        private readonly Action<IApplicationBuilder> _app = new Startup().Configure;
+        private readonly Action<IServiceCollection> _configureServices = new Startup().ConfigureServices;
+        private static readonly Assembly _resourcesAssembly = typeof(UrlResolutionTest).GetTypeInfo().Assembly;
+
+        [Fact]
+        public async Task AppRelativeUrlsAreResolvedCorrectly()
```
dougbu: Don't use `AppRelative` here either

test/Microsoft.AspNet.Mvc.FunctionalTests/UrlResolutionTest.cs
```diff
@@ -0,0 +1,71 @@
...
+            var outputFile = "compiler/resources/RazorWebSite.UrlResolution.Index.html";
+            var expectedContent =
+                await ResourceFile.ReadResourceAsync(_resourcesAssembly, outputFile, sourceFile: false);
+
+            // Act
+            var response = await client.GetAsync("http://localhost/UrlResolution/Index");
+            var responseContent = await response.Content.ReadAsStringAsync();
+
+            // Assert
+            responseContent = responseContent.Trim();
```
dougbu: If this just about avoiding a few blank lines in the baseline files, does not seem worthwhile. If not, what causes leading / trailing whitespace to change from test run to test run?

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
```diff
@@ -61,18 +63,23 @@ public ImageTagHelper(IHostingEnvironment hostingEnvironment, IMemoryCache cache
...
                 output.Attributes[SrcAttributeName] = _fileVersionProvider.AddFileVersionToPath(Src);
             }
             else
             {
                 // Pass through attribute that is also a well-known HTML attribute.
                 output.CopyHtmlAttribute(SrcAttributeName, context);
+                output.Attributes[SrcAttributeName].Value =
+                    UrlHelper.Content(output.Attributes[SrcAttributeName].Value.ToString().Trim());
```
dougbu: If the value is an `HtmlString` (likely because an earlier tag helper set the value), this will cause double-encoding. Need a special case to avoid this.

test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.Image.html
```diff
@@ -9,7 +9,7 @@
 
     <h2>Image Tag Helper Test</h2>
     <!-- Plain image tag -->
-    <img src="/images/red.png" alt="Red block" title="&lt;the title>">
+    <img src="/images/red.png" alt="Red block" title="&lt;the title>" />
```
dougbu: How is this related to your `ImageTagHelper` changes?

test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.AttributesWithBooleanValues.html
```diff
@@ -1,4 +1,3 @@
-
```
dougbu: How is this related to your changes?

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
Eilon: Application root is relative to the *web application*, which has always meant the site root. Not the root of where your source code is (which is never relevant to anything here). This is fine the way it is.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
Eilon: `app` is a perfectly reasonable abbreviation for `application` in a variable name. We don't need (or want) 20+ char variable names.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
Eilon: This doesn't matter.

test/Microsoft.AspNet.Mvc.FunctionalTests/UrlResolutionTest.cs
Eilon: This is fine. What's wrong with this?

src/Microsoft.AspNet.Mvc.Razor/Resources.resx
```diff
@@ -195,4 +195,9 @@
   <data name="RazorPage_InvalidTagHelperIndexerAssignment" xml:space="preserve">
     <value>Unable to perform '{0}' assignment. Tag helper property '{1}.{2}' must not be null.</value>
   </data>
+  <data name="CouldNotResolveApplicationRelativeUrl_TagHelper" xml:space="preserve">
+    <value>Unexpected return value from '{1}.{2}' for URL '{0}'. If you have overridden the '{1}' service, change '{2}' to replace only the '~/' prefix. Otherwise, add the following directive to your Razor page to disable application-relative URL resolution:
```
Eilon: Error messages shouldn't assign blame. Remove all the "you" stuff and make it objective: `"If the blah service has been overridden... blah blah. Add the Foo directive to the Razor page...."`

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
Eilon: Certainly *logically* `string.Join` and `string.Concat` should be the fastest - they have full information ahead of time. Whereas string interpolation and `string.Format` have only partial information (they have to either parser stuff - ewwww - or otherwise predict the consequences of format specifiers).

But anyway, regarding preserving leading or trailing whitespace: meh, not sure that that's important at all. It's super rare to ever see that, and so what if we don't preserve it.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
dougbu: Then we need to change all other use of "application relative" and "application root" in the repo. The term is currently used _only_ for paths relative to the source code e.g. `~/Views/...`. Probably a carry-over from earlier versions where we didn't have the distinct "web root" concept.

test/Microsoft.AspNet.Mvc.FunctionalTests/UrlResolutionTest.cs
dougbu: If you're fine w/ the "app" abbreviation, nothing at all. :ok: :ok_hand: :ok_woman:

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
```diff
@@ -0,0 +1,212 @@
...
+                    }
+
+                    var applicationPath = appRelativeUrl.Substring(0, appRelativeUrl.Length - postTildeSlashUrlValue.Length);
+                    var encodedApplicationPath = HtmlEncoder.HtmlEncode(applicationPath);
+
+                    resolvedUrl = string.Concat(prefix, encodedApplicationPath, postTildeSlashUrlValue, suffix);
+                }
+                else
+                {
+                    resolvedUrl = string.Concat(prefix, appRelativeUrl, suffix);
```
dougbu: Surfacing one of @Eilon's [comments](https://github.com/aspnet/Mvc/pull/2875#discussion-diff-35730321) which is buried now:
> Certainly logically string.Join and string.Concat should be the fastest - they have full information ahead of time. Whereas string interpolation and string.Format have only partial information (they have to either parser stuff - ewwww - or otherwise predict the consequences of format specifiers).

> But anyway, regarding preserving leading or trailing whitespace: meh, not sure that that's important at all. It's super rare to ever see that, and so what if we don't preserve it.

The second point touches on a consistency point I [raised](https://github.com/aspnet/Mvc/pull/2875/files#r36039607) in the next file:
> Why restore surrounding whitespace in the UrlResolutionTagHelper tag helper but drop it in the others?

Let's be consistent and use `Trim(ValidAttributeWhitespaceChars)` everywhere and skip the prefix / suffix restoration whether restoration be concatenation, joining, or formatting.

test/Microsoft.AspNet.Mvc.FunctionalTests/UrlResolutionTest.cs
NTaylorMullen: It's consistent with what we do in all the other functional tests.

test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.Image.html
NTaylorMullen: Now that all `img` tags are `TagHelper`s they can't be written with the void format.

test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.AttributesWithBooleanValues.html
NTaylorMullen: It's not. It's collateral from @ajaybhargavb not running the functional tests with generate baselines prior to checking in. I let him know.

test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.Image.html
dougbu: Again, what does this have to do with your other changes? We've had an `img` tag helper for a while. This needs to be called out in your commit description.

test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.AttributesWithBooleanValues.html
dougbu: If the CI is broken, fix it immediately. Otherwise this seems to be "fallout" from the unnecessary `Trim()` expressions.

test/Microsoft.AspNet.Mvc.FunctionalTests/UrlResolutionTest.cs
dougbu: `<snark ignore="true">Pity there's no markdown for lemmings and their cliff.</snark>`

test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.Image.html
NTaylorMullen: The `ImageTagHelper` only applied to elements with the src and append-file-version attribute combo

dougbu: The `.cshtml` file was incorrect whether or not the `UrlResolutionTagHelper` exists. And the `UrlResolutionTagHelper` doesn't care whether the element is void. So, it's a separate fix that's worth calling out in the commit description.

NTaylorMullen: This looks like part of the test. Hence the comment `Plain image tag`. It's not modified by the normal `ImageTagHelper`

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
Eilon: The views are in the `webroot` so I don't understand your comment. I also doubt an customer reading this wouldn't immediately understand what it means and how to use it.

test/Microsoft.AspNet.Mvc.FunctionalTests/UrlResolutionTest.cs
Eilon: Abbreviations are perfectly acceptable everywhere except for what customers see - namely, public APIs and resources.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
NTaylorMullen: Hu?

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
NTaylorMullen: :+1:

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
Eilon: It's doubtful we need to go make every place in Razor consistent; I just wonder whether all this extra new code is worthwhile.

test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.AttributesWithBooleanValues.html
NTaylorMullen: Not broken. The `Trim` convention used in tests hides the inconsistency with the output file.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
dougbu: `Href` isn't used again for about a 100 lines. Don't want someone to think the assignment is unnecessary.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
dougbu: @Eilon I must be confused but let's chat about this tomorrow.

What I'm seeing:
- Views (`.cshtml` files), precompiled files and a few other things are by default found using a `PhysicalFileProvider` rooted at `IApplicationEnvironment.ApplicationBasePath`.
 - Comments mentioning an "application relative path" are all about the above code.
- `FilePathResult`, the tag helpers for ``, `<link>`, and `<script>`, and other classes dealing with "webroot" use `IHostingEnvironment.WebRootFileProvider` which is also an `IFileProvider` but rooted at the `IHostingEnvironment.WebRootPath`.
- At runtime in the MVC Sample, `IApplicationEnvironment.ApplicationBasePath` is `"C:\\...\\samples\\MvcSample.Web"` while `IHostingEnvironment.WebRootPath` is `"C:\\...\\samples\\MvcSample.Web\\wwwroot"`.

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
Eilon: Ok duh I don't know why I was thinking that.

Eilon: Though I still have to say, either way I'm not [yet] convinced that there's anything actually confusing here. Generally one discussion is about paths, the other is about URLs - it's quite clear which root is used when (unless it's me and it's been a long day in the heat). Both are in some sense app relative, but with respect to different concepts.

src/Microsoft.AspNet.Mvc.Razor/Resources.resx
```diff
@@ -195,4 +195,9 @@
   <data name="RazorPage_InvalidTagHelperIndexerAssignment" xml:space="preserve">
     <value>Unable to perform '{0}' assignment. Tag helper property '{1}.{2}' must not be null.</value>
   </data>
+  <data name="CouldNotResolveApplicationRelativeUrl_TagHelper" xml:space="preserve">
+    <value>Unexpected return value from '{1}.{2}' for URL '{0}'. If the '{1}' service has been overridden, change '{2}' to replace only the '~/' prefix. Otherwise, add the following directive to the Razor page to disable application-relative URL resolution:
```
dougbu: Discussed the [now-hidden thread](https://github.com/aspnet/Mvc/pull/2875#discussion-diff-35729499) with @Eilon offline. Decided to avoid using "application-relative" in a new way.

Here "application-relative URL resolution" -> "URL resolution relative to the application's 'webroot' setting". Might want to pass "webroot" into the resource.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
dougbu: Here "Resolves application relative URLs" -> "Resolves URLs relative to the application's 'webroot' setting" or (perhaps better) "Resolves URLs starting with '~/' (relative to the application's 'webroot' setting)".

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
```diff
@@ -0,0 +1,202 @@
...
+                {
+                    ProcessUrlAttribute(attributeName, output);
+                }
+            }
+
+            // itemid can be present on any HTML element.
+            ProcessUrlAttribute("itemid", output);
+        }
+
+        protected void ProcessUrlAttribute(string attributeName, TagHelperOutput output)
```
dougbu: `protected` methods should have doc comments.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
```diff
@@ -0,0 +1,202 @@
...
+                                resolvedUrl: out resolvedUrl))
+                        {
+                            attribute.Value = new HtmlString(resolvedUrl);
+                        }
+                    }
+                }
+            }
+        }
+
+        protected bool TryResolveUrl(string url, bool tryEncodeApplicationPath, out string resolvedUrl)
```
dougbu: Since parameter names are part of the surface and there's no "try" about it, `tryEncodeApplicationPath` -> `encodeWebRoot`.

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
```diff
@@ -67,12 +74,19 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
             if (AppendVersion)
             {
                 EnsureFileVersionProvider();
+
+                string resolvedUrl;
+                if (TryResolveUrl(Src, tryEncodeApplicationPath: false, resolvedUrl: out resolvedUrl))
+                {
+                    Src = resolvedUrl;
+                }
                 output.Attributes[SrcAttributeName] = _fileVersionProvider.AddFileVersionToPath(Src);
```
dougbu: Existing issue but this doesn't play nicely with earlier tag helpers. Should look similar to other changed tag helpers e.g.
``` c#
    if (Src != null)
    {
        output.CopyHtmlAttribute(SrcAttributeName, context);

        // Resolve any application relative URLs (~/) now so they can be used in comparisons later.
        if (TryResolveUrl(Src, encodeWebRoot: false, resolvedUrl: out resolvedUrl))
        {
            Src = resolvedUrl;
        }
    }

    ProcessUrlAttribute(SrcAttributeName, output);
```

Note I moved `ProcessUrlAttribute` call out of the conditional block. This allows another tag helper to add a `src` attribute and have this one fix it up.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -205,10 +205,20 @@ public class LinkTagHelper : TagHelper
...
             {
                 output.CopyHtmlAttribute(HrefAttributeName, context);
+
+                // Resolve any application relative URLs (~/) now so they can be used in comparisons later.
+                if (TryResolveUrl(Href, tryEncodeApplicationPath: false, resolvedUrl: out resolvedUrl))
+                {
+                    Href = resolvedUrl;
+                }
+
+                ProcessUrlAttribute(HrefAttributeName, output);
```
dougbu: As mentioned in `ImageTagHelper`, useful to call `ProcessUrlAttribute()` unconditionally.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -243,6 +253,15 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
 
             if (mode == Mode.GlobbedHref || mode == Mode.Fallback && !string.IsNullOrEmpty(HrefInclude))
             {
+                if (TryResolveUrl(HrefInclude, tryEncodeApplicationPath: false, resolvedUrl: out resolvedUrl))
```
dougbu: Not correct for any globbed pattern. Should instead call `TryResolveUrl()` within the loops processing what `GlobbingUrlBuilder.BuildUrlList()` returns.

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
dougbu: Oops left out a bit after the `ProcessUrlAttribute()` call.
```
    var srcAttribute = output.Attributes[SrcAttributeName];
    if (AppendVersion && srcAttribute != null && srcAttribute.Value != null)
    {
        var wasEncoded = output.Attributes[SrcAttributeName].Value is HtmlString;
        var value = output.Attributes[SrcAttributeName].Value.ToString();
        output.Attributes[SrcAttributeName].Value= _fileVersionProvider.AddFileVersionToPath(value);
        if (wasEncoded)
        {
            output.Attributes[SrcAttributeName].Value = new HtmlString(output.Attributes[SrcAttributeName]);
        }
    }
```

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
dougbu: In addition the `AppendVersion` handling expects the `href` attribute to be either `null` or a `string`. Should handle the `HtmlString` case similar to `ProcessUrlAttribute()`. Existing issue but worth fixing while you're messing with this.

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
NTaylorMullen: https://github.com/aspnet/Mvc/issues/2902

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
NTaylorMullen: https://github.com/aspnet/Mvc/issues/2902

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
NTaylorMullen: We don't support ~/ today in globbed patterns. https://github.com/aspnet/Mvc/issues/2903

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
dougbu: @NTaylorMullen why open a new issue instead of fixing the code as part of this PR? As you said offline, the implementation would be simpler.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
```diff
@@ -0,0 +1,217 @@
...
+            }
+
+            // itemid can be present on any HTML element.
+            ProcessUrlAttribute("itemid", output);
+        }
+
+        /// <summary>
+        /// Resolves and updates URL values relative to the application's 'webroot' setting (values starting with '~/')
+        /// for <paramref name="output"/>'s <see cref="TagHelperOutput.Attributes"/> whose
+        /// <see cref="TagHelperAttribute.Name"/> is <paramref name="attributeName"/>.
```
dougbu: Wording in class' `<remarks>` are clearer e.g. `Resolves and updates URL values starting with '~/' (relative to the application's 'webroot' setting) for ...`.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
```diff
@@ -0,0 +1,217 @@
...
+        /// </summary>
+        /// <param name="attributeName">The attribute name used to lookup values to resolve.</param>
+        /// <param name="output">The <see cref="TagHelperOutput"/>.</param>
+        protected void ProcessUrlAttribute(string attributeName, TagHelperOutput output)
+        {
+            IEnumerable<TagHelperAttribute> attributes;
+            if (output.Attributes.TryGetAttributes(attributeName, out attributes))
```
dougbu: Just realized this only handles the _first_ attribute named `attributeName` in `output`. That's a loss of coverage in the (unusual) case of duplicate attribute names: Razor would previously have handled all attributes (admittedly, all attributes of any name but this goes too far).

dougbu: Neve rmind, though you were calling `TryGetAttribute`.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
```diff
@@ -0,0 +1,217 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Tries to resolve the given <paramref name="url"/> value relative to the application's 'webroot' setting.
+        /// </summary>
+        /// <param name="url">The URL to resolve.</param>
+        /// <param name="encodeWebRoot"></param>
+        /// <param name="resolvedUrl"></param>
+        /// <returns></returns>
```
dougbu: Fill in the blanks

NTaylorMullen: Ahh rest is on my box, looks like it didn't get included in commit

test/Microsoft.AspNet.Mvc.Razor.Test/TagHelpers/UrlResolutionTagHelperTest.cs
```diff
@@ -0,0 +1,170 @@
...
+                   { new HtmlString("/home/index.html ~/second/wontresolve.html") },
+                   { new HtmlString("~\\home\\index.html") },
+                   { new HtmlString("~\\/home/index.html") },
+                };
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(UnresolvableUrlData))]
+        public void Process_DoesNotResolveNonTildaSlashValues(object url)
```
dougbu: typi: `Tilde`

test/Microsoft.AspNet.Mvc.Razor.Test/TagHelpers/UrlResolutionTagHelperTest.cs
```diff
@@ -0,0 +1,170 @@
...
+                       new HtmlString("~/home/index.html ~/secondValue/index.html"),
+                       new HtmlString("HtmlEncode[[/approot/]]home/index.html ~/secondValue/index.html")
+                   },
+                };
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(ResolvableUrlData))]
+        public void Process_ResolvesTildaSlashValues(object url, object expectedHref)
```
dougbu: typi: `Tilde`

dougbu: typi: `Tilde`

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
NTaylorMullen: Did it half way, ended up not being simpler.

dougbu:
Fix #2903 in this PR. As discussed in a [hidden thread](https://github.com/aspnet/Mvc/pull/2875#discussion-diff-36106349), currently doing extra work to preserve a bug.

dougbu:
:watch:

NTaylorMullen:
Updated.

Eilon:
Let's not add more to this PR at this point. It's fine to treat other items as bugs.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
```diff
@@ -0,0 +1,218 @@
...
+        }
+
+        /// <summary>
+        /// Tries to resolve the given <paramref name="url"/> value relative to the application's 'webroot' setting.
+        /// </summary>
+        /// <param name="url">The URL to resolve.</param>
+        /// <param name="encodeWebRoot">If <c>true</c>, will encode the path to the application's root defined by the
+        /// 'webroot' setting.</param>
+        /// <param name="resolvedUrl">The URL value relative to the application's 'webroot' setting.</param>
+        /// <returns><c>true</c> if the <paramref name="url"/> could be resolved; <c>false</c> otherwise.</returns>
```
dougbu: Sorry for commenting on a commit (amateur move). My comments were on this set of doc comments.

dougbu:
:shipit: w/ last couple of doc comment fixes

PR #2878, opened 2015-07-29
Removing MvcViewEngineDescriptor and switching to use OptionsSetup to

### LINKED ISSUE #2269: Simplify MvcOptions

pranavkm:
setup RazorViewEngine.

Fixes #2269


pranavkm:
Built against volatile to pick up DNX renames.

src/Microsoft.AspNet.Mvc.Razor/MvcRazorMvcViewOptionsSetup.cs
```diff
@@ -1,21 +1,39 @@
...
+        /// Configures <paramref name="options"/> to use <see cref="RazorViewEngine"/>.
+        /// </summary>
+        /// <param name="serviceProvider">The application's <see cref="IServiceProvider"/>.</param>
+        /// <param name="options">The <see cref="MvcViewOptions"/> to configure.</param>
+        public static void ConfigureMvc(
+            [NotNull] IServiceProvider serviceProvider,
+            [NotNull] MvcViewOptions options)
         {
-            options.ViewEngines.Add(typeof(RazorViewEngine));
+            var razorViewEngine = ActivatorUtilities.CreateInstance<RazorViewEngine>(serviceProvider);
```
rynowak: Why not just register RazorViewEngine and then resolve it?

src/Microsoft.AspNet.Mvc.ViewFeatures/DependencyInjection/MvcViewFeaturesMvcBuilderExtensions.cs
```diff
@@ -55,9 +55,7 @@ internal static void AddViewServices(IServiceCollection services)
             //
             // View Engine and related infrastructure
             //
-            // The provider is inexpensive to initialize and provides ViewEngines that may require request
-            // specific services.
-            services.TryAddScoped<ICompositeViewEngine, CompositeViewEngine>();
+            services.TryAddSingleton<ICompositeViewEngine, CompositeViewEngine>();
```
rynowak: Huzzah

PR #2879, opened 2015-07-29
Build break: React to aspnet/Razor@e722f90

dougbu:
- aspnet/Razor@e722f90 removed `HashCodeCombiner` from `public` surface of Razor assembly
- not actually needed since `IReadOnlyTagHelperAttribute` implementations provide a solid `GetHashCode()`


dougbu:
FYI only since this fixes a build break

dougbu:
bb158ec

PR #2882, opened 2015-07-29
Align `TagHelperOutputExtensionsTest.CaseSensitiveTagHelperAttributeComparer` with Razor.Runtime version

dougbu:
- was missing `Minimized` comparison


dougbu:
FYI only; quick fix that @NTaylorMullen and I discussed offline

dougbu:
260ac29

PR #2886, opened 2015-07-30
Fix #2837 - Make it easier to get ParameterInfo/PropertyInfo

### LINKED ISSUE #2837: ReflectedParameterDescriptor equivalent
In MVC3, we used ReflectedParameterDescriptor (from ActionDescriptor.GetParameters()) in order to determine whether the user supplied all of the necessary arguments, and display a detailed error page if not. I see that something similar has been done in WebApiCompatShim, but it seems to happen in a stage of the pipeline that cannot modify the response. Is there a way to get this information in perhaps something like an ActionFilter (i.e. it still gets forwarded to a particular action, but will respond with a "param is required" message)?

rynowak:
For a typical configuration, it's now possible to cast a parameter
descriptor to ControllerParameterDescriptor or
ControllerBoundPropertyDescriptor to access the corresponding reflection
type.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Abstractions/ParameterDescriptor.cs
+++ b/src/Microsoft.AspNet.Mvc.Abstractions/ParameterDescriptor.cs
@@ -3,7 +3,6 @@
 
 using System;
 using Microsoft.AspNet.Mvc.ModelBinding;
-using Microsoft.AspNet.Mvc.ModelBinding.Metadata;
 
 namespace Microsoft.AspNet.Mvc
 {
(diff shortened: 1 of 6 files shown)


rynowak:
We're deliberately **not** changing `ActionDescriptor` to expose these more specialized types. The top level signatures are supposed to be abstract and free of details like reflection.

I took a look, and didn't see any opportunities to benefit from using these new properties in the existing code without a big perf-focused rewrite. This is here mostly for users.

pranavkm:
> This is here mostly for users.

I was trying to figure out what the utility for this change was.
:shipit:

rynowak:
Yeah, see the linked bug for an example. I'd think a fair number of users have code like this. It's a nice convenience and eventually we'll get to doing something with perf that will likely benefit from this change.

PR #2899, opened 2015-08-03
Making DefaultDisplayTemplates and DefaultEditorTemplates use IHtmlContent.

sornaks:
- Making TagBuilder's InnerHtml an IHtmlContent.
- Delay encoding until the content is written.
- Moving BufferedHtmlContent to Common cos it is used in aspnet/Razor also.
- This reduces the space used by around 13%.


test/Microsoft.AspNet.Mvc.TagHelpers.Test/InputTagHelperTest.cs
```diff
@@ -820,9 +823,9 @@ public static TheoryData MultiAttributeCheckBoxData
 
             Assert.True(output.SelfClosing);
             Assert.Equal(expectedAttributes, output.Attributes);
-            Assert.Empty(output.PreContent);
-            Assert.Equal(new[] { string.Empty }, output.Content);
-            Assert.Empty(output.PostContent);
+            Assert.Empty(output.PreContent.ToString());
+            Assert.Equal(string.Empty, output.Content.ToString());
+            Assert.Empty(output.PostContent.ToString());
```
dougbu: Have avoided using `ToString()` with `TagHelperContent`. Don't start now.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/TestableHtmlGenerator.cs
```diff
@@ -82,7 +83,7 @@ public override HtmlString GenerateAntiforgery(ViewContext viewContext)
                 },
             };
 
-            return tagBuilder.ToHtmlString(TagRenderMode.SelfClosing);
+            return new HtmlString(tagBuilder.ToHtmlContent(TagRenderMode.SelfClosing).ToString());
```
dougbu: This looks incorrect. Why not return an `IHtmlContent` written using the `CommonTestEncoder`?

test/Microsoft.AspNet.Mvc.TagHelpers.Test/TextAreaTagHelperTest.cs
```diff
@@ -46,32 +47,32 @@ public class TextAreaTagHelperTest
                 {
                     { null, typeof(Model), null,
                         new NameAndId("Text", "Text"),
-                        Environment.NewLine + "HtmlEncode[[]]" },
+                        Environment.NewLine },
```
dougbu: Should have absolutely no encoding differences from previous tests. This indicates an unexpected behaviour change.

src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
```diff
@@ -249,12 +249,12 @@ private void GenerateCheckBox(ModelExplorer modelExplorer, TagHelperOutput outpu
                 output.Attributes.Clear();
                 output.TagName = null;
 
-                output.Content.Append(tagBuilder.ToString(TagRenderMode.SelfClosing));
+                output.Content.Append(tagBuilder.ToHtmlContent(TagRenderMode.SelfClosing));
```
rynowak: This doesn't go far enough. TagBuilder should implement `IHtmlContent`. Or is that planned for another iteration?

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultDisplayTemplates.cs
```diff
@@ -286,12 +287,13 @@ private static bool ShouldShow(ModelExplorer modelExplorer, TemplateInfo templat
                 !templateInfo.Visited(modelExplorer);
         }
 
-        public static string StringTemplate(IHtmlHelper htmlHelper)
+        public static IHtmlContent StringTemplate(IHtmlHelper htmlHelper)
         {
-            return htmlHelper.Encode(htmlHelper.ViewData.TemplateInfo.FormattedModelValue);
+            var value = htmlHelper.ViewData.TemplateInfo.FormattedModelValue;
+            return new StringHtmlContent(value == null ? string.Empty : value.ToString());
```
rynowak: Do we need both `StringHtmlContent` and `HtmlString`?

src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
dougbu: @rynowak if `TagBuilder` implemented `IHtmlContent`, would have to decide on the `TagRenderMode` it would use when called that way and would still need `ToHtmlContent` for the other `TagRenderMode` choices. Probably better to leave as @sornaks has it here.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultEditorTemplates.cs
```diff
@@ -139,19 +136,19 @@ public static string DecimalTemplate(IHtmlHelper htmlHelper)
...
         {
             var viewData = htmlHelper.ViewData;
             var model = viewData.Model;
 
-            string result;
+            var result = new BufferedHtmlContent();
             if (viewData.ModelMetadata.HideSurroundingHtml)
             {
-                result = string.Empty;
+                result.Append(HtmlString.Empty);
```
dougbu: This line isn't useful. Change to
``` c#
  if (!viewData.ModelMetadata.HideSurroundingHtml)
  {
    result.Append(DefaultDisplayTemplates.StringTemplate(htmlHelper));
  }
```

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultDisplayTemplates.cs
dougbu: I suggest removing the `StringHtmlContent` option to treat content as already-encoded. It's a well-known MVC idiom that `HtmlString` values do not need further HTML encoding.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultEditorTemplates.cs
```diff
@@ -219,38 +216,37 @@ public static string HiddenInputTemplate(IHtmlHelper htmlHelper)
...
         }
 
-        public static string ObjectTemplate(IHtmlHelper htmlHelper)
+        public static IHtmlContent ObjectTemplate(IHtmlHelper htmlHelper)
         {
             var viewData = htmlHelper.ViewData;
             var templateInfo = viewData.TemplateInfo;
             var modelExplorer = viewData.ModelExplorer;
-            var builder = new StringBuilder();
+            var content = new BufferedHtmlContent();
```
dougbu: Suggest moving this declaration down closer to its first use.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultEditorTemplates.cs
```diff
@@ -264,27 +260,27 @@ public static string ObjectTemplate(IHtmlHelper htmlHelper)
...
-                            .ToString();
-                    if (!string.IsNullOrEmpty(label))
+                        htmlAttributes: null);
+                    if (!string.IsNullOrEmpty(label.ToString()))
                     {
                         divTag.AddCssClass("editor-label");
                         divTag.InnerHtml = label; // already escaped
-                        builder.AppendLine(divTag.ToString(TagRenderMode.Normal));
+                        content.Append(divTag.ToHtmlContent(TagRenderMode.Normal));
+                        content.Append(Environment.NewLine);
```
dougbu: Why not add `AppendLine()` to `IHtmlContent`?

src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
rynowak: Or just make `TagRenderMode` a property on the class.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -689,24 +694,26 @@ public virtual HtmlString GenerateAntiforgery([NotNull] ViewContext viewContext)
...
                         listItem.SetInnerText(errorText);
-                        htmlSummary.AppendLine(listItem.ToString(TagRenderMode.Normal));
+                        htmlSummary.Append(listItem.ToHtmlContent(TagRenderMode.Normal));
+                        htmlSummary.Append(Environment.NewLine);
                     }
                 }
             }
 
-            if (htmlSummary.Length == 0)
+            if (htmlSummary.ToString().Length == 0)
```
dougbu: Suggest we need something like `TagHelperOutput`s `IsNullOrEmpty` in either `IHtmlContent` or `BufferedHtmlContent`. This less about performance (though it's not great to create a `string` only to immediately throw it away) than straightforwardness.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -898,11 +908,11 @@ internal static string EvalString(ViewContext viewContext, string key, string fo
         /// <remarks>
         /// Not used directly in HtmlHelper. Exposed for use in DefaultDisplayTemplates.
         /// </remarks>
-        internal static TagBuilder GenerateOption(SelectListItem item, string encodedText, IHtmlEncoder htmlEncoder)
+        internal static TagBuilder GenerateOption(SelectListItem item, IHtmlContent encodedContent)
```
dougbu: Is the `encodedContent` always encoded? Does this method or the `TagBuilder` require the content to be encoded? If not, name parameter `content`.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/TagBuilder.cs
```diff
@@ -6,41 +6,23 @@
...
-        {
-        }
-
-        public TagBuilder(string tagName, [NotNull] IHtmlEncoder htmlEncoder)
+        public TagBuilder([NotNull] string tagName)
         {
-            if (string.IsNullOrEmpty(tagName))
-            {
-                throw new ArgumentException(Resources.ArgumentCannotBeNullOrEmpty, nameof(tagName));
-            }
```
dougbu: Unintended behaviour change. Do not change from checking `IsNullOrEmpty()` to using `[NotNull]`.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/HtmlString.cs
```diff
@@ -1,55 +1,31 @@
...
-using Microsoft.AspNet.Html.Abstractions;
-using Microsoft.Framework.WebEncoders;
-
 namespace Microsoft.AspNet.Mvc.Rendering
 {
     /// <summary>
     /// String content which knows how to write itself.
     /// </summary>
-    public class HtmlString : IHtmlContent
+    public class HtmlString : StringHtmlContent
```
dougbu: Inheriting from `StringHtmlContent` makes that class more complex than necessary. `HtmlString` is the only user of `encodeOnWrite: false`. Implement `IHtmlContent` here and remove `encodeOnWrite` from `StringHtmlContent`.

src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
sornaks: As discussed offline, I'll do this change as a followup.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/StringHtmlContent.cs
```diff
@@ -0,0 +1,62 @@
...
+
+using System.IO;
+using Microsoft.AspNet.Html.Abstractions;
+using Microsoft.Framework.Internal;
+using Microsoft.Framework.WebEncoders;
+
+namespace Microsoft.AspNet.Mvc.Rendering
+{
+    /// <summary>
+    /// String content which gets enoded when written.
```
dougbu: tpo: "encoded"

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/StringHtmlContent.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// </summary>
+    public class StringHtmlContent : IHtmlContent
+    {
+        private bool _encodeOnWrite;
+        private string _text;
+
+        /// <summary>
+        /// Creates a new instance of <see cref="StringHtmlContent"/>
+        /// </summary>
+        /// <param name="text"><c>string</c>to initialize <see cref="StringHtmlContent"/>.</param>
```
dougbu: `<see cref="string"/> to be HTML encoded when <see cref="WriteTo"/> is called.`

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/StringHtmlContent.cs
```diff
@@ -0,0 +1,62 @@
...
+                encoder.HtmlEncode(_text, writer);
+            }
+            else
+            {
+                writer.Write(_text);
+            }
+        }
+
+        /// <inheritdoc />
+        public override string ToString()
```
dougbu: Don't create an `HtmlEncoder()` instance for debugging purposes.
- Do the simple thing and return `_text`.
- Make sure none of our product code calls this method. Should be used only when debugging.
- Have only one or two tests that ensure this method returns the expected (raw) `string`.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/InputTagHelperTest.cs
dougbu: Also now checking something different than before. An empty `TagHelperContent` instance is similar to one that contains only a single `string.Empty` instance. But this method was confirming the _exact_ behaviour e.g. that the tag helper did not interact with `PreContent` at all. (Checking `Assert.False(output.PreContent.IsModified)` is closer to the existing code than this. But will be able to restore the original code once `TagHelperContent` again implements `IEnumerable<string>`.)

test/Microsoft.AspNet.Mvc.TagHelpers.Test/TestableHtmlGenerator.cs
dougbu: Better-yet, change return type and return `tagBuilder.ToHtmlContent(...)` directly.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/DefaultDisplayTemplatesTest.cs
```diff
@@ -268,7 +269,7 @@ public void Display_FindsViewDataMember()
             var result = helper.Display("Property1");
 
             // Assert
-            Assert.Equal("HtmlEncode[[ViewData string]]", result.ToString());
+            Assert.Equal("HtmlEncode[[ViewData string]]", HtmlContentUtilities.HtmlContentToString(result).ToString());
```
dougbu: Don't call `ToString()` on a `string`. Any place you've chained `HtmlContentToString(blah).ToString()` is confusing and will make readers wonder about the `HtmlContentToString()` return type.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/TagBuilder.cs
```diff
@@ -195,56 +177,44 @@ public void MergeAttribute(string key, string value, bool replaceExisting)
...
         public void SetInnerText(string innerText)
         {
-            InnerHtml = _htmlEncoder.HtmlEncode(innerText);
-        }
-
-        public HtmlString ToHtmlString(TagRenderMode renderMode)
-        {
-            return new HtmlString(ToString(renderMode));
+            InnerHtml = new StringHtmlContent(innerText);
         }
```
dougbu: After changing the `InnerHtml` type, this method seems redundant. It previously existed mostly to allow adding both inner text to be encoded (this method) and inner text that does not need to be encoded (assigning to `InnerHtml`).

Wondering about restoring `string InnerHtml` but changing the class' implementation to include a `bool` field that determines how `InnerHtml` is handled when `WriteTo()` is called. This `StringHtmlContent` instance wouldn't be necessary and the public surface of `TagBuilder` would change less (with no perf cost).

rynowak: I still don't understand why we need `StringHtmlContent` at all, let's just get rid of it.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/StringHtmlContentTest.cs
```diff
@@ -0,0 +1,71 @@
...
+    public class StringHtmlContentTest
+    {
+        [Fact]
+        public void CreateStringHtmlContent()
+        {
+            // Arrange & Act
+            var content = new StringHtmlContent("Hello World");
+
+            // Assert
+            var result = Assert.IsType<StringHtmlContent>(content);
```
dougbu: What is this testing? You just instantiated the `StringHtmlContent` instance so of course that's its type.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/TagBuilder.cs
dougbu: Don't we need it to indicate an entry in `BufferedHtmlContent` (or similar) needs to be encoded when `WriteTo()` is called?

dougbu: That said, we don't need a `StringHtmlContent` _instance_ here :smiley_cat:

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
Eilon: Yeah even though it's not the design I like, it does seem better for perf to change the design to have this implement the interface instead of allocating new objects.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultHtmlGenerator.cs
sornaks: No, I'll call it `content`

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultHtmlGenerator.cs
sornaks: Actually, I could work around this by setting a flag when `htmlSummary.Append()` is called with the `listItem`.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/TagBuilder.cs
sornaks: That would mean `ToString`ing wherever `InnerHtml` is used. Example - <a href="https://github.com/aspnet/Mvc/pull/2899/files#diff-0e83439eeed8ffbde4eceb0db1f6e541L721">Here</a> `BufferedHtmlContent` is set as the `TagBuilder.InnerHtml`. So `SetInnerText` acts like a shortcut when you have un-encoded text to be set as the content for the TagBuilder.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/TextAreaTagHelperTest.cs
sornaks: This encoding changed because <a href="https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.Framework.WebEncoders.Core/EncoderExtensions.cs#L28">WebEncoders' extension method</a> skips null <b>and</b> empty strings. This was not the case before because we encoded the string separately using <a href="https://github.com/aspnet/Testing/blob/dev/src/Microsoft.Framework.WebEncoders.Testing/CommonTestEncoder.cs#L88">CommonTestEncoder</a> (which returns a `HtmlEncode[[]]` when encoding empty strings) and we then write this to the writer.

sornaks:
Updated..

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -898,11 +907,11 @@ internal static string EvalString(ViewContext viewContext, string key, string fo
         /// <remarks>
         /// Not used directly in HtmlHelper. Exposed for use in DefaultDisplayTemplates.
         /// </remarks>
-        internal static TagBuilder GenerateOption(SelectListItem item, string encodedText, IHtmlEncoder htmlEncoder)
+        internal static TagBuilder GenerateOption(SelectListItem item, IHtmlContent content)
```
dougbu: This method is never called with anything except a `StringHtmlContent` instance. Why not remove the `htmlEncoder` parameter, rename `encodedText` -> `text`, and create the `StringHtmlContent` here?

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultDisplayTemplates.cs
dougbu: Suggest moving the ternary outside the constructor i.e. `return value == null ? HtmlString.Empty : new StringHtmlContent(value.ToString());`

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -1272,19 +1281,21 @@ private static string GetInputTypeString(InputType inputType)
...
             {
-                listItemBuilder.AppendLine(GenerateOption(new SelectListItem()
+                listItemBuilder.Append(GenerateOption(new SelectListItem()
                 {
                     Text = optionLabel,
                     Value = string.Empty,
                     Selected = false,
                 }));
+
+                listItemBuilder.Append(Environment.NewLine);
```
dougbu: You've got `AppendLine()` now.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/TagBuilder.cs
```diff
@@ -31,16 +23,11 @@ public TagBuilder(string tagName, [NotNull] IHtmlEncoder htmlEncoder)
...
 
         public IDictionary<string, string> Attributes { get; private set; }
 
-        public string InnerHtml
-        {
-            get { return _innerHtml ?? string.Empty; }
-            set { _innerHtml = value; }
-        }
+        public IHtmlContent InnerHtml { get; set; } = HtmlString.Empty;
```
dougbu: Previously could be set to `null` without causing an NRE in later code. Probably should remain that defensive.

Other option is to add `[param: NotNull]` above the `set`. Minor compat break but no NRE.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/HtmlString.cs
```diff
@@ -12,44 +13,32 @@ namespace Microsoft.AspNet.Mvc.Rendering
     /// </summary>
     public class HtmlString : IHtmlContent
     {
-        private static readonly HtmlString _empty = new HtmlString(string.Empty);
-        private readonly string _input;
+        private string _text;
```
dougbu: Why was this field renamed? Why is it no longer `readonly`?

test/Microsoft.AspNet.Mvc.TagHelpers.Test/TextAreaTagHelperTest.cs
dougbu: :ok:

test/Microsoft.AspNet.Mvc.TestCommon/project.json
```diff
@@ -2,5 +2,6 @@
     "version": "6.0.0-*",
     "shared": "**/*.cs",
     "dependencies": {
+        "Microsoft.Framework.WebEncoders.Testing": "1.0.0-*"
```
dougbu: We've previously not placed any dependencies in shared code `project.json` files. Had issues with resolution of these dependencies in the past. But more importantly, the semantics are a bit odd since no compilation happens in this project.

davidfowl: Yes this is wrong. Shared sources should never have dependencies.

dougbu:
:watch:

sornaks:
Updated..

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -153,7 +154,7 @@ public virtual HtmlString GenerateAntiforgery([NotNull] ViewContext viewContext)
             ModelExplorer modelExplorer,
             string expression)
         {
-            var tagBuilder = new TagBuilder("input", _htmlEncoder);
```
dougbu: Are the `Encode()` methods still used anywhere? If not, can get rid of the `_htmlEncoder` field and related constructor parameter entirely. Fortunately there are very few places where we invoke that constructor.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/HtmlHelper.cs
```diff
@@ -755,7 +753,7 @@ protected virtual string GenerateDisplayText(ModelExplorer modelExplorer)
                 htmlAttributes);
             if (tagBuilder != null)
             {
-                ViewContext.Writer.Write(tagBuilder.ToString(TagRenderMode.StartTag));
+                tagBuilder.ToHtmlContent(TagRenderMode.StartTag).WriteTo(ViewContext.Writer, HtmlEncoder);
```
dougbu: Bet this and line 798 are the only remaining uses of the `HtmlEncoder` property. If so, remove it from the `IHtmlHelper` interface and make it a `private _htmlEncoder` field instead. (Nothing similar to this property exists in MVC 5 and we shouldn't make it easy to encode outside `IHtmlContent.WriteTo()`.)

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/HtmlString.cs
```diff
@@ -12,40 +13,28 @@ namespace Microsoft.AspNet.Mvc.Rendering
...
-            _input = input;
-        }
+        public static readonly HtmlString Empty = new HtmlString(string.Empty);
 
         /// <summary>
-        /// Returns an <see cref="HtmlString"/> with empty content.
+        /// Creates a new instance of <see cref="HtmlString"/>.
         /// </summary>
-        public static HtmlString Empty
+        /// <param name="input"><c>string</c>to initialize <see cref="HtmlString"/>.</param>
```
dougbu: existing code but add space after `</c>`

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/HtmlString.cs
```diff
@@ -12,40 +13,28 @@ namespace Microsoft.AspNet.Mvc.Rendering
...
-        /// <param name="writer">The <see cref="TextWriter"/> to write contents to.</param>
-        /// <param name="encoder">The <see cref="IHtmlEncoder"/> with which the output must be encoded.</param>
-        public void WriteTo(TextWriter writer, IHtmlEncoder encoder)
+        /// <inheritdoc />
+        public void WriteTo([NotNull] TextWriter writer, [NotNull] IHtmlEncoder encoder)
         {
             writer.Write(_input);
         }
-
+
```
dougbu: Get thee the trailing whitespace visualizer (which also automagically nukes trailing whitespace).

dougbu:
:shipit:

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultHtmlGenerator.cs
sornaks: Yes it is used in a couple of other places.

sornaks:
Checked in. Thanks. https://github.com/aspnet/Mvc/commit/b6a109e2a3f6fac94a3cbec0d16ea1adf526e1d8

Issue #2901, opened 2015-08-03
ImageTagHelper showcases view-context as available HTML attribute.


NTaylorMullen:
Looks like this was missed when we introduced `[HtmlAttributeNotBound]`. 

![image](https://cloud.githubusercontent.com/assets/2008729/9045156/2d549f06-39d6-11e5-9c6d-0a57a8dd7f45.png)

NTaylorMullen:
6d228a62dc026544506f4bbf80f0d5d5457fe960

PR #2904, opened 2015-08-03
Added missing `HtmlAttributeNotBound` to `ViewContext` property.

NTaylorMullen:
- Did not add a test to validate the behavior. None of the other `TagHelper`s validate that a property is not visible/bound.

#2901

### DIFF
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
@@ -56,6 +56,7 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
 
         protected IHostingEnvironment HostingEnvironment { get; }
 
+        [HtmlAttributeNotBound]
         [ViewContext]
         public ViewContext ViewContext { get; set; }
 


src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
```diff
@@ -56,6 +56,7 @@ public ImageTagHelper(IHostingEnvironment hostingEnvironment, IMemoryCache cache
 
         protected IHostingEnvironment HostingEnvironment { get; }
 
+        [HtmlAttributeNotBound]
```
dougbu: Wot no tests?!?!!

dougbu: :smiley_cat:

dougbu:
:shipit:

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
NTaylorMullen: :smiley:

PR #2913, opened 2015-08-04
React to options (Order nuked)

HaoK:
I'm not sure how much test coverage there is around the whole MVC options ordering.  That got nuked, so I rejiggered the AddMvc services to match roughly the old order (which is perhaps why tests aren't failing?)

Someone should review this carefully who has more context about what MVC expects for the ordering...

cc @rynowak @divega


Mvc.NoFun.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
```
HaoK: I'll revert this before merge

global.json
```diff
@@ -1,3 +1,3 @@
 {
-    "projects": ["src", "test/WebSites", "samples"]
+    "projects": ["src", "test/WebSites", "samples", "C:\\Github\\Security\\src", "C:\\Github\\HttpAbstractions\\src",  "C:\\Github\\Options\\src",  "C:\\Github\\DataProtection\\src"]
```
HaoK: I'll revert this before merge

rynowak:
@HaoK - before diving into specifics... our use of order is really only important to ensure the following:

1. Options classes can be defined in an assembly that doesn't have access to the implementations that need to be added by default (Validation is a good example).
2. Any framework-provided options-setups which add 'default options' run with high priority (before the user's options setup)
3. The user-provided options-setups **always** see the 'default options' or in the case of `AddMvcCore()` always see the options that they selected 

**What's the equivalent pattern if order is being removed?**

HaoK:
The order is now simply the order returned by the container now.  

If the default options need to be added first, Then AddMvc needs to be first in startup.cs before they do any of their own configuration

Can you move any default logic into the constructors instead?

rynowak:
> If the default options need to be added first, Then AddMvc needs to be first in startup.cs before they do any of their own configuration

This is going to do wierdness with `AddMvcCore()`. It's not a big problem, but something we'll have to play with.

```
builder
    .AddMvcCore(o => o....); // This will see a virtually empty options
    .AddDataAnnotations(); // Now the validators are present
    .ConfigureMvc(o => ...); // Now you can see the validators
```

It's not as bad as a I expected... you basically get what you ask for, but the old way was a bit more idiot-proof. /cc @davidfowl 

> Can you move any default logic into the constructors instead?

Rarely an option because of 1). that I called out above

HaoK:
@rynowak does this look good?  Trying to get everything into beta 7 if I can this week

rynowak:
I still have the same questions and concerns as before. Let's run through the experience with @davidfowl

HaoK:
@rynowak Happy with this change now?

rynowak:
Yep. Go for it. If you have issues rebasing I'll be around to help.

test/Microsoft.AspNet.Mvc.Localization.Test/DependencyInjection/MvcLocalizationServiceCollectionExtensionsTest.cs
```diff
@@ -116,11 +116,13 @@ public void AddCustomLocalizers_AfterMvcLocalization_AddsNeededServices()
...
             Assert.Equal(ServiceLifetime.Singleton, services[0].Lifetime);
-            Assert.Equal(0, ((IConfigureOptions<RazorViewEngineOptions>)services[0].ImplementationInstance).Order);
+            // REVIEW: What should this be replaced with?
+            //Assert.Equal(0, ((IConfigureOptions<RazorViewEngineOptions>)services[0].ImplementationInstance).Order);
 
             Assert.Equal(typeof(IConfigureOptions<RazorViewEngineOptions>), services[1].ServiceType);
             Assert.Equal(ServiceLifetime.Singleton, services[1].Lifetime);
-            Assert.Equal(-1000, ((IConfigureOptions<RazorViewEngineOptions>)services[1].ImplementationInstance).Order);
+            // REVIEW: What should this be replaced with?
+            //Assert.Equal(-1000, ((IConfigureOptions<RazorViewEngineOptions>)services[1].ImplementationInstance).Order);
```
dougbu: Any way to confirm the previously-earlier (`-1000`) service implementation is now at `services[0]`?

rynowak:
@HaoK  was this merged?

HaoK:
Not yet identity is blocking the options pr still.

HaoK:
rebased

rynowak:
@HaoK is this still pending?

HaoK:
Will get this in this week. Rebasing repos all over :/

HaoK:
3ebdcc5f6f9c4398a73f6d26f26a7c6a77433361

PR #2915, opened 2015-08-04
Support comma separated globbed pattern in Script and Link tag helpers

ajaybhargavb:
Issue - #2841 
@dougbu @NTaylorMullen

### DIFF
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
@@ -20,6 +20,10 @@ namespace Microsoft.AspNet.Mvc.TagHelpers.Internal
     {
         private static readonly char[] PatternSeparator = new[] { ',' };
 
+        // Valid whitespace characters defined by the HTML5 spec.
+        private static readonly char[] ValidAttributeWhitespaceChars =
+            new[] { '\t', '\n', '\u000C', '\r', ' ' };
+
         private static readonly PathComparer DefaultPathComparer = new PathComparer();
 
         private readonly FileProviderGlobbingDirectory _baseGlobbingDirectory;
@@ -125,11 +129,11 @@ namespace Microsoft.AspNet.Mvc.TagHelpers.Internal
         {
             var matcher = MatcherBuilder != null ? MatcherBuilder() : new Matcher();
 
-            matcher.AddIncludePatterns(includePatterns.Select(pattern => TrimLeadingSlash(pattern)));
+            matcher.AddIncludePatterns(includePatterns.Select(pattern => TrimLeadingTildeSlash(pattern)));
 
             if (excludePatterns != null)
             {
-                matcher.AddExcludePatterns(excludePatterns.Select(pattern => TrimLeadingSlash(pattern)));
+                matcher.AddExcludePatterns(excludePatterns.Select(pattern => TrimLeadingTildeSlash(pattern)));
             }
 
             var matches = matcher.Execute(_baseGlobbingDirectory);
@@ -210,11 +214,15 @@ namespace Microsoft.AspNet.Mvc.TagHelpers.Internal
             }
         }
 
-        private static string TrimLeadingSlash(string value)
+        private static string TrimLeadingTildeSlash(string value)
         {
-            var result = value;
+            var result = value.Trim(ValidAttributeWhitespaceChars);
 
-            if (result.StartsWith("/", StringComparison.Ordinal) ||
+            if (result.StartsWith("~/", StringComparison.Ordinal))
+            {
+                result = result.Substring(2);
+            }
+            else if (result.StartsWith("/", StringComparison.Ordinal) ||
                 result.StartsWith("\\", StringComparison.Ordinal))
             {
                 // Trim the leading slash as the matcher runs from the provided root only anyway
--- a/test/Microsoft.AspNet.Mvc.TagHelpers.Test/Internal/GlobbingUrlBuilderTest.cs
+++ b/test/Microsoft.AspNet.Mvc.TagHelpers.Test/Internal/GlobbingUrlBuilderTest.cs
@@ -86,7 +86,6 @@ namespace Microsoft.AspNet.Mvc.TagHelpers.Internal
                     {
                         /* staticUrl */ "/site.css",
                         /* dirStructure */ new FileNode(null, new [] {
-                            
                             new FileNode("A", new [] {
                                 new FileNode("c.css"),
                                 new FileNode("d.css")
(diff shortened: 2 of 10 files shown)


src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
```diff
@@ -20,6 +20,10 @@ public class GlobbingUrlBuilder
     {
         private static readonly char[] PatternSeparator = new[] { ',' };
 
+        // Valid whitespace characters defined by the HTML5 spec.
+        private static readonly char[] ValidAttributeWhitespaceChars =
+            new[] { '\t', '\n', '\u000C', '\r', ' ' };
```
dougbu: Any way to share this with the definition in MVC.Razor's `UrlResolutionTagHelper`?

test/WebSites/HtmlGenerationWebSite/Views/HtmlGeneration_Home/Link.cshtml
```diff
@@ -12,6 +12,9 @@
     <!-- Globbed link tag with existing file -->
     <link asp-href-include="**/site.css" rel="stylesheet" title="&lt;the title>" />
 
+    <!-- Globbed link tag with comma separated pattern -->
+    <link asp-href-include="~/*.css, ~/sub/*.css" rel="stylesheet" title="&lt;the title" />
```
dougbu: (Not for fixing in this PR but) does `<link asp-href-include="~/joe.css, /joe.css, \\joe.css, joe.css" rel="stylesheet" title="&lt;the title" />` generate four identical `<link/>` elements?

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
ajaybhargavb: We could make it `protected` and pass it in from Link and Script tag helper when we are creating the `GlobbedUrlBuilder` instance but its does not seem like a good approach.

test/WebSites/HtmlGenerationWebSite/Views/HtmlGeneration_Home/Link.cshtml
dougbu: @Eilon we should support `~/` prefixes but is that something we should encourage i.e. highlight in this functional test web site?
@ajaybhargavb our coverage of comma-separated patterns is limited. Here or in `GlobbingUrlBuilderTest`, need to cover comma-separated patterns that do _not_ use the `~/` prefix.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
IDisposable: Why not a public readonly value somewhere, this is something many folks might need.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
```diff
@@ -210,11 +214,17 @@ public int Compare(string x, string y)
             }
         }
 
-        private static string TrimLeadingSlash(string value)
+        private static string TrimLeadingTildeSlash(string value)
         {
             var result = value;
 
-            if (result.StartsWith("/", StringComparison.Ordinal) ||
+            result = result.Trim(ValidAttributeWhitespaceChars);
```
IDisposable: move up to line 215 thus
```
var result = value.Trim(ValidAttributeWhitespaceChars);
```

ajaybhargavb: :+1:

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
dougbu: @IDisposable yeah; could e.g. be `public static readonly` in `UrlResolutionTagHelper`. We don't have many *Constants classes and a reference from here to `UrlResolutionTagHelper` isn't __too__ ugly.

test/WebSites/HtmlGenerationWebSite/Views/HtmlGeneration_Home/Link.cshtml
ajaybhargavb: >  does `<link asp-href-include="~/joe.css, /joe.css, \\joe.css, joe.css" rel="stylesheet" title="&lt;the title" />` generate four identical <link/> elements?

No. The `Matcher` returns only one match.

dougbu: Well then, since it works, include it in a test :smiley_cat: -- preferably `GlobbingUrlBuilderTest` since the repetition is certainly not something we'd like to see in a mainline scenario.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/Internal/GlobbingUrlBuilderTest.cs
```diff
@@ -299,10 +298,33 @@ public void CachesMatchResults()
...
+            // Arrange
+            var fileProvider = MakeFileProvider(MakeDirectoryContents("site.css", "blank.js", "site2.txt", "site.js"));
+            IMemoryCache cache = null;
+            var requestPathBase = PathString.Empty;
+            var globbingUrlBuilder = new GlobbingUrlBuilder(fileProvider, cache, requestPathBase);
+
+            // Act
+            var urlList = globbingUrlBuilder.BuildUrlList(
+                staticUrl: null,
+                includePattern: $"~/*.css, ~/*.txt",
```
NTaylorMullen: why `$`?

test/Microsoft.AspNet.Mvc.TagHelpers.Test/Internal/GlobbingUrlBuilderTest.cs
```diff
@@ -299,10 +298,33 @@ public void CachesMatchResults()
...
+                excludePattern: null);
+
+            // Assert
+            Assert.Collection(urlList,
```
NTaylorMullen: `Assert.Equal(new [] { "/site.css", "/site2.txt" }, urlList, StringComparer.Ordinal);`

test/Microsoft.AspNet.Mvc.TagHelpers.Test/Internal/GlobbingUrlBuilderTest.cs
```diff
@@ -348,8 +371,8 @@ public void TrimsOnlySingleLeadingSlashFromPatterns(string leadingSlash)
                 excludePattern: $"{leadingSlashes}**/*.min.css");
 
             // Assert
-            Assert.Collection(includePatterns, pattern => Assert.Equal($"{leadingSlash}**/*.css", pattern));
-            Assert.Collection(excludePatterns, pattern => Assert.Equal($"{leadingSlash}**/*.min.css", pattern));
+            Assert.Collection(includePatterns, pattern => Assert.Equal($"{prefix}**/*.css", pattern));
+            Assert.Collection(excludePatterns, pattern => Assert.Equal($"{prefix}**/*.min.css", pattern));
         }
```
NTaylorMullen: Should add a variation that has whitespace in front; since you trim the valid whitespace chars.

test/WebSites/HtmlGenerationWebSite/Views/HtmlGeneration_Home/Link.cshtml
NTaylorMullen: Since you added the whitespace trimming, it wouldn't hurt to include a few functional variations of that.

NTaylorMullen:
:watch: for tests.

test/WebSites/HtmlGenerationWebSite/Views/HtmlGeneration_Home/Link.cshtml
dougbu: @NTaylorMullen not sure about _functional_ variations since that sounds a lot like the cross-product of inputs and options that should be covered in unit tests. These tag helper functional tests area already stray a bit too much from mainline scenarios. But including more whitespace in existing functional tests (not creating new cases) sounds reasonable.

dougbu:
:watch:

test/WebSites/HtmlGenerationWebSite/Views/HtmlGeneration_Home/Link.cshtml
NTaylorMullen: Yup, that sounds fine with me.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
Eilon: This doesn't seem especially useful. This type of "refactoring" leads to more spaghetti.

Eilon: (Especially because what's valid in one place might not be valid elsewhere, and that makes code more fragile. No easy answer to this aside from don't do it.)

test/WebSites/HtmlGenerationWebSite/Views/HtmlGeneration_Home/Link.cshtml
Eilon: Having something in a test doesn't "encourage" its use. It merely validates that it works so that we don't break it...

dougbu: Agree in general but we have seen use of our functional test web sites as samples. I certainly wouldn't object if other ways of testing were more fragile exceedingly complex.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/Internal/GlobbingUrlBuilderTest.cs
ajaybhargavb: Added.

ajaybhargavb:
Updated.

NTaylorMullen:
:shipit:

test/Microsoft.AspNet.Mvc.TagHelpers.Test/Internal/GlobbingUrlBuilderTest.cs
dougbu: Where was this added?

ajaybhargavb: Added a few more cases as `InlineData`.

dougbu:
:shipit:

Never mind my commit-level comment. Thought for a sec I was looking at the whole PR.

natemcmaster:
:+1:

PR #2916, opened 2015-08-05
Add `AddHtmlAttributeValues` for `TagHelper`s.

NTaylorMullen:
- Refactored `WriteAttributeTo` to allow re-use of some of the core attribute writing logic. The refactoring was based on removing the `Begin`/`EndContext` for instrumentation bits which isn't valid in a `TagHelper` attribute scenario.
- Added unit tests to validate attributes are properly added to `TagHelperExecutionContext`.
- Added functional test to validate everything is output as expected with dynamic attributes (encoded and non-encoded).

aspnet/Razor#247

Corresponding Razor pull request: https://github.com/aspnet/Razor/pull/475


src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -519,30 +519,32 @@ public virtual void WriteLiteralTo([NotNull] TextWriter writer, string value)
...
             }
-            else if (values.Length == 1 && values[0].Prefix == "" &&
-                (values[0].Value.Value is bool || values[0].Value.Value == null))
+            else if (!ShouldRenderAttribute(values, out useAttributeNameAsValue))
+            {
+                return;
+            }
+            else if (useAttributeNameAsValue)
             {
-                // Value is either null or a bool with no prefix.
```
dougbu: Restore comment a few lines up, before the `return`

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -519,30 +519,32 @@ public virtual void WriteLiteralTo([NotNull] TextWriter writer, string value)
...
+            bool useAttributeNameAsValue;
+
             if (values.Length == 0)
             {
                 // Explicitly empty attribute, so write the prefix
                 WritePositionTaggedLiteral(writer, prefix);
             }
-            else if (values.Length == 1 && values[0].Prefix == "" &&
-                (values[0].Value.Value is bool || values[0].Value.Value == null))
+            else if (!ShouldRenderAttribute(values, out useAttributeNameAsValue))
```
dougbu: This refactoring obscures what's going on. Suggest instead splitting into two methods that each return a `bool` e.g. `IsSingleBoolOrNullValue()` and `UseAttributeNameAsValue()`.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -519,30 +519,32 @@ public virtual void WriteLiteralTo([NotNull] TextWriter writer, string value)
...
-                    return;
-                }
-
                 WritePositionTaggedLiteral(writer, prefix);
 
                 var sourceLength = suffix.Position - positionTaggedAttributeValue.Position;
+                var nameAttributeValue = new AttributeValue(
+                    attributeValue.Prefix,
+                    new PositionTagged<object>(name, attributeValue.Value.Position),
+                    literal: attributeValue.Literal);
```
dougbu: Why change `WriteAttributeValue()` parameters to force creating an object here?

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -846,6 +897,29 @@ public virtual HtmlString SetAntiforgeryCookieAndHeader()
             return HtmlString.Empty;
         }
 
+        private bool ShouldRenderAttribute(AttributeValue[] values, out bool useAttributeNameAsValue)
+        {
+            useAttributeNameAsValue = false;
+
+            if (values.Length == 1 && values[0].Prefix == "" &&
```
dougbu: `values[0].Prefix == ""` -> `string.IsNullOrEmpty()`

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -587,11 +635,7 @@ public virtual string Href([NotNull] string contentPath)
...
         }
 
-        private void WriteAttributeValue(
-            TextWriter writer,
-            AttributeValue attributeValue,
-            string stringValue,
-            int sourceLength)
+        private void WriteAttributeValue(TextWriter writer, AttributeValue attributeValue, int sourceLength)
         {
             var positionTaggedAttributeValue = attributeValue.Value;
```
dougbu: no longer used

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -602,8 +646,17 @@ public virtual string Href([NotNull] string contentPath)
...
             BeginContext(attributeValue.Value.Position, sourceLength, isLiteral: attributeValue.Literal);
 
-            // The extra branching here is to ensure that we call the Write*To(string) overload where
-            // possible.
+            WriteUnprefixedAttributeValueTo(writer, attributeValue);
+
+            EndContext();
+        }
+
+        private void WriteUnprefixedAttributeValueTo(TextWriter writer, AttributeValue attributeValue)
```
dougbu: Rather than split this out, why not add a `bool useContext` parameter to `WriteAttributeValue()`?

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -568,15 +570,61 @@ public virtual void WriteLiteralTo([NotNull] TextWriter writer, string value)
...
+
+                    if (!string.IsNullOrEmpty(value.Prefix))
+                    {
+                        WriteLiteralTo(valueBuffer, value.Prefix);
+                    }
+
+                    WriteUnprefixedAttributeValueTo(valueBuffer, value);
+                }
+
+                var htmlString = new HtmlString(valueBuffer.ToString());
```
dougbu: Ick. We should avoid smashing the `StringCollectionTextWriter`'s buffer together here. Suggest filing a follow-up bug to fix this e.g. add an `HtmlString` constructor overload that accepts an `IHtmlContext`. May be easier to clean up once @sornaks gets aspnet/Razor#473 and #2899  in.

test/Microsoft.AspNet.Mvc.FunctionalTests/TagHelpersTest.cs
```diff
@@ -58,6 +60,37 @@ public class TagHelpersTest
...
         }
 
+        [Fact]
+        public async Task CanRenderViewsWithTagHelpersAndUnboundDynamicAttributes_Encoded()
+        {
+            // Arrange
+            var server = TestHelper.CreateServer(_app, SiteName, services =>
+            {
+                _configureServices(services);
+                services.AddTransient<IHtmlEncoder, TestHtmlEncoder>();
```
dougbu: Should be moving away from `TestHtmlEncoder` (it doesn't need to exist anymore). Use `CommonTestEncoder` instead.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
NTaylorMullen: Talked to @sornaks / @Eilon  about this earlier this week. I dislike it too. Sorna is there not already one logged or should I file?

sornaks: There isn't a bug already. Can you file one?

NTaylorMullen: Yup definitely: https://github.com/aspnet/Mvc/issues/2935

test/WebSites/TagHelpersWebSite/TagHelpers/AddProcessedAttributeTagHelper.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNet.Razor.Runtime.TagHelpers;
+
+namespace TagHelpersWebSite.TagHelpers
+{
+    [TargetElement("input")]
+    public class AddProcessedAttributeTagHelper : TagHelper
```
dougbu: I don't see what would be lost if this tag helper were removed. See that the minimized `processed` attribute shows up in the generated HTML. But how does that relate to the rest of this PR?

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
NTaylorMullen: So the method signature can be used for both `TagHelper`s and non-`TagHelper`s. Otherwise you'd be passing in bs data from the `TagHelper` side; can be misleading.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
NTaylorMullen: :+1:

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
NTaylorMullen: Eh, like the different method approach better than the unneeded parameter/two if statements to go along with it.

test/Microsoft.AspNet.Mvc.FunctionalTests/TagHelpersTest.cs
NTaylorMullen: Lets move away from it in a different PR. To use `CommonTestEncoder` I'd have to add a dependency to `Microsoft.AspNet.Testing`. We can do that in one sweeping change since `TestHtmlEncoder` seems to be consistent in the rest of the functional tests.

test/WebSites/TagHelpersWebSite/TagHelpers/AddProcessedAttributeTagHelper.cs
NTaylorMullen: This `TagHelper` ensures that the input elements in `UnboundDynamicAttributes.cshtml` are run over by a `TagHelper` but not bound by one. Also adding the `processed` attribute lets us visually see that the test is working as expected.

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -524,25 +524,26 @@ public virtual void WriteLiteralTo([NotNull] TextWriter writer, string value)
                 // Explicitly empty attribute, so write the prefix
                 WritePositionTaggedLiteral(writer, prefix);
             }
-            else if (values.Length == 1 && values[0].Prefix == "" &&
-                (values[0].Value.Value is bool || values[0].Value.Value == null))
+            else if (IsSingleBoolOrNullValue(values))
+            {
+                // Value is either null or a bool with no prefix; don't render the attribute.
```
ajaybhargavb: either null or the bool 'false'

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -846,6 +893,33 @@ public virtual HtmlString SetAntiforgeryCookieAndHeader()
             return HtmlString.Empty;
         }
 
+        private bool IsSingleBoolOrNullValue(AttributeValue[] values)
```
ajaybhargavb: Maybe rename it to something like `IsSingleBoolFalseOrNullValue` (not a good name. I know). But the current name says it will return true for any bool.

NTaylorMullen: :+1:

ajaybhargavb:
:shipit:

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
dougbu: You've removed the `stingValue` parameter from `WriteAttributeValue()` and are not using it in the tag helper case. Let's discuss on Monday.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -846,6 +893,33 @@ public virtual HtmlString SetAntiforgeryCookieAndHeader()
...
+        {
+
```
dougbu: Remove extra blank line

test/WebSites/TagHelpersWebSite/TagHelpers/AddProcessedAttributeTagHelper.cs
dougbu: Ah, got it.

dougbu:
:watch:

Issue #2551, opened 2015-05-12
MVC precompilation design notes 5/12/2015


Eilon:
Today we had a design meeting with: @alxandr @danroth27 @DamianEdwards @davidfowl @dougbu @lodejard @pranavkm @NTaylorMullen @ToddGrun @muratg 

Please add your own thoughts as comments to this thread.

## Agenda

* Discuss specific goals for precompilation in ASP.NET 5
 * What kind of precompilation should we support? Updateable? Non-updateable?
 * When should it work? Dev-time (i.e. while working in VS)? Deploy-time?
* Discuss current set of issues related to precompilation, including:
 * https://github.com/aspnet/Mvc/issues/2462 - MVC ignores precompiled views where it cannot find the source file
   * And PR: https://github.com/aspnet/Mvc/pull/2463
 * https://github.com/aspnet/Mvc/issues/2498 - PrecompilationWebSite has troubles building in VS
 * https://github.com/aspnet/RazorTooling/issues/10 - TagHelper tooling does not work with pre-compilation enabled.

---------------

## Meeting notes

1. Precompilation should not be *enabled* by default in project templates (it can *exist*, but should not be used in DTH scenarios)

2. Precompilation is for:
 a. Server optimization (especially for large sites)
 b. Avoiding using a compiler on the server, which is good for security purposes
 c. Components with everything compiled in (including views) and ready to go

3. Precompilation should not require any physical files on disk

4. Precompilation behaviors:
 a. All view locators would always check for a file “on disk” (whether it’s physical or virtual)
 b. Precompilation would embed at least empty placeholder files into a virtual embedded file system
 c. You could also have a case where you embed the full files into the embedded file system, but also have precompilation enabled. You would then have updateable precompilation.

5. We should use precompilation (or something else that’s similar! E.g. a Roslyn Analyzer?) to enable showing Razor errors in VS

6. Question: Should any of these behaviors change while in DesignTimeHost (DTH)?

7. Question: How do you enable precompilation? Today it’s a CompileModule. Should it be something else?

8. Action item: Need to dig into exactly why the tag helper scenario is broken, and address that specifically.
9. Action item: Should we remove Roslyn dependency from Razor? (Not interesting right now)

Alxandr:
[...]

PR #2917, opened 2015-08-05
* Remove support for updateable precompiled views.

### LINKED ISSUE #2462: MVC ignores precompiled views where it cannot find the source file

### LINKED ISSUE #2796: Razor is not caching results for views that don't exist
Nothing is cached when looking for a view that doesn't exist, causing us to repeat the file lookup each time it's requested.

This is a big problem when using `EditorFor` because the default templates are frequently requested and rarely customized. Here's the offending code:
https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilerCache.cs#L108

Suggestion from @pranavkm is to fold this case into `OnCacheMiss` and make that method capable of storing data about a file that doesn't exist on disk.

pranavkm:
* Allow precompiled views to be served when source file does not exist in
  file system.
* Cache results for views that do not exist on disk.

Fixes #2462 and fixes #2796.
Partially addresses #2551


pranavkm:
cc @Alxandr @rynowak @loudej

test/WebSites/PrecompilationWebSite/Startup.cs
```diff
@@ -12,7 +13,9 @@ public class Startup
         public void ConfigureServices(IServiceCollection services)
         {
             // Add MVC services to the services container
-            services.AddMvc();
+            services
```
pranavkm: @rynowak what's the best way to do this?

src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilerCache.cs
```diff
@@ -27,178 +25,78 @@ public class CompilerCache : ICompilerCache
...
+        /// information for precompiled view discovery.</param>
         /// <param name="loaderContextAccessor">The <see cref="IAssemblyLoadContextAccessor"/>.</param>
         /// <param name="optionsAccessor">An accessor to the <see cref="RazorViewEngineOptions"/>.</param>
-        public CompilerCache(IAssemblyProvider assemblyProvider,
-                             IAssemblyLoadContextAccessor loadContextAccessor,
-                             IOptions<RazorViewEngineOptions> optionsAccessor)
-            : this(GetFileInfos(assemblyProvider.CandidateAssemblies),
-                  loadContextAccessor.GetLoadContext(RazorFileInfoCollectionType.Assembly),
+        public CompilerCache(
+            IEnumerable<RazorFileInfoCollection> razorFileInfoCollections,
```
pranavkm: @Alxandr, for the time we have a hybrid of what @loudej mentioned in the design notes in #2551. This approach lets us get away without having to introduce multiple `IRazorPageFactory`s or moving it into Options to allow for reordering instance. With the exception of scenario 3 from your design proposal, we do hit all the other bullet points.

Alxandr: Well, for my case, as long as I can get away with having the razor files in another project than the web project (cause the web project is in F#), I'll be happy. I'll give it a spin, and see if it works after I'm done with my exam the 10th.

rynowak:
:heart_eyes_cat:

src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilerCache.cs
```diff
@@ -27,178 +25,78 @@ public class CompilerCache : ICompilerCache
...
+            IAssemblyLoadContextAccessor loadContextAccessor,
+            IOptions<RazorViewEngineOptions> optionsAccessor)
+            : this(razorFileInfoCollections,
+                  loadContextAccessor.GetLoadContext(typeof(CompilerCache).GetTypeInfo().Assembly),
```
rynowak: would help readability to put this back in a private field. `loadContextAccessor.GetLoadContext(RazorHostAssembly)` > current

src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilerCache.cs
```diff
@@ -27,178 +25,78 @@ public class CompilerCache : ICompilerCache
...
-                    {
-                        CompilationResult = CompilationResult.Successful(cacheEntry.CompiledType),
-                        CompilerCacheEntry = cacheEntry
-                    };
+                    var relativeFileInfo = new RelativeFileInfo(fileInfo, relativePath);
+                    var compilationResult = compile(relativeFileInfo).EnsureSuccessful();
+                    cacheEntryOptions = GetMemoryCacheEntryOptions(relativePath);
+
+                    // Don't cache compilationResult since it could be UncachedCompilationResult
+                    // which has the generated view content as a string.
```
rynowak: I feel like this needs to be expanded a bit. Under what conditions might this happen?

src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilerCache.cs
```diff
@@ -27,178 +25,78 @@ public class CompilerCache : ICompilerCache
...
         {
         }
 
-        internal CompilerCache(IEnumerable<RazorFileInfoCollection> razorFileInfoCollections,
-                               IAssemblyLoadContext loadContext,
-                               IFileProvider fileProvider)
+        internal CompilerCache(
+            IEnumerable<RazorFileInfoCollection> razorFileInfoCollections,
+            IAssemblyLoadContext loadContext,
+            IFileProvider fileProvider)
```
rynowak: Does all of this work need to happen on construction? This causes cascading... services don't want to be initialized because they trigger loading of all the views. 

What we do for things like ADs is expose a 'Collection' (not a service) via a 'Collection Holder' which is a service. Then anyone who wants to can DI the service without forcing eager evaluation. Accessing the collection will trigger all the init code.

src/Microsoft.AspNet.Mvc.Razor/DependencyInjection/MvcRazorMvcBuilderExtensions.cs
```diff
@@ -91,5 +107,27 @@ internal static void AddRazorViewEngineServices(IServiceCollection services)
             // Consumed by the Cache tag helper to cache results across the lifetime of the application.
             services.TryAddSingleton<IMemoryCache, MemoryCache>();
         }
+
+        private static IEnumerable<RazorFileInfoCollection> GetFileInfos(IEnumerable<Assembly> assemblies) =>
```
rynowak: This isn't really `FileInfos` is it?

src/Microsoft.AspNet.Mvc.Razor/DependencyInjection/MvcRazorMvcBuilderExtensions.cs
```diff
@@ -91,5 +107,27 @@ internal static void AddRazorViewEngineServices(IServiceCollection services)
...
+        internal static bool IsValidRazorFileInfoCollection(Type type)
+        {
+            var inAssemblyType = typeof(RazorFileInfoCollection);
+            if (inAssemblyType.IsAssignableFrom(type))
+            {
+                var hasParameterlessConstructor = type.GetConstructor(Type.EmptyTypes) != null;
+
+                return hasParameterlessConstructor
+                    && !type.GetTypeInfo().IsAbstract
+                    && !type.GetTypeInfo().ContainsGenericParameters;
```
rynowak: Should we ignore anything inherited from `RazorFileInfoCollection` without a parameterless constructor? Would it be better to throw because it's a mistake?

src/Microsoft.AspNet.Mvc.Razor/DependencyInjection/MvcRazorMvcBuilderExtensions.cs
rynowak: This is wretched abuse of the `=>` syntax. I grudgingly award 17 PranavPoints®

test/WebSites/PrecompilationWebSite/Startup.cs
rynowak: For now, make an extension method on `IMvcBuilder` **and** on `IServiceCollection`. The `IServiceCollection` part will go away soon.

rynowak:
:watch:

src/Microsoft.AspNet.Mvc.Razor/DependencyInjection/MvcRazorMvcBuilderExtensions.cs
pranavkm: I suppose we can just let `Activator.CreateInstance` attempt to initialize it and throw if the type cannot be initialized.

src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilerCache.cs
pranavkm: We had a similar discussion in the past and at that point we decided it was saner to initialize the constructor to avoid having to perform the `EnsureInitialized` check as part of `GetOrAdd`. 

We have a test that verifies in the scenario where a View is never returned, the cache is never initialized. https://github.com/aspnet/Mvc/blob/dev/test/WebSites/RazorCompilerCacheWebSite/Services/CustomCompilerCache.cs#L20

Granted it does assume that nobody except RazorViewEngine attempts to resolve this service, but my guess is that's the most common scenario.

rynowak: My memory is a little bit fuzzy about that conversation, but the complexity of what I'm suggesting isn't very high at all.

```
public interface ICompileCacheProvider
{
    ICompilerCache Cache { get; } // This does the lazy-init
}
```

It's easy because inside the implementation, cache is just null or initialized. I don't feel very strongly about this, it just seems like a much more friendly way to do this in combination with DI.

pranavkm: Follow up item: https://github.com/aspnet/Mvc/issues/2933

rynowak:
:shipit:

Issue #2900, opened 2015-08-03
Get rid of manual body-read-state-tracking


rynowak:
MVC has custom code to track whether or not the request body has been read, which is easily circumvented. Right now this just prevents a class of common errors like putting `[FromBody]` on two parameters.

We should remove our logic for this and use what's suggested here: https://github.com/aspnet/BasicMiddleware/pull/6/files#r36109498

kichalla:
@rynowak @danroth27 and I discussed about this issue and agreed that adding a check for if a body was read or not to inform users of potential mistakes does not add much value.

Issue #2907, opened 2015-08-04
`CompositeValueProvider` does not match `IBindingSourceValueProvider` contract


dougbu:
Of the `IBindingSourceValueProvider` implementations in MVC, all but `CompositeValueProvider` have a `Filter()` method that returns `null` when the requested `BindingSource` does not match. `CompositeValueProvider` instead returns an empty `CompositeValueProvider`.

Also have a little code in `CompositeModelBinder` and `MutableObjectModelBinder` that expects the `CompositeValueProvider` behaviour of all `IBindingSourceValueProvider` impelmentations. That is, they need to change slightly to handle the expected / contracted `null` return values from `Filter()`.

Eilon:
Make 'em match!

dougbu:
Teal it is :smiling_imp:

dougbu:
4a81377

PR #2918, opened 2015-08-06
Make `CompositeValueProvider` consistent with `IBindingSourceValueProvider` contract

dougbu:
- #2907
- return `null` if `Filter()` finds no matching value provider in collection
- fix lack of defensiveness in model binders' use of `IBindingSourceValueProvider` implementations
- remove test workaround for unusual `CompositeValueProvider` behaviour

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/CompositeModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/CompositeModelBinder.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Runtime.CompilerServices;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Mvc.Core;
@@ -39,6 +40,12 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
                 !string.IsNullOrEmpty(bindingContext.ModelName);
 
             var newBindingContext = CreateNewBindingContext(bindingContext, bindingContext.ModelName);
+            if (newBindingContext == null)
+            {
+                // Unable to find a value provider for this binding source. Binding will fail.
+                return null;
+            }
+
             newBindingContext.IsFirstChanceBinding = isFirstChanceBinding;
             var modelBindingResult = await TryBind(newBindingContext);
 
@@ -46,12 +53,15 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             {
                 // Fall back to empty prefix.
                 newBindingContext = CreateNewBindingContext(bindingContext, modelName: string.Empty);
+                Debug.Assert(newBindingContext != null, "Should have failed on first attempt.");
+
                 modelBindingResult = await TryBind(newBindingContext);
             }
 
             if (modelBindingResult == null)
             {
-                return null; // something went wrong
+                // Unable to bind or something went wrong.
+                return null;
             }
 
             bindingContext.OperationBindingContext.BodyBindingState =
@@ -176,6 +186,11 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
                 if (valueProvider != null)
                 {
                     newBindingContext.ValueProvider = valueProvider.Filter(bindingSource);
+                    if (newBindingContext.ValueProvider == null)
+                    {
+                        // Unable to find a value provider for this binding source.
+                        return null;
+                    }
                 }
             }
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNet.Mvc.Core/ModelBinding/CompositeModelBinder.cs
```diff
@@ -39,6 +40,12 @@ public CompositeModelBinder([NotNull] IEnumerable<IModelBinder> modelBinders)
                 !string.IsNullOrEmpty(bindingContext.ModelName);
 
             var newBindingContext = CreateNewBindingContext(bindingContext, bindingContext.ModelName);
+            if (newBindingContext == null)
```
dougbu: Basically this and `MutableObjectModelBinder` treated all `IBindingSourceValueProvider` implementations as if they were guaranteed to act like `CompositeValueProvider`.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CompositeModelBinder.cs
```diff
@@ -176,6 +186,11 @@ public CompositeModelBinder([NotNull] IEnumerable<IModelBinder> modelBinders)
                 if (valueProvider != null)
                 {
                     newBindingContext.ValueProvider = valueProvider.Filter(bindingSource);
+                    if (newBindingContext.ValueProvider == null)
+                    {
+                        // Unable to find a value provider for this binding source.
+                        return null;
+                    }
                 }
```
dougbu: If / when we fix #2900, can move this block to the top and avoid creating the `newBindingContext` right above where we throw it away and return `null`. Reordering things today visibly changes the behaviour, hiding some expected errors.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CompositeModelBinder.cs
```diff
@@ -39,6 +40,12 @@ public CompositeModelBinder([NotNull] IEnumerable<IModelBinder> modelBinders)
                 !string.IsNullOrEmpty(bindingContext.ModelName);
 
             var newBindingContext = CreateNewBindingContext(bindingContext, bindingContext.ModelName);
+            if (newBindingContext == null)
+            {
+                // Unable to find a value provider for this binding source. Binding will fail.
+                return null;
+            }
```
rynowak: This is a really roundabout way of checking this - will we really fail if there's no VP? a model can have all greedy properties.

dougbu: Huh? The only case where `Filter()` returns `null` is if no value provider matches. And `Filter()` is only called if the requested `BindingSource` is _not_ greedy.

rynowak: You're right, but this still seems like an optimization for something that will never happen.

dougbu: This isn't an optimization. It's about the `IBindingSourceValueProvider` contract. The only possible "empty" (i.e. useless) value provider is a `CompositeValueProvider` with an empty collection. An `IBindingSourceValueProvider` implementation is supposed to return `null` instead of such a provider.

And it's fairly easy to make happen during development: Create a new non-greedy `BindingSource`, use it, but don't add an associated  factory to `MvcOptions.ValueProviderFactories`. Our tests cover this scenario, just like they cover not adding a needed `IModelBinder` to `MvcOptions.ModelBinders`.

rynowak: This would definitely make the code more consistent. Should we throw?

dougbu: Not sure. We don't throw in any other case where we don't have a value for a parameter or property. Throwing here would also make incremental development more difficult.

Suggest we separate this concern into an issue about throwing for both a "missing" value provider and model binder. Thoughts?

rynowak: > Suggest we separate this concern into an issue about throwing for both a "missing" value provider and model binder. Thoughts?

I think I'm fine with not throwing. You're right that we don't do anything else like this in MB.

rynowak:
:shipit:

dougbu:
4a81377

PR #2920, opened 2015-08-06
Update ApiController tests with StatusCodes class constants

AshrafAlam:
Few of the ApiController test methods contain checks for certain status code that is being returned from controller (i.e. Ok, NotFound status).

The current commit updates assertion of few of the test methods to have constants instead of hard-coded status code (i.e. StatusCodes.Status200OK instead of 200).


Eilon:
Hi @AshrafAlam we're generally looking for more substantial changes. Though the status codes were in some sense "hard coded," they'll always be the same value. If you have more value to add that is related to this change we could take this, but I'm not sure it's worth it for just this one change.

Issue #2745, opened 2015-06-25
ViewHierarchyUtility.GetHierarchicalPath does not handle root paths *nix file systems.


sornaks:
It trims the leading `/` and then tries to check `Path.IsPathRooted()`. This will always return `false` in *nix file systems.

sornaks:
https://github.com/aspnet/Mvc/commit/1aef84b50d8afd91d2d5e23596596724d8e5d9dd

PR #2925, opened 2015-08-06
Issue #2745 - ViewHierarchyUtility.GetHierarchicalPath does not handle root paths *nix file systems.

sornaks:
Fix: It is not necessary to check for root paths in this place because `GetHierarchicalPath` is always called with a relative path. Hence removing the check.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor.Host/ViewHierarchyUtility.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor.Host/ViewHierarchyUtility.cs
@@ -68,12 +68,6 @@ namespace Microsoft.AspNet.Mvc.Razor
                 relativePath = relativePath.Substring(1);
             }
 
-            if (Path.IsPathRooted(relativePath))
-            {
-                // If the path looks like it's not app relative, don't attempt to construct paths.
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Mvc.Razor.Host/ViewHierarchyUtility.cs
```diff
@@ -68,12 +68,6 @@ private static IEnumerable<string> GetHierarchicalPath(string relativePath, stri
                 relativePath = relativePath.Substring(1);
             }
 
-            if (Path.IsPathRooted(relativePath))
```
pranavkm: @Eilon, the paths that are used in this API are always app relative (locating `_ViewStart`, `_ViewImports`). This would be a safe change.

pranavkm:
:shipit:

src/Microsoft.AspNet.Mvc.Razor.Host/ViewHierarchyUtility.cs
Eilon: Yup, @sornaks talked to me yesterday.

Eilon:
:shipit:

PR #2926, opened 2015-08-06
Fix #2874 part 2 - Add AddLocalization() overload for MVC localization.

### LINKED ISSUE #2874: Add IMvcBuilder overloads for localization and XML formatters
See here: https://github.com/aspnet/Mvc/issues/2872#issuecomment-125750316

We don't have an `IMvcBuilder` startup experience for these pieces.

rynowak:


kirthik:
:shipit:

src/Microsoft.AspNet.Mvc.Localization/DependencyInjection/MvcLocalizationMvcBuilderExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// </summary>
+    public static class MvcLocalizationMvcBuilderExtensions
+    {
+        /// <summary>
+        /// Adds MVC localization to the application.
+        /// </summary>
+        /// <param name="builder">The <see cref="IMvcBuilder"/>.</param>
+        /// <returns>The <see cref="IMvcBuilder"/>.</returns>
+        /// <remarks>
+        /// Adding loca
```
dougbu: Might be missing something here. Do you mean "loca" as a slang term, Spanish for crazy, the plural for locus, ...

rynowak: heh. Forgot to finish typing, it's going to say "Adding localization also adds view support by calling AddViews(...)."

rynowak:
updated

src/Microsoft.AspNet.Mvc.Localization/DependencyInjection/MvcLocalizationMvcBuilderExtensions.cs
dougbu: Sounds good. Should have similar `<remarks>` on the other overload too.

dougbu:
:shipit:

PR #2946, opened 2015-08-10
Handle broader range of collection types in model binding

dougbu:
- #2793
- add `ICollectionModelBinder`, allowing `GenericModelBinder` to call `CreateEmptyCollection()`
- adjust `CollectionModelBinder` and `DictionaryModelBinder` to activate model if default types are incompatible
 - do not create default (empty) top-level collection in fallback case if Model already non-`null`
- change type checks in `GenericModelBinder` to align with `CollectionModelBinder` capabilities
 - add special case for `IEnumerable<T>`
- correct `ModelMetadata` of a few tests that previously did not need that information

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/ArrayModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/ArrayModelBinder.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
@@ -25,15 +26,16 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             return base.BindModelAsync(bindingContext);
         }
 
-        protected override object CreateEmptyCollection()
+        /// <inheritdoc />
+        public override object CreateEmptyCollection(Type targetType)
         {
             return new TElement[0];
         }
 
         /// <inheritdoc />
-        protected override object GetModel(IEnumerable<TElement> newCollection)
+        protected override object GetModel(Type targetType, IEnumerable<TElement> collection)
         {
-            return newCollection?.ToArray();
+            return collection?.ToArray();
         }
 
         /// <inheritdoc />
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/ArrayModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/ArrayModelBinder.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.Framework.Internal;
@@ -27,14 +28,27 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         }
 
         /// <inheritdoc />
-        public override object CreateEmptyCollection(Type targetType)
+        public override bool CanCreateInstance(Type targetType)
         {
+            Debug.Assert(targetType == typeof(TElement[]), "GenericModelBinder only creates this binder for arrays.");
+
+            return true;
+        }
+
+        /// <inheritdoc />
+        protected override object CreateEmptyCollection(Type targetType)
+        {
+            Debug.Assert(targetType == typeof(TElement[]), "GenericModelBinder only creates this binder for arrays.");
+
             return new TElement[0];
         }
 
         /// <inheritdoc />
-        protected override object GetModel(Type targetType, IEnumerable<TElement> collection)
+        protected override object ConvertToCollectionType(Type targetType, IEnumerable<TElement> collection)
         {
+            Debug.Assert(targetType == typeof(TElement[]), "GenericModelBinder only creates this binder for arrays.");
+
+            // If non-null, collection is a List<TElement>, never already a TElement[].
             return collection?.ToArray();
         }
 
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
@@ -7,8 +7,9 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
+#if DNXCORE50
 using System.Reflection;
-using System.Runtime.InteropServices;
+#endif
 using System.Threading.Tasks;
 using Microsoft.Framework.Internal;
 
@@ -70,7 +71,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             var boundCollection = result.Model;
             if (model == null)
             {
-                model = GetModel(bindingContext.ModelType, boundCollection);
+                model = ConvertToCollectionType(bindingContext.ModelType, boundCollection);
             }
             else
             {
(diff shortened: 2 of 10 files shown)


src/Microsoft.AspNet.Mvc.Core/ModelBinding/ArrayModelBinder.cs
```diff
@@ -25,15 +26,16 @@ public override Task<ModelBindingResult> BindModelAsync([NotNull] ModelBindingCo
...
         {
             return new TElement[0];
         }
 
         /// <inheritdoc />
-        protected override object GetModel(IEnumerable<TElement> newCollection)
+        protected override object GetModel(Type targetType, IEnumerable<TElement> collection)
         {
-            return newCollection?.ToArray();
+            return collection?.ToArray();
```
rynowak: Will the default case here be that the model is already an array? What are the cases we might get here and have it not be an array?

dougbu: It'll _never_ be an array since the base `CollectionModelBinder` uses a `List<TElement>` until it needs to copy everything into the final model. The value should not be `null` but this was written defensively originally...

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
```diff
@@ -63,13 +65,12 @@ public class CollectionModelBinder<TElement> : IModelBinder
...
                     valueProviderResult.Culture);
-                boundCollection = result.Model;
             }
 
-            model = bindingContext.Model;
+            var boundCollection = result.Model;
             if (model == null)
             {
-                model = GetModel(boundCollection);
+                model = GetModel(bindingContext.ModelType, boundCollection);
```
rynowak: Should this be named something more like `ConvertToCollectionType`? Or are there other uses?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
```diff
@@ -84,14 +85,61 @@ public class CollectionModelBinder<TElement> : IModelBinder
...
+            catch (TargetInvocationException)
+            {
+                // Constructor threw an Exception.
+            }
+            catch (TypeLoadException)
+            {
+                // Invalid type.
+            }
+
+            return null;
```
rynowak: If we're really determined to just null out here, then we should handle all exceptions. Documented exceptions thrown from a method in the CLR aren't a contract and can change.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
```diff
@@ -80,15 +84,34 @@ public class DictionaryModelBinder<TKey, TValue> : CollectionModelBinder<KeyValu
...
-        protected override object GetModel(IEnumerable<KeyValuePair<TKey, TValue>> newCollection)
+        protected override object GetModel(Type targetType, IEnumerable<KeyValuePair<TKey, TValue>> collection)
         {
-            return newCollection?.ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
+            if (collection == null)
+            {
+                return null;
+            }
+
+            if (targetType.IsAssignableFrom(typeof(Dictionary<TKey, TValue>)))
```
rynowak: should this be `!targetType.IsAssignableFrom(typeof(Dictionary<TKey, TValue>))` instead?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
```diff
@@ -13,12 +13,21 @@ public class GenericModelBinder : IModelBinder
...
+            var binderType = ResolveBinderType(bindingContext);
             if (binderType != null)
             {
                 var binder = (IModelBinder)Activator.CreateInstance(binderType);
-                var result = await binder.BindModelAsync(bindingContext);
 
+                var collectionBinder = binder as ICollectionModelBinder;
+                if (collectionBinder != null &&
+                    bindingContext.Model == null &&
+                    collectionBinder.CreateEmptyCollection(bindingContext.ModelType) == null)
```
rynowak: Rather than this, it would be nice to add a `CanBindType` or similar to `ICollectionModelBinder`.

This also seems to ignore cases where we can safely update a type, but can't construct it.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/ArrayModelBinder.cs
rynowak: Cool! Can you add that info to a comment here?

rynowak:
:watch: overall looks good, tests look good.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
dougbu: I made a fairly-squishy choice between objections against `catch (Exception)` and objections to the long list. Fine moving back in the other direction.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/ArrayModelBinder.cs
dougbu: :ok_hand:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
dougbu: Sounds fine. There are no other uses -- `GetModel()` is completely internal to the (wonderful) three `CollectionModelBinder` classes and it's called in precisely one place.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
dougbu: A `List<KeyValuePair<TKey, TValue>>`. Are you asking for a comment stating that it's never already a dictionary? Note the parameter is used only as it's declared -- `IEnumerable<KeyValuePair<TKey, TValue>>`.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
dougbu: ... where the implementation in `CollectionModelBinder` would just do this `null` check?

dougbu:
Updated

src/Microsoft.AspNet.Mvc.Core/ModelBinding/ArrayModelBinder.cs
```diff
@@ -25,15 +27,29 @@ public override Task<ModelBindingResult> BindModelAsync([NotNull] ModelBindingCo
             return base.BindModelAsync(bindingContext);
         }
 
-        protected override object CreateEmptyCollection()
+        /// <inheritdoc />
+        public override bool CanCreateInstance(Type targetType)
+        {
+            Debug.Assert(targetType == typeof(TElement[]), "GenericModelBinder only creates this binder for arrays.");
+
+            return true;
```
rynowak: Why just do the thing you put in the assert? You could say with like 20 characters what all of this does, and technically be more correct :laughing:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
rynowak: Yeah, a comment would make it must more clear :+1:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
rynowak: I tend to always lean towards `catch (Exception)` just because this isn't Java

rynowak:
:shipit:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/ArrayModelBinder.cs
dougbu: done

src/Microsoft.AspNet.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
dougbu: Yes, this iteration includes a comment I added 2 lines below.

dougbu:
d45e2ee

PR #2948, opened 2015-08-10
Remove IScopedInstance - use AsyncLocal for ActionContext and

rynowak:
ActionBindingContext

This change replaces IScopedInstance<T> in favor or IActionContextAccessor
and IActionBindingContextAccessor. In the spirit of IHttpContextAccessor,
these are both singletons which use AsyncLocal for storage.

This change allows the invoker factory to be cached which results in some
significant perf gains.


rynowak:
In both of these data, 100% is set to the time spent inside of Routing + MVC. All percentages are relative to that.

Before
--------
before


Notice here the 10% of time spent just creating the logger. Also 5% doing the service resolve for the invoker factory/providers.

After
------
after

Eilon:
I super duper don't understand this, but I like the perf gain :smile: 

@davidfowl please review.

rynowak:
> I super duper don't understand this, but I like the perf gain :smile:

**TLDR: scoped is viral. Lots of stuff we MUST use every request touches scoped stuff. If we do it less we'll have more ability to cache things we use every request, and to do further optimizations**

Eilon:
Thanks. In principle the code looks :shipit: to me - it seems like a fairly straightforward, if viral, change.

borgdylan:
This would depend on net46 being part of mono which is currently blocked due as the reference source was not updated on github.

rynowak:
> This would depend on net46 being part of mono which is currently blocked due as the reference source was not updated on github.

The code that's actually using `AsyncLocal<T>` is if-defed for coreclr. We'd consider adding an if-def and build target for 4.6 if more of these things pile up.

davidfowl:
@rynowak why's you close this?

rynowak:
> why's you close this?

Because I'm going to merge it you silly goose.

PR #2952, opened 2015-08-11
Making TagBuilder implement IHtmlContent and removing ToHtmlContent.

sornaks:
- Making `TagRenderMode` a property in `TagBuilder`.
- Modifying places where `TagBuilder` is used to suit the new model.
- This reduces space occupied by a normal application by 11.8%. (Tested over 1000 requests with 20 in parallel)

### DIFF
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
@@ -248,14 +248,11 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
                 // <input type="hidden"/> into the output's Content.
                 output.Attributes.Clear();
                 output.TagName = null;
-
-                checkBoxTag.TagRenderMode = TagRenderMode.SelfClosing;
                 output.Content.Append(checkBoxTag);
 
                 var hiddenForCheckboxTag = Generator.GenerateHiddenForCheckbox(ViewContext, modelExplorer, For.Name);
                 if (hiddenForCheckboxTag != null)
                 {
-                    hiddenForCheckboxTag.TagRenderMode = TagRenderMode.SelfClosing;
                     output.Content.Append(hiddenForCheckboxTag);
                 }
             }
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
@@ -236,25 +236,27 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
                 }
             }
 
-            var tagBuilder = Generator.GenerateCheckBox(
+            var checkBoxTag = Generator.GenerateCheckBox(
                 ViewContext,
                 modelExplorer,
                 For.Name,
                 isChecked: null,
                 htmlAttributes: htmlAttributes);
-            if (tagBuilder != null)
+            if (checkBoxTag != null)
             {
                 // Do not generate current element's attributes or tags. Instead put both <input type="checkbox"/> and
                 // <input type="hidden"/> into the output's Content.
                 output.Attributes.Clear();
                 output.TagName = null;
 
-                output.Content.Append(tagBuilder.ToHtmlContent(TagRenderMode.SelfClosing));
+                checkBoxTag.TagRenderMode = TagRenderMode.SelfClosing;
+                output.Content.Append(checkBoxTag);
 
-                tagBuilder = Generator.GenerateHiddenForCheckbox(ViewContext, modelExplorer, For.Name);
-                if (tagBuilder != null)
+                var hiddenForCheckboxTag = Generator.GenerateHiddenForCheckbox(ViewContext, modelExplorer, For.Name);
+                if (hiddenForCheckboxTag != null)
                 {
-                    output.Content.Append(tagBuilder.ToHtmlContent(TagRenderMode.SelfClosing));
+                    hiddenForCheckboxTag.TagRenderMode = TagRenderMode.SelfClosing;
+                    output.Content.Append(hiddenForCheckboxTag);
                 }
             }
         }
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultDisplayTemplates.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultDisplayTemplates.cs
@@ -256,16 +256,17 @@ namespace Microsoft.AspNet.Mvc.Rendering
                     readOnly: true,
                     additionalViewData: null);
 
+                var templateBuilderResult = templateBuilder.Build();
                 if (!propertyMetadata.HideSurroundingHtml)
                 {
-                    var containerDivTag = new TagBuilder("div");
-                    containerDivTag.AddCssClass("display-field");
-                    containerDivTag.InnerHtml = templateBuilder.Build();
-                    content.AppendLine(containerDivTag);
+                    var valueDivTag = new TagBuilder("div");
+                    valueDivTag.AddCssClass("display-field");
+                    valueDivTag.InnerHtml = templateBuilderResult;
+                    content.AppendLine(valueDivTag);
                 }
                 else
                 {
-                    content.Append(templateBuilder.Build());
+                    content.Append(templateBuilderResult);
                 }
             }
 
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultDisplayTemplates.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultDisplayTemplates.cs
@@ -41,7 +41,8 @@ namespace Microsoft.AspNet.Mvc.Rendering
                 inputTag.Attributes["checked"] = "checked";
             }
 
-            return inputTag.ToHtmlContent(TagRenderMode.SelfClosing);
+            inputTag.TagRenderMode = TagRenderMode.SelfClosing;
+            return inputTag;
         }
 
         private static IHtmlContent BooleanTemplateDropDownList(IHtmlHelper htmlHelper, bool? value)
@@ -52,17 +53,13 @@ namespace Microsoft.AspNet.Mvc.Rendering
             selectTag.Attributes["disabled"] = "disabled";
 
             var content = new BufferedHtmlContent();
-            content.Append(selectTag.ToHtmlContent(TagRenderMode.StartTag));
-
             foreach (var item in TriStateValues(value))
             {
-                content.Append(
-                    DefaultHtmlGenerator.GenerateOption(item, item.Text)
-                        .ToHtmlContent(TagRenderMode.Normal));
+                content.Append(DefaultHtmlGenerator.GenerateOption(item, item.Text));
             }
 
-            content.Append(selectTag.ToHtmlContent(TagRenderMode.EndTag));
-            return content;
+            selectTag.InnerHtml = content;
+            return selectTag;
         }
 
         // Will soon need to be shared with the default editor templates implementations.
@@ -208,7 +205,6 @@ namespace Microsoft.AspNet.Mvc.Rendering
             var viewData = htmlHelper.ViewData;
             var templateInfo = viewData.TemplateInfo;
             var modelExplorer = viewData.ModelExplorer;
-            var content = new BufferedHtmlContent();
 
             if (modelExplorer.Model == null)
             {
@@ -229,6 +225,7 @@ namespace Microsoft.AspNet.Mvc.Rendering
             var serviceProvider = htmlHelper.ViewContext.HttpContext.RequestServices;
             var viewEngine = serviceProvider.GetRequiredService<ICompositeViewEngine>();
 
+            var content = new BufferedHtmlContent();
             foreach (var propertyExplorer in modelExplorer.Properties)
             {
                 var propertyMetadata = propertyExplorer.Metadata;
@@ -237,23 +234,16 @@ namespace Microsoft.AspNet.Mvc.Rendering
                     continue;
                 }
 
-                var divTag = new TagBuilder("div");
-
                 if (!propertyMetadata.HideSurroundingHtml)
                 {
                     var label = propertyMetadata.GetDisplayName();
                     if (!string.IsNullOrEmpty(label))
                     {
-                        divTag.SetInnerText(label);
-                        divTag.AddCssClass("display-label");
-                        content.AppendLine(divTag.ToHtmlContent(TagRenderMode.Normal));
-
-                        // Reset divTag for reuse.
-                        divTag.Attributes.Clear();
-                    }
-
-                    divTag.AddCssClass("display-field");
-                    content.Append(divTag.ToHtmlContent(TagRenderMode.StartTag));
+                        var labelTag = new TagBuilder("div");
+                        labelTag.SetInnerText(label);
+                        labelTag.AddCssClass("display-label");
+                        content.AppendLine(labelTag);
+                    } 
                 }
 
                 var templateBuilder = new TemplateBuilder(
@@ -266,11 +256,16 @@ namespace Microsoft.AspNet.Mvc.Rendering
                     readOnly: true,
                     additionalViewData: null);
 
-                content.Append(templateBuilder.Build());
-
                 if (!propertyMetadata.HideSurroundingHtml)
                 {
-                    content.AppendLine(divTag.ToHtmlContent(TagRenderMode.EndTag));
+                    var containerDivTag = new TagBuilder("div");
+                    containerDivTag.AddCssClass("display-field");
+                    containerDivTag.InnerHtml = templateBuilder.Build();
+                    content.AppendLine(containerDivTag);
+                }
+                else
+                {
+                    content.Append(templateBuilder.Build());
                 }
             }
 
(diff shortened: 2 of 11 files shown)


src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
```diff
@@ -236,25 +236,27 @@ private void GenerateCheckBox(ModelExplorer modelExplorer, TagHelperOutput outpu
                 }
             }
 
-            var tagBuilder = Generator.GenerateCheckBox(
+            var checkBoxTag = Generator.GenerateCheckBox(
```
dougbu: Curious: Why did you change from reusing the one `tagBuilder` variable?

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultDisplayTemplates.cs
```diff
@@ -237,23 +234,16 @@ public static IHtmlContent ObjectTemplate(IHtmlHelper htmlHelper)
...
-                        divTag.AddCssClass("display-label");
-                        content.AppendLine(divTag.ToHtmlContent(TagRenderMode.Normal));
-
-                        // Reset divTag for reuse.
-                        divTag.Attributes.Clear();
-                    }
-
-                    divTag.AddCssClass("display-field");
-                    content.Append(divTag.ToHtmlContent(TagRenderMode.StartTag));
+                        var labelTag = new TagBuilder("div");
```
dougbu: `divTag`; it's not a `<label>`

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultDisplayTemplates.cs
```diff
@@ -237,23 +234,16 @@ public static IHtmlContent ObjectTemplate(IHtmlHelper htmlHelper)
...
                 {
                     var label = propertyMetadata.GetDisplayName();
                     if (!string.IsNullOrEmpty(label))
                     {
-                        divTag.SetInnerText(label);
-                        divTag.AddCssClass("display-label");
-                        content.AppendLine(divTag.ToHtmlContent(TagRenderMode.Normal));
-
-                        // Reset divTag for reuse.
-                        divTag.Attributes.Clear();
```
dougbu: Why not reuse the same `TagBuilder` instance?

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultDisplayTemplates.cs
```diff
@@ -266,11 +256,16 @@ public static IHtmlContent ObjectTemplate(IHtmlHelper htmlHelper)
                     readOnly: true,
                     additionalViewData: null);
 
-                content.Append(templateBuilder.Build());
-
                 if (!propertyMetadata.HideSurroundingHtml)
                 {
-                    content.AppendLine(divTag.ToHtmlContent(TagRenderMode.EndTag));
+                    var containerDivTag = new TagBuilder("div");
```
dougbu: `divTag` or perhaps `valueDivTag`; it's no more a "container" than any other `<div>` element.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultDisplayTemplates.cs
```diff
@@ -266,11 +256,16 @@ public static IHtmlContent ObjectTemplate(IHtmlHelper htmlHelper)
                     readOnly: true,
                     additionalViewData: null);
 
-                content.Append(templateBuilder.Build());
```
dougbu: stick `templateBuilder.Build()` into a variable to make it blindingly obvious it's called without regard to `HideSurroundingHtml`.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultEditorTemplates.cs
```diff
@@ -266,16 +264,11 @@ public static IHtmlContent ObjectTemplate(IHtmlHelper htmlHelper)
...
                     {
-                        divTag.AddCssClass("editor-label");
-                        divTag.InnerHtml = label; // already escaped
-                        content.AppendLine(divTag.ToHtmlContent(TagRenderMode.Normal));
-
-                        // Reset divTag for reuse.
-                        divTag.Attributes.Clear();
+                        var labelDivTag = new TagBuilder("div");
+                        labelDivTag.AddCssClass("editor-label");
+                        labelDivTag.InnerHtml = label; // already escaped
```
dougbu: Remove `// already escaped` comment since `htmlHelper.Label()` returns an `IHtmlContent`, not a `string`.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultEditorTemplates.cs
```diff
@@ -266,16 +264,11 @@ public static IHtmlContent ObjectTemplate(IHtmlHelper htmlHelper)
                         htmlAttributes: null);
                     if (!string.IsNullOrEmpty(label.ToString()))
                     {
-                        divTag.AddCssClass("editor-label");
-                        divTag.InnerHtml = label; // already escaped
-                        content.AppendLine(divTag.ToHtmlContent(TagRenderMode.Normal));
-
-                        // Reset divTag for reuse.
-                        divTag.Attributes.Clear();
+                        var labelDivTag = new TagBuilder("div");
```
dougbu: Rinse and repeat comments from `DefaultDisplayTemplates`

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultEditorTemplates.cs
```diff
@@ -288,18 +281,26 @@ public static IHtmlContent ObjectTemplate(IHtmlHelper htmlHelper)
...
-                content.Append(templateBuilder.Build());
-
                 if (!propertyMetadata.HideSurroundingHtml)
                 {
-                    content.Append(" ");
-                    content.Append(htmlHelper.ValidationMessage(
+                    var divTag = new TagBuilder("div");
+                    divTag.AddCssClass("editor-field");
+
+                    var innerContent = new BufferedHtmlContent();
```
dougbu: Not sure what refactoring to avoid `TagRenderModel.StartTag` and `EndTag` buys us. This extra buffer and the new lack of alignment between the code and generated HTML are downsides.

dougbu: Oh, I see. Can't put the same `TagBuilder` into the final content. Might do with a couple of helper methods to restore alignment between code and generated HTML.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -1302,12 +1302,26 @@ private IHtmlContent GenerateGroupsAndOptions(string optionLabel, IEnumerable<Se
...
-                TagBuilder groupBuilder = null;
                 if (optGroup != null)
                 {
-                    groupBuilder = new TagBuilder("optgroup");
+                    optGroupContent = new BufferedHtmlContent();
+                    optGroupContent.Append(Environment.NewLine);
+                }
+                else
+                {
+                    optGroupContent = listItemBuilder;
```
dougbu: This and [lines 1335,1336](https://github.com/aspnet/Mvc/pull/2952/files#diff-0e83439eeed8ffbde4eceb0db1f6e541R1336) mean the `listItemBuilder` is placed in the `listItemBuilder` content. How does this not stack dive?

src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
```diff
@@ -236,25 +236,27 @@ private void GenerateCheckBox(ModelExplorer modelExplorer, TagHelperOutput outpu
...
-            if (tagBuilder != null)
+            if (checkBoxTag != null)
             {
                 // Do not generate current element's attributes or tags. Instead put both <input type="checkbox"/> and
                 // <input type="hidden"/> into the output's Content.
                 output.Attributes.Clear();
                 output.TagName = null;
 
-                output.Content.Append(tagBuilder.ToHtmlContent(TagRenderMode.SelfClosing));
+                checkBoxTag.TagRenderMode = TagRenderMode.SelfClosing;
```
dougbu: Instead set `TagRenderMode` when creating the `TagBuilder` instance. Current separation between `TagBuilder` content and whether or not the element should be self-closing is artificial, just a result of the previous APIs. Might even be worth adding the mode or a `bool isSelfClosing` parameter to a `TagBuilder` constructor overload, saving the property setter for the (rare) `StartTag` / `EndTag` special cases.

Here for example `DefaultHtmlGenerator` knows the `<input>` tag is self-closing and this code shouldn't care.

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultDisplayTemplates.cs
Eilon: But it *is* a tag containing a label. Calling it `divTag` would honestly be more akin to Hungarian notation. Identifiers should convey *semantics*, not *types*.

dougbu: Well, most `TagBuilder` variables in product code are named `tagBuilder` which truly conveys the _type_. Even that works better here than `labelTag` because the semantics of the generated `<div>` are not the same as the contained `<label>` element.

BTW we also name `TagBuilder` variables after the generated element many other places, starting probably in MVC 1.0. The reasoning might be along the lines "The _semantics_ of this `TagBuilder` are that it will generate a `<div>` element." Are you suggesting we make a general change throughout MVC?

src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
Eilon: I don't think combining a `bool` and `TagBuilderMode` makes things any simpler. I'd like to stick to having one way to set the mode. Easier to document, easier to understand.

But one thing that I think would be nice is if it would be possible to set the mode in the generator as opposed to out here.

dougbu: Works for me. Hence my "Might even" prefix to the "any simpler" sentence :smiley_cat:

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultDisplayTemplates.cs
Eilon: Yes, but those are almost always wrong. Just like calling all StringBuilders `sb`.

In some cases there's nothing more to say about the tag builder other than `divBuilder` so it's not necessarily wrong. In this case it's not just "any" div builder - it's for one that contains a label.

dougbu: @Eilon at least for me `labelTag` is confusing because this builder will not generate a `<label>`. That name doesn't say anything about what the builder will do, just what the generated body will be.

Would `labelDivBuilder` (or perhaps `nameDivBuilder` since this is a display template) work here and `valueDivBuilder` just below? If not what names do you suggest?

Eilon: I don't think it's worth changing.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultEditorTemplates.cs
sornaks: I don't understand. Where has the alignment changed?

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultHtmlGenerator.cs
sornaks: No. <a href="https://github.com/aspnet/Mvc/pull/2952/files#diff-0e83439eeed8ffbde4eceb0db1f6e541R1336">Lines 1335,1336</a> are executed only if `optGroup` is <i>not</i> null. So `listItemBuilder` will never go into the content of `listItemBuilder`.

src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
sornaks: I think the Generator should not set the <i>Render</i> mode. I should be able to generate a tag and choose to render it as self-closing or as a normal tag with start and end tags (we do this in a few places in our code).

sornaks:
Updated..

src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
Eilon: Ah then nevermind.

dougbu: Sorry, I'm not understanding this "Won't fix". The HTML specification makes it clear that end tags are illegal for void elements e.g. `<input>`. The HTML generator deals with such HTML semantics. Why instead spread the information around?

Separately I mentioned leaving handling of start and end tags alone in my original comment. That's a separate concern from whether the end tag is valid at all.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultHtmlGenerator.cs
dougbu: This is a relatively short `foreach` block but it remains difficult to grok. Suggest repeating the inner `item in group` loop within a single `if (optGroup == null) {} else {}` statement.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultEditorTemplates.cs
```diff
@@ -288,18 +281,27 @@ public static IHtmlContent ObjectTemplate(IHtmlHelper htmlHelper)
                     readOnly: false,
                     additionalViewData: null);
 
-                content.Append(templateBuilder.Build());
-
+                var templateBuilderResult = templateBuilder.Build();
                 if (!propertyMetadata.HideSurroundingHtml)
```
dougbu: The new structure may be _too_ similar to the previous code. Suggest seeing if the loop is more clear with a single `propertyMetadata.HideSurroundingHtml` check, similar to my recent `DefaultHtmlGenerator` comment. Could do the same thing in `DefaultDisplayTemplates` but there's less inside the conditional blocks there.

src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
sornaks: <a href="https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs#L170">This</a> is the place I am referring to. Not all generated tags are self-closing except checkbox.

dougbu: What? Definitely not just `<input type="checkbox" />` alone. All void elements (``, `<input>`, `<link>`, and more) must be written out as self-closing (or start tag only once @NTaylorMullen gets void element support in). Fortunately the default rendering mode works fine for most of what `DefaultHtmlGenerator` generates.

sornaks:
Updated..

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultHtmlGenerator.cs
```diff
@@ -1302,12 +1304,17 @@ private IHtmlContent GenerateGroupsAndOptions(string optionLabel, IEnumerable<Se
             foreach (var group in groupedSelectList)
             {
                 var optGroup = group.First().Group;
-
-                // Wrap if requested.
-                TagBuilder groupBuilder = null;
                 if (optGroup != null)
                 {
-                    groupBuilder = new TagBuilder("optgroup");
+                    var optGroupContent = new BufferedHtmlContent();
```
dougbu: Suggest moving this down closer to the first use (`groupBuilder.InnerHtml = optGroupContent;`)

sornaks: This is used in <a href="https://github.com/aspnet/Mvc/pull/2952/files#diff-0e83439eeed8ffbde4eceb0db1f6e541R1314">line 1314</a>.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultEditorTemplates.cs
dougbu: thx; looks good

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultHtmlGenerator.cs
dougbu: Agreed, I meant to move that `foreach` as well.

dougbu:
:shipit:

sornaks:
Checked in. Thanks! https://github.com/aspnet/Mvc/commit/f2540f9ba19937f79c1a780ef12dd9a9d5cd84a6

PR #2955, opened 2015-08-11
Fix #2776 - Add implicit [BindRequired] for value type properties

### LINKED ISSUE #2776: Discuss adding server side validation to value type model properties by default.
Today in MVC6 having the following model:
```C#
public class Person
{
    public int Id { get; set; }
}
```
Will result in client side validation with 0 server side validation (unless specifying `BindingBehaviorAttribute` with `BindingBehavior.Required`). 

This differs from what happened in MVC5; model state would be invalid due to server side validation opting in `ValueType` properties. However, in WebAPI not providing a value for a `ValueType` property when model binding was 100% ok and did not result in any `ModelState` errors. 

Given the different behavior in the old MVC/WebAPI systems we should discuss the right approach for this in MVC6.

Related conversation that spurred this: https://github.com/aspnet/Mvc/issues/2720#issuecomment-117339050

/cc @rynowak @Eilon @dougbu

rynowak:
/cc @dougbu 

Main changes here are to tests which depending on the old behavior. I've tried to clean things up as much as possible while retaining the spirit of the tests.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
@@ -24,6 +24,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
         private ReadOnlyDictionary<object, object> _additionalValues;
         private ModelMetadata _elementMetadata;
         private bool _haveCalculatedElementMetadata;
+        private bool? _isBindingRequired;
         private bool? _isReadOnly;
         private bool? _isRequired;
         private ModelPropertyCollection _properties;
@@ -334,14 +335,24 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
         {
             get
             {
-                if (MetadataKind == ModelMetadataKind.Property)
+                if (!_isBindingRequired.HasValue)
                 {
-                    return BindingMetadata.IsBindingRequired;
+                    if (MetadataKind == ModelMetadataKind.Type)
+                    {
+                        _isBindingRequired = false;
+                    }
+                    else if (BindingMetadata.IsBindingRequired.HasValue)
+                    {
+                        _isBindingRequired = BindingMetadata.IsBindingRequired;
+                    }
+                    else
+                    {
+                        // Default to IsBindingRequired = true for value types.
+                        _isBindingRequired = !AllowsNullValue(ModelType);
+                    }
                 }
-                else
-                {
-                    return false;
-                }
+
+                return _isBindingRequired.Value;
             }
         }
 
@@ -401,6 +412,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
                     }
                     else
                     {
+                        // Default to IsRequired = true for value types.
                         _isRequired = !AllowsNullValue(ModelType);
                     }
                 }
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
@@ -381,13 +381,13 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
             {
                 if (!_isReadOnly.HasValue)
                 {
-                    if (BindingMetadata.IsReadOnly.HasValue)
-                    {
-                        _isReadOnly = BindingMetadata.IsReadOnly;
-                    }
-                    else if (MetadataKind == ModelMetadataKind.Type)
+                    if (MetadataKind == ModelMetadataKind.Type)
                     {
                         _isReadOnly = false;
+                    }
+                    else if (BindingMetadata.IsReadOnly.HasValue)
+                    {
+                        _isReadOnly = BindingMetadata.IsReadOnly;
                     }
                     else
                     {
--- a/test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/Metadata/DefaultModelMetadataTest.cs
+++ b/test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/Metadata/DefaultModelMetadataTest.cs
@@ -38,14 +38,14 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
             Assert.False(metadata.HasNonDefaultEditFormat);
             Assert.False(metadata.HideSurroundingHtml);
             Assert.True(metadata.HtmlEncode);
-            Assert.False(metadata.IsBindingRequired);
+            Assert.False(metadata.IsBindingRequired); // Defaults to false for reference types
             Assert.False(metadata.IsComplexType);
             Assert.False(metadata.IsCollectionType);
             Assert.False(metadata.IsEnum);
             Assert.False(metadata.IsFlagsEnum);
             Assert.False(metadata.IsNullableValueType);
             Assert.False(metadata.IsReadOnly);
-            Assert.False(metadata.IsRequired);
+            Assert.False(metadata.IsRequired); // Defaults to false for reference types
             Assert.True(metadata.ShowForDisplay);
             Assert.True(metadata.ShowForEdit);
 
@@ -171,6 +171,71 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
         private class DerivedDictionary : Dictionary<string, int>
         {
         }
+
+        [Theory]
+        [InlineData(typeof(string))]
+        [InlineData(typeof(int))]
+        public void IsBindingRequired_ReturnsFalse_ForTypes(Type modelType)
+        {
+            // Arrange
+            var provider = new EmptyModelMetadataProvider();
+            var detailsProvider = new EmptyCompositeMetadataDetailsProvider();
+
+            var key = ModelMetadataIdentity.ForType(modelType);
+            var cache = new DefaultMetadataDetails(key, new ModelAttributes(new object[0]));
+
+            var metadata = new DefaultModelMetadata(provider, detailsProvider, cache);
+
+            // Act
+            var isBindingRequired = metadata.IsBindingRequired;
+            
+            // Assert
+            Assert.False(isBindingRequired);
+        }
+
+        [Theory]
+        [InlineData(typeof(string))]
+        [InlineData(typeof(IDisposable))]
+        [InlineData(typeof(Nullable<int>))]
+        public void IsBindingRequired_ReturnsFalse_ForNullablePropertyTypes(Type modelType)
+        {
+            // Arrange
+            var provider = new EmptyModelMetadataProvider();
+            var detailsProvider = new EmptyCompositeMetadataDetailsProvider();
+
+            var key = ModelMetadataIdentity.ForProperty(modelType, "Test", typeof(string));
+            var cache = new DefaultMetadataDetails(key, new ModelAttributes(new object[0]));
+
+            var metadata = new DefaultModelMetadata(provider, detailsProvider, cache);
+
+            // Act
+            var isBindingRequired = metadata.IsBindingRequired;
+
+            // Assert
+            Assert.False(isBindingRequired);
+        }
+
+        [Theory]
+        [InlineData(typeof(int))]
+        [InlineData(typeof(DayOfWeek))]
+        public void IsBindingRequired_ReturnsTrue_ForNonNullablePropertyTypes(Type modelType)
+        {
+            // Arrange
+            var provider = new EmptyModelMetadataProvider();
+            var detailsProvider = new EmptyCompositeMetadataDetailsProvider();
+
+            var key = ModelMetadataIdentity.ForProperty(modelType, "Test", typeof(string));
+            var cache = new DefaultMetadataDetails(key, new ModelAttributes(new object[0]));
+
+            var metadata = new DefaultModelMetadata(provider, detailsProvider, cache);
+
+            // Act
+            var isBindingRequired = metadata.IsBindingRequired;
+
+            // Assert
+            Assert.True(isBindingRequired);
+        }
+
 
         [Theory]
         [InlineData(typeof(string))]
--- a/test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/Metadata/DefaultModelMetadataTest.cs
+++ b/test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/Metadata/DefaultModelMetadataTest.cs
@@ -175,6 +175,31 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
         [Theory]
         [InlineData(typeof(string))]
         [InlineData(typeof(int))]
+        public void IsBindingRequired_ReturnsTrue_ForTypes(Type modelType)
+        {
+            // Arrange
+            var provider = new EmptyModelMetadataProvider();
+            var detailsProvider = new EmptyCompositeMetadataDetailsProvider();
+
+            var key = ModelMetadataIdentity.ForType(modelType);
+            var cache = new DefaultMetadataDetails(key, new ModelAttributes(new object[0]));
+            cache.BindingMetadata = new BindingMetadata()
+            {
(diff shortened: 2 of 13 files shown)


test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/MutableObjectModelBinderTest.cs
```diff
@@ -762,8 +762,7 @@ public void GetPropertyValidationInfo_WithIndexerProperties_Succeeds()
             // Arrange
             var bindingContext = new ModelBindingContext
             {
-                // Any type, even an otherwise-simple POCO with an indexer property, would do here.
-                ModelMetadata = GetMetadataForType(typeof(List<Person>)),
+                ModelMetadata = GetMetadataForType(typeof(PersonCollection)),
```
rynowak: This was broken because normally MOMB wouldn't handle `List<Person>`. It was trying to bind the `Capacity` property (among others). The test was really about indexers, so I made a simpler model class.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/MutableObjectModelBinderTest.cs
```diff
@@ -942,53 +941,15 @@ public void ProcessDto_ValueTypePropertyWithBindRequired_SetToNull_CapturesExcep
         }
 
         [Fact]
-        [ReplaceCulture]
-        public void ProcessDto_MissingDataForRequiredFields_NoErrors()
```
rynowak: Deleted this one because it was just redundant. Retrofitted other test to cover important scenarios:
- value type with no value -> error
- value type with optional with no value -> no error
- nullable value type with no value -> no error

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/MutableObjectModelBinderTest.cs
```diff
@@ -1763,10 +1728,23 @@ private class EmptyModel
...
+            public int ValueTypeRequired { get; set; }
+        }
+
+        private class NullableValueTypeProperty
+        {
+            [BindingBehavior(BindingBehavior.Optional)]
+            public int? NullableValueType { get; set; }
+        }
+
         private class Person
```
rynowak: This was used in plenty of other tests, so I updated the model to behave like those tests expect it to.

test/Microsoft.AspNet.Mvc.FunctionalTests/ModelBindingFromQueryTest.cs
```diff
@@ -122,8 +122,12 @@ public class ModelBindingFromQueryTest
...
             var body = await response.Content.ReadAsStringAsync();
             var result = JsonConvert.DeserializeObject<Result>(body);
-            var error = Assert.Single(result.ModelStateErrors);
-            Assert.Equal("TestEmployees[0].EmployeeId", error);
+
+            Assert.Collection(
+                result.ModelStateErrors,
+                e => Assert.Equal("TestEmployees[0].EmployeeId", e),
+                e => Assert.Equal("TestEmployees[0].EmployeeTaxId", e),
+                e => Assert.Equal("TestEmployees[0].Age", e));
```
rynowak: This model is unfortunately used **all over**. It was simpler to just expect the errors you'd get from this set of inputs. This test was really *about* error messages anyway.

test/Microsoft.AspNet.Mvc.FunctionalTests/ModelBindingTest.cs
```diff
@@ -2218,7 +2218,7 @@ public async Task
             // Arrange
             var server = TestHelper.CreateServer(_app, SiteName, _configureServices);
             var client = server.CreateClient();
-            var url = "http://localhost/TryUpdateModel/TryUpdateModel_ClearsModelStateEntries";
+            var url = "http://localhost/TryUpdateModel/TryUpdateModel_ClearsModelStateEntries?id=5&price=1";
```
rynowak: This test was about passing valid input, so I made it valid.

test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Customer.Index.html
```diff
@@ -27,10 +27,10 @@
         <div>
             <label class="order" for="Gender">Gender</label>
             <input type="radio" value="Male" data-val="true" data-val-required="The Gender field is required." id="Gender" name="Gender" /> Male
-            <input type="radio" value="Female" id="Gender" name="Gender" /> Female
+            <input type="radio" value="Female" checked="checked" id="Gender" name="Gender" /> Female
             <span class="field-validation-valid" data-valmsg-for="Gender" data-valmsg-replace="true"></span>
         </div>
-        <div class="order validation-summary-errors" data-valmsg-summary="true"><ul><li>The value &#x27;&#x27; is invalid.</li>
+        <div class="order validation-summary-errors" data-valmsg-summary="true"><ul><li>The field Number must be between 1 and 100.</li>
```
rynowak: These changes are due to passing input that's *as-valid* as it was before my change.

test/Microsoft.AspNet.Mvc.IntegrationTests/BodyValidationIntegrationTests.cs
```diff
@@ -108,7 +108,7 @@ private class Address
         private class Person4
         {
             [FromBody]
-            [Required]
+            [BindingBehavior(BindingBehavior.Optional)]
```
rynowak: Changed an assortment of properties to have the behavior expected by integration tests (this change and all of the below).

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/Metadata/DefaultModelMetadataTest.cs
```diff
@@ -174,6 +174,71 @@ private class DerivedDictionary : Dictionary<string, int>
...
+            // Assert
+            Assert.True(isBindingRequired);
+        }
+
+
+        [Theory]
+        [InlineData(typeof(string))]
         [InlineData(typeof(IDisposable))]
         [InlineData(typeof(Nullable<int>))]
         public void IsRequired_ReturnsFalse_ForNullableTypes(Type modelType)
```
rynowak: As discussed, taking a note to play with this, and see if we have an issue.

rynowak:
ping @dougbu

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
```diff
@@ -334,14 +335,24 @@ public override bool IsBindingRequired
...
-                {
-                    return BindingMetadata.IsBindingRequired;
-                }
-                else
+                if (!_isBindingRequired.HasValue)
                 {
-                    return false;
+                    if (MetadataKind == ModelMetadataKind.Type)
+                    {
+                        _isBindingRequired = false;
```
dougbu: So a metadata provider that _wants_ to set `IsBindingRequired` for a `Type` can't? Realize the old code was doing something similar (ignoring the lack of parameter support, 'natch). But I'm wondering if it would be better to have this logic in our metadata providers rather than ignore what customer providers may do...

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/MutableObjectModelBinderTest.cs
```diff
@@ -942,53 +941,15 @@ public void ProcessDto_ValueTypePropertyWithBindRequired_SetToNull_CapturesExcep
...
-        public void ProcessDto_ValueTypeProperty_WithRequiredAttribute_SetToNull_NoError()
+        public void ProcessDto_ValueTypeProperty_WithBindingOptional_NoValueSet_NoError()
         {
             // Arrange
-            var model = new ModelWithRequired();
+            var model = new BindingOptionalProperty();
             var containerMetadata = GetMetadataForType(model.GetType());
             var bindingContext = CreateContext(containerMetadata, model);
-
+
```
dougbu: Get thee the trailing whitespace visualizer (which also nukes).

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
rynowak: Would would that even mean semantically? We have a few MM properties like this and aren't meaningful for a type.

test/Microsoft.AspNet.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
```diff
@@ -26,6 +26,7 @@ public class MutableObjectModelBinderIntegrationTest
 
         private class Order1
```
dougbu: Does this class include any tests that'll hit `IsBindingRequired` violations now?

test/Microsoft.AspNet.Mvc.IntegrationTests/CollectionModelBinderIntegrationTest.cs
```diff
@@ -548,6 +548,7 @@ private class Person4
 
         private class Address4
```
dougbu: Does this class include any tests that'll hit `IsBindingRequired` violations now?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
dougbu: The set which special case the metadata kind is `IsBindingAllowed`, `IsBindingRequired` and `IsReadOnly`.

I agree that one reading of the metadata on a `Type` would be "all properties / parameters of this `Type` have XYZ" but that's not how the rest of the system reads the information. So my question is more like "Could customers want to have this mean something on a `Type`?"

dougbu: Separately, `IsBindingAllowed` has the same problem with parameter metadata as `IsBindingRequired`. Should probably be fixed while we're looking at this.

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
rynowak: Discussed in person: The conclusion here is that we're not doing anything to support these scenarios. All of these properties `IsBindingAllowed`, `IsBindingRequired`, `IsReadOnly`, `IsRequired` all should return false for types no matter what the user does.

rynowak: Added a test for `IsBindingAllowed`, and improved unit tests for the others

test/Microsoft.AspNet.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
rynowak: MutableObjectModelBinder_WithNestedRequiredProperty_WithPartialData_GetsErrors

test/Microsoft.AspNet.Mvc.IntegrationTests/CollectionModelBinderIntegrationTest.cs
rynowak: CollectionModelBinder_BindsListOfComplexType_WithRequiredProperty_WithPrefix_PartialData

rynowak:
Updated

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
dougbu: You've added the tests but `IsBindingAllowed` is still implemented differently. Should we have a follow-up issue to make `BindingMetadata.IsBindingAllowed` nullable and align `IsBindingAllowed` here with the new `IsBindingRequired`?

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/Metadata/DefaultModelMetadataTest.cs
```diff
@@ -38,14 +38,14 @@ public void DefaultValues()
             Assert.False(metadata.HasNonDefaultEditFormat);
             Assert.False(metadata.HideSurroundingHtml);
             Assert.True(metadata.HtmlEncode);
-            Assert.False(metadata.IsBindingRequired);
+            Assert.False(metadata.IsBindingRequired); // Defaults to false for reference types
```
dougbu: Should check `IsBindingAllowed` default as well.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/Metadata/DefaultModelMetadataTest.cs
```diff
@@ -174,6 +174,99 @@ private class DerivedDictionary : Dictionary<string, int>
 
         [Theory]
         [InlineData(typeof(string))]
+        [InlineData(typeof(int))]
+        public void IsBindingRequired_ReturnsTrue_ForTypes(Type modelType)
```
dougbu: `IsBindingAllowed_...`

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/Metadata/DefaultModelMetadataTest.cs
```diff
@@ -174,6 +174,99 @@ private class DerivedDictionary : Dictionary<string, int>
...
+            var cache = new DefaultMetadataDetails(key, new ModelAttributes(new object[0]));
+            cache.BindingMetadata = new BindingMetadata()
+            {
+                IsBindingAllowed = false, // Will be ignored.
+            };
+
+            var metadata = new DefaultModelMetadata(provider, detailsProvider, cache);
+
+            // Act
+            var isBindingRequired = metadata.IsBindingAllowed;
```
dougbu: `isBindingAllowed`

dougbu:
:shipit: after fixing cut 'n paste issues

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
rynowak: I don't think so. `IsBindingRequired` is special because we want to know if the user set a value or not. We don't care in this case.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/Metadata/DefaultModelMetadataTest.cs
rynowak: sure

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/Metadata/DefaultModelMetadataTest.cs
rynowak: sure

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/Metadata/DefaultModelMetadataTest.cs
rynowak: yeah

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/Metadata/DefaultModelMetadataTest.cs
rynowak: Can't be changed. HTML helper tests go wacky if you change this behavior.

PR #2956, opened 2015-08-11
React: Enable input, image and url resolution `TagHelper`s to be written as void elements.

NTaylorMullen:
- Razor rendering now understands `TagMode` which allows void elements to be rendered.
- Added a `TagStructure.WithoutEnd` bit to `InputTagHelper`, `ImageTagHelper` and `UrlResolutionTagHelper`. This will allow users to write the various elements in the void format. Used the HTML5 spec to determine the elements appropriate.
- Added tests to ensure `TagMode.StartTagOnly` is rendered properly.
- Updated a few functional tests to showcase the void element formats.

aspnet/Razor#450

Corresponding Razor PR: https://github.com/aspnet/Razor/pull/480

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
@@ -336,7 +336,7 @@ namespace Microsoft.AspNet.Mvc.Razor
                     }
                 }
 
-                if (tagHelperOutput.SelfClosing)
+                if (tagHelperOutput.TagMode == TagMode.SelfClosing)
                 {
                     writer.Write(" /");
                 }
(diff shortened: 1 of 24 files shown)


NTaylorMullen:
Added the ability for `InputTagHelper`s generated input tags to be consistent with `output.TagMode`.

dougbu:
Need to update `LinkTagHelper` as well because `<link>` is also a [void element](http://www.w3.org/TR/html5/syntax.html#void-elements).

dougbu:
:watch:

NTaylorMullen:
Hmm could have swore I did that. Somehow got lost in translation.

Updated.

dougbu:
:shipit: once the CI clears up

ajaybhargavb:
:shipit:

PR #2961, opened 2015-08-12
Using metadata to parse information required by tag helper in precompilation

pranavkm:


pranavkm:
cc @davidfowl \ @rynowak \ @NTaylorMullen \ @dougbu

src/Microsoft.AspNet.Mvc.Razor/Precompilation/PrecompilationTagHelperTypeResolver.cs
```diff
@@ -0,0 +1,138 @@
...
+using System.Reflection;
+using Microsoft.AspNet.Razor.Runtime.TagHelpers;
+using Microsoft.CodeAnalysis;
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.AspNet.Mvc.Razor.Precompilation
+{
+    /// <summary>
+    /// <see cref="TagHelperDescriptorResolver"/> used during Razor precompilation.
+    /// </summary>
```
rynowak: not exactly right

src/Microsoft.AspNet.Mvc.Razor/Precompilation/PrecompilationTagHelperTypeResolver.cs
```diff
@@ -0,0 +1,138 @@
...
+                {
+                    _openGenericDictionaryTypeSymbol = _compilation.GetTypeByMetadataName(
+                        typeof(IDictionary<,>).FullName);
+
+                    Debug.Assert(_openGenericDictionaryTypeSymbol != null);
+                }
+
+                return _openGenericDictionaryTypeSymbol;
+            }
+        }
```
rynowak: What's the advantage of making this lazy?

src/Microsoft.AspNet.Mvc.Razor/Precompilation/PrecompilationTagHelperTypeResolver.cs
```diff
@@ -0,0 +1,138 @@
...
+                            StringComparison.Ordinal))
+                    {
+                        return GetExportedTypes(compilationReference.Compilation.Assembly);
+                    }
+
+                    var assemblyReference = reference as PortableExecutableReference;
+                    if (assemblyReference != null)
+                    {
+                        var assemblyOrModuleSymbol = _compilation.GetAssemblyOrModuleSymbol(reference);
+                        if (assemblyOrModuleSymbol != null &&
```
rynowak: What does the `else` mean here?

src/Microsoft.AspNet.Mvc.Razor/Precompilation/PrecompilationTagHelperTypeResolver.cs
```diff
@@ -0,0 +1,138 @@
...
+        private List<ITypeInfo> GetExportedTypes(IAssemblySymbol assembly)
+        {
+            var exportedTypes = new List<ITypeInfo>();
+            GetExportedTypes(assembly.GlobalNamespace, exportedTypes);
+            return exportedTypes;
+        }
+
+        private void GetExportedTypes(INamespaceSymbol namespaceSymbol, List<ITypeInfo> exportedTypes)
+        {
+            foreach (var type in namespaceSymbol.GetTypeMembers().Where(IsPublicType))
```
rynowak: Will we run into trouble with `InternalsVisibleTo`?

src/Microsoft.AspNet.Mvc.Razor/Precompilation/PrecompilationTagHelperTypeResolver.cs
```diff
@@ -0,0 +1,138 @@
...
+                }
+
+                return Enumerable.Empty<ITypeInfo>();
+            }
+        }
+
+        private List<ITypeInfo> GetExportedTypes(IAssemblySymbol assembly)
+        {
+            var exportedTypes = new List<ITypeInfo>();
+            GetExportedTypes(assembly.GlobalNamespace, exportedTypes);
```
rynowak: Are there other namespaces to consider other than `global`? Seems odd, but roughly consistent with how DIA handle symbols

rynowak: Ah nvm, I see it's recursive.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/SymbolBasedPropertyInfo.cs
```diff
@@ -0,0 +1,43 @@
...
+            _openGenericDictionaryTypeSymbol = openGenericDictionaryTypeSymbol;
+        }
+
+        public bool HasPublicGetter =>
+            _propertySymbol.GetMethod != null &&
+            _propertySymbol.GetMethod.DeclaredAccessibility == Accessibility.Public;
+
+        public bool HasPublicSetter =>
+            _propertySymbol.SetMethod != null &&
+            _propertySymbol.SetMethod.DeclaredAccessibility == Accessibility.Public;
```
rynowak: I still think these are yucky. 24 PranavPoints awarded.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/SymbolBasedPropertyInfo.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using Microsoft.AspNet.Razor.Runtime.TagHelpers;
+using Microsoft.CodeAnalysis;
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.AspNet.Mvc.Razor.Precompilation
+{
+    [DebuggerDisplay("{Name, PropertyType}")]
+    public class SymbolBasedPropertyInfo : IPropertyInfo
```
rynowak: This seems like something that would be in the box.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/SymbolBasedTypeInfo.cs
```diff
@@ -0,0 +1,99 @@
...
+            get
+            {
+                if (_fullName == null)
+                {
+                    _fullName = SymbolUtilities.GetFullName(_type);
+                }
+
+                return _fullName;
+            }
+        }
```
rynowak: What's the benefit to making this lazy?

src/Microsoft.AspNet.Mvc.Razor/Precompilation/SymbolBasedTypeInfo.cs
```diff
@@ -0,0 +1,99 @@
...
+                .Cast<IPropertySymbol>()
+                .Select(property => new SymbolBasedPropertyInfo(property, _openGenericDictionaryTypeSymbol));
+
+        public IEnumerable<TAttribute> GetCustomAttributes<TAttribute>() where TAttribute : Attribute =>
+            SymbolUtilities.GetCustomAttributes<TAttribute>(_type);
+
+        public bool ImplementsInterface(System.Reflection.TypeInfo interfaceType) =>
+            _type
+                .AllInterfaces
+                .Any(implementedInteface =>
```
rynowak: `implementedInterface`

src/Microsoft.AspNet.Mvc.Razor/Precompilation/SymbolBasedTypeInfo.cs
```diff
@@ -0,0 +1,99 @@
...
+                    SymbolUtilities.GetFullName(dictionaryInterface.TypeArguments[0]),
+                    SymbolUtilities.GetFullName(dictionaryInterface.TypeArguments[1])
+                };
+            }
+
+            return null;
+        }
+
+        private bool IsDictionaryType(INamedTypeSymbol implementedInteface) =>
+            implementedInteface.ConstructedFrom == _openGenericDictionaryTypeSymbol;
```
rynowak: `implementedInteface` again.

Wouldn't it be simpler to just use the names here to look for dictionary rather than passing the symbol around?

src/Microsoft.AspNet.Mvc.Razor/Precompilation/SymbolUtilities.cs
```diff
@@ -0,0 +1,93 @@
...
+using System.Text;
+#if DNXCORE50
+using System.Reflection;
+#endif
+using Microsoft.CodeAnalysis;
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.AspNet.Mvc.Razor.Precompilation
+{
+    public static class SymbolUtilities
```
rynowak: Gross

src/Microsoft.AspNet.Mvc.Razor/Precompilation/SymbolUtilities.cs
```diff
@@ -0,0 +1,93 @@
...
+            var constructorArguments = GetConstructorArguments(attributeData);
+            var attribute = (TAttribute)Activator.CreateInstance(typeof(TAttribute), constructorArguments);
+
+            if (attributeData.NamedArguments.Length > 0)
+            {
+                var helpers = PropertyHelper.GetVisibleProperties(attribute);
+                foreach (var item in attributeData.NamedArguments)
+                {
+                    var helper = helpers.FirstOrDefault(h => string.Equals(h.Name, item.Key, StringComparison.Ordinal));
+                    helper?.SetValue(attribute, item.Value.Value);
```
rynowak: wouldn't this be a bug if there's no match?

src/Microsoft.AspNet.Mvc.Razor/Precompilation/SymbolUtilities.cs
```diff
@@ -0,0 +1,93 @@
...
+                var constructorArgument = attributeData.ConstructorArguments[i];
+                switch  (constructorArgument.Kind)
+                {
+                    case TypedConstantKind.Enum:
+                        var enumType = Type.GetType(
+                            $"{constructorArgument.Type},{constructorArgument.Type.ContainingAssembly.Identity}", 
+                            throwOnError: true);
+                        value = Enum.Parse(enumType, constructorArgument.Value.ToString());
+                        break;
+                    case TypedConstantKind.Primitive:
```
rynowak: Does this include strings?

test/Microsoft.AspNet.Mvc.Razor.Test/Precompilation/PrecompilationTagHelperTypeResolverTest.cs
```diff
@@ -0,0 +1,230 @@
...
+    public class PrecompilationTagHelperTypeResolverTest
+    {
+        private static readonly string TypeNamespace = typeof(PrecompilationTagHelperTypeResolverTest).Namespace;
+        private static readonly Assembly ExecutingAssembly = typeof(PrecompilationTagHelperTypeResolverTest).Assembly;
+        private static readonly TypeInfo TagHelperTypeInfo = typeof(ITagHelper).GetTypeInfo();
+
+        [Theory]
+        [InlineData("TypeDerivingFromITagHelper",
+            "Microsoft.AspNet.Mvc.Razor.Precompilation.TypeDerivingFromITagHelper")]
+        [InlineData("TypeInGlobalNamespace", "TypeInGlobalNamespace")]
```
rynowak: `nameof` all the things!

rynowak: oh wait, you can't. :laughing:

src/Microsoft.AspNet.Mvc.Razor/Precompilation/PrecompilationTagHelperTypeResolver.cs
pranavkm: Probably shouldn't be.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/PrecompilationTagHelperTypeResolver.cs
pranavkm: I haven't seen a module being returned with the set of samples we have. Some of this needs hardening (especially around assemblies\types with errors)

src/Microsoft.AspNet.Mvc.Razor/Precompilation/PrecompilationTagHelperTypeResolver.cs
pranavkm: Haven't tried. Should be easy to add a test for it.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/SymbolBasedTypeInfo.cs
pranavkm: It's not necessarily always called - so we just avoid doing extra string concats.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/PrecompilationTagHelperTypeResolver.cs
```diff
@@ -0,0 +1,138 @@
...
+
+        public PrecompilationTagHelperTypeResolver([NotNull] CodeAnalysis.Compilation compilation)
+        {
+            _compilation = compilation;
+        }
+
+        /// <inheritdoc />
+        protected override IEnumerable<ITypeInfo> GetExportedTypes([NotNull] AssemblyName assemblyName)
+        {
+            lock (_assemblyLookup)
```
NTaylorMullen: `_assemblyLookupLock`

src/Microsoft.AspNet.Mvc.Razor/Precompilation/SymbolUtilities.cs
pranavkm: probably. We might want to throw just so that we know what's happening here. I need to verify how this work with more esoteric kinds of attributes \ derived ones etc.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/PrecompilationTagHelperTypeResolver.cs
```diff
@@ -0,0 +1,138 @@
...
+            {
+                IEnumerable<ITypeInfo> result;
+                if (!_assemblyLookup.TryGetValue(assemblyName.Name, out result))
```
NTaylorMullen: How often will this type be re-created? Is it once per change of a project or is it once a DTH lifetime?

src/Microsoft.AspNet.Mvc.Razor/Precompilation/SymbolUtilities.cs
pranavkm: Yup. Need to add a couple of scenarios for `Type` - I think we have to do `GetType` there.

test/Microsoft.AspNet.Mvc.Razor.Test/Precompilation/PrecompilationTagHelperTypeResolverTest.cs
pranavkm: I am going to try and see if we can get it to both compile and be used as a resource.That way we can run same tests against `RuntimeTypeInfo` and `SymbolsBasedTypeInfo` to verify expectations match.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/SymbolBasedTypeInfo.cs
pranavkm: Possibly.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/SymbolUtilities.cs
pranavkm: Might need to change the name. We already have a `SymbolUtility` which deals with pdbs.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/PrecompilationTagHelperTypeResolver.cs
```diff
@@ -0,0 +1,138 @@
...
+                        string.Equals(
+                            compilationReference.Compilation.AssemblyName,
+                            assemblyName,
+                            StringComparison.Ordinal))
+                    {
+                        return GetExportedTypes(compilationReference.Compilation.Assembly);
+                    }
+
+                    var assemblyReference = reference as PortableExecutableReference;
+                    if (assemblyReference != null)
```
NTaylorMullen: `if (reference is PortableExecutableReference)`

src/Microsoft.AspNet.Mvc.Razor/Precompilation/PrecompilationTagHelperTypeResolver.cs
pranavkm: It's once per precompilation. Although we don't necessarily do any work until we determine a view actually needs to be recompiled (i.e. view compilation gets cached the same runtime compilation does).

NTaylorMullen: Alright, awesome then. Wont have to worry about invalid cache.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/PrecompilationTagHelperTypeResolver.cs
```diff
@@ -0,0 +1,138 @@
...
+
+            foreach (var subNamespace in namespaceSymbol.GetNamespaceMembers())
+            {
+                GetExportedTypes(subNamespace, exportedTypes);
+            }
+        }
+
+        private static bool IsPublicType(INamedTypeSymbol type) =>
+            type.TypeKind == TypeKind.Class &&
+            type.DeclaredAccessibility == Accessibility.Public;
```
NTaylorMullen: This doesn't seem to have the same restrictions as the Razor `TagHelperTypeResolver` unless Roslyn's `DeclaredAccessibility` is super overloaded:

https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperTypeResolver.cs#L91-L95

NTaylorMullen: Ahh, nm, different part of the system. Ignore this comment.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/PrecompilationTagHelperTypeResolver.cs
```diff
@@ -0,0 +1,138 @@
...
+            var exportedTypes = new List<ITypeInfo>();
+            GetExportedTypes(assembly.GlobalNamespace, exportedTypes);
+            return exportedTypes;
+        }
+
+        private void GetExportedTypes(INamespaceSymbol namespaceSymbol, List<ITypeInfo> exportedTypes)
+        {
+            foreach (var type in namespaceSymbol.GetTypeMembers().Where(IsPublicType))
+            {
+                exportedTypes.Add(new SymbolBasedTypeInfo(type, OpenGenericDictionaryType));
```
NTaylorMullen: Should refactor this out. Shouldn't need to pas `OpenGenericDictionaryType` for every `SymbolBasedTypeInfo`s.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/PrecompilationTagHelperTypeResolver.cs
```diff
@@ -0,0 +1,138 @@
...
+
+                // TODO: Nested types get eventually filtered out. Could we make it part of the contract?
```
NTaylorMullen: I do like making it part of the contract. Could change the name to be `GetTopLevelExportedTypes` or something descriptive. This part sucks :sob:

src/Microsoft.AspNet.Mvc.Razor/Precompilation/SymbolBasedTypeInfo.cs
```diff
@@ -0,0 +1,99 @@
...
+{
+    [DebuggerDisplay("{FullName}")]
+    public class SymbolBasedTypeInfo : ITypeInfo
+    {
+        private readonly ITypeSymbol _type;
+        private readonly INamedTypeSymbol _openGenericDictionaryTypeSymbol;
+        private string _fullName;
+
+        public SymbolBasedTypeInfo(
+            ITypeSymbol type,
```
NTaylorMullen: `[NotNull]` for both of these

src/Microsoft.AspNet.Mvc.Razor/Precompilation/SymbolUtilities.cs
```diff
@@ -0,0 +1,93 @@
...
+                .Select(a => CreateAttribute<TAttribute>(a));
+        }
+
+        public static string GetFullName(ITypeSymbol typeSymbol)
+        {
+            var nameBuilder = new StringBuilder(typeSymbol.Name);
+            if (typeSymbol.ContainingType != null)
+            {
+                nameBuilder.Insert(0, '+')
+                    .Insert(0, typeSymbol.ContainingType.Name);
```
NTaylorMullen: `.Insert(1, typeSymbol.ContainingType.Name);` or just do append append

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationService.cs
```diff
@@ -177,12 +181,23 @@ private static string GetFilePath(string relativePath, Diagnostic diagnostic)
 
         private List<MetadataReference> GetApplicationReferences()
         {
+            return GetApplicationReferences(
+                _metadataFileCache,
+                _libraryExporter,
+                _environment.ApplicationName);
+        }
+
+        internal static List<MetadataReference> GetApplicationReferences(
```
dougbu: Is this `internal` for testing or some other reason?

src/Microsoft.AspNet.Mvc.Razor/Precompilation/SymbolUtilities.cs
```diff
@@ -0,0 +1,93 @@
...
+            var attributeFullName = typeof(TAttribute).FullName;
+            return attributes
+                .Where(
+                    attribute => string.Equals(GetFullName(attribute.AttributeClass), attributeFullName, StringComparison.Ordinal))
+                .Select(a => CreateAttribute<TAttribute>(a));
+        }
+
+        public static string GetFullName(ITypeSymbol typeSymbol)
+        {
+            var nameBuilder = new StringBuilder(typeSymbol.Name);
```
NTaylorMullen: Talked in person about robustness of this method (doesn't work for every scenario) and it's expected to not be 100% fool proof.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/PrecompilationTagHelperTypeResolver.cs
```diff
@@ -0,0 +1,138 @@
...
+{
+    /// <summary>
+    /// <see cref="TagHelperDescriptorResolver"/> used during Razor precompilation.
+    /// </summary>
+    public class PrecompilationTagHelperTypeResolver : TagHelperTypeResolver
+    {
+        private readonly object _assemblyLookupLock = new object();
+        private readonly Dictionary<string, IEnumerable<ITypeInfo>> _assemblyLookup
+            = new Dictionary<string, IEnumerable<ITypeInfo>>(StringComparer.OrdinalIgnoreCase);
+        private readonly CodeAnalysis.Compilation _compilation;
```
dougbu: Curious: What else is named `Compilation`?

src/Microsoft.AspNet.Mvc.Razor/Precompilation/PrecompilationTagHelperTypeResolver.cs
dougbu: Why add the nested types if they'll be filtered out later? Seems like extra work for to match a contract missing a doc comment or two...

src/Microsoft.AspNet.Mvc.Razor/Precompilation/PrecompilationTagHelperTypeResolver.cs
pranavkm: There's a `Microsoft.AspNet.Mvc.Razor.Compilation` namespace that collides with this.

test/Microsoft.AspNet.Mvc.Razor.Test/Precompilation/Files/TypeInGlobalNamespace.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Razor.Runtime.TagHelpers;
+
+public class TypeInGlobalNamespace : ITagHelper
```
dougbu: Make one of these a little more difficult for Roslyn and derive from `TagHelper` instead.

src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationService.cs
pranavkm: It's internal for testing. I'll add a comment.

test/Microsoft.AspNet.Mvc.Razor.Test/Precompilation/PrecompilationTagHelperTypeResolverTest.cs
```diff
@@ -0,0 +1,230 @@
...
+        {
+            // Arrange
+            var compilation = GetCompilation("TypeWithDictionaryProperties.cs");
+            var tagHelperResolver = new PrecompilationTagHelperTypeResolver(compilation);
+
+            // Act
+            var exportedTypes = tagHelperResolver.GetExportedTypes(compilation.AssemblyName);
+
+            // Assert
+            var type = exportedTypes.First();
```
dougbu: `Assert.Single()` ?

src/Microsoft.AspNet.Mvc.Razor/Precompilation/SymbolBasedPropertyInfo.cs
dougbu: Also why are these evaluated on every access? That is, why not
``` c#
public HasPublicSetter { get; } =
    _propertySymbol.SetMethod != null &&
    _propertySymbol.SetMethod.DeclaredAccessibility == Accessibility.Public;
```

Could also get rid of the `_openGenericDictionaryTypeSymbol` field and set `PropertyType` directly in the constructor. (No need to create multiple `SymbolBasedTypeInfo` instances there.)

src/Microsoft.AspNet.Mvc.Razor/Precompilation/SymbolBasedPropertyInfo.cs
```diff
@@ -0,0 +1,43 @@
...
+            _propertySymbol.SetMethod != null &&
+            _propertySymbol.SetMethod.DeclaredAccessibility == Accessibility.Public;
+
+        public string Name => _propertySymbol.Name;
+
+        public ITypeInfo PropertyType =>
+            new SymbolBasedTypeInfo(_propertySymbol.Type, _openGenericDictionaryTypeSymbol);
+
+        public IEnumerable<TAttribute> GetCustomAttributes<TAttribute>() where TAttribute : Attribute
+            => SymbolUtilities.GetCustomAttributes<TAttribute>(_propertySymbol);
```
dougbu: How expensive do you expect the lookup to be? May be worth keeping the result around if this method is called multiple times per `SymbolBasedPropertyInfo` instance.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/SymbolBasedTypeInfo.cs
```diff
@@ -0,0 +1,99 @@
...
+        public bool IsPublic => _type.DeclaredAccessibility == Accessibility.Public;
+
+        public string Name => _type.Name;
+
+        public IEnumerable<IPropertyInfo> Properties =>
+            _type.GetMembers()
+                .Where(member => member.Kind == SymbolKind.Property &&
+                    member.DeclaredAccessibility == Accessibility.Public)
+                .Cast<IPropertySymbol>()
+                .Select(property => new SymbolBasedPropertyInfo(property, _openGenericDictionaryTypeSymbol));
```
dougbu: I could imagine this getting expensive but, again, why evaluate the remaining properties on every get?

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.Razor/Precompilation/SymbolBasedTypeInfo.cs
pranavkm: Seems like there isn't a way to get back to the `Compilation` object from the symbol. Using a symbol, we are only allowed to do get GetType on a per assembly basis - which is just more work.

pranavkm:
Updated

src/Microsoft.AspNet.Mvc.Razor/Precompilation/CodeAnalysisAttributeUtilities.cs
```diff
@@ -0,0 +1,256 @@
...
+                    .Select(attribute => CreateAttribute<TAttribute>(symbolLookup, attribute))
+                    .ToArray();
+            }
+
+            return Enumerable.Empty<TAttribute>();
+        }
+
+        private static TAttribute CreateAttribute<TAttribute>(
+            CodeAnalysisSymbolLookupCache symbolLookup,
+            AttributeData attributeData)
```
NTaylorMullen: nit: Swap order of params for consistency.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/CodeAnalysisAttributeUtilities.cs
```diff
@@ -0,0 +1,256 @@
...
+                        matchInfo.Constructor,
+                        parameters),
+                    argsParameter);
+
+            return lambda.Compile();
+        }
+
+        private static ConstructorMatchInfo MatchConstructor(
+            Type type,
+            AttributeData attributeData,
```
NTaylorMullen: This method just cares about the constructor arguments.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/CodeAnalysisAttributeUtilities.cs
```diff
@@ -0,0 +1,256 @@
...
+                            index);
+                    }
+                    value = values;
+                    break;
+                default:
+                    throw new NotSupportedException(
+                        Resources.FormatCodeAnalysis_TypeConstantKindNotSupported(constructorArgument.Kind));
+            }
+
+            return value;
```
NTaylorMullen: Instead of all the breaks in the `switch` just return directly.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/CodeAnalysisAttributeUtilities.cs
```diff
@@ -0,0 +1,256 @@
...
+                        item.Value, 
+                        symbolLookup);
+                    helper.SetValue(attribute, propertyValue);
+                }
+            }
+
+            return attribute;
+        }
+
+        private static Func<object[], Attribute> MakeFastConstructorInvoker(ConstructorMatchInfo matchInfo)
```
NTaylorMullen: This method pulled from somewhere or you write it from scratch?

src/Microsoft.AspNet.Mvc.Razor/Precompilation/CodeAnalysisAttributeUtilities.cs
```diff
@@ -0,0 +1,256 @@
...
+                if (runtimeParameters.Length != symbolConstructorArguments.Length)
+                {
+                    continue;
+                }
+
+                var parametersMatched = true;
+                for (var index = 0; index < runtimeParameters.Length; index++)
+                {
+                    var runtimeParameter = runtimeParameters[index].ParameterType;
+                    if (symbolConstructorArguments[index].Kind == TypedConstantKind.Array &&
```
NTaylorMullen: You're comparing `index` for `index` here. Will this handle named arguments appropriately? Aka

```C#
[SomeAttribute(secondAttribute: 2, firstAttribute: 1)]
```

src/Microsoft.AspNet.Mvc.Razor/Precompilation/CodeAnalysisSymbolBasedPropertyInfo.cs
```diff
@@ -0,0 +1,69 @@
...
+        public string Name => _propertySymbol.MetadataName;
+
+        /// <inheritdoc />
+        public ITypeInfo PropertyType { get; }
+
+        /// <inheritdoc />
+        public IEnumerable<TAttribute> GetCustomAttributes<TAttribute>() 
+            where TAttribute : Attribute
+        {
+            return CodeAnalysisAttributeUtilities.GetCustomAttributes<TAttribute>(_propertySymbol, _symbolLookup);
```
NTaylorMullen: Be Lazy

src/Microsoft.AspNet.Mvc.Razor/Precompilation/CodeAnalysisSymbolBasedTypeInfo.cs
```diff
@@ -0,0 +1,251 @@
...
+        public string Name => _type.MetadataName;
+
+        /// <inheritdoc />
+        public IEnumerable<IPropertyInfo> Properties
+        {
+            get
+            {
+                if (_properties == null)
+                {
+                    _properties = PopulateProperties();
```
NTaylorMullen: `GetProperties();`

src/Microsoft.AspNet.Mvc.Razor/Precompilation/CodeAnalysisSymbolBasedTypeInfo.cs
```diff
@@ -0,0 +1,251 @@
...
+            GetAssemblyQualifiedName(builder, symbol);
+
+            return builder.ToString();
+        }
+
+        private List<IPropertyInfo> PopulateProperties()
+        {
+            var properties = new List<IPropertyInfo>();
+            var overridenProperties = new HashSet<IPropertySymbol>();
+            var type = _type;
```
NTaylorMullen: Make the method static and pass in the `_type`

src/Microsoft.AspNet.Mvc.Razor/Precompilation/CodeAnalysisSymbolBasedTypeInfo.cs
```diff
@@ -0,0 +1,251 @@
...
+
+                return _properties;
+            }
+        }
+
+        /// <inheritdoc />
+        public IEnumerable<TAttribute> GetCustomAttributes<TAttribute>()
+            where TAttribute : Attribute
+        {
+            return CodeAnalysisAttributeUtilities.GetCustomAttributes<TAttribute>(_type, _symbolLookup);
```
NTaylorMullen: Ditto as in `CodeAnalysisSymbolBasedPropertyInfo`

src/Microsoft.AspNet.Mvc.Razor/Precompilation/CodeAnalysisSymbolBasedTypeInfo.cs
```diff
@@ -0,0 +1,251 @@
...
+        public bool ImplementsInterface(System.Reflection.TypeInfo interfaceType)
+        {
+            var interfaceSymbol = _symbolLookup.GetSymbol(interfaceType);
+            return _type.AllInterfaces.Any(implementedInterface => implementedInterface == interfaceSymbol);
+        }
+
+        /// <inheritdoc />
+        public string[] GetGenericDictionaryParameterNames()
+        {
+            var dictionarySymbol = _symbolLookup.GetSymbol(OpenGenericDictionaryTypeInfo);
```
NTaylorMullen: Seems wasteful to do this every time. Doesn't this not change in the lifetime of the application?

src/Microsoft.AspNet.Mvc.Razor/Precompilation/CodeAnalysisSymbolBasedPropertyInfo.cs
pranavkm: I debugged this and we don't request for the same attributes on types multiple times over. Seems pointless caching it as of now.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/CodeAnalysisSymbolBasedTypeInfo.cs
```diff
@@ -0,0 +1,251 @@
...
+            return builder.ToString();
+        }
+
+        private List<IPropertyInfo> PopulateProperties()
+        {
+            var properties = new List<IPropertyInfo>();
+            var overridenProperties = new HashSet<IPropertySymbol>();
+            var type = _type;
+
+            while (type != null)
```
NTaylorMullen: do while

src/Microsoft.AspNet.Mvc.Razor/Precompilation/CodeAnalysisAttributeUtilities.cs
pranavkm: Interesting - hadn't thought about that. Will give it a try.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/CodeAnalysisAttributeUtilities.cs
```diff
@@ -0,0 +1,256 @@
...
+            for (var index = 0; index < matchInfo.ArgumentValues.Length; index++)
+            {
+                parameters[index] =
+                    Expression.Convert(
+                        Expression.ArrayIndex(
+                            argsParameter,
+                            Expression.Constant(index)),
+                        matchInfo.ArgumentValues[index].GetType());
+            }
+
```
pranavkm: Will remove extra line.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/CodeAnalysisSymbolBasedTypeInfo.cs
```diff
@@ -0,0 +1,251 @@
...
+            {
+                foreach (var member in type.GetMembers())
+                {
+                    if (member.Kind == SymbolKind.Property)
```
NTaylorMullen: Note (doesn't matter one way or the other): Could do a where clause to reduce nesting 1 level.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/CodeAnalysisSymbolBasedTypeInfo.cs
pranavkm: These things change per compilation - so we can't reliably cache symbols beyond the lifetime of a single compilation. That said, `_symbolLookup` caches the actual lookup. I guess we could special case in that type so we don't have to go to a dictionary for the actual lookup.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/CodeAnalysisSymbolBasedTypeInfo.cs
```diff
@@ -0,0 +1,251 @@
...
+                    {
+                        var propertySymbol = (IPropertySymbol)member;
+                        if (!propertySymbol.IsIndexer && !overridenProperties.Contains(propertySymbol))
```
NTaylorMullen:
```C#
public class Foo
{
    public virtual string SomeProperty { get; set; }
}

public class Fooz : Foo
{
    public override string SomeProperty { get; set; }
}

public class Bar : Fooz
{
    public string SomeProperty { get; set; }
}
```

Bar should only return 1 property. :cry:

src/Microsoft.AspNet.Mvc.Razor/Precompilation/CodeAnalysisSymbolBasedTypeInfo.cs
```diff
@@ -0,0 +1,251 @@
...
+
+            return properties;
+        }
+
+        private static string GetFullName(ITypeSymbol typeSymbol)
+        {
+            var nameBuilder = new StringBuilder();
+            GetFullName(nameBuilder, typeSymbol);
+
+            return nameBuilder.Length == 0 ? null : nameBuilder.ToString();
```
NTaylorMullen: When would there be nothing in `nameBuilder`?

src/Microsoft.AspNet.Mvc.Razor/Precompilation/CodeAnalysisSymbolBasedTypeInfo.cs
```diff
@@ -0,0 +1,251 @@
...
+        }
+
+        private static void GetFullName(StringBuilder nameBuilder, ITypeSymbol typeSymbol)
+        {
+            if (typeSymbol.Kind == SymbolKind.TypeParameter)
```
NTaylorMullen: Is this for generic types?

src/Microsoft.AspNet.Mvc.Razor/Precompilation/CodeAnalysisSymbolBasedTypeInfo.cs
```diff
@@ -0,0 +1,251 @@
...
+                if (namedSymbol.IsGenericType &&
+                    namedSymbol.ConstructedFrom != namedSymbol)
+                {
+                    nameBuilder.Append('[');
+                    foreach (var typeArgument in namedSymbol.TypeArguments)
+                    {
+                        nameBuilder.Append('[');
+                        GetAssemblyQualifiedName(nameBuilder, typeArgument);
+                        nameBuilder
+                            .Append(']')
```
NTaylorMullen: Curious: Why not just `"[,"`?

src/Microsoft.AspNet.Mvc.Razor/Precompilation/CodeAnalysisSymbolBasedTypeInfo.cs
```diff
@@ -0,0 +1,251 @@
...
+                            .Append(',');
+                    }
+
+                    nameBuilder.Length--;
```
NTaylorMullen: Comment saying it nukes the comma

src/Microsoft.AspNet.Mvc.Razor/Precompilation/CodeAnalysisSymbolLookupCache.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// <param name="typeInfo">The <see cref="System.Reflection.TypeInfo"/> to lookup.</param>
+        /// <returns></returns>
+        public INamedTypeSymbol GetSymbol([NotNull] System.Reflection.TypeInfo typeInfo)
+        {
+           lock (_lookupLock)
+            {
+                INamedTypeSymbol typeSymbol;
+                if (!_symbolLookup.TryGetValue(typeInfo, out typeSymbol))
+                {
+                    typeSymbol = _compilation.GetTypeByMetadataName(typeInfo.FullName);
```
NTaylorMullen: It doesn't need to know the assembly?

src/Microsoft.AspNet.Mvc.Razor/Precompilation/CodeAnalysisSymbolBasedTypeInfo.cs
pranavkm: Yup. `IDictionary<,>` produces the name `IDictionary;2` - this is to accommodate scenarios along those lines.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/CodeAnalysisSymbolBasedTypeInfo.cs
pranavkm: It probably could be

src/Microsoft.AspNet.Mvc.Razor/Precompilation/PrecompilationTagHelperTypeResolver.cs
```diff
@@ -0,0 +1,109 @@
...
+namespace Microsoft.AspNet.Mvc.Razor.Precompilation
+{
+    /// <summary>
+    /// <see cref="TagHelperTypeResolver"/> used during Razor precompilation.
+    /// </summary>
+    public class PrecompilationTagHelperTypeResolver : TagHelperTypeResolver
+    {
+        private readonly object _assemblyLookupLock = new object();
+        private readonly Dictionary<string, IEnumerable<ITypeInfo>> _assemblyLookup
+            = new Dictionary<string, IEnumerable<ITypeInfo>>(StringComparer.OrdinalIgnoreCase);
```
NTaylorMullen: We sure all assembly lookups are case insensitive. I somewhat remember doing `@addTagHelper` required the exact casing of the assembly.

src/Microsoft.AspNet.Mvc.Razor/Resources.resx
```diff
@@ -200,4 +200,13 @@
 
 @{3} "{4}, {5}"</value>
   </data>
+  <data name="CodeAnalysisConstructorNotFound" xml:space="preserve">
+    <value>Unable to find a suitable constructor for type'{0}'.</value>
```
NTaylorMullen: `type '{0}'`

src/Microsoft.AspNet.Mvc.Razor/Precompilation/CodeAnalysisSymbolLookupCache.cs
pranavkm: Nope. One of the issues outlined in https://github.com/dotnet/roslyn/issues/4529.

We could potentially replace this with our own assembly visiting behavior so that we match assembly names.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/CodeAnalysisSymbolBasedPropertyInfo.cs
NTaylorMullen: :+1:

NTaylorMullen:
As discussed offline. Prior to getting this PR in we should include the test files/bits in Razor for the `TagHelperTypeResolver`, `TagHelperDescriptorFactory` and `TagHelperDescriptorResolver` to 100% verify that we're covering expected scenarios with pre-compilation. Those Razor tests are pretty exhaustive in one they cover; therefore they should give us a good confidence baseline.

NTaylorMullen:
:watch:

pranavkm:
I'll address PR comments here and continue this as part of https://github.com/aspnet/Razor/issues/500

PR #2962, opened 2015-08-12
Next in a series of PR's: supporting dynamics in JsonPatch.  

KevinDockx:
This PR contains one small fix on the Add operation (.GetType() must be called on the property value on the existing object, not on the value passed in) & the new Remove method, the last "big one" (plus additional unit tests for dynamic add/remove & unit test fixes in relation to the new error messages from the previous PR/Ryan's improvements).  

The new Remove method returns the type that's being removed - that'll be used in the next PR. Having it return the type allows for faster execution of operations that use the remove op (eg: Replace).  In the next PR, the remaining operations (+ cleanup) will be implemented.  They typically reuse parts of Remove/Add, or follow the same logic; so, once Add and Remove are checked, the rest should be a breeze ;-)

### DIFF
--- a/src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
+++ b/src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
@@ -451,10 +451,11 @@ namespace Microsoft.AspNet.JsonPatch.Adapters
         /// <summary>
         /// Remove is used by various operations (eg: remove, move, ...), yet through different operations;
         /// This method allows code reuse yet reporting the correct operation on error.  The return value
-        /// contains the type of the item that has been removed - this can be used by other methods, like 
-        /// replace, to ensure that we can pass in the correctly typed value to whatever method follows.
+        /// contains the type of the item that has been removed (and a bool possibly signifying an error)
+        /// This can be used by other methods, like replace, to ensure that we can pass in the correctly 
+        /// typed value to whatever method follows.
         /// </summary>
-        private Type Remove(string path, object objectToApplyTo, Operation operationToReport)
+        private RemovedPropertyTypeResult Remove(string path, object objectToApplyTo, Operation operationToReport)
         {
             // get path result
             var pathResult = GetActualPropertyPath(
@@ -464,7 +465,7 @@ namespace Microsoft.AspNet.JsonPatch.Adapters
 
             if (pathResult == null)
             {
-                return null;
+                return new RemovedPropertyTypeResult(null, true);
             }
 
             var removeFromList = pathResult.ExecuteAtEnd;
@@ -482,7 +483,7 @@ namespace Microsoft.AspNet.JsonPatch.Adapters
                     objectToApplyTo,
                     operationToReport,
                     Resources.FormatPropertyCannotBeRemoved(path)));
-                return null;
+                return new RemovedPropertyTypeResult(null, true);
             }
 
             if (treeAnalysisResult.UseDynamicLogic)
@@ -493,8 +494,19 @@ namespace Microsoft.AspNet.JsonPatch.Adapters
                 {
                     var propertyValue = treeAnalysisResult.Container
                         .GetValueForCaseInsensitiveKey(treeAnalysisResult.PropertyPathInParent);
-                    var typeOfPathProperty = treeAnalysisResult.Container
-                            .GetValueForCaseInsensitiveKey(treeAnalysisResult.PropertyPathInParent).GetType();
+
+                    // we cannot continue when the value is null, because to be able to
+                    // continue we need to be able to check if the array is a non-string array
+                    if (propertyValue == null)
+                    {
+                        LogError(new JsonPatchError(
+                           objectToApplyTo,
+                           operationToReport,
+                           Resources.FormatCannotDeterminePropertyType(path)));
+                        return new RemovedPropertyTypeResult(null, true);
+                    }
+
+                    var typeOfPathProperty = propertyValue.GetType();
 
                     if (!IsNonStringArray(typeOfPathProperty))
                     {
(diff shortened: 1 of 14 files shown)


src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -161,7 +161,7 @@ public void Add([NotNull] Operation operation, [NotNull] object objectToApplyTo)
                     {
                         // get the actual type
                         var propertyValue = container.GetValueForCaseInsensitiveKey(treeAnalysisResult.PropertyPathInParent);
-                        var typeOfPathProperty = value.GetType();
+                        var typeOfPathProperty = propertyValue.GetType();
```
rynowak: Oops :smile:

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -447,101 +447,219 @@ public void Remove([NotNull] Operation operation, [NotNull] object objectToApply
...
+                path,
+                objectToApplyTo,
+                operationToReport);
 
-            if (path.EndsWith("/-"))
+            if (pathResult == null)
             {
-                removeFromList = true;
-                actualPathToProperty = path.Substring(0, path.Length - 2);
+                return null;
```
rynowak: Should we report that the property wasn't found? Or is that considered not an error?

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -447,101 +447,219 @@ public void Remove([NotNull] Operation operation, [NotNull] object objectToApply
...
+                // if it's not an array, we can remove the property from
+                // the dictionary.  If it's an array, we need to check the position first.
+                if (removeFromList || positionAsInteger > -1)
                 {
-                    // now, get the generic type of the IList<> from Property type.
-                    var genericTypeOfArray = GetIListType(patchProperty.Property.PropertyType);
+                    var propertyValue = treeAnalysisResult.Container
+                        .GetValueForCaseInsensitiveKey(treeAnalysisResult.PropertyPathInParent);
+                    var typeOfPathProperty = treeAnalysisResult.Container
+                            .GetValueForCaseInsensitiveKey(treeAnalysisResult.PropertyPathInParent).GetType();
```
rynowak: instead of calling into the dictionary twice just call GetType() on the object. Also what if this returns null?

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
```diff
@@ -447,101 +447,219 @@ public void Remove([NotNull] Operation operation, [NotNull] object objectToApply
...
+                            Resources.FormatInvalidIndexForArrayProperty(operationToReport.op, path)));
+                        return null;
+                    }
+
+                    // now, get the generic type of the enumerable (we'll return this type)
+                    var genericTypeOfArray = GetIListType(typeOfPathProperty);
+
+                    // get the array
+                    var array = treeAnalysisResult.Container.GetValueForCaseInsensitiveKey(
+                        treeAnalysisResult.PropertyPathInParent) as IList;
```
rynowak: This should be a c-style cast. You don't expect it to fail.

test/Microsoft.AspNet.JsonPatch.Test/Dynamic/AddOperationTests.cs
```diff
@@ -0,0 +1,509 @@
...
+            }; 
+
+            // create patch
+            JsonPatchDocument patchDoc = new JsonPatchDocument();
+            patchDoc.Add("NewInt", 1);
+
+            var serialized = JsonConvert.SerializeObject(patchDoc);
+            var deserialized = JsonConvert.DeserializeObject<JsonPatchDocument>(serialized);
+
+            Assert.Throws<JsonPatchException>(() => { deserialized.ApplyTo(doc); });
```
rynowak: Would be nice to also test that it's the right error message.

rynowak:
:watch: looks pretty good overall. Happy to see so many test adds.

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
KevinDockx: Yep indeed, that is an error and should be reported, but that's what happens after the treeAnalysisResult is returned: if the property wasn't found, IsValidPathForRemove is false, which is then reported.  Code above relates to the pathResult, which returns what we expect the path to the property to be, judging from what's passed in to the operation.  It's only null in one case: when the passed-in path is invalid due to a negative index for an array property - and that's reported in the GetActualPropertyPath method.

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
KevinDockx: check, will do :)

test/Microsoft.AspNet.JsonPatch.Test/Dynamic/AddOperationTests.cs
KevinDockx: Will do!

src/Microsoft.AspNet.JsonPatch/Adapters/ObjectAdapter.cs
KevinDockx: Right, can't believe I didn't see that.  Will fix.  

If it's null, that is something I missed... The Remove itself should be executed, but maybe we could return a TypeResult sort of object (instead of just Type), that we can then use to throw an error if this is used inside Replace?  Like that, we could also avoid returning "null" as Type for those cases where we want to stop further execution (after an error has been logged).

rynowak:
:shipit:

rynowak:
054b46c1cce2bc90eea177a0fecfcc19448ad2b8

PR #2968, opened 2015-08-13
Fix #2859 - Correct UrlHelper for special tokens

### LINKED ISSUE #2859: Unexpected Url.Action output
Given routing in beta5:
```C#
app.UseMvc(routes =>
{
    routes.MapRoute(
        name: "DefaultDouble",
        template: "{custom}/{controller}/{action}/{id?}",
        defaults: new { controller = "Home", action = "Index" },
        constraints: new { custom = "Double" });

    routes.MapRoute(
        name: "DefaultSingle",
        template: "{controller}/{action}/{id?}",
        defaults: new { custom = "Single", controller = "Home", action = "Index" },
        constraints: new { custom = "Single" });
});
```
And calling this from NotHome controller view
```C#
@Url.Action("Index", new { custom = "Double" })
```
results in
```HTTP
/Double
```
while expected output is
```HTTP
/Double/NotHome
```
Repro
https://github.com/Muchiachio/ActionUrlSample

rynowak:
This change restores a link generation behavior from MVC5 and earlier
where 'action' and 'controller' values are special cased-when using
Url.Action(...).

The change is that in-effect 'action' and 'controller' are always included
in the route values given to the routing system. Passing a null value into
the Url.Action(...) method means that the ambient value for that token
should be used explicitly. This means that the 'action' and 'controller'
tokens become sticky, even when something to the lexical left in the URL
(like area) changes.

### DIFF
--- a/test/Microsoft.AspNet.Mvc.Core.Test/UrlHelperTest.cs
+++ b/test/Microsoft.AspNet.Mvc.Core.Test/UrlHelperTest.cs
@@ -3,11 +3,13 @@
 
 using System;
 using System.Collections.Generic;
+using System.Threading.Tasks;
 using Microsoft.AspNet.Builder;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Internal;
 using Microsoft.AspNet.Mvc.Internal;
 using Microsoft.AspNet.Routing;
+using Microsoft.Framework.DependencyInjection;
 using Microsoft.Framework.Logging;
 using Microsoft.Framework.Logging.Testing;
 using Microsoft.Framework.OptionsModel;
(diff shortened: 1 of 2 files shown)


rynowak:
There are still 3 function tests that I need to fix, all of them to do with UrlHelper in HTML.

test/Microsoft.AspNet.Mvc.Core.Test/UrlHelperTest.cs
```diff
@@ -807,6 +809,83 @@ public void LinkWithCustomHostAndProtocol_ReturnsExpectedResult()
             Assert.Equal("https://myhost/named/home/newaction/someid", url);
         }
 
+        // Regression test for aspnet/Mvc#2859
```
dougbu: Haven't seen many (any?) links to closed bugs in our code. Is it about as useful as a TODO comment?

dougbu:
:shipit:

test/Microsoft.AspNet.Mvc.Core.Test/UrlHelperTest.cs
rynowak: IMO it's much more useful than a TODO. The point is that if you want to understand **all** of the context about these (odd) tests, you can find it.

PR #2978, opened 2015-08-16
React to aspnet/Razor#441 change

ajaybhargavb:
aspnet/Razor#441
@dougbu


dougbu:
:shipit: when aspnet/Razor#490 is in.

Issue #2979, opened 2015-08-16
asp-append-version not working on Linux


Daniel15:
In my _Layout.cshtml:
```
<link rel="stylesheet" href="~/cache/main.min.css" asp-append-version="true"/>
```

On my development machine (Windows 7, Visual Studio 2015, Kestrel), this correctly appends the version:
```
<link rel="stylesheet" href="/cache/main.min.css?v=GQLFOo4-np2RJQauQG_3iKMNbSjDM5osk_tVLhNAnhE" />
```

However, in my production environment (Mono 4.0.2, Kestrel), no version is appended:
```
<link rel="stylesheet" href="/cache/main.min.css" />
```

Tested on Mono 4.0.2 and ASP.NET 5 Beta 6.

I'm deploying the site by doing `dnu publish` into a temporary directory, then `rsync`ing that temporary directory to the remote server.

Daniel15:
The problematic section of code is this part in `FileVersionProvider.cs`:
```
            Uri uri;
            if (Uri.TryCreate(resolvedPath, UriKind.Absolute, out uri))
            {
				// Don't append version if the path is absolute.
				return path;
            }
```

Mono's implementation of `Uri.TryCreate` is resolving `/foo.css` to `file:///foo.css`, because technically `/foo.css` **is** an absolute path to a file on Linux. Thus, it's just returning `path` unmodified. :sob: 

I think it should either explicitly be checking `IsFile` on the Uri. I'll submit a patch.

danroth27:
@NTaylorMullen @ajaybhargavb 

Yup, looks like something we would love to get a PR for. Thanks for finding the issue!

Daniel15:
I already submitted pull request #2980 to fix this issue :)

dougbu:
a2e53e8

PR #2980, opened 2015-08-16
Explicitly check if URI is file when checking for an absolute URI

Daniel15:
Parsing `/foo.css` as a URI on Linux results in `file:///foo.css`, because `/foo.css` is a legitimate absolute path to a file on the file system. This code needs to explicitly check if the resulting URI is that of a file. 

After fixing this, all the disabled tests relating to this feature actually work on Mono, and can be re-enabled. This means that https://github.com/aspnet/External/issues/21 (which I can't see because it's private, but it was referenced in the code) can be closed as it's not an actual issue.

#2979

### DIFF
--- a/test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
+++ b/test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
@@ -135,9 +135,7 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
             }
         }
 
-        [ConditionalTheory]
-        // Mono issue - https://github.com/aspnet/External/issues/21
-        [FrameworkSkipCondition(RuntimeFrameworks.Mono)]
+        [Fact]
         public void RendersImageTag_AddsFileVersion()
         {
             // Arrange
@@ -209,9 +207,7 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
             Assert.Equal("/images/test-image.png", srcAttribute.Value);
         }
 
-        [ConditionalTheory]
-        // Mono issue - https://github.com/aspnet/External/issues/21
-        [FrameworkSkipCondition(RuntimeFrameworks.Mono)]
+        [Fact]
         public void RendersImageTag_AddsFileVersion_WithRequestPathBase()
         {
             // Arrange
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
@@ -56,7 +56,7 @@ namespace Microsoft.AspNet.Mvc.TagHelpers.Internal
             }
 
             Uri uri;
-            if (Uri.TryCreate(resolvedPath, UriKind.Absolute, out uri))
+            if (Uri.TryCreate(resolvedPath, UriKind.Absolute, out uri) && !uri.IsFile)
             {
                 // Don't append version if the path is absolute.
                 return path;
--- a/test/Microsoft.AspNet.Mvc.FunctionalTests/HtmlGenerationTest.cs
+++ b/test/Microsoft.AspNet.Mvc.FunctionalTests/HtmlGenerationTest.cs
@@ -59,12 +59,6 @@ namespace Microsoft.AspNet.Mvc.FunctionalTests
         [InlineData("AttributesWithBooleanValues", null)]
         public async Task HtmlGenerationWebSite_GeneratesExpectedResults(string action, string antiforgeryPath)
         {
-            // This uses FileVersionProvider which uses Uri.TryCreate - https://github.com/aspnet/External/issues/21
-            if (TestPlatformHelper.IsMono && (action == "Link" || action == "Script" || action == "Image"))
-            {
-                return;
-            }
-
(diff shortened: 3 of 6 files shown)


Daniel15:
Actually I should be able to just use the existing test and remove the `FrameworkSkipCondition`, since that test should work on Mono now.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
```diff
@@ -82,9 +82,6 @@ public void PreservesOrderOfSourceAttributesWhenRun()
             }
         }
 
-        [ConditionalTheory]
```
sornaks: Since you removed the `[ConditionalTheory]` attribute, these tests are currently not running on any platforms. Replace
```c#
[ConditionalTheory]
// Mono issue - https://github.com/aspnet/External/issues/21
[FrameworkSkipCondition(RuntimeFrameworks.Mono)]
```
with
```c#
[Fact]
```
to make them run on all platforms.

Daniel15: Ohhhh, good catch! l didn't see any errors during the test run so I thought it was running :( I'll fix this tonight.

Daniel15:
@sornaks I added the missing `[Fact]` attributes. Unfortunately I updated to the latest bits recently and my local environment seems pretty broken right now (Mono crashes with `* Assertion: should not be reached at class.c:6488` when I try to run these tests) so I'm going to see how the AppVeyor/TravisCI runs go.

sornaks:
Hmmm, your tests seem to have run in travis and is passing. There are a couple of other places where tests have been disabled with this issue (<a href="https://github.com/aspnet/Mvc/blob/dev/test/Microsoft.AspNet.Mvc.FunctionalTests/HtmlGenerationTest.cs#L63">here</a> and <a href="https://github.com/aspnet/Mvc/blob/dev/test/Microsoft.AspNet.Mvc.FunctionalTests/HtmlGenerationTest.cs#L128">here</a>). Can you enable them as well?

Daniel15:
Those may need separate fixes, I haven't looked through much of the MVC code. I'd rather test and re-renable those other tests in a separate diff so as to not expand the scope of this one too much from its original intention (fixing the `asp-append-version` TagHelper). What do you think?

sornaks:
The places I mentioned are using `asp-append-version`. Ex: <a href="https://github.com/aspnet/Mvc/blob/dev/test/WebSites/HtmlGenerationWebSite/Views/HtmlGeneration_Home/Script.cshtml#L120">Here</a>. I'm guessing, just removing the `if` statements in those 2 places shd have them working.

Daniel15:
Looks like I'll need to rebase this on top of ca22f1649dca2ad0e4ba0643cf871184f526d644 - I'll do that and then try to enable the tests you mentioned @sornaks

dougbu:
@Daniel15 this PR is set. Let's :shipit: 

Please squash your commits (`git rebase -i head~3`) and confirm your fork is up-to-date. We'll get it in soon after.

Daniel15:
Thanks @dougbu! Squashed it now, hopefully it worked right (first time I've ever squashed something in Git). I attempted to enable the tests in `test/Microsoft.AspNet.Mvc.FunctionalTests/HtmlGenerationTest.cs` too. The build apprears to be broken at the moment though -  I'm getting the same error that's currently on Travis for the dev branch:

```
/Users/danlo/src/AspNetMvc/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs(8,27): DNX,Version=v4.5.1 error CS0234: The type or namespace name 'Primitives' does not exist in the namespace 'Microsoft.Framework' (are you missing an assembly reference?)
/Users/danlo/src/AspNetMvc/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs(65,16): DNX,Version=v4.5.1 error CS0246: The type or namespace name 'StringValues' could not be found (are you missing a using directive or an assembly reference?)
/Users/danlo/src/AspNetMvc/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs(41,36): DNX,Version=v4.5.1 error CS0246: The type or namespace name 'StringValues' could not be found (are you missing a using directive or an assembly reference?)
/Users/danlo/src/AspNetMvc/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs(51,36): DNX,Version=v4.5.1 error CS0246: The type or namespace name 'StringValues' could not be found (are you missing a using directive or an assembly reference?)
/Users/danlo/src/AspNetMvc/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs(35,74): DNX,Version=v4.5.1 error CS1503: Argument 1: cannot convert from 'string[]' to 'StringValues'
```

davidfowl:
@Daniel15 the build was fixed, could you rebase?

dougbu:
@davidfowl @Daniel15 is fine -- his fork was only a few commits behind and rebased cleanly for me. Will push shortly...

davidfowl:
@dougbu sweet

dougbu:
a2e53e8

Daniel15:
Thank you :smile:

Issue #2836, opened 2015-07-21
TryUpdateModelAsync trounces related models instead of updating them


rynowak:
`TryUpdateModelAsync(...)` will overwrite a whole related-model instead of just setting the properties that have changed.

After a request like: `http://localhost:6046/Home/Index?name=Bill&related.Tag1=newvalue`
In MVC 5 the value of `model.Related.Tag2` will still be `"ExistingValue"`
In MVC 6 the value of `model.Related.Tag2` will be `null`

```
        public async Task<IActionResult> Index()
        {
            var model = new Entity()
            {
                Name = "Joey",
                Related = new RelatedEntity()
                {
                    Tag1 = "ExistingValue",
                    Tag2 = "ExistingValue",
                }
            };

            await TryUpdateModelAsync(model);

            return View();
        }

        public class Entity
        {
            public string Name { get; set; }
            public RelatedEntity Related { get; set; }
        }

        public class RelatedEntity
        {
            public string Tag1 { get; set; }
            public string Tag2 { get; set; }
        }
```

dougbu:
48f09d0 in `release` and merged into `dev`

PR #2984, opened 2015-08-17
Change `MutableObjectModelBinder` to pass complex property values into binding system

dougbu:
- #2836

See individual commits for details. Brief overview:
- first commit lays the groundwork, removing `ComplexModelDTO`
- third commit contains the real fix
- second and fourth commits are somewhat optional; they just clean up test classes


src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
```diff
@@ -265,24 +264,50 @@ private static bool CanUpdateReadOnlyProperty(Type propertyType)
...
+                // be resized.
+                //
+                // ModelMetadata.PropertyGetter is not null safe; use it only if Model is non-null.
+                if (bindingContext.Model != null &&
+                    propertyMetadata.PropertyGetter != null &&
+                    propertyMetadata.IsComplexType &&
+                    !propertyMetadata.ModelType.IsArray)
+                {
+                    childContext.Model = propertyMetadata.PropertyGetter(bindingContext.Model);
+                }
```
dougbu: This block had no equivalent in the old `ComplexModelDTO` world. Fixes issue for realz.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/MutableObjectModelBinderTest.cs
```diff
@@ -73,8 +73,8 @@ public class MutableObjectModelBinderTest
             };
 
             var mutableBinder = new TestableMutableObjectModelBinder();
-            bindingContext.PropertyMetadata = mutableBinder.GetMetadataForProperties(
-                bindingContext.ModelBindingContext).ToArray();
+            bindingContext.PropertyMetadata =
+                mutableBinder.GetMetadataForProperties(bindingContext.ModelBindingContext).ToArray();
```
dougbu: Part of the nit-picking in the 2nd commit; removing YOLO wrapping.

test/Microsoft.AspNet.Mvc.IntegrationTests/TryUpdateModelIntegrationTest.cs
```diff
@@ -40,17 +45,19 @@ private class Address
...
-            Assert.Empty(modelState[key].Errors);
-            Assert.Equal(ModelValidationState.Valid, modelState[key].ValidationState);
+            var entry = Assert.Single(modelState);
+            Assert.Equal("Street", entry.Key);
+            var state = entry.Value;
+            Assert.NotNull(state.Value);
+            Assert.Equal("SomeStreet", state.Value.AttemptedValue);
+            Assert.Equal("SomeStreet", state.Value.RawValue);
+            Assert.Empty(state.Errors);
+            Assert.Equal(ModelValidationState.Valid, state.ValidationState);
```
dougbu: Changes to assertions in this class are from the fourth commit. No need for the `key` indirection.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/MutableObjectModelBinderTest.cs
```diff
@@ -417,10 +419,16 @@ public async Task
         {
             // Arrange
             var mockValueProvider = new Mock<IValueProvider>();
-            mockValueProvider.Setup(o => o.ContainsPrefixAsync(It.IsAny<string>()))
-                             .Returns(Task.FromResult(true));
+            mockValueProvider
+                .Setup(o => o.ContainsPrefixAsync(It.IsAny<string>()))
+                .Returns(Task.FromResult(true));
```
dougbu: Real changes in this class start here. All are the result of internal API changes in `MutableObjectModelBinder`. New scenarios are covered in `TryUpdateModelIntegrationTest`.

dougbu:
/cc @rynowak 

This change should be done in Beta 7. I'll squash and move the commit over to the `release` branch once signed off.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/ComplexModelDto.cs
```diff
@@ -1,32 +0,0 @@
...
-
-using System.Collections.Generic;
-using System.Collections.ObjectModel;
-using System.Linq;
-using Microsoft.Framework.Internal;
-
-namespace Microsoft.AspNet.Mvc.ModelBinding
-{
-    // Describes a complex model, but uses a collection rather than individual properties as the data store.
-    public class ComplexModelDto
```
rynowak: :dancer:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
```diff
@@ -40,18 +40,22 @@ public class MutableObjectModelBinder : IModelBinder
...
+            var results = await BindPropertiesAsync(bindingContext, mutableObjectBinderContext.PropertyMetadata);
 
             var validationNode = new ModelValidationNode(
                 bindingContext.ModelName,
                 bindingContext.ModelMetadata,
-                bindingContext.Model);
+                model);
+
+            // Post-processing e.g. property setters and hooking up validation.
+            bindingContext.Model = model;
```
rynowak: nit: should we do this right away after calling `GetModel`?

dougbu: If we did that `BindPropertiesAsync()` would behave differently for no good reason.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
```diff
@@ -40,18 +40,22 @@ public class MutableObjectModelBinder : IModelBinder
                 return null;
             }
 
-            EnsureModel(bindingContext);
-            var result = await CreateAndPopulateDto(bindingContext, mutableObjectBinderContext.PropertyMetadata);
+            // Create model first (if necessary) to avoid reporting errors about properties when activation fails.
+            var model = GetModel(bindingContext);
+
+            var results = await BindPropertiesAsync(bindingContext, mutableObjectBinderContext.PropertyMetadata);
```
rynowak: it would be good to do this property-by-property rather than "bind all properties" *then* "set all properties" - I don't think it needs to happen as part of this PR, but it should eventually for perf reasons.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
```diff
@@ -265,24 +264,50 @@ private static bool CanUpdateReadOnlyProperty(Type propertyType)
...
+                    propertyModelName,
+                    propertyMetadata);
+
+                // ModelBindingContext.Model property values may be non-null when invoked via TryUpdateModel(). Pass
+                // complex (including collection) values down so that binding system does not unnecessarily recreate
+                // instances or overwrite inner properties that are not bound. No need for this with simple values
+                // because they will be overwritten if binding succeeds. Arrays are never reused because they cannot
+                // be resized.
+                //
+                // ModelMetadata.PropertyGetter is not null safe; use it only if Model is non-null.
```
rynowak: This is supposed to be null-safe. If you've observed the opposite, please open a bug.

Note that `propertyMetadata.PropertyGetter` itself will be `null` if the property doesn't have a getter

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
```diff
@@ -265,24 +264,50 @@ private static bool CanUpdateReadOnlyProperty(Type propertyType)
...
+                if (bindingContext.Model != null &&
+                    propertyMetadata.PropertyGetter != null &&
+                    propertyMetadata.IsComplexType &&
```
rynowak: should we also be checking `IsCollectionType`? or does `IsComplexType` include that case?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
```diff
@@ -298,16 +323,18 @@ protected virtual object CreateModel([NotNull] ModelBindingContext bindingContex
...
 
         /// <summary>
-        /// Ensures <see cref="ModelBindingContext.Model"/> is not <c>null</c> in given
-        /// <paramref name="bindingContext"/>.
+        /// Get <see cref="ModelBindingContext.Model"/> if that property is not <c>null</c>. Otherwise activate a
+        /// new instance of <see cref="ModelBindingContext.ModelType"/>.
         /// </summary>
         /// <param name="bindingContext">The <see cref="ModelBindingContext"/>.</param>
-        protected virtual void EnsureModel([NotNull] ModelBindingContext bindingContext)
+        protected virtual object GetModel([NotNull] ModelBindingContext bindingContext)
```
rynowak: I don't really get why any of this needed to change. The code sets `bindingContext.Model` anyway.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
dougbu: I'm curious: What's the perf advantage? Is `CreatePropertyModelName()` expensive for example? Could see a larger advantage if we stopped binding after the first error.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
```diff
@@ -365,17 +392,18 @@ internal static PropertyValidationInfo GetPropertyValidationInfo(ModelBindingCon
...
+        internal ModelValidationNode ProcessResults(
             ModelBindingContext bindingContext,
-            ComplexModelDto dto,
+            IDictionary<ModelMetadata, ModelBindingResult> results,
             ModelValidationNode validationNode)
         {
             var metadataProvider = bindingContext.OperationBindingContext.MetadataProvider;
-            var modelExplorer = metadataProvider.GetModelExplorerForType(bindingContext.ModelType, bindingContext.Model);
+            var modelExplorer =
+                metadataProvider.GetModelExplorerForType(bindingContext.ModelType, bindingContext.Model);
```
rynowak: We should get rid of `ModelExplorer` use in this code if possible. If not, I'm going to end up removing it anyway.

src/Microsoft.AspNet.Mvc.Core/MvcCoreMvcOptionsSetup.cs
```diff
@@ -36,7 +36,6 @@ public static void ConfigureMvc(MvcOptions options)
             options.ModelBinders.Add(new FormCollectionModelBinder());
             options.ModelBinders.Add(new GenericModelBinder());
             options.ModelBinders.Add(new MutableObjectModelBinder());
-            options.ModelBinders.Add(new ComplexModelDtoModelBinder());
```
rynowak: :heart_eyes_cat:

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/MutableObjectModelBinderTest.cs
```diff
@@ -417,10 +419,16 @@ public async Task
...
+
+            // Mock binder fails to bind all properties.
+            var mockBinder = new Mock<IModelBinder>();
+            mockBinder
+                .Setup(o => o.BindModelAsync(It.IsAny<ModelBindingContext>()))
+                .Returns((ModelBindingContext mbc) => Task.FromResult<ModelBindingResult>(null));
```
rynowak: BTW you can do this simpler by just using `.Returns(Task.FromResult<ModelBindingResult>(null));`

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
dougbu: #2986

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
dougbu: `IsComplexType` just checks for a conversion from a `string`. So it includes `IsCollectionType`.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
dougbu: Timing: Want to activate the model instance before attempting to bind properties to maintain error handling. But `BindPropertiesAsync()` should not set `childContext.Model` for property bindings unless current `bindingContext,Model` was already non-`null`. But another way, the special case in `BindPropertiesAsync()` is just that -- a special case.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
dougbu: Not really part of this change. Will leave to the general cleanup you're planning.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/MutableObjectModelBinderTest.cs
dougbu: :ok:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
rynowak: Cool beans

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
rynowak: Yeah, I get it now, thanks!

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
rynowak: Sure

test/Microsoft.AspNet.Mvc.IntegrationTests/TryUpdateModelIntegrationTest.cs
```diff
@@ -28,7 +29,11 @@ private class Address
             });
 
             var modelState = new ModelStateDictionary();
-            var model = new Address { Street = "DefaultStreet" };
+            var model = new Address
+            {
+                Street = "DefaultStreet",
+                City = "Toronto",
```
rynowak: :maple_leaf:

rynowak:
:shipit:

dougbu:
48f09d0 in `release` and merged into `dev`

PR #2987, opened 2015-08-17
Enable `CopyHtmlAttribute` to maintain copied attribute order.

NTaylorMullen:
- Updated implementation to do a best guess at copying an attribute back into `TagHelperOutput.Attributes`.
- Updated existing functional and unit tests to account for maintained attribute order.
- Added a set of complex order based unit tests to validate `CopyHtmlAttribute` properly maintains order.

#2639

### DIFF
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
@@ -23,9 +23,17 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
         /// <param name="tagHelperOutput">The <see cref="TagHelperOutput"/> this method extends.</param>
         /// <param name="attributeName">The name of the bound attribute.</param>
         /// <param name="context">The <see cref="TagHelperContext"/>.</param>
-        /// <remarks>Only copies the attribute if <paramref name="tagHelperOutput"/>'s
+        /// <remarks>
+        /// <para>
+        /// Only copies the attribute if <paramref name="tagHelperOutput"/>'s
         /// <see cref="TagHelperOutput.Attributes"/> does not contain an attribute with the given
-        /// <paramref name="attributeName"/>.</remarks>
+        /// <paramref name="attributeName"/>.
+        /// </para>
+        /// <para>
+        /// Duplicate attributes same name in <paramref name="context"/>'s <see cref="TagHelperContext.AllAttributes"/>
+        /// or <paramref name="tagHelperOutput"/>'s <see cref="TagHelperOutput.Attributes"/> may result in copied
+        /// attribute order not being maintained.
+        /// </para></remarks>
         public static void CopyHtmlAttribute(
             [NotNull] this TagHelperOutput tagHelperOutput,
             [NotNull] string attributeName,
@@ -33,20 +41,29 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
         {
             if (!tagHelperOutput.Attributes.ContainsName(attributeName))
             {
-                IEnumerable<IReadOnlyTagHelperAttribute> entries;
+                var copiedAttribute = false;
 
-                // We look for the original attribute so we can restore the exact attribute name the user typed.
-                // Approach also ignores changes made to tagHelperOutput[attributeName].
-                if (!context.AllAttributes.TryGetAttributes(attributeName, out entries))
+                // We iterate context.AllAttributes backwards since we prioritize TagHelperOutput values occurring
+                // before the current context.AllAttribtes[i].
+                for (var i = context.AllAttributes.Count - 1; i >= 0; i--)
+                {
+                    // We look for the original attribute so we can restore the exact attribute name the user typed in
+                    // approximately the same position where the user wrote it in the Razor source.
+                    if (string.Equals(
+                        attributeName,
+                        context.AllAttributes[i].Name,
+                        StringComparison.OrdinalIgnoreCase))
+                    {
+                        CopyHtmlAttribute(i, tagHelperOutput, context);
+                        copiedAttribute = true;
+                    }
+                }
+
+                if (!copiedAttribute)
                 {
                     throw new ArgumentException(
                         Resources.FormatTagHelperOutput_AttributeDoesNotExist(attributeName, nameof(TagHelperContext)),
                         nameof(attributeName));
-                }
-
-                foreach (var entry in entries)
-                {
-                    tagHelperOutput.Attributes.Add(entry.Name, entry.Value);
                 }
             }
         }
@@ -99,6 +116,62 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
             {
                 tagHelperOutput.Attributes.Remove(attribute);
             }
+        }
+
+        private static void CopyHtmlAttribute(
+            int allAttributeIndex,
+            TagHelperOutput tagHelperOutput,
+            TagHelperContext context)
+        {
+            var existingAttribute = context.AllAttributes[allAttributeIndex];
+            var copiedAttribute = new TagHelperAttribute
+            {
+                Name = existingAttribute.Name,
+                Value = existingAttribute.Value,
+                Minimized = existingAttribute.Minimized
+            };
+
+            // Move backwards through context.AllAttributes from the provided index until we find a familiar attribute
+            // in tagHelperOutput where we can insert the copied value after the familiar one.
+            for (var i = allAttributeIndex - 1; i >= 0; i--)
+            {
+                var previousName = context.AllAttributes[i].Name;
+                var index = IndexOfFirstMatch(previousName, tagHelperOutput.Attributes);
+                if (index != -1)
+                {
+                    tagHelperOutput.Attributes.Insert(index + 1, copiedAttribute);
+                    return;
+                }
+            }
+
+            // Move forward through context.AllAttributes from the provided index until we find a familiar attribute in
+            // tagHelperOutput where we can insert the copied value.
+            for (var i = allAttributeIndex + 1; i < context.AllAttributes.Count; i++)
+            {
+                var nextName = context.AllAttributes[i].Name;
+                var index = IndexOfFirstMatch(nextName, tagHelperOutput.Attributes);
+                if (index != -1)
+                {
+                    tagHelperOutput.Attributes.Insert(index, copiedAttribute);
+                    return;
+                }
+            }
+
+            // Couldn't determine the attribute's location, add it to the end.
+            tagHelperOutput.Attributes.Add(copiedAttribute);
+        }
+
+        private static int IndexOfFirstMatch(string name, TagHelperAttributeList attributes)
+        {
+            for (var i = 0; i < attributes.Count; i++)
+            {
+                if (string.Equals(name, attributes[i].Name, StringComparison.OrdinalIgnoreCase))
+                {
+                    return i;
+                }
+            }
+
+            return -1;
         }
     }
 }
--- a/test/Microsoft.AspNet.Mvc.TagHelpers.Test/TagHelperOutputExtensionsTest.cs
+++ b/test/Microsoft.AspNet.Mvc.TagHelpers.Test/TagHelperOutputExtensionsTest.cs
@@ -8,13 +8,345 @@ using System.Threading.Tasks;
 using Microsoft.AspNet.Mvc.Rendering;
 using Microsoft.AspNet.Razor.Runtime.TagHelpers;
 using Microsoft.AspNet.Testing;
-using Microsoft.Framework.WebEncoders.Testing;
 using Xunit;
 
 namespace Microsoft.AspNet.Mvc.TagHelpers
 {
     public class TagHelperOutputExtensionsTest
     {
+        public static TheoryData CopyHtmlAttributeData_MaintainsOrder
+        {
+            get
+            {
+                // attributeNameToCopy, outputAttributes, allAttributes, expectedAttributes
+                return new TheoryData<
+                    string,
+                    TagHelperAttributeList,
+                    TagHelperAttributeList,
+                    IEnumerable<TagHelperAttribute>>
+                {
+                    {
+                        "first",
+                        new TagHelperAttributeList
+                        {
+                            { "second", "B" },
+                        },
+                        new TagHelperAttributeList
+                        {
+                            { "first", "A" },
+                            { "second", "B" }
+                        },
+                        new[]
+                        {
+                            new TagHelperAttribute("first", "A"),
+                            new TagHelperAttribute("second", "B"),
+                        }
+                    },
+                    {
+                        "second",
+                        new TagHelperAttributeList
+                        {
+                            { "first", "A" },
+                        },
+                        new TagHelperAttributeList
+                        {
+                            { "second", "B" },
+                            { "second", "Duplicate B" },
+                            { "first", "A" },
+                        },
+                        new[]
+                        {
+                            new TagHelperAttribute("second", "B"),
+                            new TagHelperAttribute("second", "Duplicate B"),
+                            new TagHelperAttribute("first", "A"),
+                        }
+                    },
+                    {
+                        "second",
+                        new TagHelperAttributeList
+                        {
+                            { "first", "A" },
+                        },
+                        new TagHelperAttributeList
+                        {
+                            { "second", "B" },
+                            { "first", "A" },
+                            { "second", "Duplicate B" }
+                        },
+                        new[]
+                        {
+                            new TagHelperAttribute("second", "B"),
+                            new TagHelperAttribute("first", "A"),
+                            new TagHelperAttribute("second", "Duplicate B"),
+                        }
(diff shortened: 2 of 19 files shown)


NTaylorMullen:
Making a change to the algorithm. Instead of going backwards then forwards through the `context.AllAttributes` to find a familiar attribute i'm going to look both ways simultaneously to ensure we find the closest.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/TagHelperOutputExtensionsTest.cs
```diff
@@ -8,13 +8,258 @@
...
+                            new TagHelperAttribute("secondDynamic", "another value"),
+                        }
+                    },
+                    {
+                        "first",
+                        new TagHelperAttributeList
+                        {
+                            { "third", "C" },
+                            { "dynamic", "value" },
+                            { "secondDynamic", "another value"}
```
dougbu: How would `output.Attributes` get into this mixed-up state?

test/Microsoft.AspNet.Mvc.TagHelpers.Test/TagHelperOutputExtensionsTest.cs
```diff
@@ -8,13 +8,258 @@
...
```
dougbu: Again, is this mix-up realistic? Does behaviour once in this state matter much?

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
```diff
@@ -33,21 +33,27 @@ public static class TagHelperOutputExtensions
...
+                for (var i = context.AllAttributes.Count - 1; i >= 0; i--)
+                {
+                    // We look for the original attribute so we can restore the exact attribute name the user typed.
+                    // Approach also ignores changes made to tagHelperOutput[attributeName].
+                    if (string.Equals(
+                        attributeName,
+                        context.AllAttributes[i].Name,
+                        StringComparison.OrdinalIgnoreCase))
+                    {
+                        CopyHtmlAttributeMaintainOrder(i, tagHelperOutput, context);
```
dougbu: Suggest calling the called method `CopyHtmlAttribute()`. Since both the current method and the one called here maintain order, that's not a distinguishing feature.

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
```diff
@@ -33,21 +33,27 @@ public static class TagHelperOutputExtensions
         {
             if (!tagHelperOutput.Attributes.ContainsName(attributeName))
             {
-                IEnumerable<IReadOnlyTagHelperAttribute> entries;
+                var copiedAttribute = false;
+                for (var i = context.AllAttributes.Count - 1; i >= 0; i--)
```
dougbu: Add a comment about why this needs to go through `AllAttributes` backwards.

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
```diff
@@ -33,21 +33,27 @@ public static class TagHelperOutputExtensions
...
+                {
+                    // We look for the original attribute so we can restore the exact attribute name the user typed.
+                    // Approach also ignores changes made to tagHelperOutput[attributeName].
```
dougbu: Update comment to mention restoring the attribute (approximately) where the user wrote it in the Razor source.

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
```diff
@@ -100,5 +106,70 @@ public static class TagHelperOutputExtensions
                 tagHelperOutput.Attributes.Remove(attribute);
             }
         }
+
+        private static void CopyHtmlAttributeMaintainOrder(
+            int allAttributeIndex,
+            TagHelperOutput tagHelperOutput,
+            TagHelperContext context)
```
dougbu: Probably worth sticking with the `TagHelperOutput`, {specific attribute info}, `TagHelperContext` order of the (other) `CopyHtmlAttribute()` method.

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
```diff
@@ -100,5 +106,70 @@ public static class TagHelperOutputExtensions
...
+            {
+                Name = existingAttribute.Name,
+                Value = existingAttribute.Value,
+                Minimized = existingAttribute.Minimized
+            };
+
+            for (var i = allAttributeIndex - 1; i >= 0; i--)
+            {
+                // Move backwards through context.AllAttributes from the provided index until we find a familiar
+                // attribute in tagHelperOutput where we can insert the copied value in front of.
```
dougbu: Move comment above the loop since it doesn't just describe the loop body.

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
```diff
@@ -100,5 +106,70 @@ public static class TagHelperOutputExtensions
...
+                    attributes: tagHelperOutput.Attributes))
+                {
+                    return;
+                }
+            }
+
+            for (var i = allAttributeIndex + 1; i < context.AllAttributes.Count; i++)
+            {
+                // Move forward through context.AllAttributes from the provided index until we find a familiar
+                // attribute in tagHelperOutput where we can insert the copied value.
```
dougbu: Also move this comment above the loop.

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
```diff
@@ -100,5 +106,70 @@ public static class TagHelperOutputExtensions
...
+                Name = existingAttribute.Name,
+                Value = existingAttribute.Value,
+                Minimized = existingAttribute.Minimized
+            };
+
+            for (var i = allAttributeIndex - 1; i >= 0; i--)
+            {
+                // Move backwards through context.AllAttributes from the provided index until we find a familiar
+                // attribute in tagHelperOutput where we can insert the copied value in front of.
+                var previousAllAttributeName = context.AllAttributes[i].Name;
```
dougbu: Bit of a mouthful; suggest `previousName`.

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
```diff
@@ -100,5 +106,70 @@ public static class TagHelperOutputExtensions
...
+                {
+                    return;
+                }
+            }
+
+            for (var i = allAttributeIndex + 1; i < context.AllAttributes.Count; i++)
+            {
+                // Move forward through context.AllAttributes from the provided index until we find a familiar
+                // attribute in tagHelperOutput where we can insert the copied value.
+                var nextAllAttributeName = context.AllAttributes[i].Name;
```
dougbu: Suggest `nextName`.

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
```diff
@@ -100,5 +106,70 @@ public static class TagHelperOutputExtensions
...
+            tagHelperOutput.Attributes.Add(copiedAttribute);
+        }
+
+        private static bool TryFindNameAndInsertAttribute(
+            string name,
+            int insertOffset,
+            TagHelperAttribute attributeToInsert,
+            TagHelperAttributeList attributes)
+        {
+            for (var j = 0; j < attributes.Count; j++)
```
dougbu: Use `i` or `index`. Neither is ambiguous here.

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
```diff
@@ -100,5 +106,70 @@ public static class TagHelperOutputExtensions
...
+            {
+                if (string.Equals(name, attributes[j].Name, StringComparison.OrdinalIgnoreCase))
+                {
+                    attributes.Insert(j + insertOffset, attributeToInsert);
```
dougbu: This results in an odd mix of responsibilities. Instead do the insertion in `CopyHtmlAttribute()` and change this method to
``` c#
private static int FindFirstMatch(TagHelperAttributeList attributeList, string name)
{
    for (var index = 0; index < attributes.Count; index++)
    {
        if (string.Equals(name, attributeList[index].Name, StringComparison.OrdinalIgnoreCase))
        {
            return index;
        }
    }

    return -1;
}
```

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
NTaylorMullen: Nah, other `CopyHtmlAttribute` method is an extension. Otherwise it wouldn't have same order.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/TagHelperOutputExtensionsTest.cs
NTaylorMullen:
```C#
output.CopyHtmlAttribute("third", context);
...
output.Attributes.Add("dynamic", "value");
output.Attributes.Add("secondDynamic", "another value");
```

test/Microsoft.AspNet.Mvc.TagHelpers.Test/TagHelperOutputExtensionsTest.cs
NTaylorMullen: We modify attributes somewhat similarly. Same setup could occur: https://github.com/aspnet/Mvc/pull/2987/files#r37362253

NTaylorMullen:
Updated.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/TagHelperOutputExtensionsTest.cs
dougbu: :ok:

test/Microsoft.AspNet.Mvc.TagHelpers.Test/TagHelperOutputExtensionsTest.cs
dougbu: :ok:

NTaylorMullen:
This good?

test/Microsoft.AspNet.Mvc.TagHelpers.Test/TagHelperOutputExtensionsTest.cs
ajaybhargavb: Suggest adding another test case with `outputAttributes` containing some duplicates.

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
```diff
@@ -33,20 +33,30 @@ public static class TagHelperOutputExtensions
...
+                // We iterate context.AllAttributes backwards since we prioritize TagHelperOutput values occurring
+                // before the current context.AllAttribtes[i].
+                for (var i = context.AllAttributes.Count - 1; i >= 0; i--)
                 {
-                    throw new ArgumentException(
-                        Resources.FormatTagHelperOutput_AttributeDoesNotExist(attributeName, nameof(TagHelperContext)),
-                        nameof(attributeName));
+                    // We look for the original attribute so we can restore the exact attribute name the user typed in
+                    // approximately the same position where the user wrote it in the Razor source.
+                    // Approach also ignores changes made to tagHelperOutput[attributeName].
```
dougbu: What `changes made to tagHelperOutput[attributeName]`? Have already confirmed this `attributeName` does not appear in the output collection.

ajaybhargavb:
:shipit: from me

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
```diff
@@ -100,5 +110,61 @@ public static class TagHelperOutputExtensions
...
+                Name = existingAttribute.Name,
+                Value = existingAttribute.Value,
+                Minimized = existingAttribute.Minimized
+            };
+
+            // Move backwards through context.AllAttributes from the provided index until we find a familiar attribute
+            // in tagHelperOutput where we can insert the copied value in front of.
+            for (var i = allAttributeIndex - 1; i >= 0; i--)
+            {
+                var previouName = context.AllAttributes[i].Name;
```
dougbu: tyo: `previousName`

NTaylorMullen: My keyboard doesn't like `s`s after u's for whatever reason...

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
```diff
@@ -100,5 +110,61 @@ public static class TagHelperOutputExtensions
...
+                var index = IndexOfFirstMatch(previouName, tagHelperOutput.Attributes);
```
dougbu: Don't we want the _last_ attribute with same `Name` here? For example,
``` html
  <element existing="value 1" existing="value 2" toBeCopied="value 3"/>
```
will currently be restored as
``` html
  <element existing="value 1" toBeCopied="value 3" existing="value 2"/>
```

Do tests cover this case i.e. `output.Attributes` contains duplicate attributes one of which should be placed just before the copied attribute?

dougbu: Suspect this code needs to figure out how many instances of `previousName` are in `context.AllAttributes[0..allAttributeIndex -1]` and look forward through `output.Attributes` to find the same count (or the last match).

test/Microsoft.AspNet.Mvc.TagHelpers.Test/TagHelperOutputExtensionsTest.cs
dougbu: +1

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
```diff
@@ -100,5 +110,61 @@ public static class TagHelperOutputExtensions
...
+                    return;
+                }
+            }
+
+            // Move forward through context.AllAttributes from the provided index until we find a familiar attribute in
+            // tagHelperOutput where we can insert the copied value.
+            for (var i = allAttributeIndex + 1; i < context.AllAttributes.Count; i++)
+            {
+                var nextName = context.AllAttributes[i].Name;
+                var index = IndexOfFirstMatch(nextName, tagHelperOutput.Attributes);
```
dougbu: Same problem with duplicate matches for `nextName` here.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/TagHelperOutputExtensionsTest.cs
dougbu: Plus cases where `attributeNameToCopy` matches adjacent attributes in `allAttributes`.

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
NTaylorMullen: Old comment. Will delete.

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
NTaylorMullen: Talked in person.

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperOutputExtensions.cs
NTaylorMullen: Ditto.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/TagHelperOutputExtensionsTest.cs
NTaylorMullen: Done.

NTaylorMullen:
Updated.

NTaylorMullen:
Updated remarks.

dougbu:
:shipit:

Issue #2539, opened 2015-05-11
Changes to ModelMetadata provided by a ModelMetadataDetailsProvider are sometimes lost if the model property is a Nullable<T>


emisaacson:
In one of my projects I have a registered `IDisplayMetadataProvider` that examines the attributes on a model's properties and sets values in `ModelMetadata.AdditionalValues` as necessary. I find that if:

* the model property in question is a `Nullable<T>` AND
* the default value of the property is a value other than null

then the customizations to ModelMetadata are lost by the time the view is rendered.

A minimal project to reproduce the issue is here: https://github.com/emisaacson/MetadataProviderExample

I find with this change I get the expected result (`AdditionalValues` is populated when rendering the view): https://github.com/aspnet/Mvc/compare/dev...emisaacson:vd-type-checking

dougbu:
f10a071

PR #2988, opened 2015-08-18
Preserve `ViewDataDictionary.ModelType` for `Nullable<T>` properties when `Model` is non-`null`

dougbu:
- #2539
- reuse `ModelMetadata` and occasionally `ModelExplorer` when `ModelType` is `Nullable<T>`

### DIFF
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewDataDictionary.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewDataDictionary.cs
@@ -186,7 +186,9 @@ namespace Microsoft.AspNet.Mvc
         {
             // This is the core constructor called when Model is known.
             var modelType = GetModelType(model);
-            if (modelType == source.ModelMetadata.ModelType && model == source.ModelExplorer.Model)
+            var metadataModelType =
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Mvc.ViewFeatures/ViewDataDictionary.cs
```diff
@@ -186,7 +186,9 @@ protected ViewDataDictionary([NotNull] ViewDataDictionary source, object model,
         {
             // This is the core constructor called when Model is known.
             var modelType = GetModelType(model);
-            if (modelType == source.ModelMetadata.ModelType && model == source.ModelExplorer.Model)
+            var metadataModelType =
+                Nullable.GetUnderlyingType(source.ModelMetadata.ModelType) ?? source.ModelMetadata.ModelType;
```
rynowak: Should we hoist this to a method on `ModelMetadata`?

rynowak:
:shipit:

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewDataDictionary.cs
dougbu: I like the idea but suggest doing this (and a bit more) as a follow-up. Filed #2992

dougbu:
f10a071

PR #3000, opened 2015-08-19
Fix #2986 - Make ModelMetadata getters null-safe

### LINKED ISSUE #2986: `DefaultModelMetadata.PropertyGetter` is not `null`-safe
No current MVC code uses `DefaultModelMetadata.PropertyGetter`. If it were used, would find `null` values e.g. `ModelMetadata.PropertyGetter(null)` lead to NREs. Root cause may be the fact `PropertyHelper.MakeNullSafeFastPropertyGetter()` is not referenced.

Experienced this issue while developing PR #2984.

rynowak:
/cc @dougbu

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadataProvider.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadataProvider.cs
@@ -122,7 +122,8 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
                 var propertyEntry = new DefaultMetadataDetails(propertyKey, attributes);
                 if (propertyHelper.Property.CanRead && propertyHelper.Property.GetMethod?.IsPublic == true)
                 {
-                    propertyEntry.PropertyGetter = propertyHelper.ValueGetter;
+                    var getter = PropertyHelper.MakeNullSafeFastPropertyGetter(propertyHelper.Property);
+                    propertyEntry.PropertyGetter = getter;
                 }
 
                 if (propertyHelper.Property.CanWrite &&
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
@@ -287,10 +287,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
                 // instances or overwrite inner properties that are not bound. No need for this with simple values
                 // because they will be overwritten if binding succeeds. Arrays are never reused because they cannot
                 // be resized.
-                //
-                // ModelMetadata.PropertyGetter is not null safe; use it only if Model is non-null.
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadataProvider.cs
```diff
@@ -122,7 +122,8 @@ protected virtual DefaultMetadataDetails[] CreatePropertyDetails([NotNull] Model
                 var propertyEntry = new DefaultMetadataDetails(propertyKey, attributes);
                 if (propertyHelper.Property.CanRead && propertyHelper.Property.GetMethod?.IsPublic == true)
                 {
-                    propertyEntry.PropertyGetter = propertyHelper.ValueGetter;
+                    var getter = PropertyHelper.MakeNullSafeFastPropertyGetter(propertyHelper.Property);
+                    propertyEntry.PropertyGetter = getter;
```
dougbu: Why isn't `propertyHelper.ValueGetter` `null`-safe? Do we need the `Make*()` methods or the `null`-unsafe bits at all?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
```diff
@@ -287,10 +287,7 @@ private static bool CanUpdateReadOnlyProperty(Type propertyType)
                 // instances or overwrite inner properties that are not bound. No need for this with simple values
                 // because they will be overwritten if binding succeeds. Arrays are never reused because they cannot
                 // be resized.
-                //
-                // ModelMetadata.PropertyGetter is not null safe; use it only if Model is non-null.
-                if (bindingContext.Model != null &&
-                    propertyMetadata.PropertyGetter != null &&
+                if (propertyMetadata.PropertyGetter != null &&
```
dougbu: :dancers:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadataProvider.cs
dougbu: Should test `null` safety in `DefaultModelMetadataProviderTest`. At the moment nothing except product code (`MutableObjectModelBinder` i.e. no test code) does anything beyond check that `metadata.PropertyGetter` is itself non-`null`.

dougbu:
Design looks good. Ready for the real PR :smiling_imp:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadataProvider.cs
rynowak: Will add a test that actually uses the accessors.

As for propertyhelper, I don't see a reason to change it. In the validation/binding code we have a good reason to care about null-safety. In general it's something I'd avoid.

rynowak:
Added a test :+1:

dougbu:
:shipit:

PR #3004, opened 2015-08-19
Refactor ICompilerCache to be instantiated via ICompilerCacheProvider

### LINKED ISSUE #2933: Consider refactoring ICompilerCache to be consumed via a ICompilerCacheProvider
With the current approach, consuming the CompilerCache as a service would cause it to eagerly populate the cache - which performs expensive assembly loads. We could clean up the design so that it's closer to how action descriptors work.

Discussion here: https://github.com/aspnet/Mvc/pull/2917#discussion_r36452859

pranavkm:
Fixes #2933


pranavkm:
cc @rynowak

src/Microsoft.AspNet.Mvc.Razor/VirtualPathRazorPageFactory.cs
```diff
@@ -15,14 +15,20 @@ namespace Microsoft.AspNet.Mvc.Razor
...
+        /// <summary>
+        /// Initializes a new instance of <see cref="VirtualPathRazorPageFactory"/>.
+        /// </summary>
+        /// <param name="serviceProvider">The request specific <see cref="IServiceProvider"/>.</param>
+        /// <param name="compilerCacheProvider">The <see cref="ICompilerCacheProvider"/>.</param>
+        public VirtualPathRazorPageFactory(
+            IServiceProvider serviceProvider,
+            ICompilerCacheProvider compilerCacheProvider)
         {
             _serviceProvider = serviceProvider;
```
pranavkm: Slightly related but I don't think https://github.com/aspnet/Mvc/pull/3004/files#diff-88150cf385250544dc6321e6ef7c28fdR42 is true. If not, I'll clean it up as a follow up to this.

src/Microsoft.AspNet.Mvc.Razor/Compilation/DefaultCompilerCacheProvider.cs
```diff
@@ -0,0 +1,26 @@
...
+        /// </summary>
+        /// <param name="optionsAccessor">An accessor to the <see cref="RazorViewEngineOptions"/>.</param>
+        public DefaultCompilerCacheProvider(IOptions<RazorViewEngineOptions> mvcViewOptions)
+        {
+            var fileProvider = mvcViewOptions.Options.FileProvider;
+            Cache = new CompilerCache(fileProvider);
+        }
+
+        /// <inheritdoc />
+        public ICompilerCache Cache { get; }
```
rynowak: Wasn't the spirit of this change to make defer initialization until `Cache` is accessed?

src/Microsoft.AspNet.Mvc.Razor/Compilation/PrecompiledViewsCompilerCacheProvider.cs
```diff
@@ -0,0 +1,104 @@
...
+            IEnumerable<Assembly> assemblies)
+        {
+            _loadContextAccessor = loadContextAccessor;
+            _fileProvider = mvcViewOptions.Options.FileProvider;
+            _createCache = CreateCache;
+            _assemblies = assemblies.ToArray();
+        }
+
+        /// <inheritdoc />
+        public ICompilerCache Cache =>
```
rynowak: This is really annoying for multi-line code.

rynowak: Mad props for LazyInitializer, and also mad PranavPoints.

src/Microsoft.AspNet.Mvc.Razor/Compilation/DefaultCompilerCacheProvider.cs
pranavkm: I guess we could. Initializing the cache for non-precompiled is pretty inexpensive (all it does is set up an empty `MemoryCache` instance).

src/Microsoft.AspNet.Mvc.Razor/DependencyInjection/MvcRazorMvcBuilderExtensions.cs
```diff
@@ -53,13 +54,17 @@ public static class MvcRazorMvcBuilderExtensions
...
             [NotNull] this IMvcBuilder builder,
             [NotNull] params Assembly[] assemblies)
         {
-            var razorFileInfos = RazorFileInfoCollections.GetFileInfoCollections(assemblies);
-            builder.Services.TryAddEnumerable(ServiceDescriptor.Instance(razorFileInfos));
+            builder.Services.Replace(
+                ServiceDescriptor.Singleton<ICompilerCacheProvider>(serviceProvider =>
+                    ActivatorUtilities.CreateInstance<PrecompiledViewsCompilerCacheProvider>(
+                        serviceProvider,
+                        assemblies.AsEnumerable())));
```
rynowak: What as `AsEnumerable()`?

src/Microsoft.AspNet.Mvc.Razor/VirtualPathRazorPageFactory.cs
rynowak: That would be cool

src/Microsoft.AspNet.Mvc.Razor/DependencyInjection/MvcRazorMvcBuilderExtensions.cs
pranavkm: `ActivatorUtilities` doesn't like it when we expect `IEnumerable<>` but pass in `Assembly[]`.

src/Microsoft.AspNet.Mvc.Razor/VirtualPathRazorPageFactory.cs
```diff
@@ -50,7 +56,8 @@ public IRazorPage CreateInstance([NotNull] string relativePath)
                 relativePath = relativePath.Substring(1);
             }
 
-            var result = _compilerCache.GetOrAdd(
+            var compilerCache = _compilerCacheProvider.Cache;
```
rynowak: You're going to be taking a lock here right? Might want to cache the result.

src/Microsoft.AspNet.Mvc.Razor/Compilation/DefaultCompilerCacheProvider.cs
rynowak: Fair enough, yeah I think this is fine then.

rynowak:
:shipit:

src/Microsoft.AspNet.Mvc.Razor/DependencyInjection/MvcRazorMvcBuilderExtensions.cs
pranavkm: VS ends up greying out the (`IEnumerable<Assembly>`) cast to indicate it is redundant. Rather than have to document why it's there, I'll leave the `AsEnumerable` which essentially does the same thing internally:
http://referencesource.microsoft.com/#System.Core/System/Linq/Enumerable.cs,825

src/Microsoft.AspNet.Mvc.Razor/VirtualPathRazorPageFactory.cs
pranavkm: :+1:

Issue #2902, opened 2015-08-03
ImageTagHelper doesn't work well with other TagHelpers.


NTaylorMullen:
Today, `ImageTagHelper` relies too heavily on it's bound `Src` attribute which in turn makes interactions with other `TagHelper`s insufficient. Aka, if another `TagHelper` modifies `output.Attributes["src"]` the `ImageTagHelper` will trump their value instead of use it.

danroth27:
@DamianEdwards

DamianEdwards:
Hmmmm, this sounds familiar. Oh yeah, I've always said binding those was stupid :P

dougbu:
How would we handle encoding issues after making this suggested change?

NTaylorMullen:
Would there be a problem consuming the value already encoded? If not we could just ensure that values are not re-encoded.

NTaylorMullen:
Talked offline, with @dougbu, this isn't possible today.

NTaylorMullen:
Actually, we can complete this partially. Updated issue to reflect change indirection.

DamianEdwards:
Is meant to be closed still?

NTaylorMullen:
c3e2e6fa0a8888374e3d05127185cea314622ffb

PR #3006, opened 2015-08-20
Change Script, Link and Image `TagHelper`s to work better with other `TagHelper`s.

NTaylorMullen:
- `ScriptTagHelper`, `LinkTagHelper` and `ImageTagHelper` now default to using `output.Attributes["href|src"]` if it's present when they run. This enables other `TagHelper`s to run prior and add those attributes.
- Added unit tests to validate this behavior.
- Updated `ImageTagHelper` functional test resources. Now that we're always defaulting to `output.Attributes["src"]` for `ImageTagHelper.Src` we're properly copying attributes back into the `output.Attributes` collection in the correct order (isntead of appending to the end).

#2902

### DIFF
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
@@ -75,22 +75,19 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
         /// <inheritdoc />
         public override void Process(TagHelperContext context, TagHelperOutput output)
         {
+            output.CopyHtmlAttribute(SrcAttributeName, context);
+            ProcessUrlAttribute(SrcAttributeName, output);
+
             if (AppendVersion)
             {
                 EnsureFileVersionProvider();
 
-                string resolvedUrl;
-                if (TryResolveUrl(Src, encodeWebRoot: false, resolvedUrl: out resolvedUrl))
-                {
-                    Src = resolvedUrl;
-                }
+                // Retrieve the TagHelperOutput variation of the "src" attribute in case other TagHelpers in the
+                // pipeline have touched the value. If the value is already encoded this ImageTagHelper may
+                // not function properly.
+                Src = output.Attributes[SrcAttributeName].Value as string;
+
                 output.Attributes[SrcAttributeName] = _fileVersionProvider.AddFileVersionToPath(Src);
-            }
-            else
-            {
-                // Pass through attribute that is also a well-known HTML attribute.
-                output.CopyHtmlAttribute(SrcAttributeName, context);
-                ProcessUrlAttribute(SrcAttributeName, output);
             }
         }
 
--- a/test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.Image.html
+++ b/test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.Image.html
@@ -12,24 +12,24 @@
     <img src="/images/red.png" alt="Red block" title="&lt;the title>">
 
     <!-- Plain image tag with file version -->
-    <img alt="Red versioned" title="Red versioned" src="/images/red.png?v=W2F5D366_nQ2fQqUk3URdgWy2ZekXjHzHJaY5yaiOOk" />
+    <img src="/images/red.png?v=W2F5D366_nQ2fQqUk3URdgWy2ZekXjHzHJaY5yaiOOk" alt="Red versioned" title="Red versioned" />
 
     <!-- Plain image tag with file version set to false -->
     <img src="/images/red.png" alt="Red explicitly not versioned" title="Red versioned">
 
     <!-- Plain image tag with absolute path and file version -->
-    <img alt="Absolute path versioned" src="http://contoso.com/hello/world">
+    <img src="http://contoso.com/hello/world" alt="Absolute path versioned">
 
     <!-- Plain image tag with file version and path to file that does not exist -->
-    <img alt="Path to non existing file" src="/images/fake.png" />
+    <img src="/images/fake.png" alt="Path to non existing file" />
 
     <!-- Plain image tag with file version and path containing query string -->
-    <img alt="Path with query string" src="/images/red.png?abc=def&amp;v=W2F5D366_nQ2fQqUk3URdgWy2ZekXjHzHJaY5yaiOOk">
+    <img src="/images/red.png?abc=def&amp;v=W2F5D366_nQ2fQqUk3URdgWy2ZekXjHzHJaY5yaiOOk" alt="Path with query string">
 
     <!-- Plain image tag with file version and path containing fragment -->
-    <img alt="Path with query string" src="/images/red.png?v=W2F5D366_nQ2fQqUk3URdgWy2ZekXjHzHJaY5yaiOOk#abc" />
+    <img src="/images/red.png?v=W2F5D366_nQ2fQqUk3URdgWy2ZekXjHzHJaY5yaiOOk#abc" alt="Path with query string" />
 
     <!-- Plain image tag with file version and path linking to some action -->
-    <img alt="Path linking to some action" src="/controller/action">
+    <img src="/controller/action" alt="Path linking to some action">
 </body>
 </html>
--- a/test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
+++ b/test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
@@ -25,6 +25,56 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
 {
     public class ImageTagHelperTest
     {
+        [Theory]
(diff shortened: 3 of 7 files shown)


src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
```diff
@@ -75,23 +75,20 @@ public class ImageTagHelper : UrlResolutionTagHelper
         /// <inheritdoc />
         public override void Process(TagHelperContext context, TagHelperOutput output)
         {
+            output.CopyHtmlAttribute(SrcAttributeName, context);
+            ProcessUrlAttribute(SrcAttributeName, output);
```
dougbu: For complete consistency, these two lines should be in an `if (Src != null)` block. Potentially doing extra work as-is.

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
```diff
@@ -75,23 +75,20 @@ public class ImageTagHelper : UrlResolutionTagHelper
...
-                if (TryResolveUrl(Src, encodeWebRoot: false, resolvedUrl: out resolvedUrl))
-                {
-                    Src = resolvedUrl;
-                }
+                // Retrieve the TagHelperOutput variation of the "src" attribute in case other TagHelpers in the
+                // pipeline have touched the value. If the value is already encoded this ScriptTagHelper may
+                // not function properly.
+                Src = output.Attributes[SrcAttributeName]?.Value as string;
+
                 output.Attributes[SrcAttributeName] = _fileVersionProvider.AddFileVersionToPath(Src);
```
dougbu: As the other tag helpers do, need to `null`-check `Src` before calling `AddFileVersionToPath()`. That method doesn't deal w/ `null`.

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
```diff
@@ -75,23 +75,20 @@ public class ImageTagHelper : UrlResolutionTagHelper
...
                 EnsureFileVersionProvider();
 
-                string resolvedUrl;
-                if (TryResolveUrl(Src, encodeWebRoot: false, resolvedUrl: out resolvedUrl))
-                {
-                    Src = resolvedUrl;
-                }
+                // Retrieve the TagHelperOutput variation of the "src" attribute in case other TagHelpers in the
+                // pipeline have touched the value. If the value is already encoded this ScriptTagHelper may
+                // not function properly.
```
dougbu: How exactly does the system fail? For example I expect it currently would NRE in `AddFileVersionToPath()` and would no-op the cache-busting addition once `Src` is `null`-checked.

test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.Image.html
```diff
@@ -12,24 +12,24 @@
     <img src="/images/red.png" alt="Red block" title="&lt;the title>">
 
     <!-- Plain image tag with file version -->
-    <img alt="Red versioned" title="Red versioned" src="/images/red.png?v=W2F5D366_nQ2fQqUk3URdgWy2ZekXjHzHJaY5yaiOOk" />
+    <img src="/images/red.png?v=W2F5D366_nQ2fQqUk3URdgWy2ZekXjHzHJaY5yaiOOk" alt="Red versioned" title="Red versioned" />
```
dougbu: :dancers:

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
```diff
@@ -25,6 +25,47 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
...
+                Src = src,
+            };
+
+            // Act
+            helper.Process(context, output);
+
+            // Assert
+            Assert.Equal(
+                "test.jpg?v=f4OxZX_x_FO5LcGBSKHWXfwtSx-j1ncoSt3SABJtkGk",
+                output.Attributes["src"].Value as string,
```
dougbu: Conditional cast makes this assertion meaningless. Is `Value` `null` or a match for the expected string?

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
```diff
@@ -25,6 +25,47 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
...
+                new TagHelperAttributeList
+                {
+                    { "alt", new HtmlString("Testing") },
+                    { "asp-append-version", true },
+                });
+            var context = MakeTagHelperContext(allAttributes);
+            var outputAttributes = new TagHelperAttributeList
+                {
+                    { "alt", new HtmlString("Testing") },
+                    { "src", "test.jpg" },
```
dougbu: Also need to test `new HtmlString("test.jpg")` and `"~/test.jpg"` for `output.Attributes` in this test method as well.

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
NTaylorMullen: :-1: `Src` can't be `null` it's a required attribute.

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
NTaylorMullen: The only way the attribute could ever explode is if a previous `TagHelper` author did the oddity of: `output.Attributes["src"] = null;`

Essentially I'm being overprotective here and `null` checking below with `?.`. Going to remove the `?.` to not portray that functionality is proper practice.

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
NTaylorMullen: https://github.com/aspnet/Mvc/pull/3006/files#r37560599

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
NTaylorMullen: `Value` is of type `object`. Not sure I get what your comment.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
NTaylorMullen: `new HtmlString("test.jpg")` Is expected to explode/have bad behavior hence the comments in each `TagHelper`. We state that it's not specifically supported. Will add `~/test.jpg` however.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
dougbu: Do the cast unconditionally. If it fails for one member of the data set, split the test in two and `Assert.Null()` in one.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
dougbu: Explode? Should just no-op. Is there another bug here?

NTaylorMullen: Sorry, explode is too explosive :trollface:. It may not function properly due to encoding.

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Mvc.TagHelpers/ImageTagHelper.cs
dougbu: :ok: I agree it can't be `null` due to the requirements in `[TargetElement]` and the way we generate code for required `string` attributes.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
```diff
@@ -25,6 +25,56 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
...
+                    { "alt", new HtmlString("Testing") },
+                    { "src", srcOutput },
+                };
+            var output = new TagHelperOutput("img", outputAttributes);
+            var hostingEnvironment = MakeHostingEnvironment();
+            var viewContext = MakeViewContext();
+            var urlHelper = new Mock<IUrlHelper>();
+            urlHelper
+                .Setup(urlhelper => urlhelper.Content(It.IsAny<string>()))
+                .Returns(new Func<string, string>(url => url.Replace("~/", "/")));
```
dougbu: Suggest making the substitution a bit more unique since this looks like something the system might do by default e.g. `.Replace("~/", "/virtualRoot/")`.

dougbu:
:shipit:

Issue #2788, opened 2015-07-07
Tag helpers are not processed inside IE conditional comments


Taritsyn:
I use ASP.NET 5 Beta 5. In `_Layout.cshtml` file I have the following code:

	

	
		<script src="//ajax.aspnetcdn.com/ajax/jQuery/jquery-2.0.3.min.js"
			asp-fallback-src-include="~/scripts/lib/jquery/jquery.min.js"
			asp-fallback-test="window.jquery"
			asp-file-version="true">
		</script>
	

The output I get the following HTML code:

	

	
		<script src="//ajax.aspnetcdn.com/ajax/jQuery/jquery-2.0.3.min.js">
		</script>
	<script>(window.jquery||document.write("<script src=\"\/scripts\/lib\/jquery\/jquery.min.js?v=pXtSQrmprcTB74RsNlFHuJxHK5zXcPrOMx78uWU0ayU\"><\/script>"));</script>
	

From HTML code it is seen, that the tag helper in the first conditional comment was not processed.

danroth27:
@DamianEdwards 

We'd like to avoid adding support for IE conditionals in the Razor parser given that "as of Internet Explorer 10, conditional comments are no longer supported by standards mode" (see https://msdn.microsoft.com/en-us/library/ms537512(v=VS.85).aspx). One thought is that it should be straight forward to create a tag helper that will render the IE conditional for you.

@ajaybhargavb Can you put together a sample for this? Talk to @Eilon for some design ideas.

Eilon:
Something like this:

```html
<conditionalComment mode="DownlevelHidden" expression="lt IE 9">
    <script src="//ajax.aspnetcdn.com/ajax/jQuery/jquery-1.11.1.min.js"
        asp-fallback-src-include="~/scripts/lib/jquery-compat/jquery.min.js"
        asp-fallback-test="window.jquery"
        asp-file-version="true">
    </script>
</conditionalComment>

<conditionalComment mode="DownlevelRevealed" expression="gte IE 9">
    <script src="//ajax.aspnetcdn.com/ajax/jQuery/jquery-2.0.3.min.js"
        asp-fallback-src-include="~/scripts/lib/jquery/jquery.min.js"
        asp-fallback-test="window.jquery"
        asp-file-version="true">
    </script>
</conditionalComment>
```

bandico0t:
Hello guys.
Is this problem resolved?
I've found this link https://forge.typo3.org/issues/17640.
[...]

PR #3007, opened 2015-08-20
Added a sample taghelper that handles IE conditional comments

ajaybhargavb:
#2788 

@NTaylorMullen

### DIFF
--- a/samples/TagHelperSample.Web/TagHelpers/ConditionalCommentTagHelper.cs
+++ b/samples/TagHelperSample.Web/TagHelpers/ConditionalCommentTagHelper.cs
@@ -1,0 +1,37 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNet.Razor.Runtime.TagHelpers;
+
+namespace TagHelperSample.Web
+{
+    [TargetElement("iecondition")]
+    public class ConditionalCommentTagHelper : TagHelper
+    {
+        public CommentMode Mode { get; set; }
+
+        public string Condition { get; set; }
+
+        public override void Process(TagHelperContext context, TagHelperOutput output)
+        {
+            output.TagName = null;
+
+            var modeModifier = string.Empty;
+
+            if (Mode == CommentMode.DownlevelHidden)
+            {
+                modeModifier = "--";
+            }
+
+            output.PreContent.Append("<!");
+            output.PreContent.Append(modeModifier);
+            output.PreContent.Append("[if ");
+            output.PreContent.Append(Condition);
+            output.PreContent.Append("]>");
+
+            output.PostContent.Append("<![endif]");
+            output.PostContent.Append(modeModifier);
+            output.PostContent.Append(">");
+        }
+    }
+}
--- a/samples/TagHelperSample.Web/Views/TagHelper/ConditionalComment.cshtml
+++ b/samples/TagHelperSample.Web/Views/TagHelper/ConditionalComment.cshtml
@@ -1,0 +1,9 @@
+﻿@using TagHelperSample.Web
+@addTagHelper "*, TagHelperSample.Web"
+
+<iecondition mode="@CommentMode.DownlevelRevealed" condition="gt IE 7">
+    <p>Content visible to all browsers newer than Internet Explorer 7.</p>
+</iecondition>
+<iecondition mode="@CommentMode.DownlevelHidden" condition="IE 7">
+    <p>Content visible only to Internet Explorer 7 users.</p>
+</iecondition>
--- a/samples/TagHelperSample.Web/Controllers/TagHelperController.cs
+++ b/samples/TagHelperSample.Web/Controllers/TagHelperController.cs
@@ -1,0 +1,15 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNet.Mvc;
+
+namespace TagHelperSample.Web.Controllers
+{
+    public class TagHelperController : Controller
+    {
+        public IActionResult ConditionalComment()
+        {
+            return View();
+        }
+    }
+}
--- a/samples/TagHelperSample.Web/TagHelpers/CommentMode.cs
+++ b/samples/TagHelperSample.Web/TagHelpers/CommentMode.cs
@@ -1,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace TagHelperSample.Web
+{
+    public enum CommentMode
+    {
+        DownlevelHidden = 0,
+
+        DownlevelRevealed = 1
+    }
+}


samples/TagHelperSample.Web/Controllers/HomeController.cs
```diff
@@ -76,5 +76,10 @@ public IActionResult Edit(int id, User user)
             ViewBag.Items = _items;
             return View();
         }
+
+        public IActionResult ConditionalComment()
+        {
+            return View();
+        }
```
rynowak: Seems like a better fit to create a new controllers. This doesn't really jive with the other actions here.

samples/TagHelperSample.Web/TagHelpers/ConditionalCommentTagHelper.cs
```diff
@@ -0,0 +1,38 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNet.Razor.Runtime.TagHelpers;
+
+namespace TagHelperSample.Web
+{
+    [TargetElement("iecondition")]
+    public class ConditionalCommentTagHelper : TagHelper
+    {
+        [HtmlAttributeName("mode")]
```
NTaylorMullen: No need for attribute.

samples/TagHelperSample.Web/TagHelpers/ConditionalCommentTagHelper.cs
```diff
@@ -0,0 +1,38 @@
...
+        public string Mode { get; set; }
+
+        [HtmlAttributeName("condition")]
```
NTaylorMullen: ditto

samples/TagHelperSample.Web/TagHelpers/ConditionalCommentTagHelper.cs
```diff
@@ -0,0 +1,38 @@
...
+        public string Condition { get; set; }
+
+        public override void Process(TagHelperContext context, TagHelperOutput output)
+        {
+            output.TagName = null;
+            string startCondition = "";
+            string endCondition = "";
```
dougbu: `string.Empty`

samples/TagHelperSample.Web/TagHelpers/ConditionalCommentTagHelper.cs
```diff
@@ -0,0 +1,38 @@
...
+
+using Microsoft.AspNet.Razor.Runtime.TagHelpers;
+
+namespace TagHelperSample.Web
+{
+    [TargetElement("iecondition")]
+    public class ConditionalCommentTagHelper : TagHelper
+    {
+        [HtmlAttributeName("mode")]
+        public string Mode { get; set; }
```
NTaylorMullen: Since we're building this specific `TagHelper` to showcase it as something a user can do lets make this an `Enum` to showcase proper usage.

samples/TagHelperSample.Web/TagHelpers/ConditionalCommentTagHelper.cs
```diff
@@ -0,0 +1,38 @@
...
+            string endCondition = "";
+            if (Mode == "downlevelHidden")
+            {
+                startCondition = $"<!--[if {Condition}]>";
+                endCondition = "<![endif]-->";
+
+            }
+            else if (Mode == "downlevelRevealed")
+            {
+                startCondition = $"<![if {Condition}]>";
```
dougbu: What exactly are the encoding requirements for an IE conditional comment? This could have problems with HTML encoding. If so, much better to explicitly `Append()` the different parts because they have distinct encoding requirements e.g.
``` c#
output.PreContent.Append("<![if ");
output.PreContent.Append(new StringHtmlContent(Condition));
output.PreConent.Append("]>");
```

If HTML encoding were not an issue, should still be done using simple string concatenation.

samples/TagHelperSample.Web/TagHelpers/ConditionalCommentTagHelper.cs
NTaylorMullen: Actually, do we even need the `Mode` Why not transform the tag itself into the condition? Do we need to support both types of conditions (I'm unaware if they do something different)? If we're going for functionality here I don't see the value in supporting both.

samples/TagHelperSample.Web/TagHelpers/ConditionalCommentTagHelper.cs
```diff
@@ -0,0 +1,38 @@
...
+        public override void Process(TagHelperContext context, TagHelperOutput output)
+        {
+            output.TagName = null;
+            string startCondition = "";
+            string endCondition = "";
+            if (Mode == "downlevelHidden")
+            {
+                startCondition = $"<!--[if {Condition}]>";
+                endCondition = "<![endif]-->";
+
```
dougbu: remove blank line

samples/TagHelperSample.Web/TagHelpers/ConditionalCommentTagHelper.cs
```diff
@@ -0,0 +1,38 @@
...
+            }
+            else if (Mode == "downlevelRevealed")
+            {
+                startCondition = $"<![if {Condition}]>";
+                endCondition = "<![endif]>";
+            }
+
+            output.PreContent.Append(startCondition);
```
NTaylorMullen: `output.PreContent.SetCOntent(startCondition);` Same below.

samples/TagHelperSample.Web/TagHelpers/ConditionalCommentTagHelper.cs
```diff
@@ -0,0 +1,38 @@
...
+            {
+                startCondition = $"<!--[if {Condition}]>";
+                endCondition = "<![endif]-->";
+
+            }
+            else if (Mode == "downlevelRevealed")
+            {
+                startCondition = $"<![if {Condition}]>";
+                endCondition = "<![endif]>";
+            }
```
dougbu: Not easy to tell the difference between what the `if` and `else` blocks do. If `Mode` remains (seeing @NTaylorMullen's recent [comment](https://github.com/aspnet/Mvc/pull/3007#discussion-diff-37492621)), suggest instead storing `string.Empty` or `"--"` in a variable and using that in the string concatenation / `Append()` expressions.

samples/TagHelperSample.Web/Views/Home/ConditionalComment.cshtml
```diff
@@ -0,0 +1,12 @@
+﻿@addTagHelper "*, TagHelperSample.Web"
+
+<iecondition mode="downlevelRevealed" condition="gt IE 7">
+    <p>
+        Content visible to all browsers newer than Internet Explorer 7.
```
NTaylorMullen: `<p>Content visible to all browsers newer than Internet Explorer 7.</p>` Same format for next one.

NTaylorMullen:
:watch:

samples/TagHelperSample.Web/TagHelpers/ConditionalCommentTagHelper.cs
dougbu: Can't HTML encode because `&` is a special character in the [syntax](https://msdn.microsoft.com/en-us/library/ms537512.aspx). So there's no secure way to do this.

But still not great to do the extra concatenation. Updated my previous comment...

dougbu:
:watch:

samples/TagHelperSample.Web/TagHelpers/ConditionalCommentTagHelper.cs
NeelBhatt: Also,
var startCondition = string.Empty;
var endCondition = string.Empty;

If you do not mind!

samples/TagHelperSample.Web/TagHelpers/ConditionalCommentTagHelper.cs
NTaylorMullen: Talked in person: Conditional comments do different things based on the IE version. Change Mode into an enum as noted in my initial comment on this line.

ajaybhargavb:
Updated.

samples/TagHelperSample.Web/Views/TagHelper/ConditionalComment.cshtml
```diff
@@ -0,0 +1,9 @@
+﻿@using TagHelperSample.Web
+@addTagHelper "*, TagHelperSample.Web"
+
+<iecondition mode="@CommentMode.DownlevelRevealed" condition="gt IE 7">
```
NTaylorMullen: No need for the `@`.

Ditto below.

NTaylorMullen:
:shipit:

dougbu:
:shipit:

Issue #2947, opened 2015-08-10
Response.ContentType doesn't work in MVC controller


Eilon:
From @Daniel15 in https://github.com/aspnet/HttpAbstractions/issues/364:

I'm using Razor to generate some XML (an RSS feed). I have this in one of my ASP.NET MVC controllers:
```
Response.ContentType = "application/rss+xml";
return View("Foo")
```
However IIS Express is still serving the response as text/html.

This doesn't work either:
```
Response.Headers.Set("Content-Type", "application/rss+xml");
```

However, setting a custom header (eg. `X-Test`) *does* work.

(I wasn't sure whether to report this here or in the ASP.NET MVC repo)

Eilon:
From me:

We should have the view engine only set the content type if it's not already set. According to @Tratcher this is trivial to do.

PR #3009, opened 2015-08-20
Fix: User specified content type is ignored

ajaybhargavb:
#2947 

@rynowak

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ContentResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ContentResult.cs
@@ -34,31 +34,18 @@ namespace Microsoft.AspNet.Mvc
         public override async Task ExecuteResultAsync([NotNull] ActionContext context)
         {
             var response = context.HttpContext.Response;
-
             var contentTypeHeader = ContentType;
-            Encoding encoding;
-            if (contentTypeHeader == null)
-            {
-                contentTypeHeader = DefaultContentType;
-                encoding = Encoding.UTF8;
-            }
-            else
-            {
-                if (contentTypeHeader.Encoding == null)
-                {
-                    // Do not modify the user supplied content type, so copy it instead
-                    contentTypeHeader = contentTypeHeader.Copy();
-                    contentTypeHeader.Encoding = Encoding.UTF8;
 
-                    encoding = Encoding.UTF8;
-                }
-                else
-                {
-                    encoding = contentTypeHeader.Encoding;
-                }
+            if (contentTypeHeader != null && contentTypeHeader.Encoding == null)
+            {
+                // Do not modify the user supplied content type, so copy it instead
+                contentTypeHeader = contentTypeHeader.Copy();
+                contentTypeHeader.Encoding = Encoding.UTF8;
             }
-
-            response.ContentType = contentTypeHeader.ToString();
+            
+            response.ContentType = contentTypeHeader?.ToString()
+                ?? response.ContentType
+                ?? DefaultContentType.ToString();
 
             if (StatusCode != null)
             {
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNet.Mvc.ViewFeatures/ViewExecutor.cs
```diff
@@ -36,7 +36,20 @@ public static class ViewExecutor
         {
             var response = actionContext.HttpContext.Response;
 
-            contentType = contentType ?? DefaultContentType;
+            if (contentType == null)
```
kichalla: Not sure if we want to consider other action results which have similar issue (ex: https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/ContentResult.cs#L40)

ajaybhargavb: Good point. I'll update it and check for other results with the same issue.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewExecutor.cs
```diff
@@ -36,7 +36,20 @@ public static class ViewExecutor
...
+            {
+                contentType = DefaultContentType;
+
+                if (!string.IsNullOrEmpty(response.ContentType))
+                {
+                    MediaTypeHeaderValue parsedContentType;
+                    if (MediaTypeHeaderValue.TryParse(response.ContentType, out parsedContentType))
+                    {
+                        contentType = parsedContentType;
+                    }
```
rynowak: Why do we need to do this? Why does MVC care if you set an invalid content type?

ajaybhargavb: Hmm. Just curious, we already check if the encoding is set in a user supplied content type. Why not check for invalid ones?

rynowak: The WebAPI design for all of this was to make the header classes unable to accept invalid header values.

It was limiting in some scenarios where users had to interop with non-compliant other code - that's part of the reason why these are defined as strings on the http abstractions classes.

I think it's fine to add an encoding if it's valid... but we shouldn't mess with what the user set if it's invalid. That masks bugs.

ajaybhargavb: Got it. Will update.

ajaybhargavb: We still need to parse it. When the user specifies the encoding along with the content type, it needs to be used by the writer here. https://github.com/aspnet/Mvc/pull/3009/files#diff-62f5af455fe6d271af2d9522b4841df9L49

ajaybhargavb:
Updated.

src/Microsoft.AspNet.Mvc.Core/ContentResult.cs
```diff
@@ -34,31 +34,16 @@ public class ContentResult : ActionResult
...
+            if (contentTypeHeader != null && contentTypeHeader.Encoding == null)
+            {
+                // Do not modify the user supplied content type, so copy it instead
+                contentTypeHeader = contentTypeHeader.Copy();
+                contentTypeHeader.Encoding = Encoding.UTF8;
             }
-
-            response.ContentType = contentTypeHeader.ToString();
+            
+            response.ContentType = contentTypeHeader?.ToString() ?? response.ContentType ?? DefaultContentType.ToString();
```
rynowak: long line

rynowak:
:shipit:

PR #3010, opened 2015-08-20
Minor cleanup of service registration in Razor

pranavkm:
* Avoid lazyily evaluating IRazorCompilationService in VirtualPathRazorPageFactory
* Cleanup comments on the lifetime of Razor services.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor/DependencyInjection/MvcRazorMvcCoreBuilderExtensions.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/DependencyInjection/MvcRazorMvcCoreBuilderExtensions.cs
@@ -96,28 +96,22 @@ namespace Microsoft.Framework.DependencyInjection
                 return new DefaultChunkTreeCache(cachedFileProvider.Options.FileProvider);
             }));
 
-            // The host is designed to be discarded after consumption and is very inexpensive to initialize.
-            services.TryAddTransient<IMvcRazorHost, MvcRazorHost>();
-
             // Caches compilation artifacts across the lifetime of the application.
             services.TryAddSingleton<ICompilerCacheProvider, DefaultCompilerCacheProvider>();
 
-            // This caches compilation related details that are valid across the lifetime of the application
-            // and is required to be a singleton.
+            // This caches compilation related details that are valid across the lifetime of the application.
             services.TryAddSingleton<ICompilationService, RoslynCompilationService>();
 
-            // Both the compiler cache and roslyn compilation service hold on the compilation related
-            // caches. RazorCompilation service is just an adapter service, and it is transient to ensure
-            // the IMvcRazorHost dependency does not maintain state.
-            services.TryAddTransient<IRazorCompilationService, RazorCompilationService>();
-
-            // The ViewStartProvider needs to be able to consume scoped instances of IRazorPageFactory
-            services.TryAddScoped<IViewStartProvider, ViewStartProvider>();
+            // In the default scenario the following services are singleton by virtue of being initialized as part of
+            // creating the singleton RazorViewEngine instance.
             services.TryAddTransient<IRazorViewFactory, RazorViewFactory>();
-            services.TryAddSingleton<IRazorPageActivator, RazorPageActivator>();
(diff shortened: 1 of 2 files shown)


pranavkm:
cc @rynowak

src/Microsoft.AspNet.Mvc.Razor/VirtualPathRazorPageFactory.cs
```diff
@@ -14,40 +13,23 @@ namespace Microsoft.AspNet.Mvc.Razor
     /// </summary>
     public class VirtualPathRazorPageFactory : IRazorPageFactory
     {
-        private readonly IServiceProvider _serviceProvider;
+        private readonly Func<RelativeFileInfo, CompilationResult> _compileDelegate;
```
dougbu: Curious: Why use a delegate instead of `... IRazorCompilationService _razorCompilationService`?

src/Microsoft.AspNet.Mvc.Razor/DependencyInjection/MvcRazorMvcCoreBuilderExtensions.cs
```diff
@@ -96,29 +96,23 @@ internal static void AddRazorViewEngineServices(IServiceCollection services)
...
-            // This caches compilation related details that are valid across the lifetime of the application
-            // and is required to be a singleton.
+            // This caches compilation related details that are valid across the lifetime of the application.
             services.TryAddSingleton<ICompilationService, RoslynCompilationService>();
 
-            // Both the compiler cache and roslyn compilation service hold on the compilation related
-            // caches. RazorCompilation service is just an adapter service, and it is transient to ensure
-            // the IMvcRazorHost dependency does not maintain state.
+            // In the default scenario the following services are singleton by virtue of being initialized as part of
+            // creating the singleton RazorViewEngine instance.
```
dougbu: Question: If these are effectively singletons why aren't they added as singletons? Does something else reference them in a way that requires additional instances?

src/Microsoft.AspNet.Mvc.Razor/VirtualPathRazorPageFactory.cs
pranavkm: None in particular. Seemed sort of intuitive since we don't need access to `IRazorCompilationService` later on.

src/Microsoft.AspNet.Mvc.Razor/DependencyInjection/MvcRazorMvcCoreBuilderExtensions.cs
rynowak: There's not really a normative answer. If they are singletons the SP has to track their lifetime, which may entail extra work. I usually just do Transient by default, and only do something else if you have a reason.

src/Microsoft.AspNet.Mvc.Razor/VirtualPathRazorPageFactory.cs
rynowak: IMO it's good to document in this code that we're keeping a live reference to it through the `Func<>`

src/Microsoft.AspNet.Mvc.Razor/DependencyInjection/MvcRazorMvcCoreBuilderExtensions.cs
pranavkm: Nothing else I know of references these. Other services seem to have a justification as to why they are singletons, these simply don't.

src/Microsoft.AspNet.Mvc.Razor/VirtualPathRazorPageFactory.cs
pranavkm: :+1:

src/Microsoft.AspNet.Mvc.Razor/DependencyInjection/MvcRazorMvcCoreBuilderExtensions.cs
pranavkm: I'm going to leave this as is if there aren't strong arguments against doing so.

pranavkm:
Updated.

PR #3011, opened 2015-08-20
Make ValueProviderResult a string-ish struct

rynowak:
/cc @dougbu


src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
```diff
@@ -3,227 +3,361 @@
...
-            // case 4: both destination + source type are single elements, so convert
-            return ConvertSimpleType(culture, value, destinationType);
+        /// <summary>
+        /// Gets the number of submitted values.
+        /// </summary>
+        public int Length
+        {
+            get
+            {
+                return Values?.Length ?? (Value != null ? 1 : 0);
```
pranavkm: And you think my lambda properties look "eww"

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
```diff
@@ -3,227 +3,361 @@
...
+        }
+
+        /// <summary>
+        /// Converts the provided <see cref="ValueProviderResult"/> into a comma-separated string containing all
+        /// submitted values.
+        /// </summary>
+        /// <param name="result">The <see cref="ValueProviderResult"/>.</param>
+        public static implicit operator string(ValueProviderResult result)
+        {
+            if (result.Values == null)
```
pranavkm: Isn't this equivalent to `First`?

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
```diff
@@ -3,227 +3,361 @@
...
+                return true;
+            }
+        }
 
-            // In case of a Nullable object, we try again with its underlying type.
-            destinationType = UnwrapNullableType(destinationType);
+        /// <inheritdoc />
+        public override int GetHashCode()
+        {
+            return ((string)this).GetHashCode();
```
pranavkm: The implicit conversion can return `null` which would cause this to throw.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
pranavkm: Can we avoid getting the nullable int here?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/BodyModelBinder.cs
```diff
@@ -64,9 +64,9 @@ public BodyModelBinder()
                     // system to skip other model binders and never to fall back.
                     return new ModelBindingResult(modelBindingKey);
                 }
-
-                var valueProviderResult = new ValueProviderResult(rawValue: model);
-                bindingContext.ModelState.SetModelValue(modelBindingKey, valueProviderResult);
+                
+                // We don't have anything useful to put in model state.
+                bindingContext.ModelState.SetModelValue(modelBindingKey, rawValue: null, attemptedValue: null);
```
pranavkm: `SetModelValue` has `NotNull` on both those arguments. Should that requirement be removed?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/ByteArrayModelBinder.cs
```diff
@@ -31,9 +31,10 @@ public class ByteArrayModelBinder : IModelBinder
             bindingContext.ModelState.SetModelValue(bindingContext.ModelName, valueProviderResult);
 
             // Check for missing data case 2: There was an <input ... /> element but it was left blank.
-            var value = valueProviderResult.AttemptedValue;
+            var value = valueProviderResult.FirstValue;
             if (string.IsNullOrEmpty(value))
             {
+
```
pranavkm: Remove

src/Microsoft.AspNet.Mvc.Core/ModelBinding/SimpleTypeModelBinder.cs
```diff
@@ -8,12 +8,10 @@
 
 namespace Microsoft.AspNet.Mvc.ModelBinding
 {
-    public sealed class TypeConverterModelBinder : IModelBinder
+    public class SimpleTypeModelBinder : IModelBinder
     {
         public async Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindingContext)
         {
-            ModelBindingHelper.ValidateBindingContext(bindingContext);
```
pranavkm: Do we not need need this anymore?

src/Microsoft.AspNet.Mvc.Core/ModelBindingHelper.cs
```diff
@@ -530,5 +524,192 @@ public static object ConvertValuesToCollectionType<T>(Type modelType, IList<T> v
...
+        /// <summary>
+        /// Converts the provided <paramref name="value"/> to a value of <see cref="Type"/> <typeparamref name="T"/>
+        /// using the <see cref="CultureInfo.InvariantCulture"/>.
+        /// </summary>
+        /// <typeparam name="T">The <see cref="Type"/> for conversion.</typeparam>
+        /// <param name="value">The value to convert."/></param>
+        /// <returns>
+        /// The converted value or the default value of <typeparamref name="T"/> if the value could not be converted.
+        /// </returns>
+        public static T ConvertTo<T>(object value)
```
pranavkm: `TVal`

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
rynowak: I'll just rewrite as `if`

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
rynowak: good catch

src/Microsoft.AspNet.Mvc.Core/ModelBinding/BodyModelBinder.cs
rynowak: Yeah, we should. I'll fix that.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/ByteArrayModelBinder.cs
rynowak: sure

src/Microsoft.AspNet.Mvc.Core/ModelBinding/SimpleTypeModelBinder.cs
rynowak: We shouldn't have stuff like this in the code. If we care about this kind of validation, we'd choose a better approach.

src/Microsoft.AspNet.Mvc.Core/ModelBindingHelper.cs
rynowak: no

src/Microsoft.AspNet.Mvc.Core/ModelBindingHelper.cs
```diff
@@ -530,5 +524,192 @@ public static object ConvertValuesToCollectionType<T>(Type modelType, IList<T> v
...
+                    Resources.FormatValueProviderResult_NoConverterExists(value.GetType(), destinationType));
+            }
+
+            try
+            {
+                return canConvertFrom
+                           ? converter.ConvertFrom(null, culture, value)
+                           : converter.ConvertTo(null, culture, value, destinationType);
+            }
+            catch (Exception ex)
```
pranavkm:
```C#
catch (Exception ex)
     when (!(ex is FormatException))
{

}
```

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/SimpleHttpValueProvider.cs
```diff
@@ -59,11 +59,28 @@ public Task<bool> ContainsPrefixAsync(string prefix)
 
         public Task<ValueProviderResult> GetValueAsync(string key)
         {
-            ValueProviderResult result = null;
+            ValueProviderResult result = ValueProviderResult.None;
+
```
pranavkm: A bit unrelated, but we should probably rename this. There's nothing `Http` about it (anymore).

davidfowl:
Why is the ci failing?

pranavkm:
Travis: `Cannot open assembly 'packages/Sake/tools/Sake.exe': No such file or directory.`

davidfowl:
I just force a rerun

rynowak:
@pranavkm updated

src/Microsoft.AspNet.Mvc.Core/ModelBinding/BodyModelBinder.cs
rynowak: Fixed it.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
```diff
@@ -359,14 +361,30 @@ public void Merge(ModelStateDictionary dictionary)
...
 
         /// <summary>
-        /// Sets the value for the <see cref="ModelState"/> with the specified <paramref name="key"/> to the
-        /// specified <paramref name="value"/>.
+        /// Sets the value for the <see cref="ModelState"/> with the specified <paramref name="key"/>.
         /// </summary>
         /// <param name="key">The key for the <see cref="ModelState"/> entry.</param>
-        /// <param name="value">The value to assign.</param>
-        public void SetModelValue([NotNull] string key, [NotNull] ValueProviderResult value)
+        /// <param name="value">The raw values for the <see cref="ModelState"/> entry.</param>
```
dougbu: `<param name="rawValue">`

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
```diff
@@ -359,14 +361,30 @@ public void Merge(ModelStateDictionary dictionary)
         }
 
         /// <summary>
-        /// Sets the value for the <see cref="ModelState"/> with the specified <paramref name="key"/> to the
-        /// specified <paramref name="value"/>.
+        /// Sets the value for the <see cref="ModelState"/> with the specified <paramref name="key"/>.
```
dougbu: `value` is a bit vague here. Perhaps worth saying `Sets <see cref="ModelState.RawValue"> and <see cref="ModelState.AttemptedValue"> ...`

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
```diff
@@ -3,227 +3,372 @@
...
+    /// </para>
+    /// <para>
+    /// Treat <see cref="ValueProviderResult"/> as an <see cref="IEnumerable{string}"/> to consume all values,
+    /// regardless of whether a single value or multiple values were submitted.
+    /// </para>
+    /// </remarks>
+    public struct ValueProviderResult : IEquatable<ValueProviderResult>, IEnumerable<string>
     {
-        private static readonly CultureInfo _staticCulture = CultureInfo.InvariantCulture;
+        private static readonly CultureInfo _invariantCulture = CultureInfo.InvariantCulture;
```
dougbu: :+1:

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
```diff
@@ -3,227 +3,372 @@
...
 
         /// <summary>
-        /// <see cref="CultureInfo"/> to use in <see cref="ConvertTo(Type)"/> or
-        /// <see cref="ConvertTo(Type, CultureInfo)"/> if passed <see cref="CultureInfo"/> is <c>null</c>.
+        /// Creates a new <see cref="ValueProviderResult"/>..
         /// </summary>
-        public CultureInfo Culture { get; }
+        /// <param name="value">The submitted value.</param>
+        /// <param name="culture">The <see cref="CultureInfo"/> associated with this value.</param>
+        public ValueProviderResult(string value, [NotNull] CultureInfo culture)
```
dougbu: Why impose the new requirement on `culture`? Was previously fine to pass `null`.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
```diff
@@ -3,227 +3,372 @@
...
+            if (value == null)
+            {
+                Value = null;
+                Values = None.Values;
+            }
+            else
+            {
+                Value = value;
+                Values = null;
+            }
```
dougbu: Why not use `: this(value)` to avoid repeating these blocks?

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
```diff
@@ -3,227 +3,372 @@
...
+                Values = values;
+            }
+            
+            Culture = _invariantCulture;
         }
 
         /// <summary>
-        /// Converts <see cref="RawValue"/> to the given <paramref name="type"/> using the given
-        /// <paramref name="culture"/>.
+        /// Creates a new <see cref="ValueProviderResult"/>..
```
dougbu: Not dot dot.

pranavkm:
Didn't look at the tests in detail - other than the the `ValueProviderResult` one. :shipit:

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
```diff
@@ -3,227 +3,372 @@
...
-        /// <paramref name="rawValue"/>. Initializes <see cref="Culture"/> to
-        /// <see cref="CultureInfo.InvariantCulture"/>.
+        /// A <see cref="ValueProviderResult"/> that represents a lack of data.
         /// </summary>
-        /// <param name="rawValue">The <see cref="RawValue"/> value of the new instance.</param>
-        public ValueProviderResult(object rawValue)
-            : this(rawValue, attemptedValue: null, culture: _staticCulture)
-        {
-        }
+        public static ValueProviderResult None = new ValueProviderResult(new string[0]);
```
dougbu: Minor but `Enumerable.Empty<string>()` would share the array more broadly.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
```diff
@@ -3,227 +3,372 @@
...
+            if (value == null)
+            {
+                Value = null;
+                Values = None.Values;
+            }
+            else
+            {
+                Value = value;
+                Values = null;
+            }
```
dougbu: Why explicitly init properties to `null`?

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
```diff
@@ -3,227 +3,372 @@
...
+        public string Value { get; private set; }
+
+        /// <summary>
+        /// Gets or sets an array of values. Will be <c>null</c> if only a single value was provided.
+        /// </summary>
+        public string[] Values { get; private set; }
+
+        /// <summary>
+        /// Gets the first value based on the order values were provided in the request. Use <see cref="FirstValue"/>
+        /// to get a single value for processing regarless of whehter a single or multiple values were provided
```
dougbu: tpyo: "whether"

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
```diff
@@ -3,227 +3,372 @@
...
+        }
+
+        /// <inheritdoc />
+        public override string ToString()
+        {
+            return (string)this;
+        }
+
+        /// <inheritdoc />
+        public Enumerator GetEnumerator()
```
dougbu: Where would this inherit docs from? Already have explicit implementations of both methods in `IEnumerable<string>`. Might want to make this method `private Enumerator GetMyEnumerator()`.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
```diff
@@ -3,227 +3,372 @@
...
 
-            // if this is a user-input value but the user didn't type anything, return no value
-            var valueAsString = value as string;
-            if (valueAsString != null && string.IsNullOrWhiteSpace(valueAsString))
+        /// <summary>
+        /// Converts the provided <see cref="ValueProviderResult"/> into a comma-separated string containing all
+        /// submitted values.
+        /// </summary>
+        /// <param name="result">The <see cref="ValueProviderResult"/>.</param>
+        public static implicit operator string(ValueProviderResult result)
```
dougbu: Suggest leaving the casts as explicit. Will make the callers more readable and you seem to already be using `(string)` or `(string[])` where necessary.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/BodyModelBinder.cs
```diff
@@ -64,9 +64,9 @@ public BodyModelBinder()
                     // system to skip other model binders and never to fall back.
                     return new ModelBindingResult(modelBindingKey);
                 }
-
-                var valueProviderResult = new ValueProviderResult(rawValue: model);
-                bindingContext.ModelState.SetModelValue(modelBindingKey, valueProviderResult);
+                
+                // We don't have anything useful to put in model state.
```
dougbu: Huh? We changed this code not that long ago to explicitly get the `model` into the `ModelStateDictionary`. Thought one of the intents here was to separate `ValueProviderResult` from `ModelState`. Instead seem to be reproducing the new `ValueProviderResult` restrictions in `ModelState`.

Put another way, I expected `bindingContext.ModelState.SetModelValue(modelBindingKey, nonStringIshValue: model)`.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
rynowak: Fixed

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
rynowak: sure

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
rynowak: :ok:

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
rynowak: I was under the impression you didn't get to in a struct, but apparently you can. Fixed.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
rynowak: ya

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
rynowak: and then I'd have to cast it to an array

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
rynowak: it's a struct, you need to initialize every field.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
rynowak: fiex

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
rynowak: Fixed, the docs here. Making this public is intended.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
rynowak: Yeah I agree with this :+1:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/BodyModelBinder.cs
rynowak: > Huh? We changed this code not that long ago to explicitly get the model into the ModelStateDictionary. 

Yes, and this is an intentional change back from that. Storing arbitrary objects in `RawValue` adds more odd and barely tested scenarios for the conversion code + HTML helpers have to deal with. This change gets on track with the API supporting what `RawValue` was intended to be used for.

We don't gain anything from storing deserialized models in `RawValue` - it's just something we did for consistency. In the future, if we find a need to store the bound model value, we should create a new slot in MSD and do that instead of overloading `RawValue`.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
```diff
@@ -56,31 +56,13 @@ public class CollectionModelBinder<TElement> : ICollectionModelBinder
...
-                    model = bindingContext.Model ?? CreateEmptyCollection(bindingContext.ModelType);
-                    var validationNode =
-                        new ModelValidationNode(bindingContext.ModelName, bindingContext.ModelMetadata, model);
-
-                    return new ModelBindingResult(
-                        model,
-                        bindingContext.ModelName,
-                        isModelSet: true,
-                        validationNode: validationNode);
-                }
```
dougbu: What replaces this special case? Or, why can't it exist?

rynowak:
@dougbu updated with fixes for comments so far.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
```diff
@@ -116,14 +119,5 @@ protected override object CreateEmptyCollection(Type targetType)
 
             return CreateInstance(targetType);
         }
-
-        private static TKey ConvertFromString(string keyString)
-        {
-            // Use InvariantCulture to convert string since ExpressionHelper.GetExpressionText() used that culture.
-            var keyResult = new ValueProviderResult(keyString);
```
dougbu: Nice to see this workaround go bye-bye.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/FormFileModelBinder.cs
```diff
@@ -48,8 +48,10 @@ public class FormFileModelBinder : IModelBinder
                         SuppressValidation = true,
                     };
 
-                var valueProviderResult = new ValueProviderResult(rawValue: value);
-                bindingContext.ModelState.SetModelValue(bindingContext.ModelName, valueProviderResult);
+                bindingContext.ModelState.SetModelValue(
+                    bindingContext.ModelName,
+                    rawValue: null,
+                    attemptedValue: null);
```
dougbu: Another `ModelState` where we want the model.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
rynowak: There's no such thing as a `ValueProviderResult` with no value other than `ValueProviderResult.None`. Value providers have to plug in the empty string to return a "key was present but didn't have a value".

This case will now just go down the normal simple collection path. It was a nice surprise that it wasn't needed anymore.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/SimpleTypeModelBinder.cs
dougbu: Not sure removing `ValidateBindingContext()` fits in this PR at all. Let's discuss this separately.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/SimpleTypeModelBinder.cs
```diff
@@ -21,38 +19,48 @@ public sealed class TypeConverterModelBinder : IModelBinder
...
-            object newModel;
             bindingContext.ModelState.SetModelValue(bindingContext.ModelName, valueProviderResult);
+
             try
             {
-                newModel = valueProviderResult.ConvertTo(bindingContext.ModelType);
-                ModelBindingHelper.ReplaceEmptyStringWithNull(bindingContext.ModelMetadata, ref newModel);
+                var model = valueProviderResult.ConvertTo(bindingContext.ModelType);
+
+                if (bindingContext.ModelType == typeof(string))
```
dougbu: Condition seems incorrect. `TypeMatchModelBinder` was previously used only if `TypeConverterModelBinder` failed. It was conditional on `model`'s type, not the `ModelType` i.e. it checked the source and not the target.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/SimpleTypeModelBinder.cs
```diff
@@ -21,38 +19,48 @@ public sealed class TypeConverterModelBinder : IModelBinder
...
+                    }
+                }
+
                 var isModelSet = true;
 
                 // When converting newModel a null value may indicate a failed conversion for an otherwise required
                 // model (can't set a ValueType to null). This detects if a null model value is acceptable given the
                 // current bindingContext. If not, an error is logged.
-                if (newModel == null && !AllowsNullValue(bindingContext.ModelType))
+                if (model == null && !AllowsNullValue(bindingContext.ModelType))
```
dougbu: How will this case occur given the fact `valueProviderResult.ConvertTo()` doesn't return `null` except for reference types anymore? Have we unintentionally lost this error case?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/ValueProviderResultExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// </summary>
+        /// <typeparam name="T">The <see cref="Type"/> for conversion.</typeparam>
+        /// <param name="result">The <see cref="ValueProviderResult"/>.</param>
+        /// <returns>
+        /// The converted value, or the default value of <typeparamref name="T"/> if the value could not be converted.
+        /// </returns>
+        public static T ConvertTo<T>(this ValueProviderResult result)
+        {
+            var valueToConvert = (object)result.Values ?? (object)result.Value;
+            return ModelBindingHelper.ConvertTo<T>(valueToConvert, result.Culture);
```
dougbu: Why not call `ConvertTo(result, typeof(T))`?

src/Microsoft.AspNet.Mvc.Core/ModelBindingHelper.cs
```diff
@@ -530,5 +524,189 @@ public static object ConvertValuesToCollectionType<T>(Type modelType, IList<T> v
...
+            }
+            catch (FormatException)
+            {
+                throw;
+            }
+            catch (Exception ex)
+            {
+                // TypeConverter throws System.Exception wrapping the FormatException,
+                // so we throw the inner exception.
+                ExceptionDispatchInfo.Capture(ex.InnerException).Throw();
```
dougbu: What guarantees the `Exception` has a non-`null` `InnerException`?

src/Microsoft.AspNet.Mvc.Core/ModelBindingHelper.cs
```diff
@@ -530,5 +524,189 @@ public static object ConvertValuesToCollectionType<T>(Type modelType, IList<T> v
...
+        /// <typeparam name="T">The <see cref="Type"/> for conversion.</typeparam>
+        /// <param name="value">The value to convert."/></param>
+        /// <param name="culture">The <see cref="CultureInfo"/> for conversion.</param>
+        /// <returns>
+        /// The converted value or the default value of <typeparamref name="T"/> if the value could not be converted.
+        /// </returns>
+        public static T ConvertTo<T>(object value, CultureInfo culture)
+        {
+            var converted = ConvertTo(value, typeof(T), culture);
+            return converted == null ? default(T) : (T)converted;
```
dougbu: Why not `return (T)converted;`? The extra complexity seems to duplicate the `Activator.CreateInstance(type)` case below.

src/Microsoft.AspNet.Mvc.Core/ModelBindingHelper.cs
```diff
@@ -530,5 +524,189 @@ public static object ConvertValuesToCollectionType<T>(Type modelType, IList<T> v
...
+            if (destinationType.IsArray)
+            {
+                var destinationElementType = destinationType.GetElementType();
+                if (valueAsArray != null)
+                {
+                    // case 1: both destination + source type are arrays, so convert each element
+                    var converted = (IList)Array.CreateInstance(destinationElementType, valueAsArray.Length);
+                    for (var i = 0; i < valueAsArray.Length; i++)
+                    {
+                        converted[i] = ConvertSimpleType(valueAsArray.GetValue(i), destinationElementType, culture);
```
dougbu: Existing bug but this'll fail miserably if e.g. source is `string[]` containing a `null` element and target is `int[]`.

src/Microsoft.AspNet.Mvc.Core/ModelBindingHelper.cs
```diff
@@ -530,5 +524,189 @@ public static object ConvertValuesToCollectionType<T>(Type modelType, IList<T> v
...
+                // case 3: destination type is single element but source is array, so extract first element + convert
+                if (valueAsArray.Length > 0)
+                {
+                    value = valueAsArray.GetValue(0);
+                    return ConvertSimpleType(value, destinationType, culture);
+                }
+                else
+                {
+                    // case 3(a): source is empty array, so can't perform conversion
+                    return null;
```
dougbu: If doing the `Activator.CreateInstance(type)` thing above, should do it here too.

src/Microsoft.AspNet.Mvc.Core/ModelBindingHelper.cs
```diff
@@ -530,5 +524,189 @@ public static object ConvertValuesToCollectionType<T>(Type modelType, IList<T> v
...
+            }
+
+            // In case of a Nullable object, we try again with its underlying type.
+            destinationType = UnwrapNullableType(destinationType);
+
+            // if this is a user-input value but the user didn't type anything, return no value
+            var valueAsString = value as string;
+            if (valueAsString != null && string.IsNullOrWhiteSpace(valueAsString))
+            {
+                return null;
```
dougbu: `Activator.CreateInstance(type)` here too.

src/Microsoft.AspNet.Mvc.Core/ModelBindingHelper.cs
```diff
@@ -530,5 +524,189 @@ public static object ConvertValuesToCollectionType<T>(Type modelType, IList<T> v
...
+
+                throw new InvalidOperationException(
+                    Resources.FormatValueProviderResult_NoConverterExists(value.GetType(), destinationType));
+            }
+
+            try
+            {
+                return canConvertFrom
+                           ? converter.ConvertFrom(null, culture, value)
+                           : converter.ConvertTo(null, culture, value, destinationType);
```
dougbu: YOLO wrapping from @rynowak !

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
dougbu: Everything gets default values here too.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/BodyModelBinder.cs
dougbu: > it's just something we did for consistency

We wanted to consistently make the bound values available in the `ModelStateDictionary`. Without either "overloading" `RawValue` or adding that new `ModelState` property, looks like a regression.

Note users can write into `ModelStateDictionary`. Also HTML generation converts _whatever_ is in `RawValue` to `string` and the few other types it needs.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/SimpleTypeModelBinder.cs
dougbu: (Sorry if I made this suggestion offline.)

src/Microsoft.AspNet.Mvc.Core/ModelBinding/BodyModelBinder.cs
rynowak: > We wanted to consistently make the bound values available in the ModelStateDictionary. Without either "overloading" RawValue or adding that new ModelState property, looks like a regression.

But we never did that... `RawValue` had the values from the value provider in most cases, and the value created by the model binder in the case of binding `IFormFile` or `[FromBody]`. This is definitely a change and I think it's in the right direction.


>Note users can write into ModelStateDictionary. Also HTML generation converts whatever is in RawValue to string and the few other types it needs.

Yes. This is also intentionally changed to be not supported with arbitrary objects. If you want to do that, then you're responsible for `ToString()`ing the object, which seems pretty reasonable.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
rynowak: :dancer:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/FormFileModelBinder.cs
rynowak: This and body go together, lets discuss on the other comment thread.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/SimpleTypeModelBinder.cs
rynowak: In what scenario is the condition incorrect? I rewrote the ordering for clarity that this only applies when the model is a `string` (which would have been the case in `TypeMatchModelBinder`).

src/Microsoft.AspNet.Mvc.Core/ModelBinding/SimpleTypeModelBinder.cs
rynowak: `valueProviderResult.ConvertTo()` returns null if the conversion fails. You might be thinking of `ModelBindingHelper.ConvertTo<T>` which returns `default(T)`.

Would it be clearer if we renamed the latter?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/ValueProviderResultExtensions.cs
rynowak: These have different semantics. Maybe I should just delete this.

src/Microsoft.AspNet.Mvc.Core/ModelBindingHelper.cs
rynowak: Nothing, just that we expect it does. I'll add a check.

src/Microsoft.AspNet.Mvc.Core/ModelBindingHelper.cs
rynowak: Fortunately there's now no case in our code where this could happen. We use the `string`->scalar and `string[]`->scalar conversions, but not the others.

src/Microsoft.AspNet.Mvc.Core/ModelBindingHelper.cs
rynowak: I don't intend to change what this code does.

src/Microsoft.AspNet.Mvc.Core/ModelBindingHelper.cs
rynowak: I don't intend to change what this code does.

src/Microsoft.AspNet.Mvc.Core/ModelBindingHelper.cs
rynowak: I don't intend to change what this code does.

src/Microsoft.AspNet.Mvc.Core/ModelBindingHelper.cs
rynowak: More like #YOLOPasting

rynowak:
Updated with 2 small fixes.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
dougbu: True enough :ok_hand:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/BodyModelBinder.cs
dougbu: Discussed offline. Decided to restore `object` for this binder but leave the `IFormCollection` and `IFormFile` cases alone.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/FormFileModelBinder.cs
dougbu: Decided to split this and the body cases. No action needed here.

See https://github.com/aspnet/Mvc/pull/3011#discussion-diff-37593962 for more info.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/SimpleTypeModelBinder.cs
dougbu: Got a polite reminder that I was being an idiot offline. This code is :100: percent fine.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/SimpleTypeModelBinder.cs
dougbu: :ok_hand: I was overly-focused on the special case in which `ConvertTo(Type)` activates an instance.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/ValueProviderResultExtensions.cs
dougbu: Up to you but I agree this one is different (after a 3rd look).

src/Microsoft.AspNet.Mvc.Core/ModelBindingHelper.cs
dougbu: :ok_hand: I was overly-focused on the special case in which `ConvertTo(Type)` activates an instance.

src/Microsoft.AspNet.Mvc.Core/ModelBindingHelper.cs
dougbu: :ok_hand:

src/Microsoft.AspNet.Mvc.Core/ModelBindingHelper.cs
dougbu: :ok_hand: I was overly-focused on the special case in which `ConvertTo(Type)` activates an instance.

src/Microsoft.AspNet.Mvc.Core/ModelBindingHelper.cs
dougbu: :ok_hand: I was overly-focused on the special case in which `ConvertTo(Type)` activates an instance.

rynowak:
Updated again.

dougbu:
:shipit:

Issue #2941, opened 2015-08-09
Validation incorrect for dynamic collections when using `TryUpdateModelAsync()`


dougbu:
The problem described in #2294 has been fixed for controller properties and action parameters. Unfortunately that problem remains when using `TryUpdateModelAsync()`. The following form data (an example from #2294) will not be correctly validated because the `IObjectModelValidator` will not be informed that model nodes `collection[key1]` or `collection[key2]` exist.
```
Collection.index = key1,key2
Collection[key1].Name = name1
Collection[key2].Name = name2
```

Root cause: `ModelBindingHelper.TryUpdateModelAsync()` completely ignores the `ModelValidationNode` tree the model binding system returns, passing a new `ModelValidationNode` instance to the `IObjectModelValidator`. This method does not even check `ModelBindingResult.IsModelSet` before using `...Model`.

(noticed while investigating #2633)

dougbu:
02cc82a

PR #3014, opened 2015-08-21
Make validation in `TryUpdateModelAsync()` consistent with model binding elsewhere

dougbu:
- #2941
- honor `ModelBindingResult.IsModelSet` and use `ModelBindingResult.ValidationNode`
  - enable correct validation of collections or after model binding falls back to the empty prefix
  - code previously matched `Controller.TryValidateModel()`; less context available in that case

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBindingHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBindingHelper.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
@@ -313,9 +312,16 @@ namespace Microsoft.AspNet.Mvc
                 var modelValidationContext = new ModelValidationContext(modelBindingContext, modelExplorer);
 
                 var validationNode = modelBindingResult.ValidationNode;
-                Debug.Assert(
-                    validationNode != null,
-                    "ValidationNode should never be null in a successful ModelBindingResult.");
+                if (validationNode == null)
+                {
+                    validationNode = new ModelValidationNode(
+                        modelBindingResult.Key,
+                        modelMetadata,
+                        modelBindingResult.Model)
+                    {
+                        ValidateAllProperties = true,
+                    };
+                }
 
                 objectModelValidator.Validate(modelValidationContext, validationNode);
 
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBindingHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBindingHelper.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
@@ -306,16 +307,18 @@ namespace Microsoft.AspNet.Mvc
             };
 
             var modelBindingResult = await modelBinder.BindModelAsync(modelBindingContext);
-            if (modelBindingResult != null)
+            if (modelBindingResult != null && modelBindingResult.IsModelSet)
             {
                 var modelExplorer = new ModelExplorer(metadataProvider, modelMetadata, modelBindingResult.Model);
                 var modelValidationContext = new ModelValidationContext(modelBindingContext, modelExplorer);
-                objectModelValidator.Validate(
-                    modelValidationContext,
-                    new ModelValidationNode(prefix, modelBindingContext.ModelMetadata, modelBindingResult.Model)
-                    {
-                        ValidateAllProperties = true
-                    });
+
+                var validationNode = modelBindingResult.ValidationNode;
+                Debug.Assert(
+                    validationNode != null,
+                    "ValidationNode should never be null in a successful ModelBindingResult.");
+
+                objectModelValidator.Validate(modelValidationContext, validationNode);
+
                 return modelState.IsValid;
             }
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.Mvc.Core/ModelBindingHelper.cs
```diff
@@ -306,16 +307,18 @@ public static class ModelBindingHelper
...
-                    modelValidationContext,
-                    new ModelValidationNode(prefix, modelBindingContext.ModelMetadata, modelBindingResult.Model)
-                    {
-                        ValidateAllProperties = true
-                    });
+
+                var validationNode = modelBindingResult.ValidationNode;
+                Debug.Assert(
+                    validationNode != null,
+                    "ValidationNode should never be null in a successful ModelBindingResult.");
```
rynowak: We probably shouldn't have an assert here and just check for null instead. The model binder might be implemented by a user. If we want this guarantee, we need to implement it at another level.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBindingHelperTest.cs
```diff
@@ -17,15 +17,29 @@ namespace Microsoft.AspNet.Mvc.Test
...
         {
             // Arrange
             var metadataProvider = new EmptyModelMetadataProvider();
-
             var binder = new Mock<IModelBinder>();
-            binder.Setup(b => b.BindModelAsync(It.IsAny<ModelBindingContext>()))
-                  .Returns(Task.FromResult<ModelBindingResult>(null));
+            binder
+                .Setup(b => b.BindModelAsync(It.IsAny<ModelBindingContext>()))
+                .Returns(Task.FromResult<ModelBindingResult>(binderResult));
```
rynowak: Should we also verify the validation state here? I guess there wouldn't really be any entries in modelstate to verify would there?

src/Microsoft.AspNet.Mvc.Core/ModelBindingHelper.cs
dougbu: Hmm, what level? `CompositeModelBinder`, `MutableObjectModelBinder` and `GenericModelBinder` wrap pretty much every other binder and they all return with a non-`null` `ValidationNode` if successful.

rynowak: You can pass in a model binder directly to `TryUpdateModel` so it seems a little more likely than usual :laughing:. Normally I'm not paranoid about these kinds of things, but this seems very possible.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBindingHelperTest.cs
dougbu: Without changing `// Arrange`, could `Assert.Empty(modelState)`. Are you suggesting instead that this mock binder should add an error and then do the `Assert.Single(modelState)` thing?

src/Microsoft.AspNet.Mvc.Core/ModelBindingHelper.cs
dougbu: :ok_hand:

test/Microsoft.AspNet.Mvc.Core.Test/ModelBindingHelperTest.cs
rynowak: I'm trying to think of a way we could verify that validation never happened, but without doing the dumb thing of injecting a mock validator that throws.

test/Microsoft.AspNet.Mvc.IntegrationTests/TryUpdateModelIntegrationTest.cs
```diff
@@ -103,6 +103,63 @@ private class Person1
...
+            };
+
+            // Act
+            var result = await TryUpdateModel(model, string.Empty, operationContext, modelState);
+
+            // Assert
+            Assert.True(result);
+
+            // Model
+            Assert.Collection(model,
```
rynowak: #YOLO

test/Microsoft.AspNet.Mvc.IntegrationTests/TryUpdateModelIntegrationTest.cs
```diff
@@ -712,7 +768,6 @@ private class Person5
             Assert.Equal(ModelValidationState.Valid, state.ValidationState);
         }
 
-        [Fact(Skip = "Validation incorrect for collections when using TryUpdateModel, #2941")]
```
rynowak: :heart_eyes:

test/Microsoft.AspNet.Mvc.Core.Test/ModelBindingHelperTest.cs
dougbu: The mock doesn't have to `throw`. Could just `Verify()` that `Validate()` is called `Times.Zero`.

rynowak: If you can think of a simple way to accomplish this, go for it. Otherwise, unimportant.

dougbu:
Updated...

test/Microsoft.AspNet.Mvc.Core.Test/ModelBindingHelperTest.cs
dougbu: Fixed w/ `MockBehavior.Strict`.

rynowak:
:shipit:

dougbu:
02cc82a

PR #3015, opened 2015-08-21
Workaround https://github.com/aspnet/External/issues/21 in a couple more tests

dougbu:
- led to Travis build failures


dougbu:
FYI only. Just a bit more to clean up once aspnet/External#21 is fixed or we work around it.

Will push once Travis succeeds.

dougbu:
aaa47d5

Issue #194, opened 2014-04-03
Make IValueProvider.GetValue async


kichalla:
`IValueProviderFactory.GetValueProviderAsync` is async but `IValueProvider.GetValue` is not. I think this should be async and the one creating the valueproviderfactory need not.

Scenario: 
A valuer provider which gets values from database or webservice call.

PR #3017, opened 2015-08-21
Make IValueProvider sync, IValueProviderFactory async

rynowak:
/cc @dougbu

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/ByteArrayModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/ByteArrayModelBinder.cs
@@ -13,19 +13,22 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
     public class ByteArrayModelBinder : IModelBinder
     {
         /// <inheritdoc />
-        public async Task<ModelBindingResult> BindModelAsync([NotNull] ModelBindingContext bindingContext)
+        public Task<ModelBindingResult> BindModelAsync([NotNull] ModelBindingContext bindingContext)
         {
             // Check if this binder applies.
             if (bindingContext.ModelType != typeof(byte[]))
             {
-                return null;
+                return Task.FromResult<ModelBindingResult>(null);
             }
 
             // Check for missing data case 1: There was no <input ... /> element containing this data.
-            var valueProviderResult = await bindingContext.ValueProvider.GetValueAsync(bindingContext.ModelName);
+            var valueProviderResult = bindingContext.ValueProvider.GetValue(bindingContext.ModelName);
             if (valueProviderResult == ValueProviderResult.None)
             {
-                return new ModelBindingResult(model: null, key: bindingContext.ModelName, isModelSet: false);
+                return Task.FromResult(new ModelBindingResult(
+                    model: null,
+                    key: bindingContext.ModelName,
+                    isModelSet: false));
             }
 
             bindingContext.ModelState.SetModelValue(bindingContext.ModelName, valueProviderResult);
@@ -34,7 +37,10 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             var value = valueProviderResult.FirstValue;
             if (string.IsNullOrEmpty(value))
             {
-                return new ModelBindingResult(model: null, key: bindingContext.ModelName, isModelSet: false);
+                return Task.FromResult(new ModelBindingResult(
+                    model: null,
+                    key: bindingContext.ModelName,
+                    isModelSet: false));
             }
 
             try
@@ -45,11 +51,11 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
                     bindingContext.ModelMetadata,
                     model);
 
-                return new ModelBindingResult(
+                return Task.FromResult(new ModelBindingResult(
                     model,
                     bindingContext.ModelName,
                     isModelSet: true,
-                    validationNode: validationNode);
+                    validationNode: validationNode));
             }
             catch (Exception ex)
             {
@@ -58,7 +64,10 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
 
             // Matched the type (byte[]) only this binder supports. As in missing data cases, always tell the model
             // binding system to skip other model binders i.e. return non-null.
-            return new ModelBindingResult(model: null, key: bindingContext.ModelName, isModelSet: false);
+            return Task.FromResult(new ModelBindingResult(
+                model: null,
+                key: bindingContext.ModelName,
+                isModelSet: false));
         }
     }
 }
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/ByteArrayModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/ByteArrayModelBinder.cs
@@ -15,20 +15,22 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         /// <inheritdoc />
         public Task<ModelBindingResult> BindModelAsync([NotNull] ModelBindingContext bindingContext)
         {
+            return Task.FromResult(BindModel(bindingContext));
+        }
+
+        private ModelBindingResult BindModel(ModelBindingContext bindingContext)
+        {
             // Check if this binder applies.
             if (bindingContext.ModelType != typeof(byte[]))
             {
-                return Task.FromResult<ModelBindingResult>(null);
+                return null;
             }
 
             // Check for missing data case 1: There was no <input ... /> element containing this data.
             var valueProviderResult = bindingContext.ValueProvider.GetValue(bindingContext.ModelName);
             if (valueProviderResult == ValueProviderResult.None)
             {
-                return Task.FromResult(new ModelBindingResult(
-                    model: null,
-                    key: bindingContext.ModelName,
-                    isModelSet: false));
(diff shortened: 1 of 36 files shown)


src/Microsoft.AspNet.Mvc.Core/ModelBinding/ByteArrayModelBinder.cs
```diff
@@ -13,19 +13,22 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
...
             // Check for missing data case 1: There was no <input ... /> element containing this data.
-            var valueProviderResult = await bindingContext.ValueProvider.GetValueAsync(bindingContext.ModelName);
+            var valueProviderResult = bindingContext.ValueProvider.GetValue(bindingContext.ModelName);
             if (valueProviderResult == ValueProviderResult.None)
             {
-                return new ModelBindingResult(model: null, key: bindingContext.ModelName, isModelSet: false);
+                return Task.FromResult(new ModelBindingResult(
+                    model: null,
+                    key: bindingContext.ModelName,
+                    isModelSet: false));
```
dougbu: Wondering about avoiding repeated `Task.FromResult(...)` in binders that run synchronously. The following might be cleaner...
``` c#
public Task<ModelBindingResult> BindModelAsync([NotNull] ModelBindingContext bindingContext)
{
    return Task.FromResult(BindModel(bindingContext));
}

private ModelBindingResult BindModel(ModelBindingContext bindingContext)
{
    // the old BindModelAsync() content, less some awaiting
}
```

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/ElementalValueProviderTests.cs
```diff
@@ -14,7 +14,7 @@ public class ElementalValueProviderTest
         [InlineData("MyProperty", "MyProperty")]
         [InlineData("MyProperty.SubProperty", "MyProperty")]
         [InlineData("MyProperty[0]", "MyProperty")]
-        public async Task ContainsPrefixAsync_ReturnsTrue_IfElementNameStartsWithPrefix(string elementName, 
+        public void ContainsPrefix_ReturnsTrue_IfElementNameStartsWithPrefix(string elementName, 
                                                                                         string prefix)
```
dougbu: Shortened method definition messing up the old #YOLO wrapping. Not horrendous of course but please take a quick cleanup pass through the updated code.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/EnumerableValueProviderTest.cs
```diff
@@ -29,61 +29,62 @@ public abstract class EnumerableValueProviderTest
...
         }
 
-        [Fact]
-        public async Task ContainsPrefixAsync_WithNonEmptyCollection_ReturnsTrueForKnownPrefixes()
+        [Theory]
+        [InlineData("some")]
+        [InlineData("prefix")]
+        [InlineData("prefix.name")]
+        [InlineData("[index]")]
+        [InlineData("prefix[index1]")]
```
dougbu: :dancers:

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/FormValueProviderFactoryTest.cs
```diff
@@ -49,6 +50,11 @@ private static ValueProviderFactoryContext CreateContext(string contentType)
             var context = new DefaultHttpContext();
             context.Request.ContentType = contentType;
 
+            if (context.Request.HasFormContentType)
+            {
+                context.Request.Form = new FormCollection(new Dictionary<string, string[]>());
+            }
```
dougbu: Why is this only now needed?

test/WebSites/ModelBindingWebSite/Controllers/ModelBinderAttribute_ProductController.cs
```diff
@@ -97,16 +97,16 @@ private class ProductModelBinder : IModelBinder
                         string.IsNullOrEmpty(bindingContext.ModelName) ?
                         "productId" :
                         bindingContext.ModelName + "." + "productId";
-
-                    var value = await bindingContext.ValueProvider.GetValueAsync(key);
+
```
dougbu: Someone needs the trailing whitespace visualizer.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/ByteArrayModelBinder.cs
dougbu: Looks as if this would be most helpful here and in `SimpleTypeModelBinder`.

dougbu:
:watch: for a couple of questions and slight cleanup

src/Microsoft.AspNet.Mvc.Core/ModelBinding/ByteArrayModelBinder.cs
rynowak: Yep, good suggestion, made the change.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/ElementalValueProviderTests.cs
rynowak: Yep. Gotta either #YOLO everyday or not at all.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/FormValueProviderFactoryTest.cs
rynowak: The Form will be read when the VP is created now instead of when the VP is used.

test/WebSites/ModelBindingWebSite/Controllers/ModelBinderAttribute_ProductController.cs
rynowak: Not part of our coding standards...

rynowak:
@dougbu  updated

dougbu:
:shipit:

csunali:
Is there a specific reason on making the GetValue method sync again? I remember someone previously deciding to make it async intentionally. I might be missing something.

csunali:
Found the issue at https://github.com/aspnet/Mvc/issues/194

rynowak:
The change is based on a better understanding of what we think value-providers should be used for - generally a simple collection of mappings from string->string[].

If your data is anything other than that, or requires more complicated machinations to enumerate it/access it, then it's not a good fit for how value-providers are used in the system.

csunali:
I understand. What about other types? For instance, I usually bind int, long and bool using value providers as well - but not others. Is that okay to use it for these kind of primitives too?

Thank you for the details.

csunali:
By the way, `IModelBinder.BindModelAsync` method is still async. That is where one uses the `ValueProvider.GetValue` method.

That makes the method async but there are not async methods within the async method. Thus the `DNX Core 5.0 warning CS1998: This async method lacks 'await' operators and will run synchronously. Consider using the 'await' operator to await non-blocking API calls, or 'await Task.Run(...)' to do CPU-bound work on a background thread.` warning.

How would you handle that?

```
    public class SomethingModelBinder : IModelBinder
    {
        public async Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindingContext)
        {
            SomethingModel model = new SomethingModel();

            model.X = bindingContext.ValueProvider.GetValue("X");
            model.Y = bindingContext.ValueProvider.GetValue("Y");
            
            bindingContext.Model = model;

            return new ModelBindingResult(model, bindingContext.ModelName, true);
        }
    }
```

rynowak:
> For instance, I usually bind int, long and bool using value providers as well - but not others. Is that okay to use it for these kind of primitives too

Where do these values come from? Everything coming over the wire in HTTP is a string :laughing:. Generally data from a value provider is applicable **everywhere** in the model binding system, if you're handling specific model types or properties, a model binder is a better fit.

> How would you handle that?

You don't need to go async at all in this code, I'd handle it like this:
```
    public class SomethingModelBinder : IModelBinder
    {
        public Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindingContext)
        {
            SomethingModel model = new SomethingModel();

            model.X = bindingContext.ValueProvider.GetValue("X");
            model.Y = bindingContext.ValueProvider.GetValue("Y");

            bindingContext.Model = model;

            return Task.FromResult(new ModelBindingResult(model, bindingContext.ModelName, true));
        }
    }
```

csunali:
Oh, that is what you mean by a `string`, indeed, everything coming via the HTTP pipeline is a string :)

Thank you for the `Task.FromResult` hint!

Cheers!

csunali:
The `Task.FromResult` method seems to return `Task<T>` rather than `T` which results with:

`Error	CS4016	Since this is an async method, the return expression must be of type 'ModelBindingResult' rather than 'Task<ModelBindingResult>'`

I tried with `.Result` however that returns back to the previous result, which throws the warning stating that the method is async but the `await` lacks.

Do you have any other suggestions?

rynowak:
You need to remove `async` from the method definition. http://blogs.msdn.com/b/pfxteam/archive/2012/04/12/10293335.aspx

csunali:
That really helped. I will read through the article you shared.

Thank you again!

PR #3019, opened 2015-08-21
[Fix #2900] Get rid of manual body-read-state-tracking

### LINKED ISSUE #2900: Get rid of manual body-read-state-tracking

kichalla:
@rynowak


rynowak:
:shipit: **with one caveat**

Can you please check one thing! Can we remove `UseErrorReporter()` from the model binding website?

dougbu:
Why doesn't this PR kill off `BodyBindingState` entirely?

kichalla:
e12d44b40adfbc063cdf92683388b7d3919743a9

PR #3021, opened 2015-08-22
Fix #2527 - Remove FormCollection use

### LINKED ISSUE #2527: Form Collection Model Binder should not fall back to copying the form values from FormCollection
This should get us one step closer to removing MVC's dependency on the concrete implementations of the HTTP abstractions. When fixing this bug, need to check if we can switch to depend just on the HTTP `.Abstractions` package.

rynowak:
This removes the dependency on Microsoft.AspNet.Http from the Mvc.Core
code. Added the reference back to tests where needed (DefaultHttpContext).

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/FormCollectionModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/FormCollectionModelBinder.cs
@@ -1,12 +1,12 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
-using System.Reflection;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Http;
-using Microsoft.AspNet.Http.Internal;
 using Microsoft.Framework.Internal;
 
 namespace Microsoft.AspNet.Mvc.ModelBinding
@@ -19,8 +19,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         /// <inheritdoc />
         public async Task<ModelBindingResult> BindModelAsync([NotNull] ModelBindingContext bindingContext)
         {
-            if (bindingContext.ModelType != typeof(IFormCollection) &&
-                bindingContext.ModelType != typeof(FormCollection))
+            if (bindingContext.ModelType != typeof(IFormCollection))
             {
                 return null;
             }
@@ -30,19 +29,11 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             if (request.HasFormContentType)
             {
                 var form = await request.ReadFormAsync();
-                if (bindingContext.ModelType.GetTypeInfo().IsAssignableFrom(form.GetType().GetTypeInfo()))
-                {
-                    model = form;
-                }
-                else
-                {
-                    var formValuesLookup = form.ToDictionary(p => p.Key, p => p.Value);
-                    model = new FormCollection(formValuesLookup, form.Files);
-                }
+                model = form;
             }
             else
             {
-                model = new FormCollection(new Dictionary<string, string[]>());
+                model = new EmptyFormCollection();
             }
 
             var validationNode =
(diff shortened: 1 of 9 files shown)


rynowak:
VS loves to format stuff :laughing:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/FormCollectionModelBinder.cs
```diff
@@ -30,19 +29,11 @@ public class FormCollectionModelBinder : IModelBinder
             if (request.HasFormContentType)
             {
                 var form = await request.ReadFormAsync();
-                if (bindingContext.ModelType.GetTypeInfo().IsAssignableFrom(form.GetType().GetTypeInfo()))
```
dougbu: Curious: Why not do this instead of the type equality check at [line 22](https://github.com/aspnet/Mvc/pull/3021/files#diff-9ab5b561d10929ae1a601c3f090a8e03R22)?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/FormCollectionModelBinder.cs
```diff
@@ -30,19 +29,11 @@ public class FormCollectionModelBinder : IModelBinder
...
-                    model = form;
-                }
-                else
-                {
-                    var formValuesLookup = form.ToDictionary(p => p.Key, p => p.Value);
-                    model = new FormCollection(formValuesLookup, form.Files);
-                }
+                model = form;
             }
             else
```
dougbu: .. then `else if (bindingContext.ModelType.GetTypeInfo().IsAssignableFrom(typeof(EmptyFormCollection).GetTypeInfo()))` here?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/FormCollectionModelBinder.cs
dougbu: Suggest adding a bit to skip the `GetTypeInfo()` verbiage:
``` c#
#if DNXCORE50
using System.Reflection;
#endif
```

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/FormCollectionModelBinderTest.cs
```diff
@@ -77,8 +77,8 @@ public class FormCollectionModelBinderTest
 
             // Assert
             Assert.NotNull(result);
-            Assert.IsType(typeof(FormCollection), result.Model);
-            Assert.Empty((FormCollection)result.Model);
+            Assert.IsType(typeof(IFormCollection), result.Model);
+            Assert.Empty((IFormCollection)result.Model);
```
dougbu:
``` c#
var collection = Assert.IsAssignableFrom<IFormCollection>(result.Model);
Assert.Empty(collection);
```

Not sure how test would pass as-is since `IsType()` verifies the exact type "(and not a derived type)".

dougbu:
Looks as if a few more tests need to change.

dougbu:
:watch:

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/FormCollectionModelBinderTest.cs
rynowak: I don't run the tests, I just visual inspect them.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/FormCollectionModelBinder.cs
rynowak: We specifically only support using `IFormCollection` here. `FormCollection` is an 'internal' type and we don't encourage you to use it directly outside of tests.

I'd rather not build more features into this (activation of your custom type). There are far fewer useful things to do with custom types here than in other places in model binding.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/FormCollectionModelBinder.cs
rynowak: Nope, we only want to support `IFormCollection`.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/FormCollectionModelBinder.cs
dougbu: :ok:

I wasn't suggesting any activation at all, just extending the compatibilites. But I understand we don't want to support anything derived from `IFormCollection`, even if the type is compatible with the instance `await request.ReadFormAsync()` returns. In any case my approach ends up looking more cumbersome:
``` c#
    object model;
    var request = bindingContext.OperationBindingContext.HttpContext.Request;
    if (request.HasFormContentType)
    {
        var form = await request.ReadFormAsync();
        if (!bindingContext.ModelType.IsAssignableFrom(form.GetType())
        {
            return null;
        }

        model = form;
    }
    else
    {
        if (!bindingContext.ModelType.IsAssignableFrom(typeof(EmptyFormCollection))
        {
            return null;
        }

        model = new EmptyFormCollection();
    }
```

src/Microsoft.AspNet.Mvc.Core/ModelBinding/FormCollectionModelBinder.cs
dougbu: :ok_hand:

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/FormCollectionModelBinderTest.cs
dougbu: Me too; this looked wrong :smiling_imp:

rynowak:
@dougbu updated

dougbu:
:shipit:

Issue #2992, opened 2015-08-18
Hoist uses of `Nullable.GetUnderlyingType()` into `ModelMetadata`


dougbu:
MVC code contains too many instances of low-level code duplicating
``` c#
!type.GetTypeInfo().IsValueType || Nullable.GetUnderlyingType(type) != null
```
and even more duplicates of
``` c#
Nullable.GetUnderlyingType(type) ?? type
```

The first could be exposed as `ModelMetadata.AllowsNullValue`. Alternatively could provide `ModelMetadata.IsValueType` in addition to the existing `ModelMetadata.IsNullableValueType`, allowing the above to be elevated to
``` c#
!metadata.IsValueType || metadata.IsNullableValueType
```

The second could be exposed as `ModelMetadata.UnderlyingType`.

rynowak:
This, is code cleanup, just do it IMO :+1:

dougbu:
Yeah, don't disagree. Self-assigning this "task".

dougbu:
Decided to add
- `IsReferenceOrNullableType`, just to avoid a `bool` property that isn't named `IsFoo`
- `UnderlyingOrModelType`

dougbu:
bf7e0f1

PR #3023, opened 2015-08-22
Add `IsReferenceOrNullableType` and `UnderlyingOrModelType` to `ModelMetadata`

dougbu:
- #2992
- use new properties to replace common helper methods
  - largest change was to `DefaultControllerActionArgumentBinder`; previously used `PropertyHelper`
- still a few `Nullable.GetUnderlyingType()` calls
  - creating `ModelMetadata` or sites lacking `ModelMetadata` access e.g. `ModelBindingHelper.ConvertTo()`

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelMetadata.cs
+++ b/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelMetadata.cs
@@ -5,9 +5,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.ComponentModel;
-#if DNXCORE50
 using System.Reflection;
-#endif
 using Microsoft.AspNet.Mvc.ModelBinding.Metadata;
 using Microsoft.Framework.Internal;
 
@@ -139,8 +137,8 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         public abstract ModelMetadata ElementMetadata { get; }
 
         /// <summary>
-        /// Gets the ordered display names and values of all <see cref="Enum"/> values in <see cref="ModelType"/> or
-        /// <c>Nullable.GetUnderlyingType(ModelType)</c>.
+        /// Gets the ordered display names and values of all <see cref="Enum"/> values in
+        /// <see cref="UnderlyingOrModelType"/>.
         /// </summary>
         /// <value>
         /// An <see cref="IEnumerable{KeyValuePair{string, string}}"/> of mappings between <see cref="Enum"/> field names
@@ -149,8 +147,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         public abstract IEnumerable<KeyValuePair<string, string>> EnumDisplayNamesAndValues { get; }
 
         /// <summary>
-        /// Gets the names and values of all <see cref="Enum"/> values in <see cref="ModelType"/> or
-        /// <c>Nullable.GetUnderlyingType(ModelType)</c>.
+        /// Gets the names and values of all <see cref="Enum"/> values in <see cref="UnderlyingOrModelType"/>.
         /// </summary>
         /// <value>
         /// An <see cref="IReadOnlyDictionary{string, string}"/> of mappings between <see cref="Enum"/> field names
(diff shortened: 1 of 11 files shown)


src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
```diff
@@ -118,19 +117,18 @@ private static void CallPropertyAddRange<TElement>(object target, object source)
 
         private void ActivateProperties(object controller, Type containerType, Dictionary<string, object> properties)
         {
-            var propertyHelpers = PropertyHelper.GetProperties(controller);
+            var controllerMetadata = _modelMetadataProvider.GetMetadataForType(containerType);
```
dougbu: Note the two collections are slightly different. `propertyHelpers` contained properties hidden via `new` while `controllerMetadata` does not. However `.First()` in old code elides the difference.

test/Microsoft.AspNet.Mvc.Abstractions.Test/ModelBinding/ModelMetadataTest.cs
```diff
@@ -111,18 +115,42 @@ public void IsNullableValueTypeTests(Type modelType, bool expected)
             Assert.Equal(expected, modelMetadata.IsNullableValueType);
         }
 
-        private class Class1
```
dougbu: `Class1` and `Class2` were unused.

test/Microsoft.AspNet.Mvc.Abstractions.Test/ModelBinding/ModelMetadataTest.cs
```diff
@@ -50,12 +50,21 @@ public void IsComplexTypeTestsReturnsTrueForComplexTypes(Type type)
...
 
+        // IsCollectionType
+        private class NonCollectionType
+        {
+        }
+
+        private class DerivedList : List<int>
+        {
+        }
+
```
dougbu: Moved up to be consistent with other test subjects in this class.

test/Microsoft.AspNet.Mvc.Abstractions.Test/ModelBinding/ModelMetadataTest.cs
```diff
@@ -21,7 +21,7 @@ private struct IsComplexTypeModel
         [InlineData(typeof(string))]
         [InlineData(typeof(Nullable<int>))]
         [InlineData(typeof(int))]
-        public void IsComplexTypeTestsReturnsFalseForSimpleTypes(Type type)
+        public void IsComplexType_ReturnsFalseForSimpleTypes(Type type)
```
dougbu: Updated all test names in this class.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelMetadata.cs
```diff
@@ -139,8 +137,8 @@ public string PropertyName
         public abstract ModelMetadata ElementMetadata { get; }
 
         /// <summary>
-        /// Gets the ordered display names and values of all <see cref="Enum"/> values in <see cref="ModelType"/> or
-        /// <c>Nullable.GetUnderlyingType(ModelType)</c>.
+        /// Gets the ordered display names and values of all <see cref="Enum"/> values in
+        /// <see cref="UnderlyingOrModelType"/>.
```
rynowak: :dancers:

src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
rynowak: This is a bug then. If you hide a property via `new` - that doesn't make it not needed by features of the base class.

rynowak: I really don't think we should be trying to create a modelmetadata for a controller.

rynowak:
Conditional :shipit: - I'd prefer to revert the changes to the `ControllerPropertyActivator`. Other than that everything looks great! Let's discuss Monday if necessary.

src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
dougbu: > If you hide a property via `new` - that doesn't make it not needed by features of the base class.

Are you suggesting this code should assume the top-level property does not also initialize the hidden one? Sounds like a lot of complication, especially given issues when `base.Property` would be property initialized as-is.

Note this behaviour hasn't changed with this PR. The `properties` collection was a property name -> value mapping all along.

> I really don't think we should be trying to create a modelmetadata for a controller.

I could use the `GetMetadataForProperty()` extension method to hide the controller metadata.

Or are you suggesting we stick with `PropertyHelper` here? Why exactly?

Note `BindModelAsync()` already gets the `ModelMetadata` for the property types. But the getter and setter properties are `null` there.

rynowak: > Are you suggesting this code should assume the top-level property does not also initialize the hidden one? Sounds like a lot of complication, especially given issues when base.Property would be property initialized as-is.

Yes. It's unlikely anyone would hit this bug because activation only really applies to `ActionContext` and `ActionBindingContext`.

rynowak: Note that you can still do some of this cleanup by doing `GetMetadataForType(...)`

dougbu: I'll revert a bunch of this. But I'm still curious about the side discussion around `new`...

> It should be totally viable to use `new` in a controller to hide services used by the `Controller` class itself.

Services? A user can derive their controller classes from whatever they like and their base could use other model binders for properties.

But my push back is the same if focused on `Controller` in particular. In the following code, bad things could happen if `IMyMetadataProvider` and `IModelMetadataProvider` are different types, perhaps because only this controller needs or wants the extra `IMyMetadataProvider` behaviours.
``` c#
    public class MyController : Controller
    {
        [FromServices]
        public new IMyMetadataProvider MetadataProvider
        {
            get { return (IMyMetadataProvider)base.MetadataProvider; }
            set { base.MetadataProvider = value; }
        }
    }
```

Should I open an issue for further discussion of this point?

rynowak: Yeah, let's move this discussion to a separate bug.

dougbu: New issue is #3028

src/Microsoft.AspNet.Mvc.Core/DefaultControllerActionArgumentBinder.cs
```diff
@@ -124,11 +124,12 @@ private void ActivateProperties(object controller, Type containerType, Dictionar
                 var propertyHelper = propertyHelpers.First(helper =>
                     string.Equals(helper.Name, property.Key, StringComparison.Ordinal));
                 var propertyType = propertyHelper.Property.PropertyType;
+                var metadata = _modelMetadataProvider.GetMetadataForType(propertyType);
```
dougbu: I'll move this inside the `if` where it's used.

dougbu:
@rynowak updated

rynowak:
:shipit:

dougbu:
bf7e0f1

Issue #2905, opened 2015-08-03
Change in-box tag helpers to use `DefaultFrameworkSortOrder`


dougbu:
Existing tag helpers all have `Order == 0`. This is the user default and ours should lie in the documented range (-1 to -1999). Suggest `DefaultOrder.DefaultFrameworkSortOrder`.

dougbu:
8babf2b

PR #3024, opened 2015-08-22
Change in-box tag helpers to use `DefaultFrameworkSortOrder`

dougbu:
- #2905
- override `Order` implementation inherited from `TagHelper`
- only exception is `UrlResolutionTagHelper` which already overrides `Order` to execute much earlier


NTaylorMullen:
:shipit:

dougbu:
8babf2b

Issue #1888, opened 2015-01-26
Provide predictable ordering of validation summary messages and `@Html.Display()` / `@Html.Editor()` properties


dougbu:
to ensure a predictable order using MVC once #964 is fixed, users must annotate all model properties with `[Display(Order=x)]`.  this may be too high a bar but unfortunately `type.GetRuntimeProperties()` and related Reflection methods do not guarantee a predictable or even a stable ordering.

recent decision was to split this issue from #964 and to determine what guarantees exist for the `type.GetRuntimeProperties()` return value.  then, if no guarantees exist (as [MSDN](https://msdn.microsoft.com/en-us/library/kyaxdd3x.aspx) implies), ask for one or add something (perhaps not based on `PropertyName`) to MVC.

__details and history__
see also the StackOverflow [discussion](http://stackoverflow.com/questions/5473455/c-sharp-get-fieldinfos-propertyinfos-in-the-original-order/5473526#5473526) of Reflection ordering as well as the linked [MSDN documentation](https://msdn.microsoft.com/en-us/library/kyaxdd3x.aspx).

in #964:
> MVC 5.2 does not provide a predictable property order when displaying / editing a complex object or filling in the `@Html.ValidationSummary()`. We could do a bit better here and sort by `Order` then by `PropertyName`.

in PR #1866:
> I'm not sure I quite follow the reasoning for adding ordering by Name, is it the same behavior as MVC5? Can you illustrate the benefit?

--
> we had an offline discussion (sorry, not sure who all was there) where we agreed to make the property order predictable. today, it's possible for the HTML `Html.DisplayFor(...)` (for example) generates to change between CLR or Roslyn builds. with this fix in place, the UI will remain consistent no matter how the runtime and compiler choose to order properties.
> MVC 5 does not provide such a predictable property order. the offline discussion basically supported my comment in #964:
>> We could do a bit better here and sort by `Order` then by `PropertyName`.

yishaigalatzer:
Please work with the Roslyn and CLR folks to verify if:

1. Is the order consistent if you use the same compiler and same code.
2. Is the order the definition order if you have a simple POCO.
3. Can the order change at runtime

dougbu:
Too big a surprise for those used to `@Html.DisplayFor()` and so on spitting out fields in the order they placed properties in the original class.

PR #3025, opened 2015-08-23
Order `ModelMetadata.Properties` by `Order` then by `PropertyName`

dougbu:
- #1888
- add `PropertyName` to make order not only stable but predictable
  - slight breaking change from MVC 5 which uses only `Order` and what `System.Reflection` provides

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
@@ -442,7 +442,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
                 if (_properties == null)
                 {
                     var properties = _provider.GetMetadataForProperties(ModelType);
-                    properties = properties.OrderBy(p => p.Order);
+                    properties = properties.OrderBy(p => p.Order).ThenBy(p => p.PropertyName);
                     _properties = new ModelPropertyCollection(properties);
                 }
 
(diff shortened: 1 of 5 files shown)


rynowak:
Is there a good reason why we didn't do this change before?

dougbu:
> why we decided to hesitate on this in the past

Believe some thought we'd get assurance from the CLR and Roslyn teams. Nothing like that has ever happened.

The current behaviour is basically IL-ordering which _tends_ to match ordering in the source files.

rynowak:
:shipit:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
```diff
@@ -442,7 +442,7 @@ public override ModelPropertyCollection Properties
                 if (_properties == null)
                 {
                     var properties = _provider.GetMetadataForProperties(ModelType);
-                    properties = properties.OrderBy(p => p.Order);
+                    properties = properties.OrderBy(p => p.Order).ThenBy(p => p.PropertyName);
```
pranavkm: `StringComparison.Ordinal`?

pranavkm:
:shipit:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
dougbu: Agree this should have a specified comparer. Lean toward `Ordinal` as well though the order of this collection mostly affects generated HTML, where case is irrelevant.

Any votes for `OrdinalIgnoreCase`?

NeelBhatt: OrdinalIgnoreCase is good for exact comparisons but not for sorting strings with unusual characters. This is also faster because there are no extra rules to apply before comparing.

rynowak: > Lean toward Ordinal as well though the order of this collection mostly affects generated HTML, where case is irrelevant.

This is actually the use case for `InvariantCulture` :trollface:

dougbu:
Decided not to fix #1888 offline.

Issue #2993, opened 2015-08-18
Collections with different type parameters than expected interface do not bind correctly


dougbu:
The `GenericModelBinder` does not support properties that have a different number of type parameters than the expected interfaces. This occurs whether or not the property type implements that interface.

For example properties of types like the following cannot currently be bound as collections:
``` c#
public class ClosedGenericDictionary : Dictionary<string, string>
{
    // ...
}

public class ClosedGenericList : List<string>
{
    // ...
}

public class ClosedGenericKeyDictionary<TValue> : Dictionary<int, TValue>
{
    // ...
}
```

This is a follow-on issue for #2793. That was about custom collections. This extends the notion of "custom" to the exposed type parameters.

Note a fix for this issue will likely also handle oddball cases such as
``` c#
public class OddballOpenGenericList<int> : List<string>
{
    // ...
}
```

danroth27:
@dougbu Is this a regression from MVC 5?

dougbu:
@danroth27 will test...

dougbu:
@danroth27 yes, this is a regression from MVC 5.

dougbu:
829a5c9

PR #3026, opened 2015-08-23
Expand model types `GenericModelBinder` can handle

dougbu:
- #2993
- use `ClosedGenericMatcher` to handle e.g. non-generic model types implementing requested interfaces
- reduce `IsAssignableFrom()` use since created binders use explicit casts i.e. handle explicit implementations

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
@@ -4,8 +4,11 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+#if DNXCORE50
 using System.Reflection;
+#endif
 using System.Threading.Tasks;
+using Microsoft.Framework.Internal;
 
 namespace Microsoft.AspNet.Mvc.ModelBinding
 {
@@ -45,8 +48,8 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             var modelType = context.ModelType;
 
             return GetArrayBinder(modelType) ??
-                GetCollectionBinder(modelType) ??
                 GetDictionaryBinder(modelType) ??
+                GetCollectionBinder(modelType) ??
                 GetEnumerableBinder(context) ??
                 GetKeyValuePairBinder(modelType);
         }
@@ -58,6 +61,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
                 var elementType = modelType.GetElementType();
                 return typeof(ArrayModelBinder<>).MakeGenericType(elementType);
             }
+
             return null;
         }
 
@@ -99,9 +103,12 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             else
             {
                 // A non-null instance must be updated in-place. For that the instance must also implement
-                // ICollection<T>. For example an IEnumerable<T> property may have a List<T> default value.
-                var closedCollectionType = typeof(ICollection<>).MakeGenericType(modelTypeArguments);
-                if (!closedCollectionType.IsAssignableFrom(context.Model.GetType()))
+                // ICollection<T>. For example an IEnumerable<T> property may have a List<T> default value. Do not use
+                // IsAssignableFrom() because that does not handle explicit interface implementations and binders all
+                // perform explicit casts.
+                var closedGenericInterface =
+                    ClosedGenericMatcher.ExtractGenericInterface(context.Model.GetType(), typeof(ICollection<>));
+                if (closedGenericInterface == null)
                 {
                     return null;
                 }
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
```diff
@@ -45,8 +48,8 @@ private static Type ResolveBinderType(ModelBindingContext context)
             var modelType = context.ModelType;
 
             return GetArrayBinder(modelType) ??
-                GetCollectionBinder(modelType) ??
                 GetDictionaryBinder(modelType) ??
+                GetCollectionBinder(modelType) ??
```
dougbu: Order now important because a dictionary is a collection.

test/Microsoft.AspNet.Mvc.IntegrationTests/DictionaryModelBinderIntegrationTest.cs
```diff
@@ -134,6 +137,7 @@ public class DictionaryModelBinderIntegrationTest
         [Theory]
         [InlineData("?[key0]=10")]
         [InlineData("?[0].Key=key0&[0].Value=10")]
+        [InlineData("?index=low&[low].Key=key0&[low].Value=10")]
```
dougbu: Not directly related to fix but noticed these cases weren't covered.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
```diff
@@ -144,28 +153,10 @@ private static Type[] GetGenericBinderTypeArgs(Type supportedInterfaceType, Type
...
-            if (!closedInstanceType.IsAssignableFrom(modelType))
-            {
-                // modelType is not compatible with supportedInterfaceType.
-                return null;
-            }
+            var closedGenericInterface =
+                ClosedGenericMatcher.ExtractGenericInterface(modelType, supportedInterfaceType);
 
-            return modelTypeArguments;
+            return closedGenericInterface?.GenericTypeArguments;
```
rynowak: Seems like this change de-duplicates a lot of existing logic :+1:

test/Microsoft.AspNet.Mvc.IntegrationTests/DictionaryModelBinderIntegrationTest.cs
rynowak: Good catch

rynowak: Do we need to add integration tests for these cases? Are there test cases using this key-syntax that trigger validation errors?

rynowak:
:shipit:

test/Microsoft.AspNet.Mvc.IntegrationTests/DictionaryModelBinderIntegrationTest.cs
dougbu: This is an integration test.

Because the explicit index syntax is usable in scenarios using a POCO `TValue`, could test with a few data annotation attributes in place. That would ensure validation works correctly in this case. Is that what you're looking for?

rynowak: > That would ensure validation works correctly in this case. Is that what you're looking for?

Yeah, I want to make sure that we have good tests for each support key-syntax, and that those include some tests where validation errors are triggered.

dougbu:
@rynowak updated

test/Microsoft.AspNet.Mvc.IntegrationTests/CollectionModelBinderIntegrationTest.cs
```diff
@@ -94,18 +97,12 @@ public class CollectionModelBinderIntegrationTest
...
             Assert.Equal(0, modelState.ErrorCount);
             Assert.True(modelState.IsValid);
-
-            var entry = Assert.Single(modelState, kvp => kvp.Key == "prefix[0]").Value;
-            Assert.Equal("10", entry.AttemptedValue);
-            Assert.Equal("10", entry.RawValue);
-
-            entry = Assert.Single(modelState, kvp => kvp.Key == "prefix[1]").Value;
-            Assert.Equal("11", entry.AttemptedValue);
-            Assert.Equal("11", entry.RawValue);
```
dougbu: Have enough tests that confirm the exact keys in `ModelState`. Removing some redundancy enabled more test consolidation.

rynowak: I don't think that's a goal at all.

dougbu: Wasn't a goal in itself but was getting hard to tell the difference between all the `[Fact]`s. The fewer `[Theory]`s helped me see the missing cases.

rynowak:
:shipit:

dougbu:
829a5c9

Issue #2890, opened 2015-07-31
Issue with arrays in expression when using DotNetCompilerPlatform


neoGeneva:
Hey, not 100% sure if this an MVC issue or Roslyn issue, but the issue exists when using  DotNetCompilerPlatform 1.0.0.0 and MVC 5.2.3.0

The issue is that array variables used in an array get the wrong name (the end up being prefixed with "CS$<>8__locals1.") when used with HtmlHelper expression methods, eg Html.NameFor()

Steps to reproduce:
1. In VS2015 create a new MVC5 project using the ASP.NET 4.6 template "MVC" (which already has Microsoft.CodeDom.Providers.DotNetCompilerPlatform installed)
2. Replace the content /Views/Home/Index.cshtml with the razor code below
3. Start the project and view the result in a browser.

```html
@{
    Layout = null;

    var someArray = new[]
    {
        new
        {
            someProperty = ""
        }
    };
}


<html>
<body>
    @Html.NameFor(_ => someArray[0].someProperty)<br />

    @for (var i = 0; i < someArray.Length; ++i)
    {
        @Html.NameFor(_ => someArray[i].someProperty)<br />
    }
</body>
</html>
```

The content of the page looks like the following, when the two rows of the output should be identical.

```
someArray[0].someProperty
CS$<>8__locals1.someArray[0].someProperty
```

neoGeneva:
Note that commenting out the "system.codedom" in web.config resolves the issue by essentially disabling the DotNetCompilerPlatform package.

dougbu:
MVC 5.2.3 issues are tracked on [CodePlex](http://aspnetwebstack.codeplex.com/workitem/list/advanced). Have you tried this scenario using MVC 6?

neoGeneva:
Yes, I can reproduce the issue in MVC6-beta5, this time using "ASP.NET 5 Preview Templates" -> "Web Application" and change the Home/Index.cshtml view as before.

The result was exactly the same.

Eilon:
@dougbu this is presumably because `i` is being closed over, which generates a new class, and I'm guessing that's what's showing up here. Does this happen in all closure scenarios when we generate names based on expressions?

dougbu:
> Does this happen in all closure scenarios when we generate names based on expressions?

[...]

PR #3027, opened 2015-08-23
Do not include compiler-generated names in expression names

dougbu:
- #2890
- add lots of `ExpressionHelper` tests using `IdFor()` and `NameFor()` (which are thin veneers)

### DIFF
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Expressions/ExpressionHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Expressions/ExpressionHelper.cs
@@ -31,9 +31,9 @@ namespace Microsoft.AspNet.Mvc.Rendering.Expressions
                 if (part.NodeType == ExpressionType.Call)
                 {
                     var methodExpression = (MethodCallExpression)part;
-
                     if (!IsSingleArgumentIndexer(methodExpression))
                     {
+                        // Unsupported.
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Expressions/ExpressionHelper.cs
```diff
@@ -58,7 +58,18 @@ public static string GetExpressionText([NotNull] LambdaExpression expression)
                 else if (part.NodeType == ExpressionType.MemberAccess)
                 {
                     var memberExpressionPart = (MemberExpression)part;
-                    nameParts.Push("." + memberExpressionPart.Member.Name);
+                    var name = memberExpressionPart.Member.Name;
+
+                    // If identifier contains "__", it is "reserved for use by the implementation" and likely compiler-
+                    // or Razor-generated e.g. the name of a field in a delegate's generated class.
+                    if (name.Contains("__"))
```
rynowak: Another way to verify this would probably to look at `memberExpressionPart.Member` and it's declaring class, which has `[CompilerGenerated]`. 

What we have here is probably good enough, but just in case we need to be more certain in the future.

rynowak:
:shipit: looks good based on what I know about this code. if you're unsure, get @Eilon to look at it :laughing:

dougbu:
a045324

PR #3029, opened 2015-08-24
Enable `TagHelper`s to be used inside of `HelperResult`s.

NTaylorMullen:
- The corresponding Razor change results in `HelperResult`s being rendered with async lambdas.
- This change enables `TagHelper`s and other async code to exist inside of `HelperResult` blocks.
- Added test to validate Templates (they generate `HelperResult`s) can utilize `TagHelper`s correctly.
- Rename `RazorPage`s `RenderBodyDelegate` to `RenderBodyDelegateAsync`.

aspnet/Razor#494

Corresponding Razor PR: https://github.com/aspnet/Razor/pull/498

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor/HelperResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/HelperResult.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.IO;
+using System.Threading.Tasks;
 using Microsoft.AspNet.Html.Abstractions;
 using Microsoft.Framework.Internal;
 using Microsoft.Framework.WebEncoders;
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNet.Mvc.Razor/HelperResult.cs
```diff
@@ -41,7 +44,7 @@ public Action<TextWriter> WriteAction
         /// <param name="encoder">The <see cref="IHtmlEncoder"/> to encode the content.</param>
         public virtual void WriteTo([NotNull] TextWriter writer, [NotNull] IHtmlEncoder encoder)
         {
-            _action(writer);
+            _asyncAction(writer).Wait();
```
rynowak: Other places in code we're using `.GetAwaiter().GetResult()` to rethrow without the aggregate exception

src/Microsoft.AspNet.Mvc.Razor/HelperResult.cs
```diff
@@ -14,24 +15,26 @@ namespace Microsoft.AspNet.Mvc.Razor
     /// </summary>
     public class HelperResult : IHtmlContent
     {
-        private readonly Action<TextWriter> _action;
+        private readonly Func<TextWriter, Task> _asyncAction;
```
rynowak: Is this really the only option to fix this bug? Should whatever is producing a helper result today be returning a `Task<IHtmlContent>`?

pranavkm: We could do that. Maybe by defining `TemplateContent : IHtmlContent` within Razor.Runtime?

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -126,7 +126,7 @@ public ITempDataDictionary TempData
         }
 
         /// <inheritdoc />
-        public Action<TextWriter> RenderBodyDelegate { get; set; }
+        public Func<TextWriter, Task> RenderBodyDelegateAsync { get; set; }
```
pranavkm: Might end up cleaning this up as part of https://github.com/aspnet/Mvc/issues/3030

src/Microsoft.AspNet.Mvc.Razor/HelperResult.cs
NTaylorMullen: Talked in person. Result: Has to stay as is.

src/Microsoft.AspNet.Mvc.Razor/HelperResult.cs
NTaylorMullen: :+1:

NTaylorMullen:
Addressed @rynowak's [comment] (https://github.com/aspnet/Mvc/pull/3029#discussion_r37799236).

pranavkm:
:shipit:

PR #3031, opened 2015-08-24
[PERF] Invoker perf

rynowak:
We allocate a separate list for model-binding related objects when we
create the resource filter contexts, and these lists then live the
lifetime of the request. These *may* be modified by user code in
filters as a supported feature, but rarely are changed in practice.

This change adds a simple CopyOnWriteList implementation to reduce the
amount of copying that's actually done.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/FilterActionInvoker.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/FilterActionInvoker.cs
@@ -7,6 +7,7 @@ using System.Diagnostics;
 using System.Linq;
 using System.Runtime.ExceptionServices;
 using System.Threading.Tasks;
+using Microsoft.AspNet.Mvc.Internal;
 using Microsoft.AspNet.Mvc.ModelBinding;
 using Microsoft.AspNet.Mvc.ModelBinding.Validation;
 using Microsoft.Framework.Internal;
@@ -241,11 +242,11 @@ namespace Microsoft.AspNet.Mvc.Core
 
             var context = new ResourceExecutingContext(ActionContext, _filters);
 
-            context.InputFormatters = new List<IInputFormatter>(_inputFormatters);
-            context.OutputFormatters = new List<IOutputFormatter>(_outputFormatters);
-            context.ModelBinders = new List<IModelBinder>(_modelBinders);
-            context.ValidatorProviders = new List<IModelValidatorProvider>(_modelValidatorProviders);
-            context.ValueProviderFactories = new List<IValueProviderFactory>(_valueProviderFactories);
+            context.InputFormatters = new CopyOnWriteList<IInputFormatter>(_inputFormatters);
+            context.OutputFormatters = new CopyOnWriteList<IOutputFormatter>(_outputFormatters);
+            context.ModelBinders = new CopyOnWriteList<IModelBinder>(_modelBinders);
+            context.ValidatorProviders = new CopyOnWriteList<IModelValidatorProvider>(_modelValidatorProviders);
+            context.ValueProviderFactories = new CopyOnWriteList<IValueProviderFactory>(_valueProviderFactories);
 
             _resourceExecutingContext = context;
             await InvokeResourceFilterAsync();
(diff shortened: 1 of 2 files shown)


rynowak:
What we're looking at here is a profile of an async-state-machine method for `InvokeAllResourceFiltersAsync()`. This method is doing the work to set up resource filters, including copying the modelbinding 'ingredients' from options into mutable lists.  There are 15000 requests in both the sampling profile and the allocation profile.

Overall,`InvokeAllResourceFiltersAsync()->MoveNext()` is responsible for about 5% of the time spend in MVC.
**Before**:

before-execution

Notice the time spent inside of the method itself and allocating lists.

before-alloc

Notice here not just the number of allocations of these lists, but the *size* of them, because we're doing a copy. Some of these lists (`List<IModelBinder>` for one) have another copy elsewhere we can optimize out.

**After:**
after-execution

Overhead of `InvokeAllResourceFiltersAsync()->MoveNext()` reduced from 6.73% to 1.83%. Note that this is just the change in execution time, we're also saving ourselves some future GC.

after-alloc

Notice the size of these items is much smaller.

src/Microsoft.AspNet.Mvc.Core/Internal/CopyOnWriteList.cs
```diff
@@ -0,0 +1,140 @@
...
+        {
+            foreach (var obj in Readable)
+            {
+                if (object.Equals(obj, item))
+                {
+                    return true;
+                }
+            }
+
+            return false;
```
dougbu: Why not `Readable.Contains()`? Similar questions for `CopyTo()` and `IndexOf()`.

src/Microsoft.AspNet.Mvc.Core/FilterActionInvoker.cs
```diff
@@ -66,7 +67,7 @@ public abstract class FilterActionInvoker : IActionInvoker
             [NotNull] IActionBindingContextAccessor actionBindingContextAccessor,
             [NotNull] ILogger logger,
             [NotNull] INotifier notifier,
-            int maxModelValidationErrors)
+            [NotNull] int maxModelValidationErrors)
```
dougbu: Why add `[NotNull]` to an `int` parameter?

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.Core/Internal/CopyOnWriteList.cs
```diff
@@ -0,0 +1,140 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNet.Mvc.Internal
+{
+    public class CopyOnWriteList<T> : IList<T>
```
davidfowl: Should this move to common?

src/Microsoft.AspNet.Mvc.Core/Internal/CopyOnWriteList.cs
rynowak: `IReadOnlyList<T>` isn't bloated like `IList<T>`

src/Microsoft.AspNet.Mvc.Core/Internal/CopyOnWriteList.cs
rynowak: Probably will be needed. Next step is routing.

src/Microsoft.AspNet.Mvc.Core/FilterActionInvoker.cs
rynowak: So's it won't be null! duh-doi :+1:

src/Microsoft.AspNet.Mvc.Core/Internal/CopyOnWriteList.cs
davidfowl: I'm also assuming that if this were a struct you'd lose all the benefits because all of the places that call through are expecting IList right?

davidfowl:
:shipit:

src/Microsoft.AspNet.Mvc.Core/Internal/CopyOnWriteList.cs
rynowak: Yep. @eilon already made us change those to `IList<T>`

PR #3032, opened 2015-08-24
[PERF] Remove extra ModelStateDictionary allocations

rynowak:
The copy constructor is chaining to the wrong constructor. This results in
an extra 8 allocations of ModelStateDictionary per-request. All of the
various filter contexts inherit from ActionContext, that's how you get the
8 extras.

Small problem but easy fix.


rynowak:
Here's some profiling data showing 9 allocations of `ModelStateDictionary` per-request (15000 requests). 

msd

pranavkm:
:shipit:

PR #3037, opened 2015-08-25
Modify RazorPreCompileModule to use an instance of memory cache specific to the application's target framework and configuration.

pranavkm:
cc @davidfowl  \ @anurse

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor/Precompilation/RazorPreCompileModule.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/Precompilation/RazorPreCompileModule.cs
@@ -2,6 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
+using System.Runtime.Versioning;
 using Microsoft.AspNet.FileProviders;
 using Microsoft.Dnx.Compilation.CSharp;
 using Microsoft.Dnx.Runtime;
@@ -16,7 +18,9 @@ namespace Microsoft.AspNet.Mvc.Razor.Precompilation
     public abstract class RazorPreCompileModule : ICompileModule
     {
         private readonly IAssemblyLoadContext _loadContext;
-        private readonly IMemoryCache _memoryCache;
+        private readonly object _memoryCacheLookupLock = new object();
+        private readonly Dictionary<PrecompilationCacheKey, MemoryCache> _memoryCacheLookup =
+            new Dictionary<PrecompilationCacheKey, MemoryCache>();
 
         /// <summary>
         /// Instantiates a new <see cref="RazorPreCompileModule"/> instance.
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Mvc.Razor/Precompilation/RazorPreCompileModule.cs
```diff
@@ -0,0 +1,101 @@
...
+
+                    _memoryCacheLookup[cacheKey] = memoryCache;
+                }
+            }
+
+            var viewCompiler = new RazorPreCompiler(
+                context,
+                _loadContext,
+                fileProvider,
+                memoryCache)
```
pranavkm: MemoryCache is limited to using string keys - so we're stuck with either creating a composite key using the tfm and configuration - or varying the instance in itself. This seemed cleaner.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/RazorPreCompileModule.cs
```diff
@@ -0,0 +1,101 @@
...
+            public bool Equals(PrecompilationCacheKey other)
+            {
+                return
+                    other.TargetFramework == TargetFramework &&
+                    string.Equals(other.Configuration, Configuration, StringComparison.Ordinal);
+            }
+
+            public override int GetHashCode()
+            {
+                return 13 * Configuration.GetHashCode() + TargetFramework.GetHashCode();
```
anurse: HashCodeCombiner? Now in the handy-dandy `Microsoft.Framework.HashCodeCombiner.Sources` package :)

anurse:
:shipit:

The cache-per-(config,tfm) thing is a little odd but I think you're right that it's easier than changing the whole cache interface.

src/Microsoft.AspNet.Mvc.Razor/Precompilation/RazorPreCompileModule.cs
```diff
@@ -0,0 +1,101 @@
...
+        private readonly Dictionary<PrecompilationCacheKey, MemoryCache> _memoryCacheLookup =
+            new Dictionary<PrecompilationCacheKey, MemoryCache>();
+
+        /// <summary>
+        /// Instantiates a new <see cref="RazorPreCompileModule"/> instance.
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceProvider"/> for the application.</param>
+        public RazorPreCompileModule(IServiceProvider services)
+        {
+            _loadContext = services.GetRequiredService<IAssemblyLoadContext>();
```
davidfowl: Do you need this?

pranavkm: That should go away once this PR goes in - https://github.com/aspnet/Mvc/pull/2961

Issue #2931, opened 2015-08-07
Default for catchall segment doesn't work


Daniel15:
```csharp
[Route("Foo/{*path}")]
public ActionResult Foo(string path = "default")
{
	if (path == null)
	{
		return Content("NULL");
	}
	if (string.IsNullOrEmpty(path))
	{
		return Content("EMPTY");
	}
	return Content("Path: " + path);
}
```

If you hit `/Foo/` on ASP.NET MVC 5, it outputs "Path: default". However, ASP.NET 5 + MVC 6 outputs "NULL". Hitting `/Foo/Bar` correctly displays "Path: Bar" on both versions. This is the same if you change the default to an empty string - ASP.NET MVC 5 returns "EMPTY" whereas MVC 6 returns "NULL".

I also tried `[Route("Foo/{*path=default}")]` which did not work either.

Wasn't sure whether to report this in the Routing repo or here; I've reported it here since I think catchall segments are an MVC thing (and also `RouteAttribute` is an MVC thing)

Eilon:
Certainly seems like a bug...

rynowak:
I'm wondering if the route value provider is missing a null-or-empty check. My understanding is that a catch-all segment will always have a value, even if it's the empty string.

PR #3038, opened 2015-08-25
AttributeRoute does not replace existing route values with null

ajaybhargavb:
#2931 

@rynowak @pranavkm

### DIFF
--- a/test/Microsoft.AspNet.Mvc.Core.Test/Routing/InnerAttributeRouteTest.cs
+++ b/test/Microsoft.AspNet.Mvc.Core.Test/Routing/InnerAttributeRouteTest.cs
@@ -1464,6 +1464,62 @@ namespace Microsoft.AspNet.Mvc.Routing
         }
 
         [Fact]
+        public async Task AttributeRoute_ReplacesExistingRouteValues_IfNotNull()
+        {
+            // Arrange
+            var router = new Mock<IRouter>();
+            router
+                .Setup(r => r.RouteAsync(It.IsAny<RouteContext>()))
+                .Callback<RouteContext>((c) =>
+                {
+                    c.IsHandled = true;
+                })
+                .Returns(Task.FromResult(true));
+
+            var entry = CreateMatchingEntry(router.Object, "Foo/{*path}", order: 0);
+            var route = CreateAttributeRoute(router.Object, entry);
+
+            var context = CreateRouteContext("/Foo/Bar");
+
+            var originalRouteData = context.RouteData;
+            originalRouteData.Values.Add("path", "default");
+
+            // Act
+            await route.RouteAsync(context);
+
+            // Assert
+            Assert.Equal("Bar", context.RouteData.Values["path"]);
+        }
+
+        [Fact]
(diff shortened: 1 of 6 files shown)


test/Microsoft.AspNet.Mvc.Core.Test/Routing/InnerAttributeRouteTest.cs
```diff
@@ -1464,6 +1464,62 @@ public static IEnumerable<object[]> OptionalParamValues
...
+            originalRouteData.Values.Add("path", "default");
+
+            // Act
+            await route.RouteAsync(context);
+
+            // Assert
+            Assert.Equal("Bar", context.RouteData.Values["path"]);
+        }
+
+        [Fact]
```
pranavkm: Not super familiar with routing code, but is it possible for the router to route without visiting the entry (and consequently invoke `MergeAttributes`?)

rynowak: Routing creates a defensive copy at each level so that changes that aren't persisted don't dirty something else on the fallback path. That's what's on the whiteboard next to you, you silly goose! :goat:

test/Microsoft.AspNet.Mvc.FunctionalTests/RoutingTests.cs
```diff
@@ -134,6 +134,27 @@ public class RoutingTests
             Assert.Equal(HttpStatusCode.NotFound, response.StatusCode);
         }
 
+        [Theory]
+        [InlineData("", "/Home/OptionalPath/default")]
+        [InlineData("CustomPath", "/Home/OptionalPath/CustomPath")]
+        public async Task ConventionRoutedController_WithOptionalSegment(string optionalSegment, string expected)
```
rynowak: `Convention` -> `Conventional`

test/Microsoft.AspNet.Mvc.Core.Test/Routing/InnerAttributeRouteTest.cs
```diff
@@ -1464,6 +1464,62 @@ public static IEnumerable<object[]> OptionalParamValues
         }
 
         [Fact]
+        public async Task AttributeRoute_ReplacesExistingRouteValues_IfNotNull()
```
rynowak: What should be the behavior for the empty string?

ajaybhargavb: Is there a scenario where it can be an empty string? If so we should add that to the condition. But I can't think of a scenario.

rynowak: What does template route do?

ajaybhargavb: It will retain the default value(if exists) or will be set as null. It won't be set as empty string.

rynowak:
:shipit:

ajaybhargavb:
4fbaea2463cf2d15a0cb5a87a24b69af3d397ca9

PR #3047, opened 2015-08-27
Add FieldName to model binding context

rynowak:
Adds a new property, FieldName, to ModelBindingContext. The FieldName is
the name of whatever code-element is being bound, regardless of what
model-prefix is in use.

This is needed for cases like the Header model binder. We always want to
use the property/parameter name and we don't care about model prefixes.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
+++ b/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using Microsoft.AspNet.Mvc.Abstractions;
 using Microsoft.Framework.Internal;
 
 namespace Microsoft.AspNet.Mvc.ModelBinding
@@ -15,10 +14,6 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         private static readonly Func<ModelBindingContext, string, bool>
             _defaultPropertyFilter = (context, propertyName) => true;
 
-        private string _modelName;
-        private ModelStateDictionary _modelState;
-        private Func<ModelBindingContext, string, bool> _propertyFilter;
-
         /// <summary>
         /// Initializes a new instance of the <see cref="ModelBindingContext"/> class.
         /// </summary>
@@ -27,67 +22,82 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         }
 
         /// <summary>
-        /// Constructs a new instance of the <see cref="ModelBindingContext"/> class using the
-        /// <paramref name="bindingContext" />.
+        /// Creates a new <see cref="ModelBindingContext"/> for top-level model binding operation.
         /// </summary>
-        /// <param name="bindingContext">Existing <see cref="ModelBindingContext"/>.</param>
-        /// <param name="modelName">Model name of associated with the new <see cref="ModelBindingContext"/>.</param>
-        /// <param name="modelMetadata">Model metadata of associated with the new <see cref="ModelBindingContext"/>.
+        /// <param name="operationBindingContext">
+        /// The <see cref="OperationBindingContext"/> associated with the binding operation.
         /// </param>
-        public static ModelBindingContext GetChildModelBindingContext(
-            [NotNull] ModelBindingContext bindingContext,
-            [NotNull] string modelName,
-            [NotNull] ModelMetadata modelMetadata)
-        {
-            var modelBindingContext = new ModelBindingContext
-            {
-                ModelName = modelName,
-                ModelMetadata = modelMetadata,
-
-                ModelState = bindingContext.ModelState,
-                ValueProvider = bindingContext.ValueProvider,
-                OperationBindingContext = bindingContext.OperationBindingContext,
-
-                BindingSource = modelMetadata.BindingSource,
-                BinderModelName = modelMetadata.BinderModelName,
-                BinderType = modelMetadata.BinderType,
-            };
-
-            return modelBindingContext;
-        }
-
-        /// <summary>
-        /// Constructs a new instance of <see cref="ModelBindingContext"/> from given <paramref name="metadata"/>
-        /// and <paramref name="bindingInfo"/>.
-        /// </summary>
         /// <param name="metadata"><see cref="ModelMetadata"/> associated with the model.</param>
         /// <param name="bindingInfo"><see cref="BindingInfo"/> associated with the model.</param>
-        /// <param name="modelName">An optional name of the model to be used.</param>
+        /// <param name="modelName">The name of the property or parameter being bound.</param>
         /// <returns>A new instance of <see cref="ModelBindingContext"/>.</returns>
-        public static ModelBindingContext GetModelBindingContext(
+        public static ModelBindingContext CreateBindingContext(
+            [NotNull] OperationBindingContext operationBindingContext,
+            [NotNull] ModelStateDictionary modelState,
             [NotNull] ModelMetadata metadata,
             BindingInfo bindingInfo,
-            string modelName)
+            [NotNull] string modelName)
         {
             var binderModelName = bindingInfo?.BinderModelName ?? metadata.BinderModelName;
             var propertyPredicateProvider =
                 bindingInfo?.PropertyBindingPredicateProvider ?? metadata.PropertyBindingPredicateProvider;
-            return new ModelBindingContext
+
+            return new ModelBindingContext()
             {
-                ModelMetadata = metadata,
-                BindingSource = bindingInfo?.BindingSource ?? metadata.BindingSource,
-                PropertyFilter = propertyPredicateProvider?.PropertyFilter,
-                BinderType = bindingInfo?.BinderType ?? metadata.BinderType,
                 BinderModelName = binderModelName,
-                ModelName = binderModelName ?? metadata.PropertyName ?? modelName,
+                BindingSource = bindingInfo?.BindingSource ?? metadata.BindingSource,
+                BinderType = bindingInfo?.BinderType ?? metadata.BinderType,
+                PropertyFilter = propertyPredicateProvider?.PropertyFilter,
+
+                // We only support fallback to empty prefix in cases where the model name is inferred from
+                // the parameter or property being bound.
                 FallbackToEmptyPrefix = binderModelName == null,
+
+                // Because this is the top-level context, FieldName and ModelName should be the same.
+                FieldName = binderModelName ?? modelName,
+                ModelName = binderModelName ?? modelName,
+
+                IsTopLevelObject = true,
+                ModelMetadata = metadata,
+                ModelState = modelState,
+                OperationBindingContext = operationBindingContext,
+                ValueProvider = operationBindingContext.ValueProvider,
+            };
+        }
+
+        public static ModelBindingContext CreateChildBindingContext(
+            [NotNull] ModelBindingContext parent,
+            [NotNull] ModelMetadata modelMetadata,
+            [NotNull] string fieldName,
+            [NotNull] string modelName,
+            object model)
+        {
+            return new ModelBindingContext()
+            {
+                ModelState = parent.ModelState,
+                OperationBindingContext = parent.OperationBindingContext,
+                ValueProvider = parent.ValueProvider,
+
+                Model = model,
+                ModelMetadata = modelMetadata,
+                ModelName = modelName,
+                FieldName = fieldName,
+                BinderModelName = modelMetadata.BinderModelName,
+                BinderType = modelMetadata.BinderType,
+                BindingSource = modelMetadata.BindingSource,
+                PropertyFilter = modelMetadata.PropertyBindingPredicateProvider?.PropertyFilter,
             };
         }
 
         /// <summary>
         /// Represents the <see cref="OperationBindingContext"/> associated with this context.
         /// </summary>
-        public OperationBindingContext OperationBindingContext { get; set; }
+        public OperationBindingContext OperationBindingContext { get; [param:NotNull] set; }
+
+        /// <summary>
+        /// Gets or sets the name of the current field being bound.
+        /// </summary>
+        public string FieldName { get; [param: NotNull] set; }
 
         /// <summary>
         /// Gets or sets the model value for the current operation.
@@ -101,41 +111,19 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         /// <summary>
         /// Gets or sets the metadata for the model associated with this context.
         /// </summary>
-        public ModelMetadata ModelMetadata { get; set; }
+        public ModelMetadata ModelMetadata { get; [param: NotNull] set; }
 
         /// <summary>
         /// Gets or sets the name of the model. This property is used as a key for looking up values in
         /// <see cref="IValueProvider"/> during model binding.
         /// </summary>
-        public string ModelName
-        {
-            get
-            {
-                if (_modelName == null)
-                {
-                    _modelName = string.Empty;
-                }
-                return _modelName;
-            }
-            set { _modelName = value; }
-        }
+        public string ModelName { get; [param: NotNull] set; }
 
         /// <summary>
         /// Gets or sets the <see cref="ModelStateDictionary"/> used to capture <see cref="ModelState"/> values
         /// for properties in the object graph of the model when binding.
         /// </summary>
-        public ModelStateDictionary ModelState
-        {
-            get
-            {
-                if (_modelState == null)
-                {
-                    _modelState = new ModelStateDictionary();
-                }
-                return _modelState;
-            }
-            set { _modelState = value; }
-        }
+        public ModelStateDictionary ModelState { get; [param: NotNull] set; }
 
         /// <summary>
         /// Gets the type of the model.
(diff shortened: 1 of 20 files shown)


rynowak:
Other misc cleanup in this PR:
- Removed 'magical' behavior of MBC properties in favor of explicit init in tests
- Cleaned up API for creating a 'child' context, and used it more consistenly

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
```diff
@@ -27,60 +22,75 @@ public ModelBindingContext()
...
+
+                IsTopLevelObject = true,
+                ModelMetadata = metadata,
+                ModelState = modelState,
+                OperationBindingContext = operationBindingContext,
+                ValueProvider = operationBindingContext.ValueProvider,
+            };
+        }
+
+        public static ModelBindingContext CreateChildBindingContext(
```
rynowak: I didn't bother adding docs because this is going to be removed 1-2 changes from now. It's still valuable to cleanup because the code changes in the PR that replaces this will be virtually identical.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
```diff
@@ -149,12 +149,16 @@ protected object CreateInstance(Type targetType)
                 bindingContext.ModelName,
                 bindingContext.ModelMetadata,
                 boundCollection);
+
+            var innerBindingContext = ModelBindingContext.CreateChildBindingContext(
+                bindingContext,
+                elementMetadata,
+                fieldName: string.Empty,
```
rynowak: Couldn't think of anything super useful for collection model binder to put here, so it's just empty. I can't think of a scenario where it would be used.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CompositeModelBinder.cs
```diff
@@ -148,6 +148,7 @@ public CompositeModelBinder([NotNull] IEnumerable<IModelBinder> modelBinders)
                 Model = oldBindingContext.Model,
                 ModelMetadata = oldBindingContext.ModelMetadata,
                 ModelName = modelName,
+                FieldName = oldBindingContext.FieldName,
```
rynowak: This is the one location in code where we still have ad-hoc creation of an MBC that doesn't really follow an established pattern. My next changes will remove it.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
```diff
@@ -359,7 +365,7 @@ protected virtual object GetModel([NotNull] ModelBindingContext bindingContext)
                 var modelMetadataPredicate = context.ModelMetadata.PropertyBindingPredicateProvider?.PropertyFilter;
 
                 return
-                    context.PropertyFilter(context, propertyName) &&
+                    (context.PropertyFilter == null || context.PropertyFilter(context, propertyName)) &&
                     (modelMetadataPredicate == null || modelMetadataPredicate(context, propertyName));
```
rynowak: This can be null now. It's easier to just allow it to be null than to try and force it be initialize/copied everywhere.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/CompositeModelBinderTest.cs
```diff
@@ -24,14 +24,11 @@ public class CompositeModelBinderTest
...
                 ModelName = "someName",
                 ModelState = new ModelStateDictionary(),
+                OperationBindingContext = new OperationBindingContext(),
                 ValueProvider = new SimpleValueProvider
                 {
                     { "someName", "dummyValue" }
                 },
-                OperationBindingContext = new OperationBindingContext
-                {
-                    ValidatorProvider = GetValidatorProvider()
```
rynowak: Validators were still instantiated by this code even though they weren't used. I removed it.

NeelBhatt:
```private static readonly Func<ModelBindingContext, string, bool>
             _defaultPropertyFilter = (context, propertyName) => true;```

should be like:

```private static readonly Func<ModelBindingContext, string, bool>
             DefaultPropertyFilter = (context, propertyName) => true;```

due to Pascal case i guess?

File: src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs [hunk 6]
```diff
@@ -27,60 +22,75 @@ public ModelBindingContext()
...
+            [NotNull] ModelBindingContext parent,
+            [NotNull] ModelMetadata modelMetadata,
+            [NotNull] string fieldName,
+            [NotNull] string modelName,
+            object model)
+        {
+            return new ModelBindingContext()
+            {
+                IsFirstChanceBinding = parent.IsFirstChanceBinding,
```
dougbu: This changes the meaning of `IsFirstChanceBinding`. See `<remarks>` on that property: `Should not be <c>true</c> when <see cref="IsTopLevelObject"/> is <c>false</c>.`

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
```diff
@@ -27,60 +22,75 @@ public ModelBindingContext()
...
+                ModelState = parent.ModelState,
+                OperationBindingContext = parent.OperationBindingContext,
+                ValueProvider = parent.ValueProvider,
+
+                // These are never true for a child context.
+                FallbackToEmptyPrefix = false,
+                IsTopLevelObject = false,
```
dougbu: Why initialize to default values?

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
```diff
@@ -204,25 +190,10 @@ public Type ModelType
...
-            if (ModelMetadata == null)
-            {
-                throw new InvalidOperationException(Resources.ModelBindingContext_ModelMetadataMustBeSet);
-            }
-        }
+        /// <summary>
+        /// Gets or sets a predicate which will be evaluated for each property to determine if the property
+        /// is eligible for model binding.
+        /// </summary>
+        public Func<ModelBindingContext, string, bool> PropertyFilter = _defaultPropertyFilter;
```
dougbu: This should not be a field. Think you mean
``` c#
public Func<ModelBindingContext, string, bool> PropertyFilter { get; [param: NotNull] set; } =
    _defaultPropertyFilter;
```

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
```diff
@@ -90,6 +100,11 @@ public ModelBindingContext()
         public OperationBindingContext OperationBindingContext { get; set; }
 
         /// <summary>
+        /// Gets or sets the name of the current field being bound.
+        /// </summary>
+        public string FieldName { get; set; }
```
dougbu: This needs `[param: NotNull]` for the setter. Same for `ModelName`, `ModelState`, and most of the existing properties.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
```diff
@@ -222,10 +226,13 @@ protected object CreateInstance(Type targetType)
             foreach (var indexName in indexNames)
             {
                 var fullChildName = ModelNames.CreateIndexModelName(bindingContext.ModelName, indexName);
-                var childBindingContext = ModelBindingContext.GetChildModelBindingContext(
+                var childBindingContext = ModelBindingContext.CreateChildBindingContext(
                     bindingContext,
-                    fullChildName,
-                    elementMetadata);
+                    elementMetadata,
+                    fieldName: string.Empty,
```
dougbu: Suggest `indexName`

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
dougbu: Suggest `bindingContext.FieldName` since inner binder is handling (individual values for) the same property.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
```diff
@@ -57,10 +57,12 @@ public class DictionaryModelBinder<TKey, TValue> : CollectionModelBinder<KeyValu
             // Update the existing successful but empty ModelBindingResult.
             var metadataProvider = bindingContext.OperationBindingContext.MetadataProvider;
             var valueMetadata = metadataProvider.GetMetadataForType(typeof(TValue));
-            var valueBindingContext = ModelBindingContext.GetChildModelBindingContext(
+            var valueBindingContext = ModelBindingContext.CreateChildBindingContext(
                 bindingContext,
-                bindingContext.ModelName,
-                valueMetadata);
+                valueMetadata,
+                fieldName: string.Empty,
```
dougbu: Suggest `bindingContext.FieldName`

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
dougbu: If a `null` `PropertyFilter` is :ok:, why initialize the property to `_defaultPropertyFilter`? Suggest removing `_defaultPropertyFilter`.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
```diff
@@ -27,60 +22,75 @@ public ModelBindingContext()
...
-        /// <paramref name="bindingContext" />.
+        /// Creates a new <see cref="ModelBindingContext"/> for top-level model binding operation.
         /// </summary>
-        /// <param name="bindingContext">Existing <see cref="ModelBindingContext"/>.</param>
-        /// <param name="modelName">Model name of associated with the new <see cref="ModelBindingContext"/>.</param>
-        /// <param name="modelMetadata">Model metadata of associated with the new <see cref="ModelBindingContext"/>.
+        /// <param name="operationBindingContext">
+        /// The <see cref="OperationBindingContext"/> associated with the binding operation.
         /// </param>
-        public static ModelBindingContext GetChildModelBindingContext(
```
dougbu: FYI reordering these methods made reviewing more difficult.

src/Microsoft.AspNet.Mvc.Core/ModelBindingHelper.cs
```diff
@@ -293,22 +293,19 @@ public static class ModelBindingHelper
...
-            };
-
-            var modelBindingContext = new ModelBindingContext
-            {
-                Model = model,
-                ModelMetadata = modelMetadata,
-                ModelName = prefix,
-                ModelState = modelState,
+                HttpContext = httpContext,
                 ValueProvider = valueProvider,
```
dougbu: Previously `operationBindingContext.ValueProvider` would be left `null`. Doubt that was intentional but did you have to change anything else (e.g. integration tests) as a direct result of the change here?

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/CompositeModelBinderTest.cs
```diff
@@ -341,7 +321,7 @@ public class CompositeModelBinderTest
                 FallbackToEmptyPrefix = true,
                 ModelMetadata = new EmptyModelMetadataProvider().GetMetadataForType(typeof(int)),
                 ModelState = new ModelStateDictionary(),
-                OperationBindingContext = Mock.Of<OperationBindingContext>(),
+                OperationBindingContext = new OperationBindingContext(),
```
dougbu: Nice catch. Was obviously incredibly important to mock this POCO.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/CompositeModelBinderTest.cs
```diff
@@ -547,26 +526,6 @@ private static CompositeModelBinder CreateCompositeBinder(IModelBinder mockIntBi
             return shimBinder;
         }
 
-        private static IModelValidatorProvider GetValidatorProvider(params IModelValidator[] validators)
```
dougbu: Yah mean /none/ of the tests here validated? That makes sense since this is a unit test but still surprising this code was ever written. Hopefully was just over-zealous copying 'n pasting :smile:

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/ModelBindingContextTest.cs
```diff
@@ -10,14 +10,14 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Test
     public class ModelBindingContextTest
     {
         [Fact]
-        public void GetChildModelBindingContext()
+        public void CreateChildBindingContext()
```
dougbu: While you're fixing the test name, please add something about expectations to it.

dougbu:
:watch:

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/CompositeModelBinderTest.cs
rynowak: When validation used to take place *during* model binding, this test made sense. It was just never cleaned up.

src/Microsoft.AspNet.Mvc.Core/ModelBindingHelper.cs
rynowak: No, nothing had to be changed in the tests. It's possible that we had a bug in the case where a model property needs to 'reset' the value provider.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
rynowak: :sob:

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
rynowak: sure, will fix it

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
rynowak: I thought it added clarity. Honestly I wasn't really thinking about it, just making sure everything was initialized.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
rynowak: Yep.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
rynowak: Sure

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
rynowak: sure

src/Microsoft.AspNet.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
rynowak: sure

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
rynowak: Yeah, I like that idea.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/ModelBindingContextTest.cs
rynowak: :OK:

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
rynowak: Added these

rynowak:
@dougbu  updated

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
```diff
@@ -15,10 +14,6 @@ public class ModelBindingContext
         private static readonly Func<ModelBindingContext, string, bool>
             _defaultPropertyFilter = (context, propertyName) => true;
```
dougbu: Not used. Ready for :boom: :boom: :boom:

dougbu:
:shipit:

PR #3049, opened 2015-08-27
Change ModelBinding to use a single pass

rynowak:


rynowak:
@dougbu ping

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CompositeModelBinder.cs
```diff
@@ -195,6 +169,32 @@ public CompositeModelBinder([NotNull] IEnumerable<IModelBinder> modelBinders)
                 }
             }
```
dougbu: Can we now move this early-exit block above creation of `newBindingContext`? Doesn't depend on `BodyBindingState`.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CompositeModelBinder.cs
```diff
@@ -195,6 +169,32 @@ public CompositeModelBinder([NotNull] IEnumerable<IModelBinder> modelBinders)
...
 
+            if (oldBindingContext.IsTopLevelObject)
+            {
+                if (oldBindingContext.BinderModelName != null)
+                {
+                    newBindingContext.ModelName = oldBindingContext.BinderModelName;
+                }
+                else if (oldBindingContext.BindingSource != null && oldBindingContext.BindingSource.IsGreedy)
```
dougbu: Suggest using `bindingSource` here.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/FormFileModelBinder.cs
```diff
@@ -82,8 +82,15 @@ public class FormFileModelBinder : IModelBinder
...
                     }
 
-                    var modelName = HeaderUtilities.RemoveQuotes(parsedContentDisposition.Name);
-                    if (modelName.Equals(bindingContext.ModelName, StringComparison.OrdinalIgnoreCase))
+                    // If we're at the top level, then use the FieldName (paramter or property name).
+                    // This handles the fact that there will be nothing in the ValueProviders for this parameter
+                    // and so we'll do the right thing even though we 'fell-back' to the empty prefix.
+                    var modelName = bindingContext.IsTopLevelObject
+                        ? bindingContext.FieldName
+                        : bindingContext.ModelName;
```
dougbu: Do this above the loop.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/CompositeModelBinderTest.cs
```diff
@@ -65,6 +65,7 @@ public class CompositeModelBinderTest
             var bindingContext = new ModelBindingContext
             {
                 FallbackToEmptyPrefix = true,
+                IsTopLevelObject = true,
```
dougbu: Why was this addition or the one at [line 497](https://github.com/aspnet/Mvc/pull/3049/files#diff-2692e9de2424d15a779aa403c7c6c95aR497) necessary? If needed, why were the tests passing before?

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/FormFileModelBinderTest.cs
```diff
@@ -116,6 +117,35 @@ public class FormFileModelBinderTest
             Assert.Null(result.Model);
         }
 
+        [Theory]
+        [InlineData(true, "FieldName")]
```
dougbu: Why is this a `[Theory]`?

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/FormFileModelBinderTest.cs
```diff
@@ -116,6 +117,35 @@ public class FormFileModelBinderTest
...
+            bindingContext.ModelName = "ModelName";
+
+            var binder = new FormFileModelBinder();
+
+            // Act
+            var result = await binder.BindModelAsync(bindingContext);
+
+            // Assert
+            Assert.NotNull(result);
+            var file = Assert.IsAssignableFrom<IFormFile>(result.Model);
```
dougbu: suggest `Assert.True(result.IsModelSet)` before using `result.Model`

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/MutableObjectModelBinderTest.cs
```diff
@@ -65,10 +47,6 @@ public class MutableObjectModelBinderTest
                         MetadataProvider = metadataProvider,
                         ValidatorProvider = Mock.Of<IModelValidatorProvider>(),
                     },
-
-                    // CanCreateModel() ignores the BinderModelName and ModelName properties.
```
dougbu: This code was intended to ensure these properties do /not/ affect `CanCreateModel()`. Please restore the corresponding parameters and inline data.

test/Microsoft.AspNet.Mvc.FunctionalTests/ModelBindingTest.cs
```diff
@@ -860,10 +860,11 @@ public class ModelBindingTest
             // Assert
             var json = JsonConvert.DeserializeObject<Dictionary<string, string>>(response);
             Assert.Equal(3, json.Count);
+
             // Mono issue - https://github.com/aspnet/External/issues/19
-            Assert.Equal(PlatformNormalizer.NormalizeContent("The Field1 field is required."), json["Field1"]);
```
dougbu: Suggest adding a comment explaining why the prefix is used in the errors e.g. `// Prefix exists in value providers but requested values do not.`

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CompositeModelBinder.cs
rynowak: Sure.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CompositeModelBinder.cs
rynowak: sure

src/Microsoft.AspNet.Mvc.Core/ModelBinding/FormFileModelBinder.cs
rynowak: :+1:

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/FormFileModelBinderTest.cs
rynowak: There's supposed to be another case here :) OOps.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/CompositeModelBinderTest.cs
rynowak: Most of these tests cover "fallback to empty prefix" behavior. We now only fallback when `IsTopLevelModel == true`

test/Microsoft.AspNet.Mvc.IntegrationTests/ModelPrefixSelectionIntegrationTest.cs
```diff
@@ -0,0 +1,394 @@
...
+        {
+            // Arrange
+            var argumentBinder = ModelBindingTestHelper.GetArgumentBinder();
+            var parameter = new ParameterDescriptor()
+            {
+                Name = "parameter",
+                ParameterType = typeof(Person2),
+                BindingInfo = new BindingInfo()
+                {
+                    BindingSource = BindingSource.Query,
```
dougbu: Suggest an additional test almost identical to this one but with `BindingSource = BindingSource.Path`. That'll demonstrate filtering really does affect the prefix choice.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/FormFileModelBinderTest.cs
rynowak: :+1:

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/MutableObjectModelBinderTest.cs
rynowak: Discussed in person, and agreed not to add back.

test/Microsoft.AspNet.Mvc.FunctionalTests/ModelBindingTest.cs
rynowak: Sure.

test/Microsoft.AspNet.Mvc.IntegrationTests/ModelPrefixSelectionIntegrationTest.cs
```diff
@@ -0,0 +1,394 @@
...
+
+            var modelState = new ModelStateDictionary();
+
+            // Act
+            var modelBindingResult = await argumentBinder.BindModelAsync(parameter, modelState, operationContext);
+
+            // Assert
+            Assert.NotNull(modelBindingResult);
+            Assert.True(modelBindingResult.IsModelSet);
+            Assert.Equal(string.Empty, modelBindingResult.Key);
```
dougbu: Suggest `Assert.Empty(modelBindingResult.Key)` a few times in this class.

test/Microsoft.AspNet.Mvc.IntegrationTests/ModelPrefixSelectionIntegrationTest.cs
rynowak: sure

dougbu: Realize `ComplexModel_EmptyPrefixSelected_NoMatchingValueProviderValue_WithFilteredValueProviders()` is close but has an additional non-empty value provider.

test/Microsoft.AspNet.Mvc.IntegrationTests/ModelPrefixSelectionIntegrationTest.cs
rynowak: I don't think that's an improvement. `Assert.Empty` only has meaning because string happens to implement IEnumerable. `Assert.Equal(string.Empty, modelBindingResult.Key);` is clearer

test/Microsoft.AspNet.Mvc.IntegrationTests/ModelPrefixSelectionIntegrationTest.cs
```diff
@@ -0,0 +1,394 @@
...
+        }
+
+        private class PurchaseOrder1
+        {
+            [FromHeader]
+            public string CustomerId { get; set; }
+        }
+
+        [Fact]
+        public async Task ComplexModel_WithGreedyProperties_PrefixSelectedByValueInValueProvider()
```
dougbu: Is anything at all different in the assertions between this group of tests and the `PersonX` ones? If yes, I suspect something is wrong.

Greedy properties should not be relevant since `IsTopLevelObject` will be `false` when they're checked. Would however see different behaviour with `parameter.BindingSource = BindingSource.Header;` Don't see any tests like that...

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/CompositeModelBinderTest.cs
dougbu: Right but that's not new. We should _never_ have fallen back unless `IsTopLevelObject == true`.

test/Microsoft.AspNet.Mvc.IntegrationTests/ModelPrefixSelectionIntegrationTest.cs
dougbu: Discussed offline. Integration tests for the relevant model binders already test the behaviour when the top-level `BindingSource` is greedy.

test/Microsoft.AspNet.Mvc.IntegrationTests/ModelPrefixSelectionIntegrationTest.cs
dougbu: New test looks good though obviously you have a bias against `BindingSource.Path`.

test/Microsoft.AspNet.Mvc.IntegrationTests/ModelPrefixSelectionIntegrationTest.cs
rynowak: Discussed in person, will delete these tests because nothing is actually different.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/CompositeModelBinderTest.cs
rynowak: Well, then we now have the correct behavior :fireworks:

dougbu: Discussed offline. `FallbackToEmptyPrefix` should never be copied into a child `ModelBindingContext` and therefore should never be `true` when `IsTopLevelObject` is `false`. No need to check `IsTopLevelObject` in `CompositeModelBinder`.

dougbu:
:shipit: once you've whacked the hidden gremlins and the tests are working again.

davidfowl:
This was merged right?

Issue #2757, opened 2015-06-29
Exception filters are not being short circuited in mono


sornaks:
https://github.com/aspnet/Mvc/blob/dev/test/Microsoft.AspNet.Mvc.FunctionalTests/FiltersTest.cs#L509

```
Assert.Equal() Failure
                ↓ (pos 0)
      Expected: 
      Actual:   Action Exception Filter
                ↑ (pos 0)
      Stack Trace:
           at Microsoft.AspNet.Mvc.FunctionalTests.FiltersTest+<ExceptionFilterShortCircuitsAnotherExceptionFilter>d__28.MoveNext () [0x00000] in <filename unknown>:0 
         --- End of stack trace from previous location where exception was thrown ---
           at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw () [0x00000] in <filename unknown>:0 
           at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (System.Threading.Tasks.Task task) [0x00000] in <filename unknown>:0 
           at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification (System.Threading.Tasks.Task task) [0x00000] in <filename unknown>:0 
           at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd (System.Threading.Tasks.Task task) [0x00000] in <filename unknown>:0 
           at System.Runtime.CompilerServices.TaskAwaiter.GetResult () [0x00000] in <filename unknown>:0 
         --- End of stack trace from previous location where exception was thrown ---
           at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw () [0x00000] in <filename unknown>:0 
           at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (System.Threading.Tasks.Task task) [0x00000] in <filename unknown>:0 
           at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification (System.Threading.Tasks.Task task) [0x00000] in <filename unknown>:0 
           at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd (System.Threading.Tasks.Task task) [0x00000] in <filename unknown>:0 
           at System.Runtime.CompilerServices.TaskAwaiter.GetResult () [0x00000] in <filename unknown>:0 
         --- End of stack trace from previous location where exception was thrown ---
[...]

PR #3051, opened 2015-08-28
Issue #2757 - Adding Order to exception filters so that they are executed in the same order irrespective of the platform.

sornaks:
#2757


rynowak:
:shipit:

sornaks:
https://github.com/aspnet/Mvc/commit/e8043f92e5436d7e327cf4ff223c5caece5e9b46

PR #3054, opened 2015-08-28
RazorViewEngine should cache transformed paths when view cannot be found

### LINKED ISSUE #3034: [PERF] RazorViewEngine should cache transformed paths when view cannot be found
`RazorViewEngine` does caching of the 'transformed' path each time a view is searched.

This means that the relationship between inputs and outputs like:
```
{
  action = "Index",
  controller = "Home",
}

"/Views/Home/Index.cshtml"
```

We don't cache the result when we **fail** to find a view. This means that we do a `string.Format` with each of the view location search strings every time we look for the view again.

This is a most serious performance issue than it sounds like because with `Html.EditorFor(...)` we end up looking for various editor templates, and the majority of the time these won't be found.

pranavkm:
Fixes #3034

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor/RazorViewEngine.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/RazorViewEngine.cs
@@ -216,9 +216,10 @@ namespace Microsoft.AspNet.Mvc.Razor
             }
         }
 
-        private RazorPageResult LocatePageFromViewLocations(ActionContext context,
-                                                            string pageName,
-                                                            bool isPartial)
+        private RazorPageResult LocatePageFromViewLocations(
+            ActionContext context,
+            string pageName,
+            bool isPartial)
         {
             // Initialize the dictionary for the typical case of having controller and action tokens.
             var areaName = GetNormalizedRouteValue(context, AreaKey);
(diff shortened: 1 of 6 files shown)


pranavkm:
cc @rynowak

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngine.cs
```diff
@@ -240,47 +241,65 @@ internal static string GetNormalizedRouteValue(ActionContext context, string key
...
-            if (!string.IsNullOrEmpty(pageLocation))
+            var cachedResult = _viewLocationCache.Get(expanderContext);
+            IEnumerable<string> locationsToSearch = null;
+
+            if (cachedResult != null)
             {
-                var page = _pageFactory.CreateInstance(pageLocation);
+                if (cachedResult.IsFoundResult)
+                {
+                    var page = _pageFactory.CreateInstance(cachedResult.ViewLocation);
```
pranavkm: We could look at all the views we previously looked at to see if they exist now. This would resolve the issue where a view with a better location appears since we last cached it.

rynowak: Why change this behavior though? This is

rynowak: Why change this behavior as part of this change though? This really seems like a separate issue that should get logged and triaged.

src/Microsoft.AspNet.Mvc.Razor/ViewLocationCacheResult.cs
```diff
@@ -0,0 +1,61 @@
...
+
+using System.Collections.Generic;
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.AspNet.Mvc.Razor
+{
+    /// <summary>
+    /// Result of <see cref="IViewLocationCache"/> lookups.
+    /// </summary>
+    public class ViewLocationCacheResult
```
rynowak: This should probably be a struct if we're going to keep it around in memory.

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngine.cs
pranavkm: Oh sure, I haven't changed that part. Just saying we could do that.

test/Microsoft.AspNet.Mvc.Razor.Test/RazorViewEngineTest.cs
```diff
@@ -507,6 +512,63 @@ public void FindView_LooksForViewsIfCachedViewDoesNotExist()
             expander.Verify();
         }
 
+        [Fact]
+        public void FindView_DoesNotInvokeExpandViewLocations_IfCacheEntryMatchesButViewIsNotFound()
```
rynowak: I'd add a note here that this validates an important perf concern. It's easy to revisit these tests in a while and thing "gee that's a complicted test... do we need to keep it?"

pranavkm:
Updated PR

rynowak:
:shipit:

Issue #3057, opened 2015-08-29
[Perf] Don't invoke async state machine unnecessarily


benaadams:
<del>ConfigureAwait in Async methods; unless only one await and last statement when return Task rather than awaiting.</del>

Return task rather than awaiting when no extra work needs to be done after await

pranavkm:
cc @davidfowl

PinpointTownes:
Curious: do you have actual numbers showing the performance improvement? 

DNX has no default synchronization context (well, it has no synchronization context *at all* :smile:), which means that task continuations will always be started on the current task scheduler (the default one if you don't override it) when using `ConfigureAwait(true)`. Yet, `Task.SetContinuationForAwait` has special logic to avoid a performance penalty when using the default task scheduler.

I'm pretty sure there's no real gain when using `ConfigureAwait(false)` in an environment that has no synchronization and uses the default task scheduler.

benaadams:
The ConfiguredAwait follows this path: 
```csharp
SetContinuationForAwait(Action, false, false, StackCrawlMark) 
-> AwaitTaskContinuation.UnsafeScheduleAction
-> new AwaitTaskContinuation(Action, false)
-> ThreadPool.UnsafeQueueCustomWorkItem
-> Action()
```

The non-configured await (default scheduler/no synchronization context) follows this path:
```csharp
SetContinuationForAwait(Action, true, true, StackCrawlMark) 
-> new AwaitTaskContinuation (Action, true, StackCrawlMark)
-> ExecutionContext.Capture 
-> ... bunch of stuff ...
-> AwaitTaskContinuation.Run( , false)
-> ThreadPool.UnsafeQueueCustomWorkItem
-> ExecutionContext.Run(ExecutionContext, GetInvokeActionCallback(), Action, true)
-> ... bunch of stuff ...
-> Action()
```

So the execution path of the non-configured await is more complicated; but perhaps more importantly is that asp.net is a library and its performance should be independent of the caller. If the calling code changes its scheduler or synchronisation context; without ConfigureAwait, the performance of asp.net could drop though the floor as the fast paths are no longer applicable. 

Quoting @stephentoub : http://blogs.msdn.com/b/pfxteam/archive/2012/06/15/executioncontext-vs-synchronizationcontext.aspx
> In fact, as I’ve stated in other posts, most library implementers should consider using ConfigureAwait(false) on every await of a task.

PinpointTownes:
[...]

PR #3058, opened 2015-08-29
ConfigureAwait (Mvc.Razor)

benaadams:
Ref #3057  ConfigureAwait in Async methods; unless only one await and last
statement when return Task rather than awaiting


benaadams:
Premise based on asp.net <= 4.5 but no longer applies to asp.net 5; will resubmit new PR with only asp.net related changes.

PR #3059, opened 2015-08-29
ConfigureAwait (Mvc.Core)

benaadams:
Ref #3057 ConfigureAwait in Async methods; unless only one await and last
statement when return Task rather than awaiting

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ControllerActionExecutor.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ControllerActionExecutor.cs
@@ -25,16 +25,16 @@ namespace Microsoft.AspNet.Mvc
             return CastToObject<T>(task);
         }
 
-        public static async Task<object> ExecuteAsync(
+        public static Task<object> ExecuteAsync(
             MethodInfo actionMethodInfo,
             object instance,
             IDictionary<string, object> actionArguments)
         {
             var orderedArguments = PrepareArguments(actionArguments, actionMethodInfo.GetParameters());
-            return await ExecuteAsync(actionMethodInfo, instance, orderedArguments);
+            return ExecuteAsync(actionMethodInfo, instance, orderedArguments);
         }
 
-        public static async Task<object> ExecuteAsync(
+        public static Task<object> ExecuteAsync(
             MethodInfo actionMethodInfo,
             object instance,
             object[] orderedActionArguments)
@@ -51,7 +51,7 @@ namespace Microsoft.AspNet.Mvc
                 exceptionDispatchInfo.Throw();
             }
 
-            return await CoerceResultToTaskAsync(
+            return CoerceResultToTaskAsync(
                 invocationResult,
                 actionMethodInfo.ReturnType,
                 actionMethodInfo.Name,
@@ -172,7 +172,7 @@ namespace Microsoft.AspNet.Mvc
         /// </summary>
         private static async Task<object> CastToObject(Task task)
         {
-            await task;
+            await task.ConfigureAwait(false);
             return null;
         }
 
@@ -181,7 +181,7 @@ namespace Microsoft.AspNet.Mvc
         /// </summary>
         private static async Task<object> CastToObject<T>(Task<T> task)
         {
-            return (object)await task;
+            return (object)await task.ConfigureAwait(false);
         }
 
         private static Type GetTaskInnerTypeOrNull(Type type)
--- a/src/Microsoft.AspNet.Mvc.Core/FileStreamResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/FileStreamResult.cs
@@ -68,13 +68,13 @@ namespace Microsoft.AspNet.Mvc
         }
 
         /// <inheritdoc />
-        protected async override Task WriteFileAsync(HttpResponse response, CancellationToken cancellation)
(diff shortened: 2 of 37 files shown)


src/Microsoft.AspNet.Mvc.Core/ControllerActionExecutor.cs
```diff
@@ -172,7 +172,7 @@ private static void ThrowIfWrappedTaskInstance(Type actualTypeReturned, string m
         /// </summary>
         private static async Task<object> CastToObject(Task task)
         {
-            await task;
+            await task.ConfigureAwait(false);
             return null;
```
benaadams: This code is a bit exotic; was going to do something different; but probably needs a double check instead?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/JQueryFormValueProviderFactory.cs
```diff
@@ -21,10 +21,12 @@ public class JQueryFormValueProviderFactory : IValueProviderFactory
 
             if (request.HasFormContentType)
             {
+                var culture = CultureInfo.CurrentCulture;
```
benaadams: Want to use thread culture here? Request culture, as per https://github.com/aspnet/Mvc/pull/3059/files#diff-fbc3dd5ae10027009ad0c93d673e1141R19 may be more relevant?

src/Microsoft.AspNet.Mvc.Core/VirtualFileProviderResult.cs
```diff
@@ -89,23 +89,19 @@ public string FileName
...
-
-                    return;
                 }
                 else
                 {
                     var fileStream = GetFileStream(fileInfo);
                     using (fileStream)
                     {
-                        await fileStream.CopyToAsync(response.Body, DefaultBufferSize, cancellation);
+                        return fileStream.CopyToAsync(response.Body, DefaultBufferSize, cancellation);
```
davidfowl: This is broken. Can't return directly if a using is there.

src/Microsoft.AspNet.Mvc.Core/StreamOutputFormatter.cs
```diff
@@ -40,7 +40,7 @@ public bool CanWriteResult([NotNull] OutputFormatterContext context, MediaTypeHe
                     response.ContentType = context.SelectedContentType.ToString();
                 }
 
-                await valueAsStream.CopyToAsync(response.Body);
+                return valueAsStream.CopyToAsync(response.Body);
```
davidfowl: This is broken. Can't return directly if a using is there.

src/Microsoft.AspNet.Mvc.Core/FileStreamResult.cs
```diff
@@ -68,13 +68,13 @@ public Stream FileStream
...
         /// <inheritdoc />
-        protected async override Task WriteFileAsync(HttpResponse response, CancellationToken cancellation)
+        protected override Task WriteFileAsync(HttpResponse response, CancellationToken cancellation)
         {
             var outputStream = response.Body;
 
             using (FileStream)
             {
-                await FileStream.CopyToAsync(outputStream, BufferSize, cancellation);
+                return FileStream.CopyToAsync(outputStream, BufferSize, cancellation);
```
davidfowl: This is broken. Can't return directly if a using is there.

src/Microsoft.AspNet.Mvc.Core/PhysicalFileProviderResult.cs
```diff
@@ -86,10 +84,9 @@ public string FileName
 
                 using (fileStream)
                 {
-                    await fileStream.CopyToAsync(response.Body, DefaultBufferSize, cancellation);
+                    return fileStream.CopyToAsync(response.Body, DefaultBufferSize, cancellation);
```
davidfowl: This is broken. Can't return directly if a using is there.

src/Microsoft.AspNet.Mvc.Core/VirtualFileProviderResult.cs
benaadams: Very good point; didn't think that through :(

src/Microsoft.AspNet.Mvc.Core/ModelBinding/FormValueProviderFactory.cs
```diff
@@ -20,7 +20,7 @@ public class FormValueProviderFactory : IValueProviderFactory
 
                 return new ReadableStringCollectionValueProvider(
                     BindingSource.Form,
-                    await request.ReadFormAsync(),
+                    await request.ReadFormAsync().ConfigureAwait(false),
                     culture);
```
benaadams: Culture call suggests is retrieved from request; but call gets it from current thread?

rynowak:
@benaadams 

What's the measured perf impact of these changes?

benaadams:
Need to wait till the myget feed updates

PinpointTownes:
@rynowak related discussion: https://github.com/aspnet/Mvc/issues/3057

benaadams:
Premise based on asp.net <= 4.5 but no longer applies to asp.net 5; will resubmit new PR with only asp.net 5 related changes.

PR #3067, opened 2015-08-31
Don't invoke async state machine unnecessarily

### LINKED ISSUE #3057: [Perf] Don't invoke async state machine unnecessarily

benaadams:
Return task rather than awaiting when no extra work needs to be done after await

Fixes #3057

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/FilterActionInvoker.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/FilterActionInvoker.cs
@@ -186,12 +186,12 @@ namespace Microsoft.AspNet.Mvc.Core
             return context.Results.Select(item => item.Filter).Where(filter => filter != null).ToArray();
         }
 
-        private async Task InvokeAllAuthorizationFiltersAsync()
+        private Task InvokeAllAuthorizationFiltersAsync()
         {
             _cursor.SetStage(FilterStage.AuthorizationFilters);
 
             _authorizationContext = new AuthorizationContext(ActionContext, _filters);
-            await InvokeAuthorizationFilterAsync();
+            return InvokeAuthorizationFilterAsync();
         }
 
         private async Task InvokeAuthorizationFilterAsync()
@@ -236,7 +236,7 @@ namespace Microsoft.AspNet.Mvc.Core
             }
         }
 
-        private async Task InvokeAllResourceFiltersAsync()
+        private Task InvokeAllResourceFiltersAsync()
         {
             _cursor.SetStage(FilterStage.ResourceFilters);
 
@@ -249,7 +249,7 @@ namespace Microsoft.AspNet.Mvc.Core
             context.ValueProviderFactories = new CopyOnWriteList<IValueProviderFactory>(_valueProviderFactories);
 
             _resourceExecutingContext = context;
-            await InvokeResourceFilterAsync();
+            return InvokeResourceFilterAsync();
         }
 
         private async Task<ResourceExecutedContext> InvokeResourceFilterAsync()
@@ -397,11 +397,11 @@ namespace Microsoft.AspNet.Mvc.Core
             return _resourceExecutedContext;
         }
 
-        private async Task InvokeAllExceptionFiltersAsync()
+        private Task InvokeAllExceptionFiltersAsync()
         {
             _cursor.SetStage(FilterStage.ExceptionFilters);
 
-            await InvokeExceptionFilterAsync();
+            return InvokeExceptionFilterAsync();
         }
 
         private async Task InvokeExceptionFilterAsync()
@@ -654,24 +654,9 @@ namespace Microsoft.AspNet.Mvc.Core
                 {
                     await item.FilterAsync.OnResultExecutionAsync(_resultExecutingContext, InvokeResultFilterAsync);
 
-                    if (_resultExecutedContext == null)
+                    if (_resultExecutedContext == null || _resultExecutingContext.Cancel == true)
                     {
-                        // Short-circuited by not calling next
-
-                        _logger.LogVerbose(ResourceFilterShortCircuitLogMessage, item.FilterAsync.GetType().FullName);
-
-                        _resultExecutedContext = new ResultExecutedContext(
-                            _resultExecutingContext,
-                            _filters,
-                            _resultExecutingContext.Result,
-                            Instance)
-                        {
-                            Canceled = true,
-                        };
-                    }
-                    else if (_resultExecutingContext.Cancel == true)
-                    {
-                        // Short-circuited by setting Cancel == true
+                        // Short-circuited by not calling next || Short-circuited by setting Cancel == true
 
                         _logger.LogVerbose(ResourceFilterShortCircuitLogMessage, item.FilterAsync.GetType().FullName);
 
--- a/src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
@@ -146,9 +146,9 @@ namespace Microsoft.AspNet.Mvc
             FlushInternal(true, true);
         }
 
-        public override async Task FlushAsync()
+        public override Task FlushAsync()
         {
-            await FlushInternalAsync(true, true);
+            return FlushInternalAsync(true, true);
         }
 
         // Do not flush the stream on Dispose, as this will cause response to be
--- a/src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
@@ -148,7 +148,7 @@ namespace Microsoft.AspNet.Mvc
 
         public override Task FlushAsync()
         {
-            return FlushInternalAsync(true, true);
+            return FlushInternalAsync(flushStream: true, flushEncoder: true);
         }
 
         // Do not flush the stream on Dispose, as this will cause response to be
--- a/src/Microsoft.AspNet.Mvc.Core/TaskCache.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/TaskCache.cs
@@ -20,7 +20,7 @@ namespace Microsoft.AspNet.Mvc
                 var completedTask = s_completedTask;
                 if (completedTask == null)
                 {
-                    completedTask = new Task(() => { }, default(CancellationToken), (TaskCreationOptions)0x4000 /*DoNotDispose*/); // benign initialization race condition
+                    completedTask = new Task(() => { }, default(CancellationToken)); // benign initialization race condition
                     completedTask.Start();
                     s_completedTask = completedTask;
                 }
--- a/src/Microsoft.AspNet.Mvc.Core/TaskCache.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/TaskCache.cs
@@ -1,16 +1,14 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System.Threading;
 using System.Threading.Tasks;
 
 namespace Microsoft.AspNet.Mvc
 {
     public static class TaskCache
     {
-#if DNX451 || DNX452
-        /// <summary>A task that's already been completed successfully.</summary>
-        private static Task s_completedTask;
+#if DNX451
+        static Task _completedTask = Task.FromResult(0);
 #endif
 
         /// <summary>Gets a task that's already been completed successfully.</summary>
@@ -19,15 +17,8 @@ namespace Microsoft.AspNet.Mvc
         {
             get
             {
-#if DNX451 || DNX452
-                var completedTask = s_completedTask;
-                if (completedTask == null)
-                {
-                    completedTask = new Task(() => { }, default(CancellationToken)); // benign initialization race condition
-                    completedTask.Start();
-                    s_completedTask = completedTask;
-                }
-                return completedTask;
+#if DNX451
+                return _completedTask;
 #else
                 return Task.CompletedTask;
 #endif
--- a/src/Microsoft.AspNet.Mvc.Core/TaskCache.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/TaskCache.cs
@@ -1,4 +1,7 @@
-﻿using System.Threading;
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace Microsoft.AspNet.Mvc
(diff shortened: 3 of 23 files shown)


benaadams:
No fair :( Went French?

```
Microsoft.AspNet.Mvc.FunctionalTests.ViewEngineTests.RazorViewEngine_UsesExpandersForLayouts(value: "!-invalid-!", expected: "<language-layout>\nView With Layout\n</language-layo"...) [FAIL]
      Assert.Equal() Failure
                 ↓ (pos 1)
      Expected: <language-layout>\nView With Layout\n</langu···
      Actual:   <fr-language-layout>\nView With Layout\n</fr···
                 ↑ (pos 1)
```

dougbu:
> No fair :( Went French?

@kirthik ?

benaadams:
Is same error with main build rather than this PR; so feel better :)

src/Microsoft.AspNet.Mvc.Core/TaskCache.cs
```diff
@@ -0,0 +1,38 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNet.Mvc
+{
+    public static class TaskCache
+    {
+#if DNX451 || DNX452
```
davidfowl: Too complicated. Just do `Task.FromResult(0)`

src/Microsoft.AspNet.Mvc.Core/TaskCache.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// <summary>A task that's already been completed successfully.</summary>
+        private static Task s_completedTask;
```
davidfowl: We also don't `s_`

benaadams: Dropped

NeelBhatt: standard way might be 

```private static Task _sCompletedTask;```

or 

```private static Task _completedTask;```

**Comment by rynowak** [comment]
@benaadams 

What is the measured performance improvement of this change?

File: src/Microsoft.AspNet.Mvc.Core/TaskCache.cs (hunk 2 above)
Comment (NeelBhatt) [inline src/Microsoft.AspNet.Mvc.Core/TaskCache.cs]: As i just saw you have used ```static readonly``` so standard way would be :

 ```static readonly Task CompletedTask = Task.FromResult(0);```

Comment (benaadams) [inline src/Microsoft.AspNet.Mvc.Core/TaskCache.cs]: Then it would conflict with the getter name?

Comment (NeelBhatt) [inline src/Microsoft.AspNet.Mvc.Core/TaskCache.cs]: You can put it as it is now or change some names. 

The reason i suggested was just because it looks more readable!

File: src/Microsoft.AspNet.Mvc.Core/TaskCache.cs (hunk 1 above)
Comment (NeelBhatt) [inline src/Microsoft.AspNet.Mvc.Core/TaskCache.cs]: Something is wrong here: :https://github.com/aspnet/Mvc/blob/9915ea38097857cc8046b9cceaec7ad6bd0b9b11/test/Microsoft.AspNet.Mvc.FunctionalTests/ViewEngineTests.cs 

for this test:

```RazorViewEngine_UsesExpandersForLayouts```

File: src/Microsoft.AspNet.Mvc.Core/FilterActionInvoker.cs [hunk 3]
```diff
@@ -654,24 +654,9 @@ private IFilterMetadata[] GetFilters()
...
-                            _resultExecutingContext.Result,
-                            Instance)
-                        {
-                            Canceled = true,
-                        };
-                    }
-                    else if (_resultExecutingContext.Cancel == true)
-                    {
-                        // Short-circuited by setting Cancel == true
+                        // Short-circuited by not calling next || Short-circuited by setting Cancel == true
```
rynowak: If these blocks really have the same code we should look at cleaning that up across the board, there are probably 2-3 more cases in the invoker.

benaadams:
@rynowak quick slightly unscientific test using ```AsyncAction_TaskOfValueReturnType```

Parallelism | [Awaiting](https://github.com/benaadams/Mvc/commit/91578e6384ee5e0d0d688c7de3bed29787cb05cd) (ms) | [Task Passthough](https://github.com/benaadams/Mvc/commit/6e7189f235e01d92c97f37fbbc82a20d5fcf0e6a) (ms) | Slow down from awaiting
------------ | ------------ | ------------ | ------------- | -------------
1 | 1476.1121  | 1318.694 | 12%
16 | 8819.8558 | 8020.9328 | 10%
64 | 35184.4027 | 32142.7115 | 9%

So making the methods as ```async``` and having the compiler construct a state machine to ```await``` a ```Task``` which can just be returned adds 10% to the time for this path. Obviously the time will vary greatly depending on how much other work is being done in the methods - but it does show that it makes a difference.

benaadams:
@rynowak I'll run some tests on the invoker in the morning; will need to chase down the code paths. Unless you have an existing test that would be good to look at?

benaadams:
Any particular invoker? For ```ControllerActionInvoker``` could change ```InvokeActionAsync``` to use a continuation and return the task to see what effect that has?

rynowak:
@benadams 

It would be beneficial to look at the effect on throughput (requests/sec) for a "Hello, World" type of app like https://github.com/rynowak/mvc-sandbox/tree/master/src/HelloWorldMvc

src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
```diff
@@ -146,9 +146,9 @@ public override void Flush()
             FlushInternal(true, true);
         }
 
-        public override async Task FlushAsync()
+        public override Task FlushAsync()
         {
-            await FlushInternalAsync(true, true);
+            return FlushInternalAsync(true, true);
```
pranavkm: While you're changing this line of code, could you also add named parameters to the boolean parameters?

src/Microsoft.AspNet.Mvc.Core/TaskCache.cs
```diff
@@ -0,0 +1,29 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNet.Mvc
+{
+    public static class TaskCache
+    {
+#if DNX451
+        static readonly Task _completedTask = Task.FromResult(0);
```
pranavkm: cc @Eilon

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
```diff
@@ -1099,7 +1099,7 @@ public virtual Task<bool> TryUpdateModelAsync<TModel>([NotNull] TModel model)
         /// </param>
         /// <returns>A <see cref="Task"/> that on completion returns <c>true</c> if the update is successful.</returns>
         [NonAction]
-        public virtual async Task<bool> TryUpdateModelAsync<TModel>([NotNull] TModel model,
+        public virtual Task<bool> TryUpdateModelAsync<TModel>([NotNull] TModel model,
```
pranavkm: Formatting:

```C#
public virtual Task<bool> TryUpdateModelAsync<TModel>(
    [NotNull] TModel model,
    [NotNull] string prefix)
{
```

pranavkm: Same for other methods in this type.

benaadams:
@rynowak bit of a rabbit hole... post change:

image

Looking at ExecuteResultAsync it spends 47% doing WriteAsync; and 32% of the time deciding what formatter to use based on the Accept header.

image

Which between them are 69% of InvokeAllResultFiltersAsync

image

InvokeAllResultFiltersAsync (43%) and InvokeAllExecptionFiltersAsync (36%) make up 80% of FilterActionInvoker<InvokeResourceFilterAsync>

image

Probably could do with some caching of return types; though I don't know how this would work with reloading based of file changes (changing views?)

image

Looking at the 35% is lost between RouterMiddleware.Invoke and FilterActionInvoker<InvokeResourceFilterAsync>

image

If you are looking at method type caching after a quick review of ViewComponentMethodSelector I see there is some work with expressions; I've had quite good success with wrapping a Expression.Call in a strongly typed Expression.Lambda and compiling, and then caching the Func in a dictionary which may or may not help (ignoring the technical debt in the inheritance):

````csharp
public class Controller<T> where T : Controller<T>, new()
{
    ...
    static Controller()
    {
        Routes = new Dictionary<string, Func<T, ApiRequest, Task<byte[]>>>();
        var methods = typeof(T).GetMethods(BindingFlags.Public | BindingFlags.Instance);

        foreach (var method in methods)
        {
            if (!method.ReturnType.Equals(typeof(Task<byte[]>)))
            {
                continue;
            }
            var obj = Expression.Parameter(typeof(T), "obj");
            var data = Expression.Parameter(typeof(ApiRequest), "data");

            var function =
                Expression.Lambda<Func<T, ApiRequest, Task<byte[]>>>(
                    Expression.Call(obj, method, data), obj, data
                ).Compile();

            Routes.Add(method.Name.ToLowerInvariant(), function);
            ...
    }

    public Lazy<CustomerIdentity> Customer { get; private set; }
````
Then invoke with a straight dictionary lookup:
````csharp
public static Task<byte[]> Route(Lazy<CustomerIdentity> customer, string action, ApiRequest data)
{
    var obj = new T() { Customer = customer };
    // check if exists etc.
    return Routes[action](obj, data);
}
````
With this approach I can resolve and invoke 5.3M completed task controller action routes per second on a single thread and 322k controllers creating and converting POCO types->Json->211 byte[] via [Jil](https://github.com/kevin-montrose/Jil) on a single thread; all awaited on each invocation.

You'd probably need to bucket by argument count; +key with argument types etc; and the Mvc's route matching is a lot more complicated (specific vs general use cases) so it might not work at all...

Comparing with before:

image

The mscorlib await points are cut down from 11 to 7 for a shorter call stack; probably not a huge perf improvement, but is a start.

**Comment by benaadams** [comment]
Over 1,000,000 requests (via ab -c 64 -k on localhost) its 7828.62 rps vs 7789.52 rps so only a 0.5% improvement :(

Though ab isn't a very good load tester; especially running on same machine.

**Comment by benaadams** [comment]
I think my computer may be unhappy; likely too much network diagnostics entry points hanging around; will reboot and try again

**Comment by rynowak** [comment]
> Over 1,000,000 requests (via ab -c 64 -k on localhost) its 7828.62 rps vs 7789.52 rps so only a 0.5% improvement :( I think you have bigger issues going on...

That's the measurement I was looking for. I suspect that the impact of this isn't really going to noticeable in any real scenario other than benchmarking. I'm fine with taking the changes as the perf can't be worse, and the changes don't make the code worse :laughing: 

Just be aware that micro-optimizations like this aren't something the team is likely to spend more time on unless our other measurements show us that it's significant. For instance, I'm working a change now to implement the cached Task.FromResult pattern from model binding where I did find that it's a significant source of allocations. If anything it's a worthwhile exercise for us to see what kind of impact these changes have - knowledge is power!


Regarding this discussion: 

> If you are looking at method type caching after a quick review of ViewComponentMethodSelector I see there is some work with expressions; I've had quite good success with wrapping a Expression.Call in a strongly typed Expression.Lambda and compiling, and then caching the Func in a dictionary which may or may not help (ignoring the technical debt in the inheritance):

Do you mean the code for invoking the action-method itself? If there's a pattern that's better than what we're doing it's likely worth looking into. Let's take that discussion to another thread/issue.

File: Mvc.sln [hunk 7]
```diff
@@ -187,6 +187,9 @@ EndProject
 Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.AspNet.Mvc.Localization.Test", "test\Microsoft.AspNet.Mvc.Localization.Test\Microsoft.AspNet.Mvc.Localization.Test.xproj", "{8FC726B5-E766-44E0-8B38-1313B6D8D9A7}"
 EndProject
 Global
+	GlobalSection(Performance) = preSolution
+		HasPerformanceSessions = true
+	EndGlobalSection
```
rynowak: eww gross

rynowak:
:shipit: - will merge this.

rynowak:
Also regarding this:

> I think you have bigger issues going on...

We're definitely aware of that. I'm working through some of the issues right now with allocations in routing + model binding. I wanted to give your changes the best chance to show an improvement, but once you get out of "Hello, World!" type scenarios the views or model binding start to really dominate the measurements.

We're only really scratched the surface at this point with measuring MVC.

Mvc.sln
benaadams: Lol sorry :)

PinpointTownes:
It's probably too late to comment this PR (since it has already been merged), but replacing async state machines by simple `Task`-returning methods is not always harmless, specially on paths susceptible to throw an exception:

```csharp
public void Throw() {
    throw new Exception();
}
```

```csharp
public async Task<int> Method1() {
    Throw();
    return 10;
}
```

```csharp
public Task<int> Method2() {
    Throw();
    return Task.FromResult(10);
}
```

These 2 `Task`-returning methods share the same signature and - in a way - could be considered as equivalent.

There's a subtle difference though: `Method1` will actually throw the exception differently from `Method2`:
* In the first method, the exception will be thrown "asynchronously": `Method1` will immediately return a `Task` that will latter fail with `Task.Status = TaskStatus.Faulted` (that would be the equivalent of using `return Task.FromException(new Exception())`).

* In the second one, it will be thrown in a synchronous way: `Method2` will immediately crash and won't - obviously - return any task.

This may have practical consequences. For instance, this method will behave differently depending on the method invoked:

```csharp
public void Method3() {
    // This method returns a task that will latter fail but throws no exception at this point.
    var call1 = Method1();

    // This method will immediately throw an exception and you'll have no chance to get the task. The thread will crash here.
    var call2 = Method2();
}
```

In practice, awaiting **and** wrapping these 2 `Task`-returning methods in a try/catch block should produce the same result, but I think it's worth double-checking there's no regression in this PR.

benaadams:
> replacing async state machines by simple Task-returning methods is not always harmless

Agreed; though I didn't come across any generate ```Task``` now resolve later parallelism or out-of-order execution which would be effected by this. 

The execution seems to follow a serial path rather than gather+collate/map+reduce; which make sense both for caller expectation (pipeline executing steps in order) and optimising parallel requests; rather than parallelism within a single request as all the async resources are server local.

Alas ```return Task.FromException(new Exception())``` doesn't exist in 4.5.1 (though is in 4.6 and core)

benaadams:
Perf testing is a bit hard; added #3070

rynowak:
@PinpointTownes 

Put a bit more simply than what Ben said, we're awaiting all of these tasks already. If we had any code that was doing something fancy, I'd be concerned.

PR #3069, opened 2015-09-01
Make ModelBindingResult a struct.

rynowak:


src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingResult.cs
```diff
@@ -76,5 +84,54 @@ public ModelBindingResult(object model, string key, bool isModelSet, ModelValida
...
+            }
+            else
+            {
+                return Equals(other.Value);
+            }
+        }
+
+        public override int GetHashCode()
+        {
+            if (Key == null)
```
davidfowl: hash code combiner?

davidfowl:
Motivation behind this change?

rynowak:
> Motivation behind this change?

Make model binding as close to allocation-free as possible. I've done extensive profiling here the amount of allocations we do is significant.

davidfowl:
Can you share numbers?

rynowak:
From profiling I did last week on this app: https://github.com/rynowak/mvc-sandbox/tree/master/src/BasicViews

Some line-items highlighting modelbinding/validation, measurements in allocations over 10000 requests

Type | Allocated bytes | Allocated objects
-------|---------------------|---------------------
`System.Threading.Tasks.Task<ModelBindingResult>` | 17,160,000 | 390,000
`Microsoft.AspNet.Mvc.ModelBinding.ModelBindingContext` | 10,400,000 | 200,000
`Microsoft.AspNet.Mvc.ModelBinding.BindingSourceModelBinder+<BindModelAsync>d__5` | 9,880,000 | 190,000
`Microsoft.AspNet.Mvc.ModelBinding.ModelStateDictionary+<FindKeysWithPrefix>d__63` | 10,240,000 | 160,000
`Microsoft.AspNet.Mvc.ModelBinding.ModelBindingResult` | 3,840,000 | 160000
`System.Collections.Generic.Dictionary+Enumerator<String, ModelState>` | 3,840,000 |120,000
`Microsoft.AspNet.Mvc.ModelBinding.PrefixContainer+PrefixComparer` | 1,440,000 |120,000
`Microsoft.AspNet.Mvc.ModelBinding.BinderTypeBasedModelBinder+<BindModelAsync>d__2` | 5,760,000 | 90,000
`Microsoft.AspNet.Mvc.ModelBinding.CompositeModelBinder+<TryBind>d__5` | 4,680,000 | 90,000
`System.Collections.Generic.List<ModelValidationNode>` | 1,920,000 | 80,000
`Microsoft.AspNet.Mvc.ModelBinding.Validation.ModelValidationContext` | 1,920,000 | 80,000
`Microsoft.AspNet.Mvc.ModelBinding.CompositeModelBinder+<BindModelAsync>d__4` | 4,800,000 | 80,000
`Microsoft.AspNet.Mvc.ModelBinding.ModelValidationNode` | 2,240,000 | 80,000

rynowak:
Getting rid of `ModelBindingContext` and `ModelValidationNode` allocations is also on my to-do list.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingResult.cs
```diff
@@ -76,5 +84,54 @@ public ModelBindingResult(object model, string key, bool isModelSet, ModelValida
...
+            {
+                return 0;
+            }
+            else if (IsModelSet)
+            {
+                Debug.Assert(Key != null);
```
dougbu: Why assert that you're in an `else if` block? Could see asserting `!IsModelSet` in the previous block and `Model == null` here but that's about it.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingResult.cs
```diff
@@ -1,47 +1,55 @@
...
+using System.Diagnostics;
+using System.Threading.Tasks;
+
 namespace Microsoft.AspNet.Mvc.ModelBinding
 {
     /// <summary>
     /// Contains the result of model binding.
     /// </summary>
-    public class ModelBindingResult
+    public struct ModelBindingResult : IEquatable<ModelBindingResult>
```
dougbu: Class needs lots of docs.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingResult.cs
```diff
@@ -1,47 +1,55 @@
...
-        /// Creates a new <see cref="ModelBindingResult"/>.
-        /// </summary>
-        /// <param name="model">The model which was created by the <see cref="IModelBinder"/>.</param>
-        /// <param name="key">The key using which was used to attempt binding the model.</param>
-        /// <param name="isModelSet">A value that represents if the model has been set by the
-        /// <see cref="IModelBinder"/>.</param>
-        /// <param name="validationNode">A <see cref="ModelValidationNode"/> which captures the validation information.
-        /// </param>
-        public ModelBindingResult(object model, string key, bool isModelSet, ModelValidationNode validationNode)
+        public static ModelBindingResult Success(string key, object model)
```
dougbu: Should not be an encouraged pattern and would prefer if `validationNode: null` were explicit in the calling code. That is, nuke this overload and the shorter `SuccessAsync()`.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingResult.cs
```diff
@@ -76,5 +84,54 @@ public ModelBindingResult(object model, string key, bool isModelSet, ModelValida
...
+            {
+                Debug.Assert(Key != null);
+                return Key.GetHashCode();
+            }
+        }
+
+        public bool Equals(ModelBindingResult other)
+        {
+            return
+                string.Equals(Key, other.Key, StringComparison.Ordinal) &&
```
dougbu: Not sure about this. The comparable `ModelStateDictionary` keys are compared using `OrdinalIgnoreCase`. On the other hand, we don't have cases in existing code where `ModelBindingResult.Key` is ever compared (tests and that's it).

rynowak: Good catch, easy fix

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingResult.cs
rynowak: Agreed, let's nuke these and the 4-arg overload.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CompositeModelBinder.cs
```diff
@@ -77,8 +77,8 @@ public CompositeModelBinder([NotNull] IEnumerable<IModelBinder> modelBinders)
             }
 
             return new ModelBindingResult(
-                modelBindingResult.Model,
                 bindingKey,
+                modelBindingResult.Model,
                 modelBindingResult.IsModelSet,
                 modelBindingResult.ValidationNode);
```
dougbu: The other option to what we discussed about the fate of this constructor would be a copy constructor that takes `newModelName` and `dearOldResult`.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
```diff
@@ -78,7 +78,7 @@ public class DictionaryModelBinder<TKey, TValue> : CollectionModelBinder<KeyValu
                 var valueResult = await modelBinder.BindModelAsync(valueBindingContext);
 
                 // Always add an entry to the dictionary but validate only if binding was successful.
-                model[convertedKey] = ModelBindingHelper.CastOrDefault<TValue>(valueResult?.Model);
+                model[convertedKey] = ModelBindingHelper.CastOrDefault<TValue>(valueResult.Model);
                 if (valueResult != null && valueResult.IsModelSet)
```
dougbu: ~~`valueResult != null && `~~

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
```diff
@@ -425,21 +421,21 @@ internal static PropertyValidationInfo GetPropertyValidationInfo(ModelBindingCon
...
-                    validationNode.ChildNodes.Add(propertyValidationNode);
+                var propertyValidationNode = result.ValidationNode;
+                if (propertyValidationNode == null)
+                {
+                    // Make sure that irrespective of whether the properties of the model were bound with a value,
+                    // create a validation node so that these get validated.
+                    propertyValidationNode = new ModelValidationNode(result.Key, entry.Key, result.Model);
                 }
+
+                validationNode.ChildNodes.Add(propertyValidationNode);
```
dougbu: This block has different semantics than it did. Did not previously add `ModelValidationNode`s when no data existed at all (despite the comment just above). Suspect `result == ModelBindingResult.NoResult` should be handled separately from `!result.IsModelSet`.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/KeyValuePairModelBinderTest.cs
```diff
@@ -79,7 +78,7 @@ public class KeyValuePairModelBinderTest
             var result = await binder.BindModelAsync(bindingContext);
 
             // Assert
-            Assert.Null(result);
+            Assert.False(result.IsModelSet);
```
dougbu: Odd. Why isn't this one `Assert.Equal(ModelBindingResult.NoResult, result);`?

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/SimpleTypeModelBinderTest.cs
```diff
@@ -42,7 +42,7 @@ public class SimpleTypeModelBinderTest
         [InlineData(typeof(DateTimeOffset))]
         [InlineData(typeof(double))]
         [InlineData(typeof(DayOfWeek))]
-        public async Task BindModel_ReturnsNotNull_IfTypeCanBeConverted(Type destinationType)
+        public async Task BindModel_ReturnsFailure_IfTypeCanBeConverted(Type destinationType)
```
dougbu: perhaps `...IfTypeCanBeConvertedAndValueCannot` ??

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/SimpleTypeModelBinderTest.cs
```diff
@@ -171,11 +171,11 @@ public class SimpleTypeModelBinderTest
             var binder = new SimpleTypeModelBinder();
 
             // Act
-            var retVal = await binder.BindModelAsync(bindingContext);
+            var result = await binder.BindModelAsync(bindingContext);
 
             // Assert
-            Assert.NotNull(retVal);
-            Assert.Equal(42, retVal.Model);
+            Assert.NotNull(result);
```
dougbu: Either remove or compare to `NoResult`.

test/Microsoft.AspNet.Mvc.IntegrationTests/BinderTypeBasedModelBinderIntegrationTest.cs
```diff
@@ -71,9 +71,7 @@ public class BinderTypeBasedModelBinderIntegrationTest
             var modelBindingResult = await argumentBinder.BindModelAsync(parameter, modelState, operationContext);
 
             // Assert
-
-            // ModelBindingResult
-            Assert.Null(modelBindingResult);
+            Assert.False(modelBindingResult.IsModelSet);
```
dougbu: If this should be `NoResult`, check for that instead.

dougbu: Similar point a few more times in this class. Failures to find data or an appropriate binder aren't quite the same as other failures to bind.

dougbu:
Design looks good with points above and discussed offline; :watch:-ing for the non-design iteration...

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingResult.cs
rynowak: :+1: groovy

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingResult.cs
rynowak: This was replaced with `HashCodeCombiner`

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingResult.cs
rynowak: Added

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CompositeModelBinder.cs
rynowak: Added

src/Microsoft.AspNet.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
rynowak: Fixed

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
rynowak: Right on.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/KeyValuePairModelBinderTest.cs
rynowak: Yeah, looks like I just missed this one.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/SimpleTypeModelBinderTest.cs
rynowak: Love it.

test/Microsoft.AspNet.Mvc.IntegrationTests/BinderTypeBasedModelBinderIntegrationTest.cs
rynowak: yeah, I must have missed a few of these.

rynowak:
@dougbu updated and removed design tag

I recommend looking at individual commits. 439cbbc is a big mechanical change.

dougbu:
> Failed (IsModelSet == true)

Don't forget to make this `false` when squashing your commits and re-describing things.

dougbu:
:shipit: after addressing doc comment comments, updating commit description and seeing clean CI builds (looks like `dev` may be :ok: now or soon).

rynowak:
> What valid non-test success scenario involves a null validationNode?

We wrote product code to correctly handle this case **all over**. Considering that my next change is to nuke MVN for orbit, I'm not too concerned about doing the cleanup right now.

PR #3074, opened 2015-09-01
[Fixes #3016] Disable response buffering in places where the content is already built/available

### LINKED ISSUE #3016: Disable response buffering in places where the content is already built/available.
We should check for the presence of `IHttpBufferingFeature` on the context and if present should disable response buffering at least in the following cases:

- https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/FileStreamResult.cs#L77
- https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/StreamOutputFormatter.cs#L43
- https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/ContentResult.cs#L70
- https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/FileContentResult.cs#L69

@Tratcher @rynowak

kichalla:
@rynowak @Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ActionResults/ContentResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ActionResults/ContentResult.cs
@@ -4,7 +4,7 @@
 using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Http;
-using Microsoft.AspNet.Mvc.Actions;
+using Microsoft.AspNet.Http.Features;
 using Microsoft.AspNet.Mvc.Internal;
 using Microsoft.Framework.Internal;
 using Microsoft.Net.Http.Headers;
@@ -44,7 +44,7 @@ namespace Microsoft.AspNet.Mvc.ActionResults
                 contentTypeHeader = contentTypeHeader.Copy();
                 contentTypeHeader.Encoding = Encoding.UTF8;
             }
-            
+
             response.ContentType = contentTypeHeader?.ToString()
                 ?? response.ContentType
                 ?? DefaultContentType.ToString();
@@ -56,8 +56,12 @@ namespace Microsoft.AspNet.Mvc.ActionResults
 
             if (Content != null)
             {
+                var bufferingFeature = response.HttpContext.Features.Get<IHttpBufferingFeature>();
+                bufferingFeature?.DisableResponseBuffering();
+
                 return response.WriteAsync(Content, contentTypeHeader?.Encoding ?? DefaultContentType.Encoding);
             }
+
             return TaskCache.CompletedTask;
         }
     }
--- a/src/Microsoft.AspNet.Mvc.Core/ActionResults/FileContentResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ActionResults/FileContentResult.cs
@@ -5,6 +5,7 @@ using System;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Http.Features;
 using Microsoft.Framework.Internal;
 using Microsoft.Net.Http.Headers;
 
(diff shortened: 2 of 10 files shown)


src/Microsoft.AspNet.Mvc.Core/ContentResult.cs
```diff
@@ -54,6 +55,12 @@ public class ContentResult : ActionResult
 
             if (Content != null)
             {
+                var bufferingFeature = response.HttpContext.GetFeature<IHttpBufferingFeature>();
+                if (bufferingFeature != null)
```
pranavkm: `bufferingFeature?.DisableResponseBuffering();`

pranavkm:
Could this also apply to Razor views?

src/Microsoft.AspNet.Mvc.Core/ContentResult.cs
kichalla: :+1:

kichalla:
> Could this also apply to Razor views?

Currently I am disabling buffering only for scenarios where the content is already available. For  razor views or formatter serialization the final content is not available yet and the scenario is that one might want to enable buffering to catch any exceptions that happen during the generation of the final content in order to send back a friendly error message instead of partial content (just fyi...this is similar to web api 2 behavior).

src/Microsoft.AspNet.Mvc.Core/ContentResult.cs
rynowak: PranavPoints++

test/Microsoft.AspNet.Mvc.Core.Test/ContentResultTest.cs
```diff
@@ -183,9 +213,33 @@ private static ActionContext GetActionContext(HttpContext httpContext)
...
+
+            if (addBufferingFeature)
+            {
+                context.SetFeature<IHttpBufferingFeature>(new TestBufferingFeature());
+            }
+
+            return context;
+        }
+
+        private class TestBufferingFeature : IHttpBufferingFeature
```
rynowak: IMO just use Moq or lift this into a top-level class so you can reuse it.

kichalla:
Updated. cc: @pranavkm @rynowak

test/Microsoft.AspNet.Mvc.Core.Test/ContentResultTest.cs
```diff
@@ -78,6 +80,34 @@ public class ContentResultTest
             Assert.Equal("text/plain; charset=utf-7", httpContext.Response.ContentType);
         }
 
+        [Fact]
+        public async Task ContentResult_DisablesResponseBuffering_IfBufferingFeatureAvailable()
```
pranavkm: Is there a complementary scenario `IfBufferingFeatureIsNotAvailable` already covered? If not, could we add one?

test/Microsoft.AspNet.Mvc.Core.Test/ContentResultTest.cs
```diff
@@ -183,9 +213,16 @@ private static ActionContext GetActionContext(HttpContext httpContext)
                                     new ActionDescriptor());
         }
 
-        private static HttpContext GetHttpContext()
+        private static HttpContext GetHttpContext(bool addBufferingFeature = false)
```
rynowak: Another nitpick, sorry for not noticing. You might want to just put this inline in your tests that use the feature. That makes it much more apparent what you're doing.

rynowak:
:shipit:

pranavkm:
:shipit:

test/Microsoft.AspNet.Mvc.Core.Test/ContentResultTest.cs
kichalla: I did not add it explicitly as buffering is not *on* by default and all the remaining tests run without it.

PR #3081, opened 2015-09-02
[Fixes #2929] Moving IsLocalUrl implementation to UrlHelper

### LINKED ISSUE #2929: IsLocalUrl is not what I need
The implementation should be dynamic, or it should be renamed `IsRelativeUrl`, because right now it is not checking if the url is local, just relative, which is a sub-case of local.

It needs at least to check the `HOST` header, and at best a set of custom authority I could pass as a parameter. Ideally it would be a service that I can re-implement. 

Also, please explain what this sentence means, or just remove it from the comment 
> An URL with an absolute path is considered local if it does not have a host/authority part

sebastienros:
Removed `UrlUtility` class and placed the code for `IsLocalUrl` in `UrlHelper` directly so that it can be customized.


src/Microsoft.AspNet.Mvc.Core/Internal/UrlUtility.cs
```diff
@@ -1,39 +0,0 @@
...
-        /// ~/Index.html
-        /// </para>
-        /// <para>
-        /// The following URLs are non-local:
-        /// ../Index.html
-        /// http://www.contoso.com/
-        /// http://localhost/Index.html
-        /// </para>
-        /// </example>
-        public static bool IsLocalUrl(string url)
```
rynowak: Keep the documentation

sebastienros: The `UrlHelper` class inherits its documentation from `IUrlHelper` which already contains the same thing.

rynowak: Cool, sounds good then.

rynowak:
:shipit: 

Are you familiar with our rebase/merge workflow?

rynowak:
@Eilon - why is this guy showing up as cla-required? Is he trustworthy?

sebastienros:
Merged

Eilon:
Trust no one.

PR #3086, opened 2015-09-03
Disable Options order functional test

HaoK:
@rynowak Not sure if these should be reworked or just removed, I just commented them out for now


rynowak:
These can safely be removed. As you said in the comments, it doesn't really apply anymore.

HaoK:
Cool, I'll nuke them

PR #3089, opened 2015-09-03
Get four failing tests working on Linux

dougbu:
- use consistent line endings on all platforms so that hashes match
  - add another `.gitattribute` file
  - move files to avoid required index recreation in all Linux repos


dougbu:
Let's watch Travis before discussing merits of this approach.

dougbu:
Have at reviewing this PR. Functional tests now clean with Travis.

Not sure why `GetCompilationFailedResult_ReturnsCompilationResult_WithGroupedMessages()` fails in one of the Travis builds but have seen similar failures in other PRs.

NTaylorMullen:
If you rebase `GetCompilationFailedResult_ReturnsCompilationResult_WithGroupedMessages` should pass.

NTaylorMullen:
I don't entirely follow how your change fixes Linux. Out of curiosity could you explain what was wrong/how moving the files and defaulting to crlf fixed the issue?

dougbu:
> what was wrong/how moving the files and defaulting to crlf fixed the issue?

The sordid details:
- Functional tests fetching views containing `<link>` and `<script>` tag helpers using `asp-append-version="true"` were failing on Linux.
- Nearly identical tests fetching views containing `` tag helpers were _not_ failing on Linux.
- The failures were due to different SHA256 hashes of the files and therefore different `?v=someHash` portions of the generated query strings.
- The different hashes were due to different line endings on Linux.
- `` wasn't affected because the files are binary from `git`'s perspective.

Fix is to give the few relevant files the same line endings on all systems. Their content isn't read except for the hashes.

Only "gotcha" was making sure the change affects all Linux repos. Chose to move the files so they are also included in the commit. Could have edited them all but that would have made the PR less clear.

NTaylorMullen:
Ahhh :+1:. Thank you for the clarification. :shipit:

dougbu:
dc4573e

PR #3090, opened 2015-09-04
Test Cleanup: Fix Assert.NotNulls with ModelBindingResult

rynowak:
MBR is a struct now, updated tests and names to reflect this. These were missed in an earlier pass.

### DIFF
--- a/test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/BindingSourceModelBinderTest.cs
+++ b/test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/BindingSourceModelBinderTest.cs
@@ -35,8 +35,11 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         public async Task BindingSourceModelBinder_ReturnsNull_WithNoSource()
         {
             // Arrange
-            var context = new ModelBindingContext();
-            context.ModelMetadata = new EmptyModelMetadataProvider().GetMetadataForType(typeof(string));
+            var context = new ModelBindingContext()
+            {
+                ModelMetadata = new EmptyModelMetadataProvider().GetMetadataForType(typeof(string)),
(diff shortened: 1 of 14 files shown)


test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/BindingSourceModelBinderTest.cs
```diff
@@ -90,7 +90,7 @@ public void BindingSourceModelBinder_ThrowsOnNonGreedySource()
             var result = await binder.BindModelAsync(context);
 
             // Assert
-            Assert.NotNull(result);
+            Assert.NotEqual(ModelBindingResult.NoResult, result);
```
dougbu: This test is failing in the CI builds. Probably now two separate tests: In one case the system returns `NoResult` and in the other, the model is set.

dougbu:
:shipit: after getting the CI builds working

davidfowl:
Why is travis all red on MVC all of a sudden?

NeelBhatt:
I guess here something is wrong: 

https://github.com/aspnet/Mvc/commit/78c9752622526c6faa71193f0250d03a76c971cbBindingSource

for this test:

ModelBinder_ReturnsNonEmptyResult_MatchingSource

dougbu:
@NeelBhatt is correct. See my [earlier comment](https://github.com/aspnet/Mvc/pull/3090#discussion-diff-38721427).

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/BindingSourceModelBinderTest.cs
NeelBhatt: I just noticed you already wrote this comment!

Issue #3093, opened 2015-09-04
 Self referencing loop detected Error Formatters.Json 


SonOfSam:
Giving the following models listed below, a JsonSerializationException is thrown that causes no content in the body of the response but a 200 response is send back.

 	public class Order
	{
		public Order()
		{
		}

		public int Id { get; set; }

		public int BillToCustomerId { get; set; }

		public Customer BillToCustomer { get; set; }

	}

	public class Customer
	{
		public Customer()
		{
			this.Orders = new List<Order>();
		}
		public int Id { get; set; }
		public IList<Order> Orders { get; set; }
	}

Issue occur in file JsonOutputFormatter.cs WriteObject method line 57

	public void WriteObject([NotNull] TextWriter writer, object value)
	{
		using (var jsonWriter = CreateJsonWriter(writer))
		{
			var jsonSerializer = CreateJsonSerializer();
			jsonSerializer.Serialize(jsonWriter, value);
		}
	}

![complexobject](https://cloud.githubusercontent.com/assets/5281768/9696530/1298609e-533f-11e5-9395-955f4949a669.PNG)

![response](https://cloud.githubusercontent.com/assets/5281768/9696511/db4f3fea-533e-11e5-8bcb-39419582c9ca.PNG)

SonOfSam:
@davidfowl, thanks for your help, I was able to modify json settings using the following code. 

	services.Configure<MvcJsonOptions>(option =>
	{
		option.SerializerSettings.ReferenceLoopHandling = ReferenceLoopHandling.Ignore;
	});

davidfowl:
Great! For beta7 be sure to follow announcements  https://github.com/aspnet/Announcements/issues/62

PR #3097, opened 2015-09-05
Self Reference loop Ignore

SonOfSam:
When doing serialization of complex object fail do to a Self referencing loop detected Error. See issue  #3093


davidfowl:
You can customize the json settings in your application. This changes doesn't need to be inbox itself.

SonOfSam:
This is my first commit so I apologies if I broke any rules regarding MS process,  but in my opinion if MVC does not surface the error it will need to fix the way that I did in the PR

davidfowl:
This isn't a bug. MVC never handled self referencing objects before and probably shouldn't by default. You can customize that in your application.

PR #3104, opened 2015-09-07
Simplify CompositeModelBinder

rynowak:
Also adds a diagnostic ToString to ModelBindingResult to make testing
easier.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingResult.cs
@@ -76,20 +76,6 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             return Task.FromResult(Success(key, model, validationNode));
         }
 
-        /// <summary>
-        /// Creates a new <see cref="ModelBindingResult"/> using the provided <paramref name="key"/>.
-        /// </summary>
-        /// <param name="key">The key of the current model binding operation.</param>
-        /// <param name="other">The other <see cref="ModelBindingResult" /> to copy from.</param>
-        public ModelBindingResult([NotNull] string key, ModelBindingResult other)
-        {
-            Key = key;
-
-            Model = other.Model;
-            IsModelSet = other.IsModelSet;
-            ValidationNode = other.ValidationNode;
-        }
-
         private ModelBindingResult(string key, object model, bool isModelSet, ModelValidationNode validationNode)
         {
             Key = key;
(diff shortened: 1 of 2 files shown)


rynowak:
This change should remove a task allocation for every model binding operation. No test changes were needed, confirming my suspicion that the scenarios covered by this were fixed elsewhere. 

Currently working on before/after perf results.

rynowak:
Before:
image

After:
image

Removes about 15% of the allocated tasks. Not thrilling by any means, but a good savings for something that's basically code cleanup anyway.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingResult.cs
```diff
@@ -130,6 +116,23 @@ private ModelBindingResult(string key, object model, bool isModelSet, ModelValid
         public ModelValidationNode ValidationNode { get; }
 
         /// <inheritdoc />
+        public override string ToString()
```
dougbu: Is this for debugging ease?

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingResult.cs
```diff
@@ -130,6 +116,23 @@ private ModelBindingResult(string key, object model, bool isModelSet, ModelValid
...
+        {
+            if (Key == null)
+            {
+                return "No Result";
+            }
+            else if (IsModelSet)
+            {
+                return $"Success: {Key} -> {Model}";
```
dougbu: Suggest single quotes around `'{Model}'`. Will make things more clear when its `ToString()` includes spaces.

dougbu:
> No test changes were needed, confirming my suspicion that the scenarios covered by this were fixed elsewhere.

Nice. Also confirms my suspicion no code reads `Key` of the top-level `ModelBindingResult`.

dougbu:
:shipit:

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingResult.cs
rynowak: Yes, and also without this the messages printed by xUnit when you have a test failure are pretty un-actionable.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingResult.cs
rynowak: sure

PR #3105, opened 2015-09-08
Optimize async code in model binders

rynowak:
Optimize the 'no-op' path for our model binders to return a cached task
where possible without going async.


rynowak:
OK, now this is really where the sausage gets made (delicious perf-sausage). We can avoid going async in all of the cases where a model binder doesn't actually bind anything, and this helps us avoid allocating a LOT of tasks.

Model binding still isn't in a great place after this change, but it's a big improvement. I nuked `BindingSourceModelBinder` because it's an anti-pattern with this change. You **need** to care about perf if you're writing a model binder unfortunately.

Before (includes changes in #3104):
image

After:
image


**YOWZA**
This change cuts the number of `Task<ModelBindingResult>` allocations in half. Moreover we avoid allocating a bunch of async state machines we used to have to instantiate - see `BindingSourceModelBinder+<horriblename>` in the first set of results. This change is saving us 8-10mb over 3000 requests.

benaadams:
Now you are cooking with gas! :+1:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
```diff
@@ -14,9 +14,20 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
...
+                return ModelBindingResult.NoResultAsync;
+            }
+
+            return BindModelCoreAsync(bindingContext, binderType);
+        }
+
+        private async Task<ModelBindingResult> BindModelCoreAsync(ModelBindingContext bindingContext, Type binderType)
+        {
+            binderType = ResolveBinderType(bindingContext);
             if (binderType != null)
```
dougbu: Why not pass in `binderType` from `BindModelAsync()`? And why does the value need to be `null`-checked again?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/FormCollectionModelBinder.cs
```diff
@@ -18,7 +18,17 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
...
+            {
+                return ModelBindingResult.NoResultAsync;
+            }
+
+            return BindModelCoreAsync(bindingContext);
+        }
+
+        private async Task<ModelBindingResult> BindModelCoreAsync(ModelBindingContext bindingContext)
         {
             if (bindingContext.ModelType != typeof(IFormCollection))
```
dougbu: Why does this block still exist?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/BodyModelBinder.cs
```diff
@@ -14,19 +14,31 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
...
-        protected async override Task<ModelBindingResult> BindModelCoreAsync(
-            [NotNull] ModelBindingContext bindingContext)
+        /// <summary>
+        /// Attempts to bind the model using formatters.
+        /// </summary>
+        /// <param name="bindingContext">The <see cref="ModelBindingContext"/>.</param>
+        /// <returns>
+        /// A <see cref="Task{ModelBindingResult}"/> which when completed returns a <see cref="ModelBindingResult"/>.
+        /// </returns>
+        protected async Task<ModelBindingResult> BindModelCoreAsync([NotNull] ModelBindingContext bindingContext)
```
dougbu: `private` ?

dougbu:
Looks as if `TestBindingSourceModelBinder` still exists and that inherits from `BindingSourceModelBinder`. That adds up to 128 functional test failures.

Does our use of `BindingSourceModelBinder` as a basis for a binder in a functional test imply the `abstract` class would be useful to our customers? Or is `TestBindingSourceModelBinder` another anti-pattern -- a binder that should be checking the `BindingSource` match for itself?

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/GenericModelBinder.cs
rynowak: It might have decided to become null ¯\_(ツ)_/¯

davidfowl:
CI hates this change

rynowak:
> Does our use of BindingSourceModelBinder as a basis for a binder in a functional test imply the abstract class would be useful to our customers? Or is TestBindingSourceModelBinder another anti-pattern -- a binder that should be checking the BindingSource match for itself?

I don't think it's worth keeping. We should nuke it from our functional tests.

rynowak:
> CI hates this change

Good thing I haven't checked it in yet then. Making all the functional tests happy is usually something I do after addressing PR feedback.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/FormCollectionModelBinder.cs
rynowak: Cleaned up

src/Microsoft.AspNet.Mvc.Core/ModelBinding/BodyModelBinder.cs
rynowak: sure

rynowak:
Updated

dougbu:
:shipit:

davidfowl:
@rynowak This change went in right?

rynowak:
Yep. So many more perfs now.

PR #3109, opened 2015-09-09
Make a few arrays explicitly `object[]`

dougbu:
- react to aspnet/aspnet.xunit#13
- latest xUnit bits maintains array types correctly, failing some `Type`-related assertions
  - previous xUnit versions mapped all arrays to `object[]`
  - fix likely part of xunit/xunit@bd6814c


rynowak:
:shipit:

PR #3110, opened 2015-09-09
[Design] Using test setup pattern from aspnet/Stress.

pranavkm:
cc @rynowak  \ @Eilon

### DIFF
--- a/test/Microsoft.AspNet.Mvc.FunctionalTests/ActionResultTests.cs
+++ b/test/Microsoft.AspNet.Mvc.FunctionalTests/ActionResultTests.cs
@@ -8,35 +8,33 @@ using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Text;
 using System.Threading.Tasks;
-using ActionResultsWebSite;
-using Microsoft.AspNet.Builder;
-using Microsoft.Framework.DependencyInjection;
 using Xunit;
 
 namespace Microsoft.AspNet.Mvc.FunctionalTests
 {
-    public class ActionResultTests
+    public class ActionResultTests : IClassFixture<MvcFixture<ActionResultsWebSite.Startup>>
     {
-        private const string SiteName = nameof(ActionResultsWebSite);
-        private readonly Action<IApplicationBuilder> _app = new Startup().Configure;
-        private readonly Action<IServiceCollection> _configureServices = new Startup().ConfigureServices;
+        public ActionResultTests(MvcFixture<ActionResultsWebSite.Startup> fixture)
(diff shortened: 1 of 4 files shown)


test/Microsoft.AspNet.Mvc.FunctionalTests/ActionResultTests.cs
```diff
@@ -8,35 +8,33 @@
...
 using System.Threading.Tasks;
-using ActionResultsWebSite;
-using Microsoft.AspNet.Builder;
-using Microsoft.Framework.DependencyInjection;
 using Xunit;
 
 namespace Microsoft.AspNet.Mvc.FunctionalTests
 {
-    public class ActionResultTests
+    public class ActionResultTests : IClassFixture<MvcFixture<ActionResultsWebSite.Startup>>
```
pranavkm: We could create one fixture per test type - this is a lazier way of going about it.

pranavkm:
Class fixtures are created for the lifetime of all tests in a class and disposed at the end of it. Besides fewer lines of setup code per test, this also means we get lifetime management for `TestServer` without having to scatter disposals all over the place. Think @dougbu was looking into this for cross-plat tests.

Eilon:
Sorry don't know the history of this. This is just to make the tests cleaner?

rynowak:
> Sorry don't know the history of this. This is just to make the tests cleaner?

During the StressJam (come on and slam) had converged on this as a better way of doing what we're doing in functional tests, for maintainability and to increase the perf of these tests (they allocate like mad).

This is similar to outcome to what @dougbu is doing - but more normative. Any functional tests that don't fit this pattern, are good candidates for cleanup/removal.

I don't think it's a relatively high priority to do, but it's a good improvement.

Eilon:
Yeah I talked to @pranavkm yesterday and he explained it. Sounds :shipit: to me.

dougbu:
This does not appear to be a positive change compared to the `static Dictionary` approach I've been using. While things are similar,
- This approach moves all test classes using the same web site into the same collection. That's not safe in general.
- It's inflexible. I'm not talking about the ugly hacks we have reconfiguring the sites on a per-test basis. But we do have sites using different `ConfigureServices()` overloads and and different locales. (The localization middleware saves the current locale as its default when instantiated i.e. when the site is first set up.) So, this approach doesn't quite work in general.
- Due to the ugly hacks mentioned above, we have a fair number of tests that simply can't use the fixture. So, this approach should probably wait 'til we remove those hacks -- if we choose to do it eventually.

PR #3116, opened 2015-09-10
Cleanup test conditions and data

dougbu:
See individual commit descriptions for details. One-liners visible below should be enough.

I have not updated the conditions under which tests execute -- no `IsMono` -> `IsWindows` changes for example. That'll come in a later PR.

On the upside however, existing skips are better reported now.


test/Microsoft.AspNet.Mvc.Formatters.Xml.Test/XmlSerializerInputFormatterTest.cs
```diff
@@ -244,15 +240,11 @@ public void HasProperSuppportedEncodings()
             await Assert.ThrowsAsync(typeof(InvalidOperationException), () => formatter.ReadAsync(context));
         }
 
-        [Fact]
+        [ConditionalFact]
+        // ReaderQuotas are not honored on Mono
+        [FrameworkSkipCondition(RuntimeFrameworks.Mono)]
```
pranavkm: Could we set the SkipReason instead of using a comment?

pranavkm: Aww, it's not settable.

pranavkm:
:shipit:

dougbu:
b705922

PR #3124, opened 2015-09-11
Throw an exception if layouts have circular references.

### LINKED ISSUE #3100: Circular layout reference leads to OutOfMemoryException
In one of my `views` folders, I renamed a controller-specific layout from the default to something custom. The layout itself referenced a shared layout of the same name prior to renaming. DropBox then, unbeknownst to me, malfunctioned (as it is prone to do) and dropped a copy of the old layout, before it had been renamed, into my views folder, thus creating a situation where the newly-renamed layout was now erroneously trying to use the old layout which it now sat side-by-side with, instead of the old layout's namesake in the shared views folder. As a result, a circular reference was set up and I spent the next 20 minutes trying to figure out why I was getting OutOfMemoryExceptions when I hadn't changed anything significant.

I'd suggest some kind of circular reference detection in layouts and an exception being thrown to that effect, which would help prevent this kind of confusion.

pranavkm:
Fixes #3100

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor/Resources.resx
+++ b/src/Microsoft.AspNet.Mvc.Razor/Resources.resx
@@ -200,4 +200,7 @@
 
 @{3} "{4}, {5}"</value>
   </data>
+  <data name="LayoutHasCircularReference" xml:space="preserve">
+    <value>Possible circular layout reference detected when rendering '{0}'. The layout page '{1}' has already been rendered.</value>
+  </data>
 </root>
--- a/test/Microsoft.AspNet.Mvc.Razor.Test/RazorViewTest.cs
+++ b/test/Microsoft.AspNet.Mvc.Razor.Test/RazorViewTest.cs
@@ -555,11 +555,13 @@ namespace Microsoft.AspNet.Mvc.Razor
                     await v.RenderSectionAsync("sectionA");
                 });
             });
+            nestedLayout.Path = "NestedLayout";
             var baseLayout = new TestableRazorPage(v =>
             {
                 v.HtmlEncoder = htmlEncoder;
                 v.RenderSection("sectionB");
             });
+            baseLayout.Path = "Layout";
 
             var viewEngine = new Mock<IRazorViewEngine>();
             viewEngine.Setup(p => p.FindPage(It.IsAny<ActionContext>(), "NestedLayout"))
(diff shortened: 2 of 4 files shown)


src/Microsoft.AspNet.Mvc.Razor/Resources.resx
```diff
@@ -200,4 +200,7 @@
 
 @{3} "{4}, {5}"</value>
   </data>
+  <data name="LayoutHasCircularReference" xml:space="preserve">
+    <value>Possible circular layout reference detected when rendering '{0}'. The layout page '{1}' has already been rendered.</value>
```
Eilon: Why `Possible`? Isn't it definite?

`A circular layout reference was detected when rendering '{0}'. The layout page '{1}' has already been rendered.`

test/Microsoft.AspNet.Mvc.Razor.Test/RazorViewTest.cs
```diff
@@ -555,11 +555,13 @@ public class RazorViewTest
                     await v.RenderSectionAsync("sectionA");
                 });
             });
+            nestedLayout.Path = "NestedLayout";
```
Eilon: So did these start failing because the empty string was detected as a circular reference?

pranavkm: Yeah. In the general case, we set the `IRazorPage.Path` to the location where we found the cshtml file so it shouldn't be null.

Eilon: Yup.

src/Microsoft.AspNet.Mvc.Razor/Resources.resx
pranavkm: I was discussion this with @kichalla. There's an unorthodox scenario where you could eventually terminate the cycle:

```
// _Layout.cshtml
@{
   var node = (Node)Context.Items["node"];
   if (node != null)
   {
      Layout = "_Layout";
     Context.Items["node"] = node.Parent;
   }
}
@RenderBody()

```

Eilon: Right, we talked about that one ourselves too. Ultimately we disallow *any* form of circular reference, so I think the error message needs to sound stricter...

Issue #3123, opened 2015-09-11
Remove `HtmlString` special cases


dougbu:
@Eilon and I were looking through all of the places we use `as HtmlString` in the MVC product code. It appears that we're ignoring the broader `as IHtmlContent` case where this is done. That means some code will do nothing and other code will do the wrong thing when a non-`HtmlString` `IHtmlContent` is found. Since non-`HtmlString` `IHtmlContent` will (e.g. with the addition of `AppendEncoded()`) become more common, we should reserve special-casing `HtmlString` for performance hot spots -- should any turn up _later_.

dougbu:
One bit done in eb70b1c

dougbu:
Two more bits done in
- 7bb0a1a for `IJsonHelper`
- aspnet/Antiforgery@6a9b38d and 6e299d6 for `IAntiforgery`

dougbu:
Another bit in 42f3e76

dougbu:
Last bit in 1f76b3c

Issue #3122, opened 2015-09-11
Avoid calling `HtmlEncode()` where possible


dougbu:
We have at least a couple of `HtmlEncode()` calls in product code that should be replaced with a `BufferedHtmlContent` or separate additions to a `TagHelperContent`. As-is, we're creating new `string`s unnecessarily and perhaps making incorrect assumptions about what has already been HTML encoded.

Two examples:
- `LinkTagHelper` currently contains
``` c#
builder.AppendFormat(
    CultureInfo.InvariantCulture,
    "<meta name=\"x-stylesheet-fallback-test\" class=\"{0}\" />",
    HtmlEncoder.HtmlEncode(FallbackTestClass));
```
This could instead be written as
``` c#
builder
    .Append(new HtmlString("<meta name=\"x-stylesheet-fallback-test\" class=\"))
    .Append(FallbackTestClass)
    .Append(new HtmlString("\" />");
```
- `UrlResolutionTagHelper` currently contains
``` c#
var applicationPath = appRelativeUrl.Substring(0, appRelativeUrl.Length -
    postTildeSlashUrlValue.Length);
var encodedApplicationPath = HtmlEncoder.HtmlEncode(applicationPath);

resolvedUrl = string.Concat(encodedApplicationPath, postTildeSlashUrlValue);
```
This could instead be written (with appropriate changes to the `out` parameter) as
``` c#
var applicationPath = appRelativeUrl.Substring(0, appRelativeUrl.Length -
    postTildeSlashUrlValue.Length);
resolvedUrl = new BufferedHtmlContent();
resolvedUrl
    .Append(applicationPath)
    .AppendEncoded(postTildeSlashUrlValue);
```

dougbu:
1f76b3c

PR #3125, opened 2015-09-13
React to BufferedHtmlContent changes

rynowak:


src/Microsoft.AspNet.Mvc.Razor/RazorTextWriter.cs
```diff
@@ -30,9 +32,12 @@ public class RazorTextWriter : TextWriter, IBufferedTextWriter
         /// <param name="unbufferedWriter">The <see cref="TextWriter"/> to write output to when this instance
         /// is no longer buffering.</param>
         /// <param name="encoding">The character <see cref="Encoding"/> in which the output is written.</param>
-        public RazorTextWriter(TextWriter unbufferedWriter, Encoding encoding)
+        /// <param name="encoder">The HTML encoder.</param>
+        public RazorTextWriter(TextWriter unbufferedWriter, Encoding encoding, IHtmlEncoder encoder)
```
rynowak: The changes to this class maintain the current behavior - `Write(string)` does not result in encoding.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/StringCollectionTextWriter.cs
```diff
@@ -178,5 +198,46 @@ public override string ToString()
                 return writer.ToString();
             }
         }
+
+        internal class StringCollectionTextWriterContent : IHtmlContent
+        {
+            private List<object> _entries;
```
davidfowl: readonly

test/Microsoft.AspNet.Mvc.IntegrationTests/ActionParametersIntegrationTest.cs
```diff
@@ -209,6 +209,8 @@ private class Person5
         [Fact]
         public async Task ActionParameter_NonSettableCollectionModel_WithPrefix_GetsBound()
         {
+
+
```
davidfowl: nuke

src/Microsoft.AspNet.Mvc.Razor/RazorTextWriter.cs
```diff
@@ -54,7 +59,7 @@ public override Encoding Encoding
         private TextWriter TargetWriter { get; set; }
 
         [RazorInject]
```
dougbu: Shouldn't need this now that value is constructor injected. BTW I'm curious why you made that change?

rynowak: Oops for got to delete that. This was **never** working. `[RazorInject]` only applies directly on the page itself.

src/Microsoft.AspNet.Mvc.Razor/RazorTextWriter.cs
```diff
@@ -65,7 +70,7 @@ public override void Write(char value)
         /// <inheritdoc />
         public override void Write(object value)
         {
-            var htmlString = value as HtmlString;
+            var htmlString = value as IHtmlContent;
```
dougbu: :+1: this removes one of the #3123 cases and looks Just Right:tm:.

src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
```diff
@@ -9,6 +9,7 @@
 using Microsoft.AspNet.Mvc.Rendering;
 using Microsoft.AspNet.PageExecutionInstrumentation;
 using Microsoft.Framework.Internal;
+using Microsoft.Framework.WebEncoders;
```
dougbu: Changes to this class and `RazorViewFactory` are downstream of the switch from `[RazorInject]`. If required because property injection wasn't working (and the encoder is used more now), fine.

src/Microsoft.AspNet.Mvc.Razor/RazorTextWriter.cs
dougbu: ... but please fix the variable name to avoid misrepresenting things. It's not an `HtmlString`.

src/Microsoft.AspNet.Mvc.Razor/RazorTextWriter.cs
dougbu: Changes to `StringCollectionTextWriter` mean `Write(string)` does not result in encoding. Changes to this class are about moving from property to constructor injection for the `HtmlEncoder`.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -323,26 +324,27 @@ private void BuildFallbackBlock(TagHelperContent builder)
...
+                builder.Append(HtmlString.NewLine);
 
                 // Build the <meta /> tag that's used to test for the presence of the stylesheet
-                builder.AppendFormat(
-                    CultureInfo.InvariantCulture,
-                    "<meta name=\"x-stylesheet-fallback-test\" class=\"{0}\" />",
-                    HtmlEncoder.HtmlEncode(FallbackTestClass));
+                builder
+                    .AppendEncoded("<meta name=\"x-stylesheet-fallback-test\" class=\"")
+                    .Append(FallbackTestClass)
```
dougbu: Worth checking for remaining `HtmlEncoder.HtmlEncode()` calls (outside `HtmlHelper.Encode()` and similar) in your branch's product code. You may have fixed all instances of #3122 in this PR.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/StringCollectionTextWriter.cs
```diff
@@ -62,7 +70,19 @@ public override void Write([NotNull] char[] buffer, int index, int count)
...
+                var currentCount = count;
+                if (MaxCharToStringLength < currentCount)
+                {
+                    currentCount = MaxCharToStringLength;
+                }
+
+                _content.Append(new string(buffer, index, currentCount));
+                index += currentCount;
+                count -= currentCount;
+            }
```
dougbu: Messy to duplicate `BufferedHtmlContent` code here as well as in `StringCollectionTextWriterContent`. Would it be more straightforward to make `BufferedHtmlContent.WriteTo()` `virtual` and use a `private` subclass here -- overriding `WriteTo()` and `ToString()`? Don't think `StringCollectionTextWriterContent` (the proposed subclass then) would be nearly as large.

rynowak: This method is part of the textwriter API, which is why we implement it here. The better change would be to remove it from `BufferedHtmlContent`.

Thoughts?

src/Microsoft.AspNet.Mvc.Razor/RazorTextWriter.cs
rynowak: sure

test/Microsoft.AspNet.Mvc.Razor.Test/RazorPageTest.cs
```diff
@@ -1783,11 +1778,11 @@ public void WriteAttributeTo_WritesAsExpected(AttributeValue[] attributeValues,
...
-            output.PreElement.SetContent(preElement);
-            output.PreContent.SetContent(preContent);
-            output.Content.SetContent(content);
-            output.PostContent.SetContent(postContent);
-            output.PostElement.SetContent(postElement);
+            output.PreElement.SetContent(new HtmlString(preElement));
+            output.PreContent.SetContent(new HtmlString(preContent));
+            output.Content.SetContent(new HtmlString(content));
+            output.PostContent.SetContent(new HtmlString(postContent));
+            output.PostElement.SetContent(new HtmlString(postElement));
```
dougbu: Since these `TagHelperContent` instances are initially empty, could use `AppendEncoded()` here.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
rynowak: Found additional use in UrlResolutionTagHelper (encoding the application path).

test/Microsoft.AspNet.Mvc.Razor.Test/RazorPageTest.cs
rynowak: Sure, I'll do a pass for `SetContent(new HtmlString(`

rynowak:
Updated with feedback so far

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/StringCollectionTextWriterTest.cs
```diff
@@ -129,11 +129,9 @@ public void Copy_CopiesContent_IfTargetTextWriterIsAStringCollectionTextWriter()
...
             // Assert
             // Make sure content was written to the source.
-            Assert.Equal(2, source.Content.Entries.Count);
-            Assert.Equal(1, target.Content.Entries.Count);
-            var result = Assert.Single(target.Content.Entries);
-            var bufferedHtmlContent = Assert.IsType<BufferedHtmlContent>(result);
-            Assert.Same(source.Content.Entries, bufferedHtmlContent.Entries);
+            Assert.Equal(2, source.Entries.Count);
+            Assert.Equal(1, target.Entries.Count);
+            var result = Assert.Single(target.Entries);
```
dougbu: Did you intend to assert something about `result`?

test/Microsoft.AspNet.Mvc.TagHelpers.Test/InputTagHelperTest.cs
```diff
@@ -95,7 +95,7 @@ public static TheoryData MultiAttributeCheckBoxData
             {
                 TagMode = TagMode.SelfClosing,
             };
-            output.Content.SetContent(originalContent);
+            output.Content.SetContent(new HtmlString(originalContent));
```
dougbu: Suggest using `AppendEncoded()` here and in the next couple of changed classes.

rynowak: Already made this change :+1:

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/StringCollectionTextWriterTest.cs
rynowak: Yep, looks like that was missed in the shuffle.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/StringCollectionTextWriter.cs
dougbu: Perhaps the `WriteTo(char[], int, int)` method could be removed from `BufferedHtmlContent`. I don't see other uses.

Even then `StringCollectionTextWriterContent` should be cleaner as a `BufferedHtmlContent` subclass. Why did you choose to duplicate?

Separately, might be worth explaining at the top of this block _why_ limiting ourselves to 1k blocks is important.

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/StringCollectionTextWriter.cs
dougbu: Discussed offline. I'm :ok: given @rynowak's objection to subclassing in a way that inverts the `BufferedHtmlContent` behaviour. Beats my objection to duplicating some boilerplate code.

dougbu:
:shipit:

PR #3128, opened 2015-09-14
No need to add an instance in `WebApiCompatShimMvcBuilderExtensions`

dougbu:
- DI has changed since this class was written

### DIFF
--- a/src/Microsoft.AspNet.Mvc.WebApiCompatShim/WebApiCompatShimMvcBuilderExtensions.cs
+++ b/src/Microsoft.AspNet.Mvc.WebApiCompatShim/WebApiCompatShimMvcBuilderExtensions.cs
@@ -18,9 +18,7 @@ namespace Microsoft.Framework.DependencyInjection
             builder.Services.TryAddEnumerable(
                 ServiceDescriptor.Transient<IConfigureOptions<WebApiCompatShimOptions>, WebApiCompatShimOptionsSetup>());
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.Mvc.WebApiCompatShim/WebApiCompatShimMvcBuilderExtensions.cs
```diff
@@ -18,9 +18,7 @@ public static IMvcBuilder AddWebApiConventions(this IMvcBuilder builder)
             builder.Services.TryAddEnumerable(
                 ServiceDescriptor.Transient<IConfigureOptions<WebApiCompatShimOptions>, WebApiCompatShimOptionsSetup>());
 
-            // The constructors on DefaultContentNegotiator aren't DI friendly, so just
-            // new it up.
-            builder.Services.TryAdd(ServiceDescriptor.Instance<IContentNegotiator>(new DefaultContentNegotiator()));
+            builder.Services.TryAddSingleton<IContentNegotiator, DefaultContentNegotiator>();
```
dougbu: Could just as easily be transient. But that would only create more `DefaultContentNegotiator` instances without changing behaviour (that I've seen). `DefaultContentNegotiator` is not stateful.

rynowak:
:shipit:

dougbu:
29a25b2

PR #3131, opened 2015-09-14
Do not execute no-op tests

dougbu:
- consistently use `TestPlatformHelper` only to skip test _cases_ or change expectations


dougbu:
FYI only. Will merge immediately.

Issue #3066, opened 2015-08-30
Frequent `IOException`s using MVC with Core CLR on Linux


dougbu:
Functional tests on Linux often fail due to `IOException`s. I currently see 97 test failures in `Test/Microsoft.AspNet.Mvc.FunctionalTests` and 78 `IOException`s in the output.

The root cause may be an external issue but should start with confirming we're using `PhysicalFilesWatcher`s efficiently (given the number of files in the repo). Or perhaps this is a known system issue we should work around when testing on Linux (similar to setting `MONO_THREADS_PER_CPU`).

Example stack traces (think this covers all 78 `IOException`s):
```
      System.IO.IOException : No space left on device
      Stack Trace:
           at Interop.CheckIo(Int64 result, String path, Boolean isDirectory, Func`2 errorRewriter)
           at System.IO.FileSystemWatcher.RunningInstance.SysCall[TArg1,TArg2](Func`4 sysCall, TArg1 arg1, TArg2 arg2)
           at System.IO.FileSystemWatcher.RunningInstance.AddDirectoryWatchUnlocked(WatchedDirectory parent, String directoryName)
           at System.IO.FileSystemWatcher.RunningInstance.AddDirectoryWatchUnlocked(WatchedDirectory parent, String directoryName)
...
           at System.IO.FileSystemWatcher.RunningInstance..ctor(FileSystemWatcher watcher, SafeFileHandle inotifyHandle, String directoryPath, Boolean includeSubdirectories, NotifyEvents notifyFilters, CancellationToken cancellationToken)
           at System.IO.FileSystemWatcher.StartRaisingEvents()
           at Microsoft.AspNet.FileProviders.PhysicalFilesWatcher.CreateFileChangeTrigger(String filter)
           at Microsoft.AspNet.Mvc.Razor.Compilation.CompilerCache.GetOrAdd(String relativePath, Func`2 compile)
           at Microsoft.AspNet.Mvc.Razor.VirtualPathRazorPageFactory.CreateInstance(String relativePath)
```

```
      System.IO.IOException : No space left on device
      Stack Trace:
           at Interop.CheckIo(Int64 result, String path, Boolean isDirectory, Func`2 errorRewriter)
           at System.IO.FileSystemWatcher.RunningInstance.SysCall[TArg1,TArg2](Func`4 sysCall, TArg1 arg1, TArg2 arg2)
[...]

PR #3133, opened 2015-09-15
Use `MvcTextFixture` as much as possible

dougbu:
- #3066 
  - reduce `TestServer` -> `PhysicalFileProvider` -> `FileSystemWatcher` count enough to run with Core CLR on Linux
- remove use of `HttpClient.DefaultRequestHeaders`; any client change affects other tests
- remove use of `RequestBuilder` class; creates a per-test `HttpClient` and requires the `TestServer`
- updated a few expectations because `CommonTestEncoder` does JavaScript a bit differently
  - "JavaScriptEncode[[...]]" -> "JavaScriptStringEncode[[...]]"
- side benefit: xUnit reports functional tests execute for only ~12.4s; was >30s before this change

nits:
- touched-up some leftover `_app` &c declarations to be more explicit and minimize `using`s
- moved statements into correct sections of methods in `RoutingTests`
- removed `TestLoggerFactory` and related classes from `TagHelperSampleTest`


test/Microsoft.AspNet.Mvc.FunctionalTests/TagHelperSampleTest.cs
```diff
@@ -1,106 +1,55 @@
...
+                    "/Home/Edit/0",
+                    "/Home/Edit/0?Name=Bobby&Blurb=howdy&DateOfBirth=1999-11-30&YearsEmployeed=2",
+                    "/Home/Index",
+                };
             }
         }
 
-        private class TestDisposable : IDisposable
+        [Theory]
+        [MemberData(nameof(PathData))]
```
dougbu: Two overlapping changes aren't easy to read here:
- don't need `TestLoggerFactory` plus `NullLoggerFactory`
- `Home_Pages_ReturnSuccess()` should never have been a single test with a loop in it

test/Microsoft.AspNet.Mvc.FunctionalTests/WebApiCompatShimBasicTest.cs
```diff
@@ -19,21 +19,25 @@
 
 namespace Microsoft.AspNet.Mvc.FunctionalTests
 {
-    public class WebApiCompatShimBasicTest
+    public class WebApiCompatShimBasicTest : IClassFixture<MvcTestFixture<WebApiCompatShimWebSite.Startup>>
     {
         private const string SiteName = nameof(WebApiCompatShimWebSite);
         private readonly Action<IApplicationBuilder> _app = new WebApiCompatShimWebSite.Startup().Configure;
```
dougbu: A bug I haven't got to the bottom of (specific to Core CLR test runs) prevents changing 3 test methods in this class.

test/Microsoft.AspNet.Mvc.FunctionalTests/WebApiCompatShimBasicTest.cs
```diff
@@ -61,24 +61,21 @@ public class WebApiCompatShimBasicTest
                 content);
         }
 
-#if !DNXCORE50
```
dougbu: Just happened to notice this test could be enabled with Core-specific expectations.

test/Microsoft.AspNet.Mvc.FunctionalTests/project.json
```diff
@@ -1,83 +1,83 @@
...
-    "exclude": [
-        "wwwroot",
-        "node_modules",
-        "bower_components"
-    ]
+    "Microsoft.AspNet.Mvc.TestConfiguration": "1.0.0",
+    "Microsoft.AspNet.TestHost": "1.0.0-*",
+    "Microsoft.AspNet.WebUtilities": "1.0.0-*",
+    "Microsoft.Framework.Configuration.Json": "1.0.0-*",
+    "Microsoft.Framework.Logging.Testing": "1.0.0-*",
```
dougbu: Added this line. The rest is noise.

test/Microsoft.AspNet.Mvc.FunctionalTests/FilteredDefaultAssemblyProviderFixtureOfT.cs
```diff
@@ -0,0 +1,41 @@
...
+                : base(libraryManager)
+            {
+            }
+
+            protected override IEnumerable<Library> GetCandidateLibraries()
+            {
+                var libraries = base.GetCandidateLibraries();
+
+                // Filter out other WebSite projects
+                return libraries.Where(library => !library.Name.Contains("WebSite") ||
```
pranavkm: I realize this is unrelated to your change - but is there a fixed set of names we could pass to this method? This seems kinda fragile.

test/Microsoft.AspNet.Mvc.FunctionalTests/MvcTestFixture.cs
```diff
@@ -2,62 +2,105 @@
...
+                ?.CreateDelegate(typeof(Action<IApplicationBuilder>), startupInstance);
+            if (configureApplication == null)
+            {
+                var configureWithLogger = (Action<IApplicationBuilder, ILoggerFactory>)startupTypeInfo
+                    .DeclaredMethods
+                    .FirstOrDefault(m => m.Name == "Configure" && m.GetParameters().Length == 2)
+                    ?.CreateDelegate(typeof(Action<IApplicationBuilder, ILoggerFactory>), startupInstance);
+                Debug.Assert(configureWithLogger != null);
+
+                configureApplication = application => configureWithLogger(application, NullLoggerFactory.Instance);
```
pranavkm: If all we do is pass in a `NullLoggerFactory`, couldn't we simply change these websites to not use this `Configure` method?

test/Microsoft.AspNet.Mvc.FunctionalTests/TagHelperSampleTest.cs
```diff
@@ -1,106 +1,55 @@
...
-                return new TestLogger();
-            }
-
-            public void Dispose()
-            {
-            }
-        }
-
-        private class TestLogger : ILogger
+        public static TheoryData<string> PathData
```
pranavkm: :100:

test/Microsoft.AspNet.Mvc.FunctionalTests/WebApiCompatShimBasicTest.cs
pranavkm: :+1:

pranavkm:
:shipit:

test/Microsoft.AspNet.Mvc.FunctionalTests/FilteredDefaultAssemblyProviderFixtureOfT.cs
dougbu: Suggest this is fine as-is. Slightly-filtered set includes every referenced framework library, the functional test library itself, and all sample site libraries. That is, it excludes _only_ the other non-sample web sites. And the idea appears to be being inclusive enough to make the `IControllerTypeProvider` implementation work a bit.

test/Microsoft.AspNet.Mvc.FunctionalTests/MvcTestFixture.cs
dougbu: We're talking about a perfectly valid `Startup` class. And it (`TagHelperSample.Web`) is a sample and therefore more likely than most sites to be used outside the functional tests.

dougbu:
d03a851

PR #3134, opened 2015-09-15
Razor CompilerCache allocates too much in NormalizePath

### LINKED ISSUE #3035: [PERF] Razor CompilerCache allocates too much in NormalizePath
Per @pranavkm this can be simplified or done elsewhere. `CompilerCache.NormalizePath` is allocating about 13mb across 10000 requests to a bare-bones MVC app with a view.

pranavkm:
Fixes #3035

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilerCache.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilerCache.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Text;
@@ -19,9 +20,8 @@ namespace Microsoft.AspNet.Mvc.Razor.Compilation
         private readonly IFileProvider _fileProvider;
         private readonly IMemoryCache _cache;
 
-        private readonly object _cacheKeyLookupLock = new object();
-        private readonly Dictionary<string, string> _normalizedPathLookup =
-            new Dictionary<string, string>(StringComparer.Ordinal);
+        private readonly ConcurrentDictionary<string, string> _normalizedPathLookup =
+            new ConcurrentDictionary<string, string>(StringComparer.Ordinal);
 
         /// <summary>
         /// Initializes a new instance of <see cref="CompilerCache"/>.
@@ -58,6 +58,8 @@ namespace Microsoft.AspNet.Mvc.Razor.Compilation
             [NotNull] Func<RelativeFileInfo, CompilationResult> compile)
         {
             CompilerCacheResult cacheResult;
+            // Attempt to lookup the cache entry using the passed in path. This will succeed if the path is already
+            // normalized and a cache entry exists.
             if (!_cache.TryGetValue(relativePath, out cacheResult))
             {
                 var normalizedPath = GetNormalizedPath(relativePath);
@@ -71,11 +73,11 @@ namespace Microsoft.AspNet.Mvc.Razor.Compilation
         }
 
         private CompilerCacheResult CreateCacheEntry(
-            string relativePath,
+            string normalizedPath,
             Func<RelativeFileInfo, CompilationResult> compile)
         {
             CompilerCacheResult cacheResult;
-            var fileInfo = _fileProvider.GetFileInfo(relativePath);
+            var fileInfo = _fileProvider.GetFileInfo(normalizedPath);
             MemoryCacheEntryOptions cacheEntryOptions;
             CompilerCacheResult cacheResultToCache;
             if (!fileInfo.Exists)
@@ -84,13 +86,13 @@ namespace Microsoft.AspNet.Mvc.Razor.Compilation
                 cacheResult = CompilerCacheResult.FileNotFound;
 
                 cacheEntryOptions = new MemoryCacheEntryOptions();
-                cacheEntryOptions.AddExpirationTrigger(_fileProvider.Watch(relativePath));
+                cacheEntryOptions.AddExpirationTrigger(_fileProvider.Watch(normalizedPath));
             }
             else
             {
-                var relativeFileInfo = new RelativeFileInfo(fileInfo, relativePath);
+                var relativeFileInfo = new RelativeFileInfo(fileInfo, normalizedPath);
                 var compilationResult = compile(relativeFileInfo).EnsureSuccessful();
-                cacheEntryOptions = GetMemoryCacheEntryOptions(relativePath);
+                cacheEntryOptions = GetMemoryCacheEntryOptions(normalizedPath);
 
                 // By default the CompilationResult returned by IRoslynCompiler is an instance of
                 // UncachedCompilationResult. This type has the generated code as a string property and do not want
@@ -100,7 +102,7 @@ namespace Microsoft.AspNet.Mvc.Razor.Compilation
                 cacheResult = new CompilerCacheResult(compilationResult);
             }
 
-            _cache.Set(relativePath, cacheResultToCache, cacheEntryOptions);
+            _cache.Set(normalizedPath, cacheResultToCache, cacheEntryOptions);
             return cacheResult;
         }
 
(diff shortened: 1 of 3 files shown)


davidfowl:
@pranavkm please make a before and after profile and show the results.

pranavkm:
@davidfowl @rynowak walked me through the profiling for this so I thought it was cool to not have before \ after profiles for this one.

rynowak:
@pranavkm don't drag me down with you :laughing:

pranavkm:
For 3000 requests, allocated bytes and object allocations:

**Before**
before

**After**
after

rynowak:
That's a really really nice reduction in string allocations. There's another hotspot involving a string.replace that doesn't have a screenshot, but definitely is fixed by this change.

src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilerCache.cs
```diff
@@ -42,8 +48,7 @@ public CompilerCache([NotNull] IFileProvider fileProvider)
             foreach (var item in precompiledViews)
             {
                 var cacheEntry = new CompilerCacheResult(CompilationResult.Successful(item.Value));
-                var normalizedPath = NormalizePath(item.Key);
-                _cache.Set(normalizedPath, cacheEntry);
+                _cache.Set(GetNormalizedPath(item.Key), cacheEntry);
```
pranavkm: Updated + added tests

davidfowl:
Noice

src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilerCache.cs
```diff
@@ -101,15 +118,31 @@ private MemoryCacheEntryOptions GetMemoryCacheEntryOptions(string relativePath)
...
+            if (relativePath.Length == 0)
+            {
+                return relativePath;
+            }
 
-            return path;
+            lock (_cacheKeyLookupLock)
+            {
+                string normalizedPath;
+                if (!_normalizedPathLookup.TryGetValue(relativePath, out normalizedPath))
```
rynowak: curious, why a lock rather than a ConcurrentDictionary? Lock seems more error prone and you're only protecting one resource

src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilerCache.cs
```diff
@@ -17,6 +19,10 @@ public class CompilerCache : ICompilerCache
         private readonly IFileProvider _fileProvider;
         private readonly IMemoryCache _cache;
 
+        private readonly object _cacheKeyLookupLock = new object();
+        private readonly Dictionary<string, string> _normalizedPathLookup =
+            new Dictionary<string, string>(StringComparer.Ordinal);
```
rynowak: Names of these should reflect that they go together

src/Microsoft.AspNet.Mvc.Razor/Precompilation/RazorPreCompiler.cs
```diff
@@ -291,6 +293,13 @@ private static Stream GetNonDisposableStream(Stream stream)
             return new NonDisposableStream(stream);
         }
 
+        private static string CombinePath(string root, string name)
+        {
+            // We use string.Join instead of Path.Combine here to ensure that the path
+            // separator we produce matches the one used by the CompilerCache.
+            return string.Join(PathSeparator, root, name);
```
rynowak: Should we verify that the `root` isn't a rooted path?

src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilerCache.cs
pranavkm: No particular reason - I could switch it to a `ConcurrentDictionary`

src/Microsoft.AspNet.Mvc.Razor/Precompilation/RazorPreCompiler.cs
pranavkm: https://github.com/aspnet/Mvc/pull/3134/files#diff-7f27afec00025f958d6da462dfbe5c93L83 - we specify the root path as an empty string.

src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilerCache.cs
```diff
@@ -52,38 +57,50 @@ public CompilerCache([NotNull] IFileProvider fileProvider)
...
-                    cacheResult = new CompilerCacheResult(compilationResult);
+                    cacheResult = CreateCacheEntry(normalizedPath, compile);
                 }
+            }
+
+            return cacheResult;
+        }
+
+        private CompilerCacheResult CreateCacheEntry(
+            string relativePath,
```
dougbu: suggest `normalizedPath` to match how this method is called

src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilerCache.cs
```diff
@@ -52,38 +57,50 @@ public CompilerCache([NotNull] IFileProvider fileProvider)
             [NotNull] string relativePath,
             [NotNull] Func<RelativeFileInfo, CompilationResult> compile)
         {
-            var normalizedPath = NormalizePath(relativePath);
             CompilerCacheResult cacheResult;
-            if (!_cache.TryGetValue(normalizedPath, out cacheResult))
+            if (!_cache.TryGetValue(relativePath, out cacheResult))
```
dougbu: If this is the "fast path lookup" for the case where `NormalizePath()` returns the same `string`, may be worth a code comment.

src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilerCache.cs
```diff
@@ -101,15 +118,31 @@ private MemoryCacheEntryOptions GetMemoryCacheEntryOptions(string relativePath)
...
-        private static string NormalizePath(string path)
+        private string GetNormalizedPath(string relativePath)
         {
-            // We need to allow for scenarios where the application was precompiled on a machine with forward slashes
-            // but is being run in one with backslashes (or vice versa). To this effect, we'll normalize paths to
-            // use backslashes for lookups and storage in the dictionary.
-            path = path.Replace('/', '\\');
-            path = path.TrimStart('\\');
+            Debug.Assert(relativePath != null);
+            if (relativePath.Length == 0)
```
dougbu: Cases where `string.Empty` and `null` have different semantics tend to cause problems. Why not just `string.IsNullOrEmpty()` w/o the assertion?

src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilerCache.cs
```diff
@@ -101,15 +118,31 @@ private MemoryCacheEntryOptions GetMemoryCacheEntryOptions(string relativePath)
             return options;
         }
 
-        private static string NormalizePath(string path)
+        private string GetNormalizedPath(string relativePath)
         {
-            // We need to allow for scenarios where the application was precompiled on a machine with forward slashes
-            // but is being run in one with backslashes (or vice versa). To this effect, we'll normalize paths to
-            // use backslashes for lookups and storage in the dictionary.
```
dougbu: Are these comments no longer relevant?

src/Microsoft.AspNet.Mvc.Razor/Precompilation/RazorPreCompiler.cs
```diff
@@ -25,6 +25,8 @@ namespace Microsoft.AspNet.Mvc.Razor.Precompilation
 {
     public class RazorPreCompiler
     {
+        private const string PathSeparator = "/";
```
dougbu: May need a more specific name since this is not about path separation in general, just for the compiler cache e.g. `CacheKeyDirectorySeparator`.

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilerCache.cs
pranavkm: It's guaranteed to be notnull, but not empty.

pranavkm:
Updated.

dougbu:
:shipit: I'm happy if @rynowak is happy.

rynowak:
:shipit:

PR #3136, opened 2015-09-15
fix #2747: Fix Mvc localization test once dnx support reading resources

### LINKED ISSUE #2747: Fix Mvc localization test once dnx support reading resources
Localization tests today have a few hacks since dnx does not support reading resources
- TestStringLocalizer,
- Expected strings in tests have if defs

Remove these once dnx adds support for resources.

kirthik:
1. Added Localization extension methods that takes LocalizationOptions as a parameter.
2. Removed plumbing code for generating .resource files.


src/Microsoft.AspNet.Mvc.Localization/Internal/MvcLocalizationServices.cs
```diff
@@ -15,6 +17,14 @@ public static class MvcLocalizationServices
             IServiceCollection services,
             LanguageViewLocationExpanderFormat format)
         {
+            AddLocalizationServices(services, format, setupAction: null);
+        }
```
rynowak: Is this overload ever called? Just remove it if not since this is internal

src/Microsoft.AspNet.Mvc.Localization/ViewLocalizer.cs
```diff
@@ -65,7 +65,6 @@ public void Contextualize(ViewContext viewContext)
             {
                 baseName = baseName.Substring(1);
             }
-            baseName = _applicationName + "." + baseName;
```
rynowak: Due a change in how resources behave?

src/Microsoft.AspNet.Mvc.Localization/DependencyInjection/MvcLocalizationMvcBuilderExtensions.cs
```diff
@@ -40,5 +42,35 @@ public static IMvcBuilder AddViewLocalization([NotNull] this IMvcBuilder builder
             MvcLocalizationServices.AddLocalizationServices(builder.Services, format);
             return builder;
         }
+
+        /// <summary>
+        ///  Adds MVC localization to the application.
```
rynowak: Should probably say 'Adds MVC view localization..'

rynowak:
:watch: for small comments

src/Microsoft.AspNet.Mvc.Localization/ViewLocalizer.cs
kirthik: yes. This assumed resx file is always at root path. Underlying `ResourceManagerStringLocalizerFactory` now takes the basename and appends resource folder path(if any) to form full path of resx file

src/Microsoft.AspNet.Mvc.Localization/Internal/MvcLocalizationServices.cs
kirthik: Tests call it. But I will change the tests to use the other method and will remove this one.

src/Microsoft.AspNet.Mvc.Localization/DependencyInjection/MvcLocalizationMvcBuilderExtensions.cs
kirthik: I will change it.

kirthik:
@rynowak updated

rynowak:
:shipit:

kirthik:
merged ed509f65a23e5d12dfab360a44260ce4234a275b

PR #3137, opened 2015-09-15
Flow IHtmlContent through to the razor buffer

rynowak:
In the current checked in code, an `IHtmlContent` instance will write itself out when it reaches the `RazorTextWriter`. This means that anything that needs to be encoded gets encoded at this time.

Generally (unless you flush) - the `RazorTextWriter` is buffering, which means that each of these writes adds an entry to a `List<object>` held onto by `StringCollectionTextWriter`. When you encode, text is written character by character, leading to lots of little entries. This allocates a lot even for a small site: 
image

That 4 MB translates into over 1kb/request, for a very very small amount of HTML output.

---------------

The change here is to preserve the `IHtmlContent`-ness of values until they need to actually go on the wire (generally when the response is sent). This means that text is encoded character-by-character into the final `TextWriter` - which is an optimized path.

Look at the amount of `string` and `object[]` allocations with these changes

Before:
image

After:
image

`string`s are reduced by about 1.8mb
`object[]`s are reduced by about 5.5mb

This is for 3000 requests on a very small page.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
@@ -434,17 +434,35 @@ namespace Microsoft.AspNet.Mvc.Razor
                     // an attribute value that may have been quoted with single quotes, must handle any double quotes
                     // in the value. Writing the value out surrounded by double quotes.
                     //
-                    // Do not combine following condition with check of escapeQuotes; htmlContent.ToString() can be
-                    // expensive when the IHtmlContent is created with a BufferedHtmlContent.
-                    var stringValue = htmlContent.ToString();
-                    if (stringValue.Contains("\""))
+                    // Do not combine following condition with check of escapeQuotes; converting an IHtmlContent to
+                    // a string can be very expensive.
+                    using (var stringWriter = new StringWriter())
                     {
-                        writer.Write(stringValue.Replace("\"", "&quot;"));
+                        htmlContent.WriteTo(stringWriter, encoder);
+
+                        var stringValue = stringWriter.ToString();
+                        if (stringValue.Contains("\""))
+                        {
+                            stringValue = stringValue.Replace("\"", "&quot;");
+                        }
+
+                        writer.Write(stringValue);
                         return;
                     }
                 }
 
-                htmlContent.WriteTo(writer, encoder);
+                var htmlTextWriter = writer as HtmlTextWriter;
+                if (htmlTextWriter == null)
+                {
+                    htmlContent.WriteTo(writer, encoder);
+                }
+                else
+                {
+                    // This special case alows us to keep buffering as IHtmlContent until we get to the 'final'
+                    // TextWriter.
+                    htmlTextWriter.Write(htmlContent);
+                }
+
                 return;
             }
 
--- a/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
@@ -434,7 +434,7 @@ namespace Microsoft.AspNet.Mvc.Razor
                     // an attribute value that may have been quoted with single quotes, must handle any double quotes
                     // in the value. Writing the value out surrounded by double quotes.
                     //
-                    // Do not combine following condition with check of escapeQuotes; converting an IHtmlContent to
+                    // Be careful combining IHtmlContent with escapeQuotes= = true converting an IHtmlContent to
                     // a string can be very expensive.
                     using (var stringWriter = new StringWriter())
                     {
@@ -636,9 +636,13 @@ namespace Microsoft.AspNet.Mvc.Razor
                     WriteUnprefixedAttributeValueTo(valueBuffer, value);
                 }
 
-                var htmlString = new HtmlString(valueBuffer.ToString());
+                using (var stringWriter = new StringWriter())
+                {
+                    valueBuffer.Content.WriteTo(stringWriter, HtmlEncoder);
 
-                executionContext.AddHtmlAttribute(attributeName, htmlString);
+                    var htmlString = new HtmlString(stringWriter.ToString());
+                    executionContext.AddHtmlAttribute(attributeName, htmlString);
+                }
             }
         }
 
--- a/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
@@ -434,8 +434,7 @@ namespace Microsoft.AspNet.Mvc.Razor
                     // an attribute value that may have been quoted with single quotes, must handle any double quotes
                     // in the value. Writing the value out surrounded by double quotes.
                     //
-                    // Be careful combining IHtmlContent with escapeQuotes= = true converting an IHtmlContent to
-                    // a string can be very expensive.
+                    // This is really not optimal from a perf point of view, but it's the best we can do for right now.
                     using (var stringWriter = new StringWriter())
                     {
                         htmlContent.WriteTo(stringWriter, encoder);
--- a/src/Microsoft.AspNet.Mvc.Razor/RazorTextWriter.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/RazorTextWriter.cs
@@ -88,17 +88,17 @@ namespace Microsoft.AspNet.Mvc.Razor
             }
         }
 
-        /// <inheritdoct />
-        public override void Write(IHtmlContent content)
+        /// <inheritdoc />
+        public override void Write(IHtmlContent value)
         {
             var htmlTextWriter = TargetWriter as HtmlTextWriter;
             if (htmlTextWriter == null)
             {
-                content.WriteTo(TargetWriter, HtmlEncoder);
+                value.WriteTo(TargetWriter, HtmlEncoder);
             }
             else
             {
-                htmlTextWriter.Write(content);
+                htmlTextWriter.Write(value);
             }
         }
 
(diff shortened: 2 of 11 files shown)


rynowak:
Some tests are still needed here. Wanted to get this out for eyeballs.

dougbu:
Have a couple of tests that aren't working here:
- `Microsoft.AspNet.Mvc.Rendering.StringHtmlContentTest.ToString_ReturnsNullForNullInput()`
- `Microsoft.AspNet.Mvc.Rendering.StringHtmlContentTest.ToString_ReturnsAString()`

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -434,17 +434,35 @@ public virtual void WriteTo([NotNull] TextWriter writer, object value)
                     // an attribute value that may have been quoted with single quotes, must handle any double quotes
                     // in the value. Writing the value out surrounded by double quotes.
                     //
-                    // Do not combine following condition with check of escapeQuotes; htmlContent.ToString() can be
-                    // expensive when the IHtmlContent is created with a BufferedHtmlContent.
-                    var stringValue = htmlContent.ToString();
-                    if (stringValue.Contains("\""))
+                    // Be careful combining IHtmlContent with escapeQuotes= = true converting an IHtmlContent to
+                    // a string can be very expensive.
```
dougbu: Who is this comment for? I understand the code isn't optimal but it's currently the best that can be done for the `name='attribute "value"'` case today. If this turns out to be a hot spot, we'll change Razor so this isn't necessary. But for now, the method is called with `escapeQuotes==true` in only one specific situation.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -618,9 +636,13 @@ public virtual void WriteLiteralTo([NotNull] TextWriter writer, string value)
...
 
-                var htmlString = new HtmlString(valueBuffer.ToString());
+                using (var stringWriter = new StringWriter())
+                {
+                    valueBuffer.Content.WriteTo(stringWriter, HtmlEncoder);
 
-                executionContext.AddHtmlAttribute(attributeName, htmlString);
+                    var htmlString = new HtmlString(stringWriter.ToString());
+                    executionContext.AddHtmlAttribute(attributeName, htmlString);
+                }
```
dougbu: What goes South if you instead call `executionContext.AddHtmlAttribute(attributeName, valueBuffer.Content)`?

src/Microsoft.AspNet.Mvc.Razor/RazorTextWriter.cs
```diff
@@ -104,6 +89,20 @@ public override void Write(string value)
...
+            var htmlTextWriter = TargetWriter as HtmlTextWriter;
+            if (htmlTextWriter == null)
+            {
+                value.WriteTo(TargetWriter, HtmlEncoder);
+            }
+            else
+            {
+                htmlTextWriter.Write(value);
+            }
+        }
```
dougbu: Why isn't this method in `HtmlTextWriter`?

src/Microsoft.AspNet.Mvc.ViewFeatures/HtmlTextWriter.cs
```diff
@@ -0,0 +1,50 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.IO;
+using Microsoft.AspNet.Html.Abstractions;
+
+namespace Microsoft.AspNet.Mvc.Razor
```
dougbu: Think `Microsoft.AspNet.Mvc.Rendering` would be much more appropriate. This class is not doing anything Razor-related.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/HtmlHelper.cs
```diff
@@ -748,7 +749,18 @@ protected virtual string GenerateDisplayText(ModelExplorer modelExplorer)
...
+                // out character by character.
+                var htmlWriter = ViewContext.Writer as HtmlTextWriter;
+                if (htmlWriter == null)
+                {
+                    tagBuilder.WriteTo(ViewContext.Writer, _htmlEncoder);
+                }
+                else
+                {
+                    htmlWriter.Write(tagBuilder);
+                }
```
dougbu: Move the special case down into `TagBuilder.WriteTo(...)` and pick up calls to that method from `RazorPage` for free. At least, similar code is necessary in `GenerateRouteForm()`.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/StringCollectionTextWriter.cs
```diff
@@ -9,17 +9,20 @@
 using Microsoft.AspNet.Html.Abstractions;
 using Microsoft.Framework.Internal;
 using Microsoft.Framework.WebEncoders;
+using Microsoft.AspNet.Mvc.Razor;
+using System.Diagnostics;
```
dougbu: `System` `using`s first

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/StringCollectionTextWriter.cs
```diff
@@ -9,17 +9,20 @@
...
 
 namespace Microsoft.AspNet.Mvc.Rendering
 {
     /// <summary>
-    /// A <see cref="TextWriter"/> that represents individual write operations as a sequence of strings.
+    /// A <see cref="HtmlTextWriter"/> that represents individual write operations as a sequence of
```
dougbu: suggest "represents" -> "stores"

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
rynowak: Sure. I'll update this to reflect that it's not optimal. The code should document every instance of existential peril experienced by its' authors.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
rynowak: No idea, I was preserving what the existing code does while correcting the use of `ToString()`.

summoning @NTaylorMullen

src/Microsoft.AspNet.Mvc.Razor/RazorTextWriter.cs
rynowak: It's abstract there. The implementation is supposed to 'handle' `IHtmlContent`. In this case we handle it by sometimes delegating, and sometimes writing it out.

src/Microsoft.AspNet.Mvc.ViewFeatures/HtmlTextWriter.cs
rynowak: Good catch. This should `.ViewFeatures` actually. `.Rendering` is for things the user uses by name in Razor code.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/HtmlHelper.cs
rynowak: I'm not sold on that being a better choice. I don't want to put special casey stuff in the `IHtmlContent` implementations, because `HtmlTextWriter` is an mvc.razor detail.

Will address the same special case in `GenerateRouteForm`

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/StringCollectionTextWriter.cs
rynowak: Fixed

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/StringCollectionTextWriter.cs
rynowak: UGH. Thanks, that whole 'represents' thing is the WORST.

rynowak:
Updated

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
NTaylorMullen: If you don't wrap it in an `HtmlString` any already encoded values (encoding can happen by passing an `@value` as part of an attribute value) that were added to the `valueBuffer` will be encoded a second time.

rynowak: Had an offline discussion - if this change is harmless, will take it. If not will leave it as-is.

rynowak:
@dougbu @NTaylorMullen - any more feedback?

dougbu:
> any more feedback?

What was the verdict after trying the change discussed in [this thread](https://github.com/aspnet/Mvc/pull/3137#discussion-diff-39587248)?

Please try it out (if you haven't) and :shipit: w/ that fix if you can. :shipit: as is if not.

dougbu:
On the other hand, looks like the CI machines can't compile `RazorTextWriter`.

rynowak:
> On the other hand, looks like the CI machines can't compile RazorTextWriter.

I don't pay attention to what Travis is doing on my PRs :smile_cat:

rynowak:
@dougbu - I had to correct one test that was doing `Assert.IsType<HtmlString>` - other than that smooth sailing so far.

PR #3145, opened 2015-09-16
Fix #3087 - use IHtmlContentBuilder in TagBuilder

### LINKED ISSUE #3087: Appending child tag builders to parent one
I see that you changed TagBuilder class and I can't find a way to append two child tag builder contents to a parent one or just two html contents into one. In MVC5 you could do:
```C#
public String CreateTagTree()
{
    TagBuilder firstChild = new TagBuilder("input");
    firstChild.MergeAttribute("type", "hidden");
    firstChild.MergeAttribute("name", "Ids");

    TagBuilder secondChild = new TagBuilder("input");
    secondChild.MergeAttribute("type", "hidden");
    secondChild.MergeAttribute("name", "Ids");

    TagBuilder parent = new TagBuilder("div");
    parent.InnerHtml =
        firstChild.ToString(TagRenderMode.SelfClosing) +
        secondChild.ToString(TagRenderMode.SelfClosing);

    return parent.ToString();
}
```

Is this scenario (will be) supported?

rynowak:
Improves authoring experience when using TagBuilder by taking advantage of
IHtmlContentBuilder and its extension methods.

TagBuilder.InnerHtml is no longer settable.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultDisplayTemplates.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/Html/DefaultDisplayTemplates.cs
@@ -52,13 +52,11 @@ namespace Microsoft.AspNet.Mvc.Rendering
             selectTag.AddCssClass("tri-state");
             selectTag.Attributes["disabled"] = "disabled";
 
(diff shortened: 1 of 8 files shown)


Eilon:
:shipit: The new design is so much cleaner!

PR #3147, opened 2015-09-16
Reacting to TestHost changes

kichalla:
@rynowak

### DIFF
--- a/test/Microsoft.AspNet.Mvc.Core.Test/ActionResults/RedirectToActionResultTest.cs
+++ b/test/Microsoft.AspNet.Mvc.Core.Test/ActionResults/RedirectToActionResultTest.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Threading.Tasks;
 using Microsoft.AspNet.Http;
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNet.Mvc.Core.Test/ActionResults/RedirectToActionResultTest.cs
```diff
@@ -59,7 +60,7 @@ public void RedirectToAction_Execute_ThrowsOnNullUrl()
             };
 
             // Act & Assert
-            ExceptionAssert.ThrowsAsync<InvalidOperationException>(
+            await ExceptionAssert.ThrowsAsync<InvalidOperationException>(
```
kichalla: The change here is because I was receiving a Roslyn compilation exception mentioning that this method could return even before this async call is finished(which is true)...but not sure why this exception only surfaces with my change...

rynowak: A fix is a fix. Are there any other warnings??

rynowak:
:shipit:

kichalla:
f3e537ddcb104b5d17dfa9df4161c6f0421657c8

Issue #2722, opened 2015-06-21
Expand `InputFormatterContext` or `IInputFormatter.ReadAsync()` to make errors explicit


dougbu:
Need a clear way for input formatters to indicate errors -- without `throw`ing. Handling errors more clearly fixes a bug due to gaps in the current approach.

Also need more tests covering cases where the `JsonInputFormatter` adds `ModelState` errors and other input formatters `throw`.

Currently most `IInputFormatter` implementations `throw` when they encounter an error condition. However the `JsonInputFormatter` adds `ModelState` errors in many cases. As seen in c4fa402 and mentioned in a PR [comment](https://github.com/aspnet/Mvc/pull/2695#discussion_r32681923), the obvious way to check for errors (comparing `ModelState.ErrorCount` before and after) is insufficient. Further `await formatter.ReadAsync(...) == null` is incorrect because the input may be explicitly `null`.

danroth27:
Throwing due to input seems problematic - generally we should return a 400 level response (not a 500) if the request data is bad (ex. invalid JSON or XML).

dougbu:
In this case, the thrown `Exception` is normally caught almost immediately. The `BodyModelBinder` executes input formatters within a `try` block with a `catch (Exception ex)` that adds `ex` to `ModelState`.

Also input formatters themselves don't usually `throw`. It's the lower-level serializers (`DataContractSerializer`, `JsonSerializer`, and `XmlSerializer` in the box) that do. We would need that `try` block in `BodyModelBinder` or most `IInputFormatter` implementations regardless. This bug is about `Exception`s not being the _only_ way to indicate errors.

rynowak:
> Throwing due to input seems problematic - generally we should return a 400 level response (not a 500) if the request data is bad (ex. invalid JSON or XML).

This bug tracks plumbing needed to turn the response into a 400 (errors in modelstate). As @dougbu mentioned, right now errors aren't being reported correctly unless you throw.

dougbu:
42017fa

PR #3157, opened 2015-09-18
Add `InputFormatterResult` and `InputFormatterContext.ModelName`

dougbu:
- #2722
- make communication of errors from formatters to `BodyModelBinder` explicit

nits:
- improve some doc comments
- add another `BodyValidationIntegrationTests` test

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Abstractions/Formatters/IInputFormatter.cs
+++ b/src/Microsoft.AspNet.Mvc.Abstractions/Formatters/IInputFormatter.cs
@@ -6,25 +6,26 @@ using System.Threading.Tasks;
 namespace Microsoft.AspNet.Mvc.Formatters
 {
     /// <summary>
-    /// Reads an object from the request body.
+    /// Reads an <c>object</c> from the request body.
     /// </summary>
     public interface IInputFormatter
     {
         /// <summary>
-        /// Determines whether this <see cref="IInputFormatter"/> can de-serialize
-        /// an object of the specified type.
+        /// Determines whether this <see cref="IInputFormatter"/> can deserialize an <c>object</c> of the
+        /// <paramref name="context"/>'s <see cref="InputFormatterContext.ModelType"/>.
         /// </summary>
-        /// <param name="context">Input formatter context associated with this call.</param>
-        /// <returns>True if this <see cref="IInputFormatter"/> supports the passed in
-        /// request's content-type and is able to de-serialize the request body.
-        /// False otherwise.</returns>
+        /// <param name="context">The <see cref="InputFormatterContext"/>.</param>
+        /// <returns>
+        /// <c>true</c> if this <see cref="IInputFormatter"/> can deserialize an <c>object</c> of the
+        /// <paramref name="context"/>'s <see cref="InputFormatterContext.ModelType"/>. <c>false</c> otherwise.
+        /// </returns>
         bool CanRead(InputFormatterContext context);
 
         /// <summary>
-        /// Called during deserialization to read an object from the request.
+        /// Reads an <c>object</c> from the request body.
         /// </summary>
-        /// <param name="context">Input formatter context associated with this call.</param>
-        /// <returns>A task that deserializes the request body.</returns>
-        Task<object> ReadAsync(InputFormatterContext context);
+        /// <param name="context">The <see cref="InputFormatterContext"/>.</param>
+        /// <returns>A task that on completion deserializes the request body.</returns>
+        Task<InputFormatterResult> ReadAsync(InputFormatterContext context);
     }
 }
--- a/src/Microsoft.AspNet.Mvc.Abstractions/Formatters/IInputFormatter.cs
+++ b/src/Microsoft.AspNet.Mvc.Abstractions/Formatters/IInputFormatter.cs
@@ -6,26 +6,26 @@ using System.Threading.Tasks;
 namespace Microsoft.AspNet.Mvc.Formatters
 {
     /// <summary>
-    /// Reads an <c>object</c> from the request body.
+    /// Reads an object from the request body.
     /// </summary>
     public interface IInputFormatter
     {
         /// <summary>
-        /// Determines whether this <see cref="IInputFormatter"/> can deserialize an <c>object</c> of the
+        /// Determines whether this <see cref="IInputFormatter"/> can deserialize an object of the
         /// <paramref name="context"/>'s <see cref="InputFormatterContext.ModelType"/>.
         /// </summary>
         /// <param name="context">The <see cref="InputFormatterContext"/>.</param>
         /// <returns>
-        /// <c>true</c> if this <see cref="IInputFormatter"/> can deserialize an <c>object</c> of the
+        /// <c>true</c> if this <see cref="IInputFormatter"/> can deserialize an object of the
         /// <paramref name="context"/>'s <see cref="InputFormatterContext.ModelType"/>. <c>false</c> otherwise.
         /// </returns>
         bool CanRead(InputFormatterContext context);
 
         /// <summary>
-        /// Reads an <c>object</c> from the request body.
+        /// Reads an object from the request body.
         /// </summary>
         /// <param name="context">The <see cref="InputFormatterContext"/>.</param>
-        /// <returns>A task that on completion deserializes the request body.</returns>
+        /// <returns>A <see cref="Task"/> that on completion deserializes the request body.</returns>
         Task<InputFormatterResult> ReadAsync(InputFormatterContext context);
     }
 }
--- a/src/Microsoft.AspNet.Mvc.Abstractions/Formatters/InputFormatterResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Abstractions/Formatters/InputFormatterResult.cs
@@ -1,0 +1,93 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNet.Mvc.Formatters
+{
+    /// <summary>
+    /// Result of a <see cref="IInputFormatter.ReadAsync"/> operation.
+    /// </summary>
+    public class InputFormatterResult
+    {
+        private static readonly InputFormatterResult _failed = new InputFormatterResult();
+        private static readonly Task<InputFormatterResult> _failedAsync = Task.FromResult(_failed);
+
+        private InputFormatterResult()
+        {
+            HasError = true;
+        }
+
+        private InputFormatterResult(object model)
+        {
+            Model = model;
+        }
+
+        /// <summary>
+        /// Gets an indication whether the <see cref="IInputFormatter.ReadAsync"/> operation had an error.
+        /// </summary>
+        public bool HasError { get; }
+
+        /// <summary>
+        /// Gets the deserialized <see cref="object"/>.
+        /// </summary>
+        /// <value>
+        /// <c>null</c> if <see cref="HasError"/> is <c>true</c>.
+        /// </value>
+        public object Model { get; }
+
+        /// <summary>
+        /// Returns an <see cref="InputFormatterResult"/> indicating the <see cref="IInputFormatter.ReadAsync"/>
+        /// operation failed.
+        /// </summary>
+        /// <returns>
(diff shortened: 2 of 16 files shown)


src/Microsoft.AspNet.Mvc.Abstractions/Formatters/IInputFormatter.cs
```diff
@@ -6,25 +6,26 @@
...
     /// <summary>
-    /// Reads an object from the request body.
+    /// Reads an <c>object</c> from the request body.
     /// </summary>
     public interface IInputFormatter
     {
         /// <summary>
-        /// Determines whether this <see cref="IInputFormatter"/> can de-serialize
-        /// an object of the specified type.
+        /// Determines whether this <see cref="IInputFormatter"/> can deserialize an <c>object</c> of the
```
rynowak: picking nits here, but this is referring to object as a generic programming term, not as the specific class :laughing:

src/Microsoft.AspNet.Mvc.Abstractions/Formatters/InputFormatterResult.cs
```diff
@@ -0,0 +1,93 @@
...
+        /// <summary>
+        /// Returns an <see cref="InputFormatterResult"/> indicating the <see cref="IInputFormatter.ReadAsync"/>
+        /// operation was successful.
+        /// </summary>
+        /// <param name="model">The deserialized <see cref="object"/>.</param>
+        /// <returns>
+        /// An <see cref="InputFormatterResult"/> indicating the <see cref="IInputFormatter.ReadAsync"/>
+        /// operation succeeded i.e. with <see cref="HasError"/> <c>false</c>.
+        /// </returns>
+        public static InputFormatterResult Successful(object model)
```
rynowak: Can we make this `Success` instead?

src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
```diff
@@ -67,42 +67,48 @@ public virtual bool CanRead(InputFormatterContext context)
...
 
         /// <summary>
-        /// Returns a value indicating whether or not the given type can be read by this serializer.
+        /// Determines whether this <see cref="InputFormatter"/> can deserialize an <c>object</c> of the given
+        /// <paramref name="type"/>.
         /// </summary>
-        /// <param name="type">The type of object that will be read.</param>
-        /// <returns><c>true</c> if the type can be read, otherwise <c>false</c>.</returns>
+        /// <param name="type">The <see cref="Type"/> of <c>object</c> that will be read.</param>
+        /// <returns><c>true</c> if the <paramref name="type"/> can be read; <c>false</c> otherwise.</returns>
```
rynowak: This is a more obtuse way to phrase it, why the change?

src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
```diff
@@ -67,42 +67,48 @@ public virtual bool CanRead(InputFormatterContext context)
...
 
         /// <summary>
-        /// Reads the request body.
+        /// Reads an <c>object</c> from the request body.
         /// </summary>
-        /// <param name="context">The <see cref="InputFormatterContext"/> associated with the call.</param>
-        /// <returns>A task which can read the request body.</returns>
-        public abstract Task<object> ReadRequestBodyAsync(InputFormatterContext context);
+        /// <param name="context">The <see cref="InputFormatterContext"/>.</param>
+        /// <returns>A task that on completion deserializes the request body.</returns>
```
rynowak: `<see cref="Task....` ?

src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
```diff
@@ -126,7 +132,12 @@ protected Encoding SelectCharacterEncoding(MediaTypeHeaderValue contentType)
             }
 
             // No supported encoding was found so there is no way for us to start reading.
-            throw new InvalidOperationException(Resources.FormatInputFormatterNoEncoding(GetType().FullName));
+            return null;
+        }
+
+        protected string GetNoEncodingMessage()
+        {
+            return Resources.FormatInputFormatterNoEncoding(GetType().FullName);
```
rynowak: This is a wierd method to expose on the base class. I agree in general that throwing from `SelectCharacterEncoding` isn't ideal for extensibility - but why this change? What problem do you need to solve?

src/Microsoft.AspNet.Mvc.Formatters.Json/JsonInputFormatter.cs
```diff
@@ -49,16 +49,21 @@ public JsonSerializerSettings SerializerSettings
...
 
         /// <inheritdoc />
-        public override Task<object> ReadRequestBodyAsync([NotNull] InputFormatterContext context)
+        public override Task<InputFormatterResult> ReadRequestBodyAsync([NotNull] InputFormatterContext context)
         {
             var type = context.ModelType;
             var request = context.HttpContext.Request;
-            MediaTypeHeaderValue requestContentType = null;
+
+            MediaTypeHeaderValue requestContentType;
```
rynowak: :dancer:

src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
dougbu: The changes here are limited to the references e.g. "type" -> `<see cref="Type"/>` and `otherwise <c>false</c>` -> `<c>false</c> otherwise` i.e. consistent with the first phrase. What's obtuse now? In other words and with no judgement for you or anyone else, are you repeating your `<see cref="object"/>` nitpick?

src/Microsoft.AspNet.Mvc.Formatters.Json/JsonInputFormatter.cs
```diff
@@ -66,31 +71,53 @@ public override Task<object> ReadRequestBodyAsync([NotNull] InputFormatterContex
...
                     var exception = e.ErrorContext.Error;
-                    context.ModelState.TryAddModelError(e.ErrorContext.Path, e.ErrorContext.Error);
+
+                    // Handle path combinations such as "" + "Property", "Parent" + "Property", or "Parent" + "[12]".
+                    var key = e.ErrorContext.Path;
+                    if (!string.IsNullOrEmpty(context.ModelName))
+                    {
+                        if (string.IsNullOrEmpty(e.ErrorContext.Path) || e.ErrorContext.Path[0] == '[')
+                        {
+                            key = context.ModelName + e.ErrorContext.Path;
```
rynowak: It's a little wierd to parse the `IsNullOrEmpty` case falling into here. I know it's a no-op, it just seems overly clever

src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
dougbu: So you've now seen the problem I solved here? Let's discuss alternatives tomorrow.

test/Microsoft.AspNet.Mvc.Formatters.Json.Test/JsonInputFormatterTest.cs
```diff
@@ -38,7 +38,11 @@ public void CanRead_ReturnsTrueForAnySupportedContentType(string requestContentT
             var contentBytes = Encoding.UTF8.GetBytes("content");
 
             var httpContext = GetHttpContext(contentBytes, contentType: requestContentType);
-            var formatterContext = new InputFormatterContext(httpContext, new ModelStateDictionary(), typeof(string));
```
rynowak: Is there adequate coverage of the cases around `ModelName + Path` that crop up? I can't tell from the diff

src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
rynowak: Yep. I'm sleepy but I realized you need this to not-throw so you can use the validation error instead. 

I actually feel a little bit like just living with duplication of the message is a better choice than the putting this turd here.

rynowak:
:watch: this is much cleaner than I expected. Very nice!

src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
rynowak: Subtracting the markup:

> true if the type can be read, otherwise false.
> true if the type can be read; false otherwise.

A lot of these doc changes are improvements, but very tiny ones

src/Microsoft.AspNet.Mvc.Abstractions/Formatters/IInputFormatter.cs
dougbu: Agreed, that's why I used `<c>object</c>` instead of `<see cref="object"/>`. I can go back to "object" if you prefer. ??

src/Microsoft.AspNet.Mvc.Abstractions/Formatters/InputFormatterResult.cs
dougbu: Sure. And `Failure()` for the alternative -- to be consistent.

src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
dougbu: I was changing some and chose to change all. Here, do you prefer the original comma and final word order?

src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
dougbu: Should be. Thanx.

src/Microsoft.AspNet.Mvc.Formatters.Json/JsonInputFormatter.cs
dougbu: :ok: I'll add a block just for that case and do the `else if` thing.

test/Microsoft.AspNet.Mvc.Formatters.Json.Test/JsonInputFormatterTest.cs
dougbu: Good catch. I meant to vary the new `FromBodyAndRequiredOnInnerValueTypeProperty_NotBound_JsonFormatterSuccessful()` until I got deserialization until it failed lower down but didn't go back and finish after rebasing. Will add another test or two here or in `BodyValidationIntegrationTests`...

src/Microsoft.AspNet.Mvc.Abstractions/Formatters/IInputFormatter.cs
rynowak: I'm not really sure what the right thing notationally is to do here - I just wanted to point out my concern. I trust whatever you think is the right way.

src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
rynowak: yes

src/Microsoft.AspNet.Mvc.Abstractions/Formatters/IInputFormatter.cs
Eilon: It should have no notation at all really. It's just a plain English word.

dougbu:
updated

src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
```diff
@@ -126,7 +136,11 @@ protected Encoding SelectCharacterEncoding(MediaTypeHeaderValue contentType)
             }
 
             // No supported encoding was found so there is no way for us to start reading.
-            throw new InvalidOperationException(Resources.FormatInputFormatterNoEncoding(GetType().FullName));
+            context.ModelState.TryAddModelError(
+                context.ModelName,
+                Resources.FormatInputFormatterNoEncoding(GetType().FullName));
```
rynowak: yay

rynowak:
:shipit:

dougbu:
42017fa

PR #3162, opened 2015-09-18
[quick fix] Allow `TestCommon` project to build successfully

dougbu:
- otherwise may fail in VS or when using `BuildV2` from the command line


pranavkm:
:shipit:

dougbu:
0476d53

PR #3167, opened 2015-09-21
Work around aspnet/External#33, aspnet/External#41, aspnet/External#42, and aspnet/External#43

dougbu:
- do not run tests that hit known issues with Core CLR on Linux


rynowak:
:shipit: 

I'm not a fan of how the cases that don't work on linux+coreclr aren't 'skipped' so they make noise in the logs - however, I don't have a better idea that doesn't result in lots more work, so oh well... :cry:

dougbu:
895258d

PR #3169, opened 2015-09-21
[Perf] Remove use of [FromServices] on Controller

rynowak:
[FromServices] requires modelbinding to run for each of these four
properties, which allocates a lot, and ultimately just ends up calling
GetRequiredService in the end.

Also, retrieving these services is now lazy, which should be very
beneficial as few of them aren't used often.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
@@ -14,6 +14,7 @@ using Microsoft.AspNet.Mvc.ModelBinding;
 using Microsoft.AspNet.Mvc.ModelBinding.Validation;
 using Microsoft.AspNet.Mvc.ViewFeatures;
 using Microsoft.AspNet.Routing;
+using Microsoft.Framework.DependencyInjection;
 using Microsoft.Framework.Internal;
 using Microsoft.Net.Http.Headers;
 using Newtonsoft.Json;
@@ -25,9 +26,13 @@ namespace Microsoft.AspNet.Mvc
     /// </summary>
     public abstract class Controller : IActionFilter, IAsyncActionFilter, IDisposable
     {
+        private ActionContext _actionContext;
+        private IModelMetadataProvider _metadataProvider;
+        private IObjectModelValidator _objectValidator;
+        private ITempDataDictionary _tempData;
+        private IUrlHelper _url;
         private DynamicViewData _viewBag;
         private ViewDataDictionary _viewData;
-        private ActionContext _actionContext;
 
         /// <summary>
         /// Gets the request-specific <see cref="IServiceProvider"/>.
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
```diff
@@ -131,17 +136,65 @@ public ActionContext ActionContext
...
+            {
+                if (_metadataProvider == null)
+                {
+                    _metadataProvider = Resolver?.GetRequiredService<IModelMetadataProvider>();
+                }
+
+                return _metadataProvider;
+            }
+
+            [param: NotNull]
```
pranavkm: I'm going to work on removing NotNull in Mvc today. Consider throwing here instead.

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
```diff
@@ -131,17 +136,65 @@ public ActionContext ActionContext
...
+            set
+            {
+                _url = value;
+            }
+        }
+
+        public IObjectModelValidator ObjectValidator
```
pranavkm: Docs

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
```diff
@@ -25,9 +26,13 @@ namespace Microsoft.AspNet.Mvc
     /// </summary>
     public abstract class Controller : IActionFilter, IAsyncActionFilter, IDisposable
     {
+        private ActionContext _actionContext;
```
pranavkm: Undo the move?

pranavkm:
:shipit:

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
rynowak: nowai

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
rynowak: I'd rather check this in before you get to the office and let your script deal with it :+1:

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
rynowak: sure

rynowak:
Here's perf results - allocations for 3001 requests - very basic site model binding 3 fields and showing a view.

**Before:**
image

Notice these three types:
- `ModelStateDictionary+<FindKeysWithPrefix>d_63`
- `Task<ModelBindingResult>`
- `ModelBindingContext`

Collectively about 8.9mb

**After:**
image

Our three amigos are much lower on the table now, about 4.5mb total for a savings of 4.5mb or about 1.5kb/request

davidfowl:
:shipit: Seems like we need to take another look at [FromServices] if it's just that inefficient or maybe that's just a gotcha people have to be aware of /cc @lodejard

rynowak:
It's both.

Using `[FromServices]` means more modelbinding, and modelbinding has a high cost. Our use in `Controller` meant that this small sample had to run model binding 5 times even though it had one parameter.

We should think about the ergonomics of `[FromServices]` **and** the cost of model binding. There are a few more improvements I'm planning here, but I won't be able to reduce the allocs more than say 50% without us rethinking the design more.

dougbu:
> the cost of model binding

The pretty pictures above indicate our users should reduce the number of action parameters and model-bound controller properties, not that they should avoid `[FromServices]`. Do we have any perf concerns at all specifically about `[FromServices]` _within_ a model?

> think about the ergonomics of `[FromServices]`

Agreed though this is separable from the perf discussion.

PR #3170, opened 2015-09-21
[Design][Perf] Hoist simple Expression<Func<T, U>> into fields in Razor

rynowak:
This change uses Roslyn to transform the syntax tree of a Razor page and hoist all **simple** expressions into static readonly fields.

For code like:
```
@Html.EditorFor(m => m.Occupation)
```

The c# compiler inserts a bunch of code calling into `System.Linq.Expressions` to construct and verify an `Expression<Func<T, U>>` matching the lambda being passed here. This runs **every** time the page is rendered, and allocates a lot of collections to verify that the types are valid.

The transformation done here looks like this:
```
public class __Generated_RazorPage
{
    private static readonly Expression<Func<Person, string>> __Hoist_1 = m => m.Occuptation;

    public Task RenderAsync()
    {
        ...
        Write(Html.EditorFor(__Hoist_1));
        ...
    }
}
```

When we encounter a case that we can't hoist, we simply ignore it. The 90% case in Razor is a property accessor or identity func. The only common case we can't handle is indexers, as usually the index is captured.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationService.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationService.cs
@@ -13,6 +13,7 @@ using Microsoft.AspNet.FileProviders;
 using Microsoft.AspNet.Mvc.Razor.Internal;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Emit;
 using Microsoft.Dnx.Compilation;
 using Microsoft.Dnx.Compilation.CSharp;
@@ -90,10 +91,13 @@ namespace Microsoft.AspNet.Mvc.Razor.Compilation
             var compilationOptions = compilationSettings.CompilationOptions
                                                         .WithOutputKind(OutputKind.DynamicallyLinkedLibrary);
 
-            var compilation = CSharpCompilation.Create(assemblyName,
-                        options: compilationOptions,
-                        syntaxTrees: new[] { syntaxTree },
-                        references: references);
+            var compilation = CSharpCompilation.Create(
+                assemblyName,
+                options: compilationOptions,
+                syntaxTrees: new[] { syntaxTree },
+                references: references);
+
+            compilation = Rewrite(compilation);
 
             using (var ms = new MemoryStream())
             {
@@ -138,6 +142,24 @@ namespace Microsoft.AspNet.Mvc.Razor.Compilation
                     return UncachedCompilationResult.Successful(type, compilationContent);
                 }
             }
+        }
+
+        private CSharpCompilation Rewrite(CSharpCompilation compilation)
+        {
+            var rewrittenTrees = new List<SyntaxTree>();
+            foreach (var tree in compilation.SyntaxTrees)
+            {
+                var semanticModel = compilation.GetSemanticModel(tree, ignoreAccessibility: true);
+                var walker = new WalkerTexasRanger(semanticModel);
+                walker.Visit(tree.GetRoot());
+
+                var sidekick = new JamesTrivette(semanticModel, walker.Expressions);
+                var rewrittenNode = (CSharpSyntaxNode)sidekick.Visit(tree.GetRoot());
+
+                rewrittenTrees.Add(tree.WithRootAndOptions(rewrittenNode, tree.Options));
+            }
+
+            return compilation.RemoveAllSyntaxTrees().AddSyntaxTrees(rewrittenTrees);
         }
 
         // Internal for unit testing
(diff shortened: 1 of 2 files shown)


rynowak:
Perf numbers - allocations for 3001 requests on a basic site with model binding 3 fields and rendering a view. The view uses 3 `LabelFor`s and 3 `EditorFors`.

**Before:**

Filtered to types in `System.Linq.Expressions`
image

Total Allocations:

image


**After:**

No allocations in `System.Linq.Expressions`

Total Allocations:
image


Overall, there's a reduction of 8.2mb of allocations from this change on a page with 6 uses of HtmlHelpers - that's about 2.7kb/request.

It's not easy to show where the savings come from in a simple graph as dotMemory doesn't show inclusive bytes on allocations - the savings is coming from a variety of collection types used in expression compilation and type checking.

src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationService.cs
```diff
@@ -300,5 +322,156 @@ private static string ReadFileContentsSafely(IFileProvider fileProvider, string
...
+            }
+
+            private SemanticModel SemanticModel { get; }
+
+            public List<SimpleLambdaExpressionSyntax> Expressions { get; } = new List<SimpleLambdaExpressionSyntax>();
+
+            public override void VisitSimpleLambdaExpression(SimpleLambdaExpressionSyntax node)
+            {
+                var type = SemanticModel.GetTypeInfo(node);
+                if (type.ConvertedType.Name != "Expression" && type.ConvertedType.ContainingAssembly.Name != "System.Core")
```
dougbu: `||` ?

src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationService.cs
```diff
@@ -300,5 +322,156 @@ private static string ReadFileContentsSafely(IFileProvider fileProvider, string
...
+                    var memberAccess = (MemberAccessExpressionSyntax)node;
+                    var lhs = memberAccess.Expression;
+                    return IsValidForHoisting(parameter, lhs);
+                }
+
+                return false;
+            }
+        }
+
+        public class JamesTrivette : CSharpSyntaxRewriter
```
dougbu: Too obscure. Beside's that why would the re-writer be a sidekick to the walker? The partnership here is more like a sniper (the re-writer) and his spotter (the walker). :smiley_cat:

src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationService.cs
```diff
@@ -300,5 +322,156 @@ private static string ReadFileContentsSafely(IFileProvider fileProvider, string
...
+            }
+
+            public IList<SimpleLambdaExpressionSyntax> Expressions { get; }
+
+            public SemanticModel SemanticModel { get; }
+
+            public override SyntaxNode VisitClassDeclaration(ClassDeclarationSyntax node)
+            {
+                var i = 0;
+                var memberDeclaractions = new List<MemberDeclarationSyntax>();
```
dougbu: wuz a declar-action?

src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationService.cs
Eilon: True story, my uncle worked on this show for a couple of years.

src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationService.cs
rynowak: This is a prototype. The discussion about the feasibility of the approach + whether or not it adversely affects the dev experience needs to happen first :laughing: 

That said, inside the compiler all you have are names. We should be looking for full name match here, but there isn't a concept of type-equivalence that applies other than matching names. Obvi matching the assembly name doesn't work, I just wanted to bang this out so we could measure the effect.

src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationService.cs
rynowak: Yeah, I dunno. It would feel wrong giving top billing to either CD or Alex either. Also, Jimmy **did** play for the Cowboys before that career ending injury.

I realize now you don't really need a separate pass for the walker, you could do all of this in one pass.. which would have avoided all of this.

Also, @Eilon - would your uncle appear as a special guest at one of our team meetings?

Eilon: @rynowak he lives in LA but is often travelling, so it is doubtful he could make an appearance any time soon.

NTaylorMullen:
Walked through in person all my concerns. Looks solid.

rynowak:
Closing this for now. We'll consider this change for RC.

PR #3176, opened 2015-09-21
Fix #3111 - Don't trounce content-type in VCR

### LINKED ISSUE #3111: ViewComponentResult should check if the response has already been sent before setting the ContentType
Short story:
I'm trying to combine multiple different ViewComponents in one single response. It renders the first viewcomponent but always fails on any succeeding call to vc.ExecuteResultAsync(context).

Long story:
I'm trying to combine multiple different ViewComponents from a controller. The ActionResult of all combined viewcomponents will be rendered to the browser. This is based on an article which does this with PartialViews and updates the PartialViews with ajax. That article is based on previous version of MVC. For more info see: https://www.simple-talk.com/dotnet/asp.net/revisiting-partial-view-rendering-in-asp.net-mvc/

After many hours I came to the following code example. But the problem is that it works only for the first viewComponent. When I change the order of viewcomponents it still renders the first one. So it doesn't seem to have anything with my viewcomponents. Always at second loop it ends at "vc.ExecuteResultAsync(context);" with no errors. So rendering the first one is always successful.

By the way I'm using VS 2015 Enterprise with Beta7 of MVC6 and all other dependencies.
Please see this stackoverflow question: http://stackoverflow.com/questions/32468238/combine-viewcomponents-in-asp-net-5-mvc-6

I guess it is in some way related to this issue: Fix #2338 - Add ViewComponentResult

```c#
// in controller class
public async Task<IActionResult> Dashboard()
    {
        // Combine multiple viewcomponents
        return new MultipleViewResult(
            ViewComponent(typeof(OrdersViewComponent))
            , ViewComponent(typeof(AccountsViewComponent))
            );
    }

    public class MultipleViewResult : ActionResult
    {
    public const string ChunkSeparator = "---|||---";
    public IList<ViewComponentResult> ViewComponentResults { get; private set; }
    public MultipleViewResult(params ViewComponentResult[] views)
    {
        if (ViewComponentResults == null)
        {
            ViewComponentResults = new List<ViewComponentResult>();
        }

        foreach (var v in views)
            ViewComponentResults.Add(v);
    }

    public override async Task ExecuteResultAsync(ActionContext context)
    {
        if (context == null)

        byte[] chunkSeparatorBytes = System.Text.Encoding.UTF8.GetBytes(ChunkSeparator);
        var total = ViewComponentResults.Count;
        for (var index = 0; index < total; index++)
        {
            var vc = ViewComponentResults[index];

            // No matter which viewcomponent, this line works only with the first viewcomponent.
            await vc.ExecuteResultAsync(context);

            if (index < total - 1)
            {
                await context.HttpContext.Response.Body.WriteAsync(chunkSeparatorBytes, 0, chunkSeparatorBytes.Length);
            }
        }
    }
}
```

rynowak:
This change makes ViewComponentResult respect an existing Content-Type
setting on the HttpResponse. Code is very similar to the code in
ViewExector, and includes the same quirks.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponentResult.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponentResult.cs
@@ -80,22 +80,32 @@ namespace Microsoft.AspNet.Mvc
                 viewData = new ViewDataDictionary(modelMetadataProvider, context.ModelState);
             }
 
-            var contentType = ContentType ?? ViewExecutor.DefaultContentType;
-            if (contentType.Encoding == null)
+            var contentType = ContentType;
+            if (contentType != null && contentType.Encoding == null)
             {
                 // Do not modify the user supplied content type, so copy it instead
                 contentType = contentType.Copy();
                 contentType.Encoding = Encoding.UTF8;
             }
 
+            // Priority list for setting content-type/encoding:
+            //      1. this.ContentType (set by the user on the result)
+            //      2. response.ContentType (likely set by the user in controller code)
+            //      3. ViewExecutor.DefaultContentType (sensible default)
+            //
+            //
+            response.ContentType = 
+                contentType?.ToString() ?? 
+                response.ContentType ??
+                ViewExecutor.DefaultContentType.ToString();
+
             if (StatusCode != null)
             {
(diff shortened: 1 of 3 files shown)


Eilon:
Nice :shipit:

Issue #3164, opened 2015-09-19
ModelState Generic Remove


mehran63:
A generic version of remove could be more helpful and less error-prone than string hard-coded parametric version as mention here:

https://stackoverflow.com/questions/14008561/is-there-a-strongly-named-way-to-remove-modelstate-errors-in-asp-net-mvc/14008604#14008604

I'm ready to implement it.

dougbu:
Are you suggesting a thin veneer over `ModelBindingHelper.ClearValidationStateForModel(Type, ModelStateDictionary, IModelMetadataProvider, string)`? Or is that method sufficient as-is for your purpose?

Muchiachio:
Actually, having generic add and remove methods for model state would be great.
I have an example on extension implementation [here](https://github.com/NonFactors/MVC5.Template/blob/develop/src/MvcTemplate.Components/Extensions/Mvc/ModelStateDictionaryExtensions.cs#L14) too.
This could look like this:
```C#
public ActionResult Create(Account account)
{
    if (account.EmailIsAlreadyUsed())
    {
        ModelState.AddModelError<Account>(model => model.Email, "Email is already being used.");

        return View(account);
    }

    return RedirectToAction("Index");
}
```

mehran63:
Yes, I mean so.

danroth27:
Sure, do you want to send us a PR with a design proposal? Please note that this feature will have some pretty significant test cost, so we need to take that into consideration as well.

Muchiachio:
I can if @mehran63 doesn't want to take this.
What do you mean by a significant test cost?
Last question, should this be implement as an extension (if yes which namespace?) or in same class?

mehran63:
I prefer to leave it to @Muchiachio .

rynowak:
@Muchiachio - it looks like you're the lucky winner! I'd start with the following:

- Extension methods
- put extensions class in `Microsoft.AspNet.Mvc` in the `Microsoft.AspNet.Mvc.Core` project

Send a PR showing the API surface you want to add first so we can discuss.

Muchiachio:
Yes winner! :smile: But there is one problem, this functionality needs `ExpressionHelper.GetExpressionText(LambdaExpression)` from `ViewFeatures` projects which can not be included from the `Core` project, because of circular references :sob:

rynowak:
Ok, then put it in ViewFeatures for now. We might need to look at moving it.

dougbu:
[...]

PR #3182, opened 2015-09-22
Generic ModelStateDictionary add and remove extensions

Muchiachio:
This adds generic add model error and remove extension methods for ModelStateDictionary, to avoid using hard coded strings then specifying model state dictionary keys. #3164

There are few concerns what comes to mind
- Initial [thought](https://github.com/aspnet/Mvc/issues/3164#issuecomment-142053478) by @rynowak was to add this to `Microsoft.AspNet.Mvc.Core`, but I couldn't because of circular references, so for now it's in `Microsoft.AspNet.Mvc.ViewFeatures`.
- Will three method be enough for covering most of the model state dictionary usage cases?
- Does `Remove<TModel>(this ModelStateDictionary, Expression)` is clear enough, maybe extension method should be named RemoveKey?
- Is adding and removing empty key on expressions without normal text representation is correct and naturally expected behavior?

### DIFF
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/ModelStateDictionaryExtensions.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/ModelStateDictionaryExtensions.cs
@@ -2,10 +2,10 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Linq;
 using System.Linq.Expressions;
 using Microsoft.AspNet.Mvc.ModelBinding;
 using Microsoft.AspNet.Mvc.ViewFeatures;
-using System.Linq;
 
 namespace Microsoft.AspNet.Mvc
 {
@@ -67,11 +67,6 @@ namespace Microsoft.AspNet.Mvc
             if (string.IsNullOrEmpty(modelKey))
             {
                 var modelMetadata = new EmptyModelMetadataProvider().GetMetadataForType(typeof(TModel));
-                var elementMetadata = modelMetadata.ElementMetadata;
-                if (elementMetadata != null)
-                {
-                    modelMetadata = elementMetadata;
-                }
 
                 foreach (var property in modelMetadata.Properties)
                 {
@@ -95,9 +90,13 @@ namespace Microsoft.AspNet.Mvc
 
         private static string GetExpressionText(LambdaExpression expression)
         {
+            // We check if expression is wrapped with conversion to object expression
+            // and unwrap it if necessary, because Expression<Func<TModel, object>>
+            // automatically creates a convert to object expression for expresions
+            // returning value types
             var unaryExpression = expression.Body as UnaryExpression;
 
-            if (IsConvertibleToObject(unaryExpression))
+            if (IsConversionToObject(unaryExpression))
             {
                 return ExpressionHelper.GetExpressionText(Expression.Lambda(unaryExpression.Operand, expression.Parameters[0]));
             }
@@ -105,10 +104,10 @@ namespace Microsoft.AspNet.Mvc
             return ExpressionHelper.GetExpressionText(expression);
         }
 
-        private static bool IsConvertibleToObject(UnaryExpression expression)
+        private static bool IsConversionToObject(UnaryExpression expression)
         {
             return expression?.NodeType == ExpressionType.Convert &&
-                expression.Operand is MemberExpression &&
+                expression.Operand?.NodeType == ExpressionType.MemberAccess &&
                 expression.Type == typeof(object);
         }
     }
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/ModelStateDictionaryExtensions.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/ModelStateDictionaryExtensions.cs
@@ -15,13 +15,12 @@ namespace Microsoft.AspNet.Mvc
     {
         /// <summary>
         /// Adds the specified <paramref name="errorMessage"/> to the <see cref="ModelState.Errors"/> instance
-        /// that is associated with the specified <paramref name="expression"/> text key repressentation.
+        /// that is associated with the specified <paramref name="expression"/>.
         /// </summary>
         /// <typeparam name="TModel">The type of the model.</typeparam>
         /// <param name="modelState">The <see cref="ModelStateDictionary"/> instance this method extends.</param>
-        /// <param name="expression">The expression of <typeparamref name="TModel"/> from
-        /// which text representation will be used as a key of the <see cref="ModelState"/> to add errors to.</param>
-        /// <param name="errorMessage"></param>
+        /// <param name="expression">An expression to be evaluated against an item in the current model.</param>
+        /// <param name="errorMessage">The error message to add.</param>
         public static void AddModelError<TModel>(this ModelStateDictionary modelState, Expression<Func<TModel, object>> expression, string errorMessage)
         {
             modelState.AddModelError(GetExpressionText(expression), errorMessage);
@@ -29,28 +28,26 @@ namespace Microsoft.AspNet.Mvc
 
         /// <summary>
         /// Adds the specified <paramref name="exception"/> to the <see cref="ModelState.Errors"/> instance
-        /// that is associated with the specified <paramref name="expression"/> text key repressentation.
+        /// that is associated with the specified <paramref name="expression"/>.
         /// </summary>
         /// <typeparam name="TModel">The type of the model.</typeparam>
         /// <param name="modelState">The <see cref="ModelStateDictionary"/> instance this method extends.</param>
-        /// <param name="expression">The expression of <typeparamref name="TModel"/> from
-        /// which text representation will be used as a key of the <see cref="ModelState"/> to add errors to.</param>
-        /// <param name="exception">The error message to add.</param>
+        /// <param name="expression">An expression to be evaluated against an item in the current model.</param>
+        /// <param name="exception">The <see cref="Exception"/> to add.</param>
         public static void AddModelError<TModel>(this ModelStateDictionary modelState, Expression<Func<TModel, object>> expression, Exception exception)
         {
             modelState.AddModelError(GetExpressionText(expression), exception);
         }
 
         /// <summary>
-        /// Removes the specified <paramref name="expression"/> text key repressentation from the <see cref="ModelStateDictionary"/>.
+        /// Removes the specified <paramref name="expression"/> from the <see cref="ModelStateDictionary"/>.
         /// </summary>
         /// <typeparam name="TModel">The type of the model.</typeparam>
         /// <param name="modelState">The <see cref="ModelStateDictionary"/> instance this method extends.</param>
-        /// <param name="expression">The expression of <typeparamref name="TModel"/> from
-        /// which text representation will be used as a key of the <see cref="ModelState"/> to add errors to.</param>
+        /// <param name="expression">An expression to be evaluated against an item in the current model.</param>
         /// <returns>
         /// true if the element is successfully removed; otherwise, false.
-        /// This method also returns false if expression's text key was not found in the model-state dictionary.
+        /// This method also returns false if <paramref name="expression"/> was not found in the model-state dictionary.
         /// </returns>
         public static bool Remove<TModel>(this ModelStateDictionary modelState, Expression<Func<TModel, object>> expression)
         {
@@ -61,12 +58,15 @@ namespace Microsoft.AspNet.Mvc
         {
             var unaryExpression = expression.Body as UnaryExpression;
 
-            if (IsConvertToObject(unaryExpression))
+            if (IsConvertibleToObject(unaryExpression))
+            {
                 return ExpressionHelper.GetExpressionText(Expression.Lambda(unaryExpression.Operand, expression.Parameters[0]));
+            }
 
             return ExpressionHelper.GetExpressionText(expression);
         }
-        private static bool IsConvertToObject(UnaryExpression expression)
+
+        private static bool IsConvertibleToObject(UnaryExpression expression)
         {
             return expression?.NodeType == ExpressionType.Convert &&
                 expression.Operand is MemberExpression &&
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/ModelStateDictionaryExtensions.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/ModelStateDictionaryExtensions.cs
@@ -5,6 +5,7 @@ using System;
 using System.Linq.Expressions;
 using Microsoft.AspNet.Mvc.ModelBinding;
 using Microsoft.AspNet.Mvc.ViewFeatures;
+using System.Linq;
 
 namespace Microsoft.AspNet.Mvc
 {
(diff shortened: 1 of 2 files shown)


Eilon:
@Muchiachio that was fast, thanks for the PR!

@dougbu can you review this?

dougbu:
@Eilon yes

src/Microsoft.AspNet.Mvc.ViewFeatures/ModelStateDictionaryExtensions.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// <param name="modelState">The <see cref="ModelStateDictionary"/> instance this method extends.</param>
+        /// <param name="expression">The expression of <typeparamref name="TModel"/> from
+        /// which text representation will be used as a key of the <see cref="ModelState"/> to add errors to.</param>
+        /// <returns>
+        /// true if the element is successfully removed; otherwise, false.
+        /// This method also returns false if expression's text key was not found in the model-state dictionary.
+        /// </returns>
+        public static bool Remove<TModel>(this ModelStateDictionary modelState, Expression<Func<TModel, object>> expression)
+        {
+            return modelState.Remove(GetExpressionText(expression));
```
dougbu: This choice looks incorrect. With a POCO model for example, will remove only the entry for the top-level model and not for any bound property. Should use an approach similar to `ModelBindingHelper.ClearValidationStateForModel()` or perhaps provide both `Remove()` and `RemoveAll()` methods.

src/Microsoft.AspNet.Mvc.ViewFeatures/ModelStateDictionaryExtensions.cs
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        private static string GetExpressionText(LambdaExpression expression)
+        {
+            var unaryExpression = expression.Body as UnaryExpression;
+
+            if (IsConvertToObject(unaryExpression))
+                return ExpressionHelper.GetExpressionText(Expression.Lambda(unaryExpression.Operand, expression.Parameters[0]));
```
dougbu: Braces around this expression block please.

src/Microsoft.AspNet.Mvc.ViewFeatures/ModelStateDictionaryExtensions.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// <summary>
+        /// Adds the specified <paramref name="errorMessage"/> to the <see cref="ModelState.Errors"/> instance
+        /// that is associated with the specified <paramref name="expression"/> text key repressentation.
+        /// </summary>
+        /// <typeparam name="TModel">The type of the model.</typeparam>
+        /// <param name="modelState">The <see cref="ModelStateDictionary"/> instance this method extends.</param>
+        /// <param name="expression">The expression of <typeparamref name="TModel"/> from
+        /// which text representation will be used as a key of the <see cref="ModelState"/> to add errors to.</param>
+        /// <param name="errorMessage"></param>
+        public static void AddModelError<TModel>(this ModelStateDictionary modelState, Expression<Func<TModel, object>> expression, string errorMessage)
```
dougbu: Please use the style of most other classes in this repo: wrap parameters on their own lines when the method declaration would otherwise be significantly longer than 120 chars.

src/Microsoft.AspNet.Mvc.ViewFeatures/ModelStateDictionaryExtensions.cs
```diff
@@ -0,0 +1,76 @@
...
+    {
+        /// <summary>
+        /// Adds the specified <paramref name="errorMessage"/> to the <see cref="ModelState.Errors"/> instance
+        /// that is associated with the specified <paramref name="expression"/> text key repressentation.
+        /// </summary>
+        /// <typeparam name="TModel">The type of the model.</typeparam>
+        /// <param name="modelState">The <see cref="ModelStateDictionary"/> instance this method extends.</param>
+        /// <param name="expression">The expression of <typeparamref name="TModel"/> from
+        /// which text representation will be used as a key of the <see cref="ModelState"/> to add errors to.</param>
+        /// <param name="errorMessage"></param>
```
dougbu: Don't invent new terms for this documentation, especially not "text key representation". Suggest being consistent with what we have in https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/IHtmlHelperOfT.cs That includes simply saying `specified <paramref name="expression"/>` and documenting the parameter as `An expression to be evaluated against an item in the current model.`. (Realize the expression won't be executed but it is "evaluated" in a very general sense.)

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ModelStateDictionaryExtensionsTest.cs
```diff
@@ -0,0 +1,241 @@
...
+using System.Linq;
+using Microsoft.AspNet.Mvc.ModelBinding;
+using Xunit;
+
+namespace Microsoft.AspNet.Mvc
+{
+    public class ModelStateDictionaryExtensionsTest
+    {
+        [Fact]
+        public void AddsModelErrorMessage_ForSingleExpression()
```
dougbu: We try to include expectations in test names. So here the name should be `AddModelError_ForSingleExpression_AddsExpectedMessage`.

src/Microsoft.AspNet.Mvc.ViewFeatures/ModelStateDictionaryExtensions.cs
```diff
@@ -0,0 +1,76 @@
...
+        private static string GetExpressionText(LambdaExpression expression)
+        {
+            var unaryExpression = expression.Body as UnaryExpression;
+
+            if (IsConvertToObject(unaryExpression))
+                return ExpressionHelper.GetExpressionText(Expression.Lambda(unaryExpression.Operand, expression.Parameters[0]));
+
+            return ExpressionHelper.GetExpressionText(expression);
+        }
+        private static bool IsConvertToObject(UnaryExpression expression)
```
dougbu: Add a blank line above this.

dougbu: Also rename to `IsConvertibleToObject`

dougbu:
:watch:
Note CI checks are unrelated to your changes. Should settle down once we update the MyGet feed.

src/Microsoft.AspNet.Mvc.ViewFeatures/ModelStateDictionaryExtensions.cs
Muchiachio: > will remove only the entry for the top-level model and not for any bound property.

Could you elaborate a bit more on this scenario? Because I couldn't completely understand what you meant here.

dougbu: Say the current model type is
``` c#
public class MyModel
{
   int A { get; set; }
   int B { get; set; }
   int C { get; set; }
   int D { get; set; }
}
```
Pass `model => model` to this method and it will remove only the `ModelStateDictionary` entry with `Key == string.Empty`. It ignores entries with names "A", "B", "C" and "D".

I'm suggesting users need an option to clean up everything for the given expression. `RemoveAll()` would provide this.

Muchiachio:
@dougbu I updated PR with your comments.
I took similar approach to `ModelBindingHelper.ClearValidationStateForModel`, but empty key would not delete all keys, just the model property ones.

src/Microsoft.AspNet.Mvc.ViewFeatures/ModelStateDictionaryExtensions.cs
```diff
@@ -0,0 +1,115 @@
...
+        {
+            string modelKey = GetExpressionText(expression);
+            if (string.IsNullOrEmpty(modelKey))
+            {
+                var modelMetadata = new EmptyModelMetadataProvider().GetMetadataForType(typeof(TModel));
+                var elementMetadata = modelMetadata.ElementMetadata;
+                if (elementMetadata != null)
+                {
+                    modelMetadata = elementMetadata;
+                }
```
dougbu: I can't think of a case where property names of an element will appear in the `ModelStateDictionary` at the top level. A top-level collection of simple elements e.g. `string[]` might be bound as `string.Empty` (though HTML generation disallows the empty key) but will normally appear with prefixes such as `"[0]"`.

Suggest removing this block for now. Worst case is the block below will attempt to clean out entries with the `"Count"`, `"IsReadOnly"` or similar prefix. File a follow-up issue to find all collection prefixes and do the right thing. For example, we could eventually find the collection prefixes borrowing code from `JQueryFormValueProviderFactory.NormalizeJQueryToMvc()`. No incredible rush -- top-level collections are unusual.

src/Microsoft.AspNet.Mvc.ViewFeatures/ModelStateDictionaryExtensions.cs
```diff
@@ -0,0 +1,115 @@
...
+                return ExpressionHelper.GetExpressionText(Expression.Lambda(unaryExpression.Operand, expression.Parameters[0]));
+            }
+
+            return ExpressionHelper.GetExpressionText(expression);
+        }
+
+        private static bool IsConvertibleToObject(UnaryExpression expression)
+        {
+            return expression?.NodeType == ExpressionType.Convert &&
+                expression.Operand is MemberExpression &&
```
dougbu: `expression.Operand?.NodeType == ExpressionType.MemberAccess`

src/Microsoft.AspNet.Mvc.ViewFeatures/ModelStateDictionaryExtensions.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            if (IsConvertibleToObject(unaryExpression))
+            {
+                return ExpressionHelper.GetExpressionText(Expression.Lambda(unaryExpression.Operand, expression.Parameters[0]));
+            }
+
+            return ExpressionHelper.GetExpressionText(expression);
+        }
+
+        private static bool IsConvertibleToObject(UnaryExpression expression)
```
dougbu: Sorry, my name suggestion didn't make sense. Prefer `IsConversionToObject()`.

src/Microsoft.AspNet.Mvc.ViewFeatures/ModelStateDictionaryExtensions.cs
```diff
@@ -0,0 +1,115 @@
...
+                foreach (var entry in entries)
+                {
+                    modelState.Remove(entry.Key);
+                }
+            }
+        }
+
+        private static string GetExpressionText(LambdaExpression expression)
+        {
+            var unaryExpression = expression.Body as UnaryExpression;
```
dougbu: Add code comments informing reader that this is about ignoring an outermost conversion to `object`.

Forgot to ask last time: What scenarios hit this case? We don't have similar handling elsewhere. If it's an important scenario, probably need to handle it at a lower level instead.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ModelStateDictionaryExtensionsTest.cs
```diff
@@ -0,0 +1,336 @@
...
+        public void AddModelError_ForSingleExpression_AddsExpectedMessage()
+        {
+            // Arrange
+            var modelState = new ModelStateDictionary();
+
+            // Act
+            modelState.AddModelError<TestModel>(model => model.Text, "Message");
+
+            // Assert
+            Assert.Equal("Text", modelState.Single().Key);
```
dougbu: FYI only: `Assert.Single()` results in significantly better test failures than the `Exception` Linq's `Single()` throws. If you're kind enough to tackle other issues in the future, do this as `Assert.Equal("Text", Assert.Single(modelState).Key);`. Or use a temporary variable for the `Assert.Single()` result so that each assertion is on a single line and the exact failure is easier to find.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ModelStateDictionaryExtensionsTest.cs
```diff
@@ -0,0 +1,336 @@
...
+
+        [Fact]
+        public void AddModelError_ForExplicitlyCastedToObjectExpression_AddsExpectedException()
+        {
+            // Arrange
+            var exception = new Exception();
+            var modelState = new ModelStateDictionary();
+
+            // Act
+            modelState.AddModelError<TestModel>(model => (object)model.Child.Value, exception);
```
dougbu: Note this expression is not supported elsewhere in MVC. The leftmost expression can't be a cast.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ModelStateDictionaryExtensionsTest.cs
```diff
@@ -0,0 +1,336 @@
...
+
+        [Fact]
+        public void AddModelError_ForExpressionWithoutStringRepresentation_AddsExpectedException()
+        {
+            // Arrange
+            var exception = new Exception();
+            var modelState = new ModelStateDictionary();
+
+            // Act
+            modelState.AddModelError<TestModel>(model => model.ToString(), exception);
```
dougbu: This expression is also not supported in MVC.

src/Microsoft.AspNet.Mvc.ViewFeatures/ModelStateDictionaryExtensions.cs
Muchiachio: It's all about not having to specify to two generic types then calling an extension method, so we don't have to have boilerplate code like this:
`ModelState.AddModelError<Account, *int*>(model => model.Age, "You are too young.")`
And instead can write
`ModelState.AddModelError<Account>(model => model.Age, "You are too young.")`
which on structs result in automatic cast to object, because
`Expression<Func<TModel, object>> expression` converts stuff like `model => model.Age` to `model => Convert(model.Age)` which can not be converted to proper text by `ExpressionHelper.GetExpressionText` method.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ModelStateDictionaryExtensionsTest.cs
Muchiachio: Check my comment above, about automatic struct casting to object. I just wanted to explicitly add this, so that it would be visible. I can remove the cast if you think it's okay.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ModelStateDictionaryExtensionsTest.cs
Muchiachio: Definitely will change that to `Assert.Single` :+1:

src/Microsoft.AspNet.Mvc.ViewFeatures/ModelStateDictionaryExtensions.cs
dougbu: If this is needed only when the expression starts with `(object)` explicitly, please remove. If it's required because Linq generates a `Convert` node for a boxing operation or other implicit conversion, :ok:

Muchiachio: Yes, Linq generates a `Convert` node. I will add a comment describing an issue this code block is solving.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ModelStateDictionaryExtensionsTest.cs
dougbu: Suggest removing the tests involving explicit casts and `ToString()`. The implicit cast test seems sufficient.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ModelStateDictionaryExtensionsTest.cs
```diff
@@ -0,0 +1,336 @@
...
+using System.Linq;
+using Microsoft.AspNet.Mvc.ModelBinding;
+using Xunit;
+
+namespace Microsoft.AspNet.Mvc
+{
+    public class ModelStateDictionaryExtensionsTest
+    {
+        [Fact]
+        public void AddModelError_ForSingleExpression_AddsExpectedMessage()
```
dougbu: Optional but may be worth testing field or variable references in the expression e.g.
``` c#
  var something = "whatever value";
  modelState.AddModelError<TestModel>(model => something, ...);
```
The above would add an error with key "something" and it's a supported case.

dougbu:
@Muchiachio looking quite good. The current checks appear to be failing for reasons unrelated to your changes -- should clear up soon.

dougbu:
:watch:

Muchiachio:
I updated the code based on our conversation, and replaced explicit and `.ToString()` tests to variable reference ones. Seems like I didn't get a CLA check.

dougbu:
Cycling closed / open to see if the CLA bot wakes up...

dougbu:
:shipit: I'll get this in soon.

dougbu:
squashed to b28debf with slight touch-up in eff10cd (move to same namespace as `ModelStateDictionary`)

PR #3183, opened 2015-09-22
Asynchronously flush the HttpResponseStreamWriter once a view has been rendered.

pranavkm:
This solves a perf issue for views which produce content that is smaller
than the buffer size of HttpResponseStreamWriter. In this case, the writer
ends up synchronously writing to the Response as part of Dispose which
affects perf.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/ViewExecutor.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/ViewExecutor.cs
@@ -62,6 +62,9 @@ namespace Microsoft.AspNet.Mvc.ViewFeatures
                     htmlHelperOptions);
 
                 await view.RenderAsync(viewContext);
+                // Invoke FlushAsync to ensure any buffered content is asynchronously written to the underlying response. In the absence
+                // of this line, the buffer gets synchronously written to the response as part of the Dispose which has a perf impact.
+                await writer.FlushAsync();
             }
         }
     }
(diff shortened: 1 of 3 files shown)


pranavkm:
cc @rynowak

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/ViewExecutor.cs
```diff
@@ -62,6 +62,7 @@ public static class ViewExecutor
                     htmlHelperOptions);
 
                 await view.RenderAsync(viewContext);
+                await writer.FlushAsync();
```
rynowak: This deserves a comment explaining why this is necessary, and at least a unit test. /grumpy

pranavkm:
Updated.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/ViewExecutor.cs
```diff
@@ -62,6 +62,9 @@ public static class ViewExecutor
                     htmlHelperOptions);
 
                 await view.RenderAsync(viewContext);
+                // Invoke FlushAsync to ensure any buffered content is asynchronously written to the underlying response. In the absence
+                // of this line, the buffer gets synchronously written to the response as part of the Dispose which has a perf impact.
+                await writer.FlushAsync();
```
rynowak: Lines look a little long here

rynowak:
:shipit:

PR #3186, opened 2015-09-23
Allow @model specified in a page work with @inherits directive in _ViewImports

### LINKED ISSUE #3144: Specifying @model in a page ignores value specified by @inherit from a _ViewImport
We have code that prevents `@model` and `@inherits` from being specified in the same page. However this check happens at parse time. Consequently specifying an `@inherits MyBase<TModel>` in _ViewImports and `@model CustomModel` doesn't produce an error. Instead, the `@model` wins and resets the base type to the default (`RazorPage<>`).

This seems like a reasonable thing to do - we shouldn't error on it.

Related: https://github.com/aspnet/Mvc/issues/3143

pranavkm:
Fixes #3144


pranavkm:
cc @NTaylorMullen

test/Microsoft.AspNet.Mvc.Razor.Host.Test/TestFiles/Output/DesignTime/ModelExpressionTagHelper.cs
```diff
@@ -8,13 +8,7 @@ namespace Asp
     using Microsoft.AspNet.Mvc.Rendering;
     using System.Threading.Tasks;
 
-    public class ASPV_TestFiles_Input_ModelExpressionTagHelper_cshtml : Microsoft.AspNet.Mvc.Razor.RazorPage<
-#line 1 "TestFiles/Input/ModelExpressionTagHelper.cshtml"
```
pranavkm: Discussed with @NTaylorMullen offline - I'll fix the missing line mappings for this in a follow up PR.

NTaylorMullen:
Gotta fix travis :watch:

pranavkm:
Updated

NTaylorMullen:
:shipit: once @rynowak has addressed the build issue.

Issue #3158, opened 2015-09-18
Create a common abstraction for Application/Action/Controller -Models


Grinderofl:
Currently in MVC there exists 3 types of application models:
`ApplicationModel`, `ControllerModel`, `ActionModel`

All of the models share some common properties, like `ApiExplorer`, `Filters`, and `Properties`. I would like for the models to inherit from a `ModelBase` that has `ApiExplorer`, `Filters` and `Properties` as base. This would facilitate support for creating generic conventions that wouldn't necessarily have to apply to one specific model type.

Sample use case:
I have added the functionality to MVC to register filters programmatically without using attributes, via fluent interface. To achieve this functionality, I have a generic base FilterModelConvention, with three methods:
`public void Apply(TModel model)`
`protected abstract IEnumerable<IFilterRegistration> CreateRegistrations(TModel model);`
`protected abstract Func<TModel, IList<IFilterMetadata>> CollectionPropertySelector()`

In the end, the `Apply` method does this:
`CollectionPropertySelector()(model).Add(filter);`

I would much rather be able to specify `TModel` as a generic type ModelConvention, and have in `Apply`

`model.Filters.Add(filter);`

Resolving these sorts of problems is tricky, as first one has to figure out the smallest common denominator to start building up from - more difficult for those that don't use the more complex language features on a daily basis. Having an abstract base that has at least the most basic properties makes a good starting point, and looking at codebase I can't find any immediate pitfalls to this approach - however would make developers' lives better (by keeping us sane by not forcing us to fall back on reflection or write same code three times...).

If this is something that you would consider having in there, but is not in the team's sights for near future, I'm willing to implement and PR, as the parallel development of our internal kit could make huge use of this approach.

danroth27:
@rynowak

rynowak:
I really like the idea of this. 

> Having an abstract base that has at least the most basic properties makes a good starting point, and looking at codebase I can't find any immediate pitfalls to this approach - however would make developers' lives better (by keeping us sane by not forcing us to fall back on reflection or write same code three times...).

Agreed. I considered briefly adding a feature/facet paradigm to solve this, but it seems like overengineering. I think starting with the abc is the right thing to do.

Would like to come up with a slightly better name than `ModelBase` as `PropertyModel` and `ParameterModel` don't overlap with this.

Please go ahead and send a PR - ideally this week as we're locking for beta8 on Friday. I'd love to see some examples of what you're using this for when you're done.

[...]

PR #3188, opened 2015-09-23
Adds common interfaces around Models (implements #3158)

Grinderofl:
To facilitate support for creating more generic model conventions, separated the most common properties into interfaces and had appropriate classes implement them. This'll allow creating conventions like:

Implements #3158 

````c#
public abstract class FilterModelConvention<TModel> where TModel : ICommonModel
{
    public virtual void Apply(TModel model)
    {
        model.Filters.Add(...);
    }
}

public class ApplicationFilterModelConvention : FilterModelConvention<ApplicationModel>, IApplicationModelConvention
{
}
```

Originally suggested base class(es), but interfaces allow for more flexibility.


rynowak:
Since you went this way instead of the base-class, some feedback on how to make this more complete.

=====================

`ICommonModel` is too specific and doesn't really capture all of things that most models have. `IAttributeModel` on the other hand includes attributes, which all models except `ApplicationModel` have.

I'd suggest something like the following:
```
public interface ICommonModel
{
    IReadOnlyList<object> Attributes { get; }

    // Implement explicitly, keep existing properties for things like MethodInfo
    MemberInfo MemberInfo { get; } 

    // Implement explicitly, keep existing names for things like ActionName
    string Name { get; } 
    IDictionary<object, object> Properties { get; } // Add this to ParameterModel and PropertyModel
}
```

This really defines the 'base' of everything besides `ApplicationModel` - I think it's an oversight that we never added properties to parameter/property models.

==============================

Separate filters and ApiExplorer into their own interfaces. They aren't related features and don't belong together.

===========================

The piece that I'm not thrilled about is that there are still several properties that can't really be fit into a coherent model. *I think for the purposes of your PR it's fine, but it's a work item for the team to complete this design.*

Ex:
```
public interface IRoutingModel
{
    AttributeRouteModel AttributeRoute { get; }
    
    IList<ActionConstraintMetadata> ActionConstraints { get; }

    IList<IRouteConstraintProvider> RouteConstraints { get; }
}
```

All of these are related to routing... the problem is that the picture here isn't as neat as I'm making it seem. `ActionModel` also has `HttpMethods` - which overlap with `IActionConstraintMetadata`, and `ControllerModel` lets you define multiple attribute routes.

Grinderofl:
Ahh! This explains so much! I was wondering why parameter and property models were a little lackluster. I'll update this hopefully within today.

Grinderofl:
I've done some changes there as per your recommendations, but refrained from touching the routing for time being, too ill to dive head first in that side of things. I'm not sure if it's feasible though, going by what you're saying... maybe routing part needs to just wait until the team's figured out the design?

rynowak:
> maybe routing part needs to just wait until the team's figured out the design?

Yeah, please don't try to fix that for now. Those changes are going to have a big impact and we need to think about it a little more.

src/Microsoft.AspNet.Mvc.Core/ApplicationModels/IApiExploringModel.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Microsoft.AspNet.Mvc.ApplicationModels
+{
+    public interface IApiExploringModel
```
rynowak: Why the verb in the name? I'd prefer `IApiExplorerModel`. Were you looking to avoid confusion between `ApiExplorerModel` and `IApiExplorerModel`?

src/Microsoft.AspNet.Mvc.Core/ApplicationModels/IBindingModel.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Microsoft.AspNet.Mvc.ModelBinding;
```
rynowak: These all need the standard doc header.

src/Microsoft.AspNet.Mvc.Core/ApplicationModels/PropertyModel.cs
```diff
@@ -25,7 +25,7 @@ public class PropertyModel
             [NotNull] IReadOnlyList<object> attributes)
         {
             PropertyInfo = propertyInfo;
-
+            Properties = new Dictionary<object, object>();
```
rynowak: Take a look at travis. You have some failing tests to update

src/Microsoft.AspNet.Mvc.Core/ApplicationModels/IFilteringModel.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Collections.Generic;
+using Microsoft.AspNet.Mvc.Filters;
+
+namespace Microsoft.AspNet.Mvc.ApplicationModels
```
rynowak: Similar comment to `IApiExploringModel` - I'd prefer `IFilterModel` here

rynowak:
:watch: - looking pretty close. Need to fix failing tests

src/Microsoft.AspNet.Mvc.Core/ApplicationModels/IApiExploringModel.cs
Grinderofl: Yes, I did try to avoid confusion... but ended up having some inconsistencies instead.

Grinderofl:
On it :+1:

Grinderofl:
Looks like the Linux build is failing due to package download issues...

rynowak:
:shipit: I'm going to clean this up an merge it so it can get in for beta8.

FYI your fix for the tests was the wrong fix. It's supposed to be a copy of the dictionary, the tests need to be updated for the addition of the new property.

rynowak:
49acfd562e8cc2b1c0284e7a1c4ffe4227c2d22b merged. Thanks!

PR #3191, opened 2015-09-23
TModel substitution in Razor pages has broken intellisense

### LINKED ISSUE #3185: TModel substitution in Razor pages has broken intellisense
Mvc supports inheriting injected properties (`@inject`) from applicabe _ViewImports. A feature we support as part of this injection is the substitution of any `TModel` tokens that appear in the type name of `@inject` with the current model type. For e.g

```html
// _ViewImport
@inject CssHelper<TModel> CssHelper

// Index
@model IndexModel
```
The `CssHelper` property on Index would be of type `CssHelper<IndexModel>`. However since `TModel` isn't a real C# token, VS tooling for this is broken:

capture

Note squiggly lines under `TModel`. In addition, you do not get any auto completion for this token. Perhaps a fix for this would be to generate a `TModel` identifier in the page at design time. It would also be useful to figure out if there's a nicer way to perform this substitution.

pranavkm:
Fixes #3185

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor.Host/Directives/ChunkHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor.Host/Directives/ChunkHelper.cs
@@ -16,7 +16,8 @@ namespace Microsoft.AspNet.Mvc.Razor.Directives
         /// Token that is replaced by the model name in <c>@inherits</c> and <c>@inject</c>
         /// chunks as part of <see cref="ChunkInheritanceUtility"/>.
         /// </summary>
-        public static readonly string TModelToken = "<TModel>";
+        public static readonly string TModelToken = "TModel";
+        private static readonly string TModelReplaceToken = $"<{TModelToken}>";
 
         /// <summary>
         /// Returns the <see cref="ModelChunk"/> used to determine the model name for the page generated
@@ -86,7 +87,7 @@ namespace Microsoft.AspNet.Mvc.Razor.Directives
                 throw new ArgumentNullException(nameof(modelName));
             }
 
-            return value.Replace(TModelToken, modelName);
+            return value.Replace(TModelReplaceToken, modelName);
         }
     }
 }
--- a/src/Microsoft.AspNet.Mvc.Razor.Host/MvcCSharpCodeGenerator.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor.Host/MvcCSharpCodeGenerator.cs
@@ -2,6 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.IO;
+using Microsoft.AspNet.Mvc.Razor.Directives;
 using Microsoft.AspNet.Razor.CodeGenerators;
 using Microsoft.AspNet.Razor.CodeGenerators.Visitors;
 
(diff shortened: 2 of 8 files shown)


src/Microsoft.AspNet.Mvc.Razor.Host/MvcCSharpCodeGenerator.cs
```diff
@@ -45,6 +47,22 @@ public class MvcCSharpCodeGenerator : CSharpCodeGenerator
             _injectAttribute = injectAttribute;
         }
 
+        protected override CSharpCodeWritingScope BuildClassDeclaration(CSharpCodeWriter writer)
+        {
+            if (Context.Host.DesignTimeMode &&
+                !string.IsNullOrEmpty(Context.SourceFile) &&
```
NTaylorMullen: When would we not have a source file?

src/Microsoft.AspNet.Mvc.Razor.Host/MvcRazorHost.cs
```diff
@@ -45,7 +45,7 @@ public class MvcRazorHost : RazorEngineHost, IMvcRazorHost
             new SetBaseTypeChunk
             {
                 // Microsoft.Aspnet.Mvc.Razor.RazorPage<TModel>
-                TypeName = BaseType + ChunkHelper.TModelToken,
+                TypeName =  $"{BaseType}<{ChunkHelper.TModelToken}>",
```
NTaylorMullen: Got an extra space after `=`

src/Microsoft.AspNet.Mvc.Razor.Host/Directives/ChunkHelper.cs
```diff
@@ -16,7 +16,8 @@ public static class ChunkHelper
         /// Token that is replaced by the model name in <c>@inherits</c> and <c>@inject</c>
         /// chunks as part of <see cref="ChunkInheritanceUtility"/>.
         /// </summary>
-        public static readonly string TModelToken = "<TModel>";
+        public static readonly string TModelToken = "TModel";
+        private static readonly string TModelReplaceToken = $"<{TModelToken}>";
```
NTaylorMullen: Wish static properties didn't have to be ordered like this :cry:

test/Microsoft.AspNet.Mvc.Razor.Host.Test/MvcRazorHostTest.cs
```diff
@@ -221,6 +222,32 @@ public void BasicVisitor_GeneratesCorrectLineMappings()
...
         [Fact]
+        public void MvcRazorHost_GeneratesCorrectLineMappingsAndUsingStatementsForViewImports()
+        {
+            // Arrange
+            var fileProvider = new TestFileProvider();
+            var host = new MvcRazorHostWithNormalizedNewLine(new DefaultChunkTreeCache(fileProvider))
+            {
+                DesignTimeMode = true
+            };
+            host.NamespaceImports.Clear();
```
NTaylorMullen: What's the value in clearing?

src/Microsoft.AspNet.Mvc.Razor.Host/MvcCSharpCodeGenerator.cs
NTaylorMullen: :boom:

NTaylorMullen:
:shipit: once small comments are addressed.

test/Microsoft.AspNet.Mvc.Razor.Host.Test/MvcRazorHostTest.cs
pranavkm: No idea. I just copied it from the test above.

PR #3192, opened 2015-09-24
Include ViewEngine Type in ViewResult Telemetry Source Notifications

avanderhoorn:
Provide a way so that we know which ViewEngine was used to check if it could handle the request.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewResult.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewResult.cs
@@ -79,7 +79,7 @@ namespace Microsoft.AspNet.Mvc
                             actionContext = context,
                             result = this,
                             viewName = viewName,
-                            viewEgineType = viewEngine.GetType(),
+                            viewEngineType = viewEngine.GetType(),
                             searchedLocations = viewEngineResult.SearchedLocations
                         });
                 }
@@ -100,7 +100,7 @@ namespace Microsoft.AspNet.Mvc
                         actionContext = context,
                         result = this,
                         viewName,
-                        viewEgineType = viewEngine.GetType(),
+                        viewEngineType = viewEngine.GetType(),
                         view = view
                     });
             }
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewResult.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewResult.cs
@@ -79,6 +79,7 @@ namespace Microsoft.AspNet.Mvc
                             actionContext = context,
                             result = this,
                             viewName = viewName,
+                            viewEgineType = viewEngine.GetType(),
                             searchedLocations = viewEngineResult.SearchedLocations
                         });
                 }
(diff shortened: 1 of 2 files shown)


avanderhoorn:
// @rynowak

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewResult.cs
```diff
@@ -94,7 +95,14 @@ public class ViewResult : ActionResult
             {
                 telemetry.WriteTelemetry(
                     "Microsoft.AspNet.Mvc.ViewResultViewFound",
-                    new { actionContext = context, result = this, viewName, view = view });
+                    new
+                    {
+                        actionContext = context,
+                        result = this,
+                        viewName,
+                        viewEgineType = viewEngine.GetType(),
```
davidfowl: viewEgineType -> viewEngineType

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewResult.cs
```diff
@@ -79,6 +79,7 @@ public class ViewResult : ActionResult
                             actionContext = context,
                             result = this,
                             viewName = viewName,
+                            viewEgineType = viewEngine.GetType(),
```
rynowak: This is almost certainly not going to do what you want. What is this needed for? Why didn't it come up in our requirements scrub that we did earlier in beta8.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewResult.cs
avanderhoorn: Dam... don't know how I missed that. Updating now.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewResult.cs
avanderhoorn: Only found it today. I don't mind if it goes in post Beta8... I'm not trying to sneak anything past. Was going to talk to you about it this afternoon but you busy, so because it was a "simple" change I thought I would start the conversation with a PR.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewResult.cs
```diff
@@ -79,6 +79,7 @@ public class ViewResult : ActionResult
                             actionContext = context,
                             result = this,
                             viewName = viewName,
+                            viewEngineType = viewEngine.GetType(),
```
pranavkm: It's going to print `CompositeViewEngine` all the time since thats the wrapper. You'd need to log this separately or change `ViewEngineResult` to include the `IViewEngine` instance that found the view.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewResult.cs
avanderhoorn: Realized I didn't answer you original question. We are wanting to let the user know which view engine made the decision... i.e. in v1 we would tell people that it was the WebForm ViewEngine that did a given action and then it went to the Razor ViewEngine. 

viewstab

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewResult.cs
avanderhoorn: Yep, see that... So the next point past this was to see if we can get the `ViewEngineResult` to include `IViewEngine` or if there is a better way all together. Any thoughts?

rynowak:
I recalled some earlier discussion on this, and I'm not really convinced that showing it is all the relevant right now. I'm not aware of any porting efforts of other view engines right now. 

If you really feel like this is useful, I'd suggest looking at the `IView` itself - this will be an instance of `RazorView` - that at least lets you know that razor is involved.

avanderhoorn:
Closing down for the moment.

PR #3197, opened 2015-09-24
Refactor of validation

rynowak:


src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Validation/ValidationStateDictionary.cs
```diff
@@ -0,0 +1,61 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+
+namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
+{
+    public class ValidationStateDictionary : Dictionary<object, ValidationState>
```
rynowak: Implement IDictionary rather than subclassing

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Validation/ValidationStateDictionary.cs
```diff
@@ -0,0 +1,61 @@
...
+            }
+
+            public int GetHashCode(object obj)
+            {
+                return RuntimeHelpers.GetHashCode(obj);
+            }
+        }
+    }
+
+    public class ValidationState
```
rynowak: Rename to ValidationStateEntry

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/ValidationVisitor.cs
```diff
@@ -0,0 +1,341 @@
...
+
+        // Sorts validators based on whether or not they are 'required'. We want to run
+        // 'required' validators first so that we get the best possible error message.
+        private class ValidatorOrderComparer : IComparer<IModelValidator>
+        {
+            public static readonly ValidatorOrderComparer Instance = new ValidatorOrderComparer();
+
+            public int Compare(IModelValidator x, IModelValidator y)
+            {
+                var xScore = x.IsRequired ? 0 : 1;
```
pranavkm: Could we add a `Order` property on validators instead? It'd be in line with the way we do ordering in other parts of the application.

src/Microsoft.AspNet.Mvc.Core/Controllers/DefaultControllerActionArgumentBinder.cs
```diff
@@ -82,20 +82,22 @@ public class DefaultControllerActionArgumentBinder : IControllerActionArgumentBi
...
-                    operationContext.ValidatorProvider,
-                    modelState,
-                    modelExplorer);
-
-                _validator.Validate(validationContext, modelBindingResult.ValidationNode);
+                if (modelBindingResult.Model == null && 
+                    modelState.GetValidationState(modelBindingResult.Key) != ModelValidationState.Valid)
+                {
+                    modelState.MarkFieldValid(modelBindingResult.Key);
+                }
```
rynowak: Take out this `if`.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CancellationTokenModelBinder.cs
```diff
@@ -17,13 +17,7 @@ public Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindingContex
...
             {
                 var model = bindingContext.OperationBindingContext.HttpContext.RequestAborted;
-                var validationNode =
-                    new ModelValidationNode(bindingContext.ModelName, bindingContext.ModelMetadata, model)
-                    {
-                        SuppressValidation = true,
-                    };
-
-                return ModelBindingResult.SuccessAsync(bindingContext.ModelName, model, validationNode);
+                return ModelBindingResult.SuccessAsync(bindingContext.ModelName, model);
```
rynowak: Double check this is covered by an exclude filter

src/Microsoft.AspNet.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
```diff
@@ -79,12 +80,14 @@ public class DictionaryModelBinder<TKey, TValue> : CollectionModelBinder<KeyValu
...
-                if (valueResult.IsModelSet)
-                {
-                    validationNode.ChildNodes.Add(valueResult.ValidationNode);
-                }
+                keyMappings.Add(kvp.Key, convertedKey);
             }
 
+            bindingContext.ValidationState.Add(model, new ValidationState()
+            {
+                Strategy = new WierdExplicitIndexDictionaryValidationStrategy<TKey, TValue>(keyMappings, valueMetadata),
```
rynowak: `ShortFormDictionaryValidationStrategy`

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/ValidationVisitor.cs
rynowak: I'd rather not even do that. The validator providers get a list, and they should be responsible for putting them in the right order.

Not part of this PR though

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CancellationTokenModelBinder.cs
rynowak: This is already covered.

ilmax:
@rynowak are there any chanches to allow async model validation? ;)

rynowak:
@ilmax - not part of this change. Feel free to open an issue but it's something we're very unlikely to do because it increases the overhead significantly in a very hot path.

A more workable solution would to be implement whatever you need in an action filter, or replace the `IControllerActionArgumentBinder` which are both async.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
```diff
@@ -576,30 +558,6 @@ private static void CallPropertyAddRange<TElement>(object target, object source)
             }
         }
 
-        // Returns true if validator execution adds a model error.
-        private static bool RunValidator(
```
rynowak: This was unused.

ilmax:
Thak you @rynowak for your quick response, I've already opened an issue in the past around a sync model validation. Performance is a fair argument to exclude the feature, but sometimes is useful (or even required) to hit some external resources to perform model validation. 
What I would like to have (as a opt-in feature maybe via a different package) is the ability to hook into the validation phase and write async logic without rewrite the whole validation components, is that possible in some way?

rynowak:
> What I would like to have (as a opt-in feature maybe via a different package) is the ability to hook into the validation phase and write async logic without rewrite the whole validation components, is that possible in some way?

You can do this today by replacing `IControllerActionArgumentBinder` with your own implementation. You'd need to replace this code with your own validation system: https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/Controllers/DefaultControllerActionArgumentBinder.cs#L98

We'd be open to changes to make this kind of extensibility easier, but it's unlikely we'd put it in the box.

test/Microsoft.AspNet.Mvc.IntegrationTests/ActionParametersIntegrationTest.cs
```diff
@@ -107,13 +107,13 @@ private class Person6
             // Read-only collection should not be updated.
             Assert.Empty(boundModel.Address);
 
-            // ModelState (data is valid but is not copied into Address).
-            Assert.True(modelState.IsValid);
+            // ModelState (data is can't be validated).
+            Assert.False(modelState.IsValid);
```
rynowak: @dougbu Let's talk about this case when you get to it.

rynowak:
@dougbu  - updated with docs and such. My only outstanding TODO that I know of here is to write unit tests for each of the validation strategies.

test/Microsoft.AspNet.Mvc.IntegrationTests/ActionParametersIntegrationTest.cs
rynowak: Discussed in person, agreed that `Unvalidated` is fine.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Validation/ValidationEntry.cs
```diff
@@ -0,0 +1,51 @@
...
+
+            Metadata = metadata;
+            Key = key;
+            Model = model;
+        }
+
+        /// <summary>
+        /// The model prefix associated with <see cref="Model"/>.
+        /// </summary>
+        public string Key { get; private set; }
```
NTaylorMullen: `{ get; }` For this and other properties.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Validation/ValidationEntry.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// Contains data needed for validating a child entry of a model object. See <see cref="IValidationStrategy"/>.
+    /// </summary>
+    public struct ValidationEntry
+    {
+        /// <summary>
+        /// Creates a new <see cref="ValidationEntry"/>.
+        /// </summary>
+        /// <param name="metadata">The <see cref="ModelMetadata"/> associated with <paramref name="model"/>.</param>
+        /// <param name="key">The model prefix associated with <paramref name="model"/>.</param>
+        /// <param name="model">the model object.</param>
```
NTaylorMullen: `The model object.`

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Validation/ValidationEntry.cs
```diff
@@ -0,0 +1,51 @@
...
+        public ValidationEntry( ModelMetadata metadata, string key, object model)
```
NTaylorMullen: `(ModelMetadata ...`

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Validation/ValidationStateDictionary.cs
```diff
@@ -0,0 +1,176 @@
...
+    /// <summary>
+    /// Used for tracking validation state to customize validation behavior for a model object.
+    /// </summary>
+    public class ValidationStateDictionary : 
+        IDictionary<object, ValidationStateEntry>, 
+        IReadOnlyDictionary<object, ValidationStateEntry>
+    {
+        private readonly Dictionary<object, ValidationStateEntry> _inner;
+
+        public ValidationStateDictionary()
```
NTaylorMullen: docs

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CollectionModelBinder.cs
```diff
@@ -252,17 +239,26 @@ private Task<CollectionResult> BindComplexCollection(ModelBindingContext binding
 
             return new CollectionResult
             {
-                ValidationNode = validationNode,
-                Model = boundCollection
+                Model = boundCollection,
+
+                // If we're working with a fixed set of indexes then this is the format like:
+                //
+                //  ?parameter.index=zero,one,two&parameter[zero]=0&&parameter[one]=1&parameter[two]=1...
```
NTaylorMullen: `[two]=2...`

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultCollectionValidationStrategy.cs
```diff
@@ -0,0 +1,123 @@
...
+    /// 
+    ///     In this case the elements of the collection are identified in the input data set by an incrementing
+    ///     integer index.
+    /// </example>
+    /// 
+    /// or:
+    /// 
+    /// <example>
+    ///     Model: IDictionary&lt;string, int&gt; 
+    ///     Query String: ?students[0].Name=Joey&amp;students[0].Age=8
```
NTaylorMullen: Interesting, don't have to do `[0].Value.Name`?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultCollectionValidationStrategy.cs
```diff
@@ -0,0 +1,123 @@
...
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
+{
+    /// <summary>
+    /// The default implementation of <see cref="IValidationStrategy"/> for a collection.
+    /// </summary>
+    /// <remarks>
+    /// This implemenation handles cases like:
```
rynowak: implementation

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultCollectionValidationStrategy.cs
rynowak: Yep, that's right. Overaggressive copy pasting of the examples.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/ExplicitIndexCollectionValidationStrategy.cs
```diff
@@ -0,0 +1,116 @@
...
+    ///     class will yield two 'Student' objects with corresponding keys 'students[Joey]' and 'students[Katherine]'.
+    /// </example>
+    /// 
+    /// Using this key format, the enumerator enumerates model objects of type matching
+    /// <see cref="ModelMetadata.ElementMetadata"/>. The keys captured during model binding are mapped to the elements
+    /// in the collection to compute the model prefix keys.
+    /// </remarks>
+    public class ExplicitIndexCollectionValidationStrategy : IValidationStrategy
+    {
+        public ExplicitIndexCollectionValidationStrategy(IEnumerable<string> elementKeys)
```
NTaylorMullen: docs

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/ExplicitIndexCollectionValidationStrategy.cs
```diff
@@ -0,0 +1,116 @@
...
+        {
+            ElementKeys = elementKeys;
```
NTaylorMullen: This able to be `null`?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/ValidationVisitor.cs
```diff
@@ -0,0 +1,356 @@
...
+            public static StateManager Recurse(
+                ValidationVisitor visitor,
+                string key,
+                ModelMetadata metadata,
+                object model,
+                IValidationStrategy strategy)
+            {
+                var recursifier = new StateManager(visitor, model);
+
+                visitor._container = visitor._model;
```
NTaylorMullen: container is the model?

NTaylorMullen:
Only found small issues. You gots a :shipit: from me :smile:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/ExplicitIndexCollectionValidationStrategy.cs
rynowak: Nope, that wouldn't work at all!

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/ValidationVisitor.cs
rynowak: Container is what we're inside. If we're validating a property, then model is the object we evaluated the property from.

NTaylorMullen: :+1:

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
```diff
@@ -174,5 +178,11 @@ public ModelBindingContext()
         /// is eligible for model binding.
         /// </summary>
         public Func<ModelBindingContext, string, bool> PropertyFilter { get; set; }
+
+        /// <summary>
+        /// Gets or sets the <see cref="ValidationStateDictionary"/>. Used for tracking validation state to
+        /// customize validation behavior for a model object.
+        /// </summary>
+        public ValidationStateDictionary ValidationState { get; set; }
```
dougbu: Why does this property have a setter? Is this just for the odd special-case almost-a-copy-constructor code in `CompositeModelBinder`?

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Validation/ValidationStateDictionary.cs
```diff
@@ -0,0 +1,179 @@
...
+            get
+            {
+                ValidationStateEntry entry;
+                TryGetValue(key, out entry);
+                return entry;
+            }
+
+            set
+            {
+                _inner[key] = value;
```
dougbu: In what scenario is a `null` `ValidationStateEntry` interesting? Put another way, why no `null` check here?

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Validation/ValidationStateDictionary.cs
```diff
@@ -0,0 +1,179 @@
...
+                return ((IDictionary<object, ValidationStateEntry>)_inner).IsReadOnly;
+            }
+        }
+
+        /// <inheritdoc />
+        public ICollection<object> Keys
+        {
+            get
+            {
+                return ((IDictionary<object, ValidationStateEntry>)_inner).Keys;
```
dougbu: Curious: Why not `_inner.Keys`?

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Validation/ValidationStateDictionary.cs
```diff
@@ -0,0 +1,179 @@
...
+            return _inner.TryGetValue(key, out value);
+        }
+
+        /// <inheritdoc />
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            return ((IDictionary<object, ValidationStateEntry>)_inner).GetEnumerator();
+        }
+        
+        private class ReferenceEqualityComparer : IEqualityComparer<object>
```
dougbu: Two copies of this code now. Perhaps the version in Mvc.Core should be public and in Mvc.Abstractions?

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Validation/ValidationStateDictionary.cs
```diff
@@ -0,0 +1,179 @@
...
+        {
+            public static readonly ReferenceEqualityComparer Instance = new ReferenceEqualityComparer();
+
+            public new bool Equals(object x, object y)
```
dougbu: `private`

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Validation/ValidationStateDictionary.cs
```diff
@@ -0,0 +1,179 @@
...
+            {
+                return Object.ReferenceEquals(x, y);
```
dougbu: lowercase `object`

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Validation/ValidationStateEntry.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// <summary>
+    /// An entry in a <see cref="ValidationStateDictionary"/>. Records state information to override the default
+    /// behavior of validation for an object.
+    /// </summary>
+    public class ValidationStateEntry
+    {
+        /// <summary>
+        /// Gets or sets the model prefix associated with the entry.
+        /// </summary>
+        public string Key { get; set; }
```
dougbu: See the need to modify an existing entry in `CompositeModelBinder`. But, pity this `class` can't be immutable, like `ValidationEntry`.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultCollectionValidationStrategy.cs
```diff
@@ -0,0 +1,123 @@
...
+namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
+{
+    /// <summary>
+    /// The default implementation of <see cref="IValidationStrategy"/> for a collection.
+    /// </summary>
+    /// <remarks>
+    /// This implementation handles cases like:
+    /// <example>
+    ///     Model: IList&lt;Student&gt; 
+    ///     Query String: ?students[0].Age=8&students[1].Age=9
```
dougbu: Invalid `&` here. You caught some of these just below and later in the PR but should be `&amp;` in all new examples.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultObjectValidator.cs
```diff
@@ -27,324 +21,49 @@ public class DefaultObjectValidator : IObjectModelValidator
...
-        {
-            public static readonly ValidatorOrderComparer Instance = new ValidatorOrderComparer();
-
-            public int Compare(IModelValidator x, IModelValidator y)
-            {
-                var xScore = x.IsRequired ? 0 : 1;
-                var yScore = y.IsRequired ? 0 : 1;
-                return xScore.CompareTo(yScore);
-            }
+            var metadata = model == null ? null : _modelMetadataProvider.GetMetadataForType(model.GetType());
```
dougbu: Shouldn't this get the `ModelMetadata` from the `ValidationStateDictionary`, when that's available?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/ExplicitIndexCollectionValidationStrategy.cs
```diff
@@ -0,0 +1,128 @@
...
+{
+    /// <summary>
+    /// An implementation of <see cref="IValidationStrategy"/> for a collection bound using 'explict indexing'
+    /// style keys.
+    /// </summary>
+    /// <remarks>
+    /// This implemenation handles cases like:
+    /// <example>
+    ///     Model: IList&lt;Student&gt; 
+    ///     Query String: ?students.index=Joey,Katherine&amp;students[Joey].Age=8&students[Katherine].Age=9
```
dougbu: Another invalid `&`

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/ExplicitIndexCollectionValidationStrategy.cs
```diff
@@ -0,0 +1,128 @@
...
+                throw new ArgumentNullException(nameof(elementKeys));
+            }
+
+            ElementKeys = elementKeys;
+        }
+
+        /// <summary>
+        /// Gets the keys of collection elements that were used during model binding.
+        /// </summary>
+        public IEnumerable<string> ElementKeys { get; }
```
dougbu: `internal` with "for testing" comment?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/ShortFormDictionaryValidationStrategy.cs
```diff
@@ -0,0 +1,135 @@
...
+            ModelMetadata valueMetadata)
+        {
+            KeyMappings = keyMappings;
+            _valueMetadata = valueMetadata;
+        }
+
+        /// <summary>
+        /// Gets the mapping from model prefix key to dictionary key.
+        /// </summary>
+        public IEnumerable<KeyValuePair<string, TKey>> KeyMappings { get; }
```
dougbu: `internal` with "for testing" comment?

PR #3200, opened 2015-09-25
Add localizers to validation attributes

kirthik:
cc @rynowak, @NTaylorMullen


src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsClientModelValidatorProvider.cs
```diff
@@ -19,16 +22,26 @@ public class DataAnnotationsClientModelValidatorProvider : IClientModelValidator
     {
         // A factory for validators based on ValidationAttribute.
         internal delegate IClientModelValidator
-            DataAnnotationsClientModelValidationFactory(ValidationAttribute attribute);
+            DataAnnotationsClientModelValidationFactory(ValidationAttribute attribute, IStringLocalizer stringLocalizer);
```
rynowak: this would be much easier to read if you put the line break after `DataAnnotationsClientModelValidationFactory(`

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsClientModelValidatorProvider.cs
```diff
@@ -36,6 +49,15 @@ public void GetValidators(ClientValidatorProviderContext context)
             {
                 throw new ArgumentNullException(nameof(context));
             }
+            IStringLocalizer stringLocalizer = null;
+            if (_options != null &&
```
rynowak: When would options be null?

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsClientModelValidatorProvider.cs
```diff
@@ -19,16 +22,26 @@ public class DataAnnotationsClientModelValidatorProvider : IClientModelValidator
...
         {
             get { return _attributeFactories; }
         }
 
+        public DataAnnotationsClientModelValidatorProvider(
+            IOptions<MvcDataAnnotationsLocalizationOptions> options,
+            IStringLocalizerFactory stringLocalizerFactory)
+        {
+            _options = options;
+            _stringLocalizerFactory = stringLocalizerFactory;
```
rynowak: What's a string localizer factory? Is it a service?

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsClientModelValidatorProvider.cs
```diff
@@ -36,6 +49,15 @@ public void GetValidators(ClientValidatorProviderContext context)
             {
                 throw new ArgumentNullException(nameof(context));
             }
+            IStringLocalizer stringLocalizer = null;
+            if (_options != null &&
+                _options.Value.DataAnnotationLocalizerProvider != null &&
+                _stringLocalizerFactory != null)
+            {
+                stringLocalizer = _options.Value.DataAnnotationLocalizerProvider(
+                    context.ModelMetadata.ContainerType ?? context.ModelMetadata.ModelType,
```
rynowak: Might want to document what this fallback represents

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsModelValidator.cs
```diff
@@ -5,12 +5,17 @@
...
 using System.ComponentModel.DataAnnotations;
 using System.Linq;
+using Microsoft.Framework.Localization;
 
 namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
 {
     public class DataAnnotationsModelValidator : IModelValidator
     {
-        public DataAnnotationsModelValidator(ValidationAttribute attribute)
+        public DataAnnotationsModelValidator(ValidationAttribute attribute): this(attribute, null)
```
rynowak: fix spacing

rynowak: Why is this constructor still needed?

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsClientModelValidatorProvider.cs
```diff
@@ -46,14 +68,14 @@ public void GetValidators(ClientValidatorProviderContext context)
                 DataAnnotationsClientModelValidationFactory factory;
                 if (_attributeFactories.TryGetValue(attribute.GetType(), out factory))
                 {
-                    context.Validators.Add(factory(attribute));
+                    context.Validators.Add(factory(attribute, stringLocalizer));
```
rynowak: Does this need the same logic as the server-side validators?

src/Microsoft.AspNet.Mvc.DataAnnotations/DependencyInjection/MvcDataAnnotationsMvcBuilderExtensions.cs
```diff
@@ -0,0 +1,50 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNet.Mvc.DataAnnotations.Internal;
+using Microsoft.AspNet.Mvc.ModelBinding.Validation;
+
+namespace Microsoft.Framework.DependencyInjection
+{
+    public static class MvcDataAnnotationsMvcBuilderExtensions
```
rynowak: docs

src/Microsoft.AspNet.Mvc.DataAnnotations/DependencyInjection/MvcDataAnnotationsMvcBuilderExtensions.cs
```diff
@@ -0,0 +1,50 @@
...
+        public static IMvcBuilder AddDataAnnotationsLocalization(
+            this IMvcBuilder builder,
+            Action<MvcDataAnnotationsLocalizationOptions> setupAction)
+        {
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            DataAnnotationsLocalizationService.AddDataAnnotationsLocalizationServices(
```
rynowak: `DataAnnotationsLocalizationServices`

src/Microsoft.AspNet.Mvc.DataAnnotations/Internal/MvcDataAnnotationsMvcOptionsSetup.cs
```diff
@@ -12,15 +15,23 @@ namespace Microsoft.AspNet.Mvc.DataAnnotations.Internal
...
         {
         }
 
-        public static void ConfigureMvc(MvcOptions options)
+        public static void ConfigureMvc(
+            MvcOptions options,
+            IServiceProvider serviceProvider)
         {
+            var dataAnnotationLocalizationOptions = serviceProvider.GetService<IOptions<MvcDataAnnotationsLocalizationOptions>>();
+            var stringLocalizerFactory = serviceProvider.GetService<IStringLocalizerFactory>();
```
rynowak: DI these directly rather than the DI'ing the service provider.

src/Microsoft.AspNet.Mvc.DataAnnotations/Internal/MvcDataAnnotationsMvcOptionsSetup.cs
```diff
@@ -12,15 +15,23 @@ namespace Microsoft.AspNet.Mvc.DataAnnotations.Internal
...
+            : base(options => ConfigureMvc(options, serviceProvider))
         {
         }
 
-        public static void ConfigureMvc(MvcOptions options)
+        public static void ConfigureMvc(
+            MvcOptions options,
+            IServiceProvider serviceProvider)
         {
+            var dataAnnotationLocalizationOptions = serviceProvider.GetService<IOptions<MvcDataAnnotationsLocalizationOptions>>();
```
rynowak: This can be a `GetRequiredService` it will always be there.

src/Microsoft.AspNet.Mvc.DataAnnotations/Internal/MvcDataAnnotationsMvcOptionsSetup.cs
rynowak: Document that this might not be registered

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/KeyValuePairModelBinderTest.cs
```diff
@@ -242,7 +242,9 @@ private static ModelBindingContext CreateContext()
                 {
                     ModelBinder = innerBinder ?? CreateIntBinder(),
                     MetadataProvider = metataProvider,
-                    ValidatorProvider = new DataAnnotationsModelValidatorProvider()
+                    ValidatorProvider = new DataAnnotationsModelValidatorProvider(
+                        options: null,
```
rynowak: new `TestOptionsManager<T>()`

test/Microsoft.AspNet.Mvc.DataAnnotations.Test/MvcDataAnnotationsMvcBuilderExtensionsTest.cs
```diff
@@ -0,0 +1,89 @@
...
+                },
+                service =>
+                {
+                    Assert.Equal(typeof(IConfigureOptions<MvcDataAnnotationsLocalizationOptions>), service.ServiceType);
+                    Assert.Equal(ServiceLifetime.Transient, service.Lifetime);
+                });
+        }
+
+        [Fact]
+        public void AddDataAnnotationsLocalizationServicesWithOptios_AddsNeededServices()
```
rynowak: WithOptions

test/Microsoft.AspNet.Mvc.DataAnnotations.Test/MvcDataAnnotationsMvcBuilderExtensionsTest.cs
```diff
@@ -0,0 +1,89 @@
...
+
+using Microsoft.AspNet.Mvc.DataAnnotations.Internal;
+using Microsoft.AspNet.Mvc.Localization;
+using Microsoft.AspNet.Mvc.ModelBinding.Validation;
+using Microsoft.Framework.DependencyInjection;
+using Microsoft.Framework.Localization;
+using Microsoft.Framework.OptionsModel;
+using Xunit;
+
+namespace Microsoft.AspNet.Mvc.DataAnnotations.Test
```
rynowak: Drop the `.Test`

test/Microsoft.AspNet.Mvc.DataAnnotations.Test/MvcDataAnnotationsMvcBuilderExtensionsTest.cs
```diff
@@ -0,0 +1,89 @@
...
+{
+    public class DataAnnotationsLocalizationServiceTest
+    {
+        [Fact]
+        public void AddDataAnnotationsLocalizationServices_AddsNeededServices()
```
rynowak: This seems like the kind of test that would change if you made an insignificant change to the product (lifetime, localization adds another service). Do you think there's a lot of value coming from these?

I wouldn't bother as this seems really fragile.

test/Microsoft.AspNet.Mvc.FunctionalTests/LocalizationTest.cs
```diff
@@ -121,5 +122,36 @@ Learn More
...
+<span class=""field-validation-error"" data-valmsg-for=""Product.ProductName"" data-valmsg-replace=""true"">Nom du produit est invalide</span>
+<div class=""editor-label""><label for=""Name"">Name</label></div>
+<div class=""editor-field""><input class=""input-validation-error text-box single-line"" data-val=""true"" data-val-minlength=""Nom non valide. Longueur minimale de nom est 4"" data-val-minlength-min=""4"" id=""Name"" name=""Name"" type=""text"" value=""A"" /> <span class=""field-validation-error"" data-valmsg-for=""Name"" data-valmsg-replace=""true"">Nom non valide. Longueur minimale de nom est 4</span></div>
+
+<div class=""editor-label""><label for=""Product_ProductName"">ProductName</label></div>
+<div class=""editor-field""><input class=""input-validation-error text-box single-line"" data-val=""true"" data-val-required=""Nom du produit est invalide"" id=""Product_ProductName"" name=""Product.ProductName"" type=""text"" value="""" /> <span class=""field-validation-error"" data-valmsg-for=""Product.ProductName"" data-valmsg-replace=""true"">Nom du produit est invalide</span></div>";
```
rynowak: :dancer:

test/Microsoft.AspNet.Mvc.FunctionalTests/LocalizationTest.cs
```diff
@@ -121,5 +122,36 @@ Learn More
...
+                new CookieHeaderValue("ASPNET_CULTURE", cultureCookie).ToString());
+
+            // Act
+            var response = await Client.SendAsync(request);
+            var body = await response.Content.ReadAsStringAsync();
+
+            // Assert
+            // Mono issue - https://github.com/aspnet/External/issues/19
+            Assert.Equal(
+                PlatformNormalizer.NormalizeContent(expected),
```
rynowak: I don't think this will handle your localized strings. What does travis say?

test/WebSites/LocalizationWebSite/Resources/User.fr.resx
```diff
@@ -0,0 +1,123 @@
...
+    <value>2.0</value>
+  </resheader>
+  <resheader name="reader">
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="Name" xml:space="preserve">
+    <value>Nom non valide. Longueur minimale de nom est 4</value>
```
rynowak: Whoa! a `4`!

Does the resource lookup get applied before formatting? How is this going to work?

test/WebSites/LocalizationWebSite/Startup.cs
```diff
@@ -10,7 +10,14 @@ public class Startup
     {
         public void ConfigureServices(IServiceCollection services)
         {
-            services.AddMvc().AddViewLocalization(options => options.ResourcesPath = "Resources");
+            services
+                .AddMvc()
+                .AddViewOptions(options =>
+                {
+                    options.HtmlHelperOptions.ClientValidationEnabled = true;
```
rynowak: I believe this is the default

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsClientModelValidatorProvider.cs
kirthik: My bad. They wont be. Will remove this condition.

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsClientModelValidatorProvider.cs
kirthik: Yes, its a service.

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsClientModelValidatorProvider.cs
kirthik: Din't feel it adds a lot of value here. The adapters already have `IStringLocalizer` as constructor argument. So passing stringLocalizer here. If resource is not configured, `stringLocalizer` would be null. Let me know if you disagree

rynowak: Seems fine. So then the logic to localize or not is inside that adapter?

kirthik: Its in `DataAnnotationsClientModelValidator<TAttribute>` class `GetErrorMessage` method. Call `stringLocalizer` if it is not null else just format the error message

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsClientModelValidatorOfTAttribute.cs
```diff
@@ -48,6 +59,11 @@ protected virtual string GetErrorMessage(ModelMetadata modelMetadata)
                 throw new ArgumentNullException(nameof(modelMetadata));
             }
 
+            if (StringLocalizer != null)
```
NTaylorMullen:
```C#
var displayName = modelMetadata.GetDisplayName();

if (StringLocalizer != null)
{
    return StringLocalizer[displayName]
}

return Attribute.FormatErrorMessage(displayName);
```

test/Microsoft.AspNet.Mvc.FunctionalTests/LocalizationTest.cs
kirthik: Works fine on mono (Linux). Mac failed because of a random error. Re-running.

rynowak: Then it's likely not needed. This thing does a find and replace for certain strings which for which mono doesn't match.

kirthik:
@NTaylorMullen, @rynowak : updated

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsClientModelValidatorProvider.cs
```diff
@@ -19,16 +22,28 @@ public class DataAnnotationsClientModelValidatorProvider : IClientModelValidator
     {
         // A factory for validators based on ValidationAttribute.
         internal delegate IClientModelValidator
-            DataAnnotationsClientModelValidationFactory(ValidationAttribute attribute);
+            DataAnnotationsClientModelValidationFactory(
+            ValidationAttribute attribute,
+            IStringLocalizer stringLocalizer);
```
rynowak: > this would be much easier to read if you put the line break after `DataAnnotationsClientModelValidationFactory(`

And now `DataAnnotationsClientModelValidationFactory(` can move up to the previous line and it looks like normal code!

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsClientModelValidatorOfTAttribute.cs
```diff
@@ -31,6 +34,14 @@ public TAttribute Attribute
             get;
         }
 
+        /// <summary>
+        /// Gets the <see cref="IStringLocalizer"/>.
+        /// </summary>
+        public IStringLocalizer StringLocalizer
```
NTaylorMullen: Why expose this? I don't see it accessed anywhere.

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsModelValidator.cs
```diff
@@ -18,10 +23,13 @@ public DataAnnotationsModelValidator(ValidationAttribute attribute)
             }
 
             Attribute = attribute;
+            StringLocalizer = stringLocalizer;
         }
 
         public ValidationAttribute Attribute { get; private set; }
 
+        public IStringLocalizer StringLocalizer { get; private set; }
```
rynowak: This guy and the one above it can be get-only

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsClientModelValidatorProvider.cs
```diff
@@ -19,16 +22,28 @@ public class DataAnnotationsClientModelValidatorProvider : IClientModelValidator
...
             BuildAttributeFactoriesDictionary();
+        private IOptions<MvcDataAnnotationsLocalizationOptions> _options;
+        private IStringLocalizerFactory _stringLocalizerFactory;
 
         internal Dictionary<Type, DataAnnotationsClientModelValidationFactory> AttributeFactories
         {
             get { return _attributeFactories; }
         }
 
+        public DataAnnotationsClientModelValidatorProvider(
```
NTaylorMullen: Move this ctor above the internal property above

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsClientModelValidatorProvider.cs
NTaylorMullen: It's possible for `stringLocalizer` here to be `null` correct? Is that ok?

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsModelValidator.cs
NTaylorMullen: `{ get; }` also same question [here](https://github.com/aspnet/Mvc/pull/3200/files#r40462779)

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsClientModelValidatorProvider.cs
```diff
@@ -5,6 +5,9 @@
 using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
 using System.Linq;
+using Microsoft.Framework.DependencyInjection;
```
NTaylorMullen: Don't need this do we?

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsModelValidatorProvider.cs
```diff
@@ -2,10 +2,12 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
 using System.Linq;
 using System.Reflection;
+using Microsoft.Framework.DependencyInjection;
```
NTaylorMullen: Don't need this do we?

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsModelValidatorProvider.cs
```diff
@@ -16,11 +18,42 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
     /// </summary>
     public class DataAnnotationsModelValidatorProvider : IModelValidatorProvider
     {
+        private IOptions<MvcDataAnnotationsLocalizationOptions> _options;
```
NTaylorMullen: `private readonly` here and in next line.

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsModelValidatorProvider.cs
```diff
@@ -16,11 +18,42 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
...
+
         public void GetValidators(ModelValidatorProviderContext context)
         {
+            IStringLocalizer stringLocalizer = null;
+            if (_options != null &&
+                _options.Value.DataAnnotationLocalizerProvider != null &&
+                _stringLocalizerFactory != null)
+            {
+                stringLocalizer = _options.Value.DataAnnotationLocalizerProvider(
+                    context.ModelMetadata.ContainerType ?? context.ModelMetadata.ModelType,
```
NTaylorMullen: Curious: Why do we prefer the `ContainerType` here as opposed to the model type?

src/Microsoft.AspNet.Mvc.DataAnnotations/MvcDataAnnotationsLocalizationOptions.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    /// <summary>
+    /// Provides programmatic configuration for DataAnnotations localization in the MVC framework.
+    /// </summary>
+    public class MvcDataAnnotationsLocalizationOptions
+    {
+        /// <summary>
+        /// The delegate to invoke for creating <see cref="IStringLocalizer"/>.
+        /// </summary>
+        public Func<Type, IStringLocalizerFactory, IStringLocalizer> DataAnnotationLocalizerProvider;
```
NTaylorMullen: Do we typically have `Func`s in options for creating types/utilities? Seems more like a service provider factory concern.

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsModelValidatorProvider.cs
```diff
@@ -16,11 +18,42 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
...
+                    _stringLocalizerFactory);
+            }
+
             foreach (var attribute in context.ValidatorMetadata.OfType<ValidationAttribute>())
             {
-                context.Validators.Add(new DataAnnotationsModelValidator(attribute));
+                if (stringLocalizer != null &&
```
NTaylorMullen: This logic seems odd here. If there's a `stringLocalizer` we should always pass it, the validator itself should determine which validation logic it should use (the validation attributes or the string localizer's).

src/Microsoft.AspNet.Mvc.DataAnnotations/DataTypeAttributeAdapter.cs
```diff
@@ -16,8 +17,9 @@ public class DataTypeAttributeAdapter : DataAnnotationsClientModelValidator<Data
     {
         public DataTypeAttributeAdapter(
```
NTaylorMullen: nit

```C#
public DataTypeAttributeAdapter(DataTypeAttribute attribute, string ruleName, IStringLocalizer stringLocalizer)
    : base(attribute, stringLocalizer)
```

src/Microsoft.AspNet.Mvc.DataAnnotations/Internal/MvcDataAnnotationsLocalizationOptionsSetup.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// Sets up default options for <see cref="MvcDataAnnotationsLocalizationOptions"/>.
+    /// </summary>
+    public class MvcDataAnnotationsLocalizationOptionsSetup : ConfigureOptions<MvcDataAnnotationsLocalizationOptions>
+    {
+        public MvcDataAnnotationsLocalizationOptionsSetup()
+            : base(ConfigureMvc)
+        {
+        }
+
+        public static void ConfigureMvc(MvcDataAnnotationsLocalizationOptions options)
```
NTaylorMullen: docs

src/Microsoft.AspNet.Mvc.DataAnnotations/DependencyInjection/MvcDataAnnotationsMvcCoreBuilderExtensions.cs
```diff
@@ -22,11 +23,32 @@ public static IMvcCoreBuilder AddDataAnnotations(this IMvcCoreBuilder builder)
             return builder;
         }
 
+        public static IMvcCoreBuilder AddDataAnnotationsLocalization(
```
NTaylorMullen: Oddd, none of our extensions have docs?

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsModelValidatorProvider.cs
```diff
@@ -16,11 +18,42 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
     /// </summary>
     public class DataAnnotationsModelValidatorProvider : IModelValidatorProvider
     {
+        private IOptions<MvcDataAnnotationsLocalizationOptions> _options;
+        private IStringLocalizerFactory _stringLocalizerFactory;
+
+        public DataAnnotationsModelValidatorProvider(
```
NTaylorMullen: docs

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsClientModelValidatorProvider.cs
NTaylorMullen: docs

src/Microsoft.AspNet.Mvc.DataAnnotations/Internal/MvcDataAnnotationsMvcOptionsSetup.cs
```diff
@@ -12,15 +15,26 @@ namespace Microsoft.AspNet.Mvc.DataAnnotations.Internal
     /// </summary>
     public class MvcDataAnnotationsMvcOptionsSetup : ConfigureOptions<MvcOptions>
     {
-        public MvcDataAnnotationsMvcOptionsSetup()
-            : base(ConfigureMvc)
+        public MvcDataAnnotationsMvcOptionsSetup(
+           IServiceProvider serviceProvider)
```
NTaylorMullen: nit: move up to previous line.

src/Microsoft.AspNet.Mvc.DataAnnotations/Internal/MvcDataAnnotationsMvcOptionsSetup.cs
```diff
@@ -12,15 +15,26 @@ namespace Microsoft.AspNet.Mvc.DataAnnotations.Internal
...
 
-        public static void ConfigureMvc(MvcOptions options)
+        public static void ConfigureMvc(
+            MvcOptions options,
+            IServiceProvider serviceProvider)
         {
+            var dataAnnotationLocalizationOptions =
+                serviceProvider.GetRequiredService<IOptions<MvcDataAnnotationsLocalizationOptions>>();
+
+            //This service will be registered only if AddDataAnnotationsLocalization() is added to service collection.
```
NTaylorMullen: `// This`

src/Microsoft.AspNet.Mvc.DataAnnotations/Internal/MvcDataAnnotationsMvcOptionsSetup.cs
```diff
@@ -12,15 +15,26 @@ namespace Microsoft.AspNet.Mvc.DataAnnotations.Internal
...
         {
         }
 
-        public static void ConfigureMvc(MvcOptions options)
+        public static void ConfigureMvc(
+            MvcOptions options,
+            IServiceProvider serviceProvider)
         {
+            var dataAnnotationLocalizationOptions =
+                serviceProvider.GetRequiredService<IOptions<MvcDataAnnotationsLocalizationOptions>>();
```
NTaylorMullen: It's odd that you bring in the `IServiceProvider` in the ctor, just to use it to further pull in additional services that may or may not exist. Is there a reason why we can't just specify in the ctor the dependencies this class needs then pass them through?

src/Microsoft.AspNet.Mvc.ViewFeatures/Internal/MvcViewOptionsSetup.cs
```diff
@@ -14,16 +17,24 @@ public class MvcViewOptionsSetup : ConfigureOptions<MvcViewOptions>
         /// <summary>
         /// Initializes a new instance of <see cref="MvcViewOptionsSetup"/>.
         /// </summary>
-        public MvcViewOptionsSetup()
-            : base(ConfigureMvc)
+        public MvcViewOptionsSetup(IServiceProvider serviceProvider)
```
NTaylorMullen: [Question](https://github.com/aspnet/Mvc/pull/3200/files#r40466825) also applies here since it's pas through.

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsClientModelValidatorOfTAttribute.cs
kirthik: Yes, I will make this private.

test/Microsoft.AspNet.Mvc.DataAnnotations.Test/DataAnnotationsModelValidatorTest.cs
```diff
@@ -217,6 +218,36 @@ public void ValidateReturnsMemberNameIfItIsDifferentFromDisplayName()
...
+
+            var attribute = new Mock<ValidationAttribute> { CallBase = true };
+            attribute.Setup(a => a.IsValid(modelExplorer.Model)).Returns(false);
+
+            attribute.Object.ErrorMessage = "Length";
+
+            var localizedString = new LocalizedString("Length", "Longueur est invalide");
+            var stringLocalizer = new Mock<IStringLocalizer>();
+            stringLocalizer.Setup(s => s["Length"]).Returns(localizedString);
+
```
NTaylorMullen: :boom:

test/WebSites/LocalizationWebSite/Controllers/HomeController.cs
```diff
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Collections.Generic;
```
NTaylorMullen: Need this?

test/WebSites/LocalizationWebSite/Models/User.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        [MinLength(4, ErrorMessage = "Name")]
+        public string Name { get; set; }
+
+        public Product Product { get; set; }
+    }
+
+    public class Product
+    {
+        [Required(ErrorMessage ="ProductName")]
```
NTaylorMullen: ` = `

test/WebSites/LocalizationWebSite/Models/User.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.ComponentModel.DataAnnotations;
+
+namespace LocalizationWebSite.Models
+{
+    public class User
```
NTaylorMullen: Add a variation where there's another `User` class.

src/Microsoft.AspNet.Mvc.DataAnnotations/Internal/MvcDataAnnotationsLocalizationOptionsSetup.cs
NTaylorMullen: options can't be `null`

NTaylorMullen:
:watch: Also, are the travis failures due to missing changes from Localization?

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsModelValidatorProvider.cs
kirthik: Primitive data types like `string`, `int` are also treated as `ModelType`. We don't want to create `.resx` for each data types/property. `ContainerType` is the Complex type. So we want to look at `ModelType` only if `ContainerType` is `null`

src/Microsoft.AspNet.Mvc.DataAnnotations/Internal/MvcDataAnnotationsMvcOptionsSetup.cs
kirthik: This can get called even outside of locaization in which case, service provider will not have `IOptions<MvcDataAnnotationsLocalizationOptions>` and `IStringLocalizerFactory` services. If I add them to constructor, I get `Unable to resolve service errors`.

src/Microsoft.AspNet.Mvc.DataAnnotations/Internal/MvcDataAnnotationsMvcOptionsSetup.cs
kirthik: Yes they can.

src/Microsoft.AspNet.Mvc.DataAnnotations/MvcDataAnnotationsLocalizationOptions.cs
kirthik: This was part of @DamianEdwards design.

src/Microsoft.AspNet.Mvc.ViewFeatures/Internal/MvcViewOptionsSetup.cs
kirthik: Answered above.

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsModelValidatorProvider.cs
kirthik: Will sync up offline about this one.

src/Microsoft.AspNet.Mvc.DataAnnotations/MvcDataAnnotationsLocalizationOptions.cs
DamianEdwards: We have it in the logger providers for filtering. This is the simplest thing we can do that enables anything without having to introduce an entire provider pattern (like we have in the localization middleware for example).

src/Microsoft.AspNet.Mvc.DataAnnotations/Internal/MvcDataAnnotationsLocalizationOptionsSetup.cs
kirthik: yes?

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsModelValidatorProvider.cs
rynowak: That doesn't have anything to do with what kind of type it is. 

Validators are either retrieved for a property or a type. If you're looking at a property, then you'll have a container type.

test/WebSites/LocalizationWebSite/Models/User.cs
kirthik: This will not work today because of a bug in https://github.com/aspnet/Localization/blob/dev/src/Microsoft.Framework.Localization/ResourceManagerStringLocalizerFactory.cs#L53

I filed a bug https://github.com/aspnet/Localization/issues/90

kirthik:
@rynowak, @NTaylorMullen updated

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsClientModelValidatorProvider.cs
```diff
@@ -18,11 +20,27 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
...
-        internal delegate IClientModelValidator
-            DataAnnotationsClientModelValidationFactory(ValidationAttribute attribute);
+        internal delegate IClientModelValidator DataAnnotationsClientModelValidationFactory(
+            ValidationAttribute attribute,
+            IStringLocalizer stringLocalizer);
 
         private readonly Dictionary<Type, DataAnnotationsClientModelValidationFactory> _attributeFactories =
             BuildAttributeFactoriesDictionary();
+        private IOptions<MvcDataAnnotationsLocalizationOptions> _options;
+        private IStringLocalizerFactory _stringLocalizerFactory;
```
rynowak: These can both be made `readonly`

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsClientModelValidatorProvider.cs
rynowak: This is still here.

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsClientModelValidatorProvider.cs
```diff
@@ -36,6 +54,16 @@ public void GetValidators(ClientValidatorProviderContext context)
             {
                 throw new ArgumentNullException(nameof(context));
             }
+            IStringLocalizer stringLocalizer = null;
+            if (_options != null &&
+                _options.Value.DataAnnotationLocalizerProvider != null &&
+                _stringLocalizerFactory != null)
+            {
+                // This will pass first non-null type (either containerType or modelType) to delegate.
```
rynowak: This isn't a useful comment. - **WHY does it do this and why is that the right thing to do?**

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsModelValidatorProvider.cs
rynowak: What was the outcome of this dicussion?

test/Microsoft.AspNet.Mvc.DataAnnotations.Test/DataAnnotationsModelValidatorProviderTest.cs
```diff
@@ -21,7 +21,7 @@ public class DataAnnotationsModelValidatorProviderTest
         public void GetValidators_ReturnsValidatorForIValidatableObject()
         {
             // Arrange
-            var provider = new DataAnnotationsModelValidatorProvider();
+            var provider = new DataAnnotationsModelValidatorProvider(options: null, stringLocalizerFactory: null);
```
rynowak: why null?

test/WebSites/LocalizationWebSite/Resources/User.fr.resx
rynowak: Discussed in person. We don't think there's any case where data from the model makes its way into one of these strings.

rynowak:
:shipit: from me after fixing minor comments

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsModelValidatorProvider.cs
kirthik: Move this logic to `DataAnnotationsModelValidator`

src/Microsoft.AspNet.Mvc.DataAnnotations/Internal/MvcDataAnnotationsLocalizationOptionsSetup.cs
kirthik: forgot updating this. I will update it next iteration

NTaylorMullen:
:shipit:

kirthik:
Thanks @rynowak and @NTaylorMullen. Merged 0889b18f9523c4a1418e74a446814eea9c9084d0

Issue #2945, opened 2015-08-10
Make removing a formatter a 1-liner


rynowak:
See code sample here: https://github.com/aspnet/Mvc/issues/2944#issuecomment-129499220

When we were using descriptors for formatters we had a 1-liner extension method for removing a formatter by-type from the collection. We should add this back by defining custom collection types for input formatters and output formatters.

Krusen:
Hi, I'm new to contributing to other projects but would like to give this a shot.

Were you thinking something like this?

```C#
namespace Microsoft.AspNet.Mvc.Core.Formatters
{
    public class FormatterCollection<TFormatter> : Collection<TFormatter>
    {
        /// <summary>
        /// Removes all formatters of the specified type.
        /// </summary>
        /// <typeparam name="T">The type to remove.</typeparam>
        public void RemoveType<T>() where T : TFormatter
        {
            var formatters = this.OfType<T>();
            foreach (var formatter in formatters)
            {
                Remove(formatter);
            }
        }
    }
}
```

```C#
public MvcOptions()
{
    // ...
    InputFormatters = new FormatterCollection<IInputFormatter>();
    OutputFormatters = new FormatterCollection<IOutputFormatter>();
    // ...
}

public FormatterCollection<IInputFormatter> InputFormatters { get; }
public FormatterCollection<IOutputFormatter> OutputFormatters { get; }

// ...
```

Should the `IList<IInputFormatter>`/`IList<IOutputFormatter>` properties in `Microsoft.AspNet.Mvc.Filters.ResourceExecutingContext` and `Microsoft.AspNet.Mvc.Filters.ActionBindingContext` be changed as well?

rynowak:
@Krusen - yes, that's pretty much exactly what I had in mind.

/cc @Eilon in case he has feedback on the approach. This also follows the design of some other places where we were getting rid of extension methods that you need to use from options/startup code.

Eilon:
I'm cool with this. BTW I'm not sure the proposed code will necessarily work: you might have to force `.ToList()` after calling `.OfType()` so that all the LINQ evaluation will be complete before modifying the collection.

Krusen:
Great. I'll go ahead and make a pull request later today.

@Eilon You are right, it needs to be changed.

davidfowl:
Avoid using LINQ if you can. If you use a `List<T>` you can use RemoveAll

rynowak:
[...]

PR #3205, opened 2015-09-25
Add FormatterCollection<T> with RemoveType<T> method

Krusen:
#2945

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/Formatters/FormatterCollection.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Formatters/FormatterCollection.cs
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Collections.ObjectModel;
-using System.Linq;
 
 namespace Microsoft.AspNet.Mvc.Formatters
 {
@@ -19,10 +18,13 @@ namespace Microsoft.AspNet.Mvc.Formatters
         /// <typeparam name="T">The type to remove.</typeparam>
         public void RemoveType<T>() where T : TFormatter
         {
-            var formatters = this.OfType<T>().ToList();
-            foreach (var formatter in formatters)
+            for (var i = Count - 1; i >= 0; i--)
             {
-                Remove(formatter);
+                var formatter = this[i];
+                if (formatter is T)
+                {
+                    RemoveAt(i);
+                }
             }
         }
     }
--- a/src/Microsoft.AspNet.Mvc.Core/Formatters/FormatterCollection.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Formatters/FormatterCollection.cs
@@ -1,0 +1,19 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.AspNet.Mvc.Formatters
+{
+    public class FormatterCollection<TFormatter> : List<TFormatter>
+    {
+        /// <summary>
+        /// Removes all formatters of the specified type.
+        /// </summary>
+        /// <typeparam name="T">The type to remove.</typeparam>
+        public void RemoveType<T>() where T : TFormatter
+        {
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.Mvc.Core/Formatters/FormatterCollection.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.AspNet.Mvc.Formatters
+{
+    public class FormatterCollection<TFormatter> : List<TFormatter>
```
rynowak: I'd really prefer that this inherit from `Collection<T>` like in the sample you provided. `List<T>` exposes extra surface area that we have to assume users take a dependency on.

rynowak: Also, doc header on the class.

rynowak:
:watch:

src/Microsoft.AspNet.Mvc.Core/Formatters/FormatterCollection.cs
rynowak: And if you really still want to be able to call `RemoveAll` you can use this constructor: https://msdn.microsoft.com/en-us/library/ms132401(v=vs.110).aspx and keep a reference to the list you create.

Krusen:
Changed it back to my original proposal using `Collection<T>` and LINQ and added a little documentation on the class.

src/Microsoft.AspNet.Mvc.Core/Formatters/FormatterCollection.cs
```diff
@@ -0,0 +1,29 @@
...
+    public class FormatterCollection<TFormatter> : Collection<TFormatter>
+    {
+
+        /// <summary>
+        /// Removes all formatters of the specified type.
+        /// </summary>
+        /// <typeparam name="T">The type to remove.</typeparam>
+        public void RemoveType<T>() where T : TFormatter
+        {
+            var formatters = this.OfType<T>().ToList();
```
davidfowl:
```
            for (int i = Count - 1; i >= 0; i--)
            {
                var formatter = this[i];
                if(formatter is T)
                {
                    Remove(formatter);
                }
            }
```

Krusen:
Removed LINQ usage as proposed by @davidfowl.

davidfowl:
@Krusen can you add a test?

Krusen:
@davidfowl Yes, of course. I have added a test. Let me know if it needs more tests or anything else.

rynowak:
:shipit: - looks good to me. I'll merge it shortly.

test/Microsoft.AspNet.Mvc.Core.Test/Formatters/FormatterCollectionTests.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            public override Task WriteResponseBodyAsync([NotNull] OutputFormatterContext context)
+            {
+                throw new NotImplementedException();
+            }
+        }
+
+        private class AnotherTestOutputFormatter : OutputFormatter
+        {
+            public override Task WriteResponseBodyAsync([NotNull] OutputFormatterContext context)
```
dougbu: @rynowak could you include a commit to remove the `[NotNull]` use in test code? Not something we've done before and now's an odd time to start.

dougbu:
@rynowak also probably better to hold off pushing 'til we get Travis clean again.

test/Microsoft.AspNet.Mvc.Core.Test/Formatters/FormatterCollectionTests.cs
rynowak: Sure

rynowak:
Yeah, I'm looking at travis now. I suspect their `RuntimeHelpers.GetHashCode` doesn't handle null gracefully.

rynowak:
Merged: ec18b3512307f7cac66c26c0668b9bca05b52101 thanks!

PR #3206, opened 2015-09-25
Replace NotNullAttribute with thrown exceptions

pranavkm:


src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
```diff
@@ -1104,10 +1238,15 @@ internal static object GetModelStateValue(ViewContext viewContext, string key, T
         }
 
         protected virtual TagBuilder GenerateLink(
-            [NotNull] string linkText,
-            [NotNull] string url,
+            string linkText,
+            string url,
```
pranavkm: Design change: the AnchorTagHelper can produce links that do not have urls.

rynowak: I'm not sure about that. Isn't this supposed to throw?

`IUrlHelper` can indeed return null.

pranavkm: Ok. We'll need to update a couple of functional tests in that case. Essentially if you have an anchor tag helper with incomplete information such as: `<a asp-controller="something">Some value</a>` - we'll throw a Arg Null Exception.

cc @NTaylorMullen

NTaylorMullen: Oooo definitely not good, we can't be throwing in `TagHelper`s like that. If we get incomplete data we'll typically use logging to notify the user that incomplete information was provided to the `TagHelper`. For example the `ScriptTagHelper` uses a mode matcher to [log details](https://github.com/aspnet/Mvc/blob/a68d9e4cb1328b4f25683ac7eab422c190cd6eb6/src/Microsoft.AspNet.Mvc.TagHelpers/Internal/ModeMatchResult.cs#L59-L68). `AnchorTagHelper` shouldn't do the same thing but it definitely shouldn't explode :smile:

pranavkm: I'll leave this as is for now and file a bug to change the design of anchor tag helper which includes a follow up to add a ArgNotNull here.

pranavkm: Filed - https://github.com/aspnet/Mvc/issues/3208

pranavkm:
Checking this in to avoid ongoing merge conficts. I'll leave the PR open

Issue #2969, opened 2015-08-14
Need support for specifying custom error message for client validation


ajaybhargavb:
Related issue - #2817 
We now have specific error messages for client validation. 
Eg: `The field FieldName must be a number`
There needs to be a way for users to customize the error message that are sent with the `data-val` attributes.

MVC 5 has `GetUserResourceString` for this purpose.
http://aspnetwebstack.codeplex.com/SourceControl/latest#src/System.Web.Mvc/ClientDataTypeModelValidatorProvider.cs
```C#
return GetUserResourceString(controllerContext, "FieldMustBeNumeric") ?? MvcResources.ClientDataTypeModelValidatorProvider_FieldMustBeNumeric;
```

Need to have something similar in MVC 6.

cc' @pranavkm @Eilon

dougbu:
@Eilon I could do one message in Beta 8 for #2303 and fill it in for this bug during RC1. Was that the intent in placing this in RC1? If not I'll bring this into Beta 8.

Separately @Eilon and @rynowak do you agree we no longer need the "needs design" label?

Eilon:
I removed the needs design label.

Yeah I was thinking do just the one for beta8, and the rest in RC. If it's somehow completely trivial to do all at once, that's fine with me too.

dougbu:
28aec3f

dougbu:
Reopening. 28aec3f doesn't cover the specific message mentioned above.

jln1989:
@ajaybhargavb @dougbu 

Is customizing the error message available for any one message in Beta8 or else it will be available in RC1 completely? If its available in Beta8 can you tell me what is the specific message and how to use it to override the default message?

dougbu:
@jln1989 customizing error messages is currently available only for a few messages. This issue remains Open because we have not done the work for client-side validation messages.

Have a look at the `Microsoft.AspNet.Mvc.ModelBinding.Metadata.IModelBindingMessageProvider` interface for the few customizations currently available. The easiest way to make changes is e.g.
``` c#
    public void ConfigureServices(IServiceCollection services)
    {
        // Add MVC services to the services container
        services.AddMvc(mvcOptions =>
        {
            mvcOptions.ModelBindingMessageProvider.MissingBindRequiredValueAccessor =
                name => $"A value for the '{ name }' property was not provided.";
        });
    }
```

Of course resources or `IStringLocalizer` implementations may be better in your scenario.

dougbu:
[...]

PR #3209, opened 2015-09-25
Support user overrides of system-provided `ModelError` messages

dougbu:
- #2969
- add ~~`ValidationErrorMetadata`~~ ~~`ModelBindingMessages`~~ `ModelBindingMessageProvider` for configuration or ~~`IValidationMetadataProvider`~~ `IBindingMetadataProvider` overrides
- add ~~`ValidationErrorMessages`~~ ~~`IModelBindingMessages`~~ `IModelBindingMessageProvider` property to `ModelMetadata` for use in rest of the product code
- plumb the various bits through the system

~~WIP~~
- ~~names: tempted to rename to `ModelBindingErrorMetadata` and `ModelBindingErrorMessages`~~ (done)
  - ~~move `ModelBindingErrorMetadata` into `BindingMetadata`~~
  - ~~move some plumbing into `DefaultBindingMetadataProvider`~~
  - ~~Thoughts? !!~~
- ~~src compiles but test does not~~ (done)
- ~~test new override capabilities~~ (done)
- ~~doc comments~~ (done)
- ~~move commented-out reminders from `ValidationErrorMessages` into a follow-up issue~~ #3215
  - ~~or figure out how to plumb `ModelMetadata` into the relevant product code~~


src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
```diff
@@ -528,5 +529,35 @@ public override IReadOnlyList<object> ValidatorMetadata
...
             }
         }
+
+        public override ValidationErrorMessages ValidationErrorMessages
+        {
+            get
+            {
+                if (_validationErrorMessages == null)
+                {
+                    _validationErrorMessages = new DefaultErrorMessages(ValidationMetadata.ValidationErrorMetadata);
```
dougbu: I'm somewhat ambivalent about creating this read-only wrapper. Could instead change `ValidationErrorMetadata` to `DefaultErrorMetadata` and have it subclass `ValidationErrorMessages` but provide setters. Thoughts?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultValidationMetadataProvider.cs
```diff
@@ -12,15 +11,24 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
...
+
+        public DefaultValidationMetadataProvider(ValidationErrorMetadata validationErrorMetadata)
+        {
+            _validationErrorMetadata = validationErrorMetadata;
+        }
+
         /// <inheritdoc />
         public void GetValidationMetadata([NotNull] ValidationMetadataProviderContext context)
         {
+            context.ValidationMetadata.ValidationErrorMetadata = _validationErrorMetadata;
```
dougbu: Main oddity here is that a later `IValidationMetadataProvider` that wants to change a single `Func<string>` must first replace the containing `ValidationErrorMetadata` with a new instance. Thoughts on the following options?

1. Lots of docs about per-metadata customization of this property.
2. Create a new `ValidationErrorMetadata` instance here, avoiding the issue but creating a fair number of new objects that will be identical in most cases. (This option should be very safe: Hard to imagine a configuration delegate running after the first invocation of this method. Just not sure we need that level of safety here.)

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Metadata/ValidationErrorMessages.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// <value>Default <see cref="string"/> is "A value is required.".</value>
+        public abstract Func<string> MissingKeyOrValueResource { get; }
+
+        // See use of Resources.Common_ValueNotValidForProperty in ValidationHelpers.
+        // <summary>
+        // Fallback error message HTML and tag helpers display when a property is invalid but the
+        // <see cref="ModelError"/>s have <c>null</c> <see cref="ModelError.ErrorMessage"/>s.
+        // </summary>
+        // <value>Default <see cref="string"/> is "The value '{0}' is invalid.".</value>
+        ////public abstract Func<string> ValueInvalid_UnknownErrorResource { get; }
```
dougbu: One way to avoid this framework-provided string entirely would be to use the next two `Func`s to ensure `ModelError`s never have a `null` or empty `ErrorMessage`. Then `DefaultHtmlGenerator` and `ValidationHelpers` would never need to fill a gap the model binding system left.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Metadata/ValidationErrorMessages.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        // See how Resources.ModelError_InvalidValue_MessageWithModelValue is used
+        // <summary>
+        // Replacement <see cref="ModelError.ErrorMessage"/> used in <see cref="ModelError"/> when
+        // <see cref="ModelError.Exception"/> is of type <see cref="FormatException"/> and value is known i.e. when
+        // the bound value could not be converted and the "replacer" has the attempted value.
+        // </summary>
+        // <value>Default <see cref="string"/> is "The value '{0}' is not valid for {1}.".</value>
+        ////public abstract Func<string> ValueInvalid_WithValueResource { get; }
```
dougbu: The relevant code here is `ModelStateDictionary.AddModelError(... Exception ...)`. That method isn't currently passed the associated `ModelMetadata` but could be. Would affect all the call sites but keep the special cases in one place.

Note the vagueness of the message added given that code currently runs only when passed a `FormatException`. Should we use a different resource for that case e.g. `The value '{0}' could not be converted for assignment to {1}.`?

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Metadata/ValidationErrorMessages.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        // See how Resources.ModelError_InvalidValue_GenericMessage is used.
+        // <summary>
+        // Replacement <see cref="ModelError.ErrorMessage"/> used in <see cref="ModelError"/> when
+        // <see cref="ModelError.Exception"/> is of type <see cref="FormatException"/> and value is unknown i.e. when
+        // the bound value could not be converted and the "replacer" does not have the attempted value.
+        // </summary>
+        // <value>Default <see cref="string"/> is "The supplied value is invalid for {0}.".</value>
+        ////public abstract Func<string> ValueInvalid_WithoutValueResource { get; }
```
dougbu: If changing the `AddModelError(... Exception ...)` method to pass the `ModelMetadata` in as well, could also require the `AttemptedValue`. Might only need `ValueInvalid_WithValueResource` (well, renamed to `ValueInvalidResource` :smiley_cat: ).

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Metadata/ValidationErrorMessages.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// <see cref="BindRequiredAttribute"/> is not bound.
+        /// </summary>
+        /// <value>Default <see cref="string"/> is "A value for the '{0}' property was not provided.".</value>
+        public abstract Func<string> MissingBindRequiredValueResource { get; }
+
+        /// <summary>
+        /// Error message the model binding system adds when either the key or the value of a
+        /// <see cref="System.Collections.Generic.KeyValuePair{TKey, TValue}"/> is bound but not both.
+        /// </summary>
+        /// <value>Default <see cref="string"/> is "A value is required.".</value>
```
dougbu: I plan to change this resource and `ValueInvalid_MustNotBeNullResource` to include the key. As-is these are not very useful in a validation summary.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
```diff
@@ -420,7 +420,10 @@ internal static PropertyValidationInfo GetPropertyValidationInfo(ModelBindingCon
 
                 bindingContext.ModelState.TryAddModelError(
                     modelStateKey,
-                    Resources.FormatModelBinding_MissingBindRequiredMember(propertyName));
+                    string.Format(
+                        CultureInfo.CurrentCulture,
+                        propertyExplorer.Metadata.ValidationErrorMessages.MissingBindRequiredValueResource(),
+                        propertyName));
```
dougbu: Tempting to make the resource `Func`s take the arguments and avoid these shenanigans. Any objection?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultValidationMetadataProvider.cs
rynowak: I'd just create the copy. It's an extra object, but it happens once per model and it's cached.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/ValidationErrorMetadata.cs
```diff
@@ -0,0 +1,31 @@
...
+namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
+{
+    public class ValidationErrorMetadata
+    {
+        /// <summary>
+        /// Error message the model binding system adds when a property with an associated
+        /// <see cref="BindRequiredAttribute"/> is not bound.
+        /// </summary>
+        /// <value>Default <see cref="string"/> is "A value for the '{0}' property was not provided.".</value>
+        public Func<string> MissingBindRequiredValueResource { get; set; }
```
rynowak: We need to make sure that these can't be set to null. It would be really hard to debug (for a user) if they did something silly here and got random 500s because of a nullref.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
rynowak: Yeah I think your suggested change is a good choice if you decide to make a defensive copy inside of `DefaultValidationMetadataProvider`.

Due to how `new` and `abstract` interact, I think you'll need 3 classes.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
rynowak: Defining funcs of the proper arity is better than demanding that you give us a resource string with the right number of parameters in it. :+1:

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Metadata/ValidationErrorMessages.cs
rynowak: I think the message is fine, we're intentionally vague about everything the use didn't ask us to expose.

rynowak:
:watch: - this looks like it's in a pretty good place already.

As far as distinguishing between binding/validation - I don't feel strongly. I lean towards binding because all these errors come from missing data or failed conversions, but I think there's an argument to be made for either.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/ValidationErrorMetadata.cs
dougbu: :100: % agreed. Should have included `null`-checking in my WIP list.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Metadata/ValidationErrorMessages.cs
dougbu: #3215 for this and other comments on the code comments from this iteration

dougbu:
Updated; no longer in design phase. /cc @rynowak

test/Microsoft.AspNet.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
```diff
@@ -1689,7 +1731,7 @@ private class Person11
             Assert.Null(entry.RawValue);
             Assert.Null(entry.AttemptedValue);
             var error = Assert.Single(modelState["Customer.Name"].Errors);
-            Assert.Equal("A value for the 'Name' property was not provided.", error.ErrorMessage);
+            Assert.Equal("A value for the 'Customer.Name' property was not provided.", error.ErrorMessage);
```
dougbu: This illustrates the utility of the full path better than all of the tests with prefixes and custom binder names.

dougbu:
Not sure exactly what's going on w/ Travis but it looks like the problems started with your "Rewrite of validation" push (8a502db) @rynowak  See for example the failures in https://travis-ci.org/aspnet/Mvc/jobs/82253265 That was the next `dev` push after @kirthik's successful Travis builds and failures since then look consistent. The Mac failures include a `DefaultObjectValidator` stack trace. Likely those changes hit up against a Mono issue though there's nothing obvious (open or closed) at [aspnet/External/issues](https://github.com/aspnet/External/issues?utf8=%E2%9C%93&q=is%3Aissue+label%3Amono+).

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Metadata/IModelBindingMessages.cs
```diff
@@ -0,0 +1,34 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
+{
+    /// <summary>
+    /// Messages for errors the model binding system detects.
+    /// </summary>
+    public interface IModelBindingMessages
```
rynowak: An interface for this kinda makes me sad. I understand why you ended up going that route

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Metadata/IModelBindingMessages.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        /// <summary>
+        /// Error message the model binding system adds when a property with an associated
+        /// <c>BindRequiredAttribute</c> is not bound.
+        /// </summary>
+        /// <value>Default <see cref="string"/> is "A value for the '{0}' property was not provided.".</value>
+        Func<object, string> MissingBindRequiredValueResource { get; }
```
rynowak: Why are these object? Don't we always pass in the property name?

dougbu: Matching the signature of the `Resources` properties. IOW doing as resx generators do.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Metadata/IModelBindingMessages.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// </summary>
+        /// <value>Default <see cref="string"/> is "A value is required for '{0}'.".</value>
+        Func<object, string> MissingKeyOrValueResource { get; }
+
+        /// <summary>
+        /// Error message the model binding system adds when a <c>null</c> value is bound to a
+        /// non-<see cref="Nullable"/> property.
+        /// </summary>
+        /// <value>Default <see cref="string"/> is "A null value is invalid for '{0}'.".</value>
+        Func<object, string> ValueInvalid_MustNotBeNullResource { get; }
```
rynowak: Underscore in a property name is teh wonky

src/Microsoft.AspNet.Mvc.Core/Resources.resx
```diff
@@ -147,14 +144,8 @@
...
-  <data name="ActionSelector_GetCandidateActionsIsAmbiguous" xml:space="preserve">
-    <value>The supplied route values are ambiguous and can select multiple sets of actions.</value>
-  </data>
-  <data name="ArgumentPropertyUnexpectedType" xml:space="preserve">
-    <value>Property '{0}' is of type '{1}', but this method requires a value of type '{2}'.</value>
-  </data>
-  <data name="Common_ValueNotValidForProperty" xml:space="preserve">
-    <value>The value '{0}' is invalid.</value>
+  <data name="ModelBinding_NullValueNotValid" xml:space="preserve">
+    <value>A null value is invalid for '{0}'.</value>
```
rynowak: Is this really actionable for the end user?

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/Metadata/DefaultBindingMetadataProviderTest.cs
```diff
@@ -514,6 +515,16 @@ public void GetBindingDetails_BindingBehaviorLeftAlone_ForAttributeOnPropertyTyp
...
 
+        private static ModelBindingMessages CreateBindingMessages()
+        {
+            return new ModelBindingMessages
+            {
+                MissingBindRequiredValueResource = Resources.FormatModelBinding_MissingBindRequiredMember,
+                MissingKeyOrValueResource = Resources.FormatKeyValuePair_BothKeyAndValueMustBePresent,
+                ValueInvalid_MustNotBeNullResource = Resources.FormatModelBinding_NullValueNotValid,
+            };
+        }
```
rynowak: Should we put this in testcommon somewhere? Maybe expose off of `TestMetadataProvider`?

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/MutableObjectModelBinderTest.cs
```diff
@@ -1139,7 +1139,7 @@ public void ProcessResults_ValueTypePropertyWithBindRequired_RequiredValidatorIg
                 .Value;
             var error = Assert.Single(entry.Errors);
             Assert.Null(error.Exception);
-            Assert.Equal("A value for the 'ValueTypeProperty' property was not provided.", error.ErrorMessage);
+            Assert.Equal("A value for the 'theModel.ValueTypeProperty' property was not provided.", error.ErrorMessage);
```
rynowak: Doesn't this make these error messages deviate from what things like `LabelFor(...)` would generate? Let's chat tomorrow, I'm not convinced that adding the prefix into these messages helps the end user.

dougbu: Yeah let's talk tomorrow since I don't understand what you mean. Note this change to `MutableObjectModelBinder` makes it more consistent with for example the `ErrorMessage` that `TryAddModelError()` adds in the `FormatException` special case.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/Metadata/DefaultBindingMetadataProviderTest.cs
dougbu: Can't do that because `Resources` isn't exposed (through IVF) to most test code. In any case there are only four `CreateBindingMessages()` methods, two using `Resources` and two not.

src/Microsoft.AspNet.Mvc.Core/Resources.resx
dougbu: Sure e.g. an `<input/>` element left blank can hit this.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Metadata/IModelBindingMessages.cs
dougbu: Will change.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Metadata/IModelBindingMessages.cs
Eilon: `IModelBindingMessageProvider`? This doesn't actually contain any messages.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Metadata/IModelBindingMessages.cs
Eilon: The resx generators do that only because they can't do anything else. This should be a string.

Eilon: Also, regarding naming, these properties are not themselves resources. They provide actual messages.

src/Microsoft.AspNet.Mvc.Core/Resources.resx
Eilon: End users don't understand `null`. Maybe more like `"The {0} field cannot be empty."`?

dougbu: Point is moot in next iteration. I've gone back to the previous messages.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Metadata/IModelBindingMessages.cs
dougbu: Are you thinking along `MissingBindRequiredValueMessage` or `MissingBindRequiredValueFunc` lines?

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Metadata/IModelBindingMessages.cs
dougbu: Hmm, you seem to be arguing the opposite w.r.t. property naming?

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Metadata/IModelBindingMessages.cs
Eilon: Do we have a naming pattern elsewhere for funcs? These are neither messages nor resources.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Metadata/IModelBindingMessages.cs
Eilon: I'm suggesting a new name: `IModelBindingMessageProvider`

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Metadata/IModelBindingMessages.cs
dougbu: Within the MVC repo, we don't currently have many `Func<...>` properties. (Most `Func` mentions are method parameters.) We have one `Func<..., bool>` property named `PropertyFilter` and another 3 or 4 named `...Delegate` or `...DelegateAsync`. So does `MissingBindRequiredValueDelegate` (and similar for the other two) win?

If so, should the interface be named `IModelBindingMessageDelegates`? (I'm fine w/ `IModelBindingMessageProvider` but am wondering about consistency.)

Eilon: `Provider` is fine for the type name.

For the properties, how about `SuchAndSuchMessageAccessor`?

dougbu: Sold

dougbu:
Updated @Eilon, @rynowak

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/ModelBindingMessageProvider.cs
```diff
@@ -0,0 +1,95 @@
...
+namespace Microsoft.AspNet.Mvc.ModelBinding.Metadata
+{
+    /// <summary>
+    /// Read / write <see cref="IModelBindingMessageProvider"/> implementation.
+    /// </summary>
+    public class ModelBindingMessageProvider : IModelBindingMessageProvider
+    {
+        private Func<string, string> _missingBindRequiredValueResource;
+        private Func<string> _missingKeyOrValueResource;
+        private Func<object, string> _valueInvalid_MustNotBeNullResource;
```
dougbu: Will rename backing fields to match properties...

src/Microsoft.AspNet.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
```diff
@@ -414,7 +412,7 @@ internal static PropertyValidationInfo GetPropertyValidationInfo(ModelBindingCon
 
                 bindingContext.ModelState.TryAddModelError(
                     modelStateKey,
-                    Resources.FormatModelBinding_MissingBindRequiredMember(propertyName));
+                    bindingContext.ModelMetadata.ModelBindingMessageProvider.MissingBindRequiredValueAccessor(propertyName));
```
dougbu: Will wrap

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Metadata/IModelBindingMessageProvider.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// </summary>
+        /// <value>Default <see cref="string"/> is "A value is required.".</value>
+        Func<string> MissingKeyOrValueAccessor { get; }
+
+        /// <summary>
+        /// Error message the model binding system adds when a <c>null</c> value is bound to a
+        /// non-<see cref="Nullable"/> property.
+        /// </summary>
+        /// <value>Default <see cref="string"/> is "The value '{0}' is invalid.".</value>
+        Func<object, string> ValueMustNotBeNullAccessor { get; }
```
rynowak: Is this supposed to have been changed to a string?

rynowak:
:shipit: from me after fixing small comments

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Metadata/IModelBindingMessageProvider.cs
dougbu: Yes. Previously we were passing the `model` into this one. Forgot to change this when I updated the code to pass `AttemptedValue` (really a `ValueProviderResult`'s `ToString()`, which is the same thing).

Fixing...

dougbu:
28aec3f

Issue #3043, opened 2015-08-26
ValidationSummary does not showcase all errors.


NTaylorMullen:
**Repro:**
- File => New Web Application.
- Add the following properties to `LoginPasswordViewModel.cs`:
```C#
public int Age { get; set; }

public uint Money { get; set; }
```  
- Add the following lines to `Login.cshtml`:
```HTML
<div class="form-group">
    <label asp-for="Age" class="col-md-2 control-label"></label>
    <div class="col-md-10">
        <input asp-for="Age" class="form-control" />
        <span asp-validation-for="Age" class="text-danger"></span>
    </div>
</div>

<div class="form-group">
    <label asp-for="Money" class="col-md-2 control-label"></label>
    <div class="col-md-10">
        <input asp-for="Money" class="form-control" />
        <span asp-validation-for="Money" class="text-danger"></span>
    </div>
</div>
```
- Run the application and go to Login
- Fill out the form as follows:  
![image](https://cloud.githubusercontent.com/assets/2008729/9502544/d8fe9d12-4be6-11e5-8a25-0a6436545952.png)
- Hit `Log In`
- You get this page:  
![image](https://cloud.githubusercontent.com/assets/2008729/9502564/fa710548-4be6-11e5-85e8-61bc2009ca73.png)
Note the lacking `-123` error in the validation summary at the top.

Due to lacking symbols I was only able to capture the call stacks of each server created error to help diagnose the impact of the error

`Age` error call stack:
![image](https://cloud.githubusercontent.com/assets/2008729/9502606/48020744-4be7-11e5-8f5f-ed87d6372338.png)

`Money` error call stack (thrown multiple times, here's just one of the call stacks):
![image](https://cloud.githubusercontent.com/assets/2008729/9502657/82e185f6-4be7-11e5-8b75-5ae70f747a34.png)

Most likely this error is due to the error message not being populated for `Money` in the model state. After all errors have been created the two errors on model state look as so:

![image](https://cloud.githubusercontent.com/assets/2008729/9502695/add41382-4be7-11e5-9c4f-0277aa51b1f1.png)

PR #3214, opened 2015-09-27
Validation summary displays generic error messages for ignored exceptions

ajaybhargavb:
#3043 

@dougbu / @NTaylorMullen

### DIFF
--- a/test/Microsoft.AspNet.Mvc.IntegrationTests/ValidationIntegrationTests.cs
+++ b/test/Microsoft.AspNet.Mvc.IntegrationTests/ValidationIntegrationTests.cs
@@ -992,6 +992,94 @@ namespace Microsoft.AspNet.Mvc.IntegrationTests
             Assert.True(modelState.IsValid);
         }
 
+        private class User
+        {
+            public int Id { get; set; }
+
+            public uint Zip { get; set; }
+
+        }
+
+        [Fact]
+        public async Task Validation_FormatException_ShowsInvalidValueMessage_OnSimpleTypeProperty()
+        {
+            // Arrange
+            var argumentBinder = ModelBindingTestHelper.GetArgumentBinder();
+            var parameter = new ParameterDescriptor()
+            {
+                Name = "parameter",
+                ParameterType = typeof(User)
+            };
+
+            var operationContext = ModelBindingTestHelper.GetOperationBindingContext(request =>
+            {
+                request.QueryString = new QueryString("?Id=bill");
+            });
+
+            var modelState = new ModelStateDictionary();
+
+            // Act
+            var modelBindingResult = await argumentBinder.BindModelAsync(parameter, modelState, operationContext);
+
+            // Assert
+            Assert.True(modelBindingResult.IsModelSet);
+
+            var model = Assert.IsType<User>(modelBindingResult.Model);
+            Assert.Equal(0, model.Id);
+            Assert.Equal(1, modelState.ErrorCount);
+            Assert.False(modelState.IsValid);
+
+            var state = Assert.Single(modelState);
+            Assert.Equal("Id", state.Key);
+            var entry = state.Value;
+            Assert.Equal("bill", entry.AttemptedValue);
+            Assert.Equal("bill", entry.RawValue);
+            Assert.Single(entry.Errors);
+
+            var error = entry.Errors[0];
+            Assert.Equal("The value 'bill' is not valid for Id.", error.ErrorMessage);
+        }
+
+        [Fact]
+        public async Task Validation_OverflowException_ShowsInvalidValueMessage_OnSimpleTypeProperty()
+        {
+            // Arrange
+            var argumentBinder = ModelBindingTestHelper.GetArgumentBinder();
+            var parameter = new ParameterDescriptor()
+            {
+                Name = "parameter",
+                ParameterType = typeof(User)
+            };
+
+            var operationContext = ModelBindingTestHelper.GetOperationBindingContext(request =>
+            {
+                request.QueryString = new QueryString("?Zip=-123");
+            });
+
+            var modelState = new ModelStateDictionary();
+
+            // Act
+            var modelBindingResult = await argumentBinder.BindModelAsync(parameter, modelState, operationContext);
+
+            // Assert
+            Assert.True(modelBindingResult.IsModelSet);
+
+            var model = Assert.IsType<User>(modelBindingResult.Model);
+            Assert.Equal<uint>(0, model.Zip);
+            Assert.Equal(1, modelState.ErrorCount);
+            Assert.False(modelState.IsValid);
+
+            var state = Assert.Single(modelState);
+            Assert.Equal("Zip", state.Key);
+            var entry = state.Value;
+            Assert.Equal("-123", entry.AttemptedValue);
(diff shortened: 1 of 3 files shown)


test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperValidationSummaryTest.cs
```diff
@@ -342,21 +347,21 @@ private void AddMultipleErrors(ModelStateDictionary modelState)
             modelState.AddModelError("Property3.OrderedProperty2", "This is an error for Property3.OrderedProperty2.");
 
             modelState.AddModelError("Property3", "This is an error for Property3.");
-            modelState.AddModelError("Property3", new InvalidCastException("Exception will be ignored."));
+            modelState.AddModelError("Property3", new InvalidCastException("Produces generic error message"));
```
dougbu: Why change these ignored messages?

ajaybhargavb: Because they are not actually ignored but are displayed as generic errors.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
```diff
@@ -802,7 +802,7 @@ public virtual IHtmlContent GenerateAntiforgery(ViewContext viewContext)
             {
                 foreach (var modelError in modelState.Errors)
                 {
-                    var errorText = ValidationHelpers.GetUserErrorMessageOrDefault(modelError, modelState: null);
+                    var errorText = ValidationHelpers.GetUserErrorMessageOrDefault(modelError, modelState);
```
dougbu: This is a change from validation summaries provided in MVC 5 and earlier. It's been consistent for validation messages (at field level) to include default "is invalid" message but to hide such messages in validation summaries. What I expected to fix #3043 was another special case in `TryAddModelError(... Exception ...)` for the  `Exception` type thrown in that scenario (`OverflowException`?), similar to `FormatException` handling but with a different message e.g. `Value '{0}' is out of range for '{1}'.`. @Eilon your thoughts?

If we decide to use current approach, should remove the `modelState == null` special case in `GetUserErrorMessageOrDefault()`. Also need an integration test demonstrating new behaviour for exact #3043 scenario (attempt to bind a negative value to a `uint` property).

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperValidationSummaryTest.cs
dougbu: You're quibbling but :ok:

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
ajaybhargavb: Checked MVC 5 behavior. It does not ignore the errors in validation summary.

image


Looked at the code and there is a special casing in `DefaultModelBinder` for Format and Overflow exceptions which adds the invalid messages. http://aspnetwebstack.codeplex.com/SourceControl/latest#src/System.Web.Mvc/DefaultModelBinder.cs
```C#
// We only consider "known" type of exception and do not make too aggressive changes here
if (exception is FormatException || exception is OverflowException)
{
    string displayName = propertyMetadata.GetDisplayName();
    string errorMessageTemplate = GetValueInvalidResource(controllerContext);
    string errorMessage = String.Format(CultureInfo.CurrentCulture, errorMessageTemplate, modelState.Value.AttemptedValue, displayName);
    modelState.Errors.Remove(error);
    modelState.Errors.Add(errorMessage);
    break;
}
```

dougbu: Let's start with fixing that issue.

We can research why the validation summary has historically been shorter than the sum of all validation messages and tackle that separately. Worth filing a separate Investigate item for that.

NTaylorMullen: :+1:

ajaybhargavb:
Updated.

NTaylorMullen:
Since the new change addresses [this](https://github.com/aspnet/Mvc/pull/3214#discussion_r40627250). :shipit:. 

Should track investigation of the second part of that comment to determine what we should or should not hide in MVC6.

test/Microsoft.AspNet.Mvc.IntegrationTests/ValidationIntegrationTests.cs
```diff
@@ -1086,5 +1162,13 @@ private static void AssertRequiredError(string key, ModelError error)
                 string.Format("The {0} field is required.", key)), error.ErrorMessage);
             Assert.Null(error.Exception);
         }
+
+        private class User
```
dougbu: Everywhere else in this test class, the target types are defined just before their first use. Not good practise in product code but it makes this long file easier to use. So move this above `Validation_FormatException_ShowsInvalidValueMessage_OnSimpleTypeProperty()`.

test/Microsoft.AspNet.Mvc.IntegrationTests/ValidationIntegrationTests.cs
```diff
@@ -992,6 +992,82 @@ private class Order10
...
+            Assert.Equal(0, model.Id);
+
+            Assert.Equal(1, modelState.Count);
+            Assert.Equal(1, modelState.ErrorCount);
+            Assert.False(modelState.IsValid);
+
+            var entry = Assert.Single(modelState, e => e.Key == "Id").Value;
+            Assert.Equal("bill", entry.AttemptedValue);
+            Assert.Equal("bill", entry.RawValue);
+            Assert.Single(entry.Errors);
```
dougbu: Better to use `var error =` here than `entry.Errors[0]` on the next line.

test/Microsoft.AspNet.Mvc.IntegrationTests/ValidationIntegrationTests.cs
```diff
@@ -992,6 +992,82 @@ private class Order10
...
+            Assert.True(modelBindingResult.IsModelSet);
+
+            var model = Assert.IsType<User>(modelBindingResult.Model);
+            Assert.Equal(0, model.Id);
+
+            Assert.Equal(1, modelState.Count);
+            Assert.Equal(1, modelState.ErrorCount);
+            Assert.False(modelState.IsValid);
+
+            var entry = Assert.Single(modelState, e => e.Key == "Id").Value;
```
dougbu: Checks here are redundant. Don't need `Assert.Equal(1, modelState.Count)`:
``` c#
var state = Assert.Single(modelState);
Assert.Equal("Id", state.Key);
var entry = state.Value;
```

dougbu:
:shipit:

PR #3216, opened 2015-09-28
Move ModelExplorer to ViewFeatures

rynowak:
This change removes the last vestiges of `ModelExplorer` from the model binding code, and moves it to ViewFeatures

### DIFF
--- a/test/Microsoft.AspNet.Mvc.DataAnnotations.Test/DataAnnotationsModelValidatorTest.cs
+++ b/test/Microsoft.AspNet.Mvc.DataAnnotations.Test/DataAnnotationsModelValidatorTest.cs
@@ -21,7 +21,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
         private static IModelMetadataProvider _metadataProvider = TestModelMetadataProvider.CreateDefaultProvider();
 
         [Fact]
-        public void ValuesSet()
+        public void Constructor_SetsAttribute()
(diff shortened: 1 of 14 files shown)


test/Microsoft.AspNet.Mvc.DataAnnotations.Test/DataAnnotationsModelValidatorTest.cs
```diff
@@ -83,15 +94,20 @@ public static IEnumerable<object[]> ValidateSetsMemberNamePropertyDataSet
         public void ValidateWithIsValidTrue()
```
pranavkm: Could you give this a meaningful name while you're modifying this type?

`ValidateReturnsEmptyValidationResultSequence_IfIsValidReturnsTrue`?

pranavkm: A bunch of tests are named this way. Maybe leave it as is

pranavkm:
:shipit:

test/Microsoft.AspNet.Mvc.DataAnnotations.Test/DataAnnotationsModelValidatorTest.cs
rynowak: I cleaned up more of these guys to have better names

PR #3219, opened 2015-09-28
Replace NotNullAttribute with thrown exceptions

pranavkm:


pranavkm:
cc @rynowak

src/Microsoft.AspNet.Mvc.WebApiCompatShim/ApiController.cs
```diff
@@ -160,8 +168,13 @@ public virtual ConflictResult Conflict()
         /// <param name="value">The content value to negotiate and format in the entity body.</param>
         /// <returns>A <see cref="NegotiatedContentResult{T}"/> with the specified values.</returns>
         [NonAction]
-        public virtual NegotiatedContentResult<T> Content<T>(HttpStatusCode statusCode, [NotNull] T value)
+        public virtual NegotiatedContentResult<T> Content<T>(HttpStatusCode statusCode, T value)
         {
+            if (value == null)
```
pranavkm: This surprisingly doesn't cause a compilation issue and works at runtime for value types.

dougbu: @pranavkm what does "works at runtime" mean? Does it throw if `value == default(T)` for example?

Not surprising there's no compilation warning. Compilers tend to be silent unless they're _sure_ you're doing something wrong.

rynowak: This isn't surprising at all - how else would you write null-safe generic code?

dougbu:
:shipit:

src/Microsoft.AspNet.Mvc.WebApiCompatShim/ApiController.cs
pranavkm: @dougbu, it just no-ops for value types.

PR #3221, opened 2015-09-28
Workaround Mono bugs to fix travis

rynowak:
This change works around two mono issues that are blocking travis. I'd
like to have tests skip instead, but unfortunately that would mean not
running any validation tests on mono so this seems better.

- RuntimeHelpers.GetHashCode will sometimes crash the runtime
- PropertyHelper sometimes throws null-ref

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultComplexObjectValidationStrategy.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultComplexObjectValidationStrategy.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Reflection;
 
 namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
 {
@@ -12,6 +13,8 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
     /// </summary>
     public class DefaultComplexObjectValidationStrategy : IValidationStrategy
     {
+        private static readonly bool IsMono = Type.GetType("Mono.Runtime") != null;
+
         /// <summary>
         /// Gets an instance of <see cref="DefaultComplexObjectValidationStrategy"/>.
         /// </summary>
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultComplexObjectValidationStrategy.cs
```diff
@@ -78,7 +81,34 @@ public bool MoveNext()
...
+                // This is a workaround for what the PropertyGetter does in the background.
+                if (IsMono)
+                {
+                    if (_model == null)
+                    {
+                        model = null;
+                    }
+                    else
+                    {
+                        var propertyInfo = _model.GetType().GetRuntimeProperty(property.PropertyName);
```
rynowak: Note that neither version of this code gracefully handles set-only properties. Will address separately.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/ReferenceEqualityComparer.cs
```diff
@@ -25,6 +28,13 @@ public new bool Equals(object x, object y)
 
         public int GetHashCode(object obj)
         {
+            // RuntimeHelpers.GetHashCode sometimes crashes the runtime on Mono 4.0.4
+            // The workaround here is to just not hash anything, and fall back to an equality check.
+            if (IsMono)
+            {
+                return 0;
+            }
+
```
rynowak: Will follow up separately on moving this to common.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultComplexObjectValidationStrategy.cs
```diff
@@ -78,7 +81,34 @@ public bool MoveNext()
                 var property = _properties[_index];
                 var propertyName = property.BinderModelName ?? property.PropertyName;
                 var key = ModelNames.CreatePropertyModelName(_key, propertyName);
-                var model = property.PropertyGetter(_model);
+
+                object model;
+
+                // Our property accessors don't work on Mono 4.0.4 - see https://bugzilla.xamarin.com/show_bug.cgi?id=25717
```
dougbu: Please reference an aspnet/External#xyz bug rather than the Mono issue.

rynowak: Sure, I just created one, so that will fit here nicely.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultComplexObjectValidationStrategy.cs
```diff
@@ -12,6 +13,8 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
     /// </summary>
     public class DefaultComplexObjectValidationStrategy : IValidationStrategy
     {
+        private static readonly bool IsMono = Type.GetType("Mono.Runtime") != null;
```
dougbu: The `IsMono` field in `SymbolsUtility` is `Lazy<bool>` but it's unlikely someone would use MVC without validation. Not sure whether that optimization is useful in `SymbolsUtility` or necessary here. Thoughts?

rynowak: static readonly fields are already lazy. They are evaluated on first use of the class.

dougbu: Yeah, the question here is: Will anything end up using `DefaultComplexObjectValidationStrategy` without hitting the `IsMono` check?

rynowak: The question here is, why are we having a discussion about performance without evidence that there's an issue that needs solving :confounded: This is a total non-issue.

dougbu:
:shipit: once Travis actually passes :smiling_imp:

PR #3222, opened 2015-09-28
Fix #3217 - Optimize IList.GetEnumerator allocations

### LINKED ISSUE #3217: Optimize IList.GetEnumerator allocations
Unfortunately, this code allocates:
```
var list = (IList<object>)new List<object>() { ... };
foreach (var item in list)
{
    ...
}
```

This is easily fixed by using a for-loop instead.

These add up to significant amount of allocations for an MVC app if you look at the framework as a whole:
image

 This profile includes allocations for 3001 requests on a small page doing model binding, validation, and rendering a view. It's about 2.5% of allocated bytes.

We should go through all of the instances that show up in this kind of profile, and do the trivial fix of turning it into a for-loop. By my count there are about 20 instances. Some of them use linq, which means we should be able to get a further reduction.

rynowak:
This change fixes call sites on the main request path for a simple site
(model binding, validation, views) that allocate boxed list enumerators.

Some cases aren't addressed by this change because the fix is too invasive
or requires changing an important contract to take IList instead of
IEnumerable. Will follow up on those case by case in order of importance.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.DataAnnotations/DefaultClientModelValidatorProvider.cs
+++ b/src/Microsoft.AspNet.Mvc.DataAnnotations/DefaultClientModelValidatorProvider.cs
@@ -22,9 +22,10 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
                 throw new ArgumentNullException(nameof(context));
             }
 
-            foreach (var metadata in context.ValidatorMetadata)
+            //Perf: Avoid allocations here
+            for (var i = 0; i < context.ValidatorMetadata.Count; i++)
             {
-                var validator = metadata as IClientModelValidator;
+                var validator = context.ValidatorMetadata[i] as IClientModelValidator;
                 if (validator != null)
                 {
                     context.Validators.Add(validator);
--- a/src/Microsoft.AspNet.Mvc.Razor/RazorViewEngine.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/RazorViewEngine.cs
@@ -198,17 +198,24 @@ namespace Microsoft.AspNet.Mvc.Razor
             }
             else
             {
-                // For traditional routes, lookup the key in RouteConstraints if the key is RequireKey.
-                var match = actionDescriptor.RouteConstraints.FirstOrDefault(
-                    constraint => string.Equals(constraint.RouteKey, key, StringComparison.OrdinalIgnoreCase));
-                if (match != null && match.KeyHandling != RouteKeyHandling.CatchAll)
+                // Perf: Avoid allocations
+                for (var i = 0; i < actionDescriptor.RouteConstraints.Count; i++)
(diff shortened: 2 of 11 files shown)


rynowak:
**Before**
image


**After**
image

That's about 830 bytes/request

src/Microsoft.AspNet.Mvc.DataAnnotations/DefaultClientModelValidatorProvider.cs
```diff
@@ -22,9 +22,10 @@ public void GetValidators(ClientValidatorProviderContext context)
                 throw new ArgumentNullException(nameof(context));
             }
 
-            foreach (var metadata in context.ValidatorMetadata)
+            //Perf: Avoid allocations here
```
pranavkm: Space after comment

pranavkm:
:shipit:

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngine.cs
```diff
@@ -198,17 +198,24 @@ public virtual IEnumerable<string> AreaViewLocationFormats
             }
             else
             {
-                // For traditional routes, lookup the key in RouteConstraints if the key is RequireKey.
```
dougbu: Is this comment no longer relevant?

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
```diff
@@ -800,8 +800,10 @@ public virtual IHtmlContent GenerateAntiforgery(ViewContext viewContext)
             var htmlSummary = new TagBuilder("ul");
             foreach (var modelState in modelStates)
             {
-                foreach (var modelError in modelState.Errors)
+                // Perf: Avoid allocations
+                for (var i = 0; i < modelState.Errors.Count; i++)
```
dougbu: This one shouldn't be a hot path. Never have more than one validation summary in a page.

dougbu:
Couple of minor late comments but would have said :shipit:

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngine.cs
rynowak: It was never super relevant to begin with

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
rynowak: It's inside a for-loop. That's one enumerator allocation per modelstate entry. This was allocating 3 enumerators per request for a very simple site.

PR #3223, opened 2015-09-28
[Fixes #3068] TempData fails silently without sessions middleware

### LINKED ISSUE #3068: TempData fails silently without sessions middleware
Following action results in empty response:
```c#
[HttpGet]
public async Task<IActionResult> Index()
{
    TempData["test"] = "test";

    return View(await messageBus.Query(new FindUsers()));
}
```

kichalla:
@rynowak @ajaybhargavb @Eilon

### DIFF
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/SessionStateTempDataProvider.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/SessionStateTempDataProvider.cs
@@ -59,17 +59,8 @@ namespace Microsoft.AspNet.Mvc.ViewFeatures
                 throw new ArgumentNullException(nameof(context));
             }
 
-            if (!IsSessionEnabled(context))
-            {
-                // Session middleware is not enabled. No-op
-                return null;
-            }
-
+            // Accessing Session property will throw if the session middleware is not enabled.
             var session = context.Session;
-            if (session == null)
-            {
-                return null;
-            }
 
             var tempDataDictionary = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
             byte[] value;
@@ -162,6 +153,9 @@ namespace Microsoft.AspNet.Mvc.ViewFeatures
                 throw new ArgumentNullException(nameof(context));
             }
 
+            // Accessing Session property will throw if the session middleware is not enabled.
+            var session = context.Session;
+
             var hasValues = (values != null && values.Count > 0);
             if (hasValues)
             {
@@ -170,9 +164,6 @@ namespace Microsoft.AspNet.Mvc.ViewFeatures
                     // We want to allow only simple types to be serialized in session.
                     EnsureObjectCanBeSerialized(item);
                 }
-
-                // Accessing Session property will throw if the session middleware is not enabled.
-                var session = context.Session;
 
                 using (var memoryStream = new MemoryStream())
                 {
@@ -183,16 +174,10 @@ namespace Microsoft.AspNet.Mvc.ViewFeatures
                     }
                 }
             }
-            else if (IsSessionEnabled(context))
+            else
             {
-                var session = context.Session;
                 session.Remove(TempDataSessionStateKey);
             }
-        }
-
-        private static bool IsSessionEnabled(HttpContext context)
-        {
-            return context.Features.Get<ISessionFeature>() != null;
         }
 
         internal void EnsureObjectCanBeSerialized(object item)
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/TempDataDictionary.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/TempDataDictionary.cs
@@ -101,7 +101,13 @@ namespace Microsoft.AspNet.Mvc.ViewFeatures
         /// <inheritdoc />
         public void Keep()
         {
-            Load();
+            // if the data is not loaded, we can assume none of it has been read
+            // and so silently return.
+            if (!_loaded)
+            {
+                return;
+            }
+
             _retainedKeys.Clear();
             _retainedKeys.UnionWith(_data.Keys);
         }
--- a/test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewFeatures/SessionStateTempDataProviderTest.cs
+++ b/test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewFeatures/SessionStateTempDataProviderTest.cs
@@ -16,56 +16,7 @@ namespace Microsoft.AspNet.Mvc.ViewFeatures
     public class SessionStateTempDataProviderTest
     {
         [Fact]
-        public void Load_NullSession_ReturnsEmptyDictionary()
-        {
-            // Arrange
-            var testProvider = new SessionStateTempDataProvider();
-
-            // Act
-            var tempDataDictionary = testProvider.LoadTempData(
-                GetHttpContext(session: null, sessionEnabled: true));
-
-            // Assert
-            Assert.Null(tempDataDictionary);
-        }
-
-        [Fact]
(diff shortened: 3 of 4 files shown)


src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/TempDataDictionary.cs
```diff
@@ -101,6 +101,13 @@ public ICollection<object> Values
         /// <inheritdoc />
         public void Keep()
         {
+            // if the data is not loaded, we can assume none of it has been read
+            // and so silently return.
+            if (!_loaded)
```
kichalla: This check is needed to prevent from throwing exception (ex: 'Session feature not enabled') in cases where use doesn't even use TempData...example: you could be redirecting to another action (this was caught in some failing tests).
https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/SaveTempDataFilter.cs#L45

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/SessionStateTempDataProvider.cs
```diff
@@ -59,16 +59,10 @@ public class SessionStateTempDataProvider : ITempDataProvider
...
-            {
-                // Session middleware is not enabled. No-op
-                return null;
-            }
-
             var session = context.Session;
             if (session == null)
             {
-                return null;
+                throw new InvalidOperationException("Session cannot be null.");
```
pranavkm: Localized string.

kichalla: good catch

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/SessionStateTempDataProvider.cs
```diff
@@ -59,16 +59,10 @@ public class SessionStateTempDataProvider : ITempDataProvider
                 throw new ArgumentNullException(nameof(context));
             }
 
-            if (!IsSessionEnabled(context))
-            {
-                // Session middleware is not enabled. No-op
-                return null;
-            }
-
             var session = context.Session;
```
ajaybhargavb: Wouldn't it just throw right here if session is not enabled?

rynowak: I seem to remember that detail from a previous PR. Do we have existing tests for this?

kichalla: yes and that is what we want in this scenario. So when the user does `TempData['key1']='value1'`, an exception is thrown to indicate that something has not been configured.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewFeatures/SessionStateTempDataProviderTest.cs
```diff
@@ -16,52 +16,54 @@ namespace Microsoft.AspNet.Mvc.ViewFeatures
...
             var testProvider = new SessionStateTempDataProvider();
 
-            // Act
-            var tempDataDictionary = testProvider.LoadTempData(
-                GetHttpContext(session: null, sessionEnabled: true));
-
-            // Assert
-            Assert.Null(tempDataDictionary);
+            // Act & Assert
+            var exception = Assert.Throws<InvalidOperationException>(() => testProvider.LoadTempData(
```
rynowak: Just use

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/SessionStateTempDataProvider.cs
ajaybhargavb: In what scenarios will the session be null?

kichalla: actually i can't think of any scenarios where you have session feature but session is null...

ajaybhargavb: Was thinking of the same thing. We don't need the condition below.

kichalla:
@rynowak @pranavkm @ajaybhargavb updated

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/TempDataDictionary.cs
pranavkm: Would you still need the call to `Load()` after this line?

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewFeatures/SessionStateTempDataProviderTest.cs
```diff
@@ -16,52 +16,45 @@ namespace Microsoft.AspNet.Mvc.ViewFeatures
...
             var testProvider = new SessionStateTempDataProvider();
 
-            // Act
-            var tempDataDictionary = testProvider.LoadTempData(
-                GetHttpContext(session: null, sessionEnabled: true));
-
-            // Assert
-            Assert.Null(tempDataDictionary);
+            // Act & Assert
+            var exception = Assert.Throws<InvalidOperationException>(() =>
```
pranavkm: Is this an exception from our stack? If so, could you verify the message?

pranavkm: If not, remove the unused `exception` variable. Similar changes for other tests.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewFeatures/TempDataDictionaryTest.cs
```diff
@@ -214,9 +239,8 @@ public void SaveTempData(HttpContext context, IDictionary<string, object> values
 
         private static IHttpContextAccessor GetHttpContextAccessor()
         {
-            var httpContext = new Mock<HttpContext>();
             var httpContextAccessor = new Mock<IHttpContextAccessor>();
```
pranavkm: `new HttpContextAccessor();`

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewFeatures/SessionStateTempDataProviderTest.cs
kichalla: Right, this exception is from HttpAbstractions and so did not want to verify the message..sure, will remove the variable

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/TempDataDictionary.cs
kichalla: good catch...it need not.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewFeatures/TempDataDictionaryTest.cs
kichalla: :+1:

kichalla:
Updated. @pranavkm @rynowak @ajaybhargavb

pranavkm:
:shipit:

kichalla:
@Eilon yeah, right...I based this PR's branch on the session PR...looks like something got messed up...will fix before checking in

kichalla:
130d94eb7e490508df88633d4dbbe6c061be2fdd

PR #3229, opened 2015-09-29
fix #3172: Remove mono check from localization test

### LINKED ISSUE #3172: Investigate the reason for localization functional test errors on Mono
`Localization_Resources_ReturnExpectedValues` test in `LocalizationTest` of functional test project fails on Mono. The same website when run separately works returns right results. So there is something going on with how functional tests retrieve resources. We need to investigate this issue more.

kirthik:
Removing the check since this issue could have been happening because of https://github.com/aspnet/dnx/issues/2802. I am not seeing this issue on mono now.


dougbu:
:shipit: once last Travis build succeeds (I retried one because failure _seemed_ unrelated)

kirthik:
Travis build succeeds. Merged fa5f9399e891f60fb9802c89696ba5b30ca2a4b6

PR #3231, opened 2015-09-30
[Design] Custom type for handling dictionaries of HTML attributes

rynowak:
TagBuilder currently uses `SortedDictionary<string, string>` for HTML attributes. Its memory usage pattern is pretty bad for our usage pattern (lots of small, short-lived objects).

Here's some numbers for a very small site with model binding and views - showing allocations for 3001 requests. This is all the flotsam created by these dictionaries:
image

Rough math says it's about 5.2mb or about 4.5% of total allocations. **something must be done**

===============================
The first approach that I tried here is use a `List<>` inside of a `IDictionary<,>` List is a fairly lean data structure to begin with.

image
image

You can see here that a number of these lists need to grow (reallocate) as items are added since the count of arrays > count of lists.

==============================
The second attempt is much more extreme, and uses fields inside the `AttributeDictionary` for the first 4 elements, only allocating the list after those 4 have been filled.

image
image

If you look at the second, you'll see far fewer of either are allocated. I have a bug in this example where the list is always being allocated - accounting for the discrepancy between arrays and lists. Assume that a good version of this would have only 9003 lists allocated.

=============================

At this point, more research and testing is needed. I wanted to get the idea out there to get feedback on the craziness of the idea. It's likely that the sweet spot here is around 6-8. Upping the number of 'inline' items to 6 might turn out to be much better for practical usage.

I'm happy to conclude that if we can't make the second option show a really **significant** improvement, then we should stick with the first one. :+1:

### DIFF
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/TagBuilder.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/TagBuilder.cs
@@ -17,6 +17,8 @@ namespace Microsoft.AspNet.Mvc.Rendering
     [DebuggerDisplay("{DebuggerToString()}")]
     public class TagBuilder : IHtmlContent
     {
+        private AttributeDictionary _attributes;
+
         public TagBuilder(string tagName)
         {
             if (string.IsNullOrEmpty(tagName))
@@ -25,15 +27,25 @@ namespace Microsoft.AspNet.Mvc.Rendering
             }
 
             TagName = tagName;
-            Attributes = new SortedDictionary<string, string>(StringComparer.OrdinalIgnoreCase);
-
             InnerHtml = new BufferedHtmlContent();
         }
 
         /// <summary>
         /// Gets the set of attributes that will be written to the tag.
         /// </summary>
-        public IDictionary<string, string> Attributes { get; }
+        public AttributeDictionary Attributes
+        {
+            get
+            {
+                // Perf: Avoid allocating `_attributes` if possible
+                if (_attributes == null)
+                {
+                    _attributes = new AttributeDictionary();
+                }
+
+                return _attributes;
+            }
+        }
 
         public IHtmlContentBuilder InnerHtml { get; }
 
@@ -151,20 +163,24 @@ namespace Microsoft.AspNet.Mvc.Rendering
 
         private void AppendAttributes(TextWriter writer, IHtmlEncoder encoder)
         {
-            foreach (var attribute in Attributes)
+            // Perf: Avoid allocating enumerator for `_attributes` if possible
+            if (_attributes != null && _attributes.Count > 0)
             {
-                var key = attribute.Key;
-                if (string.Equals(key, "id", StringComparison.OrdinalIgnoreCase) &&
-                    string.IsNullOrEmpty(attribute.Value))
+                foreach (var attribute in Attributes)
                 {
-                    continue;
-                }
+                    var key = attribute.Key;
+                    if (string.Equals(key, "id", StringComparison.OrdinalIgnoreCase) &&
+                        string.IsNullOrEmpty(attribute.Value))
+                    {
+                        continue;
+                    }
 
-                writer.Write(" ");
-                writer.Write(key);
-                writer.Write("=\"");
-                encoder.HtmlEncode(attribute.Value, writer);
-                writer.Write("\"");
+                    writer.Write(" ");
+                    writer.Write(key);
+                    writer.Write("=\"");
+                    encoder.HtmlEncode(attribute.Value, writer);
+                    writer.Write("\"");
+                }
             }
         }
 
@@ -193,7 +209,8 @@ namespace Microsoft.AspNet.Mvc.Rendering
 
         public void MergeAttributes<TKey, TValue>(IDictionary<TKey, TValue> attributes, bool replaceExisting)
         {
-            if (attributes != null)
+            // Perf: Avoid allocating enumerator for `attributes` if possible
+            if (attributes != null && attributes.Count > 0)
             {
                 foreach (var entry in attributes)
                 {
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/AttributeDictionary.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/AttributeDictionary.cs
@@ -1,0 +1,681 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNet.Mvc.ViewFeatures
+{
+    /// <summary>
+    /// A dictionary for HTML attributes.
+    /// </summary>
+    public class AttributeDictionary : IDictionary<string, string>, IReadOnlyDictionary<string, string>
+    {
+        private List<KeyValuePair<string, string>> _items;
+
+        /// <inheritdoc />
+        public string this[string key]
+        {
+            get
+            {
+                if (key == null)
+                {
+                    throw new ArgumentNullException(nameof(key));
+                }
+
+                var index = Find(key);
+                if (index < 0)
+                {
+                    throw new KeyNotFoundException();
+                }
+                else
+                {
+                    return Get(index).Value;
+                }
+            }
+
+            set
+            {
+                if (key == null)
+                {
+                    throw new ArgumentNullException(nameof(key));
+                }
+
+                var item = new KeyValuePair<string, string>(key, value);
+                var index = Find(key);
+                if (index < 0)
+                {
+                    Insert(~index, item);
+                }
+                else
+                {
+                    Set(index, item);
+                }
+            }
+        }
+
+        /// <inheritdoc />
+        public int Count => _items == null ? 0 : _items.Count;
+
+        /// <inheritdoc />
+        public bool IsReadOnly
+        {
+            get
+            {
+                return false;
+            }
+        }
+
+        /// <inheritdoc />
+        public ICollection<string> Keys
+        {
+            get
+            {
+                return new KeyCollection(this);
+            }
+        }
+
+        /// <inheritdoc />
+        public ICollection<string> Values
+        {
+            get
+            {
+                return new ValueCollection(this);
+            }
+        }
+
+        /// <inheritdoc />
+        IEnumerable<string> IReadOnlyDictionary<string, string>.Keys
+        {
+            get
+            {
+                return new KeyCollection(this);
+            }
+        }
+
+        /// <inheritdoc />
+        IEnumerable<string> IReadOnlyDictionary<string, string>.Values
+        {
+            get
+            {
+                return new ValueCollection(this);
+            }
+        }
+
+        private KeyValuePair<string, string> Get(int index)
+        {
+            if (index >= Count)
+            {
+                throw new ArgumentOutOfRangeException();
+            }
+
+            return _items[index];
+        }
+
+        private void Set(int index, KeyValuePair<string, string> value)
+        {
+            if (index > Count)
+            {
+                throw new ArgumentOutOfRangeException();
+            }
+
+            if (_items == null)
+            {
+                _items = new List<KeyValuePair<string, string>>();
+            }
+
+            _items[index] = value;
+        }
+
+        private void Insert(int index, KeyValuePair<string, string> value)
+        {
+            if (index > Count)
+            {
+                throw new ArgumentOutOfRangeException();
+            }
+
+            if (_items == null)
+            {
+                _items = new List<KeyValuePair<string, string>>();
+            }
+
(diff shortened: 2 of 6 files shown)


src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/TagBuilder.cs
```diff
@@ -25,15 +27,25 @@ public TagBuilder(string tagName)
...
-            Attributes = new SortedDictionary<string, string>(StringComparer.OrdinalIgnoreCase);
-
             InnerHtml = new BufferedHtmlContent();
         }
 
         /// <summary>
         /// Gets the set of attributes that will be written to the tag.
-        /// </summary>
-        public IDictionary<string, string> Attributes { get; }
+        /// </summary>B
```
dougbu: Have you hidden the `r` and the `o` somewhere else?

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/AttributeDictionary.cs
```diff
@@ -0,0 +1,653 @@
...
+                    end = pivot - 1;
+                }
+            }
+
+            return ~start;
+        }
+
+        public void Clear()
+        {
+            _items0 = default(KeyValuePair<string, string>);
```
dougbu: add `Count = 0;` to this method

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/AttributeDictionary.cs
```diff
@@ -0,0 +1,653 @@
...
+        {
+            if (index > Count)
+            {
+                throw new ArgumentOutOfRangeException();
+            }
+
+            if (_items == null)
+            {
+                _items = new List<KeyValuePair<string, string>>();
+            }
```
dougbu: Why is this block here? Case covered in `Count > FixedEntryCount` block.

dougbu: Could you please rerun the numbers after removing this? (Assuming it's the extra lists bug.)

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/AttributeDictionary.cs
```diff
@@ -0,0 +1,653 @@
...
+        {
+            return GetEnumerator();
+        }
+
+        IEnumerator<KeyValuePair<string, string>> IEnumerable<KeyValuePair<string, string>>.GetEnumerator()
+        {
+            return GetEnumerator();
+        }
+
+        public struct Enumerator : IEnumerator<KeyValuePair<string, string>>
```
Eilon: OM nitpick: Please no public nested types. So gross.

Eilon: Yeah I know it's a design PR but I can't help myself :smile:

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/TagBuilder.cs
```diff
@@ -25,15 +27,25 @@ public TagBuilder(string tagName)
...
+        public AttributeDictionary Attributes
```
Eilon: Do you intend for the return type to be this or keep it `IDictionary<string, string`?

Eilon: (I think it should stay IDictionary)

Eilon:
This looks pretty cool. It's certainly a fairly common trick. I think it's even been used in ASP.NET before.

Eilon:
BTW this is I think almost exactly what you implemented: http://referencesource.microsoft.com/#System/compmod/system/collections/specialized/hybriddictionary.cs,23 (read the doc comment at the top)

But unfortunately that type is non-generic so we can't use it directly. I wonder if we could use that as the backing store for this type instead of having lots of code?

It also reminds me of ASP.NET's OccasionalFields and RareFields (http://referencesource.microsoft.com/#System.Web/UI/Control.cs,81 and http://referencesource.microsoft.com/#System.Web/UI/Control.cs,3644)

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/TagBuilder.cs
davidfowl: Yar

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/AttributeDictionary.cs
rynowak: This is the standard pattern for Enumerators'n'such - should we open a bug on the BCL?

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/TagBuilder.cs
rynowak: :-1: No, the intent is that it should be a concrete type that returns a `struct` enumerator. It's a big problem with using dictionaries in general and there are three ways to squash it.
- don't ever use the enumerator (not likely)
- only use the enumerator from places where you know the concrete type
- do this

All of the arguments about why you should prefer `IDictionary<>` over `Dictionary<>` can also be made for defining your own type. You guys just hate custom collections :trollface:

rynowak:
@Eilon - this is actually a simplified and sorted version of `FrugalMap` - http://referencesource.microsoft.com/#WindowsBase/Shared/MS/Utility/FrugalMap.cs

If you look at the commits in this PR - the first one is a simple listdictionary - it doesn't go hybrid because we never never put more than like 10 things in there. If you decide to do it, then on your own head. This is a reasonable conservative path for balancing complexity vs benefit and also flexibility vs optimization for our usage.

The second commit uses fields of the object for the first 4 items and then grows to use a list if you need more space. We'd likely see a better result from going up to 6, but I want to use a more representative app to make that kind of call.

I'm leaning at this point towards moving forward with the first version, and then revisiting this if it becomes important for a bigger app.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/TagBuilder.cs
rynowak: If you guys are really opposed to the idea of exposing the concrete type then the right thing to do is to start over with this API altogether.

davidfowl: I'm +1 for the type. It make sense if this is the common use case.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/AttributeDictionary.cs
rynowak: Not really relevant anymore because I'm going with the simple list version for now.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/AttributeDictionary.cs
```diff
@@ -0,0 +1,704 @@
...
+                }
+                else
+                {
+                    Set(index, item);
+                }
+            }
+        }
+
+        /// <inheritdoc />
+        public int Count => _items == null ? 0 : _items.Count;
```
pranavkm: :100:

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/AttributeDictionary.cs
```diff
@@ -0,0 +1,704 @@
...
+        public string this[string key]
+        {
+            get
+            {
+                if (key == null)
+                {
+                    throw new ArgumentNullException(nameof(key));
+                }
+
+                if (Count == 0)
```
pranavkm: What's the current behavior when you have no entries and you attempt to look it up? I'd think it throws a `KeyNotFoundEx`

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/AttributeDictionary.cs
```diff
@@ -0,0 +1,704 @@
...
+            if (index >= Count)
+            {
+                throw new ArgumentOutOfRangeException();
+            }
+
+            
+            _items.RemoveAt(index);
+        }
+
+        private int Find(string key)
```
pranavkm: Could you add a comment that calls out the not found behavior. Pretty nifty.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/AttributeDictionary.cs
rynowak: Yeah good point, this should throw.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/AttributeDictionary.cs
rynowak: sure

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/AttributeDictionary.cs
```diff
@@ -0,0 +1,704 @@
...
+            }
+
+            var start = 0;
+            var end = Count - 1;
+
+            while (start <= end)
+            {
+                int pivot = start + (end - start >> 1);
+
+                int compare = StringComparer.OrdinalIgnoreCase.Compare(Get(pivot).Key, key);
```
pranavkm: var

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewFeatures/AttributeDictionaryTest.cs
```diff
@@ -0,0 +1,432 @@
...
+            // Act
+            attributes.Clear();
+
+            // Assert
+            Assert.Equal(0, attributes.Count);
+            Assert.Empty(attributes);
+        }
+
+        [Fact]
+        public void AttributeDictionary_Contains()
```
pranavkm: Can you add negative cases for the `Contains` \ `ContainsKeys` (the `Assert.False` scenarios?)

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewFeatures/AttributeDictionaryTest.cs
```diff
@@ -0,0 +1,432 @@
...
+            var attributes = new AttributeDictionary();
+
+            // Act
+            var result = attributes.IsReadOnly;
+
+            // Assert
+            Assert.False(result);
+        }
+
+
```
pranavkm: Empty line

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewFeatures/AttributeDictionaryTest.cs
```diff
@@ -0,0 +1,432 @@
...
+            attributes.Add(new KeyValuePair<string, string>("zero", "0"));
+            attributes.Add(new KeyValuePair<string, string>("one", "1"));
+            attributes.Add(new KeyValuePair<string, string>("two", "2"));
+
+            string value;
+
+            // Act
+            var result = attributes.TryGetValue("two", out value);
+
+
```
pranavkm: remove

pranavkm:
:shipit:

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewFeatures/AttributeDictionaryTest.cs
rynowak: sure

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/AttributeDictionary.cs
```diff
@@ -0,0 +1,681 @@
...
+        {
+            get
+            {
+                return new ValueCollection(this);
+            }
+        }
+
+        private KeyValuePair<string, string> Get(int index)
+        {
+            if (index >= Count)
```
dougbu: This should also `throw` if `index < 0`.

dougbu: Error case missed elsewhere in this class.

pranavkm: Wouldn't the backing list take care of this?

PR #3232, opened 2015-09-30
React to WriteAttribute \ AddHtmlAttribute API changes

pranavkm:
Reaction to https://github.com/aspnet/Razor/pull/554


pranavkm:
For a page with 4 `<a href="~/Home">Home</a>` links over 3000 requests. The number of allocations increases linearly with the number of conditional attributes the page has

### Before

tuple

attributevalue


### After
tupleafter

NTaylorMullen:
Waiting for reaction to Razor PR comments.

rynowak:
:clap: - what's the total allocations for your sample?

pranavkm:
@rynowak it's pretty insignificant in this application. Let me try out a more realistic page and get back to you on it.

rynowak:
I still wouldn't call 1mb insignificant - for the kinds of samples I've been investigating, it's approaching 1% of the total bytes. 

The questions I'm asking are:
- Is the fix yucky? (would we want this change without any perf improvement)
- If it is yucky, is the yuckiness balanced with the benefit?

src/Microsoft.AspNet.Mvc.Razor.Host/MvcRazorHost.cs
```diff
@@ -107,7 +107,6 @@ internal MvcRazorHost(IChunkTreeCache chunkTreeCache, RazorPathNormalizer pathNo
 
                     // Can't use nameof because IHtmlHelper is (also) not accessible here.
                     MarkAsHtmlEncodedMethodName = HtmlHelperPropertyName + ".Raw",
-                    AddHtmlAttributeValuesMethodName = "AddHtmlAttributeValues",
```
NTaylorMullen: We are always explicit about the configuration of the methods. We should add the new values.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -1030,15 +1052,12 @@ public virtual HtmlString SetAntiforgeryCookieAndHeader()
...
-            if (values.Length == 1 && string.IsNullOrEmpty(values[0].Prefix) &&
-                (values[0].Value.Value is bool || values[0].Value.Value == null))
+            if (string.IsNullOrEmpty(prefix) &&
+                (value is bool || value == null))
             {
-                var attributeValue = values[0];
-                var positionTaggedAttributeValue = attributeValue.Value;
-
-                if (positionTaggedAttributeValue.Value == null || !(bool)positionTaggedAttributeValue.Value)
+                if (value == null || !(bool)value)
```
NTaylorMullen: Can pull this up into the above statement.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -1047,13 +1066,11 @@ private bool IsSingleBoolFalseOrNullValue(AttributeValue[] values)
             return false;
         }
 
-        private bool UseAttributeNameAsValue(AttributeValue[] values)
+        private bool UseAttributeNameAsValue(string prefix, object value)
```
NTaylorMullen: Should change name to reflect new behavior since we're not consuming all of the attributes anymore (how many there are controls if we use the attribute name as value).

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -1063,5 +1080,59 @@ private void EnsureMethodCanBeInvoked(string methodName)
                 throw new InvalidOperationException(Resources.FormatRazorPage_MethodCannotBeCalled(methodName, Path));
             }
         }
+
+        private struct AttributeInfo
```
NTaylorMullen: Seems wasteful to have this attribute info reflect TagHelper and non-TagHelper attributes. The TagHelper variation would be much simpler if they weren't combined.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -1063,5 +1080,59 @@ private void EnsureMethodCanBeInvoked(string methodName)
...
+
+            public int PrefixOffset { get; }
+
+            public string Suffix { get; }
+
+            public int SuffixOffset { get; }
+
+            public TagHelperExecutionContext ExecutionContext { get; }
+
+            public bool Processed { get; set; }
```
NTaylorMullen: `Suppressed`

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
pranavkm: Sure, I could do that.

NTaylorMullen:
:watch:

pranavkm:
@rynowak, yeah the API surface at the consumption end is much nicer and requires fewer data types to grock. We'd want it regardless of the API change. 

That said, for the BasicVIew application (with the addition 4 anchor links to the Index view) from https://github.com/rynowak/mvc-sandbox, we're looking at 0.68% reduced allocations.

pranavkm:
Updated

NTaylorMullen:
:shipit:

PR #3238, opened 2015-09-30
Minor cleanup/refactor of ViewResult

rynowak:
- Make ViewExecutor a service
- Add facades for ViewResult/PartialViewResult
- Add eventing for ViewFound/NotFound in PartialViewResult
- Add eventing around view execution
- Cleanup of some various eventing & our tests code

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs
@@ -636,7 +636,7 @@ namespace Microsoft.AspNet.Mvc.Controllers
                         {
                             _telemetry.WriteTelemetry(
                                 "Microsoft.AspNet.Mvc.AfterActionMethod",
-                                new { actionContext = ActionContext, result });
+                                new { actionContext = ActionContext, result = result });
                         }
                     }
 
@@ -789,7 +789,7 @@ namespace Microsoft.AspNet.Mvc.Controllers
             {
                 _telemetry.WriteTelemetry(
                     "Microsoft.AspNet.Mvc.BeforeActionResult",
-                    new { actionContext = ActionContext, result });
+                    new { actionContext = ActionContext, result = result });
             }
 
             try
@@ -802,7 +802,7 @@ namespace Microsoft.AspNet.Mvc.Controllers
                 {
                     _telemetry.WriteTelemetry(
                         "Microsoft.AspNet.Mvc.AfterActionResult",
-                        new { actionContext = ActionContext, result });
+                        new { actionContext = ActionContext, result = result });
                 }
             }
         }
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/PartialViewResult.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/PartialViewResult.cs
@@ -3,12 +3,10 @@
 
 using System;
 using System.Threading.Tasks;
-using Microsoft.Framework.DependencyInjection;
-using Microsoft.Framework.Logging;
-using Microsoft.Net.Http.Headers;
-using Microsoft.Framework.OptionsModel;
 using Microsoft.AspNet.Mvc.ViewEngines;
 using Microsoft.AspNet.Mvc.ViewFeatures;
+using Microsoft.Framework.DependencyInjection;
+using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNet.Mvc
 {
@@ -60,41 +58,16 @@ namespace Microsoft.AspNet.Mvc
                 throw new ArgumentNullException(nameof(context));
             }
 
-            var viewEngine = ViewEngine ??
-                             context.HttpContext.RequestServices.GetRequiredService<ICompositeViewEngine>();
+            var services = context.HttpContext.RequestServices;
+            var executor = services.GetRequiredService<PartialViewResultExecutor>();
 
-            var logger = context.HttpContext.RequestServices.GetRequiredService<ILogger<PartialViewResult>>();
+            var result = executor.FindView(context, this);
+            result.EnsureSuccessful();
 
-            var options = context.HttpContext.RequestServices.GetRequiredService<IOptions<MvcViewOptions>>();
-
-            var viewName = ViewName ?? context.ActionDescriptor.Name;
-            var viewEngineResult = viewEngine.FindPartialView(context, viewName);
-            if (!viewEngineResult.Success)
-            {
-                logger.LogError(
-                    "The partial view '{PartialViewName}' was not found. Searched locations: {SearchedViewLocations}",
-                    viewName,
-                    viewEngineResult.SearchedLocations);
-            }
-
-            var view = viewEngineResult.EnsureSuccessful().View;
-
-            logger.LogVerbose("The partial view '{PartialViewName}' was found.", viewName);
-
-            if (StatusCode != null)
-            {
-                context.HttpContext.Response.StatusCode = StatusCode.Value;
-            }
-
+            var view = result.View;
             using (view as IDisposable)
             {
-                await ViewExecutor.ExecuteAsync(
-                    view,
-                    context,
-                    ViewData,
-                    TempData,
-                    options.Value.HtmlHelperOptions,
-                    ContentType);
+                await executor.ExecuteAsync(context, view, this);
             }
         }
     }
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/PartialViewResultExecutor.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/PartialViewResultExecutor.cs
@@ -1,0 +1,143 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics.Tracing;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Mvc.ViewEngines;
+using Microsoft.Framework.Logging;
+using Microsoft.Framework.OptionsModel;
+
+namespace Microsoft.AspNet.Mvc.ViewFeatures
+{
+    /// <summary>
+    /// Finds and executes an <see cref="IView"/> for a <see cref="PartialViewResult"/>.
+    /// </summary>
+    public class PartialViewResultExecutor : ViewExecutor
+    {
+        /// <summary>
+        /// Creates a new <see cref="PartialViewResultExecutor"/>.
+        /// </summary>
+        /// <param name="viewOptions">The <see cref="IOptions{MvcViewOptions}"/>.</param>
+        /// <param name="viewEngine">The <see cref="ICompositeViewEngine"/>.</param>
+        /// <param name="telemetry">The <see cref="TelemetrySource"/>.</param>
+        /// <param name="loggerFactory">The <see cref="ILoggerFactory"/>.</param>
+        public PartialViewResultExecutor(
+            IOptions<MvcViewOptions> viewOptions,
+            ICompositeViewEngine viewEngine,
+            TelemetrySource telemetry,
+            ILoggerFactory loggerFactory)
+            : base(viewOptions, viewEngine, telemetry)
+        {
+            if (loggerFactory == null)
+            {
+                throw new ArgumentNullException(nameof(loggerFactory));
+            }
+
+            Logger = loggerFactory.CreateLogger<PartialViewResultExecutor>();
+        }
+
+        /// <summary>
+        /// Gets the <see cref="ILogger"/>.
+        /// </summary>
+        protected ILogger Logger { get; }
+
+        /// <summary>
+        /// Attempts to find the <see cref="IView"/> associated with <paramref name="viewResult"/>.
+        /// </summary>
+        /// <param name="actionContext">The <see cref="ActionContext"/> associated with the current request.</param>
+        /// <param name="viewResult">The <see cref="PartialViewResult"/>.</param>
+        /// <returns>A <see cref="ViewEngineResult"/>.</returns>
+        public virtual ViewEngineResult FindView(ActionContext actionContext, PartialViewResult viewResult)
+        {
+            if (actionContext == null)
+            {
+                throw new ArgumentNullException(nameof(actionContext));
+            }
+
+            if (viewResult == null)
+            {
+                throw new ArgumentNullException(nameof(viewResult));
+            }
+
+            var viewEngine = viewResult.ViewEngine ?? ViewEngine;
+            var viewName = viewResult.ViewName ?? actionContext.ActionDescriptor.Name;
+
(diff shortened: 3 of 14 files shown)


rynowak:
What's **not done** here is all of the updates to unit tests

src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs
```diff
@@ -587,7 +587,7 @@ private Task InvokeAllExceptionFiltersAsync()
                         {
                             _telemetry.WriteTelemetry(
                                 "Microsoft.AspNet.Mvc.AfterActionMethod",
-                                new { actionContext = ActionContext, result });
+                                new { actionContext = ActionContext, result = result });
```
rynowak: using `new { result }` is legal c#, but not a practice we should use with event notifications. The parameter names are significant, and we don't want to break consumers with a refactor

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/PartialViewResultExecutor.cs
```diff
@@ -0,0 +1,141 @@
...
+                        isPartial = false,
+                        result = viewResult,
+                        viewName,
+                        view = result.View,
+                    });
+            }
+
+            Logger.LogVerbose("The partial view '{PartialViewName}' was found.", viewName);
+
+            return result;
```
rynowak: This code *looks* identical to `ViewResultExecutor`, but it's actually just different enough IMO to deserve its own method.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/PartialViewResultExecutor.cs
```diff
@@ -0,0 +1,141 @@
...
+            }
+            
+            if (Telemetry.IsEnabled("Microsoft.AspNet.Mvc.ViewFound"))
+            {
+                Telemetry.WriteTelemetry(
+                    "Microsoft.AspNet.Mvc.ViewFound",
+                    new
+                    {
+                        actionContext = actionContext,
+                        isPartial = false,
```
rynowak: This should be `true`

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/PartialViewResultExecutor.cs
```diff
@@ -0,0 +1,141 @@
...
+            if (!result.Success)
+            {
+                if (Telemetry.IsEnabled("Microsoft.AspNet.Mvc.ViewNotFound"))
+                {
+                    Telemetry.WriteTelemetry(
+                        "Microsoft.AspNet.Mvc.ViewNotFound",
+                        new
+                        {
+                            actionContext = actionContext,
+                            isPartial = false,
```
rynowak: This should be `true`

src/Microsoft.AspNet.Mvc.ViewFeatures/PartialViewResult.cs
```diff
@@ -60,41 +58,16 @@ public class PartialViewResult : ActionResult
                 throw new ArgumentNullException(nameof(context));
             }
 
-            var viewEngine = ViewEngine ??
-                             context.HttpContext.RequestServices.GetRequiredService<ICompositeViewEngine>();
+            var services = context.HttpContext.RequestServices;
+            var executor = services.GetRequiredService<PartialViewResultExecutor>();
 
-            var logger = context.HttpContext.RequestServices.GetRequiredService<ILogger<PartialViewResult>>();
```
rynowak: Wave :wave: bye bye to all of these `GetRequiredService` calls

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/ViewExecutor.cs
```diff
@@ -84,12 +122,27 @@ public static class ViewExecutor
...
-                // response. In the absence of this line, the buffer gets synchronously written to the response
-                // as part of the Dispose which has a perf impact.
+
+                if (Telemetry.IsEnabled("Microsoft.AspNet.Mvc.AfterView"))
+                {
+                    Telemetry.WriteTelemetry(
+                        "Microsoft.AspNet.Mvc.AfterView",
+                        new { view = view, viewContext = viewContext, });
+                }
+
```
rynowak: @avanderhoorn is this what you're looking for?

src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs
avanderhoorn: Good thinking!

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/PartialViewResultExecutor.cs
```diff
@@ -0,0 +1,141 @@
...
+            if (Telemetry.IsEnabled("Microsoft.AspNet.Mvc.ViewFound"))
+            {
+                Telemetry.WriteTelemetry(
+                    "Microsoft.AspNet.Mvc.ViewFound",
+                    new
+                    {
+                        actionContext = actionContext,
+                        isPartial = false,
+                        result = viewResult,
+                        viewName,
```
avanderhoorn: Based on what you said about could this be `viewName = viewName`?

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/ViewExecutor.cs
avanderhoorn: Yes, Thanks!

avanderhoorn: Is there any way of determining of the `BeforeView`/`AfterView` is a partial or not?

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/PartialViewResultExecutor.cs
Eilon: Yes that needs to be fixed here, good catch.

Eilon: BTW a few other of these still elsewhere in this PR as well.

Eilon:
Looks pretty nice.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/PartialViewResultExecutor.cs
rynowak: Yes, missed this one too!

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/PartialViewResultExecutor.cs
```diff
@@ -0,0 +1,141 @@
...
+        /// </summary>
+        protected ILogger Logger { get; }
+
+        /// <summary>
+        /// Attempts to find the <see cref="IView"/> associated with <paramref name="viewResult"/>.
+        /// </summary>
+        /// <param name="actionContext">The <see cref="ActionContext"/> associated with the current request.</param>
+        /// <param name="viewResult">The <see cref="PartialViewResult"/>.</param>
+        /// <returns>A <see cref="ViewEngineResult"/>.</returns>
+        public virtual ViewEngineResult FindView(ActionContext actionContext, PartialViewResult viewResult)
```
pranavkm: Couldnt `FindView` + `ExecuteViewAsync` be combined?

pranavkm:
Design looks good

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/PartialViewResultExecutor.cs
rynowak: All of ASP.NET could be one method with callbacks you override. Personally, I like to see the 5-6 lines of code that end up in the result class, and having it all in one method would make this less flexible. What do you think would be better if they were combined?

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/ViewExecutor.cs
rynowak: We can also include the actionresult here if you want.

rynowak: Not easily. Why is this important?

avanderhoorn: Just how many messages I need to processes to determine if the timing I have (between BeforeView and AfterView) was for a partial view or not - we show them differently in the ui.

avanderhoorn: > We can also include the actionresult here if you want.

It would be nice... getting access to the `ActionId` helps with correlation.

rynowak: > We can also include the actionresult here if you want.

I deleted this comment because it's not that simple. If the actionID is what you want, you have that the `ViewContext` here already.

> we show them differently in the ui.

This is for top-level results **only** not for `@Html.Partial` - `RazorView` actually already knows if it's a partial or not.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/PartialViewResultExecutor.cs
rynowak: There's also IMO a win for testability with this design.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/ViewExecutor.cs
avanderhoorn: Great on the `ViewContext` front. If it has `actionId` then that should be enough. Regarding top-level results, are you saying that `BeforeView`/`AfterView` only fires for the top-level results?

rynowak: > Regarding top-level results, are you saying that BeforeView/AfterView only fires for the top-level results?

Correct.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/PartialViewResultExecutor.cs
pranavkm: Fair enough if it helps testing.

rynowak:
@pranavkm updated - removed design tag

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/PartialViewResultExecutor.cs
```diff
@@ -0,0 +1,145 @@
...
+
+namespace Microsoft.AspNet.Mvc.ViewFeatures
+{
+    /// <summary>
+    /// Finds and executes an <see cref="IView"/> for a <see cref="PartialViewResult"/>.
+    /// </summary>
+    public class PartialViewResultExecutor : ViewExecutor
+    {
+        /// <summary>
+        /// Creates a new <see cref="ViewResultExecutor"/>.
```
pranavkm: > PartialViewResultExecutor

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/PartialViewResultExecutor.cs
```diff
@@ -0,0 +1,145 @@
...
+                        });
+                }
+
+                Logger.LogError(
+                    "The partial view '{PartialViewName}' was not found. Searched locations: {SearchedViewLocations}",
+                    viewName,
+                    result.SearchedLocations);
+            }
+            
+
```
pranavkm: Extra lines.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/PartialViewResultExecutor.cs
```diff
@@ -0,0 +1,145 @@
...
+            {
+                throw new ArgumentNullException(nameof(view));
+            }
+
+            if (viewResult == null)
+            {
+                throw new ArgumentNullException(nameof(viewResult));
+            }
+
+            await ExecuteAsync(
```
pranavkm: Shouldn't need the await here.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/ViewExecutor.cs
```diff
@@ -13,54 +15,84 @@ namespace Microsoft.AspNet.Mvc.ViewFeatures
     /// <summary>
     /// Utility type for rendering a <see cref="IView"/> to the response.
```
pranavkm: > Executes an `<see cref="IView" />` to the response.

To be in sync with derived types.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/ViewExecutor.cs
```diff
@@ -13,54 +15,84 @@ namespace Microsoft.AspNet.Mvc.ViewFeatures
...
+        /// <param name="tempData">The <see cref="ITempDataDictionary"/>.</param>
+        /// <param name="contentType">
+        /// The content-type header value to set in the response. If <c>null</c>, <see cref="DefaultContentType"/> will be used.
+        /// </param>
+        /// <param name="statusCode">
+        /// The HTTP status code to set in the response. May be <c>null</c>.
+        /// </param>
+        /// <returns>A <see cref="Task"/> which will complete when view execution is completed.</returns>
+        public virtual async Task ExecuteAsync(
+            ActionContext actionContext,
```
pranavkm: These would need null checks.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/ViewResultExecutor.cs
```diff
@@ -0,0 +1,143 @@
...
+            {
+                throw new ArgumentNullException(nameof(view));
+            }
+
+            if (viewResult == null)
+            {
+                throw new ArgumentNullException(nameof(viewResult));
+            }
+
+            await ExecuteAsync(
```
pranavkm: `return ExecuteAsync`

pranavkm:
:shipit:

PR #3241, opened 2015-10-01
Renamed PhysicalFileProviderResult to PhysicalFileResult

kichalla:
@pranavkm


pranavkm:
Do we need to change `VirtualFileProviderResult`?

kichalla:
good point...i updated the PR to include this name change too...

pranavkm:
:shipit:

Eilon:
:shipit: but FYI your branch looks messed up (it has your session state changes in it).

kichalla:
eef6c3883a7e27b8387b0925f0b6a88df0a484c5

Shazwazza:
would be great to put this on the announcements board as a breaking change

Eilon:
@Shazwazza the result types aren't used very frequently because most people use the factory methods on `Controller`. We want to avoid "too many" announcements so that most people will continue to find it useful.

Shazwazza:
Fair enough, I found this so no worries ;)

PR #3248, opened 2015-10-02
Use pooled memory for the streamwriter

rynowak:
/cc @pranavkm @kichalla

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
@@ -5,6 +5,7 @@ using System;
 using System.IO;
 using System.Text;
 using System.Threading.Tasks;
+using Microsoft.Framework.MemoryPool;
 
 namespace Microsoft.AspNet.Mvc
 {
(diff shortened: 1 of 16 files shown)


src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
```diff
@@ -168,6 +216,16 @@ public override Task FlushAsync()
         protected override void Dispose(bool disposing)
         {
             FlushInternal(flushStream: false, flushEncoder: true);
+
+            if (_leasedByteBuffer != null)
```
pranavkm: `_leasedByteBuffer?.Owner.Return(_leasedByteBuffer);`

pranavkm:
:shipit:

src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
rynowak: Wouldn't this result in a nullref?
```
var temp = _leasedByteBuffer?.Owner;
temp.Return(_leasedByteBuffer);
```

rynowak:
Some perf numbers for this:

**Before**
image


**After**
image

Allocations for 3001 requests to a site with model binding and a very minimal view.

Issue #3252, opened 2015-10-03
Exception while encoding/writing view using pooled buffer


gvx111:
on running controller /home/index showing this error
<pre>ArgumentException: The output byte buffer is too small to contain the encoded data, encoding 'Unicode (UTF-8)' fallback 'System.Text.EncoderReplacementFallback'.
Parameter name: bytes
System.Text.Encoding.ThrowBytesOverflow()</pre>

Error 2
<pre>Unhandled Exception: System.InvalidOperationException: A LeasedArraySegment was
collected without being returned to the pool. This is a memory leak.
   at Microsoft.Framework.MemoryPool.DefaultArraySegmentPool`1.DefaultLeasedArra
ySegment.Finalize()</pre>

rynowak:
Can you provide some more details? What platform is this on? What's your controller/view code look like?

gvx111:
this is my view.cshtml
<pre>@{
    ViewBag.Title = "Auctions";
    Layout = Url.ModuleArea("Views\\Layouts\\_Layout");
}
@using ITN.TS.UserStore.Extensions
@using ITN.TS.UserStore.Services
@model List<Auction>

<section id="content" class="content_index">
    <div class="title-line">
        <div class="title">მიმდინარე აუქციონები</div>
    </div>
    <div class="product-content">
        @foreach (var item in Model)
        {
            await Html.RenderPartialAsync("_AuctionBox", item);
        }
    </div>
</section>
<div class="modal fade" id="myModal" tabindex="-1" role="dialog" aria-labelledby="myModalLabel" data-backdrop="static" data-keyboard="false">
    <div class="modal-dialog" role="document">
        <div class="modal-content">
            <div class="modal-body">
                <i class="fa fa-minus-circle"></i> Attempting to restore connection. Changes made now may not be saved.
            </div>
        </div>
    </div>
</div>
@section scripts
{
    <script src="http://ajax.aspnetcdn.com/ajax/signalr/jquery.signalr-2.0.3.js"></script>
    <script src="~/signalr/hubs" type="text/javascript"></script>
    <script type="text/javascript">
        jQuery(document).ready(function($) {
            $('.bxslider').bxSlider();
            var auctionHub = $.connection.Auction;
            $.connection.hub.reconnecting(function() {
                $('#myModal').modal();
            });

            $.connection.hub.reconnected(function() {
                $('#myModal').modal('hide');
            });

            $.connection.hub.disconnected(function() {

            });
[...]

PR #3259, opened 2015-10-05
Fix for #3252 - Issues with pooled buffer + unicode

rynowak:
- Dispose buffers when Flush throws inside the Dispose method
- Compute size of buffers correctly
- Throw earlier when handed invalid-sized buffers


src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
```diff
@@ -58,6 +59,7 @@ public HttpResponseStreamWriter(Stream stream, Encoding encoding, int bufferSize
         public HttpResponseStreamWriter(
```
pranavkm: Docs for the ctor

dougbu:
looks good :shipit:

pranavkm:
:shipit:

PR #3260, opened 2015-10-05
Bring arguments inline between before/after events - Fixes #3204

### LINKED ISSUE #3204: Telemetry Source Before/After events feedback/thoughts
Currently there are a few places where we have various Before/After events - `BeforeAction`/`AfterAction` and `BeforeActionMethod`/`AfterActionMethod` being amongst them. In both of these cases the `Before` events has more details than the `After` event. This means that for people interested in getting that before data in conjunction with the timing data have to subscribe to both events and either independently store the data from the begin event and carry it though to the after, or correlate the data after the fact. 

In most cases, if the `After` event had the same data as the `Before` event, it would simplify things for the developer with the only thing they would need to store/flow between the two events is the timing information. In our case, it would also mean that we would only need to publish the `After` event to our internal store and/or not have to store this subset of data independently whilst waiting for the post event if we only wanted the one event going through to our internal store. 

For the current Before/After events where there is a difference in parameters that are published, it would seem that the extra data the `Before` event has could easily be made available to the `After` event as well. And moving forward it would seem this pattern would continue as the `After` usual is executed in the same scope as the `Before`.

avanderhoorn:
This is designed to address the few points mentioned in #3204 @rynowak

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs
@@ -625,7 +625,12 @@ namespace Microsoft.AspNet.Mvc.Controllers
                         {
                             _telemetry.WriteTelemetry(
                                 "Microsoft.AspNet.Mvc.BeforeActionMethod",
-                                new { actionContext = ActionContext, arguments = _actionExecutingContext.ActionArguments });
+                                new
+                                {
+                                    actionContext = ActionContext, 
+                                    arguments = _actionExecutingContext.ActionArguments,
+                                    controller = _actionExecutingContext.Controller
+                                });
                         }
 
                         result = await InvokeActionAsync(_actionExecutingContext);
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs
```diff
@@ -636,7 +636,7 @@ private Task InvokeAllExceptionFiltersAsync()
                         {
                             _telemetry.WriteTelemetry(
                                 "Microsoft.AspNet.Mvc.AfterActionMethod",
-                                new { actionContext = ActionContext, result = result });
+                                new { actionContext = ActionContext, arguments = _actionExecutingContext.ActionArguments, result = result });
```
rynowak: break up the long line :+1:

rynowak: Do you want access to the controller instance here? You can also add that to both of these if that appeals to you.

rynowak:
:shipit:

rynowak:
a0764faa86e00c77189fb563039a9abbee08ca66

PR #3261, opened 2015-10-05
CR feedback from attribute dictionary

rynowak:
@dougbu - cleaning up bounds checking, and turning these into asserts since it shouldn't actually ever happen.


dougbu:
:shipit:

PR #3266, opened 2015-10-06
Allow ViewFeature tests to run in dnxcore50

pranavkm:


test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewComponents/DefaultViewComponentDescriptorProviderTest.cs
```diff
@@ -81,7 +81,10 @@ protected override IEnumerable<TypeInfo> GetCandidateTypes()
                     GetAssemblyProvider()
                     .CandidateAssemblies
                     .SelectMany(a => a.DefinedTypes)
-                    .Select(t => t.GetTypeInfo());
+#if DNX451
+                    .Select(t => t.GetTypeInfo())
+#endif
```
dougbu: Why not do this unconditionally? Where necessary could add
```
#if DNXCORE50
using System.Reflection;
#endif
```

dougbu:
Looks good. What are the test counts when run with `dnxcore50` and `dnx451`?

pranavkm:
DNX451 1732
DNXCore50: 238

There's probably a lot more that can be made to run in Dnx451 but I just wanted to get some test coverage for the view component helper test for now.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewComponents/DefaultViewComponentDescriptorProviderTest.cs
pranavkm: `DefinedTypes` returns `TypeInfo` in DNXCore50 and `Type` in DNX451. It's kinda weird they broke this API.

rynowak:
Is this checked in now?

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewComponents/DefaultViewComponentDescriptorProviderTest.cs
dougbu: ick but :ok:

dougbu:
:shipit:

PR #3267, opened 2015-10-06
Use Type.GetMethod in ViewComponentMethodSelector

### LINKED ISSUE #3211: Can not pass nullable to ViewComponent Invoke method
Beta7 Calling a method like:

```c#
 public async Task<IViewComponentResult> InvokeAsync(int id, bool? commentAllowed, int? commentCount)
```

Generates error 

```
InvalidOperationException: Could not find an 'Invoke' or 'InvokeAsync' method matching the parameters.

Microsoft.AspNet.Mvc.ViewComponents.DefaultViewComponentInvoker.<InvokeAsync>d__4.MoveNext()
```

From View as

```
@await Component.InvokeAsync("Comment", Model.Id, Model.CommentAllowed, Model.CommentCount) 
```

pranavkm:
Fixes #3211

### DIFF
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/Resources.resx
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/Resources.resx
@@ -277,4 +277,7 @@
   <data name="Dictionary_DuplicateKey" xml:space="preserve">
     <value>The collection already contains an entry with key '{0}'.</value>
   </data>
+  <data name="ViewComponent_SyncMethod_ShouldNotReturnTask" xml:space="preserve">
+    <value>The view component method '{0}' should not return a Task.</value>
+  </data>
 </root>
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/Resources.resx
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/Resources.resx
@@ -277,7 +277,7 @@
   <data name="Dictionary_DuplicateKey" xml:space="preserve">
     <value>The collection already contains an entry with key '{0}'.</value>
   </data>
-  <data name="ViewComponent_SyncMethod_ShouldNotReturnTask" xml:space="preserve">
-    <value>The view component method '{0}' should not return a Task.</value>
+  <data name="ViewComponent_SyncMethod_CannotReturnTask" xml:space="preserve">
+    <value>The view component method '{0}' cannot return a {1}.</value>
   </data>
 </root>
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/ViewComponentMethodSelector.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/ViewComponentMethodSelector.cs
@@ -55,42 +55,44 @@ namespace Microsoft.AspNet.Mvc.ViewComponents
                 throw new InvalidOperationException(
                     Resources.FormatViewComponent_SyncMethod_ShouldReturnValue(SyncMethodName));
             }
+            else if (method.ReturnType.IsAssignableFrom(typeof(Task)))
+            {
+                throw new InvalidOperationException(
+                    Resources.FormatViewComponent_SyncMethod_ShouldNotReturnTask(SyncMethodName));
+            }
 
             return method;
         }
 
         private static MethodInfo GetMethod(TypeInfo componentType, object[] args, string methodName)
         {
-            args = args ?? new object[0];
-            var argumentExpressions = new Expression[args.Length];
-            for (var i = 0; i < args.Length; i++)
+            Type[] types;
+            if (args == null || args.Length == 0)
             {
-                argumentExpressions[i] = Expression.Constant(args[i], args[i].GetType());
+                types = Type.EmptyTypes;
+            }
+            else
+            {
+                types = new Type[args.Length];
+                for (var i = 0; i < args.Length; i++)
+                {
+                    types[i] = args[i]?.GetType() ?? typeof(object);
+                }
             }
 
-            try
-            {
-                // We're currently using this technique to make a call into a component method that looks like a
-                // regular method call.
-                //
-                // Ex: @Component.Invoke<Cart>("hello", 5) => cart.Invoke("hello", 5)
-                //
-                // This approach has some drawbacks, namely it doesn't account for default parameters, and more
-                // noticably, it throws if the method is not found.
-                //
-                // Unfortunely the overload of Type.GetMethod that we would like to use is not present in CoreCLR.
-                // Item #160 in Jira tracks these issues.
-                var expression = Expression.Call(
-                    Expression.Constant(null, componentType.AsType()),
-                    methodName,
-                    null,
-                    argumentExpressions);
-                return expression.Method;
-            }
-            catch (InvalidOperationException)
-            {
-                return null;
-            }
+#if DNX451
+            return componentType.AsType().GetMethod(
+                methodName,
+                BindingFlags.Public | BindingFlags.Instance,
+                binder: null,
+                types: types,
+                modifiers: null);
+#else
+            var method = componentType.AsType().GetMethod(methodName, types: types);
+            // At most one method (including static and instance methods) with the same parameter types can exist
+            // per type.
+            return method != null && method.IsStatic ? null : method;
+#endif
         }
     }
 }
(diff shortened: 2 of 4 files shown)


src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/ViewComponentMethodSelector.cs
```diff
@@ -55,42 +55,44 @@ public static MethodInfo FindSyncMethod(TypeInfo componentType, object[] args)
...
+
+#if DNX451
+            return componentType.AsType().GetMethod(
+                methodName,
+                BindingFlags.Public | BindingFlags.Instance,
+                binder: null,
+                types: types,
+                modifiers: null);
+#else
+            var method = componentType.AsType().GetMethod(methodName, types: types);
```
pranavkm: This is still deficient but it is slightly more robust than what we do. It filters non-public methods. The scenario that affects this is ifdefed in the test but involves having a non-static method in the base class and a `new` static method in the derived class with the same arguments. 

I'll follow up with the CoreCLR team to see if they can add a `Type.GetType(string methodName, BindingFlags, Type[] types)` overload which should address this issue.

dougbu: Why not use a `Linq` expression to check what you need in `MethodInfo`? e.g.
``` c#
componentType.AsType().GetMethods()
    .Where(methodInfo =>
    {
        if (!string.Equals(methodInfo.Name, methodName, StringComparison.Ordinal) ||
            !methodInfo.IsPublic ||
            methodInfo.IsStatic)
        {
            return false;
        }

        var parameters = methodInfo .GetParameters();
        var parameterTypes = parameters.Select(parameter => parameter.ParameterType);
        return parameterTypes.SequenceEqual(types);
    }
```

pranavkm: It doesn't do the same method resolution you get with C# overload resolution. That's what this selector is attempting to ensure.

src/Microsoft.AspNet.Mvc.ViewFeatures/Resources.resx
```diff
@@ -277,4 +277,7 @@
   <data name="Dictionary_DuplicateKey" xml:space="preserve">
     <value>The collection already contains an entry with key '{0}'.</value>
   </data>
+  <data name="ViewComponent_SyncMethod_ShouldNotReturnTask" xml:space="preserve">
+    <value>The view component method '{0}' should not return a Task.</value>
```
Eilon: Should not -> could not?

Eilon: And pass in `nameof(Task)`?

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/ViewComponentMethodSelector.cs
```diff
@@ -55,42 +55,44 @@ public static MethodInfo FindSyncMethod(TypeInfo componentType, object[] args)
                 throw new InvalidOperationException(
                     Resources.FormatViewComponent_SyncMethod_ShouldReturnValue(SyncMethodName));
             }
+            else if (method.ReturnType.IsAssignableFrom(typeof(Task)))
+            {
+                throw new InvalidOperationException(
+                    Resources.FormatViewComponent_SyncMethod_ShouldNotReturnTask(SyncMethodName));
+            }
```
rynowak: Good catch

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewComponents/ViewComponentMethodSelectorTest.cs
```diff
@@ -0,0 +1,217 @@
...
+            var ex = Assert.Throws<InvalidOperationException>(
+                () => ViewComponentMethodSelector.FindAsyncMethod(typeInfo, args));
+            Assert.Equal("The async view component method 'InvokeAsync' should be declared to return Task<T>.",
+                ex.Message);
+        }
+
+        [Theory]
+        [InlineData(new object[] { 4 }, "The view component method 'Invoke' should be declared to return a value.")]
+        [InlineData(new object[] { "" }, "The view component method 'Invoke' should not return a Task.")]
+        public void FindSyncMethod_ThrowsIfInvokeSyncDoesNotHaveCorrectReturnType(object[] args, string expectedMessage)
```
rynowak: why is this a theory test?

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/ViewComponentMethodSelector.cs
rynowak: > This is still deficient but it is slightly more robust than what we do.

I'm not convinced this is really a lateral move - did you test cases with implicit value-conversions? Conversion to a base-class? Conversion to an interface?

Basically I expect that all of **this stuff** is broken: https://msdn.microsoft.com/en-us/library/system.type.defaultbinder(v=vs.110).aspx#Anchor_2

I'm not suggesting a fix to any of these, I'm suggesting that this fix regresses some things we already have.

pranavkm: The conversions do seem to work - added additional test for those.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewComponents/ViewComponentMethodSelectorTest.cs
pranavkm: Probably shouldn't.

src/Microsoft.AspNet.Mvc.ViewFeatures/Resources.resx
pranavkm: Ok.

pranavkm:
Updated

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/ViewComponentMethodSelector.cs
```diff
@@ -55,42 +55,44 @@ public static MethodInfo FindSyncMethod(TypeInfo componentType, object[] args)
                 throw new InvalidOperationException(
                     Resources.FormatViewComponent_SyncMethod_ShouldReturnValue(SyncMethodName));
             }
+            else if (method.ReturnType.IsAssignableFrom(typeof(Task)))
+            {
+                throw new InvalidOperationException(
+                    Resources.FormatViewComponent_SyncMethod_CannotReturnTask(SyncMethodName, nameof(Task)));
```
dougbu: We tend to use `typeof(T).FullName` elsewhere.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewComponents/ViewComponentMethodSelectorTest.cs
```diff
@@ -0,0 +1,284 @@
...
+
+namespace Microsoft.AspNet.Mvc.ViewComponents
+{
+    public class ViewComponentMethodSelectorTest
+    {
+        [Theory]
+        [InlineData(typeof(ViewComponentWithSyncInvoke), new object[0])]
+        [InlineData(typeof(ViewComponentWithSyncInvoke), new object[] { "" })]
+        [InlineData(typeof(ViewComponentWithSyncInvoke), new object[] { 42 })]
+        [InlineData(typeof(ViewComponentWithAsyncInvoke), new object[] { 42, false })]
```
dougbu: Isn't this the only data that leads to failure due to type mismatches? Put another way, looks like all `[InlineData]` using `ViewComponentWithSyncInvoke` are the same test. Either that or the data for this test and `FindSyncMethod_ReturnsNull_IfMatchCannotBeFound()` were switched at birth.

dougbu:
:shipit:

rynowak:
:shipit:

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/ViewComponentMethodSelector.cs
pranavkm: There's another usage for this on https://github.com/aspnet/Mvc/pull/3267#discussion-diff-41322297R34 where it prints `Task<T>` (without nameof etc). Just wanted these two to be consistent.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewComponents/ViewComponentMethodSelectorTest.cs
pranavkm: Yeah it started off being different. I can trim a case or two since they test the same behavior

PR #3268, opened 2015-10-06
Avoid state machine and allocation for 0-parameters

rynowak:
This change avoids going async and allocating various context objects when no binding needs to be performed.

Here's some allocation data from aspnet/benchmarks for the MVC plaintext benchmark. 3000 requests here: 
image

These two state machine types are the most relevant, but there's also `OperationBindingContext` which is below the fold here, and a bunch of framework plumbing that gets skipped over.

The total reduction is about 1.4mb for 3000 requests. Total allocations for the benchmark are on the order of 33mb total with some other changes I've got brewing, so this is a significant chunk for an action with no arguments.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/Controllers/DefaultControllerActionArgumentBinder.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Controllers/DefaultControllerActionArgumentBinder.cs
@@ -35,7 +35,7 @@ namespace Microsoft.AspNet.Mvc.Controllers
             _validator = validator;
         }
 
-        public async Task<IDictionary<string, object>> BindActionArgumentsAsync(
+        public Task<IDictionary<string, object>> BindActionArgumentsAsync(
             ActionContext actionContext,
             ActionBindingContext actionBindingContext,
             object controller)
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.Mvc.Core/Controllers/DefaultControllerActionArgumentBinder.cs
```diff
@@ -64,6 +64,28 @@ public class DefaultControllerActionArgumentBinder : IControllerActionArgumentBi
                         nameof(actionContext));
             }
 
+            if (actionDescriptor.BoundProperties.Count == 0 &&
+                actionDescriptor.Parameters.Count == 0)
+            {
+                return Task.FromResult<IDictionary<string, object>>(
+                    new Dictionary<string, object>(StringComparer.Ordinal));
```
rynowak: I also considered pulling this check out into the invoker, but it seems like a good thing to have the binder at least be called - for extensibility's sake. The code is less artificial if I put the `if` in the invoker, but that seemed like a bad choice for future proofing.

YuvalItzchakov: Is there a reason you're not caching the dictionary allocation as well?

dougbu: Why allocate the `Dictionary` or the `Task` in this mainline case? Suggest sticking the `Task` in a `static` field.

dougbu:
@rynowak I'm curious: Down to 33mb for this simplified action ... and how much for something like the `Index` action in the same controller?

src/Microsoft.AspNet.Mvc.Core/Controllers/DefaultControllerActionArgumentBinder.cs
davidfowl: cached task?

rynowak: @YuvalItzchakov - the dictionary is exposed later though a public contract and needs to be mutable. We could try to avoid it with further shenanigans, but that would be a much more impactful change for a smaller benefit.

rynowak:
> Down to 33mb for this simplified action ... and how much for something like the Index action in the same controller?

Not sure what you mean. This is what's being benchmarked - https://github.com/aspnet/benchmarks/blob/master/src/Benchmarks/Controllers/HomeController.cs#L16 The numbers are super low because it's super minimal. 

Right now my site with a simple view + model binding is allocating around 110mb. In that context, this is still a reasonably worthwhile improvement.

dougbu:
I meant using [this other action](https://github.com/aspnet/benchmarks/blob/master/src/Benchmarks/Controllers/HomeController.cs#L29). Think 110mb" is the answer to my question.

dougbu:
:shipit:

rynowak:
For posterity the answer to your question is 49.5mb for 3000 requests. That's the price of admission right now for views.

rynowak:
1ed22e59390505f9e145371cdd0ac31e5692ac3e

PR #3269, opened 2015-10-06
Optimize allocations in GetFilters

rynowak:
Doing it the old fashioned way to get rid of some allocations.

Before:
image

After:
image

This is worth about .9mb for 3000 requests. For aspnet/benchmarks mvc plaintext benchmark, this is about 2.7% of total allocations.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs
@@ -5,7 +5,6 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.Tracing;
-using System.Linq;
 using System.Runtime.ExceptionServices;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Mvc.Abstractions;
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs
```diff
@@ -237,7 +242,33 @@ private IFilterMetadata[] GetFilters()
                 _filterProviders[i].OnProvidersExecuted(context);
             }
 
-            return context.Results.Select(item => item.Filter).Where(filter => filter != null).ToArray();
+            var count = 0;
+            for (var i = 0; i < items.Count; i++)
```
rynowak: Counting up front to avoid resize when using a list, and to avoid the need to call `ToArray`

rynowak:
Ping - need someone with courage and stoutness of heart to give this a review

src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs
```diff
@@ -223,13 +224,17 @@ private set
 
         private IFilterMetadata[] GetFilters()
         {
-            var context = new FilterProviderContext(
-                ActionContext,
-                ActionContext.ActionDescriptor.FilterDescriptors.Select(fd => new FilterItem(fd)).ToList());
+            var filterDescriptors = ActionContext.ActionDescriptor.FilterDescriptors;
+            var items = new List<FilterItem>(filterDescriptors.Count);
```
pranavkm: Could this be an array?

rynowak: No, filter providers can `Add` items as well as set them.

src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs
pranavkm: Could you add a comment saying this is done this way for perf so that this doesn't get replaced with `.Count(i => i != null)`

pranavkm:
:shipit:

PR #3270, opened 2015-10-06
Remove dead code in invoker (SetStage)

rynowak:
Just some cleanup, the `StillAt` method was used in a much earlier version of this code, but now it's just vestigial.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs
@@ -178,7 +178,7 @@ namespace Microsoft.AspNet.Mvc.Controllers
             _cursor = new FilterCursor(_filters);
 
             ActionContext.ModelState.MaxAllowedErrors = _maxModelValidationErrors;
-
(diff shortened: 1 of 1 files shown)


rynowak:
Going to go ahead and merge this since it's just some deletions and a rename.

src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs
```diff
@@ -178,7 +178,7 @@ private set
             _cursor = new FilterCursor(_filters);
 
             ActionContext.ModelState.MaxAllowedErrors = _maxModelValidationErrors;
-
+
```
Eilon: Erm...

rynowak: It may display as a space on github, but it's a rarely used unicode character that increases your performances. `\u1337`

Eilon: O I C!!!!!!!!

PR #3271, opened 2015-10-06
Reducing allocations in value providers

rynowak:
- Don't go async in formdata providers unless we need to
- Remove unnecessary defensive copy in CompositeValueProvider

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/CompositeValueProvider.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/CompositeValueProvider.cs
@@ -30,8 +30,8 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         /// </summary>
         /// <param name="valueProviders">The sequence of <see cref="IValueProvider"/> to add to this instance of
         /// <see cref="CompositeValueProvider"/>.</param>
-        public CompositeValueProvider(IEnumerable<IValueProvider> valueProviders)
-            : base(valueProviders.ToList())
+        protected CompositeValueProvider(IList<IValueProvider> valueProviders)
+            : base(valueProviders)
         {
         }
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNet.Mvc.Core/ModelBinding/CompositeValueProvider.cs
```diff
@@ -30,8 +30,8 @@ public CompositeValueProvider()
         /// </summary>
         /// <param name="valueProviders">The sequence of <see cref="IValueProvider"/> to add to this instance of
         /// <see cref="CompositeValueProvider"/>.</param>
-        public CompositeValueProvider(IEnumerable<IValueProvider> valueProviders)
-            : base(valueProviders.ToList())
+        protected CompositeValueProvider(IList<IValueProvider> valueProviders)
+            : base(valueProviders)
```
rynowak: Turning this into `IList<T>` avoids the copy - moreover we never even call this constructor outside of the class and some tests.

rynowak: The important note here is that when using uber-binding attributes like `[FromForm]` we'll instantiate a new CVP for each attribute. So you can allocate N+1 copies of this per request, and N+1 calls to trivial ToList()

rynowak:
Working on getting some perf numbers for this change.

rynowak:
**Before:**
image

**After**
image


This ended up being only worth about 400kb across 3000 requests, so not super impactful. Still seems worth taking given how simple the code changes are.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CompositeValueProvider.cs
dougbu: Suggest `IReadOnlyList` and sub-classing `ReadOnlyCollection` to express intent a bit better. That is, unless I'm forgetting something that updates the `CompositeValueProvider`.

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/CompositeValueProvider.cs
rynowak: We have product code that updates the `CompositeValueProvider`. I don't see much reason to change this detail or the usage pattern in that part of the code

rynowak:
@dougbu - any other feedback?

dougbu:
:shipit:

PR #3272, opened 2015-10-06
Optimize allocations for creating ActionConstraints

rynowak:
**Before**
image

**After**
image

This change is very similar to the `GetFilters` change. This is worth about .9mb for 3000 requests to an action with a single `[HttpGet]` attribute. This is fairly constrained and simple, and has a good outcome for just removing the linq code.

There's another task here to look at the allocations for the **execution** side of action constraints.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ActionConstraints/DefaultActionConstraintProvider.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ActionConstraints/DefaultActionConstraintProvider.cs
@@ -29,9 +29,9 @@ namespace Microsoft.AspNet.Mvc.ActionConstraints
                 throw new ArgumentNullException(nameof(context));
             }
 
-            foreach (var item in context.Results)
+            for (var i = 0; i < context.Results.Count; i++)
             {
-                ProvideConstraint(item, context.HttpContext.RequestServices);
+                ProvideConstraint(context.Results[i], context.HttpContext.RequestServices);
             }
         }
 
--- a/src/Microsoft.AspNet.Mvc.Core/ActionConstraints/HttpMethodConstraint.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ActionConstraints/HttpMethodConstraint.cs
@@ -13,7 +13,7 @@ namespace Microsoft.AspNet.Mvc.ActionConstraints
     {
         public static readonly int HttpMethodConstraintOrder = 100;
 
-        private readonly IReadOnlyList<string> _methods;
+        private readonly IReadOnlyList<string> _httpMethods;
 
         private readonly string OriginHeader = "Origin";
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNet.Mvc.Core/ActionConstraints/DefaultActionConstraintProvider.cs
```diff
@@ -29,9 +29,9 @@ public void OnProvidersExecuting(ActionConstraintProviderContext context)
                 throw new ArgumentNullException(nameof(context));
             }
 
-            foreach (var item in context.Results)
+            for (var i = 0; i < context.Results.Count; i++)
```
rynowak: This was missed in the earlier boxed list enumerator cleanup.

src/Microsoft.AspNet.Mvc.Core/ActionConstraints/HttpMethodConstraint.cs
```diff
@@ -83,7 +83,16 @@ public bool Accept(ActionConstraintContext context)
...
 
-            return (HttpMethods.Any(m => m.Equals(method, StringComparison.Ordinal)));
+            for (var i = 0; i < _httpMethods.Count; i++)
+            {
+                var supportedMethod = _httpMethods[i];
+                if (string.Equals(supportedMethod, method, StringComparison.Ordinal))
+                {
+                    return true;
+                }
+            }
```
rynowak: Removed capturing lambda

src/Microsoft.AspNet.Mvc.Core/Infrastructure/DefaultActionSelector.cs
```diff
@@ -273,11 +277,31 @@ private IReadOnlyList<IActionConstraint> GetConstraints(HttpContext httpContext,
...
-                .Select(item => item.Constraint)
-                .ToList();
+            var count = 0;
+            for (var i = 0; i < context.Results.Count; i++)
+            {
+                if (context.Results[i].Constraint != null)
+                {
+                    count++;
+                }
+            }
```
rynowak: Does a first pass to find out the final count so we can avoid reallocating

src/Microsoft.AspNet.Mvc.Core/ActionConstraints/DefaultActionConstraintProvider.cs
davidfowl: Add a comment so this doesn't regress

src/Microsoft.AspNet.Mvc.Core/ActionConstraints/HttpMethodConstraint.cs
davidfowl: :smile:

dougbu:
:shipit:

PR #3288, opened 2015-10-08
Add expression rewriting to Razor

rynowak:
This change rewrites simple and safe `Expression<Func<T, U>>` expressions
into accesses to readonly fields. This allows us to cache the actual
expression and avoid repeatedly allocating and compiling it.

The rewrite is limited to cases where we know that the expression doesn't
capture, and where we support that kind of expression for evaluating
viewdata. In practice this means 'indentity' and property accessors are
allowed.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor/Compilation/ExpressionRewriter.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/Compilation/ExpressionRewriter.cs
@@ -1,0 +1,200 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq.Expressions;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace Microsoft.AspNet.Mvc.Razor.Compilation
+{
+    /// <summary>
+    /// An expression rewriter which can hoist a simple expression lambda into a private field.
+    /// </summary>
+    public class ExpressionRewriter : CSharpSyntaxRewriter
+    {
+        private static readonly string FieldNameTemplate = "__h{0}";
+
+        public ExpressionRewriter(SemanticModel semanticModel)
+        {
+            SemanticModel = semanticModel;
+
+            Expressions = new List<KeyValuePair<SimpleLambdaExpressionSyntax, IdentifierNameSyntax>>();
+        }
+
+        // We only want to rewrite expressions for the top-level class definition.
+        private bool IsInsideClass { get; set; }
+
+        private SemanticModel SemanticModel { get; }
+
+        private List<KeyValuePair<SimpleLambdaExpressionSyntax, IdentifierNameSyntax>> Expressions { get; }
+
+        public override SyntaxNode VisitClassDeclaration(ClassDeclarationSyntax node)
+        {
+            if (IsInsideClass)
+            {
+                // Avoid recursing into nested classes. 
+                return node;
+            }
+
+            Expressions.Clear();
+
+            IsInsideClass = true;
+
+            // Call base first to visit all the children and populate Expressions.
+            var classDeclaration = (ClassDeclarationSyntax)base.VisitClassDeclaration(node);
+
+            IsInsideClass = false;
+
+            var memberDeclarations = new List<MemberDeclarationSyntax>();
+            foreach (var kvp in Expressions)
+            {
+                var expression = kvp.Key;
+                var memberName = kvp.Value.GetFirstToken();
+
+                var expressionType = SemanticModel.GetTypeInfo(expression).ConvertedType;
+                var declaration = SyntaxFactory.FieldDeclaration(
+                    SyntaxFactory.List<AttributeListSyntax>(),
+                    SyntaxFactory.TokenList(
+                        SyntaxFactory.Token(SyntaxKind.PrivateKeyword),
+                        SyntaxFactory.Token(SyntaxKind.StaticKeyword),
+                        SyntaxFactory.Token(SyntaxKind.ReadOnlyKeyword)),
+                    SyntaxFactory.VariableDeclaration(
+                        SyntaxFactory.ParseTypeName(expressionType.ToDisplayString(
+                            SymbolDisplayFormat.FullyQualifiedFormat)),
+                        SyntaxFactory.SingletonSeparatedList(
+                            SyntaxFactory.VariableDeclarator(
+                                memberName,
+                                SyntaxFactory.BracketedArgumentList(),
+                                SyntaxFactory.EqualsValueClause(expression)))))
+                    .WithTriviaFrom(expression);
+                memberDeclarations.Add(declaration);
+            }
+
+            return classDeclaration.AddMembers(memberDeclarations.ToArray());
+        }
+
+        public override SyntaxNode VisitSimpleLambdaExpression(SimpleLambdaExpressionSyntax node)
+        {
+            Debug.Assert(IsInsideClass);
+
+            // If this lambda is an Expression and is suitable for hoisting, we rewrite this into a field access.
+            //
+            //  Before:
+            //      public Task ExecuteAsync(...)
+            //      {
+            //          ...
+            //          Html.EditorFor(m => m.Price);
+            //          ...
+            //      }
+            //
+            //
+            //  After:
+            //      private static readonly Expression<Func<Product, decimal>> __h0 = m => m.Price;
+            //      public Task ExecuteAsync(...)
+            //      {
+            //          ...
+            //          Html.EditorFor(__h0);
+            //          ...
+            //      }
+            //
+            var type = SemanticModel.GetTypeInfo(node);
+            if (type.ConvertedType.Name != typeof(Expression).Name &&
+                type.ConvertedType.ContainingNamespace.Name != typeof(Expression).Namespace)
+            {
+                return node;
+            }
+
+            if (!node.Parent.IsKind(SyntaxKind.Argument))
+            {
+                return node;
+            }
+
+            var parameter = node.Parameter;
+            if (IsValidForHoisting(parameter, node.Body))
+            {
+                // Replace with a MemberAccess
+                var memberName = string.Format(FieldNameTemplate, Expressions.Count);
+                var memberAccess = PadMemberAccess(node, SyntaxFactory.IdentifierName(memberName));
+                Expressions.Add(new KeyValuePair<SimpleLambdaExpressionSyntax, IdentifierNameSyntax>(node, memberAccess));
+                return memberAccess;
+            }
+
+            return node;
+        }
+
+        private static IdentifierNameSyntax PadMemberAccess(
+            SimpleLambdaExpressionSyntax node,
+            IdentifierNameSyntax memberAccess)
+        {
+            // We want to make the new span 
+            var originalSpan = node.GetLocation().GetMappedLineSpan();
+
+            // Start by collecting all the trivia 'inside' the expression - we need to tack that on the end, but
+            // if it ends with a newline, don't include that.
+            var innerTrivia = SyntaxFactory.TriviaList(node.DescendantTrivia(descendIntoChildren: n => true));
+            if (innerTrivia.Count > 0 && innerTrivia[innerTrivia.Count - 1].IsKind(SyntaxKind.EndOfLineTrivia))
+            {
+                innerTrivia = innerTrivia.RemoveAt(innerTrivia.Count - 1);
+            }
+
+            memberAccess = memberAccess.WithTrailingTrivia(innerTrivia);
+
+            // If everything is all on one line, then make sure the spans are the same, to compensate
+            // for the expression potentially being longer than the variable name.
+            var lineSpan = originalSpan.EndLinePosition.Line - originalSpan.StartLinePosition.Line;
+            if (lineSpan == 0)
+            {
+                var padding = node.Span.Length - memberAccess.FullSpan.Length;
+                var trailingTrivia =
+                    SyntaxFactory.TriviaList(memberAccess.GetTrailingTrivia())
+                    .Add(SyntaxFactory.Whitespace(new string(' ', padding)))
+                    .AddRange(node.GetTrailingTrivia());
+
+                return
+                    memberAccess
+                    .WithLeadingTrivia(node.GetLeadingTrivia())
+                    .WithTrailingTrivia(trailingTrivia);
+            }
+            else
+            {
+                // If everything isn't on the same line, we need to pad out the last line.
+                var padding =
+                    originalSpan.EndLinePosition.Character -
+                    originalSpan.StartLinePosition.Character;
+
+                var trailingTrivia =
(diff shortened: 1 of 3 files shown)


rynowak:
Allocations from a 3001 requests to a view with 3 `LabelFor(...)`s and 3 `EditorFor(...)`s

**Before**
image

**After**
image


All of the Expression types and their related objects are gone. The overall reduction in allocations is about 9mb - which is huuuuuuuuuuuuuge.

src/Microsoft.AspNet.Mvc.Razor/Compilation/ExpressionRewriter.cs
```diff
@@ -0,0 +1,161 @@
...
+            Expressions = new List<KeyValuePair<SimpleLambdaExpressionSyntax, IdentifierNameSyntax>>();
+        }
+
+        private SemanticModel SemanticModel { get; }
+
+        private List<KeyValuePair<SimpleLambdaExpressionSyntax, IdentifierNameSyntax>> Expressions { get; }
+
+        public override SyntaxNode VisitClassDeclaration(ClassDeclarationSyntax node)
+        {
+            Expressions.Clear();
```
Eilon: Interesting. Is the visitor guaranteed to do "one class at a time" (e.g. the way you think of classes containing members)? What happens with nested types? Or partial classes? Surely a bit esoteric but we need to be careful.

src/Microsoft.AspNet.Mvc.Razor/Compilation/ExpressionRewriter.cs
```diff
@@ -0,0 +1,161 @@
...
+
+                var expressionType = SemanticModel.GetTypeInfo(expression);
+                var declaration = SyntaxFactory.FieldDeclaration(
+                    SyntaxFactory.List<AttributeListSyntax>(),
+                    SyntaxFactory.TokenList(
+                        SyntaxFactory.Token(SyntaxKind.PrivateKeyword),
+                        SyntaxFactory.Token(SyntaxKind.StaticKeyword),
+                        SyntaxFactory.Token(SyntaxKind.ReadOnlyKeyword)),
+                    SyntaxFactory.VariableDeclaration(
+                        SyntaxFactory.ParseTypeName(expressionType.ConvertedType.ToDisplayString()),
```
Eilon: Couldn't easily find the docs, but is a `DisplayString` the right thing here?

src/Microsoft.AspNet.Mvc.Razor/Compilation/ExpressionRewriter.cs
```diff
@@ -0,0 +1,161 @@
...
+            else
+            {
+                // If everything isn't on the same line, we need to pad out the last line.
+                var padding =
+                    originalSpan.EndLinePosition.Character -
+                    originalSpan.StartLinePosition.Character;
+
+                var trailingTrivia =
+                    SyntaxFactory.TriviaList(memberAccess.GetTrailingTrivia())
+                    .Add(SyntaxFactory.Whitespace(new string(' ', padding)))
```
Eilon: Just had a crazy thought. I heard some crazy people use tabs - yes, **TABS** - in their code. If we take a *character count* that had tabs in it, and replace it with an equivalent number of spaces, then horrible things might happen! Is it really our place in life to punish those who might dare to use tabs??

src/Microsoft.AspNet.Mvc.Razor/Compilation/ExpressionRewriter.cs
rynowak: You can't define classes in Razor :laughing:

rynowak: If this is a concern we can make it more robust, or at least we could add some validation that it's not being used that way. I left it simple on purpose since there's no way to to run this on non-Razor code.

src/Microsoft.AspNet.Mvc.Razor/Compilation/ExpressionRewriter.cs
rynowak: Tabs intermingled with the code would be handled by L98. The padding that we manufacture is only there to accommodate the difference in spans between our expression and 'non-trivia' code.

src/Microsoft.AspNet.Mvc.Razor/Compilation/ExpressionRewriter.cs
```diff
@@ -0,0 +1,161 @@
...
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace Microsoft.AspNet.Mvc.Razor.Compilation
+{
+    /// <summary>
+    /// An expression rewriter which can hoist a simple expression lambda into a private field.
+    /// </summary>
+    public class ExpressionRewriter : CSharpSyntaxRewriter
+    {
+        private static readonly string FieldNameTemplate = "__h{0}";
```
Eilon: Is there any padding issue that can happen is the expression we rewrite *to* is longer than the original expression? It's definitely difficult for that to happen, but the identity expression `x=>x` (4 chars) could be rewritten to `__h10` (5 chars).

src/Microsoft.AspNet.Mvc.Razor/Compilation/ExpressionRewriter.cs
Eilon: Sure you can. In `@functions` you can declare nested types.

src/Microsoft.AspNet.Mvc.Razor/Compilation/ExpressionRewriter.cs
rynowak: Looks like we should be passing https://github.com/dotnet/roslyn/blob/master/src/Compilers/Core/Portable/SymbolDisplay/SymbolDisplayFormat.cs#L125 to include `global::` but otherwise seems right.

src/Microsoft.AspNet.Mvc.Razor/Compilation/ExpressionRewriter.cs
rynowak: that's wierd. OK will correct that logic and add a test.

src/Microsoft.AspNet.Mvc.Razor/Compilation/ExpressionRewriter.cs
Eilon: Ah, I see...

src/Microsoft.AspNet.Mvc.Razor/Compilation/ExpressionRewriter.cs
rynowak: Absolutely. You won't see an artifact like that if the expression is on a single line because the debugger highlights the line.

If we need to address it we can increment 'h' as well.

src/Microsoft.AspNet.Mvc.Razor/Compilation/ExpressionRewriter.cs
Eilon: Weird, but not entirely uncommon. Especially in Web Pages where you might declare helper data structures.

src/Microsoft.AspNet.Mvc.Razor/Compilation/ExpressionRewriter.cs
Eilon: I'm fine logging that as something to look at later.

src/Microsoft.AspNet.Mvc.Razor/Compilation/ExpressionRewriter.cs
pranavkm: You could use `IMvcRazorHost.MainClassPrefix` to determine if you're the the actual view being rendered.

src/Microsoft.AspNet.Mvc.Razor/Compilation/ExpressionRewriter.cs
pranavkm: AFAIK, it's the string as it would appear in code and not the GetType().FullName. In this case, it would be the right string to use.

pranavkm: Would it be easier to follow this if we this was written as a string and parsed? It'd be a lot easier to read than looking at a SyntaxTree structure.

src/Microsoft.AspNet.Mvc.Razor/Compilation/ExpressionRewriter.cs
```diff
@@ -0,0 +1,161 @@
...
+                memberDeclarations.Add(declaration);
+            }
+            
+            return classDeclaration.AddMembers(memberDeclarations.ToArray());
+        }
+
+        public override SyntaxNode VisitSimpleLambdaExpression(SimpleLambdaExpressionSyntax node)
+        {
+            var type = SemanticModel.GetTypeInfo(node);
+            if (type.ConvertedType.Name!= typeof(Expression).Name &&
```
pranavkm: formatting

src/Microsoft.AspNet.Mvc.Razor/Compilation/ExpressionRewriter.cs
```diff
@@ -0,0 +1,161 @@
...
+                    .WithTriviaFrom(expression);
+                memberDeclarations.Add(declaration);
+            }
+            
+            return classDeclaration.AddMembers(memberDeclarations.ToArray());
+        }
+
+        public override SyntaxNode VisitSimpleLambdaExpression(SimpleLambdaExpressionSyntax node)
+        {
+            var type = SemanticModel.GetTypeInfo(node);
```
pranavkm: Could we make a comment on what the generated code looks like before and after?

src/Microsoft.AspNet.Mvc.Razor/Compilation/ExpressionRewriter.cs
```diff
@@ -0,0 +1,161 @@
...
+                    .AddRange(node.GetTrailingTrivia());
+
+                return 
+                    memberAccess
+                    .WithLeadingTrivia(node.GetLeadingTrivia())
+                    .WithTrailingTrivia(trailingTrivia);
+            }
+        }
+
+        private bool IsValidForHoisting(ParameterSyntax parameter, CSharpSyntaxNode node)
```
pranavkm: `static`

src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationService.cs
```diff
@@ -90,10 +90,13 @@ public CompilationResult Compile(RelativeFileInfo fileInfo, string compilationCo
             var compilationOptions = compilationSettings.CompilationOptions
```
pranavkm: Care to also format the indentation for this while you're modifying this file?

test/Microsoft.AspNet.Mvc.Razor.Test/Compilation/ExpressionRewriterTest.cs
```diff
@@ -0,0 +1,420 @@
...
+            var compilation = Compile(tree);
+            var semanticModel = compilation.GetSemanticModel(tree, ignoreAccessibility: true);
+
+            var rewriter = new ExpressionRewriter(semanticModel);
+
+            // Act
+            var result = rewriter.Visit(tree.GetRoot());
+
+            // Assert
+            var fields = FindFields(result);
```
pranavkm: Could we instead Assert `tree.GetText()` matches the expected output?

test/Microsoft.AspNet.Mvc.Razor.Test/Compilation/ExpressionRewriterTest.cs
```diff
@@ -0,0 +1,420 @@
...
+            var services = CallContextServiceLocator.Locator.ServiceProvider;
+            var libraryExporter = services.GetRequiredService<ILibraryExporter>();
+            var environment = services.GetRequiredService<IApplicationEnvironment>();
+
+            var references = new List<MetadataReference>();
+
+            var libraryExports = libraryExporter.GetAllExports(environment.ApplicationName);
+            foreach (var export in libraryExports.MetadataReferences)
+            {
+                references.Add(export.ConvertMetadataReference(MetadataReferenceExtensions.CreateAssemblyMetadata));
```
pranavkm: Sweet

pranavkm:
:watch:

src/Microsoft.AspNet.Mvc.Razor/Compilation/ExpressionRewriter.cs
Eilon: Yeah if we want to restrict it to just the "page" then maybe there are ways of doing that.

rynowak: Two restrictions I'm planning on adding here:
- Only consider candidates inside an `ArgumentSyntax` - skip over members and variable declarations
- Only consider candidates inside a inside a top-level class

src/Microsoft.AspNet.Mvc.Razor/Compilation/ExpressionRewriter.cs
rynowak: Gross. Really not a fan of that idea.

test/Microsoft.AspNet.Mvc.Razor.Test/Compilation/ExpressionRewriterTest.cs
rynowak: I'd really rather not. That ends up being a lot like baselining.

src/Microsoft.AspNet.Mvc.Razor/Compilation/ExpressionRewriter.cs
rynowak: sure

rynowak:
Updated. MVC doesn't build right now so I can't verify the test changes.

src/Microsoft.AspNet.Mvc.Razor/Compilation/ExpressionRewriter.cs
Eilon: Yeah top-level only will be safe with Razor.

src/Microsoft.AspNet.Mvc.Razor/Compilation/ExpressionRewriter.cs
```diff
@@ -0,0 +1,200 @@
...
+        // We only want to rewrite expressions for the top-level class definition.
+        private bool IsInsideClass { get; set; }
+
+        private SemanticModel SemanticModel { get; }
+
+        private List<KeyValuePair<SimpleLambdaExpressionSyntax, IdentifierNameSyntax>> Expressions { get; }
+
+        public override SyntaxNode VisitClassDeclaration(ClassDeclarationSyntax node)
+        {
+            if (IsInsideClass)
```
Eilon: Is there a way to just "know" that you're in a top-level class? Like `node.IsTopLevelMember` or something like that?

Eilon:
Aside from the one small comment, :shipit:

pranavkm:
:shipit:

src/Microsoft.AspNet.Mvc.Razor/Compilation/ExpressionRewriter.cs
rynowak: There's a few ways to do this with the semantic model, but it all seems less straightforward to me than what I'm doing.

PR #3303, opened 2015-10-11
Test `[FromServices]` with service that is not available

dougbu:
- also test `[FromServices]` for a defined type without a `BinderModelName`

### DIFF
--- a/test/Microsoft.AspNet.Mvc.IntegrationTests/ServicesModelBinderIntegrationTest.cs
+++ b/test/Microsoft.AspNet.Mvc.IntegrationTests/ServicesModelBinderIntegrationTest.cs
@@ -1,6 +1,8 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+using System.Collections.Generic;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Mvc.Abstractions;
 using Microsoft.AspNet.Mvc.Formatters;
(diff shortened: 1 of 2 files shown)


dougbu:
@pranavkm updated

pranavkm:
:shipit:

Eilon:
@dougbu is this the one where you were going to add a case for `IEnumerable<SomeService>`?

dougbu:
@Eilon yup. I thought I'd successfully force-pushed my branch before I added the "updated" comment above. That failed but the extra methods are now visible here.

dougbu:
In particular see `BindEnumerableParameterFromService_NoPrefix_GetsBound()` and `BindEnumerableParameterFromService_NoService_GetsBound()`.

test/Microsoft.AspNet.Mvc.IntegrationTests/ServicesModelBinderIntegrationTest.cs
```diff
@@ -132,5 +134,139 @@ private class Address
...
+
+            // Act
+            var modelBindingResult = await argumentBinder.BindModelAsync(parameter, modelState, operationContext);
+
+            // Assert
+            // ModelBindingResult
+            Assert.True(modelBindingResult.IsModelSet);
+
+            // Model
+            var formatterArray = Assert.IsType<IActionResult[]>(modelBindingResult.Model);
```
Eilon: Wrong variable name?

dougbu: Nice catch!

Eilon:
:shipit:

dougbu:
2e2043f

PR #3310, opened 2015-10-12
 * Allow null ViewData and TempData

ryanbrandenburg:
If you're manually constructing a ViewResult you shouldn't have to pass in a ViewData and TempData just for us. This change allows them to be left null, the behavior on a resulting page being that references to ViewData just don't put anything out and references to TempData cause a 500.

### DIFF
--- a/test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewFeatures/DefaultDisplayTemplatesTest.cs
+++ b/test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewFeatures/DefaultDisplayTemplatesTest.cs
@@ -122,12 +122,11 @@ namespace Microsoft.AspNet.Mvc.ViewFeatures
         public void ObjectTemplate_IgnoresPropertiesWith_ScaffoldColumnFalse()
         {
             // Arrange
-            var expected =
-@"<div class=""HtmlEncode[[display-label]]"">HtmlEncode[[Property1]]</div>
-<div class=""HtmlEncode[[display-field]]""></div>
-<div class=""HtmlEncode[[display-label]]"">HtmlEncode[[Property3]]</div>
-<div class=""HtmlEncode[[display-field]]""></div>
-";
+            var expected = "<div class=\"HtmlEncode[[display-label]]\">HtmlEncode[[Property1]]</div>" + Environment.NewLine +
+                "<div class=\"HtmlEncode[[display-field]]\"></div>"+ Environment.NewLine +
+                "<div class=\"HtmlEncode[[display-label]]\">HtmlEncode[[Property3]]</div>"+ Environment.NewLine +
+                "<div class=\"HtmlEncode[[display-field]]\"></div>"+ Environment.NewLine;
+
             var model = new DefaultTemplatesUtilities.ObjectWithScaffoldColumn();
             var viewEngine = new Mock<ICompositeViewEngine>();
             viewEngine.Setup(v => v.FindPartialView(It.IsAny<ActionContext>(), It.IsAny<string>()))
--- a/test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewFeatures/ViewExecutorTest.cs
+++ b/test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewFeatures/ViewExecutorTest.cs
@@ -8,12 +8,14 @@ using System.IO;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Internal;
 using Microsoft.AspNet.Mvc.Abstractions;
 using Microsoft.AspNet.Mvc.ModelBinding;
 using Microsoft.AspNet.Mvc.Rendering;
 using Microsoft.AspNet.Mvc.ViewEngines;
 using Microsoft.AspNet.Routing;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Net.Http.Headers;
 using Moq;
 using Xunit;
@@ -110,6 +112,62 @@ namespace Microsoft.AspNet.Mvc.ViewFeatures
             // Assert
             Assert.Equal(expectedContentType, context.Response.ContentType);
             Assert.Equal("abcd", Encoding.UTF8.GetString(memoryStream.ToArray()));
+        }
+
+        private static IServiceProvider GetServiceProvider()
+        {
+            var httpContext = new HttpContextAccessor() { HttpContext = new DefaultHttpContext() };
+            var serviceCollection = new ServiceCollection();
+            serviceCollection.AddInstance<IModelMetadataProvider>(new EmptyModelMetadataProvider());
+            var tempDataProvider = new SessionStateTempDataProvider();
+            serviceCollection.AddInstance<ITempDataDictionary>(new TempDataDictionary(httpContext, tempDataProvider));
+
+            return serviceCollection.BuildServiceProvider();
+        }
+
+        [Fact]
+        public async Task ExecuteAsync_ViewResultAllowNull()
+        {
+            // Arrange
+            var tempDataNull = false;
+            var viewDataNull = false;
+            var deligateHit = false;
+
+            var view = CreateView(async (v) =>
+            {
+                deligateHit = true;
+                tempDataNull = v.TempData == null;
+                viewDataNull = v.ViewData == null;
+
+                await v.Writer.WriteAsync("abcd");
+            });
+            var context = new DefaultHttpContext();
+
+            var memoryStream = new MemoryStream();
+            context.Response.Body = memoryStream;
+
+            var actionContext = new ActionContext(
+                context,
+                new RouteData(),
+                new ActionDescriptor());
+            
+            context.RequestServices = GetServiceProvider();
+            var viewExecutor = CreateViewExecutor();
+
+            // Act
+            await viewExecutor.ExecuteAsync(
+                actionContext,
+                view,
+                null,
+                null,
+                contentType: null,
+                statusCode: 200);
+
+            // Assert
+            Assert.Equal(200, context.Response.StatusCode);
+            Assert.True(deligateHit);
+            Assert.False(viewDataNull);
+            Assert.False(tempDataNull);
         }
 
         [Fact]
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponentResult.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponentResult.cs
@@ -84,6 +84,12 @@ namespace Microsoft.AspNet.Mvc
                 viewData = new ViewDataDictionary(modelMetadataProvider, context.ModelState);
             }
 
+            var tempData = TempData;
+            if (tempData == null)
+            {
+                tempData = services.GetRequiredService<ITempDataDictionary>();
+            }
+
             var contentType = ContentType;
             if (contentType != null && contentType.Encoding == null)
             {
@@ -98,8 +104,8 @@ namespace Microsoft.AspNet.Mvc
             //      3. ViewExecutor.DefaultContentType (sensible default)
             //
             //
-            response.ContentType = 
-                contentType?.ToString() ?? 
+            response.ContentType =
+                contentType?.ToString() ??
                 response.ContentType ??
                 ViewExecutor.DefaultContentType.ToString();
 
@@ -115,7 +121,7 @@ namespace Microsoft.AspNet.Mvc
                     context,
                     NullView.Instance,
                     viewData,
-                    TempData,
+                    tempData,
                     writer,
                     htmlHelperOptions);
 
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/ViewExecutor.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/ViewExecutor.cs
@@ -6,8 +6,10 @@ using System.Diagnostics;
 using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Mvc.Infrastructure;
+using Microsoft.AspNet.Mvc.ModelBinding;
 using Microsoft.AspNet.Mvc.Rendering;
 using Microsoft.AspNet.Mvc.ViewEngines;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.OptionsModel;
 using Microsoft.Net.Http.Headers;
 
@@ -117,14 +119,16 @@ namespace Microsoft.AspNet.Mvc.ViewFeatures
                 throw new ArgumentNullException(nameof(view));
             }
 
+            var services = actionContext.HttpContext.RequestServices;
             if (viewData == null)
             {
-                throw new ArgumentNullException(nameof(viewData));
+                var metadataProvider = services.GetRequiredService<IModelMetadataProvider>();
+                viewData = new ViewDataDictionary(metadataProvider);
             }
 
             if (tempData == null)
             {
-                throw new ArgumentNullException(nameof(tempData));
+                tempData = services.GetRequiredService<ITempDataDictionary>();
             }
 
             var response = actionContext.HttpContext.Response;
--- a/test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewComponentResultTest.cs
+++ b/test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewComponentResultTest.cs
@@ -30,6 +30,32 @@ namespace Microsoft.AspNet.Mvc
             new TempDataDictionary(new HttpContextAccessor(), new SessionStateTempDataProvider());
 
         [Fact]
+        public async Task ExecuteAsync_ViewComponentResult_AllowsNullViewDataAndTempData()
+        {
+            // Arrange
+            var descriptor = new ViewComponentDescriptor()
+            {
+                FullName = "Full.Name.Text",
+                ShortName = "Text",
+                Type = typeof(TextViewComponent),
+            };
+
+            var actionContext = CreateActionContext(descriptor);
+            
+            var viewComponentResult = new ViewComponentResult
+            {
+                Arguments = new object[] { "World!" },
+                ViewData = null,
+                TempData = null,
+                ViewComponentName = "Text"
+            };
+
+            // Act
+            await viewComponentResult.ExecuteResultAsync(actionContext);
+            // No assert, just confirm it didn't throw
+        }
+
+        [Fact]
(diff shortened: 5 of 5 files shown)


test/Microsoft.AspNet.Mvc.Core.Test/PhysicalFileResultTest.cs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
```
ryanbrandenburg: This snuck in from another PR, I'll remove it when I rebase.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewResultTest.cs
```diff
@@ -97,6 +97,54 @@ public class ViewResultTest
             viewEngine.Verify();
         }
 
+        [Fact]
+        public async Task ExecuteResultAsync_ViewResultAllowNullViewDataAndTempData()
+        {
+            var context = GetActionContext();
```
pranavkm: we usually decorate tests with `// Arrange`, `// Act` and `// Assert`. Have a look at other tests in this class.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewResultTest.cs
```diff
@@ -97,6 +97,54 @@ public class ViewResultTest
...
+            var controller = new TestController(context);
+
+            await controller.Index().ExecuteResultAsync(context);
+            Assert.Null(controller.ViewContext.ViewData);
+            Assert.Null(controller.ViewContext.TempData);
+        }
+
+        private class TestController
+        {
+            public ViewContext ViewContext { get; set; }
```
Eilon: Re-sort these a bit:

```c#
private class Foo
{
    // fields here

    // ctors here

    // props here

    // methods here
}
```

Including an empty line between each.

Eilon:
Aside from the small comments, :shipit:

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewResultTest.cs
```diff
@@ -97,6 +97,54 @@ public class ViewResultTest
...
+        {
+            var context = GetActionContext();
+            var controller = new TestController(context);
+
+            await controller.Index().ExecuteResultAsync(context);
+            Assert.Null(controller.ViewContext.ViewData);
+            Assert.Null(controller.ViewContext.TempData);
+        }
+
+        private class TestController
```
pranavkm: Move this test to `ViewExecutorTest.cs`. You shouldn't need a Controller to verify this.

Eilon: Ah that's actually a good point. Can just new up the type and call Execute from within the test.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/ViewContext.cs
```diff
@@ -58,16 +58,6 @@ public ViewContext()
                 throw new ArgumentNullException(nameof(view));
             }
 
-            if (viewData == null)
```
dougbu: Expect lots of tag and HTML helpers to blow up if `ViewContext.ViewData` is `null`. May also experience NREs if `TempData` is allowed to be null here.

Why not instead leave `ViewContext` unchanged and have `ViewExecutor` new up instances if it's passed `null`?

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/ViewContext.cs
ryanbrandenburg: That's what I was originally doing, but @Eilon suggested we just leave the things null. He thought it should only become a problem if they tried to use it explicitly in their template, if there are calls using it which might NRE then I think that's a strong reason to just new-up an empty one, but on the other hand this also is a scenario you get to by returning a ViewResult directly, rather than letting us do all the plumbing, so I guess I can see it both ways. @Eilon, thoughts?

Eilon: @dougbu ah that's interesting. I suppose it may very well be better to have dummy ViewData/TempData for those cases. When I talked to @ryanbrandenburg I was indeed thinking only of explicit calls to ViewData/TempData from the view code, which I would be fine with failing.

rynowak:
>  I suppose it may very well be better to have dummy ViewData/TempData for those cases. 

^^^ - was thinking the same thought earlier about this bug. There's too much stuff that will *maybe* read `ViewData` to give a consistent experience around reporting when it's `null` and not without just requiring `ViewExector` to create one if needed.

ryanbrandenburg:
Am I signed off then?

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/ViewContext.cs
```diff
@@ -58,24 +59,24 @@ public ViewContext()
...
             {
-                throw new ArgumentNullException(nameof(writer));
+                viewData = new ViewDataDictionary(new EmptyModelMetadataProvider());
             }
 
-            if (htmlHelperOptions == null)
+            if (tempData == null)
             {
-                throw new ArgumentNullException(nameof(htmlHelperOptions));
+                tempData = new TempDataDictionary(new HttpContextAccessor(), new SessionStateTempDataProvider());
```
pranavkm: The `HttpContextAccessor` would need to point to the current `ActionContext.HttpContext`

rynowak: We should be service-locating `ITempDataDictionary` and `IModelMetadataProvider` here - this isn't a unit-test-only scenario.

ryanbrandenburg:
@rynowak Do those changes satisfy the dependency injection problems?

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/ViewContext.cs
```diff
@@ -58,24 +59,26 @@ public ViewContext()
...
+                throw new ArgumentNullException(nameof(writer));
             }
 
-            if (tempData == null)
+            if (htmlHelperOptions == null)
             {
-                throw new ArgumentNullException(nameof(tempData));
+                throw new ArgumentNullException(nameof(htmlHelperOptions));
             }
+            var service = actionContext.HttpContext.RequestServices;
```
rynowak: `services`

ryanbrandenburg:
@rynowak is that better?

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/ViewContext.cs
```diff
@@ -58,24 +59,24 @@ public ViewContext()
...
             {
-                throw new ArgumentNullException(nameof(writer));
+                throw new ArgumentNullException(nameof(viewData));
             }
 
-            if (htmlHelperOptions == null)
+            if (tempData == null)
             {
-                throw new ArgumentNullException(nameof(htmlHelperOptions));
+                throw new ArgumentNullException(nameof(tempData));
```
dougbu: Restore these `null` checks in their original order. That nicely matched the parameter order.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/ViewContext.cs
```diff
@@ -116,7 +117,8 @@ public ViewContext()
 
             if (viewData == null)
             {
-                throw new ArgumentNullException(nameof(viewData));
+                var metadata = viewContext.HttpContext.RequestServices.GetRequiredService<IModelMetadataProvider>();
+                viewData = new ViewDataDictionary(metadata);
             }
```
dougbu: Keep all of the `null` checks together. Move this down below the last of them.

src/Microsoft.AspNet.Mvc.ViewFeatures/project.json
```diff
@@ -12,6 +12,7 @@
         "Microsoft.AspNet.Antiforgery": "1.0.0-*",
         "Microsoft.AspNet.Diagnostics.Abstractions": "1.0.0-*",
         "Microsoft.AspNet.Html.Abstractions": "1.0.0-*",
+        "Microsoft.AspNet.Http": "1.0.0-*",
```
dougbu: What breaks without this dependency?

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewFeatures/DefaultDisplayTemplatesTest.cs
```diff
@@ -138,7 +138,7 @@ public void ObjectTemplate_IgnoresPropertiesWith_ScaffoldColumnFalse()
             var result = DefaultDisplayTemplates.ObjectTemplate(htmlHelper);
 
             // Assert
-            Assert.Equal(expected, HtmlContentUtilities.HtmlContentToString(result));
+            Assert.Equal(expected, HtmlContentUtilities.HtmlContentToString(result), ignoreLineEndingDifferences: true);
```
dougbu: There's a lot of `Assert.Equal()` calls in this class and only one needs `ignoreLineEndingDifferences`? Why?

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/ViewContext.cs
dougbu: Actually, this fallback is either unnecessary or incorrect. Since this copy constructor is only used internally for `HtmlHelper` and `ViewViewComponentResult` scenarios where the passed `ViewDataDictionary` should never be `null`, I lean toward the first. (Am I missing a scenario?) Otherwise the VDD should be created from the one in the passed `ViewContext`.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/ViewContext.cs
```diff
@@ -58,24 +59,24 @@ public ViewContext()
...
             {
-                throw new ArgumentNullException(nameof(tempData));
+                throw new ArgumentNullException(nameof(htmlHelperOptions));
             }
 
-            if (writer == null)
+            if (viewData == null)
             {
-                throw new ArgumentNullException(nameof(writer));
+                throw new ArgumentNullException(nameof(viewData));
```
dougbu: This is the constructor that `ViewComponentResult` and `ViewExecutor` use but `ViewExecutor` only provides a fallback `ITempDataDictionary` instance. Can't we still hit this `Exception` if `ViewData` is left `null` in a manually-crafted `ViewResult`?

dougbu:
Wondering whether we should leave all but the main `ViewContext` constructor unchanged. That is, place the `TempData` and `ViewData` fallbacks in that constructor. May need minor `ViewComponentResult` and `ViewExecutor` updates to remove `null` checks. Then the fallback logic would then be in just one place.

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/ViewContext.cs
rynowak: `ViewData` calls `ViewExecutor`

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewFeatures/ViewExecutorTest.cs
```diff
@@ -112,6 +114,59 @@ public class ViewExecutorTest
...
+            serviceCollection.AddInstance<ITempDataDictionary>(new TempDataDictionary(httpContext, tempDataProvider));
+
+            return serviceCollection.BuildServiceProvider();
+        }
+
+        [Fact]
+        public async Task ExecuteAsync_ViewResultAllowNull()
+        {
+            bool tempDataNull = false;
+            bool viewDataNull = false;
```
rynowak: var

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewFeatures/ViewExecutorTest.cs
```diff
@@ -112,6 +114,59 @@ public class ViewExecutorTest
...
+            var serviceCollection = new ServiceCollection();
+            serviceCollection.AddInstance<IModelMetadataProvider>(new EmptyModelMetadataProvider());
+            var tempDataProvider = new SessionStateTempDataProvider();
+            serviceCollection.AddInstance<ITempDataDictionary>(new TempDataDictionary(httpContext, tempDataProvider));
+
+            return serviceCollection.BuildServiceProvider();
+        }
+
+        [Fact]
+        public async Task ExecuteAsync_ViewResultAllowNull()
```
rynowak: Need the same unit test for a `ViewComponentResult`

rynowak:
:watch:

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/ViewContext.cs
ryanbrandenburg: You're correct, this should just be a null check, I missed it when cleaning up.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewFeatures/DefaultDisplayTemplatesTest.cs
ryanbrandenburg: Because the rest of them are constructed using Environment.NewLine and this one is constructed using @"". I'll rewrite it to more closely match the other tests.

ryanbrandenburg:
Doug, about consolidating into the ViewContext, @rynowak is of the opinion, and he convinced me, that contexts should be dumb bags of objects that don't have any logic in them. Your suggestion would certainly simplify things a bit, but at the expense of having the Context violate its dumbness.

ryanbrandenburg:
I believe I've addressed everything now, is there a ship it?

rynowak:
:shipit:

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/ViewContext.cs
```diff
@@ -6,6 +6,7 @@
 using Microsoft.AspNet.Mvc.ModelBinding;
 using Microsoft.AspNet.Mvc.ViewEngines;
 using Microsoft.AspNet.Mvc.ViewFeatures;
+using Microsoft.Extensions.DependencyInjection;
```
dougbu: Minor but you don't need this anymore. That is, revert everything in this file.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewFeatures/DefaultDisplayTemplatesTest.cs
```diff
@@ -122,12 +122,10 @@ public void ObjectTemplateDisplaysNullDisplayTextWhenObjectIsNull()
...
-@"<div class=""HtmlEncode[[display-label]]"">HtmlEncode[[Property1]]</div>
-<div class=""HtmlEncode[[display-field]]""></div>
-<div class=""HtmlEncode[[display-label]]"">HtmlEncode[[Property3]]</div>
-<div class=""HtmlEncode[[display-field]]""></div>
-";
+            var expected = "<div class=\"HtmlEncode[[display-label]]\">HtmlEncode[[Property1]]</div>" + Environment.NewLine +
+"<div class=\"HtmlEncode[[display-field]]\"></div>"+Environment.NewLine +
+"<div class=\"HtmlEncode[[display-label]]\">HtmlEncode[[Property3]]</div>"+Environment.NewLine+
+"<div class=\"HtmlEncode[[display-field]]\"></div>"+Environment.NewLine;
             var model = new DefaultTemplatesUtilities.ObjectWithScaffoldColumn();
```
dougbu: Strange indentation now that you're not using a here string. Also try to get into the habit of putting spaces around binary operators even in test code. That way you won't mess up the product code.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewFeatures/ViewExecutorTest.cs
```diff
@@ -112,6 +114,59 @@ public class ViewExecutorTest
...
+
+            var memoryStream = new MemoryStream();
+            context.Response.Body = memoryStream;
+
+            var actionContext = new ActionContext(
+                context,
+                new RouteData(),
+                new ActionDescriptor());
+
+
```
dougbu: minor: One blank line is plenty.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewFeatures/ViewExecutorTest.cs
```diff
@@ -112,6 +114,59 @@ public class ViewExecutorTest
...
+            return serviceCollection.BuildServiceProvider();
+        }
+
+        [Fact]
+        public async Task ExecuteAsync_ViewResultAllowNull()
+        {
+            var tempDataNull = false;
+            var viewDataNull = false;
+            // Arrange
```
dougbu: Move up 2 lines.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewFeatures/ViewExecutorTest.cs
```diff
@@ -112,6 +114,59 @@ public class ViewExecutorTest
...
+                view,
+                null,
+                null,
+                contentType: null,
+                statusCode: 200);
+
+            // Assert
+            Assert.Equal(200, context.Response.StatusCode);
+            Assert.False(viewDataNull);
+            Assert.False(tempDataNull);
```
dougbu: Please add an assertion showing the delegate calling `WriteAsync()` was actually invoked. Might be worth another Boolean variable.

dougbu:
:shipit: once the CI builds start working. Problem looks like something propagating through our builds and not at Coherence yet. But it could hide something real.
```
C:\projects\mvc\test\WebSites\FiltersWebSite\Controllers\AuthorizeUserController.cs(31,10): DNX,Version=v4.5.1 error CS0104: 'AllowAnonymous' is an ambiguous reference between 'Microsoft.AspNet.Mvc.AllowAnonymousAttribute' and 'Microsoft.AspNet.Authorization.AllowAnonymousAttribute'
```

Issue #2994, opened 2015-08-18
Change order of generated input elements in asp-for checkbox helper


rblackman:
When rendering a checkbox input. There is a hidden input with a false value rendered after the checkbox input. See the example below.

```html
<input asp-for="Foo" type="checkbox"/>
<label asp-for="Foo">Foo?</label>
```

becomes

```html
<input data-val="true" data-val-required="The Foo field is required." id="Foo" name="Foo" type="checkbox" value="true" />
<input name="Foo" type="hidden" value="false" />
<label for="Foo">Foo?</label>
```

It is relatively common to style a label based on a checkboxes checked state e.g.
```css
input[type="checkbox"] + label {
    /* style a */
}
input[type="checkbox"]:checked + label {
     /* style b */
}
```
The addition of the hidden input breaks this sort of selector.

I would suggest, putting the hidden input first (since it doesn't seem to matter which order the inputs are rendered in). So the above code would become:
```html
<input name="Foo" type="hidden" value="false" />
<input data-val="true" data-val-required="The Foo field is required." id="Foo" name="Foo" type="checkbox" value="true" />
<label for="Foo">Foo?</label>
```

danroth27:
@dougbu @DamianEdwards

Eilon:
See http://stackoverflow.com/questions/2860940/why-does-the-checkboxfor-render-an-additional-input-tag-and-how-can-i-get-the-v for an explanation and the history of this.

Eilon:
Solution:

1. The `Html.CheckBox` HTML helpers and `<input type="checkbox">` tag helper will render *only* the checkbox inline
2. These helpers will *not* render the hidden input, and instead add it to the current form's context
3. The form HTML helper will render all the extra hidden inputs at the bottom of the form
4. A new "hidden" `<form>` tag helper (editorbrowsable=never) will capture all `<form>` elements regardless of attributes and its sole job is to render the extra hidden inputs at the bottom of the form
5. If the HTML/tag helper cannot find a form context then it will fall back to the current behavior of rendering the hidden input immediately after the checkbox

ajaybhargavb:
d40e6612a4e77f701fe99ceab40a238564fec238

PR #3311, opened 2015-10-12
Moving the generated hidden tags for checkbox to the end of the form

ajaybhargavb:
#2994 
@dougbu @pranavkm

### DIFF
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
@@ -278,7 +278,15 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
                 if (hiddenForCheckboxTag != null)
                 {
                     hiddenForCheckboxTag.TagRenderMode = renderingMode;
-                    output.Content.Append(hiddenForCheckboxTag);
+
+                    if (ViewContext.FormContext.CanRenderAtEndOfForm)
+                    {
+                        ViewContext.FormContext.EndOfFormContent.Add(hiddenForCheckboxTag);
+                    }
+                    else
+                    {
+                        output.Content.Append(hiddenForCheckboxTag);
+                    }
                 }
             }
         }
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/RenderAtEndOfFormTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/RenderAtEndOfFormTagHelper.cs
@@ -1,0 +1,61 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.ComponentModel;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Mvc.Rendering;
+using Microsoft.AspNet.Mvc.ViewFeatures;
+using Microsoft.AspNet.Razor.Runtime.TagHelpers;
+
+namespace Microsoft.AspNet.Mvc.TagHelpers
+{
+    /// <summary>
+    /// <see cref="ITagHelper"/> implementation targeting all form elements
+    /// to generate content before the form end tag.
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    [HtmlTargetElement("form")]
+    public class RenderAtEndOfFormTagHelper : TagHelper
+    {
+        public override int Order => -1000;
+
+        [HtmlAttributeNotBound]
+        [ViewContext]
+        public ViewContext ViewContext { get; set; }
+
+        /// <inheritdoc />
+        public override async Task ProcessAsync(TagHelperContext context, TagHelperOutput output)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            if (output == null)
+            {
+                throw new ArgumentNullException(nameof(output));
+            }
+
+            // Push the new FormContext.
+            ViewContext.FormContext = new FormContext
+            {
+                CanRenderAtEndOfForm = true
+            };
+
+            await context.GetChildContentAsync();
+
+            var formContext = ViewContext.FormContext;
+            if (formContext.HasEndOfFormContent)
+            {
+                foreach (var content in formContext.EndOfFormContent)
+                {
+                    output.PostContent.Append(content);
+                }
+            }
+
+            // Reset the FormContext
+            ViewContext.FormContext = null;
+        }
+    }
+}
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/MvcForm.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/MvcForm.cs
@@ -3,6 +3,8 @@
 
 using System;
 using Microsoft.AspNet.Mvc.ViewFeatures;
+using Microsoft.Extensions.WebEncoders;
+using Microsoft.Extensions.DependencyInjection;
 
 namespace Microsoft.AspNet.Mvc.Rendering
 {
@@ -19,9 +21,6 @@ namespace Microsoft.AspNet.Mvc.Rendering
             }
 
             _viewContext = viewContext;
-
-            // Push the new FormContext; GenerateEndForm() does the corresponding pop.
-            _viewContext.FormContext = new FormContext();
         }
 
         public void Dispose()
@@ -40,6 +39,7 @@ namespace Microsoft.AspNet.Mvc.Rendering
 
         protected virtual void GenerateEndForm()
         {
+            RenderEndOfFormContent();
             _viewContext.Writer.Write("</form>");
             _viewContext.FormContext = null;
         }
@@ -50,6 +50,34 @@ namespace Microsoft.AspNet.Mvc.Rendering
             {
                 _disposed = true;
                 GenerateEndForm();
+            }
+        }
+
+        private void RenderEndOfFormContent()
+        {
+            var formContext = _viewContext.FormContext;
+            if (formContext.HasEndOfFormContent)
+            {
+                var writer = _viewContext.Writer;
+                var htmlWriter = writer as HtmlTextWriter;
+
+                IHtmlEncoder htmlEncoder = null;
+                if (htmlWriter == null)
+                {
+                    htmlEncoder = _viewContext.HttpContext.RequestServices.GetRequiredService<IHtmlEncoder>();
+                }
+
+                foreach (var content in formContext.EndOfFormContent)
+                {
+                    if (htmlWriter == null)
+                    {
+                        content.WriteTo(writer, htmlEncoder);
+                    }
+                    else
+                    {
+                        htmlWriter.Write(content);
+                    }
+                }
             }
         }
     }
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/FormContext.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/FormContext.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using Microsoft.AspNet.Html.Abstractions;
 
 namespace Microsoft.AspNet.Mvc.ViewFeatures
 {
@@ -11,6 +12,7 @@ namespace Microsoft.AspNet.Mvc.ViewFeatures
         private readonly Dictionary<string, bool> _renderedFields =
             new Dictionary<string, bool>(StringComparer.Ordinal);
         private Dictionary<string, object> _formData;
+        private IList<IHtmlContent> _endOfFormContent;
 
         /// <summary>
         /// Property bag for any information you wish to associate with a &lt;form/&gt; in an
@@ -28,6 +30,25 @@ namespace Microsoft.AspNet.Mvc.ViewFeatures
                 return _formData;
             }
         }
+
+        public bool HasFormData => _formData != null;
+
+        public bool HasEndOfFormContent => _endOfFormContent != null;
+
+        public IList<IHtmlContent> EndOfFormContent
+        {
+            get
+            {
+                if (_endOfFormContent == null)
+                {
+                    _endOfFormContent = new List<IHtmlContent>();
+                }
+
+                return _endOfFormContent;
+            }
+        }
+
+        public bool CanRenderAtEndOfForm { get; set; }
 
         public bool RenderedField(string fieldName)
         {
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/HtmlHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/HtmlHelper.cs
@@ -272,12 +272,24 @@ namespace Microsoft.AspNet.Mvc.ViewFeatures
             FormMethod method,
             object htmlAttributes)
         {
+            // Push the new FormContext; MvcForm.GenerateEndForm() does the corresponding pop.
+            _viewContext.FormContext = new FormContext
+            {
+                CanRenderAtEndOfForm = true
+            };
+
             return GenerateForm(actionName, controllerName, routeValues, method, htmlAttributes);
         }
 
         /// <inheritdoc />
         public MvcForm BeginRouteForm(string routeName, object routeValues, FormMethod method, object htmlAttributes)
         {
+            // Push the new FormContext; MvcForm.GenerateEndForm() does the corresponding pop.
+            _viewContext.FormContext = new FormContext
+            {
+                CanRenderAtEndOfForm = true
+            };
+
             return GenerateRouteForm(routeName, routeValues, method, htmlAttributes);
         }
 
@@ -708,13 +720,24 @@ namespace Microsoft.AspNet.Mvc.ViewFeatures
                 isChecked,
                 htmlAttributes);
 
-            var hidden = _htmlGenerator.GenerateHiddenForCheckbox(ViewContext, modelExplorer, expression);
-            if (checkbox == null || hidden == null)
+            var hiddenForCheckboxTag = _htmlGenerator.GenerateHiddenForCheckbox(ViewContext, modelExplorer, expression);
+            if (checkbox == null || hiddenForCheckboxTag == null)
             {
                 return HtmlString.Empty;
             }
 
-            return new BufferedHtmlContent().Append(checkbox).Append(hidden);
+            var checkboxContent = new BufferedHtmlContent().Append(checkbox);
+
+            if (ViewContext.FormContext.CanRenderAtEndOfForm)
+            {
+                ViewContext.FormContext.EndOfFormContent.Add(hiddenForCheckboxTag);
+            }
+            else
+            {
+                checkboxContent.Append(hiddenForCheckboxTag);
+            }
+
+            return checkboxContent;
         }
 
         protected virtual string GenerateDisplayName(ModelExplorer modelExplorer, string expression)
--- a/test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.EmployeeList.html
+++ b/test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.EmployeeList.html
@@ -24,7 +24,7 @@ EmployeeName_0</textarea>
 </div>
 <div>
     <label class="employee" for="z0__Remote">Remote</label>
-    <input data-val="true" data-val-required="The Remote field is required." id="z0__Remote" name="[0].Remote" type="checkbox" value="true" /><input name="[0].Remote" type="hidden" value="false" />
+    <input data-val="true" data-val-required="The Remote field is required." id="z0__Remote" name="[0].Remote" type="checkbox" value="true" />
 </div>
 <div>
     <label class="employee" for="z0__OfficeNumber">OfficeNumber</label>
@@ -55,7 +55,7 @@ EmployeeName_1</textarea>
 </div>
 <div>
     <label class="employee" for="z1__Remote">Remote</label>
-    <input data-val="true" data-val-required="The Remote field is required." id="z1__Remote" name="[1].Remote" type="checkbox" value="true" /><input name="[1].Remote" type="hidden" value="false" />
+    <input data-val="true" data-val-required="The Remote field is required." id="z1__Remote" name="[1].Remote" type="checkbox" value="true" />
 </div>
 <div>
     <label class="employee" for="z1__OfficeNumber">OfficeNumber</label>
@@ -86,7 +86,7 @@ EmployeeName_2</textarea>
 </div>
 <div>
     <label class="employee" for="z2__Remote">Remote</label>
-    <input checked="checked" data-val="true" data-val-required="The Remote field is required." id="z2__Remote" name="[2].Remote" type="checkbox" value="true" /><input name="[2].Remote" type="hidden" value="false" />
+    <input checked="checked" data-val="true" data-val-required="The Remote field is required." id="z2__Remote" name="[2].Remote" type="checkbox" value="true" />
 </div>
 <div>
     <label class="employee" for="z2__OfficeNumber">OfficeNumber</label>
@@ -94,5 +94,5 @@ EmployeeName_2</textarea>
 <option>1002</option>
 </select>
 </div>            <input type="submit" />
-</form></body>
+<input name="[0].Remote" type="hidden" value="false" /><input name="[1].Remote" type="hidden" value="false" /><input name="[2].Remote" type="hidden" value="false" /></form></body>
 </html>
--- a/test/Microsoft.AspNet.Mvc.TagHelpers.Test/RenderAtEndOfFormTagHelperTest.cs
+++ b/test/Microsoft.AspNet.Mvc.TagHelpers.Test/RenderAtEndOfFormTagHelperTest.cs
@@ -1,0 +1,90 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Mvc.Rendering;
+using Microsoft.AspNet.Razor.Runtime.TagHelpers;
+using Xunit;
+
+namespace Microsoft.AspNet.Mvc.TagHelpers
+{
+    public class RenderAtEndOfFormTagHelperTest
+    {
+        public static TheoryData RenderAtEndOfFormTagHelperData
+        {
+            get
+            {
+                // tagBuilderList, expectedOutput
+                return new TheoryData<List<TagBuilder>, string>
+                {
+                    {
+                        new List<TagBuilder>
+                        {
+                            GetTagBuilder("input", "SomeName", "hidden", "false", TagRenderMode.SelfClosing)
+                        },
+                        @"<input name=""SomeName"" type=""hidden"" value=""false"" />"
+                    },
+                    {
+                        new List<TagBuilder>
+                        {
+                            GetTagBuilder("input", "SomeName", "hidden", "false", TagRenderMode.SelfClosing),
+                            GetTagBuilder("input", "SomeOtherName", "hidden", "false", TagRenderMode.SelfClosing)
+                        },
+                        @"<input name=""SomeName"" type=""hidden"" value=""false"" />" + 
+                        @"<input name=""SomeOtherName"" type=""hidden"" value=""false"" />"
+                    }
+                };
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(RenderAtEndOfFormTagHelperData))]
+        public async Task Process_AddsHiddenInputTag_FromEndOfFormContent(List<TagBuilder> tagBuilderList, string expectedOutput)
+        {
+            // Arrange
+            var viewContext = new ViewContext();
+            var tagHelperOutput = new TagHelperOutput(
+                tagName: "form",
+                attributes: new TagHelperAttributeList());
+
+            var tagHelperContext = new TagHelperContext(
+                Enumerable.Empty<IReadOnlyTagHelperAttribute>(),
+                new Dictionary<object, object>(),
+                "someId",
+                (useCachedResult) =>
+                {
+                    Assert.True(viewContext.FormContext.CanRenderAtEndOfForm);
+                    foreach (var item in tagBuilderList)
+                    {
+                        viewContext.FormContext.EndOfFormContent.Add(item);
+                    }
+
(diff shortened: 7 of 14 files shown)


src/Microsoft.AspNet.Mvc.Razor/TagHelpers/HiddenFormTagHelper.cs
```diff
@@ -0,0 +1,45 @@
...
+using Microsoft.AspNet.Mvc.Rendering;
+using Microsoft.AspNet.Mvc.ViewFeatures;
+using Microsoft.AspNet.Razor.Runtime.TagHelpers;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNet.Mvc.Razor.TagHelpers
+{
+    [EditorBrowsable(EditorBrowsableState.Never)]
```
dougbu: + docs

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/HiddenFormTagHelper.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            // Push the new FormContext.
+            ViewContext.FormContext = new FormContext();
+
+            await context.GetChildContentAsync();
+
+            var formContext = ViewContext.FormContext;
+            if (formContext.HasFormData && formContext.FormData.ContainsKey("hiddenForCheckBoxList"))
+            {
+                var hiddenForCheckBoxList = formContext.FormData["hiddenForCheckBoxList"] as List<TagBuilder>;
```
dougbu: use `TryGetValue()` to do the lookup just once

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/HiddenFormTagHelper.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        public override async Task ProcessAsync(TagHelperContext context, TagHelperOutput output)
+        {
+            // Push the new FormContext.
+            ViewContext.FormContext = new FormContext();
+
+            await context.GetChildContentAsync();
+
+            var formContext = ViewContext.FormContext;
+            if (formContext.HasFormData && formContext.FormData.ContainsKey("hiddenForCheckBoxList"))
```
dougbu: Use a constant key that's visible to this class, `MvcForm`, and `DefaultHtmlGenerator`. Suggest `nameof(DefaultHtmlGenerator.GenerateHiddenForCheckbox)` -- though I suggest (separately) that method should be renamed.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/HiddenFormTagHelper.cs
```diff
@@ -0,0 +1,45 @@
...
+            {
+                var hiddenForCheckBoxList = formContext.FormData["hiddenForCheckBoxList"] as List<TagBuilder>;
+                if (hiddenForCheckBoxList != null)
```
dougbu: Shouldn't this be a `Debug.Assert()`? What correct scenario involves an instance of another type in this `FormData` slot?

src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
```diff
@@ -268,7 +268,13 @@ private void GenerateCheckBox(ModelExplorer modelExplorer, TagHelperOutput outpu
...
                 {
                     hiddenForCheckboxTag.TagRenderMode = renderingMode;
-                    output.Content.Append(hiddenForCheckboxTag);
+
+                    var hiddenForCheckboxList = ViewContext.FormContext.FormData.ContainsKey("hiddenForCheckBoxList") ?
+                        ViewContext.FormContext.FormData["hiddenForCheckBoxList"] as List<TagBuilder> :
+                        new List<TagBuilder>();
+
+                    hiddenForCheckboxList.Add(hiddenForCheckboxTag);
+                    ViewContext.FormContext.FormData["hiddenForCheckBoxList"] = hiddenForCheckboxList;
```
dougbu: This code should not be duplicated between here and `HtmlHelper`. Instead move it into `DefaultHtmlGenerator.GenerateHiddenForCheckbox()`, change that method to return `void`, and rename it to something like `SaveHiddenForCheckbox()`. This class and `HtmlHelper` will then get simpler rather than more complicated.

src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
```diff
@@ -268,7 +268,13 @@ private void GenerateCheckBox(ModelExplorer modelExplorer, TagHelperOutput outpu
                 if (hiddenForCheckboxTag != null)
                 {
                     hiddenForCheckboxTag.TagRenderMode = renderingMode;
-                    output.Content.Append(hiddenForCheckboxTag);
+
+                    var hiddenForCheckboxList = ViewContext.FormContext.FormData.ContainsKey("hiddenForCheckBoxList") ?
+                        ViewContext.FormContext.FormData["hiddenForCheckBoxList"] as List<TagBuilder> :
```
dougbu: Either this cast should be unconditional or you need a `Debug.Assert()` before using the value on line 276.

Also don't do the lookup twice: Use `TryGetValue()`.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/MvcForm.cs
```diff
@@ -3,15 +3,18 @@
...
 namespace Microsoft.AspNet.Mvc.Rendering
 {
     public class MvcForm : IDisposable
     {
         private readonly ViewContext _viewContext;
+        private readonly IHtmlEncoder _htmlEncoder;
         private bool _disposed;
 
-        public MvcForm(ViewContext viewContext)
+        public MvcForm(ViewContext viewContext, IHtmlEncoder htmlEncoder)
```
dougbu: Ick. For the rare case when the `ViewContext.TextWriter` isn't an `HtmlTextWriter`, use `ViewContext.HttpContext.RequestServices.GetRequiredService<IHtmlEncoder>()`.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/MvcForm.cs
```diff
@@ -52,5 +54,30 @@ protected virtual void Dispose(bool disposing)
...
+
+        private void AddHiddenTagForCheckBox()
+        {
+            var writer = _viewContext.Writer;
+            var htmlWriter = writer as HtmlTextWriter;
+            var formContext = _viewContext.FormContext;
+            if (formContext.HasFormData && formContext.FormData.ContainsKey("hiddenForCheckBoxList"))
+            {
+                var hiddenForCheckBoxList = formContext.FormData["hiddenForCheckBoxList"] as List<TagBuilder>;
+                if (hiddenForCheckBoxList != null)
```
dougbu: See comments on similar code in `HiddenFormTagHelper`.

test/Microsoft.AspNet.Mvc.Razor.Test/TagHelpers/HiddenFormTagHelperTest.cs
```diff
@@ -0,0 +1,87 @@
...
+using Microsoft.AspNet.Razor.Runtime.TagHelpers;
+using Moq;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Microsoft.AspNet.Mvc.Razor.Test.TagHelpers
+{
+    public class HiddenFormTagHelperTest
```
dougbu: Instead test with a `CommonTestEncoder` instance. Should confirm double-encoding doesn't happen in unit tests.

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.Razor.Host/MvcRazorHost.cs
```diff
@@ -42,6 +42,10 @@ public class MvcRazorHost : RazorEngineHost, IMvcRazorHost
             {
                 LookupText = "Microsoft.AspNet.Mvc.Razor.TagHelpers.UrlResolutionTagHelper, Microsoft.AspNet.Mvc.Razor"
             },
+            new AddTagHelperChunk
```
pranavkm: Spoke to @Eilon - instead of doing this, move it to `Microsoft.AspNet.Mvc.TagHelpers`. Users will import it via `_ViewImports`.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/HiddenFormTagHelper.cs
pranavkm: Why?

ajaybhargavb: So that it doesn't show up in intellisense.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/HiddenFormTagHelper.cs
```diff
@@ -0,0 +1,45 @@
...
+{
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    [HtmlTargetElement("form")]
+    public class HiddenFormTagHelper : TagHelper
+    {
+        [HtmlAttributeNotBound]
+        [ViewContext]
+        public ViewContext ViewContext { get; set; }
+
+        public override async Task ProcessAsync(TagHelperContext context, TagHelperOutput output)
```
pranavkm: Null checks

src/Microsoft.AspNet.Mvc.Razor.Host/MvcRazorHost.cs
dougbu: @pranavkm @Eilon if we do this, `@Html.CheckBox()` and `@Html.CheckBoxFor()` will not work correctly inside a regular `<form>` element.

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/HiddenFormTagHelper.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            // Push the new FormContext.
+            ViewContext.FormContext = new FormContext();
```
pranavkm: I know nested forms aren't a supported browser feature, but what's the behavior when you do make a malformed page?

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/FormContext.cs
```diff
@@ -29,6 +29,14 @@ public class FormContext
             }
         }
 
+        public bool HasFormData
```
pranavkm: `public bool HasFormData => _formData != null;`

pranavkm:
:watch:

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/HiddenFormTagHelper.cs
dougbu: Same thing it's always been -- information added in the outer `<form>` element is lost. `FormContext` has always behaved this way except we've had a bug in that the `FormTagHelper` never set it.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/FormContext.cs
dougbu: @pranavkm why? Nothing in our guidelines requires Pranav points and nothing else in this class uses newer C# features.

pranavkm: It's less verbose. No reason not to adopt features when they are the better alternative.

ajaybhargavb:
Updated.

src/Microsoft.AspNet.Mvc.TagHelpers/HiddenFormTagHelper.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNet.Mvc.TagHelpers
+{
+    /// <summary>
+    /// <see cref="ITagHelper"/> implementation targeting all form elements to generate hidden tag for checkboxes.
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Advanced)]
+    [HtmlTargetElement("form")]
+    public class HiddenFormTagHelper : TagHelper
```
dougbu: Need to be more general w/ these descriptions and names. Check boxes are one use of the infrastructure this helper provides. Perhaps `... to generate deferred content before the &lt;/form&gt; end tag.` and `RenderBeforeFormEndTagTagHelper`?

src/Microsoft.AspNet.Mvc.TagHelpers/HiddenFormTagHelper.cs
```diff
@@ -0,0 +1,47 @@
...
+    {
+        [HtmlAttributeNotBound]
+        [ViewContext]
+        public ViewContext ViewContext { get; set; }
+
+        public override async Task ProcessAsync(TagHelperContext context, TagHelperOutput output)
```
dougbu: At least `/// <inheritdoc />`

src/Microsoft.AspNet.Mvc.TagHelpers/HiddenFormTagHelper.cs
```diff
@@ -0,0 +1,47 @@
...
+    /// <summary>
+    /// <see cref="ITagHelper"/> implementation targeting all form elements to generate hidden tag for checkboxes.
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Advanced)]
+    [HtmlTargetElement("form")]
+    public class HiddenFormTagHelper : TagHelper
+    {
+        [HtmlAttributeNotBound]
+        [ViewContext]
+        public ViewContext ViewContext { get; set; }
```
dougbu: Add an override of the `Order` property to reliably run just before or after the regular `FormTagHelper`.

src/Microsoft.AspNet.Mvc.TagHelpers/HiddenFormTagHelper.cs
```diff
@@ -0,0 +1,47 @@
...
+            // Push the new FormContext.
+            ViewContext.FormContext = new FormContext
+            {
+                CanRenderAtEndOfForm = true
+            };
+
+            await context.GetChildContentAsync();
+
+            var formContext = ViewContext.FormContext;
+            if (formContext.HasRenderAtEndOfFormContent)
```
dougbu: Is this the name @Eilon liked? Not just `HasEndOfFormContent`?

src/Microsoft.AspNet.Mvc.TagHelpers/HiddenFormTagHelper.cs
dougbu: Actually, fine if order relative to `FormTagHelper` is not predetermined. `Order` should be `-1000` to keep execution away from user tag helpers.

src/Microsoft.AspNet.Mvc.TagHelpers/HiddenFormTagHelper.cs
```diff
@@ -0,0 +1,47 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNet.Mvc.Rendering;
+using Microsoft.AspNet.Mvc.ViewFeatures;
+using Microsoft.AspNet.Razor.Runtime.TagHelpers;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Threading.Tasks;
```
dougbu: `System` namespaces first. Check your VS options and "remove and sort usings".

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/MvcForm.cs
```diff
@@ -3,6 +3,9 @@
 
 using System;
 using Microsoft.AspNet.Mvc.ViewFeatures;
+using System.Collections.Generic;
+using Microsoft.Extensions.WebEncoders;
+using Microsoft.Extensions.DependencyInjection;
```
dougbu: Sort these.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/MvcForm.cs
```diff
@@ -52,5 +53,27 @@ protected virtual void Dispose(bool disposing)
                 GenerateEndForm();
             }
         }
+
+        private void RenderEndOfFormContent()
+        {
+            var writer = _viewContext.Writer;
+            var htmlWriter = writer as HtmlTextWriter;
```
dougbu: Move these two declarations closer to first use -- inside the first `if`.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/MvcForm.cs
```diff
@@ -52,5 +53,27 @@ protected virtual void Dispose(bool disposing)
...
+            var formContext = _viewContext.FormContext;
+            if (formContext.HasRenderAtEndOfFormContent)
+            {
+                foreach (var hiddenTag in formContext.RenderAtEndOfFormContent)
+                {
+                    if (htmlWriter == null)
+                    {
+                        var htmlEncoder = _viewContext.HttpContext.RequestServices.GetRequiredService<IHtmlEncoder>();
```
dougbu: Compiler won't optimize this and will execute the expression every time through the loop. Move this declaration right after `var htmlWriter` and leave it `null` if `htmlWriter` is not `null`.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/FormContext.cs
```diff
@@ -29,6 +31,8 @@ public class FormContext
             }
         }
 
+        public bool HasFormData => _formData != null;
+
         public bool RenderedField(string fieldName)
```
dougbu: Properties before methods. The two `RenderedField()` overloads should remain at the bottom of this file.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/HtmlHelper.cs
```diff
@@ -272,12 +272,24 @@ public IHtmlContent AntiForgeryToken()
             FormMethod method,
             object htmlAttributes)
         {
+            // Push the new FormContext; GenerateEndForm() does the corresponding pop.
```
dougbu: Update these comments to include location of `GenerateEndForm()`. That method isn't in this class.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/HiddenFormTagHelperTest.cs
```diff
@@ -0,0 +1,92 @@
...
+                        @"<input name=""SomeName"" type=""hidden"" value=""false"" />"
+                    },
+                    {
+                        new List<TagBuilder>
+                        {
+                            GetTagBuilder("input", "SomeName", "hidden", "false", TagRenderMode.SelfClosing),
+                            GetTagBuilder("input", "SomeOtherName", "hidden", "false", TagRenderMode.SelfClosing)
+                        },
+                        @"<input name=""SomeName"" type=""hidden"" value=""false"" />" +
+                @"<input name=""SomeOtherName"" type=""hidden"" value=""false"" />"
```
dougbu: Line looks odd. Is this out-dented for a reason?

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperCheckboxTest.cs
```diff
@@ -128,6 +128,26 @@ public void CheckBoxCheckedWithOnlyName_GeneratesExpectedValue()
...
+                @"name=""HtmlEncode[[Property1]]"" type=""HtmlEncode[[checkbox]]"" " +
+                @"value=""HtmlEncode[[true]]"" />");
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper(GetTestModelViewData());
+            helper.ViewContext.FormContext.CanRenderAtEndOfForm = true;
+
+            // Act
+            var html = helper.CheckBox("Property1", isChecked: true, htmlAttributes: null);
+
+            // Assert
+            Assert.Equal(expected, HtmlContentUtilities.HtmlContentToString(html));
```
dougbu: Since this is a unit test, add assertions about `FormContext.HasRenderAtEndOfFormContent` and the `RenderAtEndOfFormContent` itself.

src/Microsoft.AspNet.Mvc.TagHelpers/HiddenFormTagHelper.cs
dougbu: Missing `null` check for `output`. (`context` isn't used so it's fine to ignore.) Know other tag helpers lack the checks but let's not expand that error.

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.TagHelpers/HiddenFormTagHelper.cs
```diff
@@ -0,0 +1,47 @@
...
+
+namespace Microsoft.AspNet.Mvc.TagHelpers
+{
+    /// <summary>
+    /// <see cref="ITagHelper"/> implementation targeting all form elements to generate hidden tag for checkboxes.
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Advanced)]
```
NTaylorMullen: `EditorBrowsableState.Advanced` wont do anything. Is there a follow up bug to make Razor treat `Advanced` like `Never`?

src/Microsoft.AspNet.Mvc.TagHelpers/HiddenFormTagHelper.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        public override async Task ProcessAsync(TagHelperContext context, TagHelperOutput output)
+        {
+            // Push the new FormContext.
+            ViewContext.FormContext = new FormContext
+            {
+                CanRenderAtEndOfForm = true
+            };
+
+            await context.GetChildContentAsync();
```
NTaylorMullen: When looking at this, I discovered we have a pretty significant design-flaw: https://github.com/aspnet/Razor/issues/571

src/Microsoft.AspNet.Mvc.TagHelpers/HiddenFormTagHelper.cs
Eilon: This should just be Never.

src/Microsoft.AspNet.Mvc.TagHelpers/HiddenFormTagHelper.cs
Eilon: Per our discussion yesterday, this shouldn't say anything at all about anything being deferred. It's really entirely just about rendering stuff before the end of the form.

src/Microsoft.AspNet.Mvc.TagHelpers/HiddenFormTagHelper.cs
Eilon: I think `HasEndOfFormContent` is a nice, short name.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/FormContext.cs
```diff
@@ -29,6 +31,8 @@ public class FormContext
             }
         }
 
+        public bool HasFormData => _formData != null;
```
Eilon: This property is no longer needed, right?

ajaybhargavb: Yes. Spoke to Doug about this and we thought it is a useful perf optimization.

dougbu: ... for _customers_ using `FormData` (while we're here).

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperFormTest.cs
```diff
@@ -10,6 +10,9 @@
 using Microsoft.AspNet.Mvc.ViewFeatures;
 using Moq;
 using Xunit;
+using System;
+using Microsoft.Extensions.WebEncoders;
+using Microsoft.Extensions.WebEncoders.Testing;
```
Eilon: Sort & Remove in all the files :trollface:

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/FormContext.cs
Eilon: Yeah I'm sure it gets used a lot :smile: But fine, leave it.

Eilon:
Looking really nice!

src/Microsoft.AspNet.Mvc.TagHelpers/HiddenFormTagHelper.cs
dougbu: My change (#3321) to fix other tag helpers will `null`-check both parameters consistently, whether or not they're used. Do the same here.

test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.EmployeeList.html
```diff
@@ -94,5 +94,5 @@
 <option>1002</option>
 </select>
 </div>            <input type="submit" />
-</form></body>
+<input name="[0].Remote" type="hidden" value="false" /><input name="[1].Remote" type="hidden" value="false" /><input name="[2].Remote" type="hidden" value="false" /></form></body>
```
NTaylorMullen: Not important here but at some point we should probably change the rendering to output void input tags.

dougbu: @NTaylorMullen we test both with and without the slash. Are you suggesting we test only without?

NTaylorMullen: Nooo, testing isn't what I'm referring to. I'm saying that we should always generate without.

dougbu: Thought we decided long ago not to mess w/ how the user wrote an element in the `.cshtml` file?

NTaylorMullen: These aren't what the user wrote though. These are generated to pass data to the server on post.

dougbu: Ah you're talking specifically about revisiting our decision to write these additional `<input/>` (or `<input>`) elements identically to the check boxes. Right?

NTaylorMullen: correct

dougbu: Well, I'm curious what has changed since we made that decision but let's take this offline. Unrelated to @ajaybhargavb's work here.

dougbu:
@ajaybhargavb is this updated and ready for another review round?

ajaybhargavb:
@dougbu yes. Forgot to mention.

ajaybhargavb:
Updated.

src/Microsoft.AspNet.Mvc.TagHelpers/RenderAtEndOfFormTagHelper.cs
```diff
@@ -0,0 +1,67 @@
...
+            {
+                CanRenderAtEndOfForm = true
+            };
+
+            await context.GetChildContentAsync();
+
+            var formContext = ViewContext.FormContext;
+            if (formContext.HasEndOfFormContent)
+            {
+                foreach (var hiddenTag in formContext.EndOfFormContent)
```
dougbu: `hiddenTag` -> `content`

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/MvcForm.cs
```diff
@@ -52,5 +52,33 @@ protected virtual void Dispose(bool disposing)
...
+                var writer = _viewContext.Writer;
+                var htmlWriter = writer as HtmlTextWriter;
+
+                IHtmlEncoder htmlEncoder = null;
+                if (htmlWriter == null)
+                {
+                    htmlEncoder = _viewContext.HttpContext.RequestServices.GetRequiredService<IHtmlEncoder>();
+                }
+
+                foreach (var hiddenTag in formContext.EndOfFormContent)
```
dougbu: `hiddenTag` -> `content`

dougbu:
:shipit: with just a couple of teensy changes

test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.EmployeeList.html
Eilon: We should be generating `<self-closing />` tags to be XHTML compliant, which some people care about. It's never wrong, is it?

test/Microsoft.AspNet.Mvc.TagHelpers.Test/RenderAtEndOfFormTagHelperTest.cs
```diff
@@ -0,0 +1,90 @@
...
+                        @"<input name=""SomeName"" type=""hidden"" value=""false"" />" + 
+                        @"<input name=""SomeOtherName"" type=""hidden"" value=""false"" />"
+                    }
+                };
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(RenderAtEndOfFormTagHelperData))]
+        public async Task Process_AddsHiddenInputTag_FromFormData(List<TagBuilder> tagBuilderList, string expectedOutput)
```
Eilon: Update test name to not say `FormData`

Eilon:
Looking great, :shipit: after tiny comments.

Issue #3250, opened 2015-10-03
Routes values are being added in duplicate


mdmoura:
Hello,

I created a TagHelper and needed to access the current route so I ended up with:

```c#
    [TargetElement("a", Attributes = RouteHasName)]
    public class RouteActiveTagHelper : TagHelper {

      private const String RouteHasName = "route-has";

      private IActionContextAccessor _actionContextAccessor;
      private IUrlHelper _urlHelper;   

      [HtmlAttributeName(RouteHasName)]
      public String RouteHas { get; set; }
  
      public RouteActiveTagHelper(IActionContextAccessor actionContextAccessor, IUrlHelper urlHelper) {
        _actionContextAccessor = actionContextAccessor;
        _urlHelper = urlHelper;
    } // RouteActiveTagHelper

    public override void Process(TagHelperContext context, TagHelperOutput output) {

      var values = _actionContextAccessor.ActionContext.RouteData.Values;

      String route = _urlHelper.RouteUrl(values.Distinct());  

      // Removed remaining code for sake of simplicity

      } // Process
    } // RouteActiveTagHelper
```

When I create the route I get the error "ArgumentException: An item with the same key has already been added." on every route which is defined using Attribute Routing.

To fix this I needed to apply Distinct to values before creating a route.

Is there any reason for having duplicated objects in route values? Is this a bug?

If this is the expected result is there a better way to get the current route values?

Note: I get the same error using ViewContext ...

Thank You,
Miguel

rynowak:
This seems like a bug. Can you include a callstack for the exception?

As a workaround, if you're just trying to link to the 'current values' you don't need to pass them in to `RouteUrl`.
``
_urlHelper.RouteUrl(new { })`
``

mdmoura:
Here is the stack. Does it help?

		ArgumentException: An item with the same key has already been added.
			System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
				System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
				System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)
				Microsoft.AspNet.Mvc.Routing.InnerAttributeRoute.GenerateVirtualPath(VirtualPathContext context, AttributeRouteLinkGenerationEntry entry)
				Microsoft.AspNet.Mvc.Routing.InnerAttributeRoute.GetVirtualPath(VirtualPathContext context)
[...]

PR #3313, opened 2015-10-13
Remove RouteGroup from RouteValues

ajaybhargavb:
#3250 
@rynowak @pranavkm 

Currently this does not break any tests. Will add tests once the fix is accepted.


Eilon:
Talked to @ajaybhargavb , there might be a better fix for this.

ajaybhargavb:
Updated with new fix and added a unit test.

src/Microsoft.AspNet.Mvc.Core/Infrastructure/MvcRouteHandler.cs
```diff
@@ -66,6 +67,7 @@ public VirtualPathData GetVirtualPath(VirtualPathContext context)
             // without affecting something upstream.
             var oldRouteData = context.RouteData;
             var newRouteData = new RouteData(oldRouteData);
+            newRouteData.Values.Remove(AttributeRouting.RouteGroupKey);
```
rynowak: Might want to group this with the next blok and make a omment that this is done to 'simulate' the output of onventional routing.

rynowak:
:shipit:

PR #3316, opened 2015-10-13
Modify ViewStartUtility to generate always view paths with forward sl…

### LINKED ISSUE #3294: Modify ViewStartUtility to generate always view paths with forward slashes
Related to https://github.com/aspnet/FileSystem/issues/129. From Mvc's end we could modify the path generation for ViewStart and ViewImports to always have forward slashes. These would bring it in line with the templated view and area paths we specify in RazorViewEngine.

pranavkm:
…ashes

Fixes #3294

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor.Host/ViewHierarchyUtility.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor.Host/ViewHierarchyUtility.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text;
 
 namespace Microsoft.AspNet.Mvc.Razor
 {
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNet.Mvc.Razor.Host/ViewHierarchyUtility.cs
```diff
@@ -77,18 +78,31 @@ private static IEnumerable<string> GetHierarchicalPath(string relativePath, stri
...
                 }
             }
 
+            var builder = new StringBuilder(relativePath);
+            builder.Replace('\\', '/');
+
+            if (builder.Length > 0 && builder[0] != '/')
+            {
+                builder.Insert(0, '/');
+            }
```
rynowak: You may want to do this first to avoid all the blitting needed with an `Insert(0...)` - it's probably not significant though.  Is this a common case or uncommon?

src/Microsoft.AspNet.Mvc.Razor.Host/ViewHierarchyUtility.cs
```diff
@@ -77,18 +78,31 @@ private static IEnumerable<string> GetHierarchicalPath(string relativePath, stri
...
             {
-                relativePath = Path.GetDirectoryName(relativePath);
-                var path = Path.Combine(relativePath, fileName);
-                locations.Add(path);
+                if (builder[index] == '/')
+                {
+                    builder.Length = index + 1;
+                    builder.Append(fileName);
+
+                    locations.Add(builder.ToString());
```
rynowak: is this change being done for perf? Seems like a smart way to do this

src/Microsoft.AspNet.Mvc.Razor.Host/ViewHierarchyUtility.cs
rynowak: On second thought - we do we normalize in favor of adding the `/`. That just seems off from an xplat pov since it looks like an absolute path.

pranavkm: I was trying to get it to look like all the other paths we generate by default - https://github.com/aspnet/Mvc/pull/3316/files#diff-91b20801581acc92340a5d67c51d172dL2. We could live with out it, but we assume all `/` rooted view paths are app rooted and not absolute paths.

src/Microsoft.AspNet.Mvc.Razor.Host/ViewHierarchyUtility.cs
pranavkm: It seemed easy to do it this way rather than doing string concats. I guess there's some perf benefit here but it might become negligible once we figure out some way to cache these.

Issue #3138, opened 2015-09-15
content-type check seems to be wrong in InputFormatter


andresmoschini:
From [InputFormatter.cs](https://github.com/aspnet/Mvc/blob/6.0.0-beta7/src/Microsoft.AspNet.Mvc.Core/InputFormatter.cs#L71)

```csharp
public virtual bool CanRead(InputFormatterContext context)
{
    // . . .
    return SupportedMediaTypes
        .Any(supportedMediaType => supportedMediaType.IsSubsetOf(requestContentType));
}
```

If I am not wrong, to be able to read a request, `requestContentType` should be a subset of any of `InputFormatter`'s `SupportedMediaTypes`.

andresmoschini:
For example, I set:

```csharp
formatter.SupportedMediaTypes.Add(new MediaTypeHeaderValue("*/*"));
```

and I get `Unsupported content type 'text/plain; charset=UTF-8'.`.

But it works if I set: 

```csharp
formatter.SupportedMediaTypes.Add(new MediaTypeHeaderValue("text/plain"));
```

Is this behavior expected?

dougbu:
It's possible there's a bug in how `IsSubsetOf()` handles `*/*` however the real answer for your scenario is to override `CanRead()` to return `true` since that seems to be what you want.

dougbu:
Put another way, `InputFormatter` handles a specific selection of concrete formatters. A catch-all formatter should probably implement `IInputFormatter` directly.

andresmoschini:
Thanks @dougbu, but it was only an example.

I think that an `InputFormatter` should be capable to read all request with `content-type` that are subtypes of `SupportedMediaTypes`. Am I missing something?

dougbu:
Do you have a concrete example that fails which doesn't involve `*/*`? Better yet do you have a small repro project we could use to investigate what you're seeing -- even if with `*/*`?

Note what you're doing with the catch-all formatter (i.e. that accepts `*/*`) _should_ work. My alternate implementation was based on the semantics you need rather than a thought your issue was invalid.

andresmoschini:
Thanks for your help @dougbu, since I need to ignore request content type, in my project I will override `CanRead()` as you mention.

BUT, I am pretty sure that I am not understanding how to _InputFormaters_ deal with `content-type` header, OR there is a mistake in the implementation.

[...]

PR #3317, opened 2015-10-13
Correct polarity of MediaTypeHeaderValue.IsSubsetOf()` checks and remove one conneg fallback

dougbu:
- aspnet/Mvc#3138 part 2/2
- request's Content-Type header must be a subset of what an `IInputFormatter` can consume
  - `[Consumes]` is similar
- what an `IOutputFormatter` produces must be a subset of the request's Accept header
  - `FormatFilter` and `ObjectResult` are similar
- `ObjectResult` no longer falls back to `Content-Type` header if no `Accept` value is acceptable
- left `WebApiCompatShim` code alone for consistency with down-level `System.Net.Http.Formatting`
- correct tests to match new behaviour
  - do not test `Accept` values containing a `charset` parameter; that case is not valid

nits:
- add `InputFormatterTests`
- add / update comments and doc comments
- correct xUnit attributes in `ActionResultTest`; odd it doesn't show up in command-line runs

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Abstractions/project.json
+++ b/src/Microsoft.AspNet.Mvc.Abstractions/project.json
@@ -25,7 +25,8 @@
     "Microsoft.Extensions.PropertyHelper.Sources": {
       "version": "1.0.0-*",
       "type": "build"
-    }
+    },
+    "Microsoft.Net.Http.Headers": "1.0.0-*"
   },
   "frameworks": {
     "dnx451": { },
--- a/src/Microsoft.AspNet.Mvc.Core/ConsumesAttribute.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ConsumesAttribute.cs
@@ -56,10 +56,10 @@ namespace Microsoft.AspNet.Mvc
                 MediaTypeHeaderValue requestContentType = null;
                 MediaTypeHeaderValue.TryParse(context.HttpContext.Request.ContentType, out requestContentType);
 
-                // Only execute if this is the last filter before calling the action.
-                // This ensures that we only run the filter which is closest to the action.
+                // Confirm the request's content type is more specific than a media type this action supports e.g. OK
+                // if client sent "text/plain" data and this action supports "text/*".
                 if (requestContentType != null &&
-                    !ContentTypes.Any(contentType => contentType.IsSubsetOf(requestContentType)))
+                    !ContentTypes.Any(contentType => requestContentType.IsSubsetOf(contentType)))
                 {
                     context.Result = new UnsupportedMediaTypeResult();
                 }
@@ -102,7 +102,9 @@ namespace Microsoft.AspNet.Mvc
                 return !isActionWithoutConsumeConstraintPresent;
             }
 
-            if (ContentTypes.Any(c => c.IsSubsetOf(requestContentType)))
+            // Confirm the request's content type is more specific than a media type this action supports e.g. OK
+            // if client sent "text/plain" data and this action supports "text/*".
+            if (ContentTypes.Any(contentType => requestContentType.IsSubsetOf(contentType)))
             {
                 return true;
             }
--- a/src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
@@ -62,7 +62,7 @@ namespace Microsoft.AspNet.Mvc.Formatters
 
             var contentType = context.HttpContext.Request.ContentType;
             MediaTypeHeaderValue requestContentType;
-            if (!MediaTypeHeaderValue.TryParse(contentType, out requestContentType) || requestContentType == null)
+            if (!MediaTypeHeaderValue.TryParse(contentType, out requestContentType))
             {
                 return false;
             }
--- a/src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
@@ -62,12 +62,17 @@ namespace Microsoft.AspNet.Mvc.Formatters
 
             var contentType = context.HttpContext.Request.ContentType;
             MediaTypeHeaderValue requestContentType;
-            if (!MediaTypeHeaderValue.TryParse(contentType, out requestContentType))
+            if (!MediaTypeHeaderValue.TryParse(contentType, out requestContentType) || requestContentType == null)
             {
                 return false;
             }
 
-            return SupportedMediaTypes.Any(supportedMediaType => supportedMediaType.IsSubsetOf(requestContentType));
+            // Confirm the request's content type is more specific than a media type this formatter supports e.g. OK if
+            // client sent "text/plain" data and this formatter supports "text/*".
+            return SupportedMediaTypes.Any(supportedMediaType =>
+            {
+                return requestContentType.IsSubsetOf(supportedMediaType);
+            });
         }
 
         /// <summary>
--- a/src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
@@ -203,7 +203,7 @@ namespace Microsoft.AspNet.Mvc.Formatters
             // Copy the media type as we don't want it to affect the next request
             selectedMediaType = selectedMediaType.Copy();
 
-            // Note text-based media types will use an encoding/charset - binary formats just ignore it. We want to
+            // Note: Text-based media types will use an encoding/charset - binary formats just ignore it. We want to
             // make this class work with media types that use encodings, and those that don't.
             //
             // The default implementation of SelectCharacterEncoding will read from the list of SupportedEncodings
--- a/src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
@@ -77,6 +77,8 @@ namespace Microsoft.AspNet.Mvc.Formatters
             {
                 List<MediaTypeHeaderValue> mediaTypes = null;
 
+                // Confirm this formatter supports a more specific media type than requested e.g. OK if "text/*"
+                // requested and formatter supports "text/plain". Treat contentType like it came from an Accept header.
                 foreach (var mediaType in _supportedMediaTypes)
                 {
                     if (mediaType.IsSubsetOf(contentType))
@@ -119,8 +121,7 @@ namespace Microsoft.AspNet.Mvc.Formatters
                 {
                     var requestCharset = requestContentType.Charset;
                     encoding = SupportedEncodings.FirstOrDefault(
-                                                            supportedEncoding =>
-                                                                requestCharset.Equals(supportedEncoding.WebName));
+                        supportedEncoding => requestCharset.Equals(supportedEncoding.WebName));
                 }
             }
 
@@ -151,10 +152,11 @@ namespace Microsoft.AspNet.Mvc.Formatters
             }
             else
             {
-                // Since supportedMedia Type is going to be more specific check if supportedMediaType is a subset
-                // of the content type which is typically what we get on acceptHeader.
-                mediaType = SupportedMediaTypes
-                                  .FirstOrDefault(supportedMediaType => supportedMediaType.IsSubsetOf(contentType));
+                // Confirm this formatter supports a more specific media type than requested e.g. OK if "text/*"
+                // requested and formatter supports "text/plain". contentType is typically what we got in an Accept
+                // header.
+                mediaType = SupportedMediaTypes.FirstOrDefault(
+                    supportedMediaType => supportedMediaType.IsSubsetOf(contentType));
             }
 
             if (mediaType != null)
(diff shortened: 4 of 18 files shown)


src/Microsoft.AspNet.Mvc.Abstractions/project.json
```diff
@@ -25,7 +25,8 @@
     "Microsoft.Extensions.PropertyHelper.Sources": {
       "version": "1.0.0-*",
       "type": "build"
-    }
+    },
+    "Microsoft.Net.Http.Headers": "1.0.0-*"
```
dougbu: Ignore this change from the first commit. Did this only to ensure build picks up the local changes to `HttpAbstractions`.

test/Microsoft.AspNet.Mvc.FunctionalTests/RespectBrowserAcceptHeaderTests.cs
```diff
@@ -21,7 +21,7 @@ public RespectBrowserAcceptHeaderTests(MvcTestFixture<FormatterWebSite.Startup>
         public HttpClient Client { get; }
 
         [Theory]
-        [InlineData("application/xml,*/*;0.2")]
+        [InlineData("application/xml,*/*;q=0.2")]
```
dougbu: syntax was incorrect

dougbu:
03625c3

pranavkm:
No :shipit:s?

Eilon:
I talked to @dougbu this does still need to be reviewed.

src/Microsoft.AspNet.Mvc.Core/ConsumesAttribute.cs
```diff
@@ -56,10 +56,10 @@ public void OnResourceExecuting(ResourceExecutingContext context)
                 MediaTypeHeaderValue requestContentType = null;
                 MediaTypeHeaderValue.TryParse(context.HttpContext.Request.ContentType, out requestContentType);
 
-                // Only execute if this is the last filter before calling the action.
```
rynowak: Does this comment go with another block in this method? Your new comment is great, but methinks the previous comment wandered away from home.

src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
```diff
@@ -62,12 +62,17 @@ public virtual bool CanRead(InputFormatterContext context)
 
             var contentType = context.HttpContext.Request.ContentType;
             MediaTypeHeaderValue requestContentType;
-            if (!MediaTypeHeaderValue.TryParse(contentType, out requestContentType))
+            if (!MediaTypeHeaderValue.TryParse(contentType, out requestContentType) || requestContentType == null)
```
rynowak: How can this happen? What are the conditions where parsing will succeed and return null?

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
```diff
@@ -201,11 +203,11 @@ public virtual void WriteResponseHeaders(OutputFormatterContext context)
             // Copy the media type as we don't want it to affect the next request
             selectedMediaType = selectedMediaType.Copy();
 
-            // Not text-based media types will use an encoding/charset - binary formats just ignore it. We want to
+            // Note text-based media types will use an encoding/charset - binary formats just ignore it. We want to
```
rynowak: `Note: `

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
```diff
@@ -237,21 +239,17 @@ private Encoding MatchAcceptCharacterEncoding(IList<StringWithQualityHeaderValue
...
                     if (!string.IsNullOrWhiteSpace(charset))
                     {
-                        var encoding = SupportedEncodings.FirstOrDefault(
-                                                        supportedEncoding =>
-                                                            charset.Equals(supportedEncoding.WebName,
-                                                                           StringComparison.OrdinalIgnoreCase) ||
-                                                            charset.Equals("*", StringComparison.Ordinal));
+                        var encoding = SupportedEncodings.FirstOrDefault(supportedEncoding =>
+                            charset.Equals(supportedEncoding.WebName, StringComparison.OrdinalIgnoreCase) ||
+                            charset.Equals("*", StringComparison.Ordinal));
```
rynowak: Is there a constant somewhere for this?

src/Microsoft.AspNet.Mvc.Core/ObjectResult.cs
```diff
@@ -122,25 +116,12 @@ public override Task ExecuteResultAsync(ActionContext context)
...
-                }
+                selectedFormatter = SelectFormatterUsingSortedAcceptHeaders(
+                    formatterContext,
+                    formatters,
+                    sortedAcceptHeaderMediaTypes);
 
-                // 2. No formatter was found based on accept headers, fall back on request Content-Type header.
-                if (selectedFormatter == null && requestContentType != null)
-                {
-                    selectedFormatter = SelectFormatterUsingAnyAcceptableContentType(
```
rynowak: :dancers:

src/Microsoft.AspNet.Mvc.Core/ObjectResult.cs
```diff
@@ -122,25 +116,12 @@ public override Task ExecuteResultAsync(ActionContext context)
...
-                                                                                formatterContext,
-                                                                                formatters,
-                                                                                new[] { requestContentType });
-                }
-
-                // 3. No formatter was found based on Accept and request Content-Type headers, so
-                // fallback on type based match.
+                // 2. No formatter was found based on Accept header. Fallback to type-based match.
```
rynowak: A "type-based match" is really a specialized scenario, and isn't a very usual thing outside of things like `HttpResponseMessage` or `Stream`. I'd say something like "let the formatters choose any content type they want"

src/Microsoft.AspNet.Mvc.Core/ObjectResult.cs
```diff
@@ -177,9 +158,9 @@ public override Task ExecuteResultAsync(ActionContext context)
                     // In any of these cases, if the user has specified content types,
                     // do a last effort to find a formatter which can write any of the user specified content type.
                     selectedFormatter = SelectFormatterUsingAnyAcceptableContentType(
-                                                                        formatterContext,
-                                                                        formatters,
-                                                                        ContentTypes);
+                        formatterContext,
+                        formatters,
+                        ContentTypes);
```
rynowak: As formatting changes go, this one deserves at least a round of applause :clap:

test/Microsoft.AspNet.Mvc.Core.Test/Formatters/InputFormatterTest.cs
```diff
@@ -0,0 +1,292 @@
...
+using Microsoft.AspNet.Http.Internal;
+using Microsoft.AspNet.Mvc.ModelBinding;
+using Microsoft.Net.Http.Headers;
+using Xunit;
+
+namespace Microsoft.AspNet.Mvc.Formatters
+{
+    public class InputFormatterTest
+    {
+        private class CatchAllFormatter : TestFormatter
```
rynowak:
```
for (var i = 0; i < Int32.MaxValue; i++)
{
    Console.WriteLine("Thank you!");
}
```

is deserved for writing these tests in a way that doesn't depend on the `JsonFormatter` implementation.

test/Microsoft.AspNet.Mvc.FunctionalTests/MvcSampleTests.cs
```diff
@@ -100,7 +100,7 @@ public MvcSampleTests(MvcTestFixture<MvcSample.Web.Startup> fixture)
         {
             // Arrange
             var request = new HttpRequestMessage(HttpMethod.Get, "http://localhost/Home/ReturnUser");
-            request.Headers.Accept.Add(MediaTypeWithQualityHeaderValue.Parse("application/xml;charset=utf-8"));
+            request.Headers.Accept.Add(MediaTypeWithQualityHeaderValue.Parse("application/xml"));
```
rynowak: Why does this need to change?

rynowak:
:watch:

src/Microsoft.AspNet.Mvc.Core/ConsumesAttribute.cs
dougbu: It was a duplicate comment. Open your window on this diff just a bit wider, look at https://github.com/aspnet/Mvc/pull/3317/files#diff-1b5675f1b083c83b1f98dd3e87cf08f6R52

src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
dougbu: I don't know but hit an NRE on the call to `IsSubsetOf()` just below. Seemed better to code defensively than narrow down or fix a bug in `MediaTypeHeaderValue` code I wasn't changing.

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
dougbu: nit: :smiling_imp:

src/Microsoft.AspNet.Mvc.Core/ConsumesAttribute.cs
rynowak: :+1:

src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
rynowak: It might be the case that a `null` value here propagates with `true`. This check would be less confusing as `contentType != null && ...TryParse(...)`

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
rynowak: If you're going to take the time to change it then it's fair game for feedback :trollface:

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
dougbu: There is in HttpAbstraction's `MediaTypeHeaderValue` and MVC's `MediaTypeHeaderValueComparer+ParsedMediaTypeHeaderValue` but both are `private`.

src/Microsoft.AspNet.Mvc.Core/ObjectResult.cs
dougbu: :ok: main thing is the formatters are passed `null` for their `MediaTypeHeaderValue` parameter. They can't make any decisions based on _that_ parameter and are left with just the `Type` parameter. Agree they can choose to ignore that.

src/Microsoft.AspNet.Mvc.Core/ObjectResult.cs
dougbu: :bow:ing

test/Microsoft.AspNet.Mvc.FunctionalTests/MvcSampleTests.cs
dougbu: Because an `Accept` header containing a `charset` is both invalid and _more_ specific than anything a formatter would admit to supporting.

dougbu:
Reopening so we can more-easily see this discussion 'til it completes.

test/Microsoft.AspNet.Mvc.FunctionalTests/MvcSampleTests.cs
rynowak: Should some of these tests use `Accept-Charset`? Or does that already have adequate coverage?

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
rynowak: Drat

test/Microsoft.AspNet.Mvc.FunctionalTests/MvcSampleTests.cs
dougbu: The functional tests don't use `Accept-Charset` much but unit tests such as [test/Microsoft.AspNet.Mvc.Core.Test/ObjectResultTests.cs](https://github.com/aspnet/Mvc/blob/e075a58ee8437300b560714ddae28dd0c45926ca/test/Microsoft.AspNet.Mvc.Core.Test/ObjectResultTests.cs) and [test/Microsoft.AspNet.Mvc.Formatters.Json.Test/JsonOutputFormatterTests.cs](https://github.com/aspnet/Mvc/blob/e075a58ee8437300b560714ddae28dd0c45926ca/test/Microsoft.AspNet.Mvc.Formatters.Json.Test/JsonOutputFormatterTests.cs) cover things fairly well.

rynowak: That seems fine.

rynowak:
:shipit:

src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
dougbu: Reverted. I must have broken something elsewhere at an intermediate point in working on this issue. `TryParse()` can't return `true` with the `out` parameter left `null`.

dougbu:
@rynowak know you've signed off. But I'd like your opinion on method renaming in new 3rd commit.

Addressed other comments as well.

dougbu:
AppVeyor and Travis should fail here. Pushed `dougbu/followup.contenttype` branch with just 3rd commit on a `dev` base for real checks.

rynowak:
> But I'd like your opinion on method renaming in new 3rd commit.

Name change is a small improvement over what we have IMO. You should go ahead with your changes and we can discuss when you're in the office.

dougbu:
Comments addressed in 9b00461

PR #3321, opened 2015-10-14
Add `null` checks for MVC tag helpers

dougbu:
- chose not to check `context` parameter in `Process[Async]()` methods that ignore it

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
@@ -107,6 +107,11 @@ namespace Microsoft.AspNet.Mvc.Razor.TagHelpers
         /// <inheritdoc />
         public override void Process(TagHelperContext context, TagHelperOutput output)
         {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
             if (output == null)
             {
                 throw new ArgumentNullException(nameof(output));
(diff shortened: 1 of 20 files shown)


src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
```diff
@@ -108,6 +107,11 @@ public override int Order
         /// <inheritdoc />
         public override void Process(TagHelperContext context, TagHelperOutput output)
         {
+            if (output == null)
```
pranavkm: Is this because we don't use context here?

pranavkm: Seems like it. Maybe it's just less work to throw in the null checks here so that we don't have to add that check if `content` starts being used?

pranavkm:
:shipit:

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
dougbu: Doubt this helper would change _that_ much but I'm not too fussed either way. Will add the extra check.

dougbu: Will add `context` check in other tag helpers where I left it out. Will also update the `protected` methods in this class -- `protected` is the new `public`.

dougbu:
Updated. FYI only since I've got a squirrel.

dougbu:
9aed5ef)

Eilon:
Wait a sec, why do we want to do this? These methods are called only by the framework and by our own tests. Not saying it's wrong, but why bother?

dougbu:
@Eilon interesting point. I made this change because the classes are `public`, can be subclassed, and looked inconsistent with other framework-provided classes. But it definitely wasn't mandatory.

Eilon:
It made the unit tests a bit uglier :crying_cat_face: 

I'm fine keeping this, but let's track these changes as bugs. This is a behavioral change, not just cleanup.

Eilon:
(In the future, that is.)

dougbu:
@Eilon in the near future, should we make a general announcement looking back to Beta8 for all the good `[NotNull]` -> `ArgumentNullException` work @pranavkm did?

Eilon:
Shouldn't affect anyone in theory.

Issue #3247, opened 2015-10-02
Optimize ModelStateDictionary allocations


rynowak:
Here's allocation data from a 3000 requests to a very simple API site. There's only one parameter and it's bound using `[FromBody]` - so this is very much the **best case** for MSD's allocations.

![image](https://cloud.githubusercontent.com/assets/1430011/10251559/7367c416-68e6-11e5-81f0-d7b320052ea6.png)

In context, the total allocations at 110mb, so this is just over 2%. Once we have buffer pooling for this sample, the total allocations will be closer to 70mb :grinning: 

The biggest issue here is the boxed enumerator and the `FindKeysWithPrefix` method, which uses `yield return`. `FindKeysWithPrefix` should probably be recast as `GetEnumerator(string)` returning a struct enumerator.

Secondly, we'd very likely benefit from turing MSD into a tree that implements `IDictionary<>`. The values stored in the dictionary are `ModelState` values which we must allocate anyway. We'd need to add a 'children' and 'next' field to each `ModelState` entry, and that would tradeoff some allocations of arrays for the dictionary for 2 fields on something we're already allocating. The usage pattern of MSD is more like a tree anyway, we rarely want to find a single entry, we're usually looking at all entries related to the prefix, which just screams treeeeeeeeeeeeeeeeeeeeeeeeee :scream: 

In this context, the allocations from tree vs dictionary aren't very significant. We should look at a sample with more data model bound.

rynowak:
054b39013c17face3da992e07497b326b73f9720 addressed the iterator allocations.

Going to close this for now and reopen if we find evidence that using a tree would net a significant benefit.

PR #3327, opened 2015-10-15
Reduce allocations in MSD enumeration

rynowak:
Introduces a struct `Enumerable` and `Enumerator` to remove a common source of allocations.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
+++ b/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
@@ -4,9 +4,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
-using System.Linq;
 using Microsoft.AspNet.Mvc.Abstractions;
-using Microsoft.Extensions.Internal;
 
 namespace Microsoft.AspNet.Mvc.ModelBinding
 {
@@ -22,7 +20,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         /// </summary>
         public static readonly int DefaultMaxAllowedErrors = 200;
 
-        private readonly IDictionary<string, ModelState> _innerDictionary;
+        private readonly Dictionary<string, ModelState> _innerDictionary;
         private int _maxAllowedErrors;
 
         /// <summary>
@@ -55,7 +53,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
                 throw new ArgumentNullException(nameof(dictionary));
             }
 
-            _innerDictionary = new CopyOnWriteDictionary<string, ModelState>(
+            _innerDictionary = new Dictionary<string, ModelState>(
                 dictionary,
                 StringComparer.OrdinalIgnoreCase);
 
@@ -124,7 +122,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         /// <inheritdoc />
         public bool IsReadOnly
         {
-            get { return _innerDictionary.IsReadOnly; }
+            get { return ((ICollection<KeyValuePair<string, ModelState>>)_innerDictionary).IsReadOnly; }
         }
 
         /// <inheritdoc />
@@ -153,7 +151,11 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         /// <inheritdoc />
         public ModelValidationState ValidationState
         {
-            get { return GetValidity(_innerDictionary); }
+            get
+            {
+                var entries = FindKeysWithPrefix(string.Empty);
+                return GetValidity(entries, defaultState: ModelValidationState.Valid);
+            }
         }
 
         /// <inheritdoc />
@@ -344,12 +346,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             }
 
             var entries = FindKeysWithPrefix(key);
-            if (!entries.Any())
-            {
-                return ModelValidationState.Unvalidated;
-            }
-
-            return GetValidity(entries);
+            return GetValidity(entries, defaultState: ModelValidationState.Unvalidated);
         }
 
         /// <summary>
@@ -496,9 +493,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         {
             // If key is null or empty, clear all entries in the dictionary
             // else just clear the ones that have key as prefix
-            var entries = (string.IsNullOrEmpty(key)) ?
-                _innerDictionary : FindKeysWithPrefix(key);
-
+            var entries = FindKeysWithPrefix(key ?? string.Empty);
             foreach (var entry in entries)
             {
                 entry.Value.Errors.Clear();
@@ -523,11 +518,16 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
             return modelState;
         }
 
-        private static ModelValidationState GetValidity(IEnumerable<KeyValuePair<string, ModelState>> entries)
+        private static ModelValidationState GetValidity(PrefixEnumerable entries, ModelValidationState defaultState)
         {
+
+            var hasEntries = false;
             var validationState = ModelValidationState.Valid;
+
             foreach (var entry in entries)
             {
+                hasEntries = true;
+
                 var entryState = entry.Value.ValidationState;
                 if (entryState == ModelValidationState.Unvalidated)
                 {
@@ -539,7 +539,8 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
                     validationState = entryState;
                 }
             }
-            return validationState;
+
+            return hasEntries ? validationState : defaultState;
         }
 
         private void EnsureMaxErrorsReachedRecorded()
@@ -591,7 +592,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         /// <inheritdoc />
         public bool Contains(KeyValuePair<string, ModelState> item)
         {
-            return _innerDictionary.Contains(item);
+            return ((ICollection<KeyValuePair<string, ModelState>>)_innerDictionary).Contains(item);
         }
 
         /// <inheritdoc />
@@ -613,13 +614,13 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
                 throw new ArgumentNullException(nameof(array));
             }
 
-            _innerDictionary.CopyTo(array, arrayIndex);
+            ((ICollection<KeyValuePair<string, ModelState>>)_innerDictionary).CopyTo(array, arrayIndex);
         }
 
         /// <inheritdoc />
         public bool Remove(KeyValuePair<string, ModelState> item)
         {
-            return _innerDictionary.Remove(item);
+            return ((ICollection<KeyValuePair<string, ModelState>>)_innerDictionary).Remove(item);
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
```diff
@@ -656,70 +653,204 @@ IEnumerator IEnumerable.GetEnumerator()
...
+                return false;
             }
 
-            foreach (var entry in _innerDictionary)
+            if (!key.StartsWith(prefix, StringComparison.OrdinalIgnoreCase))
             {
-                var key = entry.Key;
+                if (key.StartsWith("[", StringComparison.OrdinalIgnoreCase))
+                {
+                    var subKey = key.Substring(key.IndexOf('.') + 1);
```
rynowak: This could obviously be optimized further, this isn't the part I'm going after. Will wait for some more data to motivate further changes.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
```diff
@@ -656,70 +653,204 @@ IEnumerator IEnumerable.GetEnumerator()
...
+            }
+        }
+
+        public struct PrefixEnumerator : IEnumerator<KeyValuePair<string, ModelState>>
+        {
+            private readonly ModelStateDictionary _dictionary;
+            private readonly string _prefix;
+
+            private bool _exactMatchUsed;
+            private IEnumerator<KeyValuePair<string, ModelState>> _enumerator;
```
rynowak: The other issue here is that this still allocates due to boxing. The changes in https://github.com/aspnet/Common/pull/53 will allow us to eliminate this change will still preserving the copy on write semantic.

rynowak:
Will add some perf numbers tomorrow based on https://github.com/aspnet/Mvc/issues/3247

I expect all of these enumerator allocations to disappear.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
```diff
@@ -656,70 +653,204 @@ IEnumerator IEnumerable.GetEnumerator()
...
+                {
+                    var subKey = key.Substring(key.IndexOf('.') + 1);
 
-                if (key.Length <= prefix.Length)
+                    if (!subKey.StartsWith(prefix, StringComparison.OrdinalIgnoreCase))
+                    {
+                        return false;
+                    }
+
+                    if (string.Equals(prefix, subKey, StringComparison.Ordinal))
```
Eilon: Any idea why this one is Ordinal and not IgnoreCase? (I realize it didn't change, but not sure why it was that way.)

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
```diff
@@ -656,70 +653,204 @@ IEnumerator IEnumerable.GetEnumerator()
...
                 }
-                else
+
+                _dictionary = dictionary;
+                _prefix = prefix;
+            }
+
+            public PrefixEnumerator GetEnumerator()
+            {
+                return _dictionary == null ? new PrefixEnumerator() : new PrefixEnumerator(_dictionary, _prefix);
```
dougbu: Why does this ternary exist when the only constructor that's used requires a non-`null` dictionary? Is this for some test scenarios of the future.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
```diff
@@ -656,70 +653,204 @@ IEnumerator IEnumerable.GetEnumerator()
...
+                    {
+                        return false;
+                    }
+
+                    if (string.Equals(prefix, subKey, StringComparison.Ordinal))
+                    {
+                        return true;
+                    }
+
+                    key = subKey;
```
dougbu: This is a strange case that makes `"[index].prefix"`, `"[index].prefix.property"` and so on a match for `"prefix"`. Do we have the history on why?

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
```diff
@@ -656,70 +653,204 @@ IEnumerator IEnumerable.GetEnumerator()
...
+                }
+            }
+
+            public void Dispose()
+            {
+            }
+
+            public bool MoveNext()
+            {
+                if (_dictionary == null)
```
dougbu: Won't need this if the ternary at [line 754](https://github.com/aspnet/Mvc/pull/3327/files#diff-a2b0a5845b6336b458622ec4dacfb762R754) goes away.

dougbu:
Is this [design] because you don't have low-level tests of the new enumerator / enumerable? Or for the perf numbers?

Can't see how these changes would break existing tests but want to confirm since the CI builds weren't 100% successful.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
rynowak: No, I don't know the history of this. @Eilon

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
rynowak: This is a struct. It's always possible to call the default constructor

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
rynowak: This is a struct. It's always possible to call the default constructor

rynowak:
> Is this [design] because you don't have low-level tests of the new enumerator / enumerable? Or for the perf numbers?

I haven't run the perf numbers yet, because this was dependent on the outcome of our discussion Friday.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
rynowak: No I really don't know. See discussion here https://github.com/aspnet/Mvc/pull/3327#discussion-diff-42335129 - there's plenty of wierdness here in general.

rynowak:
Allocations for 3000 request to a simple API site - post JSON + return JSON

**Before**
image

**After**
image

dougbu:
Is this still a designer PR?

rynowak:
> Is this still a designer PR?

NO WAY

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
```diff
@@ -55,7 +53,7 @@ public ModelStateDictionary(ModelStateDictionary dictionary)
                 throw new ArgumentNullException(nameof(dictionary));
             }
 
-            _innerDictionary = new CopyOnWriteDictionary<string, ModelState>(
```
dougbu: What still calls this copy constructor? I'd rather remove it than simplify it.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
```diff
@@ -153,7 +151,7 @@ public bool IsValid
         /// <inheritdoc />
         public ModelValidationState ValidationState
         {
-            get { return GetValidity(_innerDictionary); }
+            get { return GetValidity(FindKeysWithPrefix(string.Empty), defaultState: ModelValidationState.Valid); }
```
dougbu: :dancers:

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
```diff
@@ -344,12 +342,7 @@ public ModelValidationState GetFieldValidationState(string key)
             }
 
             var entries = FindKeysWithPrefix(key);
```
dougbu: minor: Like having this temporary variable for debugging. Could you make the `ValidationState` property look similar?

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
```diff
@@ -496,10 +489,7 @@ public void ClearValidationState(string key)
         {
             // If key is null or empty, clear all entries in the dictionary
             // else just clear the ones that have key as prefix
-            var entries = (string.IsNullOrEmpty(key)) ?
-                _innerDictionary : FindKeysWithPrefix(key);
-
-            foreach (var entry in entries)
+            foreach (var entry in FindKeysWithPrefix(key ?? string.Empty))
```
dougbu: minor: Let's keep the `entries` temporary variable around for debugging.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
dougbu: Well, it's an inner `struct` that's `public` only because it has to be and we don't use it that way but :ok:

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
dougbu: If we don't hear from @Eilon before you push this change, please file a follow-up issue about this odd complexity and the inconsistency [discussed above](https://github.com/aspnet/Mvc/pull/3327#discussion-diff-42316029).

dougbu:
:shipit:, just minor comments

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
rynowak: Yeah, I think this is just a mistake. Going to fix it as long as nothing breaks.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
rynowak: Going to follow up on this item immediately - this was added as a sattelite change of https://github.com/aspnet/Mvc/commit/d0927bdc75cd88d1cba0a0e9264cb2b3ec81de76 and it seems dubious.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
rynowak: Discussed in person and agreed to leave the copy constructor in place even though we don't call it.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
rynowak: Sure

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
rynowak: sure

Issue #3332, opened 2015-10-16
Rename ViewComponent.Context to HttpContext


DmitrySikorsky:
Because now we have this property named 'HttpContext' in the Controller, so it is strange to have different name in the ViewComponent.

Muchiachio:
And if we are on this, then [RazorPage](https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs) and [ApiController](https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.WebApiCompatShim/ApiController.cs) too?

DmitrySikorsky:
Yes. I think they all have to have some one name.

rynowak:
For `RazorPage` and `ApiController` we need to be consistent with previous versions

henkmollema:
:+1: this is pretty confusing at the moment: http://stackoverflow.com/q/33212908

Eilon:
@pranavkm can you have a look at the PR and merge?

pranavkm:
Thanks for the PR!

henkmollema:
You're welcome :smile: 
Does this need an announcement or some other breaking change notice?

pranavkm:
@danroth27, thoughts on the announcement question?

rynowak:
@pranavkm Might as well make an announcement grouping all of the similar changes we did like this, it's a change affecting a very common surface area :+1:

pranavkm:
@rynowak could you point me to the "similar changes" so I could add it to the announcement?

PR #3344, opened 2015-10-19
Rename ViewComponent.Context to HttpContext

henkmollema:
Rename the property to `HttpContext` to be consistent with the `Controller` class.

#3332

### DIFF
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponent.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponent.cs
@@ -27,9 +27,9 @@ namespace Microsoft.AspNet.Mvc
         private ICompositeViewEngine _viewEngine;
 
         /// <summary>
-        /// Gets the <see cref="HttpContext"/>.
+        /// Gets the <see cref="Http.HttpContext"/>.
         /// </summary>
-        public HttpContext Context
+        public HttpContext HttpContext
         {
             get
             {


Eilon:
@pranavkm

henkmollema:
Are there any more places where `Context` is referenced you know of? I think I got them all in the Mvc repo.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -46,16 +46,11 @@ public RazorPage()
             _writerScopes = new Stack<TextWriter>();
         }
 
-        public HttpContext Context
+        public HttpContext HttpContext
```
pranavkm: If you feel like making the change (since we don't actually have a guideline or preference):

`public HttpContext HttpContext => ViewContext?.HttpContext;`

pranavkm: I just noticed @rynowak's comment here - https://github.com/aspnet/Mvc/issues/3332#issuecomment-148739038 and we don't want to break compat here. Could you undo changes to this file?

pranavkm:
:watch:

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
henkmollema: Oops, missed that one too. I'll revert it.

henkmollema:
@pranavkm I reverted the expression body changes in `RazorPage` too, is guess this is out of scope now.

henkmollema:
@pranavkm I squashed and rebased. Want me to do anything else?

henkmollema:
Merged: https://github.com/aspnet/Mvc/commit/34d4dfdef8a6555c8ce841263893823891a0aeed

henkmollema:
s

PR #3350, opened 2015-10-19
React to aspnet/localization #90

kirthik:
`.resx` files now should provide full name.

cc @Eilon and @dougbu


kirthik:
merged this change since it was breaking tests on CI. Feel free to provide comments. I will address them in subsequent check ins.

dougbu:
one after-the-fact :shipit:

Issue #3274, opened 2015-10-06
Diagnostics Source events for View Components


avanderhoorn:
Originally we had intended to show child actions. View Components are the successor to what child actions where intended to do and we would like to have Diagnostics Source events for the "action"/component resolution and corresponding view resolution.

// @rynowak @Eilon

rynowak:
Would love to see more detail about what this would look like and what kinds of things you expect users to troubleshoot/understand with this experience.

avanderhoorn:
As demonstrated by the various View Components in [MusicStore](https://github.com/aspnet/MusicStore/blob/master/src/MusicStore/Components/CartSummaryComponent.cs#L30-L34), I think it is imagined that a good percentage of users will be doing some sort of out-of-proc work in these (probably mainly db stuff) and because of that it makes sense to show them what is causing that work to occur and what the result it. Hence shoring which View Component have executed, what happened within them and what resulted is really useful. Given this, we would like a before/after for the View Component itself and then for the "view" that results from that. Does that make sense/help?

PR #3351, opened 2015-10-19
Added DiagnosticsSource events for ViewComponent

ajaybhargavb:
#3274 

@rynowak

### DIFF
--- a/test/Microsoft.AspNet.Mvc.TestDiagnosticListener.Sources/TestDiagnosticListener.cs
+++ b/test/Microsoft.AspNet.Mvc.TestDiagnosticListener.Sources/TestDiagnosticListener.cs
@@ -220,5 +220,109 @@ namespace Microsoft.AspNet.Mvc
                 ViewContext = viewContext,
             };
         }
+
+        public class OnBeforeViewComponentEventData
+        {
+            public IProxyActionDescriptor ActionDescriptor { get; set; }
+
+            public IProxyViewComponentContext ViewComponentContext { get; set; }
+
+            public object ViewComponent { get; set; }
+        }
+
+        public OnBeforeViewComponentEventData BeforeViewComponent { get; set; }
+
+        [DiagnosticName("Microsoft.AspNet.Mvc.BeforeViewComponent")]
+        public virtual void OnBeforeViewComponent(
+            IProxyActionDescriptor actionDescriptor,
+            IProxyViewComponentContext viewComponentContext,
+            object viewComponent)
+        {
+            BeforeViewComponent = new OnBeforeViewComponentEventData()
+            {
+                ActionDescriptor = actionDescriptor,
+                ViewComponentContext = viewComponentContext,
+                ViewComponent = viewComponent
+            };
+        }
+
+        public class OnAfterViewComponentEventData
+        {
+            public IProxyActionDescriptor ActionDescriptor { get; set; }
+
+            public IProxyViewComponentContext ViewComponentContext { get; set; }
+
+            public IProxyViewComponentResult ViewComponentResult { get; set; }
+
+            public object ViewComponent { get; set; }
+        }
+
+        public OnAfterViewComponentEventData AfterViewComponent { get; set; }
+
+        [DiagnosticName("Microsoft.AspNet.Mvc.AfterViewComponent")]
+        public virtual void OnAfterViewComponent(
+            IProxyActionDescriptor actionDescriptor,
+            IProxyViewComponentContext viewComponentContext,
+            IProxyViewComponentResult viewComponentResult,
+            object viewComponent)
+        {
+            AfterViewComponent = new OnAfterViewComponentEventData()
+            {
+                ActionDescriptor = actionDescriptor,
+                ViewComponentContext = viewComponentContext,
+                ViewComponentResult = viewComponentResult,
+                ViewComponent = viewComponent
+            };
+        }
+
+        public class OnViewComponentBeforeViewExecuteEventData
+        {
+            public IProxyActionDescriptor ActionDescriptor { get; set; }
+
+            public IProxyViewComponentContext ViewComponentContext { get; set; }
+
+            public IProxyView View { get; set; }
+        }
+
+        public OnViewComponentBeforeViewExecuteEventData ViewComponentBeforeViewExecute { get; set; }
+
+        [DiagnosticName("Microsoft.AspNet.Mvc.ViewComponentBeforeViewExecute")]
+        public virtual void OnViewComponentBeforeViewExecute(
+            IProxyActionDescriptor actionDescriptor,
+            IProxyViewComponentContext viewComponentContext,
+            IProxyView view)
+        {
+            ViewComponentBeforeViewExecute = new OnViewComponentBeforeViewExecuteEventData()
+            {
+                ActionDescriptor = actionDescriptor,
+                ViewComponentContext = viewComponentContext,
+                View = view
+            };
+        }
+
+        public class OnViewComponentAfterViewExecuteEventData
+        {
+            public IProxyActionDescriptor ActionDescriptor { get; set; }
+
+            public IProxyViewComponentContext ViewComponentContext { get; set; }
+
+            public IProxyView View { get; set; }
+        }
+
+        public OnViewComponentAfterViewExecuteEventData ViewComponentAfterViewExecute { get; set; }
+
+        [DiagnosticName("Microsoft.AspNet.Mvc.ViewComponentAfterViewExecute")]
+        public virtual void OnViewComponentAfterViewExecute(
+            IProxyActionDescriptor actionDescriptor,
+            IProxyViewComponentContext viewComponentContext,
+            IProxyView view)
+        {
+            ViewComponentAfterViewExecute = new OnViewComponentAfterViewExecuteEventData()
+            {
+                ActionDescriptor = actionDescriptor,
+                ViewComponentContext = viewComponentContext,
+                View = view
+            };
+        }
     }
 }
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/DiagnosticSource/ViewComponentDiagnosticSourceExtensions.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/DiagnosticSource/ViewComponentDiagnosticSourceExtensions.cs
@@ -1,0 +1,87 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Diagnostics;
+using Microsoft.AspNet.Mvc.Abstractions;
+using Microsoft.AspNet.Mvc.ViewComponents;
+using Microsoft.AspNet.Mvc.ViewEngines;
+
+namespace Microsoft.AspNet.Mvc.Diagnostics
+{
+    public static class ViewComponentDiagnosticSourceExtensions
+    {
+        public static void BeforeViewComponent(
+            this DiagnosticSource diagnosticSource,
+            ViewComponentContext context,
+            object viewComponent)
+        {
+            if (diagnosticSource.IsEnabled("Microsoft.AspNet.Mvc.BeforeViewComponent"))
+            {
+                diagnosticSource.Write(
+                "Microsoft.AspNet.Mvc.BeforeViewComponent",
+                new
+                {
+                    actionDescriptor = context.ViewContext.ActionDescriptor,
+                    viewComponentContext = context,
+                    viewComponent = viewComponent
+                });
+            }
+        }
+
+        public static void AfterViewComponent(
+            this DiagnosticSource diagnosticSource,
+            ViewComponentContext context,
+            IViewComponentResult result,
+            object viewComponent)
+        {
+            if (diagnosticSource.IsEnabled("Microsoft.AspNet.Mvc.AfterViewComponent"))
+            {
+                diagnosticSource.Write(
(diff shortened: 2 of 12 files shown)


src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentHelper.cs
```diff
@@ -210,7 +216,13 @@ private ViewComponentDescriptor SelectComponent(Type componentType)
                     Resources.FormatViewComponent_IViewComponentFactory_ReturnedNull(descriptor.Type.FullName));
             }
 
-            return invoker.InvokeAsync(context);
+            WriteTelemetryIfEnabled("Microsoft.AspNet.Mvc.BeforeViewComponent", context);
```
rynowak: No.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentHelper.cs
```diff
@@ -237,7 +249,23 @@ private ViewComponentDescriptor SelectComponent(Type componentType)
...
+        }
+
+        private void WriteTelemetryIfEnabled(string telemetryName, ViewComponentContext context)
+        {
+#pragma warning disable 0618
+            if (_telemetry.IsEnabled(telemetryName))
+            {
+                _telemetry.WriteTelemetry(
+                    telemetryName,
+                    new { viewComponentContext = context });
```
rynowak: Do we have the action context here?

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentHelper.cs
```diff
@@ -237,7 +249,23 @@ private ViewComponentDescriptor SelectComponent(Type componentType)
                     Resources.FormatViewComponent_IViewComponentFactory_ReturnedNull(descriptor.Type.FullName));
             }
 
+            WriteTelemetryIfEnabled("Microsoft.AspNet.Mvc.BeforeViewComponent", context);
```
avanderhoorn: What is the difference between this `InvokeCore` Before/AfterViewComponent event and the `Contextualize` events of the same name?

rynowak: There are two code paths here, `InvokeCore` and `InvokeCoreAsync` - it depends on whether the calling code is async or not. `Contextualize` is something different entirely and isn't related.

avanderhoorn: lol ok, I just saw why I got that screwed up... github had folded the commit. Sorry about that. On a slightly different note, do you see any benefit in letting the user know when something is async like this or not?

rynowak: I don't see much benefit, and I'd personally just lean towards us removing the sync version - which may still happen before release.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentHelper.cs
ajaybhargavb: No.

rynowak: It looks like it's included as a member of the VCC so we're good.

Let's add the actiondescriptor as its own top level item here

ajaybhargavb:
Updated.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentHelper.cs
```diff
@@ -210,7 +216,25 @@ private ViewComponentDescriptor SelectComponent(Type componentType)
                     Resources.FormatViewComponent_IViewComponentFactory_ReturnedNull(descriptor.Type.FullName));
             }
 
-            return invoker.InvokeAsync(context);
+#pragma warning disable 0618
+            if (_telemetry.IsEnabled("Microsoft.AspNet.Mvc.BeforeViewComponent"))
+            {
+                _telemetry.WriteTelemetry(
+                    "Microsoft.AspNet.Mvc.BeforeViewComponent",
+                    new { actionDescriptor = _viewContext.ActionDescriptor, viewComponentContext = context });
```
avanderhoorn: Is the `ActionDescriptor` here that of root/origin action?

rynowak: Yes

ajaybhargavb:
Updated.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentHelper.cs
```diff
@@ -15,12 +16,15 @@ public class DefaultViewComponentHelper : IViewComponentHelper, ICanHasViewConte
         private readonly IViewComponentDescriptorCollectionProvider _descriptorProvider;
         private readonly IViewComponentInvokerFactory _invokerFactory;
         private readonly IViewComponentSelector _selector;
+#pragma warning disable 0618
```
rynowak: Remove all of these pragmas. They were needed because the old type was `[Obsolete]`

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentHelper.cs
```diff
@@ -210,7 +216,25 @@ private ViewComponentDescriptor SelectComponent(Type componentType)
...
+                    new { actionDescriptor = _viewContext.ActionDescriptor, viewComponentContext = context });
+            }
+
+            var result = invoker.InvokeAsync(context);
+
+            if (_diagnosticSource.IsEnabled("Microsoft.AspNet.Mvc.AfterViewComponent"))
+            {
+                _diagnosticSource.Write(
+                    "Microsoft.AspNet.Mvc.AfterViewComponent",
+                    new { actionDescriptor = _viewContext.ActionDescriptor, viewComponentContext = context });
```
rynowak: Include the result here as well

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentHelper.cs
```diff
@@ -210,7 +216,25 @@ private ViewComponentDescriptor SelectComponent(Type componentType)
...
-            return invoker.InvokeAsync(context);
+#pragma warning disable 0618
+            if (_diagnosticSource.IsEnabled("Microsoft.AspNet.Mvc.BeforeViewComponent"))
+            {
+                _diagnosticSource.Write(
+                    "Microsoft.AspNet.Mvc.BeforeViewComponent",
+                    new { actionDescriptor = _viewContext.ActionDescriptor, viewComponentContext = context });
+            }
+
+            var result = invoker.InvokeAsync(context);
```
rynowak: you need to await this call now.

rynowak: Should we have done this change in the invoker? That would allow us to get at the result returned from the VC.

test/Microsoft.AspNet.Mvc.TestDiagnosticListener.Sources/TestDiagnosticListener.cs
```diff
@@ -220,5 +220,91 @@ public virtual void OnAfterView(IProxyView view, IProxyViewContext viewContext)
...
+            public IProxyActionDescriptor ActionDescriptor { get; set; }
+
+            public IProxyViewComponentContext ViewComponentContext { get; set; }
+        }
+
+        public OnBeforeViewComponentEventData BeforeViewComponent { get; set; }
+
+        [DiagnosticName("Microsoft.AspNet.Mvc.BeforeViewComponent")]
+        public virtual void OnBeforeViewComponent(
+            IProxyActionDescriptor actionDescriptor, IProxyViewComponentContext viewComponentContext)
```
rynowak: no #YOLOing

rynowak: put everything on its own line

rynowak:
:watch:

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentHelper.cs
ajaybhargavb: The reason I put it here was putting it in invoker might lead to duplication because we call `InvokeSyncCore` in two places. But If we want the result, we should put it there. I'll move it.

ajaybhargavb:
Updated.

rynowak:
:shipit:

avanderhoorn:
Quick question, for VC's, is there something like the `ActionId` which can be used for correlating all of the VC events like we do for Actions?

rynowak:
> Quick question, for VC's, is there something like the ActionId which can be used for correlating all of the VC events like we do for Actions?

You can use the type name (on the descriptor). Have a gander at that and we'll talk if that doesn't meet your requirements.

ajaybhargavb:
Updated. (Added extension methods)

avanderhoorn:
Looking good... just a couple of questions in general about the objects that I'm getting.

 - `ViewComponentContext`
     - Is `Arguments` the arguments that matchup to what is being passed into the VC?
         - If so, is there any way to match these back to the names of the arguments like we do for action?
     - Is `ViewContext` the root view context... as in the view that corresponds the the action that started the whole show?
 - `ViewComponentDescriptor`
     - The [same way](https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Abstractions/Abstractions/ActionDescriptor.cs#L18) that `ActionDescriptor` has an id, can the VS one have this as well... it will help with corilation
     - Currently the VC Descriptor is accessible in any of the `ViewComponentDiagnosticSourceExtensions` events... is it possible to have it accessible via the view events that executed as a result a VC running? This will help with correlation
 - `ViewComponent` 
     - At the moment we don't have access tot he `ViewComponent` itself, how important is this? Meaning the main things of interest are things like `ViewData`, `ModelState`, etc... is this going to be the same as the root view, or can it be different for each VC? If its the later, it would be good to have access to it.

rynowak:
> If so, is there any way to match these back to the names of the arguments like we do for action?

You can look at the VCD and get the `MethodInfo` I believe.

> Is ViewContext the root view context... as in the view that corresponds the the action that started the whole show?

Not likely, what do you need?

>The same way that ActionDescriptor has an id, can the VS one have this as well... it will help with corilation (sic)

We can certainly add something like this, but the class name should be able to do the same. Isn't that better since it's readable? If you still want it, let's do it in a follow up.

rynowak:
> is it possible to have it accessible via the view events that executed as a result a VC running? This will help with correlation

Can you give a source reference to what you're talking about?

rynowak:
> At the moment we don't have access tot he ViewComponent itself, how important is this? Meaning the main things of interest are things like ViewData, ModelState, etc... is this going to be the same as the root view, or can it be different for each VC? If its the later, it would be good to have access to it.

Yeah, we should find a way to do this @ajaybhargavb

avanderhoorn:
>If so, is there any way to match these back to the names of the arguments like we do for action?
You can look at the VCD and get the MethodInfo I believe.

Pity, would have been nice if it was just there like it is for action and not have to handle this independently. 

>Is ViewContext the root view context... as in the view that corresponds the the action that started the whole show?
Not likely, what do you need?

No, it was just a general question.

> The same way that ActionDescriptor has an id, can the VS one have this as well... it will help with correlation (sic)
We can certainly add something like this, but the class name should be able to do the same. Isn't that better since it's readable? If you still want it, let's do it in a follow up.

Was more for consistency with Actions and how correlation is done generally... For actions I was using the Id and here, if it doesn't have an id, I'm probably best to update my action logic to match. But not a huge huge deal.

> is it possible to have it accessible via the view events that executed as a result a VC running? This will help with correlation
Can you give a source reference to what you're talking about?

In cases like [this](https://github.com/aspnet/Mvc/blob/telemetry/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/PartialViewResultExecutor.cs#L74) where we are rendering the view for a View Component, it would be nice to be able to correlate it back to the VC... I was thinking `ViewComponentDescriptor` would be the easiest way to do this.

>At the moment we don't have access tot he ViewComponent itself, how important is this? Meaning the main things of interest are things like ViewData, ModelState, etc... is this going to be the same as the root view, or can it be different for each VC? If its the later, it would be good to have access to it.
Yeah, we should find a way to do this

Great!

ajaybhargavb:
Updated (Added ViewComponent object to the notification)

rynowak:
:shipit: 

@avanderhoorn - I'm currently working on adding an Id and DisplayName to VCD in another PR

avanderhoorn:
@ajaybhargavb Great!
@rynowak Great again!

Looks like it :clap:'s all around :D

PR #3354, opened 2015-10-20
Refactor of OutputFormatterContext

rynowak:
This change applies the expanding context pattern to output formatters.

Also simplifies the boilerplate around `CanWriteType`. Instead of providing the declared type and runtime type, and asking the formatter to figure it out, we now just provide the declared type **or** runtime type (whichever is meaningful). This really simplifies things.

/cc @lodejard


src/Microsoft.AspNet.Mvc.Abstractions/Formatters/OutputFormatterCanWriteContext.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// and expect to see the same value provided in
+        /// <see cref="IOutputFormatter.WriteAsync(OutputFormatterWriteContext)"/>
+        /// </remarks>
+        public virtual MediaTypeHeaderValue ContentType { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value inidicating that content-negotiation could not find a formatter based on the 
+        /// information on the <see cref="Http.HttpRequest"/>.
+        /// </summary>
+        public virtual bool? FailedContentNegotiation { get; protected set; }
```
rynowak: Note: I didn't remove this yet since this isn't really what this PR is about. Will address that feedback later.

src/Microsoft.AspNet.Mvc.Core/ApiExplorer/IApiResponseFormatMetadataProvider.cs
```diff
@@ -11,25 +11,18 @@ namespace Microsoft.AspNet.Mvc.ApiExplorer
...
-        /// type can be used.
+        /// <param name="context">The <see cref="OutputFormatterCanWriteContext"/></param>
         /// </param>
-        /// <returns>Content types which are supported by the <see cref="Formatters.IOutputFormatter"/>.</returns>
-        IReadOnlyList<MediaTypeHeaderValue> GetSupportedContentTypes(
-            Type declaredType,
-            Type runtimeType,
-            MediaTypeHeaderValue contentType);
+        /// <returns>Content types which are supported by the <see cref="IOutputFormatter"/>.</returns>
+        IReadOnlyList<MediaTypeHeaderValue> GetSupportedContentTypes(OutputFormatterCanWriteContext context);
```
rynowak: It's now possible to reuse this context type here now, since we don't need the provide an `HttpContext`.

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
```diff
@@ -112,95 +109,97 @@ public virtual Encoding SelectCharacterEncoding(OutputFormatterContext context)
...
+                        // This is supported.
+                        return context.ContentType.Encoding;
+                    }
                 }
             }
 
-            encoding = encoding ?? SupportedEncodings.FirstOrDefault();
-            return encoding;
+            // A formatter for a non-text media-type won't have any supported encodings.
+            return SupportedEncodings.Count > 0 ? SupportedEncodings[0] : null;
```
rynowak: Changes here are because there's no more `SelectedEncoding` property.

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
```diff
@@ -215,16 +214,29 @@ public virtual void WriteResponseHeaders(OutputFormatterContext context)
             var selectedEncoding = SelectCharacterEncoding(context);
             if (selectedEncoding != null)
             {
-                context.SelectedEncoding = selectedEncoding;
-
                 // Override the content type value even if one already existed.
-                selectedMediaType.Charset = selectedEncoding.WebName;
+                selectedMediaType.Encoding = selectedEncoding;
```
rynowak: Since there's no more `SelectedEncoding` - we just apply it to the CT directly. This is more straightforward than what we used to do :+1:

src/Microsoft.AspNet.Mvc.Abstractions/Formatters/IOutputFormatter.cs
```diff
@@ -21,13 +20,13 @@ public interface IOutputFormatter
         /// <paramref name="contentType"/> and is able to serialize the object
         /// represent by <paramref name="context"/>'s Object property.
         /// False otherwise.</returns>
-        bool CanWriteResult(OutputFormatterContext context, MediaTypeHeaderValue contentType);
+        bool CanWriteResult(OutputFormatterCanWriteContext context);
```
NTaylorMullen: nuke `<param>` for `contentType`

src/Microsoft.AspNet.Mvc.Abstractions/Formatters/OutputFormatterCanWriteContext.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// <summary>
+        /// Gets or sets the object to write to the response.
+        /// </summary>
+        public virtual object Object { get; protected set; }
+
+        /// <summary>
+        /// Gets or sets the <see cref="Type"/> of the object to write to the response.
+        /// </summary>
+        public virtual Type ObjectType { get; protected set; }
+    }
```
NTaylorMullen: Curious: Is it uncommon/not useful for users to have access to the `HttpContext` during their `CanWrite` phase of output formatting? I ask because I see that it was removed.

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
```diff
@@ -112,95 +109,97 @@ public virtual Encoding SelectCharacterEncoding(OutputFormatterContext context)
...
-            MediaTypeHeaderValue mediaType = null;
-            if (contentType == null)
+            
+            if (context.ContentType == null)
             {
-                // If the desired content type is set to null, the current formatter is free to choose the
-                // response media type.
-                mediaType = SupportedMediaTypes.FirstOrDefault();
+                // If the desired content type is set to null, then the current formatter can write anything
+                // it want.
```
NTaylorMullen: wants

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
```diff
@@ -112,95 +109,97 @@ public virtual Encoding SelectCharacterEncoding(OutputFormatterContext context)
...
-            // If content type is not set then set it based on supported media types.
-            selectedMediaType = selectedMediaType ?? SupportedMediaTypes.FirstOrDefault();
+            var selectedMediaType = context.ContentType;
             if (selectedMediaType == null)
             {
-                throw new InvalidOperationException(Resources.FormatOutputFormatterNoMediaType(GetType().FullName));
+                // If content type is not set then set it based on supported media types.
+                if (SupportedEncodings.Count > 0)
+                {
+                    selectedMediaType = SupportedMediaTypes[0];
```
NTaylorMullen: What scenario would result in this being hit? I'd think this would be hit prior to this method via `CanWriteResult`

src/Microsoft.AspNet.Mvc.Core/Formatters/StreamOutputFormatter.cs
```diff
@@ -25,7 +24,6 @@ public bool CanWriteResult(OutputFormatterContext context, MediaTypeHeaderValue
             // Ignore the passed in content type, if the object is a Stream.
             if (context.Object is Stream)
             {
-                context.SelectedContentType = contentType;
```
NTaylorMullen: ?

src/Microsoft.AspNet.Mvc.Core/Formatters/StringOutputFormatter.cs
```diff
@@ -59,7 +59,7 @@ public override Task WriteResponseBodyAsync(OutputFormatterContext context)
 
             var response = context.HttpContext.Response;
 
-            return response.WriteAsync(valueAsString, context.SelectedEncoding);
+            return response.WriteAsync(valueAsString, context.ContentType.Encoding ?? Encoding.UTF8);
```
NTaylorMullen: Not possible for `ContentType` to be `null`?

src/Microsoft.AspNet.Mvc.Formatters.Json/JsonOutputFormatter.cs
```diff
@@ -106,7 +106,7 @@ public override Task WriteResponseBodyAsync(OutputFormatterContext context)
             }
 
             var response = context.HttpContext.Response;
-            var selectedEncoding = context.SelectedEncoding;
+            var selectedEncoding = context.ContentType.Encoding ?? Encoding.UTF8;
```
NTaylorMullen: Same [Q](https://github.com/aspnet/Mvc/pull/3354/files#r42544439)

src/Microsoft.AspNet.Mvc.Formatters.Xml/XmlDataContractSerializerOutputFormatter.cs
```diff
@@ -88,41 +89,27 @@ public DataContractSerializerSettings SerializerSettings
...
         protected virtual Type GetSerializableType(Type type)
         {
-            var wrapperProvider = WrapperProviderFactories.GetWrapperProvider(
-                                                        new WrapperProviderContext(type, isSerialization: true));
+            if (type == null)
+            {
+                throw new ArgumentNullException(nameof(type));
+            }
+
+            var wrapperProvider = WrapperProviderFactories.GetWrapperProvider(new WrapperProviderContext(
```
NTaylorMullen: nit: 

```C#
var wrapperProvider = WrapperProviderFactories.GetWrapperProvider(
    new WrapperProviderContext(type, isSerialization: true));
```

NTaylorMullen:
:watch: just for answers to questions

src/Microsoft.AspNet.Mvc.Formatters.Xml/XmlDataContractSerializerOutputFormatter.cs
rynowak: My way is better

src/Microsoft.AspNet.Mvc.Core/Formatters/StringOutputFormatter.cs
rynowak: Yes, you're right, this should be `?.`

src/Microsoft.AspNet.Mvc.Formatters.Json/JsonOutputFormatter.cs
rynowak: Yes, you're right, this should be `?.`

src/Microsoft.AspNet.Mvc.Core/Formatters/StreamOutputFormatter.cs
rynowak: There is no `contentType` anymore. `context.ContentType` is already set

src/Microsoft.AspNet.Mvc.Abstractions/Formatters/OutputFormatterCanWriteContext.cs
dougbu: Well, could you at least spell "indicating" correctly?

src/Microsoft.AspNet.Mvc.Abstractions/Formatters/OutputFormatterCanWriteContext.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// <summary>
+        /// Gets or sets a value inidicating that content-negotiation could not find a formatter based on the 
+        /// information on the <see cref="Http.HttpRequest"/>.
+        /// </summary>
+        public virtual bool? FailedContentNegotiation { get; protected set; }
+
+        /// <summary>
+        /// Gets or sets the object to write to the response.
+        /// </summary>
+        public virtual object Object { get; protected set; }
```
dougbu: Why does `CanWriteResult()` need this?

rynowak: I would love to remove this, but that isn't the topic of this PR. You'll find a few examples of formatters that use it.

src/Microsoft.AspNet.Mvc.Abstractions/Formatters/OutputFormatterCanWriteContext.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        /// <summary>
+        /// Gets or sets the <see cref="Type"/> of the object to write to the response.
+        /// </summary>
+        public virtual Type ObjectType { get; protected set; }
```
dougbu: Why aren't the setters `public`?

src/Microsoft.AspNet.Mvc.Abstractions/Formatters/OutputFormatterCanWriteContext.cs
rynowak: It's not common, and wasn't supported in WebAPI

src/Microsoft.AspNet.Mvc.Core/ApiExplorer/IApiResponseFormatMetadataProvider.cs
```diff
@@ -1,8 +1,8 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
 using System.Collections.Generic;
+using Microsoft.AspNet.Mvc.Formatters;
```
dougbu: This is a bit iffy because Roslyn doesn't detect the namespace is used in doc comments. Will trip up the next person who removes supposedly-unused `using`s.

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
```diff
@@ -49,26 +49,23 @@ public IList<MediaTypeHeaderValue> SupportedMediaTypes
         /// <summary>
         /// Returns a value indicating whether or not the given type can be written by this serializer.
         /// </summary>
-        /// <param name="declaredType">The declared type.</param>
-        /// <param name="runtimeType">The runtime type.</param>
+        /// <param name="type">The object type.</param>
         /// <returns><c>true</c> if the type can be written, otherwise <c>false</c>.</returns>
```
dougbu: Do these comments add any value over `<inheritdoc />`?

src/Microsoft.AspNet.Mvc.Abstractions/Formatters/OutputFormatterCanWriteContext.cs
rynowak: It's not intended that formatters set these properties.

src/Microsoft.AspNet.Mvc.Core/ApiExplorer/IApiResponseFormatMetadataProvider.cs
rynowak: The namespace is used in the code, I didn't just do this as cleanup.

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
rynowak: This would happen if the user overrides `CanWriteResult` and returns true without setting a content-type. There is a phase in conneg where `context.ContentType` will be `null`.

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
dougbu: This is fine but double-check no later code reconfirms `SupportedMediaTypes.Any(supportedType => supportedType.IsSubsetOf(context.ContentType))` or similar. You're making `ContentType` more specific than it was.

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatterContext.cs
```diff
@@ -0,0 +1,52 @@
...
+        public new object Object
+        {
+            get { return base.Object; }
+            set { base.Object = value; }
+        }
+
+        /// <summary>
+        /// Gets or sets the <see cref="Type"/> of the object to write to the response.
+        /// </summary>
+        public new Type ObjectType
```
dougbu: Could you avoid `new` here if the properties in the `abstract` classes were `abstract` instead of `virtual`?

src/Microsoft.AspNet.Mvc.Abstractions/Formatters/OutputFormatterCanWriteContext.cs
rynowak: Sure

src/Microsoft.AspNet.Mvc.Core/Infrastructure/ObjectResultExecutor.cs
```diff
@@ -272,7 +276,8 @@ public virtual Task ExecuteAsync(ActionContext context, ObjectResult result)
 
             foreach (var formatter in formatters)
             {
-                if (formatter.CanWriteResult(formatterContext, contentType: null))
+                formatterContext.ContentType = null;
```
dougbu: Wouldn't it be a bug if `CanWriteResult()` updated the `formatterContext` then returned `false`? If I'm correct, would prefer a `Debug.Assert(formatterContext.ContentType == null);` at the bottom of this loop.

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
rynowak: It's not inherited, it's `new`

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatterContext.cs
rynowak: That would require making them get/set.

src/Microsoft.AspNet.Mvc.Core/Infrastructure/ObjectResultExecutor.cs
rynowak: The bug would be in a user-implemented formatter at that point. I don't think it's sensible to throw or assert here when we can just do the right thing always.

src/Microsoft.AspNet.Mvc.Core/Infrastructure/ObjectResultExecutor.cs
```diff
@@ -314,22 +319,20 @@ public virtual Task ExecuteAsync(ActionContext context, ObjectResult result)
...
-                // Loop through each of the formatters and see if any one will support this
-                // mediaType Value.
-                selectedFormatter = formatters.FirstOrDefault(
-                    formatter => formatter.CanWriteResult(formatterContext, contentType));
-                if (selectedFormatter != null)
+                foreach (var formatter in formatters)
                 {
-                    // we found our match.
-                    break;
+                    formatterContext.ContentType = contentType;
```
dougbu: Suspect `Debug.Assert(formatterContext.ContentType == contentType); at the bottom of this loop would match the intent better. Then can init `ContentType` once.

src/Microsoft.AspNet.Mvc.Core/Infrastructure/ObjectResultExecutor.cs
```diff
@@ -366,11 +369,19 @@ public virtual Task ExecuteAsync(ActionContext context, ObjectResult result)
                 throw new ArgumentNullException(nameof(acceptableContentTypes));
             }
 
-            var selectedFormatter = formatters.FirstOrDefault(
-                formatter => acceptableContentTypes.Any(
-                    contentType => formatter.CanWriteResult(formatterContext, contentType)));
+            foreach (var formatter in formatters)
+            {
+                foreach (var contentType in acceptableContentTypes)
```
dougbu: Thought you were going to make this the outer loop since users usually control the `[Produces]` list directly?

rynowak: That's not part of this refactor.

src/Microsoft.AspNet.Mvc.Formatters.Xml/XmlDataContractSerializerOutputFormatter.cs
```diff
@@ -185,41 +172,37 @@ protected virtual DataContractSerializer CreateSerializer(Type type)
...
-                if (wrappingType != null && wrappingType != resolvedType)
+                if (wrappingType != null && wrappingType != context.ObjectType)
                 {
-                    var wrapperProvider = WrapperProviderFactories.GetWrapperProvider(
-                                                            new WrapperProviderContext(
-                                                                                declaredType: resolvedType,
-                                                                                isSerialization: true));
+                    var wrapperProvider = WrapperProviderFactories.GetWrapperProvider(new WrapperProviderContext(
+                        declaredType: context.ObjectType,
+                        isSerialization: true));
```
dougbu: :clap:

src/Microsoft.AspNet.Mvc.Formatters.Xml/XmlDataContractSerializerOutputFormatter.cs
```diff
@@ -11,6 +11,7 @@
 using System.Xml;
 using Microsoft.AspNet.Mvc.Formatters.Xml;
 using Microsoft.AspNet.Mvc.Formatters.Xml.Internal;
+using Microsoft.AspNet.Mvc.Internal;
```
dougbu: Just to confirm: Is this for `TaskCache`?

rynowak: Yes

test/Microsoft.AspNet.Mvc.Core.Test/Formatters/OutputFormatterTests.cs
```diff
@@ -184,18 +163,17 @@ public void GetSupportedContentTypes_ReturnsNull_ForUnsupportedType()
...
             var formatter = new TypeSpecificFormatter();
-
             formatter.SupportedMediaTypes.Add(MediaTypeHeaderValue.Parse("application/json"));
-
             formatter.SupportedTypes.Add(typeof(int));
 
+            var context = new OutputFormatterContext();
+            context.ObjectType = typeof(string);
+            context.Object = "Hello, world!";
+            context.ContentType = formatter.SupportedMediaTypes[0];
```
dougbu: Prefer the consistent use of object init syntax elsewhere in this PR.

dougbu:
:watch: for experiment discussed offline

test/Microsoft.AspNet.Mvc.Core.Test/Formatters/OutputFormatterTests.cs
rynowak: We changed it to be a constructor, so that's used throughout.

rynowak:
Updated

src/Microsoft.AspNet.Mvc.Core/Infrastructure/ObjectResultExecutor.cs
dougbu: :ok_hand:

test/Microsoft.AspNet.Mvc.Core.Test/Formatters/OutputFormatterTests.cs
```diff
@@ -50,8 +50,8 @@ public static IEnumerable<object[]> SelectResponseCharacterEncodingData
             // Arrange
             var mockHttpContext = new Mock<HttpContext>();
             var httpRequest = new DefaultHttpContext().Request;
-            httpRequest.Headers["Accept-Charset"] = acceptCharsetHeaders;
-            httpRequest.ContentType = "application/acceptCharset;charset=" + requestEncoding;
+            httpRequest.Headers[HeaderNames.AcceptCharset] = acceptCharsetHeaders;
+            httpRequest.Headers[HeaderNames.Accept] = "application/acceptCharset;charset=" + requestEncoding;
```
dougbu: This isn't the same test anymore. Now giving the formatter something it should pay attention to. Suggest just removing the `ContentType` assignment; don't add an `Accept` header.

test/Microsoft.AspNet.Mvc.Core.Test/Formatters/OutputFormatterTests.cs
```diff
@@ -62,9 +62,10 @@ public static IEnumerable<object[]> SelectResponseCharacterEncodingData
 
             var formatterContext = new OutputFormatterContext()
```
dougbu: There's still an `OutputFormatterContext` somewhere?

dougbu:
Product code looks good but the test code hasn't caught up yet. Bang that into shape and you're good to go. :shipit: once the CI is green.

Let me know if you want one last pass of the test code once you've updated it.

test/Microsoft.AspNet.Mvc.Core.Test/Formatters/OutputFormatterTests.cs
rynowak: Yeah this is wierd, my VS missed these files somehow. I'm working on fixing it.

test/Microsoft.AspNet.Mvc.Core.Test/Formatters/OutputFormatterTests.cs
rynowak: The formatter used to choose the encoding from the request `ContentType` which was removed in this PR. I'll change `context.ContentType` to be `application/acceptCharset;`

src/Microsoft.AspNet.Mvc.Abstractions/Formatters/OutputFormatterWriteContext.cs
```diff
@@ -0,0 +1,37 @@
...
+
+            HttpContext = httpContext;
+            ObjectType = objectType;
+            Object = @object;
+        }
+
+        /// <summary>
+        /// Gets or sets the <see cref="HttpContext"/> context associated with the current operation.
+        /// </summary>
+        public virtual HttpContext HttpContext { get; protected set; }
```
NTaylorMullen: Still need to be `protected set`?

src/Microsoft.AspNet.Mvc.Abstractions/Formatters/OutputFormatterCanWriteContext.cs
NTaylorMullen: Could do away with the various `protected set`s if you added a ctor to this context object too.

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
```diff
@@ -215,16 +215,29 @@ public virtual void WriteResponseHeaders(OutputFormatterContext context)
...
+        public virtual void WriteResponseHeaders(OutputFormatterWriteContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
 
             var response = context.HttpContext.Response;
-            response.ContentType = selectedMediaType.ToString();
+            response.ContentType = context.ContentType.ToString();
```
NTaylorMullen: `context.ContentType` may be `null`

src/Microsoft.AspNet.Mvc.Formatters.Xml/XmlDataContractSerializerOutputFormatter.cs
NTaylorMullen: lies :smile:

NTaylorMullen:
Smaller comments. :shipit:

src/Microsoft.AspNet.Mvc.Abstractions/Formatters/OutputFormatterWriteContext.cs
rynowak: It's nice this way for mockability.

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
rynowak: Fixed

PR #3356, opened 2015-10-20
Use buffer pooling in IOutputFormatters

rynowak:
Allocation data from 3000 request to a simple API site. Post JSON -> Return 201 Created with JSON.

**Before**
image

**After**
image

The overall reduction in allocations here is about 16mb - based on a total of 106mb. The vast majority of the remaining `char[]` and `byte[]` allocations come from input formatters, which I'll address next.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Abstractions/Formatters/OutputFormatterWriteContext.cs
+++ b/src/Microsoft.AspNet.Mvc.Abstractions/Formatters/OutputFormatterWriteContext.cs
@@ -2,6 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.IO;
+using System.Text;
 using Microsoft.AspNet.Http;
 
 namespace Microsoft.AspNet.Mvc.Formatters
@@ -15,13 +17,19 @@ namespace Microsoft.AspNet.Mvc.Formatters
         /// Creates a new <see cref="OutputFormatterWriteContext"/>.
         /// </summary>
         /// <param name="httpContext">The <see cref="Http.HttpContext"/> for the current request.</param>
+        /// <param name="writerFactory">The delegate used to create a <see cref="TextWriter"/> for writing the response.
         /// <param name="objectType">The <see cref="Type"/> of the object to write to the response.</param>
         /// <param name="@object">The object to write to the response.</param>
-        public OutputFormatterWriteContext(HttpContext httpContext, Type objectType, object @object)
+        public OutputFormatterWriteContext(HttpContext httpContext, Func<Stream, Encoding, TextWriter> writerFactory, Type objectType, object @object)
         {
             if (httpContext == null)
             {
                 throw new ArgumentNullException(nameof(httpContext));
+            }
+
+            if (writerFactory == null)
+            {
+                throw new ArgumentNullException(nameof(writerFactory));
             }
 
             HttpContext = httpContext;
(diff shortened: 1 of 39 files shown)


src/Microsoft.AspNet.Mvc.Abstractions/Formatters/OutputFormatterWriteContext.cs
```diff
@@ -14,5 +17,10 @@ public abstract class OutputFormatterWriteContext : OutputFormatterCanWriteConte
         /// Gets or sets the <see cref="HttpContext"/> context associated with the current operation.
         /// </summary>
         public virtual HttpContext HttpContext { get; protected set; }
+
+        /// <summary>
+        /// Gets or sets a delegate used to create a <see cref="TextWriter"/> for writing the response.
+        /// </summary>
+        public virtual Func<Stream, Encoding, TextWriter> WriterFactory { get; protected set; }
```
rynowak: Considered creating a delegate type for this... thoughts?

src/Microsoft.AspNet.Mvc.Formatters.Xml/XmlDataContractSerializerOutputFormatter.cs
```diff
@@ -148,17 +148,20 @@ protected virtual DataContractSerializer CreateSerializer(Type type)
...
+        /// <see cref="WriterSettings"/>.
         /// </summary>
-        /// <param name="writeStream">The stream on which the XmlWriter should operate on.</param>
+        /// <param name="writer">
+        /// The underlying <see cref="TextWriter"/> which the <see cref="XmlWriter"/> should write to.
+        /// </param>
         /// <returns>A new instance of <see cref="XmlWriter"/></returns>
         public virtual XmlWriter CreateXmlWriter(
-            Stream writeStream,
+            TextWriter writer,
```
rynowak: Made this change so we could do the 'nested using' pattern. We don't want to have problems if the user throws here and doesn't dispose the writer.

src/Microsoft.AspNet.Mvc.Formatters.Xml/XmlDataContractSerializerOutputFormatter.cs
```diff
@@ -182,24 +186,26 @@ public override Task WriteResponseBodyAsync(OutputFormatterWriteContext context)
...
+                    isSerialization: true));
 
-                // Wrap the object only if there is a wrapping type.
-                if (wrappingType != null && wrappingType != context.ObjectType)
-                {
-                    var wrapperProvider = WrapperProviderFactories.GetWrapperProvider(new WrapperProviderContext(
-                        declaredType: context.ObjectType,
-                        isSerialization: true));
+                value = wrapperProvider.Wrap(value);
+            }
```
rynowak: Pulled this code out of the using.

test/Microsoft.AspNet.Mvc.WebApiCompatShimTest/TestHttpResponseStreamWriterFactory.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.IO;
+using System.Text;
+using Microsoft.AspNet.Mvc.Infrastructure;
+
+namespace Microsoft.AspNet.Mvc
+{
+    public class TestHttpResponseStreamWriterFactory : IHttpResponseStreamWriterFactory
```
rynowak: Had to copy-paste this from testcommon, which the shim doesn't/can't reference.

rynowak:
Will resend based on dev

PR #3362, opened 2015-10-20
Allow enums to be serialized by SessionStateTempDataProvider

### LINKED ISSUE #3141: TempData support for Enum types
`SessionStateTempDataProvider` does not support `Enum` types but enums always can be converted to the underlying type and serialized to JSON. May be temp data provider should support enums?

pranavkm:
Fixes #3141

### DIFF
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/SessionStateTempDataProvider.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/SessionStateTempDataProvider.cs
@@ -9,7 +9,6 @@ using System.IO;
 using System.Linq;
 using System.Reflection;
 using Microsoft.AspNet.Http;
-using Microsoft.AspNet.Http.Features;
 using Microsoft.Extensions.Internal;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Bson;
@@ -73,10 +72,13 @@ namespace Microsoft.AspNet.Mvc.ViewFeatures
                     tempDataDictionary = _jsonSerializer.Deserialize<Dictionary<string, object>>(writer);
                 }
 
-                var convertedDictionary = new Dictionary<string, object>(tempDataDictionary, StringComparer.OrdinalIgnoreCase);
+                var convertedDictionary = new Dictionary<string, object>(
+                    tempDataDictionary,
+                    StringComparer.OrdinalIgnoreCase);
                 foreach (var item in tempDataDictionary)
                 {
                     var jArrayValue = item.Value as JArray;
+                    var jObjectValue = item.Value as JObject;
                     if (jArrayValue != null && jArrayValue.Count > 0)
                     {
                         var arrayType = jArrayValue[0].Type;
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewFeatures/SessionStateTempDataProviderTest.cs
```diff
@@ -185,7 +186,7 @@ public void SaveAndLoad_IntCanBeStoredAndLoaded()
             var TempData = testProvider.LoadTempData(context);
 
             // Assert
-            var intVal = Convert.ToInt32(TempData["int"]);
+            var intVal = Assert.IsType<int>(TempData["int"]);
```
pranavkm: @Eilon, casting for `int`s was currently broken.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/SessionStateTempDataProvider.cs
```diff
@@ -128,6 +129,16 @@ public class SessionStateTempDataProvider : ITempDataProvider
                             throw new InvalidOperationException(message);
                         }
                     }
+                    else if (item.Value is long)
```
pranavkm: Thoughts?

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewFeatures/SessionStateTempDataProviderTest.cs
dougbu: What breaks if `int`s round-trip as `long`s with the same numeric value?

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/SessionStateTempDataProvider.cs
dougbu: Meh. Unclear why we'd guess the original type was `int` and not `short` et cetera.

dougbu:
:watch:

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewFeatures/SessionStateTempDataProviderTest.cs
pranavkm: The casting.
```C#
TempData["foo"] = 1;
...
var id = (int)TempData["foo"];
```

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/SessionStateTempDataProvider.cs
pranavkm: You're far more likely to use `int`s than any other data type.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewFeatures/SessionStateTempDataProviderTest.cs
Eilon: Right, can't unbox a boxed `long` to an `int`.

Eilon:
Looks :shipit: to me...

PR #3363, opened 2015-10-20
Added DiagnosticSource for filters

ajaybhargavb:
#3279 

@rynowak @avanderhoorn

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs
@@ -8,6 +8,7 @@ using System.Runtime.ExceptionServices;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Mvc.Abstractions;
 using Microsoft.AspNet.Mvc.Core;
+using Microsoft.AspNet.Mvc.Diagnostics;
 using Microsoft.AspNet.Mvc.Filters;
 using Microsoft.AspNet.Mvc.Formatters;
 using Microsoft.AspNet.Mvc.Infrastructure;
@@ -288,7 +289,15 @@ namespace Microsoft.AspNet.Mvc.Controllers
             var current = _cursor.GetNextFilter<IAuthorizationFilter, IAsyncAuthorizationFilter>();
             if (current.FilterAsync != null)
             {
+                _diagnosticSource.BeforeOnAuthorizationAsync(
+                    _authorizationContext,
+                    current.FilterAsync);
+
                 await current.FilterAsync.OnAuthorizationAsync(_authorizationContext);
+
+                _diagnosticSource.AfterOnAuthorizationAsync(
+                    _authorizationContext,
+                    current.FilterAsync);
 
                 if (_authorizationContext.Result == null)
                 {
@@ -302,7 +311,15 @@ namespace Microsoft.AspNet.Mvc.Controllers
             }
             else if (current.Filter != null)
             {
+                _diagnosticSource.BeforeOnAuthorization(
+                    _authorizationContext,
+                    current.Filter);
+
                 current.Filter.OnAuthorization(_authorizationContext);
+
+                _diagnosticSource.AfterOnAuthorization(
+                    _authorizationContext,
+                    current.Filter);
 
                 if (_authorizationContext.Result == null)
                 {
@@ -358,9 +375,18 @@ namespace Microsoft.AspNet.Mvc.Controllers
             {
                 if (item.FilterAsync != null)
                 {
+                    _diagnosticSource.BeforeOnResourceExecution(
+                        _resourceExecutingContext,
+                        item.FilterAsync);
+
                     await item.FilterAsync.OnResourceExecutionAsync(
                         _resourceExecutingContext,
                         InvokeResourceFilterAsync);
+
+                    _diagnosticSource.AfterOnResourceExecution(
+                        _resourceExecutingContext.ActionDescriptor,
+                        _resourceExecutedContext,
+                        item.FilterAsync);
 
                     if (_resourceExecutedContext == null)
                     {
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs
```diff
@@ -241,6 +241,17 @@ private IFilterMetadata[] GetFilters()
...
+            if (_diagnosticSource.IsEnabled("Microsoft.AspNet.Mvc.RegisteredFilters"))
+            {
+                _diagnosticSource.Write(
+                    "Microsoft.AspNet.Mvc.RegisteredFilters",
+                    new
+                    {
+                        actionDescriptor = ActionContext.ActionDescriptor,
+                        filterProviderContext = context
+                    });
+            }
```
rynowak: This is totally out of place, what is this for?

src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs
```diff
@@ -301,8 +312,30 @@ private Task InvokeAllAuthorizationFiltersAsync()
...
             else if (current.Filter != null)
             {
+                if (_diagnosticSource.IsEnabled("Microsoft.AspNet.Mvc.BeforeOnAuthorization"))
+                {
+                    _diagnosticSource.Write(
+                        "Microsoft.AspNet.Mvc.BeforeOnAuthorization",
+                        new
+                        {
+                            actionDescriptor = _authorizationContext.ActionDescriptor,
+                            authorizationContext = _authorizationContext
```
rynowak: Include the filter here as well

src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs
```diff
@@ -357,10 +390,34 @@ private Task InvokeAllResourceFiltersAsync()
...
 
+                    if (_diagnosticSource.IsEnabled("Microsoft.AspNet.Mvc.AfterOnResourceExecution"))
+                    {
+                        _diagnosticSource.Write(
+                            "Microsoft.AspNet.Mvc.AfterOnResourceExecution",
+                            new
+                            {
+                                actionDescriptor = _resourceExecutingContext.ActionDescriptor,
+                                resourceExecutingContext = _resourceExecutingContext,
+                                resourceExecutedContext = _resourceExecutedContext
```
rynowak: Isn't this always null here?

src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs
rynowak: IN all of these

src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs
```diff
@@ -357,10 +390,34 @@ private Task InvokeAllResourceFiltersAsync()
...
                         InvokeResourceFilterAsync);
 
+                    if (_diagnosticSource.IsEnabled("Microsoft.AspNet.Mvc.AfterOnResourceExecution"))
+                    {
+                        _diagnosticSource.Write(
+                            "Microsoft.AspNet.Mvc.AfterOnResourceExecution",
+                            new
+                            {
+                                actionDescriptor = _resourceExecutingContext.ActionDescriptor,
+                                resourceExecutingContext = _resourceExecutingContext,
```
rynowak: I think you can leave this out in the 'after'

src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs
ajaybhargavb: At this point the `FilterProviderContext` contains a list of `FIlterItem` which provides the FilterDescriptors and Filters

rynowak: but what is this for?

ajaybhargavb: I thought @avanderhoorn needed the list of registered filters separately.

src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs
ajaybhargavb: Won't it have the updated `Result` property?

src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs
rynowak: This event on it's own is out of place, all of these other events already provide this.

src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs
rynowak: `resourceExecutedContext` also has the updated `.Result` property

ajaybhargavb:
Updated.

avanderhoorn:
In terms of working out `ExceptionFilter` vs `ActionFilter` vs `ResultFilter` etc is the recommended approach here to derive this from the `type` of the filter?

rynowak:
> In terms of working out ExceptionFilter vs ActionFilter vs ResultFilter etc is the recommended approach here to derive this from the type of the filter?

From the name of the event - if you want to consolidate that and track it as 'generic filter' feel free.

avanderhoorn:
Thanks

ajaybhargavb:
Updated (with extension methods)

src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs
```diff
@@ -288,8 +289,18 @@ private Task InvokeAllAuthorizationFiltersAsync()
             var current = _cursor.GetNextFilter<IAuthorizationFilter, IAsyncAuthorizationFilter>();
             if (current.FilterAsync != null)
             {
+                _diagnosticSource.BeforeOnAuthorizationAsync(
+                    _authorizationContext.ActionDescriptor,
+                    _authorizationContext,
+                    current.FilterAsync);
```
rynowak: Simplify this method to just take the `_authorizationContext` and `current.FilterAsync`

rynowak:
:shipit: after removing a parameter from all of the methods :+1:

PR #3367, opened 2015-10-21
Fixes for failing OutputFormatter functional tests

rynowak:
The missing piece here is is that StringOutputFormatter needs to set the
ContentType so that it gets overridden. The problem is that the formatter
is likely called with something like application/json, but decides to
write a string anyway. So because we're saying 'yes we can write' we also
need to override what we're writing.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/Formatters/StringOutputFormatter.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Formatters/StringOutputFormatter.cs
@@ -31,13 +31,9 @@ namespace Microsoft.AspNet.Mvc.Formatters
 
             // Ignore the passed in content type, if the object is string
             // always return it as a text/plain format.
-            if (context.ObjectType == typeof(string))
+            if (context.ObjectType == typeof(string) || context.Object is string)
             {
-                return true;
-            }
-
-            if (context.Object is string)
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.Mvc.Core/Formatters/StringOutputFormatter.cs
```diff
@@ -31,13 +31,9 @@ public override bool CanWriteResult(OutputFormatterCanWriteContext context)
 
             // Ignore the passed in content type, if the object is string
             // always return it as a text/plain format.
-            if (context.ObjectType == typeof(string))
-            {
-                return true;
-            }
-
-            if (context.Object is string)
+            if (context.ObjectType == typeof(string) || context.Object is string)
```
dougbu: Curious: When is the RHS of the `||` needed? `ObjectType` can't be `object` when `Object` is non-`null`, can it?

rynowak: Yeah, that's true that it shouldn't be needed. Being cavalier about removing things is what broke these in the first place. Let's sync up on removing `context.Object` tomorrow if possible and revisit this as a forcing function.

PR #3378, opened 2015-10-21
Fix functional tests

pakrym:
After https://github.com/aspnet/Hosting/pull/435 
@rynowak @davidfowl

### DIFF
--- a/test/Microsoft.AspNet.Mvc.Core.Test/Infrastructure/DefaultAssemblyProviderTests.cs
+++ b/test/Microsoft.AspNet.Mvc.Core.Test/Infrastructure/DefaultAssemblyProviderTests.cs
@@ -253,8 +253,7 @@ namespace Microsoft.AspNet.Mvc.Infrastructure
 
             private static ILibraryManager GetLibraryManager()
             {
-                return CallContextServiceLocator.Locator.ServiceProvider
-                    .GetRequiredService<ILibraryManager>();
+                return PlatformServices.Default.LibraryManager;
             }
         }
     }
--- a/test/Microsoft.AspNet.Mvc.FunctionalTests/MvcTestFixture.cs
+++ b/test/Microsoft.AspNet.Mvc.FunctionalTests/MvcTestFixture.cs
@@ -91,8 +91,7 @@ namespace Microsoft.AspNet.Mvc.FunctionalTests
             Assembly startupAssembly,
             Func<IServiceCollection, IServiceProvider> buildServices)
         {
-            var applicationServices = CallContextServiceLocator.Locator.ServiceProvider;
-            var libraryManager = applicationServices.GetRequiredService<ILibraryManager>();
+            var libraryManager = PlatformServices.Default.LibraryManager;
 
             // When an application executes in a regular context, the application base path points to the root
             // directory where the application is located, for example .../samples/MvcSample.Web. However, when
@@ -104,7 +103,7 @@ namespace Microsoft.AspNet.Mvc.FunctionalTests
             var library = libraryManager.GetLibrary(applicationName);
             var applicationRoot = Path.GetDirectoryName(library.Path);
 
-            var applicationEnvironment = applicationServices.GetRequiredService<IApplicationEnvironment>();
+            var applicationEnvironment = PlatformServices.Default.Application;
 
             return (services) =>
             {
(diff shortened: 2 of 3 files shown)


rynowak:
If you do a search, I think there are two other places where CallContextSP is used in MVC

test/Microsoft.AspNet.Mvc.FunctionalTests/TestHelper.cs
```diff
@@ -60,7 +60,7 @@ public static TestServer CreateServer(Action<IApplicationBuilder> builder, strin
 
             // Get current IApplicationEnvironment; likely added by the host.
```
rynowak: This comment isn't relevant anymore

test/Microsoft.AspNet.Mvc.FunctionalTests/TestHelper.cs
```diff
@@ -58,9 +59,8 @@ public static TestServer CreateServer(Action<IApplicationBuilder> builder, strin
         {
             applicationPath = applicationPath ?? WebsitesDirectoryPath;
 
-            // Get current IApplicationEnvironment; likely added by the host.
             var provider = services.BuildServiceProvider();
-            var originalEnvironment = provider.GetRequiredService<IApplicationEnvironment>();
+            var originalEnvironment = PlatformServices.Default.Application;
```
dougbu: We get `originalEnvironment` to wrap it in our `TestApplicationEnvironment`. Will the `TestApplicationEnvironment` instance added just below be picked up correctly?

test/Microsoft.AspNet.Mvc.FunctionalTests/TestHelper.cs
```diff
@@ -47,7 +47,8 @@ public static TestServer CreateServer(Action<IApplicationBuilder> builder, strin
         {
             return TestServer.Create(
                 builder,
-                services => AddTestServices(services, applicationWebSiteName, applicationPath, configureServices));
+                configureServices: null,
+                configureHostServices: services => AddTestServices(services, applicationWebSiteName, applicationPath, configureServices));
```
dougbu: Harder to read here. Please wrap the parameters to their own lines.

dougbu:
+1 to @rynowak's comment. Please fix [`DefaultAssemblyProviderTests`](https://github.com/aspnet/Mvc/blob/dev/test/Microsoft.AspNet.Mvc.Core.Test/Infrastructure/DefaultAssemblyProviderTests.cs#L256) as well

test/Microsoft.AspNet.Mvc.FunctionalTests/TestHelper.cs
pakrym: Yes, new environment will be picked up, it is executed in WebHostBuilder.BuildHostingServices and default services are added using TryAdd so it won't be overwritten

test/Microsoft.AspNet.Mvc.FunctionalTests/TestHelper.cs
pakrym: What do you mean?

AddTestServices(
services, 
applicationWebSiteName, applicationPath, configureServices)

dougbu: Yes. With relative indentation, it should look like:
``` c#
return TestServer.Create(
    builder,
    configureServices: null,
    configureHostServices: services => AddTestServices(
        services,
        applicationWebSiteName,
        applicationPath,
        configureServices));
```

test/Microsoft.AspNet.Mvc.Core.Test/Infrastructure/DefaultAssemblyProviderTests.cs
```diff
@@ -253,8 +253,7 @@ private static IEnumerable<string> GetAssemblies(IEnumerable<string> libraries)
 
             private static ILibraryManager GetLibraryManager()
             {
-                return CallContextServiceLocator.Locator.ServiceProvider
-                    .GetRequiredService<ILibraryManager>();
+                return PlatformServices.Default.LibraryManager;
```
dougbu: thx

dougbu:
:shipit: once you're able to verify the complete build succeeds.

davidfowl:
:shipit:

Issue #3307, opened 2015-10-12
Support relative Layout values and view paths


mikeandersun:
I Have one controller action that return Views according to an input parameter like this code:

        [HttpGet]
        public async Task<IActionResult> RenderAViewInFoler(string foldername)
        {...
I can set the view path grammatically but my issue is setting the `layout.cshtml` and ` @Html.Partial("_footer")` that come from where ever `RazorViewEngine` returns.
I need to `layout.cshtml` and ` @Html.Partial("_footer")` come from the `foldername` path.
For temprory solution I set the path of those in View file like this:
`@{
     Layout = "~/foldername/_layout.cshtml";
}
`
FYI, in MVC 5 we could set full qualified name of layout for `View(...)` method.

dougbu:
Could you give an example of a "fully qualified name" that doesn't look like `"~/foldername/_layout.cshtml"`?

mikeandersun:
"~/folder1/_layout.cshtml"
"~/folder2/_layout.cshtml"
What all I needs is that when i set a path for a view like `View("~/folder1/View1.cshtml)"` then all partial view and the layout come from `folder1` not come from `Views/shared` or others place that searched by RazorViewEngine.
It seam RazorViewEngine  doesn't run per request to a controller action.

dougbu:
#### tl;dr
The view and layout systems in MVC 5 and 6 are slightly different. You've found one of the differences -- `Controller` in MVC 6 lacks the `View(string viewName, string masterName)` overload available in MVC 5. In addition `Layout` values are not handled as relative paths as they are in MVC 5: They're either view names or fully-qualified paths (similar to `masterName` in the previous version).

We're going to look at adding relative path support throughout this system in MVC 6, providing the best of both words and avoiding odd searches that check for `/Views/Home/About.cshtml.cshtml`.

#### MVC 5 details
The MVC 5 lookup is approximately:

1. Find view page.
  1. If `viewName` passed to `View()` is `null`, search `~/Views/{controller}/{action}.cshtml` and `~/Views/Shared/{action}.cshtml`. (Area lookup checks `~/Areas/{area}/Views/{controller}/{action}.cshtml` and `~/Areas/{area}/Views/Shared/{action}.cshtml` first. I ignore areas from here on.) 
  2. Otherwise if `viewName` is fully-qualified, use given path.
[...]

PR #3380, opened 2015-10-21
Add relative view path support

dougbu:
- #3307 ~~WIP~~

~~This is not for current milestone. Getting this out for discussion if you're interested.~~


dougbu:
bump
:arrows_counterclockwise: (rebased FYI)

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngine.cs
```diff
@@ -233,14 +235,43 @@ public virtual IEnumerable<string> AreaViewLocationFormats
...
         {
-            if (IsApplicationRelativePath(pageName))
+            var applicationRelativePath = pageName;
+            if (!IsApplicationRelativePath(applicationRelativePath) && IsRelativePath(applicationRelativePath))
             {
-                var applicationRelativePath = pageName;
-                if (!pageName.EndsWith(ViewExtension, StringComparison.OrdinalIgnoreCase))
+                // Given a relative path that is not yet app-relative.
+                // Interpret given path relative to currently-executing view, if any.
+                var executingPath = (context as ViewContext)?.ExecutingFilePath;
```
pranavkm: Is this to support relative paths for partials? If it's specifically meant for Layouts, could we change `FindPage` to specify the `basePath` and calculate the Layout using that value?

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngine.cs
```diff
@@ -233,14 +235,43 @@ public virtual IEnumerable<string> AreaViewLocationFormats
...
+                }
+                else
+                {
+                    applicationRelativePath = executingPath + "/" + applicationRelativePath;
+                }
+            }
+
+            if (IsApplicationRelativePath(applicationRelativePath))
+            {
+                // **Reviewers**: Removed code converted ~/Views/Folder/View to ~/View/Folder/View.cshtml. Discussed w/
```
pranavkm: :+1:

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngine.cs
```diff
@@ -361,5 +392,16 @@ private static bool IsApplicationRelativePath(string name)
             Debug.Assert(!string.IsNullOrEmpty(name));
             return name[0] == '~' || name[0] == '/';
         }
+
+        private static bool IsRelativePath(string name)
+        {
+            Debug.Assert(!string.IsNullOrEmpty(name));
+
+            // **Reviewers**: Though ./ViewName looks like a relative path, checking for '/' leads to problems with
```
pranavkm: Do we need a better data type to represents paths here so that things like view components can be more explicit that the path is relative?

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngine.cs
dougbu: Yes, this is meant to support relative paths for partials, top-level views, top-level views for view components, ...

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngine.cs
dougbu: I think of the view component case as _using_ a feature that's available to everyone rater than as a unique requirement. Further view searches with slashes in the name work in MVC 5.x.y.

dougbu:
:up::date: no longer in `[Design]`

test/Microsoft.AspNet.Mvc.Razor.Test/RazorViewEngineTest.cs
```diff
@@ -345,6 +348,227 @@ public void FindView_UsesAreaViewLocationFormat_IfRouteContainsArea()
...
+            viewFactory
+                .Setup(v => v.GetView(viewEngine, page, false))
+                .Returns(Mock.Of<IView>());
+
+            // Act
+            var result = viewEngine.FindView(context, viewName);
+
+            // Assert
+            Assert.True(result.Success);
+            Assert.Equal(viewName, result.ViewName);    // Result gets original view name.
```
dougbu: Getting the original view name is not a behaviour change. `RazorViewEngine` does this consistently whether it uses view locations or not.

src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
```diff
@@ -162,10 +163,21 @@ private Task RenderPageCoreAsync(IRazorPage page, ViewContext context)
...
                     // Copy the layout value from the previous view start (if any) to the current.
                     viewStart.Layout = layout;
+
                     await RenderPageCoreAsync(viewStart, context);
+
                     layout = viewStart.Layout;
+                    if (!string.IsNullOrEmpty(layout) && !IsApplicationRelativePath(layout) && IsRelativePath(layout))
+                    {
+                        // Fix up relative layout to be app-relative. Interpret layout relative to this view start.
+                        // Get directory name but do not use Path.GetDirectoryName() to preserve path normalization.
```
pranavkm: Could we commonize this a `CombinePath` method? Pattern's identical to Line 228-231

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngine.cs
```diff
@@ -233,14 +234,30 @@ public virtual IEnumerable<string> AreaViewLocationFormats
...
         {
-            if (IsApplicationRelativePath(pageName))
+            var applicationRelativePath = pageName;
+            if (!IsApplicationRelativePath(applicationRelativePath) && IsRelativePath(applicationRelativePath))
             {
-                var applicationRelativePath = pageName;
-                if (!pageName.EndsWith(ViewExtension, StringComparison.OrdinalIgnoreCase))
+                // Given a relative path (i.e. ending with ".cshtml") that is not yet application-relative (i.e.
+                // starting with "~/" or "/"), interpret given path relative to currently-executing view, if any.
+                var executingPath = (context as ViewContext)?.ExecutingFilePath;
```
pranavkm: This still feels very weird to me - it relies heavily on the way we construct `ViewContext` instances from `ActionContext` which happens to be an implementation detail and not a contract. How did this work for partials in the past? Did we resolve it in the HtmlHelper? 

cc @Eilon for thoughts?

pranavkm:
Could we get functional tests for partials with relative path? In particular partials invoking other partials and specifying Layouts with relative path.

:watch:

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngine.cs
dougbu: Relative paths did not work for partials in the past.

src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
dougbu: Do you mean a `private static` method in this class? If so, sure. But I'd rather not attempt to commonize things between `RazorView` and `RazorViewEngine`.

pranavkm: `private static` method for this type.

dougbu:
:up::date: @Eilon @pranavkm

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewEngines/CompositeViewEngine.cs
```diff
@@ -23,62 +26,69 @@ public CompositeViewEngine(IOptions<MvcViewOptions> optionsAccessor)
...
-            {
-                throw new ArgumentNullException(nameof(partialViewName));
-            }
+            Debug.Assert(!string.IsNullOrEmpty(name));
 
-            return FindView(context, partialViewName, partial: true);
+            // A page path starts with "~" or "/" (if absolute) or ends with ".cshtml" (if relative).
+            return name[0] == '~' ||
+                name[0] == '/' ||
+                name.EndsWith(ViewExtension, StringComparison.OrdinalIgnoreCase);
```
pranavkm: This would be problematic for non-Razor view engines. Maybe we should check if the file has any extension?

dougbu: This was the semantic we designed. Note the existing system sticks `.cshtml` on the end unconditionally.

pranavkm: It works fine for Razor specifically, but we can't use that semantic for the `CompositeViewEngine`. That sounds quite incorrect

dougbu: If @Eilon agrees w/ this concern, `IViewEngine` would need to expose `IsPagePath()` and this class would determine truthiness based on its contained `IViewEngine` implementations. E.g. first `IViewEngine` that says `true` determines the return value.

Of course this is perilous because the view engine that said something was a path (and not a name) will not necessarily be called first when `GetView()` is called. Might need a `IViewEngine HandleAsPath(string path)` method on `IViewEngine` instead.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/HtmlHelper.cs
```diff
@@ -542,7 +542,16 @@ public Task RenderPartialAsync(string partialViewName, object model, ViewDataDic
 
             var newViewData = new ViewDataDictionary(baseViewData, model);
 
-            var viewEngineResult = _viewEngine.FindPartialView(ViewContext, partialViewName);
+            ViewEngineResult viewEngineResult;
+            if (CompositeViewEngine.IsPagePath(partialViewName))
+            {
+                viewEngineResult = _viewEngine.GetView(ViewContext.ExecutingFilePath, partialViewName, isPartial: true);
```
Eilon: change pattern to
``` c#
   var viewEngineResult = _viewEngine.GetView(ViewContext.ExecutingFilePath, partialViewName, isPartial: true);
   if (!viewEngineResult.Success)
       // FindView
   if (!viewEngineResult.Success)
      // fail
```

dougbu:
:up::date: to agreed design. Working on tests.

dougbu:
:up::date: once again no longer in `[Design]`

src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
```diff
@@ -233,13 +238,18 @@ private Task RenderPageCoreAsync(IRazorPage page, ViewContext context)
...
         }
 
-        private IRazorPage GetLayoutPage(ViewContext context, string layoutPath)
+        private IRazorPage GetLayoutPage(ViewContext context, string executingFilePath, string layoutPath)
         {
-            var layoutPageResult = _viewEngine.FindPage(context, layoutPath);
+            var layoutPageResult = _viewEngine.GetPage(executingFilePath, layoutPath, isPartial: true);
+            if (layoutPageResult.Page == null)
+            {
+                layoutPageResult = _viewEngine.FindPage(context, layoutPath, isPartial: true);
```
dougbu: All of the tests work fine. But I'm still working on the details of this `GetPage()` / `GetView()` then `FindPage()` / `FindView()` pattern. Errors may use an empty `SearchedLocations` collection if the `GetX()` method tries and the `FindX()` method doesn't.

Most likely fix is to check `!layoutPageResult.SearchedLocations.Any()` before entering this block. At most one of the `GetX()` and `FindX()` methods will set `SearchedLocations` to anything but `Enumerable.Empty<string>()`.

src/Microsoft.AspNet.Mvc.Razor/IRazorViewEngine.cs
```diff
@@ -11,12 +11,38 @@ namespace Microsoft.AspNet.Mvc.Razor
...
+
+        /// <summary>
+        /// Gets the page with the given <paramref name="pagePath"/>, relative to <paramref name="executingFilePath"/>
+        /// unless <paramref name="pagePath"/> is already absolute.
+        /// </summary>
+        /// <param name="executingFilePath">The absolute path to the currently-executing page, if any.</param>
+        /// <param name="pagePath">The path to the page.</param>
+        /// <param name="isPartial">Determines if the page being found is a partial.</param>
+        /// <returns>The <see cref="RazorPageResult"/> of locating the page.</returns>
+        /// <remarks>See also <see cref="IViewEngine.GetView"/>.</remarks>
```
Eilon: There's a `<seealso>` element, no?

src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
```diff
@@ -233,13 +238,18 @@ private Task RenderPageCoreAsync(IRazorPage page, ViewContext context)
             }
         }
 
-        private IRazorPage GetLayoutPage(ViewContext context, string layoutPath)
+        private IRazorPage GetLayoutPage(ViewContext context, string executingFilePath, string layoutPath)
         {
-            var layoutPageResult = _viewEngine.FindPage(context, layoutPath);
+            var layoutPageResult = _viewEngine.GetPage(executingFilePath, layoutPath, isPartial: true);
```
Eilon: Why is `isPartial` true?

dougbu: Because `FindPage()` previously hard-coded that value when it was successful. More generally because a layout page can't exist separate from the main page.

src/Microsoft.AspNet.Mvc.Razor/IRazorViewEngine.cs
dougbu: Don't see it used much but :ok: Do you prefer `<seealso/>` inside the `<summary/>` or `<remarks/>`?

src/Microsoft.AspNet.Mvc.Razor/IRazorViewEngine.cs
```diff
@@ -11,12 +11,38 @@ namespace Microsoft.AspNet.Mvc.Razor
...
+        /// <paramref name="executingFilePath"/> unless <paramref name="pagePath"/> is already absolute.
+        /// </summary>
+        /// <param name="executingFilePath">The absolute path to the currently-executing page, if any.</param>
+        /// <param name="pagePath">The path to the page.</param>
+        /// <returns>
+        /// The combination of <paramref name="executingFilePath"/> and <paramref name="pagePath"/> if
+        /// <paramref name="pagePath"/> is a relative path. The <paramref name="pagePath"/> value (unchanged)
+        /// otherwise.
+        /// </returns>
+        string MakePathAbsolute(string executingFilePath, string pagePath);
```
pranavkm: `GetAbsolutePath`? `Make` sounds a bit unusual.

src/Microsoft.AspNet.Mvc.Razor/IRazorViewEngine.cs
Eilon: I didn't think it goes inside *any* tag, does it? I thought it's top-level. Using semantically-accurate tags is better than not using semantically-accurate tags...

Eilon: Oh weird MSDN shows it inside the `<summary>` tag. But I think doc comment Intellisense shows it outside of it. I would still use the tag.

src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
Eilon: Not sure I understand. Is the parameter name just bad? What behavior are we trying to get (or avoid) by passing in this seemingly-wrong parameter?

dougbu: The parameter name lines up with the `IsPartial` properties in `IRazorPage` and `RazorView`. I have not changed what pages / views are considered partial, just exposed the possibility of creating a non-partial page using `GetPage()` and `FindPage()`.

FYI only top-level views i.e. those `ViewResultExecutor` processes are _not_ considered partial.

Eilon: My question is a bit different: why would locating a *layout page* involve telling the system that we're looking for a *partial*? Those are two completely different unrelated things. Does it just happen to give the correct behavior as opposed to the "other" behavior for non-partial pages? If so, what is that behavior?

pranavkm: The previous edition of the API was `IRazorViewEngine.FindPage(ActionContext context, string path);` - it didn't have a notion of `partial` at least not in the API surface. We shouldn't have it here

dougbu: @pranavkm the previous version didn't _expose_ the `IsPartial` notion. Under the covers however, it returned a page with `IsPartial` set to `true`. What I've done here is make the notion visible and make the `[Find|Get]Page` methods directly usable from `[Find|Get]View`. That's already been useful because it spurred the conversation @Eilon and I have been having about whether `true` is correct for layouts.

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngine.cs
```diff
@@ -228,31 +172,116 @@ public virtual IEnumerable<string> AreaViewLocationFormats
             return stringRouteValue;
         }
 
-        private RazorPageResult GetRazorPageResult(
-            ActionContext context,
-            string pageName,
-            bool isPartial)
+        /// <inheritdoc />
+        public RazorPageResult FindPage(ActionContext context, string pageName, bool isPartial)
```
pranavkm: Might want to rebase on dev since this type's mutated a bit.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewEngines/CompositeViewEngine.cs
```diff
@@ -11,6 +10,8 @@ namespace Microsoft.AspNet.Mvc.ViewEngines
     /// <inheritdoc />
     public class CompositeViewEngine : ICompositeViewEngine
     {
+        private const string ViewExtension = ".cshtml";
```
pranavkm: I guess this is no longer needed?

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/ViewResultExecutor.cs
```diff
@@ -68,7 +68,12 @@ public virtual ViewEngineResult FindView(ActionContext actionContext, ViewResult
             var viewEngine = viewResult.ViewEngine ?? ViewEngine;
             var viewName = viewResult.ViewName ?? actionContext.ActionDescriptor.Name;
 
-            var result = viewEngine.FindView(actionContext, viewName);
+            var result = viewEngine.GetView(executingFilePath: null, viewPath: viewName, isPartial: false);
```
pranavkm: Isn't this essentially moving the check one layer up (out of the `IViewEngine`)? For any `IViewEngine` lookup aren't you essentially forced to perform both these checks (or is there a scenario where you can only have to call `GetView` \ `FindView` but not the other?

dougbu: Yes, moving the check(s) one layer up was exactly the point of @Eilon's request. We can now easily add for example `LayoutPath` and `LayoutName` properties that expose this separation to users.

src/Microsoft.AspNet.Mvc.Razor/IRazorViewEngine.cs
dougbu: Based on the use of `<seealso/>` in the System.Net.Http.Formatting package (`HttpClientExtensions`'s `PostAsync` and `PutAsync` overloads and `MediaTypeFormatter`'s `ReadFromStreamAsync` and `WriteToStreamAsync` overloads), looks as if the element simply doesn't work outside other top-level elements. I checked MSDN, IntelliSense, and the object browser and saw nothing added.

src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
dougbu: Fixed this issue.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewEngines/CompositeViewEngine.cs
dougbu: Removed.

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngine.cs
dougbu: Done. Ick.

src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
dougbu: Conversation continuing offline. Hope to resolve this tomorrow.

src/Microsoft.AspNet.Mvc.Razor/IRazorViewEngine.cs
dougbu: Not sure `GetAbsolutePath` is significantly better but :ok:

dougbu:
:up::date:

src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
dougbu: Resolved. Decisions reflected in 94a32f6.

dougbu:
:up::date: (feels done)

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewEngines/CompositeViewEngine.cs
```diff
@@ -24,61 +25,88 @@ public CompositeViewEngine(IOptions<MvcViewOptions> optionsAccessor)
...
-        /// <inheritdoc />
-        public ViewEngineResult FindView(
-            ActionContext context,
-            string viewName)
-        {
-            if (context == null)
+            // Do not allocate in the common cases: ViewEngines contains one entry or initial attempt is successful.
+            IEnumerable<string> searchedLocations = null;
+            List<string> searchedList = null;
+            for (var index = 0; index < ViewEngines.Count; index++)
```
Eilon: Any reason to change these loops from a `foreach` to a `for`? And why use `index` instead of good old `i`?

dougbu: 1. Because this is two lines after a comment saying what follows is allocation free. But `foreach` allocates at least the enumerator.
2. I like `index` :smiley_cat: but will switch if you feel strongly. Do you?

Eilon: 1. Who cares about allocations here?
2. We always use `i`.

src/Microsoft.AspNet.Mvc.Razor/IRazorViewEngine.cs
```diff
@@ -11,12 +11,38 @@ namespace Microsoft.AspNet.Mvc.Razor
...
         /// </summary>
         /// <param name="context">The <see cref="ActionContext"/>.</param>
-        /// <param name="page">The name or full path to the view.</param>
-        /// <returns>A result representing the result of locating the <see cref="IRazorPage"/>.</returns>
-        RazorPageResult FindPage(ActionContext context, string page);
+        /// <param name="pageName">The name of the page.</param>
+        /// <param name="isMainPage">Determines if the page being found is the main page for an action.</param>
+        /// <returns>The <see cref="RazorPageResult"/> of locating the page.</returns>
+        /// <remarks><seealso cref="IViewEngine.FindView"/>.</remarks>
+        RazorPageResult FindPage(ActionContext context, string pageName, bool isMainPage);
```
pranavkm: Is there any instance of this API where `isMainPage` is true?

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewEngines/CompositeViewEngine.cs
```diff
@@ -24,61 +25,88 @@ public CompositeViewEngine(IOptions<MvcViewOptions> optionsAccessor)
...
-            return FindView(context, partialViewName, partial: true);
-        }
-
-        /// <inheritdoc />
-        public ViewEngineResult FindView(
-            ActionContext context,
-            string viewName)
-        {
-            if (context == null)
+            // Do not allocate in the common cases: ViewEngines contains one entry or initial attempt is successful.
```
pranavkm: Could we get before \ after allocations for this if it's a perf change?

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewEngines/ViewEngineResult.cs
```diff
@@ -64,14 +65,30 @@ public static ViewEngineResult Found(string viewName, IView view)
...
 
-        public ViewEngineResult EnsureSuccessful()
+        /// <summary>
+        /// Ensure this <see cref="ViewEngineResult"/> was successful.
+        /// </summary>
+        /// <param name="originalLocations">
+        /// Additional <see cref="SearchedLocations"/> to include in the thrown <see cref="InvalidOperationException"/>
+        /// if <see cref="Success"/> is <c>false</c>.
+        /// </param>
+        /// <exception cref="InvalidOperationException">
```
pranavkm: We typically don't document thrown exceptions. Why do it here?

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/PartialViewResultExecutor.cs
```diff
@@ -69,16 +71,51 @@ public virtual ViewEngineResult FindView(ActionContext actionContext, PartialVie
...
+                    if (result.SearchedLocations.Any())
+                    {
+                        // Return a new ViewEngineResult listing all searched locations.
+                        var locations = new List<string>(originalResult.SearchedLocations);
+                        locations.AddRange(result.SearchedLocations);
+                        result = ViewEngineResult.NotFound(viewName, locations);
+                    }
+                    else
+                    {
+                        // GetView() searched locations but FindView() did not. Use first ViewEngineResult.
```
pranavkm: Can we change this to be a bit more explicit - `ViewEngineResult.UnsupportedPath` or something along those lines? Seems cleaner than having to look at the `SearchedLocations` to determine if the view engine actually did any work.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/TemplateRenderer.cs
```diff
@@ -109,9 +109,14 @@ public IHtmlContent Render()
 
             foreach (string viewName in GetViewNames())
             {
-                var fullViewName = modeViewPath + "/" + viewName;
+                var viewEngineResult = _viewEngine.GetView(_viewContext.ExecutingFilePath, viewName, isMainPage: false);
+                if (!viewEngineResult.Success)
+                {
+                    // Success here is more common than with GetView() but GetView() is less expensive.
```
pranavkm: Isn't this a pattern requirement and not a choice based on what's more expensive? We can't say `GetView` is necessarily less expensive for all `IViewEngine` implementations.

test/Microsoft.AspNet.Mvc.Razor.Test/RazorViewEngineTest.cs
```diff
@@ -32,7 +32,7 @@ public class RazorViewEngineTest
             {"controller", "bar"},
         };
 
-        public static IEnumerable<string[]> InvalidViewNameValues
+        public static IEnumerable<string[]> AbsoluteViewPathData
```
pranavkm: :+1:

src/Microsoft.AspNet.Mvc.Razor/IRazorViewEngine.cs
dougbu: At the moment `GetPage()` and `FindPage()` are each used exactly once. I had previously implemented `GetView()` and `FindView()` on top of them. But this is a public API and the behaviour is now consistent with the corresponding `IViewEngine` methods and no semantic is hidden.

test/Microsoft.AspNet.Mvc.Razor.Test/RazorViewEngineTest.cs
```diff
@@ -73,50 +73,82 @@ public static IEnumerable<object[]> ViewLocationExpanderTestData
         [Theory]
         [InlineData(null)]
         [InlineData("")]
-        public void FindView_ThrowsIfViewNameIsNullOrEmpty(string viewName)
+        public void FindView_IsMainPage_ThrowsIfViewNameIsNullOrEmpty(string viewName)
```
pranavkm: We aren't varying the method name based on other parameters (`FindView_ForAreas_WithMainView_Throws` etc). That bit in the name shouldn't be necessary.

test/Microsoft.AspNet.Mvc.Razor.Test/RazorViewEngineTest.cs
```diff
@@ -967,6 +1208,166 @@ public void FindPage_LooksForPages_UsesRouteValuesAsFallback(bool isAttributeRou
...
+        }
+
+        [Theory]
+        [InlineData(null, "/Page")]
+        [InlineData(null, "~/Folder/Page.cshtml")]
+        [InlineData(null, "/Folder1/Folder2/Page.rzr")]
+        [InlineData("/Home/Index.cshtml", "~/Page")]
+        [InlineData("/Home/Index.cshtml", "/Folder/Page.cshtml")]
+        [InlineData("/Home/Index.cshtml", "~/Folder1/Folder2/Page.rzr")]
+        public void GetAbsolutePath_ReturnsPageUnchanged_IfAppRelative(string executingFilePath, string pagePath)
```
pranavkm: `GetAbsolutePath_ReturnsPagePathUnchanged_IfAppRelative`

test/WebSites/RazorWebSite/Components/ComponentWithRelativePath.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using Microsoft.AspNet.Mvc;
+
+namespace RazorWebSite.Components
+{
+    public class ComponentWithRelativePath : ViewComponent
+    {
+        public IViewComponentResult Invoke(Person person)
+        {
+            return View("../Shared/Components/ComponentWithRelativePath.cshtml", person);
```
pranavkm: Would we need tests to verify the behavior of relative paths outside app root? I'm guessing that's testing the implementation of `IFileProvider` at that point.

pranavkm:
:watch: for questions.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewEngines/CompositeViewEngine.cs
dougbu: 1. agreed offline this code path is hot.
2. I will switch to `i`.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/ViewViewComponentResult.cs
```diff
@@ -76,18 +77,21 @@ public void Execute(ViewComponentContext context)
                 throw new ArgumentNullException(nameof(context));
             }
 
-            var viewEngine = ViewEngine ?? ResolveViewEngine(context);
```
dougbu: I'll revert. I must have sorted the `view*` rows alphabetically because reasons but ignored `isNullOrEmptyViewName` because other reasons.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewEngines/CompositeViewEngine.cs
dougbu: Agreed offline that this is a hot path. E.g. think of the `DisplayFor()` / `Editor()` template lookups.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewEngines/ViewEngineResult.cs
dougbu: Because throwing or not throwing is the point of this method.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/PartialViewResultExecutor.cs
dougbu: "did any work" isn't the point and `UnsupportedPath` wouldn't add value. The important thing is just that one `ViewEngineResult` has useful information for the caller and the other doesn't. This code avoids unnecessarily creating a new `ViewEngineResult` and contained `List<string>` as in the `true` side of this particular condition.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/TemplateRenderer.cs
dougbu: The pattern is flexible and we could call `FindView()` first. However 98.7% of the time, the only view engine is the `RazorViewEngine` and the difference between `GetView()` and `FindView()` is rather larger. But really the point is that the `Display(... templateName: "relative/or/absolute/path.cshtml")` cases are unusual but not enough to reverse the pattern here.

Think I may just delete the comment.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/PartialViewResultExecutor.cs
pranavkm: > The important thing is just that one ViewEngineResult has useful information for the caller and the other doesn't.

But the current API doesn't express this. We're repurposing the `SearchedLocations` to determine this. A proper API that expressively does this would be better. Avoiding extra `ViewEngineResult` instances seems tangential.

Eilon:
:shipit:

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/PartialViewResultExecutor.cs
pranavkm: Discussed this offline.

src/Microsoft.AspNet.Mvc.Razor/IRazorViewEngine.cs
pranavkm: But there isn't a property `IsMainPage` on RazorPage (and probably shouldn't be one). We shouldn't need this parameter.

dougbu: Discussed offline. We don't expect any additional uses of `IRazorViewEngine`. I'll simplify the interface and remove the `isMainPage` parameter.

test/Microsoft.AspNet.Mvc.Razor.Test/RazorViewEngineTest.cs
dougbu: Reduced this somewhat but still have (just) a few with unique names, similar to old `_IsPartial` name qualifications.

test/WebSites/RazorWebSite/Components/ComponentWithRelativePath.cs
dougbu: Right: No need for such tests of the `IFileProvider`.

dougbu:
:up::date:

pranavkm:
:shipit:

dougbu:
Commits 08dd77d through 95b3d2c pushed individually :frowning:

PR #3381, opened 2015-10-21
[Fixes #3177] Clean up logic for encoding in ViewResult and friends

### LINKED ISSUE #3177: Clean up logic for encoding in ViewResult and friends
Example here: #3176

There is a basic priority system to how we determine the content-type/encoding for writing out a `ViewResult`

**content-type**
1. If there is a content-type specified on the result we prefer that - but we add `charset=utf8` if no encoding is specified
2. Else, if there is a content-type specified on the `HttpResponse` we use it verbatim (don't add `charset`)
3. Else, we specify `text/html; charset=utf-8` 

**encoding**
1. If we're here, we use the provided encoding or UTF8 if unset
2. If we're here, we **always** use UTF8 - we don't attempt to parse `response.ContentType`
3. If we're here, we always use UTF8 which matches our charset

The issue of course here is that case 1 and case 2 seem out of whack. If we're going to add `charset` to the content type user-code sets, we should do it consistently and add it in case 2. However, if we're going to detect the encoding from the content-type we should also do the same in case 2.

My suggestion, remove the fixup from case 1, and update case 2 to use the encoding specified in the content-type assuming it's valid.

kichalla:
@rynowak

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ContentResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ContentResult.cs
@@ -47,32 +47,31 @@ namespace Microsoft.AspNet.Mvc
             var logger = loggerFactory.CreateLogger<ContentResult>();
 
             var response = context.HttpContext.Response;
-            var contentTypeHeader = ContentType;
 
-            if (contentTypeHeader != null && contentTypeHeader.Encoding == null)
-            {
-                // Do not modify the user supplied content type, so copy it instead
-                contentTypeHeader = contentTypeHeader.Copy();
-                contentTypeHeader.Encoding = Encoding.UTF8;
-            }
+            string resolvedContentType;
+            Encoding resolvedContentTypeEncoding;
+            ResponseContentTypeHelper.ResolveContentTypeAndEncoding(
+                ContentType,
+                response.ContentType,
+                DefaultContentType,
+                out resolvedContentType,
+                out resolvedContentTypeEncoding);
 
-            response.ContentType = contentTypeHeader?.ToString()
-                ?? response.ContentType
-                ?? DefaultContentType.ToString();
+            response.ContentType = resolvedContentType;
 
             if (StatusCode != null)
             {
                 response.StatusCode = StatusCode.Value;
             }
 
-            logger.ContentResultExecuting(contentTypeHeader);
+            logger.ContentResultExecuting(resolvedContentType);
 
             if (Content != null)
             {
                 var bufferingFeature = response.HttpContext.Features.Get<IHttpBufferingFeature>();
                 bufferingFeature?.DisableResponseBuffering();
 
-                return response.WriteAsync(Content, contentTypeHeader?.Encoding ?? DefaultContentType.Encoding);
+                return response.WriteAsync(Content, resolvedContentTypeEncoding);
             }
 
             return TaskCache.CompletedTask;
--- a/src/Microsoft.AspNet.Mvc.Core/Internal/ResponseContentTypeHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Internal/ResponseContentTypeHelper.cs
@@ -1,0 +1,73 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Diagnostics;
+using System.Text;
+using Microsoft.Net.Http.Headers;
+using Microsoft.AspNet.Http;
+
+namespace Microsoft.AspNet.Mvc.Internal
+{
+    public static class ResponseContentTypeHelper
+    {
+        /// <summary>
+        /// Gets the content type and encoding that need to be used for the response.
+        /// The priority for selecting the content type is:
+        /// 1. ContentType property set on the action result
+        /// 2. <see cref="HttpResponse.ContentType"/> property set on <see cref="HttpResponse"/>
+        /// 3. Default content type set on the action result
+        /// </summary>
+        /// <remarks>
+        /// The user supplied content type is not modified and is used as is. For example, if user
+        /// sets the content type to be "text/plain" without any encoding, then the default content type's
+        /// encoding is used to write the response and the ContentType header is set to be "text/plain" without any
+        /// "charset" information.
+        /// </remarks>
+        /// <param name="actionResultContentType">ContentType set on the action result</param>
+        /// <param name="httpResponseContentType"><see cref="HttpResponse.ContentType"/> property set
+        /// on <see cref="HttpResponse"/></param>
+        /// <param name="defaultContentType">The default content type of the action result.</param>
+        /// <param name="resolvedContentType">The content type to be used for the response content type header</param>
+        /// <param name="resolvedContentTypeEncoding">Encoding to be used for writing the response</param>
+        public static void ResolveContentTypeAndEncoding(
+            MediaTypeHeaderValue actionResultContentType,
+            string httpResponseContentType,
+            MediaTypeHeaderValue defaultContentType,
+            out string resolvedContentType,
+            out Encoding resolvedContentTypeEncoding)
+        {
+            Debug.Assert(defaultContentType != null);
+            Debug.Assert(defaultContentType.Encoding != null);
+
+            // 1. User sets the ContentType property on the action result
+            if (actionResultContentType != null)
+            {
+                resolvedContentType = actionResultContentType.ToString();
+                resolvedContentTypeEncoding = actionResultContentType.Encoding ?? defaultContentType.Encoding;
+                return;
+            }
+
+            // 2. User sets the ContentType property on the http response directly
+            if (!string.IsNullOrEmpty(httpResponseContentType))
+            {
+                MediaTypeHeaderValue mediaType;
+                if (MediaTypeHeaderValue.TryParse(httpResponseContentType, out mediaType))
+                {
+                    resolvedContentType = httpResponseContentType;
+                    resolvedContentTypeEncoding = mediaType.Encoding ?? defaultContentType.Encoding;
+                }
+                else
+                {
+                    resolvedContentType = httpResponseContentType;
+                    resolvedContentTypeEncoding = defaultContentType.Encoding;
+                }
+
+                return;
+            }
+
+            // 3. Fall-back to the default content type
+            resolvedContentType = defaultContentType.ToString();
+            resolvedContentTypeEncoding = defaultContentType.Encoding;
+        }
+    }
+}
--- a/src/Microsoft.AspNet.Mvc.Core/Logging/ContentResultLoggerExtensions.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Logging/ContentResultLoggerExtensions.cs
@@ -19,9 +19,9 @@ namespace Microsoft.AspNet.Mvc.Logging
                 "Executing ContentResult with HTTP Response ContentType of {ContentType}");
         }
 
-        public static void ContentResultExecuting(this ILogger logger, MediaTypeHeaderValue contentType)
+        public static void ContentResultExecuting(this ILogger logger, string contentType)
         {
-            _contentResultExecuting(logger, contentType?.MediaType, null);
+            _contentResultExecuting(logger, contentType, null);
         }
     }
 }
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/ViewExecutor.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/ViewExecutor.cs
@@ -7,6 +7,7 @@ using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Mvc.Diagnostics;
 using Microsoft.AspNet.Mvc.Infrastructure;
+using Microsoft.AspNet.Mvc.Internal;
 using Microsoft.AspNet.Mvc.ModelBinding;
 using Microsoft.AspNet.Mvc.Rendering;
 using Microsoft.AspNet.Mvc.ViewEngines;
@@ -96,7 +97,8 @@ namespace Microsoft.AspNet.Mvc.ViewFeatures
         /// <param name="viewData">The <see cref="ViewDataDictionary"/>.</param>
         /// <param name="tempData">The <see cref="ITempDataDictionary"/>.</param>
         /// <param name="contentType">
-        /// The content-type header value to set in the response. If <c>null</c>, <see cref="DefaultContentType"/> will be used.
+        /// The content-type header value to set in the response. If <c>null</c>,
+        /// <see cref="DefaultContentType"/> will be used.
         /// </param>
         /// <param name="statusCode">
         /// The HTTP status code to set in the response. May be <c>null</c>.
@@ -134,26 +136,23 @@ namespace Microsoft.AspNet.Mvc.ViewFeatures
 
             var response = actionContext.HttpContext.Response;
 
-            if (contentType != null && contentType.Encoding == null)
-            {
-                // Do not modify the user supplied content type, so copy it instead
-                contentType = contentType.Copy();
-                contentType.Encoding = Encoding.UTF8;
-            }
+            string resolvedContentType = null;
+            Encoding resolvedContentTypeEncoding = null;
+            ResponseContentTypeHelper.ResolveContentTypeAndEncoding(
+                contentType,
+                response.ContentType,
+                DefaultContentType,
+                out resolvedContentType,
+                out resolvedContentTypeEncoding);
 
-            // Priority list for setting content-type:
-            //      1. passed in contentType (likely set by the user on the result)
-            //      2. response.ContentType (likely set by the user in controller code)
-            //      3. ViewExecutor.DefaultContentType (sensible default)
-            response.ContentType = contentType?.ToString() ?? response.ContentType ?? DefaultContentType.ToString();
+            response.ContentType = resolvedContentType;
 
             if (statusCode != null)
             {
                 response.StatusCode = statusCode.Value;
             }
 
-            var encoding = contentType?.Encoding ?? DefaultContentType.Encoding;
-            using (var writer = WriterFactory.CreateWriter(response.Body, encoding))
+            using (var writer = WriterFactory.CreateWriter(response.Body, resolvedContentTypeEncoding))
             {
                 var viewContext = new ViewContext(
                     actionContext,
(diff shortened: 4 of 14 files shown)


src/Microsoft.AspNet.Mvc.Core/Internal/ResponseContentTypeHelper.cs
```diff
@@ -0,0 +1,54 @@
...
+        {
+            if (defaultContentType == null)
+            {
+                throw new ArgumentNullException(nameof(defaultContentType));
+            }
+
+            if (defaultContentType.Encoding == null)
+            {
+                throw new InvalidOperationException(
+                    Resources.FormatDefaultContentTypeMustHaveEncoding(defaultContentType.ToString()));
```
dougbu: This should be an `ArgumentException` instead. Look at where `Resources.ModelBinderUtil_ModelMetadataCannotBeNull` is used in `ModelBindingHelper.ValidateBindingContext()`.

src/Microsoft.AspNet.Mvc.Core/Internal/ResponseContentTypeHelper.cs
```diff
@@ -0,0 +1,54 @@
...
+            }
+
+            // 1. User sets the ContentType property on the action result
+            if (actionResultContentType != null)
+            {
+                if (actionResultContentType.Encoding == null)
+                {
+                    // Do not modify the user supplied content type, so copy it instead
+                    var contentType = actionResultContentType.Copy();
+                    contentType.Encoding = defaultContentType.Encoding;
```
dougbu: Why does this case ignore an encoding specified in the `httpResponseContentType` (if any)?

src/Microsoft.AspNet.Mvc.Core/Resources.resx
```diff
@@ -319,4 +319,7 @@
     <value>Path '{0}' was not rooted.</value>
     <comment>{0} is the path which wasn't rooted</comment>
   </data>
+  <data name="DefaultContentTypeMustHaveEncoding" xml:space="preserve">
+    <value>The default content type '{0}' must have an encoding.</value>
```
dougbu: `Argument must have a non-null {1} property.` The argument name is already visible in an `ArgumentException`'s `Message`.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/ViewExecutor.cs
```diff
@@ -104,7 +106,7 @@ public class ViewExecutor
             IView view,
             ViewDataDictionary viewData,
             ITempDataDictionary tempData,
-            MediaTypeHeaderValue contentType,
+            MediaTypeHeaderValue actionResultContentType,
```
dougbu: Odd rename. Why?

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.Core/ContentResult.cs
```diff
@@ -64,7 +59,7 @@ public override Task ExecuteResultAsync(ActionContext context)
                 var bufferingFeature = response.HttpContext.Features.Get<IHttpBufferingFeature>();
                 bufferingFeature?.DisableResponseBuffering();
 
-                return response.WriteAsync(Content, contentTypeHeader?.Encoding ?? DefaultContentType.Encoding);
+                return response.WriteAsync(Content, contentTypeHeader.Encoding);
```
rynowak: This doesn't do what we want. If the user set an explicit content type that doesn't include `charset`, we're aren't trying to 'fix' it for them.

src/Microsoft.AspNet.Mvc.Core/Internal/ResponseContentTypeHelper.cs
rynowak: :+1:

rynowak: We prioritize determining the encoding from a content type on the action result over anything the user did the response.

src/Microsoft.AspNet.Mvc.Core/Internal/ResponseContentTypeHelper.cs
```diff
@@ -0,0 +1,54 @@
...
+                    return contentType;
+                }
+
+                return actionResultContentType;
+            }
+
+            // 2. User sets the ContentType property on the http response directly
+            if (!string.IsNullOrEmpty(httpResponseContentType))
+            {
+                var contentType = MediaTypeHeaderValue.Parse(httpResponseContentType);
```
rynowak: What if it's invalid? We should let you do that.

src/Microsoft.AspNet.Mvc.Core/Internal/ResponseContentTypeHelper.cs
```diff
@@ -0,0 +1,54 @@
...
+                contentType.Encoding = contentType.Encoding ?? defaultContentType.Encoding;
```
rynowak: Is there a case where our default content type doesn't have an encoding? For all practical purposes the 'default' is just UTF8

src/Microsoft.AspNet.Mvc.Core/Resources.resx
rynowak: All of this seems like going to far to attempt to consolidate this code. How many copies were you looking to dedupe?

src/Microsoft.AspNet.Mvc.Core/Internal/ResponseContentTypeHelper.cs
rynowak: Also, we **do not** intend to 'fix' the content type if you set it to something invalid, or that doesn't contain a `charset`.

My concern is that once you've corrected the behavior problems here, the API it won't be too cleanly.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/ViewExecutor.cs
kichalla: Just for more clarity while reading the code...consider the following piece of code which we try to figure the content type

```
            var resolvedContentType = ResponseContentTypeHelper.GetContentType(
                actionResultContentType,
                response.ContentType,
                DefaultContentType);
```

src/Microsoft.AspNet.Mvc.Core/ContentResult.cs
kichalla: Sure

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/ViewExecutor.cs
dougbu: Not sure I buy this name is any better. The new name provides unnecessary information about the context in which `ExecuteAsync()` is called but says nothing more about how it's used. All this method should infer about the context is that the `Response` and its headers can be changed.

src/Microsoft.AspNet.Mvc.Core/Internal/ResponseContentTypeHelper.cs
```diff
@@ -0,0 +1,57 @@
...
+using System;
+using System.Text;
+using Microsoft.AspNet.Mvc.Core;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNet.Mvc.Internal
+{
+    public static class ResponseContentTypeHelper
+    {
+        public static Tuple<string, Encoding> GetResponseContentTypeAndEncoding(
```
dougbu: `Tuple` is an ugly type that makes the calling code less readable. Use separate methods, two `out` parameters with a `void` return value, ...

src/Microsoft.AspNet.Mvc.Core/Resources.resx
dougbu: I suggested `{1}` because we normally pass in names from the code, not to dedupe messages.

kichalla:
:up: :date: @ryan @dougbu

src/Microsoft.AspNet.Mvc.Core/ContentResult.cs
```diff
@@ -47,32 +47,31 @@ public override Task ExecuteResultAsync(ActionContext context)
...
 
-            if (contentTypeHeader != null && contentTypeHeader.Encoding == null)
-            {
-                // Do not modify the user supplied content type, so copy it instead
-                contentTypeHeader = contentTypeHeader.Copy();
-                contentTypeHeader.Encoding = Encoding.UTF8;
-            }
+            string resolvedContentType = null;
+            Encoding resolvedContentTypeEncoding = null;
+            ResponseContentTypeHelper.ResolveContentTypeAndEncoding(
```
rynowak: You don't need to initialize these variables. `out` variables are assigned when you call the function.

src/Microsoft.AspNet.Mvc.Core/Internal/ResponseContentTypeHelper.cs
```diff
@@ -0,0 +1,62 @@
...
+using System;
+using System.Text;
+using Microsoft.AspNet.Mvc.Core;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNet.Mvc.Internal
+{
+    public static class ResponseContentTypeHelper
+    {
+        public static void ResolveContentTypeAndEncoding(
```
rynowak: I'm not asking for a doc comment here, but a little explanation of what the purpose of this is, and what scenarios it handles.

src/Microsoft.AspNet.Mvc.Core/Internal/ResponseContentTypeHelper.cs
```diff
@@ -0,0 +1,62 @@
...
+            if (defaultContentType == null)
+            {
+                throw new ArgumentNullException(nameof(defaultContentType));
+            }
+
+            if (defaultContentType.Encoding == null)
+            {
+                throw new ArgumentException(
+                    Resources.FormatDefaultContentTypeMustHaveEncoding(defaultContentType.ToString()));
+            }
```
rynowak: I don't understand why this adds any value. This is an internal namespace

src/Microsoft.AspNet.Mvc.Core/Internal/ResponseContentTypeHelper.cs
```diff
@@ -0,0 +1,62 @@
...
+
+            // 1. User sets the ContentType property on the action result
```
rynowak: These comments need to be beefed up. The original comments benefitted from the context of the surrounding code.

test/Microsoft.AspNet.Mvc.Core.Test/Internal/ResponseContentTypeHelperTest.cs
```diff
@@ -0,0 +1,218 @@
...
+                        null,
+                        "text/bar",
+                        "text/bar"
+                    },
+                    {
+                        null,
+                        "text/bar; p1=p1-value",
+                        "text/bar; p1=p1-value"
+                    },
+                                        {
```
rynowak: wonky-donky

test/Microsoft.AspNet.Mvc.Core.Test/Internal/ResponseContentTypeHelperTest.cs
```diff
@@ -0,0 +1,218 @@
...
+                    {
+                        MediaTypeHeaderValue.Parse("text/foo; charset=us-ascii"),
+                        "text/bar",
+                        "text/foo; charset=us-ascii"
+                    },
+                    {
+                        MediaTypeHeaderValue.Parse("text/foo; charset=us-ascii"),
+                        "text/bar; charset=utf-8",
+                        "text/foo; charset=us-ascii"
+                    }
```
rynowak: It would be nice to document user-scenarios each of these test inputs represent. I don't feel like using `[Theory]` here makes these tests better, it's hard to tell what's tested and what's not.

test/Microsoft.AspNet.Mvc.Core.Test/Internal/ResponseContentTypeHelperTest.cs
```diff
@@ -0,0 +1,218 @@
...
+                    httpResponseContentType,
+                    defaultContentType,
+                    out resolvedContentType,
+                    out resolvedContentTypeEncoding);
+            });
+            Assert.Equal(
+                $"The default content type '{defaultContentType.ToString()}' must have an encoding.",
+                exception.Message);
+        }
+    }
```
rynowak: Why is there 100 lines of test code for these exception messages? This should just be a `Debug.Assert`.

rynowak:
:watch: - this is looking much better

src/Microsoft.AspNet.Mvc.Core/Internal/ResponseContentTypeHelper.cs
```diff
@@ -0,0 +1,62 @@
...
+namespace Microsoft.AspNet.Mvc.Internal
+{
+    public static class ResponseContentTypeHelper
+    {
+        public static void ResolveContentTypeAndEncoding(
+            MediaTypeHeaderValue actionResultContentType,
+            string httpResponseContentType,
+            MediaTypeHeaderValue defaultContentType,
+            out string resolvedResponseContentType,
+            out Encoding resolvedContentTypeEncoding)
```
dougbu: Suggest removing assumptions and redundancies from these names. `actionResultContentType` says something about the caller, not this code. Here it's more like `contentTypeOverride`. And the word "Response" doesn't add much in the `ResponseContentTypeHelper` class.

src/Microsoft.AspNet.Mvc.Core/Logging/ContentResultLoggerExtensions.cs
```diff
@@ -19,9 +19,9 @@ static ContentResultLoggerExtensions()
                 "Executing ContentResult with HTTP Response ContentType of {ContentType}");
         }
 
-        public static void ContentResultExecuting(this ILogger logger, MediaTypeHeaderValue contentType)
+        public static void ContentResultExecuting(this ILogger logger, string contentType)
         {
-            _contentResultExecuting(logger, contentType?.MediaType, null);
+            _contentResultExecuting(logger, contentType, null);
```
dougbu: Name the `null` parameter.

test/Microsoft.AspNet.Mvc.Core.Test/ContentResultTest.cs
```diff
@@ -131,14 +91,14 @@ public class ContentResultTest
                         new MediaTypeHeaderValue("text/foo"),
                         "κόσμε",
                         null,
-                        "text/foo; charset=utf-8",
+                        "text/foo",
```
dougbu: Why are `responseContentType` values losing their `charset` values?

test/Microsoft.AspNet.Mvc.Core.Test/ContentResultTest.cs
```diff
@@ -158,15 +118,15 @@ public class ContentResultTest
...
+                        "text/bar; p1=p1-value",
                         new byte[] { 97, 98, 99, 100 }
                     },
                     {
                         null,
                         "abcd",
-                        "Invalid content type",
-                        "Invalid content type",
+                        "application/xml; p1=p1-value; charset=us-ascii",
+                        "application/xml; p1=p1-value; charset=us-ascii",
```
dougbu: Why can't "Invalid content type" be tested anymore?

test/Microsoft.AspNet.Mvc.Core.Test/ContentResultTest.cs
```diff
@@ -158,15 +118,15 @@ public class ContentResultTest
                     {
                         null,
                         "abcd",
-                        "application/xml; charset=us-ascii",
-                        "application/xml; charset=us-ascii",
+                        "text/bar; p1=p1-value",
+                        "text/bar; p1=p1-value",
```
dougbu: This looks like a new test. Why was the old one removed?

test/Microsoft.AspNet.Mvc.Core.Test/ContentResultTest.cs
kichalla: Because we do not want to modify what the user has set explicitly, which is just `text/foo` without any charset

dougbu: Generally scary how much existing tests are changing in a change mostly about refactoring existing code.

test/Microsoft.AspNet.Mvc.Core.Test/ContentResultTest.cs
kichalla: good catch...this should be a valid test now as I reverted the behavior from the previous commit

test/Microsoft.AspNet.Mvc.Core.Test/ContentResultTest.cs
kichalla: But this is not just refactoring though right...we have a behavior change as mentioned earlier...

dougbu: Reasoning is a bit odd. User asked for _a_ `text/foo` content type and the framework chose to encode using UTF8 but didn't indicate that in the response headers.

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.Core/Internal/ResponseContentTypeHelper.cs
rynowak: this is worse IMO

src/Microsoft.AspNet.Mvc.Core/Logging/ContentResultLoggerExtensions.cs
rynowak: Nope. The name is `arg2`

src/Microsoft.AspNet.Mvc.Core/Resources.resx
rynowak: Your feedback about the names is :100: correct. However we shouldn't even be throwing exceptions here. This is internal code.

test/Microsoft.AspNet.Mvc.Core.Test/ContentResultTest.cs
dougbu: :ok: After offline conversations, I'm fine w/ the product code change leading to this test change. :ok:

dougbu:
:shipit: from me but please get @rynowak's squirrel too.

rynowak:
:shipit:

Issue #3227, opened 2015-09-29
ModelState error for FormatException should show DisplayName of the problematic property if available


danroth27:
`ModelStateDictionary` adds `Resources.ModelError_InvalidValue_GenericMessage` and `Resources.ModelError_InvalidValue_MessageWithModelValue` messages when a `FormatException` occurs. This message includes the full key (property / collection path) e.g. "model.Customer.Name". For consistency with MVC 5 and usability, should instead use the `DisplayName` of the problematic property and fall back to the property name.

dougbu:
40b7636

PR #3385, opened 2015-10-22
Use `ModelMetadata.GetDisplayName()` in error message replacing `FormatException` and `OverflowException`

dougbu:
- #3227
- much of change is to tests, creating and passing `ModelMetadata`
- updated `InputFormatterContext` to make `ModelMetadata` available to `JsonInputFormatter`
  - walk `ModelMetadata` tree to get information about property with an issue
- add missing `null` checks in `ModelStateDictionaryExtensions`

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Formatters.Json/JsonInputFormatter.cs
+++ b/src/Microsoft.AspNet.Mvc.Formatters.Json/JsonInputFormatter.cs
@@ -6,7 +6,7 @@ using System.IO;
 using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Mvc.Internal;
-using Microsoft.Net.Http.Headers;
+using Microsoft.AspNet.Mvc.ModelBinding;
 using Newtonsoft.Json;
 
 namespace Microsoft.AspNet.Mvc.Formatters
@@ -101,7 +101,8 @@ namespace Microsoft.AspNet.Mvc.Formatters
                         }
                     }
 
-                    context.ModelState.TryAddModelError(key, eventArgs.ErrorContext.Error);
+                    var metadata = GetPathMetadata(context.Metadata, eventArgs.ErrorContext.Path);
+                    context.ModelState.TryAddModelError(key, eventArgs.ErrorContext.Error, metadata);
 
                     // Error must always be marked as handled
                     // Failure to do so can cause the exception to be rethrown at every recursive level and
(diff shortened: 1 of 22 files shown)


rynowak:
Design overall LGTM - I'll let @ajaybhargavb sign off on the specifics :laughing:

src/Microsoft.AspNet.Mvc.Formatters.Json/JsonInputFormatter.cs
```diff
@@ -171,5 +172,51 @@ protected virtual JsonSerializer CreateJsonSerializer()
         {
             return JsonSerializer.Create(SerializerSettings);
         }
+
+        private ModelMetadata GetPathMetadata(ModelMetadata metadata, string path)
+        {
+            var index = 0;
+            while (index >= 0 && index < path.Length)
```
ajaybhargavb: Is there any case that `index` might go below 0?

dougbu: Not in a path we understand. However we can't guarantee Json.NET will never provide a path such as `"property["`.

test/Microsoft.AspNet.Mvc.Formatters.Json.Test/JsonInputFormatterTest.cs
```diff
@@ -150,6 +158,61 @@ public static IEnumerable<object[]> JsonFormatterReadSimpleTypesData
...
+            Assert.True(result.HasError);
+            Assert.Equal("The supplied value is invalid for Byte.", modelState["[2]"].Errors[0].ErrorMessage);
+            Assert.Null(modelState["[2]"].Errors[0].Exception);
+        }
+
+        [Fact]
+        public async Task ReadAsync_InvalidComplexArray_AddsOverflowErrorsToModelState()
+        {
+            // Arrange
+            var content = "[{name: 'Name One', Age: 30}, {name: 'Name Two', Small: 300}]";
```
ajaybhargavb: Suggest adding a functional test for this scenario.

ajaybhargavb:
:watch: for test

test/Microsoft.AspNet.Mvc.Formatters.Json.Test/JsonInputFormatterTest.cs
dougbu: Why add that redundancy? Functional tests are about mainline scenarios and not uncommon binding errors.

ajaybhargavb: Spoke offline. There are already some integration tests to cover this scenario.

ajaybhargavb:
:shipit:

dougbu:
40b7636

PR #3391, opened 2015-10-22
Improve logging behavior for actions scope

rynowak:


src/Microsoft.AspNet.Mvc.Core/Controllers/ControllerActionDescriptor.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace Microsoft.AspNet.Mvc.Controllers
 {
-    [DebuggerDisplay("CA {DisplayName}(RC-{RouteConstraints.Count})")]
+    [DebuggerDisplay("{DisplayName}")]
```
rynowak: There was really no point in the extra stuff we were putting here, this is much cleaner.

src/Microsoft.AspNet.Mvc.Core/Logging/MvcRouteHandlerLoggerExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+                }
+
+                _action = action;
+            }
+
+            public IEnumerable<KeyValuePair<string, object>> GetValues()
+            {
+                return new KeyValuePair<string, object>[]
+                {
+                new KeyValuePair<string, object>("ActionId", _action.Id),
```
kichalla: nit: indent

kichalla:
:shipit:

Issue #3347, opened 2015-10-19
TagHelper and non-TagHelper select tags in a form don't work well together.


NTaylorMullen:
If you have:
```HTML
<form asp-action="...">
    <select asp-for="...">
        ...
    </select>
    <select>
        <option value="...">...</option>
    </select>
</form>
```

The `<option>` tags in the second select will think that they're running under the first `<select asp-for="...">`. This can be fixed by transitioning the `SelectTagHelper` to properly use `TagHelperContext.Items` instead of relying on `ViewContext.FormContext.FormData`.

Eilon:
Wow, call that OOPS!

@NTaylorMullen any idea if other tag helpers suffer from cross-talk like this?

dougbu:
@Eilon this is a to-do item we've had since we added the `Items` property bag -- which came after the `<select>` and `<option>` tag helpers. The problem is a corner case related to lifetime of the information relative to `FormContext`. And the new `<form>` tag helper for rendering at the end is the only other one that communicates to children but that one is definitely using the correct mechanism (`FormContext`). So, no chance of similar issues elsewhere.

Eilon:
@dougbu thanks, good to know.

NTaylorMullen:
911dfc57b0a40b8feb415e9afcc90a9845d0ba5b

PR #3392, opened 2015-10-22
Transition `SelectTagHelper` and `OptionTagHelper` to use `context.Items`.

NTaylorMullen:
- Added functional tests to validate data created from a `SelectTagHelper` does not impact following `<select>` tags.

#3347

### DIFF
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/OptionTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/OptionTagHelper.cs
@@ -54,8 +54,8 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
 
         /// <inheritdoc />
         /// <remarks>
-        /// Does nothing unless <see cref="FormContext.FormData"/> contains a
-        /// <see cref="SelectTagHelper.SelectedValuesFormDataKey"/> entry and that entry is a non-empty
+        /// Does nothing unless <see cref="TagHelperContext.Items"/> contains a
+        /// <see cref="SelectTagHelper"/> <see cref="Type"/> entry and that entry is a non-empty
         /// <see cref="ICollection{string}"/> instance. Also does nothing if the associated &lt;option&gt; is already
         /// selected.
         /// </remarks>
@@ -82,9 +82,7 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
             {
                 // Is this <option/> element a child of a <select/> element the SelectTagHelper targeted?
                 object formDataEntry;
-                ViewContext.FormContext.FormData.TryGetValue(
-                    SelectTagHelper.SelectedValuesFormDataKey,
-                    out formDataEntry);
+                context.Items.TryGetValue(typeof(SelectTagHelper), out formDataEntry);
 
                 // ... And did the SelectTagHelper determine any selected values?
                 var selectedValues = formDataEntry as ICollection<string>;
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/SelectTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/SelectTagHelper.cs
@@ -21,16 +21,8 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
     {
         private const string ForAttributeName = "asp-for";
         private const string ItemsAttributeName = "asp-items";
-
-        /// <summary>
-        /// Key used for selected values in <see cref="FormContext.FormData"/>.
-        /// </summary>
-        /// <remarks>
-        /// Value for this dictionary entry will either be <c>null</c> (indicating no <see cref="SelectTagHelper"/> has
-        /// executed within this &lt;form/&gt;) or an <see cref="ICollection{string}"/> instance. Elements of the
-        /// collection are based on current <see cref="ViewDataDictionary.Model"/>.
-        /// </remarks>
-        public static readonly string SelectedValuesFormDataKey = nameof(SelectTagHelper) + "-SelectedValues";
+        private bool _allowMultiple;
+        private IReadOnlyCollection<string> _currentValues;
 
         /// <summary>
         /// Creates a new <see cref="SelectTagHelper"/>.
@@ -70,6 +62,42 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
         public IEnumerable<SelectListItem> Items { get; set; }
 
         /// <inheritdoc />
+        public override void Init(TagHelperContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            // Note null or empty For.Name is allowed because TemplateInfo.HtmlFieldPrefix may be sufficient.
+            // IHtmlGenerator will enforce name requirements.
+            if (For.Metadata == null)
+            {
+                throw new InvalidOperationException(Resources.FormatTagHelpers_NoProvidedMetadata(
+                    "<select>",
+                    ForAttributeName,
+                    nameof(IModelMetadataProvider),
+                    For.Name));
+            }
+
+            // Base allowMultiple on the instance or declared type of the expression to avoid a
+            // "SelectExpressionNotEnumerable" InvalidOperationException during generation.
+            // Metadata.IsEnumerableType is similar but does not take runtime type into account.
+            var realModelType = For.ModelExplorer.ModelType;
+            _allowMultiple = typeof(string) != realModelType &&
+                typeof(IEnumerable).GetTypeInfo().IsAssignableFrom(realModelType.GetTypeInfo());
+            _currentValues = Generator.GetCurrentValues(
+                ViewContext,
+                For.ModelExplorer,
+                expression: For.Name,
+                allowMultiple: _allowMultiple);
+
+            // Whether or not (not being highly unlikely) we generate anything, could update contained <option/>
+            // elements. Provide selected values for <option/> tag helpers.
+            context.Items[typeof(SelectTagHelper)] = _currentValues;
+        }
+
+        /// <inheritdoc />
         /// <remarks>Does nothing if <see cref="For"/> is <c>null</c>.</remarks>
         /// <exception cref="InvalidOperationException">
         /// Thrown if <see cref="Items"/> is non-<c>null</c> but <see cref="For"/> is <c>null</c>.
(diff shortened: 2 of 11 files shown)


src/Microsoft.AspNet.Mvc.TagHelpers/OptionTagHelper.cs
```diff
@@ -54,7 +54,7 @@ public override int Order
 
         /// <inheritdoc />
         /// <remarks>
-        /// Does nothing unless <see cref="FormContext.FormData"/> contains a
+        /// Does nothing unless <see cref="TagHelperContext.Items"/> contains a
         /// <see cref="SelectTagHelper.SelectedValuesFormDataKey"/> entry and that entry is a non-empty
```
dougbu: Throw `SelectTagHelper.SelectedValuesFormDataKey` away -- far away. This is an inefficient pattern that `Items` was designed to avoid. Instead use `typeof(SelectTagHelper)`.

dougbu:
:watch:

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Mvc.TagHelpers/SelectTagHelper.cs
```diff
@@ -131,7 +121,7 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
 
             // Whether or not (not being highly unlikely) we generate anything, could update contained <option/>
             // elements. Provide selected values for <option/> tag helpers. They'll run next.
-            ViewContext.FormContext.FormData[SelectedValuesFormDataKey] = currentValues;
+            context.Items[typeof(SelectTagHelper)] = currentValues;
```
dougbu: This will need to move to new `Init()` method as part of your reaction PR. Could you rebase that branch on top of this one? Or is that almost ready to get in first?

test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.Order.html
```diff
@@ -41,6 +41,9 @@
                 <option value="Credit">Credit</option>
                 <option value="Check" selected="selected">Check</option>
             </select>
+            <select>
+                <option value="Check">Check</option>
+            </select>
```
dougbu: Do we have no way outside the functional tests to confirm the second `<select>` gets a different `Items` collection than the first? Hard to argue two `<select>` lists with overlapping `<option>` values and inconsistent tag helper use is mainline enough to show here.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/SelectTagHelperTest.cs
```diff
@@ -243,10 +243,9 @@ public class SelectTagHelperTest
             Assert.Equal(expectedPostContent, output.PostContent.GetContent());
             Assert.Equal(expectedTagName, output.TagName);
 
-            Assert.NotNull(viewContext.FormContext?.FormData);
             Assert.Single(
-                viewContext.FormContext.FormData,
-                entry => entry.Key == SelectTagHelper.SelectedValuesFormDataKey);
+                tagHelperContext.Items,
+                entry => (Type)entry.Key == typeof(SelectTagHelper));
```
dougbu: Why is the cast necessary? Looks messy.

test/WebSites/HtmlGenerationWebSite/Views/HtmlGeneration_Home/Order.cshtml
```diff
@@ -52,6 +52,9 @@
                 <option value="Credit">Credit</option>
                 <option value="Check">Check</option>
             </select>
+            <select>
+                <option value="Check">Check</option>
```
dougbu: This doesn't prove the `<option>` tag helpers are working correctly. (If it must be tested in a functional test ...) Need to have the "Check" `<option>` be selected just above and not selected here.

Suggest you change the `.cshtml` file so this one has `selected="selected"`. That attribute is cleared in the above `<select>` and won't be here.

test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.Order.html
dougbu: Note our general direction is toward unit testing and integration testing.

test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.Order.html
```diff
@@ -41,6 +41,9 @@
                 <option value="Credit">Credit</option>
```
dougbu: CI builds are failing because you didn't update `HtmlGenerationWebSite.HtmlGeneration_Home.Order.Encoded.html`.

test/WebSites/HtmlGenerationWebSite/Views/HtmlGeneration_Home/Order.cshtml
```diff
@@ -52,6 +52,9 @@
                 <option value="Credit">Credit</option>
```
dougbu: This particular set of files is "special" in that `Order.cshtml` and `OrderUsingHtmlHelpers.cshtml` should generate almost the same HTML. So, if you must change functional tests, change that and its output files too.

dougbu:
:watch:

test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.Order.html
NTaylorMullen: No, there's no good way to verify that the context doesn't flow from one to the other. Reason being a lot of what we're verifying is due to the overall `TagHelper` execution flow when going across two separate `TagHelper`s.

test/WebSites/HtmlGenerationWebSite/Views/HtmlGeneration_Home/Order.cshtml
NTaylorMullen: Hu? It verifies it, in the generated output the first is `selected` the second is not.

test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.Order.html
NTaylorMullen: :+1: Can thank this due to the flakey builds over the past two days haha

test/WebSites/HtmlGenerationWebSite/Views/HtmlGeneration_Home/Order.cshtml
NTaylorMullen: Will do

test/WebSites/HtmlGenerationWebSite/Views/HtmlGeneration_Home/Order.cshtml
dougbu: :ok: I mixed myself up.

src/Microsoft.AspNet.Mvc.TagHelpers/SelectTagHelper.cs
dougbu: Now `Init()`s in, this should be there.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/SelectTagHelperTest.cs
NTaylorMullen: image

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Mvc.TagHelpers/SelectTagHelper.cs
```diff
@@ -69,27 +61,16 @@ public override int Order
         [HtmlAttributeName(ItemsAttributeName)]
         public IEnumerable<SelectListItem> Items { get; set; }
 
-        /// <inheritdoc />
-        /// <remarks>Does nothing if <see cref="For"/> is <c>null</c>.</remarks>
-        /// <exception cref="InvalidOperationException">
-        /// Thrown if <see cref="Items"/> is non-<c>null</c> but <see cref="For"/> is <c>null</c>.
-        /// </exception>
-        public override void Process(TagHelperContext context, TagHelperOutput output)
+        public override void Init(TagHelperContext context)
```
dougbu: + `/// <inheritdoc />`

src/Microsoft.AspNet.Mvc.TagHelpers/SelectTagHelper.cs
```diff
@@ -102,25 +83,47 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
...
+            _allowMultiple = typeof(string) != realModelType &&
                 typeof(IEnumerable).GetTypeInfo().IsAssignableFrom(realModelType.GetTypeInfo());
+            _currentValues = Generator.GetCurrentValues(
+                ViewContext,
+                For.ModelExplorer,
+                expression: For.Name,
+                allowMultiple: _allowMultiple);
+
+            // Whether or not (not being highly unlikely) we generate anything, could update contained <option/>
+            // elements. Provide selected values for <option/> tag helpers. They'll run next.
```
dougbu: Remove `They'll run next.` since it's no longer relevant exactly when any child tag helpers execute.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/SelectTagHelperTest.cs
```diff
@@ -235,6 +235,7 @@ public class SelectTagHelperTest
             };
 
             // Act
+            tagHelper.Init(tagHelperContext);
```
dougbu: Like that you added this to every test in this class.

dougbu:
:shipit: but may want to wait for Coherence to catch up since we can't confirm MVC builds at the moment.

Issue #3365, opened 2015-10-21
Razor compilation uses a custom load context on CoreCLR and regular Assembly.Load on .NET Framework 


NTaylorMullen:
@davidfowl @Eilon

NTaylorMullen:
faa78a3d91f64770444a05841d20774db4b4f64e

PR #3393, opened 2015-10-22
Transition from `IAssemblyLoadContext` to `AssemblyLoadContext`.

NTaylorMullen:
- `IAssemblyLoadContext` is a DNX defined type that will eventually go away.

#3365

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationService.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationService.cs
@@ -9,6 +9,9 @@ using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Reflection.PortableExecutable;
+#if DOTNET5_4
+using System.Runtime.Loader;
+#endif
 using Microsoft.AspNet.FileProviders;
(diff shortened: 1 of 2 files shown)


pranavkm:
:shipit:

davidfowl:
:shipit:

src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationService.cs
```diff
@@ -146,6 +154,15 @@ public CompilationResult Compile(RelativeFileInfo fileInfo, string compilationCo
             }
         }
 
+        private Assembly LoadStream(MemoryStream ms, MemoryStream assemblySymbols)
+        {
+#if NET451
+            return Assembly.Load(ms.ToArray(), assemblySymbols?.ToArray());
+#else
+            return _razorLoadContext.Load(ms, assemblySymbols);
```
davidfowl: Make sure you can pass null here

NTaylorMullen: It's all good.

PR #3404, opened 2015-10-23
Rybrande/url helper multi action name

ryanbrandenburg:


src/Microsoft.AspNet.Mvc.Core/Internal/Routing/LinkGenerationDecisionTree.cs
```diff
@@ -131,13 +133,30 @@ private class LinkGenerationMatchComparer : IComparer<LinkGenerationMatch>
...
             {
-                // For these comparisons lower is better.
-
                 if (x.Entry.Order != y.Entry.Order)
                 {
                     return x.Entry.Order.CompareTo(y.Entry.Order);
                 }
+                
+                //By number of parameters
+                if(x.Entry.Template.Parameters != y.Entry.Template.Parameters)
```
ryanbrandenburg: This needs to be .Parameters.Count obviously.

src/Microsoft.AspNet.Mvc.Core/Internal/Routing/LinkGenerationDecisionTree.cs
```diff
@@ -131,19 +133,36 @@ private class LinkGenerationMatchComparer : IComparer<LinkGenerationMatch>
...
-
+                //For this comparison lower is better
                 if (x.Entry.Order != y.Entry.Order)
                 {
                     return x.Entry.Order.CompareTo(y.Entry.Order);
                 }
 
+                if(x.ParameterCount != y.ParameterCount)
+                {
+                    if(x.Entry?.Template.Segments?.Count != y.Entry?.Template.Segments?.Count)
```
NTaylorMullen: What's happening here? Why does `x.ParameterCount.CompareTo(y.ParameterCount)` not suffice?

src/Microsoft.AspNet.Mvc.Core/Internal/Routing/LinkGenerationMatch.cs
```diff
@@ -9,15 +11,50 @@ public struct LinkGenerationMatch
...
+            _routeValues = routeValues;
         }
 
         public AttributeRouteLinkGenerationEntry Entry { get { return _entry; } }
 
         public bool IsFallbackMatch { get { return _isFallbackMatch; } }
+
+        public IDictionary<string, object> RouteValues { get { return _routeValues; } }
+
+        public int ParameterCount { get {
```
NTaylorMullen: ctrl + k + d

src/Microsoft.AspNet.Mvc.Core/Internal/Routing/LinkGenerationMatch.cs
```diff
@@ -9,15 +11,50 @@ public struct LinkGenerationMatch
...
+                var routeValues = _routeValues;
```
NTaylorMullen: Why capture in a local?

src/Microsoft.AspNet.Mvc.Core/Internal/Routing/LinkGenerationMatch.cs
```diff
@@ -9,15 +11,50 @@ public struct LinkGenerationMatch
...
+                {
+                    return 0;
+                }
+            } }
+
+        public int DefaultParameters
+        {
+            get
+            {
+                var routeValues = _routeValues;
```
NTaylorMullen: ditto question

src/Microsoft.AspNet.Mvc.Core/Internal/Routing/LinkGenerationMatch.cs
```diff
@@ -9,15 +11,50 @@ public struct LinkGenerationMatch
...
 
         public bool IsFallbackMatch { get { return _isFallbackMatch; } }
+
+        public IDictionary<string, object> RouteValues { get { return _routeValues; } }
+
+        public int ParameterCount { get {
+                var routeValues = _routeValues;
+                if((Entry?.Template?.Parameters) != null)
+                {
+                    return Entry.Template.Parameters.Count(p => routeValues.ContainsKey(p.Name));
```
NTaylorMullen: Could simplify the whole method to `return Entry?.Template?.Parameters.Count(p => routeValues.ContainsKey(p.Name)) ?? 0;`

src/Microsoft.AspNet.Mvc.Core/Internal/Routing/LinkGenerationMatch.cs
```diff
@@ -9,15 +11,50 @@ public struct LinkGenerationMatch
...
+            } }
+
+        public int DefaultParameters
+        {
+            get
+            {
+                var routeValues = _routeValues;
+                if (Entry?.Defaults != null)
+                {
+                    return Entry.Defaults.Count(p => routeValues.ContainsKey(p.Key));
```
NTaylorMullen: ditto

test/Microsoft.AspNet.Mvc.Core.Test/Routing/UrlHelperTest.cs
```diff
@@ -741,7 +741,7 @@ public void LinkWithCustomHostAndProtocol_ReturnsExpectedResult()
         [Fact]
         public void Action_RouteValueInvalidation_DoesNotAffectActionAndController()
         {
-            // Arrage
+            // Arrange
```
NTaylorMullen: :+1:

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Mvc.Core/Internal/Routing/LinkGenerationDecisionTree.cs
```diff
@@ -131,19 +133,36 @@ private class LinkGenerationMatchComparer : IComparer<LinkGenerationMatch>
 
             public int Compare(LinkGenerationMatch x, LinkGenerationMatch y)
             {
-                // For these comparisons lower is better.
-
+                //For this comparison lower is better
```
rynowak: spacing

src/Microsoft.AspNet.Mvc.Core/Internal/Routing/LinkGenerationDecisionTree.cs
```diff
@@ -131,19 +133,36 @@ private class LinkGenerationMatchComparer : IComparer<LinkGenerationMatch>
...
                 if (x.Entry.Order != y.Entry.Order)
                 {
                     return x.Entry.Order.CompareTo(y.Entry.Order);
                 }
 
+                if(x.ParameterCount != y.ParameterCount)
```
rynowak: spacing

src/Microsoft.AspNet.Mvc.Core/Internal/Routing/LinkGenerationDecisionTree.cs
rynowak: :space_invader:

rynowak: You don't want to early exit if they are equal, there are more criteria.

rynowak: Why are you comparing number of segments? The sense does not make this.

rynowak: This all deserves some comments.

src/Microsoft.AspNet.Mvc.Core/Internal/Routing/LinkGenerationMatch.cs
```diff
@@ -9,15 +11,50 @@ public struct LinkGenerationMatch
...
             _entry = entry;
             _isFallbackMatch = isFallbackMatch;
+            _routeValues = routeValues;
         }
 
         public AttributeRouteLinkGenerationEntry Entry { get { return _entry; } }
 
         public bool IsFallbackMatch { get { return _isFallbackMatch; } }
+
+        public IDictionary<string, object> RouteValues { get { return _routeValues; } }
```
rynowak: `{ get; }` #PranavPoints

src/Microsoft.AspNet.Mvc.Core/Internal/Routing/LinkGenerationMatch.cs
rynowak: This is totally the wrong name.

src/Microsoft.AspNet.Mvc.Core/Internal/Routing/LinkGenerationDecisionTree.cs
ryanbrandenburg: Imagine we have these routes:
help/general
help
help/{id}

The reason for this PR is that when someone generates a url from urlhelper.Action pointing at the Action help with a parameter Id we want to get "help/1234" not "help?id=1234" (that is, it should match the third route, not the second). But we also want to have the more concrete match try to match first. So if someone goes looking for  "help/general" it needs to match the first route, not the third.

src/Microsoft.AspNet.Mvc.Core/Internal/Routing/LinkGenerationMatch.cs
rynowak: This is the number of parameters with values. You also need to consider defaults in this check.

src/Microsoft.AspNet.Mvc.Core/Internal/Routing/LinkGenerationDecisionTree.cs
```diff
@@ -64,7 +66,7 @@ public IList<LinkGenerationMatch> GetMatches(VirtualPathContext context)
             // to the results.
             for (var i = 0; i < node.Matches.Count; i++)
             {
-                results.Add(new LinkGenerationMatch(node.Matches[i], isFallbackPath));
+                results.Add(new LinkGenerationMatch(node.Matches[i], isFallbackPath, context.Values));
```
rynowak: what is `context.Values`? Does this include ambient values?

test/Microsoft.AspNet.Mvc.FunctionalTests/CustomUrlHelperTests.cs
```diff
@@ -27,6 +32,62 @@ public CustomUrlHelperTests(MvcTestFixture<UrlHelperWebSite.Startup> fixture)
...
+
+            var idPath = await idResult.Content.ReadAsStringAsync();
+            var basePath = await baseResult.Content.ReadAsStringAsync();
+
+            // Assert
+            Assert.NotNull(idPath);
+            Assert.Equal("/api/routevalueusage/base/1234", idPath);
+
+            Assert.NotNull(basePath);
+            Assert.Equal("/api/routevalueusage/base", basePath);
```
rynowak: This is two tests.

test/Microsoft.AspNet.Mvc.FunctionalTests/CustomUrlHelperTests.cs
```diff
@@ -27,6 +32,62 @@ public CustomUrlHelperTests(MvcTestFixture<UrlHelperWebSite.Startup> fixture)
...
+            var idPath = await idResult.Content.ReadAsStringAsync();
+            var basePath = await baseResult.Content.ReadAsStringAsync();
+
+            // Assert
+            Assert.NotNull(idPath);
+            Assert.Equal("/api/routevalueusage/default/1234", idPath);
+
+            Assert.NotNull(basePath);
+            Assert.Equal("/api/routevalueusage/default", basePath);
+        }
```
rynowak: Also two tests

test/Microsoft.AspNet.Mvc.FunctionalTests/CustomUrlHelperTests.cs
```diff
@@ -27,6 +32,62 @@ public CustomUrlHelperTests(MvcTestFixture<UrlHelperWebSite.Startup> fixture)
...
+            var services = new ServiceCollection();
+            services.AddTransient<IUrlHelper, UrlHelper>();
+            services.AddSingleton<IActionContextAccessor, ActionContextAccessor>();
+            services.AddTransient<IActionSelector, DefaultActionSelector>();
+            services.AddSingleton<IActionDescriptorsCollectionProvider, DefaultActionDescriptorsCollectionProvider>();
+            services.AddSingleton<IActionSelectorDecisionTreeProvider, ActionSelectorDecisionTreeProvider>();
+            services.AddTransient<ILoggerFactory, LoggerFactory>();
+
+            return services;
+        }
```
rynowak: Where is any of this junk used?

rynowak:
This needs unit tests. Functional tests are too difficult to diagnose and don't scale well for testing this.

rynowak:
:watch:

src/Microsoft.AspNet.Mvc.Core/Internal/Routing/LinkGenerationDecisionTree.cs
rynowak: > What's happening here? Why does x.ParameterCount.CompareTo(y.ParameterCount) not suffice?

BTW questions like this popping up in your PR is a great hint of where naming is bad or comments need to be added. (probably both)

src/Microsoft.AspNet.Mvc.Core/Internal/Routing/LinkGenerationMatch.cs
ryanbrandenburg: Because "this" isn't available in a lambda which is in a struct. That one was news to me too.

NTaylorMullen: :+1: Didn't see it was a struct.

ryanbrandenburg:
Updated.

src/Microsoft.AspNet.Mvc.Core/Internal/Routing/LinkGenerationDecisionTree.cs
```diff
@@ -131,19 +131,38 @@ private class LinkGenerationMatchComparer : IComparer<LinkGenerationMatch>
...
-
+                // For this comparison lower is better.
                 if (x.Entry.Order != y.Entry.Order)
                 {
                     return x.Entry.Order.CompareTo(y.Entry.Order);
                 }
 
+                if (x.ParametersWithValues != y.ParametersWithValues)
+                {
+                    //If one match has more variables but they have the same segments we want the non-variable one
```
NTaylorMullen: Note: for consistency we typically do comments as `// abc` with a space inbetween the `//` and text.

src/Microsoft.AspNet.Mvc.Core/Internal/Routing/LinkGenerationMatch.cs
```diff
@@ -10,14 +13,42 @@ public struct LinkGenerationMatch
...
             _isFallbackMatch = isFallbackMatch;
+            RouteValues = pathContext.Values;
+            AmbientValues = pathContext.AmbientValues;
         }
 
         public AttributeRouteLinkGenerationEntry Entry { get { return _entry; } }
 
         public bool IsFallbackMatch { get { return _isFallbackMatch; } }
+
+        public IDictionary<string, object> RouteValues { get; }
```
NTaylorMullen: May want to consider exposing this and `AmbientValues` as an `IReadOnlyDictionary<string, object>`.

NTaylorMullen: Ahh, looks like you can't without copying (not worth it).

rynowak: Yeah, this is in an internal namespace + there's a lot of difficultly practically with using `IReadOnlyDictionary<>`

test/Microsoft.AspNet.Mvc.FunctionalTests/CustomUrlHelperTests.cs
```diff
@@ -1,9 +1,14 @@
...
 
+using System;
 using System.Net;
 using System.Net.Http;
 using System.Threading.Tasks;
+using Microsoft.AspNet.Mvc.Infrastructure;
+using Microsoft.AspNet.Mvc.Routing;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
 using Xunit;
```
rynowak: Are these new usings needed?

test/Microsoft.AspNet.Mvc.Core.Test/Routing/UrlHelperTest.cs
```diff
@@ -851,6 +857,210 @@ public void Action_RouteValueInvalidation_DoesNotAffectActionAndController_Actio
...
+                    {
+                        ApplicationServices = serviceMock.Object,
+                        RequestServices = serviceMock.Object,
+                    },
+                    RouteData = new RouteData(),
+                },
+            };
+
+            var serviceCollection = new ServiceCollection();
+            serviceCollection.AddMvc();
```
rynowak: What is this adding that you're not adding manually right after

test/Microsoft.AspNet.Mvc.Core.Test/Routing/UrlHelperTest.cs
```diff
@@ -851,6 +857,210 @@ public void Action_RouteValueInvalidation_DoesNotAffectActionAndController_Actio
...
+
+            [HttpGet("help/{id}")]
+            public void Get(int id)
+            { }
+        }
+
+        private static ControllerActionDescriptorProvider GetActionDescriptorProvider()
+        {
+            //Get all the controllers in this class
+            var controllerTypes = typeof(UrlHelperTest)
```
NTaylorMullen: Use `TypeInfo` reflection based methods.
```C#
var controllerInfo = typeof(Controller).GetTypeInfo();
var controllerTypes = typeof(UrlHelperTest).GetTypeInfo()
    .DeclaredNestedTypes
    .Where(nestedType => controllerInfo.IsAssignableFrom(nestedType))
    .ToList();
```

test/Microsoft.AspNet.Mvc.Core.Test/Routing/UrlHelperTest.cs
```diff
@@ -851,6 +857,210 @@ public void Action_RouteValueInvalidation_DoesNotAffectActionAndController_Actio
...
+
+            var provider = new ControllerActionDescriptorProvider(
+                controllerTypeProvider,
+                new[] { modelProvider },
+                options);
+
+            return provider;
+        }
+
+        private static ILibraryManager CreateLibraryManager()
```
NTaylorMullen: Inline this, looks like it's only ever called once.

test/Microsoft.AspNet.Mvc.Core.Test/Routing/UrlHelperTest.cs
```diff
@@ -851,6 +857,210 @@ public void Action_RouteValueInvalidation_DoesNotAffectActionAndController_Actio
...
+                    },
+                    RouteData = new RouteData(),
+                },
+            };
+
+            var serviceCollection = new ServiceCollection();
+            serviceCollection.AddMvc();
+
+            serviceCollection.AddInstance<ILibraryManager>(CreateLibraryManager());
+            serviceCollection.AddInstance<IActionInvokerFactory>(invokerFactory.Object);
```
NTaylorMullen: You've added a lot of services to this method aren't used in existing; also, there's services here that aren't needed for your tests. Makes me worry about conflating the test expectations.

test/Microsoft.AspNet.Mvc.Core.Test/Routing/UrlHelperTest.cs
```diff
@@ -895,6 +1105,21 @@ private static UrlHelper CreateUrlHelper(IServiceProvider services)
...
+
+            return CreateUrlHelper(appBase, router, services);
+        }
+
+        private static UrlHelper CreateUrlHelper(string appBase, IRouter router, IServiceProvider services)
+        {
+            var context = CreateHttpContext(services, appBase);
+            var actionContext = CreateActionContext(context, router);
+
+            var actionSelector = new Mock<IActionSelector>(MockBehavior.Strict);
```
NTaylorMullen: Should probably get this from the services since it has it now.

test/Microsoft.AspNet.Mvc.Core.Test/Routing/UrlHelperTest.cs
rynowak: Yeah, there's a lot of shenanigans here

NTaylorMullen:
:watch: for test comments.

test/Microsoft.AspNet.Mvc.Core.Test/Routing/UrlHelperTest.cs
```diff
@@ -851,6 +857,210 @@ public void Action_RouteValueInvalidation_DoesNotAffectActionAndController_Actio
             Assert.Equal("/b/Store/Checkout", url);
         }
 
+        [Fact]
+        public void UrlAction_DefaultIgnoredForPrecidence()
```
rynowak: Precedence

test/Microsoft.AspNet.Mvc.Core.Test/Routing/UrlHelperTest.cs
```diff
@@ -851,6 +857,210 @@ public void Action_RouteValueInvalidation_DoesNotAffectActionAndController_Actio
...
+
+            // Act
+            var url = urlHelper.Action(
+                action: "Get",
+                controller: "MatchVariable");
+
+            // Assert
+            Assert.Equal("/api/MatchVariable/help", url);
+        }
+
```
rynowak: This is a good set of tests. Look at testing this more with the AttributeRoute directly - there are existing tests of this flavor, and it should be simpler.

I'd also like to see some tests with more variables.

src/Microsoft.AspNet.Mvc.Core/Internal/Routing/LinkGenerationDecisionTree.cs
ryanbrandenburg: Ah ok, I had seen it the other way a couple times in code and thought it was a personal preference thing.

test/Microsoft.AspNet.Mvc.Core.Test/Routing/UrlHelperTest.cs
ryanbrandenburg: There are certainly services in the new function which didn't exist in the old one. I actually went back and forth on weather I wanted to merge the two together, if we want to separate them I'm fine with that. Unfortunately everything I added here I ended up needing because even if it doesn't get used by our case, the underlying code will still complain that the service isn't registered. IActionContextAccessor, IActionContextAccessor, IActionSelector, ILoggerFactory, and DiagnosticSource for example are all needed so that EnsureServices in MvcRouteHandler doesn't blow up. I can have a second go at Mocking out more thing without mocking out the behavior we need to try to decrease the number if we want.

NTaylorMullen: I don't feel strongly here. @rynowak usually has a stronger opinion on tests having too much going on.

ryanbrandenburg:
:up::date:

test/Microsoft.AspNet.Mvc.Core.Test/Routing/InnerAttributeRouteTest.cs
```diff
@@ -361,6 +361,63 @@ public class InnerAttributeRouteTest
...
+        [InlineData("template", "template/{parameter:int=1033}", "/template", false)]
+        [InlineData("template", "template/{parameter:int=5}", "/template/5", true)]
+        public void AttributeRoute_GenerateLink_UseAvailableVariables(
+            string firstTemplate,
+            string secondTemplate,
+            string expectedPath,
+            bool useValues)
+        {
+            // Arrange
+            var expectedGroup = CreateRouteGroup(0, useValues ? secondTemplate : firstTemplate);
```
NTaylorMullen: When you have conditional statements in the arrange section of your tests it usually indicates the need to break out the test (even if it means duplicating lots of code).

ryanbrandenburg: Yeah, I definitely feel a little gross doing conditionals like that, but have usually erred on the side of not duplicating code just because of it in the past, but I'll split it out if that's the general opinion.

test/Microsoft.AspNet.Mvc.Core.Test/Routing/InnerAttributeRouteTest.cs
```diff
@@ -361,6 +361,63 @@ public class InnerAttributeRouteTest
         }
 
         [Theory]
+        [InlineData("template", "template/{parameter}", "/template/5", true)]
+        [InlineData("template", "template/{parameter}", "/template", false)]
+        [InlineData("template/{parameter}", "template/{parameter}/{id}", "/template/5/1234", true)]
+        [InlineData("template", "template/{parameter:int=1033}", "/template", false)]
+        [InlineData("template", "template/{parameter:int=5}", "/template/5", true)]
+        public void AttributeRoute_GenerateLink_UseAvailableVariables(
```
NTaylorMullen: Once broken out names should reflect desired functionality

NTaylorMullen:
:shipit: once test is broken out and @rynowak is happy with it.

src/Microsoft.AspNet.Mvc.Core/Internal/Routing/LinkGenerationDecisionTree.cs
```diff
@@ -131,19 +131,38 @@ private class LinkGenerationMatchComparer : IComparer<LinkGenerationMatch>
...
+                // For this comparison lower is better.
                 if (x.Entry.Order != y.Entry.Order)
                 {
                     return x.Entry.Order.CompareTo(y.Entry.Order);
                 }
 
+                if (x.ParametersWithValues != y.ParametersWithValues)
+                {
+                    // If one match has more variables but they have the same segments we want the non-variable one
+                    // e.g. for 'help/general' and 'help/{id}' we want to make sure help/general is the first option
```
rynowak: Your example isn't accurate - these have different numbers of parameters.

src/Microsoft.AspNet.Mvc.Core/Internal/Routing/LinkGenerationDecisionTree.cs
```diff
@@ -131,19 +131,38 @@ private class LinkGenerationMatchComparer : IComparer<LinkGenerationMatch>
...
+                    if (x.Entry?.Template.Segments?.Count != y.Entry?.Template.Segments?.Count)
```
rynowak: If `Entry` is null we could have already thrown.

Did you actually find a case where any of these are null? I don't think `.Segments` can be null either.

rynowak: Also null != 0 - if this really can be null, you want a `?? 0` on both

src/Microsoft.AspNet.Mvc.Core/Internal/Routing/LinkGenerationMatch.cs
```diff
@@ -10,14 +13,42 @@ public struct LinkGenerationMatch
...
+        public IDictionary<string, object> AmbientValues { get; }
+
+        public int ParametersWithValues
+        {
+            get
+            {
+                var t = this;
+                return Entry?.Template?.Parameters?.Count(p => (t.RouteValues.ContainsKey(p.Name) && t.RouteValues[p.Name] != p.DefaultValue) ||
+                    (t.AmbientValues.ContainsKey(p.Name) && t.AmbientValues[p.Name] != p.DefaultValue)) ?? 0;
+            }
```
rynowak: This really needs to be simplied.

src/Microsoft.AspNet.Mvc.Core/Internal/Routing/LinkGenerationMatch.cs
```diff
@@ -10,14 +13,42 @@ public struct LinkGenerationMatch
...
+        public IDictionary<string, object> RouteValues { get; }
+
+        public IDictionary<string, object> AmbientValues { get; }
+
+        public int ParametersWithValues
+        {
+            get
+            {
+                var t = this;
+                return Entry?.Template?.Parameters?.Count(p => (t.RouteValues.ContainsKey(p.Name) && t.RouteValues[p.Name] != p.DefaultValue) ||
```
rynowak: Don't use `Linq` here

src/Microsoft.AspNet.Mvc.Core/Internal/Routing/LinkGenerationMatch.cs
```diff
@@ -10,14 +13,42 @@ public struct LinkGenerationMatch
...
+
+        public IDictionary<string, object> RouteValues { get; }
+
+        public IDictionary<string, object> AmbientValues { get; }
+
+        public int ParametersWithValues
+        {
+            get
+            {
+                var t = this;
```
rynowak: lolwut

src/Microsoft.AspNet.Mvc.Core/Internal/Routing/LinkGenerationMatch.cs
```diff
@@ -10,14 +13,42 @@ public struct LinkGenerationMatch
...
+                    (t.AmbientValues.ContainsKey(p.Name) && t.AmbientValues[p.Name] != p.DefaultValue)) ?? 0;
+            }
+        }
+
+        public int DefaultParameters
+        {
+            get
+            {
+                var t = this;
+                return Entry?.Defaults?.Count(p => t.RouteValues.ContainsKey(p.Key)) ?? 0;
```
rynowak: You want the number of parameters that have their default values. This doesn't match values, just keys.

test/Microsoft.AspNet.Mvc.Core.Test/Routing/InnerAttributeRouteTest.cs
```diff
@@ -361,6 +361,63 @@ public class InnerAttributeRouteTest
...
+            {
+                selectedGroup = (string)ctx.ProvidedValues[AttributeRouting.RouteGroupKey];
+                ctx.IsBound = true;
+            })
+            .Returns((VirtualPathData)null);
+
+            var matchingRoutes = Enumerable.Empty<AttributeRouteMatchingEntry>();
+
+            var firstEntry = CreateGenerationEntry(firstTemplate, requiredValues: null);
+            var secondEntry = CreateGenerationEntry(secondTemplate, requiredValues: null, order: 0);
```
rynowak: Do these values end up with a different `Order`? if it's the same then leave it out :+1:

test/Microsoft.AspNet.Mvc.Core.Test/Routing/InnerAttributeRouteTest.cs
```diff
@@ -361,6 +361,63 @@ public class InnerAttributeRouteTest
...
+        [InlineData("template", "template/{parameter}", "/template/5", true)]
+        [InlineData("template", "template/{parameter}", "/template", false)]
+        [InlineData("template/{parameter}", "template/{parameter}/{id}", "/template/5/1234", true)]
+        [InlineData("template", "template/{parameter:int=1033}", "/template", false)]
+        [InlineData("template", "template/{parameter:int=5}", "/template/5", true)]
+        public void AttributeRoute_GenerateLink_UseAvailableVariables(
+            string firstTemplate,
+            string secondTemplate,
+            string expectedPath,
+            bool useValues)
```
rynowak: What is `useValues` - what does it mean?

rynowak:
:watch:

test/Microsoft.AspNet.Mvc.Core.Test/Routing/InnerAttributeRouteTest.cs
ryanbrandenburg: Got remove after @NTaylorMullen 's comments.

src/Microsoft.AspNet.Mvc.Core/Internal/Routing/LinkGenerationMatch.cs
ryanbrandenburg: If you're complaining about aliasing this that's because this isn't available for a lambda inside a struct. It'll get removed when I get rid of the Linq anyway.

test/Microsoft.AspNet.Mvc.Core.Test/Routing/InnerAttributeRouteTest.cs
rynowak: currently :watch: for test updates

ryanbrandenburg:
Updated most of these, still trying to figure out how to figure out if a parameter was received from default or not.

ryanbrandenburg:
:up: :date: I moved to @rynowak 's suggestion of a more clear model for calculating these, seems much cleaner now, and reduced the amount of code I had to touch.

src/Microsoft.AspNet.Mvc.Core/Routing/AttributeRoutePrecedence.cs
```diff
@@ -13,7 +13,7 @@ namespace Microsoft.AspNet.Mvc.Routing
     /// </summary>
     public static class AttributeRoutePrecedence
     {
-        public static decimal Compute(RouteTemplate template)
+        public static decimal ComputeMatched(RouteTemplate template)
```
rynowak: Add a brief comment explaining the jargon'ey `Matched` term

src/Microsoft.AspNet.Mvc.Core/Routing/AttributeRoutePrecedence.cs
```diff
@@ -32,13 +32,66 @@ public static decimal Compute(RouteTemplate template)
             return precedence;
         }
 
+        public static decimal ComputeGenerated(RouteTemplate template)
```
rynowak: Add a brief comment explaining the jargon'ey `Generated` term

rynowak: Should we add a few examples to each of these in comments?

src/Microsoft.AspNet.Mvc.Core/Routing/AttributeRoutePrecedence.cs
rynowak: Not asking for a doc comment with all the pomp and circumstance - just a normal comment will do.

src/Microsoft.AspNet.Mvc.Core/Routing/AttributeRoutePrecedence.cs
```diff
@@ -32,13 +32,66 @@ public static decimal Compute(RouteTemplate template)
...
+                precedence += decimal.Divide(digit, (decimal)Math.Pow(10, i));
+            }
+
+            return precedence;
+        }
+
+        // Segments have the following order:
+        // 6 - Literal segments
+        // 5 - Multi-part segments
+        // 4 - Constrained parameter segments ts
```
rynowak: ts?

src/Microsoft.AspNet.Mvc.Core/Routing/AttributeRoutePrecedence.cs
```diff
@@ -32,13 +32,66 @@ public static decimal Compute(RouteTemplate template)
...
+
+                precedence += decimal.Divide(digit, (decimal)Math.Pow(10, i));
+            }
+
+            return precedence;
+        }
+
+        // Segments have the following order:
+        // 6 - Literal segments
+        // 5 - Multi-part segments
```
rynowak: Why treat multi-part any differently than we do for matching?

test/Microsoft.AspNet.Mvc.Core.Test/Routing/AttributeRoutePrecedenceTests.cs
```diff
@@ -63,7 +63,7 @@ private static decimal Compute(string template)
             options.SetupGet(o => o.Value).Returns(new RouteOptions());
 
             var parsed = TemplateParser.Parse(template);
-            return AttributeRoutePrecedence.Compute(parsed);
+            return AttributeRoutePrecedence.ComputeMatched(parsed);
```
rynowak: Tests for the new method?

test/Microsoft.AspNet.Mvc.Core.Test/Routing/InnerAttributeRouteTest.cs
```diff
@@ -360,6 +360,121 @@ public class InnerAttributeRouteTest
             Assert.False(context.IsHandled);
         }
 
+        [Fact]
+        public void AttributeRoute_GenerateLink_DefaultHidesBase()
```
rynowak: `DefaultHidesBase` -> `LongerTemplateWithDefaultIsMoreSpecific`

test/Microsoft.AspNet.Mvc.Core.Test/Routing/InnerAttributeRouteTest.cs
```diff
@@ -360,6 +360,121 @@ public class InnerAttributeRouteTest
...
+            // The Binder binds to /template
+            Assert.Equal(new PathString($"/template"), result.VirtualPath);
+            Assert.Same(route, result.Router);
+            Assert.Empty(result.DataTokens);
+            // Even though the path was /template, the group generated from was /template/{paramter:int=1003}
+            Assert.Equal(expectedGroup, selectedGroup);
+        }
+        
+        [Theory]
+        [InlineData("template/{parameter:int=5}", "template", "/template/5")]
```
rynowak: `Theory` with only set of inputs? this kid is a loose cannon

test/Microsoft.AspNet.Mvc.Core.Test/Routing/InnerAttributeRouteTest.cs
```diff
@@ -360,6 +360,121 @@ public class InnerAttributeRouteTest
             Assert.False(context.IsHandled);
         }
 
+        [Fact]
```
rynowak: Do we need more tests with things like catch-all or complex-segments?

We didn't really need to thoroughly test this feature before because it wasn't sophisticated at all. Now we want to have tests covering more complex scenarios as validation that 'feels' right.

rynowak:
:watch:

src/Microsoft.AspNet.Mvc.Core/Routing/AttributeRoutePrecedence.cs
ryanbrandenburg: I was lead by this comment: "// Multi-part segments should appear after literal segments but before parameter segments" which was already in the code to believe that we did want to treat multi-part segments and constrained parameters differently. If that's not the case I'll modify ComputeGenerationDigit and alter the existing comment so it doesn't contradict the code.

ryanbrandenburg:
:up: :date:

src/Microsoft.AspNet.Mvc.Core/Routing/AttributeRouteLinkGenerationEntry.cs
```diff
@@ -34,9 +34,9 @@ public class AttributeRouteLinkGenerationEntry
         public int Order { get; set; }
 
         /// <summary>
-        /// The precedence of the template.
+        /// The precedence of the template for link generation.
```
rynowak: Add that higher is better :+1:

src/Microsoft.AspNet.Mvc.Core/Routing/AttributeRoutePrecedence.cs
```diff
@@ -32,17 +37,73 @@ public static decimal Compute(RouteTemplate template)
             return precedence;
         }
 
+        // Compute the precedence for generating a url
+        // e.g.: /api/template          == 6.6
+        //       /api/template/{id}     == 6.63
+        //       /api/{id:int}          == 6.4
+        //       /api/template/{id:int} == 6.64
```
rynowak: These examples aren't right with the change to the values

test/Microsoft.AspNet.Mvc.Core.Test/Routing/AttributeRoutePrecedenceTests.cs
```diff
@@ -16,11 +17,25 @@ public class AttributeRoutePrecedenceTests
         [InlineData("Employees/{id}", "Employees/{employeeId}")]
         [InlineData("abc", "def")]
         [InlineData("{x:alpha}", "{x:int}")]
-        public void Compute_IsEqual(string xTemplate, string yTemplate)
+        public void ComputeMatched_IsEqual(string xTemplate, string yTemplate)
```
rynowak: ComputeMatches

rynowak:

test/Microsoft.AspNet.Mvc.Core.Test/Routing/AttributeRoutePrecedenceTests.cs
ryanbrandenburg: Actually, the name of the underlying function we're testing is ComputeMatched, it's the local helper function that's named wrong.

PR #3420, opened 2015-10-27
Fix build after DNX change 167fff5.

cesarbs:
@pranavkm @NTaylorMullen @davidfowl


cesarbs:
Nevermind, @pakrym already took care of it.

PR #3421, opened 2015-10-27
Rename config.json to appsettings.json for Beta 8

scottaddie:
The `config.json` file in the MvcSample.Web project needs to be updated for the Beta 8 file name change. The new name is `appsettings.json`.


Eilon:
@scottaddie what is this change for?

scottaddie:
@Eilon - See [this issue](https://github.com/aspnet/Templates/issues/261) in the Templates repo for some context around this PR.

Eilon:
Ah, I see. But presumably something else in the app needs to change in order to use the newly named file. I don't think the name is magical - it has to be referenced somewhere.

scottaddie:
It's typically referenced by name in `Startup.cs`, as explained in the [docs](https://docs.asp.net/en/latest/fundamentals/configuration.html). I didn't find any references to it in the entire solution.

Eilon:
Maybe this one just needs to be deleted, then?

scottaddie:
Would it be useful to keep `appsettings.json` and add some code to `Startup.cs` to demonstrate its use? I would guess that most developers will have this JSON file in their projects to store their custom configuration values.

Eilon:
This "sample" is more of a test playground to verify the behavior of features, and not so much a sample in the typical sense of demonstrating how features can be used. If something isn't used at all, it might be best to just delete it...

scottaddie:
Agreed. I'll close this PR and open a new one for deleting the unused `config.json` file.

pranavkm:
Merged in https://github.com/aspnet/Mvc/commit/3c46662d91b1950e078509482bc0bce454858346

PR #3436, opened 2015-10-28
React to aspnet/Razor#580.

NTaylorMullen:
- Updated `MvcRazorHost` to configure `Type` full names.
- Updated code generation tests.

Razor PR: https://github.com/aspnet/Razor/pull/587


ajaybhargavb:
:shipit:

dougbu:
:shipit:

src/Microsoft.AspNet.Mvc.Razor.Host/MvcRazorHost.cs
```diff
@@ -95,7 +95,7 @@ internal MvcRazorHost(IChunkTreeCache chunkTreeCache, RazorPathNormalizer pathNo
                     ScopeManagerBeginMethodName = nameof(TagHelperScopeManager.Begin),
                     ScopeManagerEndMethodName = nameof(TagHelperScopeManager.End),
 
-                    TagHelperContentTypeName = nameof(TagHelperContent),
+                    TagHelperContentTypeName = typeof(TagHelperContent).FullName,
```
Eilon: Needs to say `global::Microsoft.Foo.Bar` to avoid conflicts with sub-namespaces named `Microsoft`.

NTaylorMullen: Super solid :+1: will make the change. It's going to be a bit broader of a change with that modification.

NTaylorMullen:
:up: :date: to include `global::` changes.

PR #3442, opened 2015-10-29
Fix MVC master branch

kichalla:
- Updated Autofac package reference from beta6 to beta8
- Removed package references to "Ketrel" and "Microsoft.AspNet.Server.IIS"

@pranavkm


pranavkm:
:shipit:

Issue #3140, opened 2015-09-15
Content negotiation fails in subsequent requests accepting same media type


dougbu:
The @d03a851 could not remove the functional test `TestHelper` class in part because three test methods in `WebApiCompatShimBasicTest` cannot share a `TestServer` instance when using Core CLR. This issue is separate from #3139 because nothing about these tests should require a one-off server and we could have an underlying product issue here. The problem does not appear to involve inconsistent contexts, `HttpRequestMessage` or header reuse, or content negotiation state held in the server. Even so the tests fail consistently.

#### Details
Failures when sharing a `TestServer` instance and using Core CLR affect the following test methods:
- `ApiController_CreateResponse_Conneg()`, 2 of 4 test cases fail
- `ApiController_CreateResponse_HardcodedMediaType()`, 2 of 2 test cases fail
- `ApiController_CreateResponse_Conneg_Error()`, 2 of 4 test cases fail

The specific failures do not change between running every functional test and running just those in `WebApiCompatShimBasicTest`. On the other hand, failures change if the `[InlineData]` order changes or test methods are run individually. Removing the apparent failure cases from one run does not remove all failures in the next, just shifts to one of the remaining cases. Failures at least seem consistent between runs for a particular setup.

Nothing should be running in parallel when running just tests in this class. xUnit treats each test class as a collection and runs tests within collections alphabetically, not in parallel. The output sometimes looks mixed-up when I add `Console.WriteLine()` debugging statements but that probably relates to flushing and not concurrency.

The dougbu/max.fixtures.with.debugging server branch uses fixtures for these three test methods and includes some `WriteLine()` and `Assert()` checks to eliminate possibilities.

The `ApiController_CreateResponse_HardcodedMediaType()` tests fail when asserting the response status code e.g.
```
    Microsoft.AspNet.Mvc.FunctionalTests.WebApiCompatShimBasicTest.ApiController_CreateResponse_HardcodedMediaType(mediaType: "text/xml") [FAIL]
      Assert.Equal() Failure
      Expected: OK
      Actual:   InternalServerError
      Stack Trace:
        C:\dd\dnx\MVC\test\Microsoft.AspNet.Mvc.FunctionalTests\WebApiCompatShimBasicTest.cs(326,0): at Microsoft.AspNet.Mvc.FunctionalTests.WebApiCompatShimBasicTest.<ApiController_CreateResponse_HardcodedMediaType>d__19.MoveNext()
[...]

PR #3450, opened 2015-10-30
Content negotiation fails in subsequent requests accepting same media type

dougbu:
- #3140
- clone `MediaTypeHeaderValue` instance before updating it when content negotiation succeeds
  - avoids changes to `MediaTypeConstants` properties and `OutputFormatter.SupportedMediaTypes` entries
  - `MediaTypeHeaderValue.Clone()` does not exist in our DNX Core fork of this class
  - in previous implementation, was called defensively rather than when required
- update `WebApiCompatShimBasicTest` functional tests to use `MvcTestFixture<TStartup>` everywhere
  - #3140 blocked that final migration
  - remove `TestHelper` since it's no longer referenced

nits:
- remove comments mentioning `TestHelper`
- correct spelling of "negotiation"


dougbu:
# remove `TestHelper` since it's no longer referenced

Just in case you missed it.

src/Microsoft.AspNet.Mvc.WebApiCompatShim/ContentNegotiator/DefaultContentNegotiator.cs
```diff
@@ -85,14 +83,25 @@ public DefaultContentNegotiator(bool excludeMatchOnTypeOnly)
...
                 {
-                    bestFormatterMatch.MediaType.CharSet = bestEncodingMatch.WebName;
+                    // Clone media type value since this is not done defensively in this implementation.
+                    // `MediaTypeHeaderValue` lacks a Clone() method in this runtime. Fortunately, this is the only
+                    // update to an existing instance we need.
+                    var clonedMediaType = new MediaTypeHeaderValue(bestMediaType.MediaType);
+                    foreach (var parameter in bestMediaType.Parameters)
+                    {
+                        clonedMediaType.Parameters.Add(new NameValueHeaderValue(parameter.Name, parameter.Value));
+                    }
```
rynowak: Annoying that the the copy constructor is protected here.

rynowak:
:shipit:

dougbu:
53060be

PR #3452, opened 2015-10-31
Fixing tests broken due to encoding changes

pranavkm:
cc @dougbu \ @Tratcher


dougbu:
Do not do this. The result will be unstable tests of a scenario that will never happen in real life. `null` and `Empty` `string`s are not encoded and should not be encoded in tests.

I am about to send out a PR for HttpAbstractions fixing the test encoders to work as the old system did.

dougbu:
Why was this checked in?

Eilon:
Yeah seems like this should be closed and addressed in https://github.com/aspnet/Mvc/pull/3453...

PR #3461, opened 2015-11-02
Adding Controller.Challenge and Controller.Forbid

pranavkm:
cc @Eilon \ @Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
@@ -1245,64 +1245,64 @@ namespace Microsoft.AspNet.Mvc
             => new ChallengeResult(authenticationSchemes, properties);
 
         /// <summary>
-        /// Creates a <see cref="ForbiddenResult"/>.
+        /// Creates a <see cref="ForbidResult"/>.
         /// </summary>
-        /// <returns>The created <see cref="ForbiddenResult"/> for the response.</returns>
+        /// <returns>The created <see cref="ForbidResult"/> for the response.</returns>
         [NonAction]
-        public virtual ForbiddenResult Forbid()
-            => new ForbiddenResult();
+        public virtual ForbidResult Forbid()
+            => new ForbidResult();
 
         /// <summary>
-        /// Creates a <see cref="ForbiddenResult"/> with the specified authentication scheme.
+        /// Creates a <see cref="ForbidResult"/> with the specified authentication scheme.
         /// </summary>
         /// <param name="authenticationScheme">The authentication scheme to challenge.</param>
-        /// <returns>The created <see cref="ForbiddenResult"/> for the response.</returns>
+        /// <returns>The created <see cref="ForbidResult"/> for the response.</returns>
         [NonAction]
-        public virtual ForbiddenResult Forbid(string authenticationScheme)
-            => new ForbiddenResult(authenticationScheme);
+        public virtual ForbidResult Forbid(string authenticationScheme)
+            => new ForbidResult(authenticationScheme);
 
         /// <summary>
-        /// Creates a <see cref="ForbiddenResult"/> with the specified authentication schemes.
+        /// Creates a <see cref="ForbidResult"/> with the specified authentication schemes.
         /// </summary>
         /// <param name="authenticationScheme">The authentication schemes to challenge.</param>
-        /// <returns>The created <see cref="ForbiddenResult"/> for the response.</returns>
+        /// <returns>The created <see cref="ForbidResult"/> for the response.</returns>
         [NonAction]
-        public virtual ForbiddenResult Forbid(IList<string> authenticationSchemes)
-            => new ForbiddenResult(authenticationSchemes);
+        public virtual ForbidResult Forbid(IList<string> authenticationSchemes)
+            => new ForbidResult(authenticationSchemes);
 
         /// <summary>
-        /// Creates a <see cref="ForbiddenResult"/> with the specified <paramref name="properties" />.
+        /// Creates a <see cref="ForbidResult"/> with the specified <paramref name="properties" />.
         /// </summary>
         /// <param name="properties"><see cref="AuthenticationProperties"/> used to perform the authentication
         /// challenge.</param>
-        /// <returns>The created <see cref="ForbiddenResult"/> for the response.</returns>
+        /// <returns>The created <see cref="ForbidResult"/> for the response.</returns>
         [NonAction]
-        public virtual ForbiddenResult Forbid(AuthenticationProperties properties)
-            => new ForbiddenResult(properties);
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
```diff
@@ -1181,6 +1183,128 @@ public virtual CreatedAtRouteResult CreatedAtRoute(string routeName, object rout
         }
 
         /// <summary>
+        /// Creates a <see cref="ChallengeResult"/>.
+        /// </summary>
+        /// <returns>The created <see cref="ChallengeResult"/> for the response.</returns>
+        [NonAction]
+        public virtual ChallengeResult Challenge()
+            => new ChallengeResult();
```
Eilon: :100: PranavPoints awarded.

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
```diff
@@ -1181,6 +1183,128 @@ public virtual CreatedAtRouteResult CreatedAtRoute(string routeName, object rout
...
+
+        /// <summary>
+        /// Creates a <see cref="ChallengeResult"/> with the specified authentication scheme.
+        /// </summary>
+        /// <param name="authenticationScheme">The authentication scheme to challenge.</param>
+        /// <returns>The created <see cref="ChallengeResult"/> for the response.</returns>
+        [NonAction]
+        public virtual ChallengeResult Challenge(string authentiationScheme)
```
Eilon: -:100: PranavPoints for typo in param name.

Eilon: And -:100: PranavPoints for each of the subsequent methods that has this error :smile:

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
```diff
@@ -1181,6 +1183,128 @@ public virtual CreatedAtRouteResult CreatedAtRoute(string routeName, object rout
...
+        public virtual ForbiddenResult Forbid()
+            => new ForbiddenResult();
+
+        /// <summary>
+        /// Creates a <see cref="ForbiddenResult"/> with the specified authentication scheme.
+        /// </summary>
+        /// <param name="authenticationScheme">The authentication scheme to challenge.</param>
+        /// <returns>The created <see cref="ForbiddenResult"/> for the response.</returns>
+        [NonAction]
+        public virtual ForbiddenResult Forbid(string authentiationScheme)
```
Eilon: Why not `Forbidden`? I guess it sounds weird either way... but do we ever have much of a deviation between the factory name and the result type?

Eilon:
:clock1130:

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
pranavkm: Oops

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
pranavkm: It calls `AuthenticationManager.ForbidAsync` so I thought calling it `Forbid` made sense?

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
```diff
@@ -1181,6 +1183,128 @@ public virtual CreatedAtRouteResult CreatedAtRoute(string routeName, object rout
...
+        public virtual ForbiddenResult Forbid()
+            => new ForbiddenResult();
+
+        /// <summary>
+        /// Creates a <see cref="ForbiddenResult"/> with the specified authentication scheme.
+        /// </summary>
+        /// <param name="authenticationScheme">The authentication scheme to challenge.</param>
+        /// <returns>The created <see cref="ForbiddenResult"/> for the response.</returns>
+        [NonAction]
+        public virtual ForbiddenResult Forbid(string authenticationScheme)
```
pranavkm: @Eilon - `Forbidden`?

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
```diff
@@ -1181,6 +1183,128 @@ public virtual CreatedAtRouteResult CreatedAtRoute(string routeName, object rout
...
+        public virtual ChallengeResult Challenge(string authenticationScheme)
+            => new ChallengeResult(authenticationScheme);
+
+        /// <summary>
+        /// Creates a <see cref="ChallengeResult"/> with the specified authentication schemes.
+        /// </summary>
+        /// <param name="authenticationScheme">The authentication schemes to challenge.</param>
+        /// <returns>The created <see cref="ChallengeResult"/> for the response.</returns>
+        [NonAction]
+        public virtual ChallengeResult Challenge(IList<string> authenticationSchemes)
```
Tratcher: Does it make sense to include every possible overload here?

Eilon: @pranavkm and I chatted and I said "eh, why not" have them all... or are some of them truly more obscure?

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
Eilon: Ah, then maybe the type should be `ForbidResult`? Yeah, also sounds weird at least at first, but perhaps that's more accurately what it is.

pranavkm:
I'll change the type to `ForbidResult`.

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
rynowak: IMO `Forbidden` > `Forbid` - we have existing precedent for using the name of the status code as the method name. Doing something not exactly equivalent just leads to more confusion.

Eilon: These ones are more about the security operation and less about the HTTP status code. They do *also* set the status code, but it's far more than that. E.g. the Challenge factory doesn't return a Challenge status code (because there isn't any status code with that name).

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
Tratcher: *shrug* it just seems verbose.

Eilon: Then just set it to `.Informational` :trollface:

pranavkm:
:up: :date:

Eilon:
:shipit:

Issue #3449, opened 2015-10-30
LinkTagHelper's fallback meta tag isn't W3C compliant


madskristensen:
This generated markdup isn't valid W3C compliant markup:

```html
<meta name="x-stylesheet-fallback-test" class="invisible" />
```
Running it through the W3C validator produces this error:

`Element meta is missing one or more of the following attributes: "content", "property".`

NTaylorMullen:
This should be rc2. /cc @Eilon @danroth27

NTaylorMullen:
Played around with this and I'm proposing we go with: `<meta property="x-stylesheet-fallback-test" content class="..." />`.

Meta tags aren't valid in the body with a `name` tag sadly.

NTaylorMullen:
decf8823414146fe6feaba0ed9fcfe4044258e04

PR #3463, opened 2015-11-03
Update `LinkTagHelper`s `meta` tag to be w3c compliant.

NTaylorMullen:
- Prior to this change the `<meta ...>` generated had a name attribute. Problem with this is `meta` tags aren't allowed to have `name` attributes in the body of an HTML page. As an alternative we're using the `property` attribute which is used similarly to `name` but for social graphs. Lastly, having a minimized content attribute showcases that there's no value that's trying to be portrayed.

#3449


src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -338,7 +338,7 @@ private void BuildFallbackBlock(TagHelperContent builder)
 
                 // Build the <meta /> tag that's used to test for the presence of the stylesheet
                 builder
-                    .AppendHtml("<meta name=\"x-stylesheet-fallback-test\" class=\"")
+                    .AppendHtml("<meta property=\"x-stylesheet-fallback-test\" content class=\"")
```
Eilon: Having a minimized `content` attribute seems wrong. What does it even mean?

Is this issue really a matter of "person who wrote the `<meta>` tag wasn't HTML-compliant, so too bad"?

dougbu: Agree w/ @Eilon. Where HTML and tag helpers add content, the content should be XHTML-compliant. The minimized choice for this attribute goes against that. (And I'm not sure minimization is legal for anything except Boolean attributes.)

Separately where is the `property` attribute defined? Not mentioned in http://www.w3.org/TR/html5/document-metadata.html#the-meta-element

Eilon: Yeah, but my main concern is that I think this is perhaps just user error. It's certainly not the job of tag helpers to correct erroneous markup. Am I correct in arguing that it's erroneous markup to begin with?

NTaylorMullen: > Having a minimized content attribute seems wrong. What does it even mean?

A minimized `content` here is just less code and HTML5 w3c compliant; ultimately this `meta` tag is used as a dummy element to verify that a CSS class properly applies. 

> the content should be XHTML-compliant. The minimized choice for this attribute goes against that. (And I'm not sure minimization is legal for anything except Boolean attributes.)


There's a lot of things about this `<meta>` tag that is non-XHTML compliant. For instance it can't have a `class` attribute or even exist in the `<body>`, sadly, we do both. I'm open to other "dummy" elements but `meta` tags/HTML5 compliance seem to be the most correct (purely for metadata) / consistent (been using HTML5 compliance as a baseline) approach.

> Separately where is the property attribute defined? Not mentioned in http://www.w3.org/TR/html5/document-metadata.html#the-meta-element

The `property` attribute is an open graph protocol attribute which w3c seems to validate for yet is rarely used or consumed. It's rarity is valuable to us since the tag is used in a "dummy" manner; the spec is [here](http://ogp.me/).

dougbu: > There's a lot of things about this <meta> tag that is non-XHTML compliant. For instance ...

Interesting but not quite the same concern. Those are semantic issues. This attribute is invalid XML syntax. If we go with using the obscure open graph micro-format, why not use `content=""`?

DamianEdwards: @Eilon this is markup that our Tag Helper produces.

@NTaylorMullen we don't put it in the body, we put it next to the `<link />` which should always be in the `<head>`. And who says it can't have a `class` attribute? The report was that it's missing some required attributes, not that it has some illegal attributes.

DamianEdwards: Just adding `content=""` fixes the validation error. That's all we should be doing here.

NTaylorMullen: @DamianEdwards If you run a meta tag with a class attribute through w3c doesn't like it:

**Input:**
```HTML
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.1//EN" "http://www.w3.org/TR/xhtml11/DTD/xhtml11.dtd"><!-- <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.1//EN" "http://www.w3.org/TR/xhtml11/DTD/xhtml11.dtd"> -->
<html>
<head>
	<title>Test</title>
	<meta name="hello" content="world" class="foo" />
</head>
<body>
</body>
</html>
```

**Error related to meta tag:**
image

DamianEdwards: that's 'cos you're validating against XHTML 1.1. We don't care about that. Validate it with `<!doctype html>` in order to invoke the HTML5 rules.

DamianEdwards: This validates just fine:
``` HTML
<!doctype html>
<html xmlns="http://www.w3.org/1999/xhtml">
<head>
<meta charset="utf-8" />
<title>Foo</title>
<link rel="stylesheet" href="foo.css" />
<meta name="x-stylesheet-check" content="" class="invisible" />
</head>
<body>
</body>
</html>
```

NTaylorMullen: @DamianEdwards I agree, @dougbu was worrying about being XHTML compliant so I ran it through the XHTML validator. I then chose to do the `property` instead of `name` approach so if a user for some resaon adds it to the body it doesn't barf.

NTaylorMullen: Note that the version in this PR is HTML5 compliant whether its in the `head` or `body`.

DamianEdwards: I **think** he meant be well-formed XML, rather than actually adhere to the strict XHTML 1.1 doctype. That is practically impossible in today's web applications and in practice doesn't matter at all. The validator understands the concept of an HTML5 doctype with an XHTML xmlns declaration (like my sample above) that invokes the XML parsing rules, as opposed to SGML. I agree that we should generally emit XHTML rather than SGML.

DamianEdwards: @NTaylorMullen that's nice, but `<link />` is only valid in the `<head>` so it doesn't mean much :smile:

DamianEdwards: And the version in this PR looks strange. Change it.

NTaylorMullen: Ahhh If @dougbu was referring to the well-formed XML then I completely understand :+1:. I'll make the change to do `content=""` and revert back to `name="..."`.

NTaylorMullen:
:up: :date:

test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.Link.Encoded.html
```diff
@@ -1,4 +1,4 @@
-﻿<!doctype html>
+<!doctype html>
```
DamianEdwards: What changed here? BOM?

DamianEdwards:
:shipit:

test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.Link.Encoded.html
dougbu: BOOM

dougbu:
:shipit: and, yes, I was concerned about well-formed XML not :100:% XHTML compliance.

NTaylorMullen:
Will get this in once the build settles down. :smile:

Eilon:
:shipit: me too.

PR #3480, opened 2015-11-04
fix #3414 : Change DataAnnotationsClientModelValidator to localize error message

### LINKED ISSUE #3414: RegularExpressionAttribute generates wrong data-val-regex message
Using the RegularExpressionAttribute with beta8 like C# data object:
```
[RegularExpression("X4[27]|X8[0-9]", ErrorMessage="Should be like RegEx ....")]
public string ProjectCode { get; set; }
```
and cshtml view: 
```
<input asp-for="ProjectCode" class="form-control" />
```
generates html:
```
<input name="ProjectCode" class="form-control valid" id="ProjectCode" type="text" value="" 
    data-val="true" data-val-regex-pattern="X4[27]|X8[0-9]" data-val-regex="ProjectCode" >
```
instead of: 
```
data-val-regex="Should be like RegEx ...."
```
During validation failure the field Name is displayed instead of the ErrorMessage text.
If the ErrorMessage parameter is **NOT** used, then the standard text for data-val-regex is correct.

kirthik:


src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsClientModelValidatorOfTAttribute.cs
```diff
@@ -52,16 +52,15 @@ protected virtual string GetErrorMessage(ModelMetadata modelMetadata)
                 throw new ArgumentNullException(nameof(modelMetadata));
             }
 
-            var displayName = modelMetadata.GetDisplayName();
             if (_stringLocalizer != null &&
                     !string.IsNullOrEmpty(Attribute.ErrorMessage) &&
                     string.IsNullOrEmpty(Attribute.ErrorMessageResourceName) &&
                     Attribute.ErrorMessageResourceType == null)
```
dougbu: Please fix the wonky indentation while you're here.

dougbu:
:shipit:

kirthik:
merged 454ff9f52a29ffdb50d7d987e1c183ad2e75a672

Issue #3379, opened 2015-10-21
Optimize Allocations in retrieving validators


rynowak:
Here's allocations from 3000 requests to an API with a very simple model (3 properties):

![image](https://cloud.githubusercontent.com/assets/1430011/10650772/1df05ad8-7800-11e5-98ec-e1dd7aff6cf0.png)

The total allocations for this sample is about 90mb, so the 3mb of allocations here is about 3% and pretty significant. Consider also that when we have buffer pooling for input formatters, the total allocations for this will be around 70mb.

Notice the multiples here - anything allocated 27000 times is allocated per-property-per-request. The `List<IModelValidator>` that's allocated 54000 times means that it's twice per property. This is bad scaling for a central infrastructure piece.

Suggested improvements - 

Don't sort validators by `IModelValidator.IsRequired`. We should remove this property and have the validator provider put validators in the correct order. This can avoid duplicating the list and doing the sort using Linq.

Have the `ValidationVisitor` cache the `ModelValidatorProviderContext` - this would mean that we'd allocate it once per request instead of N-times (N is the number of nodes we need to validate). We could do something more exotic in the future if needed. Reducing the number of `MVPC` instances would also allow us to reuse lists and arrays that they in-turn allocate.

rynowak:
> Have the ValidationVisitor cache the ModelValidatorProviderContext

Some more evidence here - sampling data from 3000 requests of the same site.
![image](https://cloud.githubusercontent.com/assets/1430011/10775437/89496b70-7cc5-11e5-8cad-6fae89f47192.png)

Creating new lists and using Linq inside the validation code is too expensive.

ajaybhargavb:
f7a211c09531f145329d76fe71d6676d5c4f3685

PR #3481, opened 2015-11-05
Removed use of LINQ in ValidationVisitor

ajaybhargavb:
#3379 

@rynowak

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/ValidationVisitor.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/ValidationVisitor.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Runtime.CompilerServices;
 
 namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
@@ -23,6 +22,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
         private string _key;
         private object _model;
         private ModelMetadata _metadata;
+        private ModelValidatorProviderContext _context;
         private IValidationStrategy _strategy;
 
         private HashSet<object> _currentPath;
@@ -94,7 +94,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
             var state = _modelState.GetValidationState(_key);
             if (state == ModelValidationState.Unvalidated)
             {
-                var validators = GetValidators(_metadata);
+                var validators = GetValidators();
 
                 var count = validators.Count;
                 if (count > 0)
@@ -255,11 +255,23 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
             return isValid;
         }
 
-        private IList<IModelValidator> GetValidators(ModelMetadata metadata)
+        private IList<IModelValidator> GetValidators()
         {
-            var context = new ModelValidatorProviderContext(metadata);
-            _validatorProvider.GetValidators(context);
-            return context.Validators.OrderBy(v => v, ValidatorOrderComparer.Instance).ToList();
+            if (_context == null)
+            {
+                _context = new ModelValidatorProviderContext(_metadata);
+            }
+            else
+            {
+                // Reusing the context so we don't allocate a new context and list
+                // for every property that gets validated.
+                _context.ModelMetadata = _metadata;
+                _context.Validators.Clear();
+            }
+
+            _validatorProvider.GetValidators(_context);
+
+            return _context.Validators;
         }
 
         private void SuppressValidation(string key)
@@ -346,20 +358,6 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
                 _visitor._strategy = _strategy;
 
                 _visitor._currentPath.Remove(_newModel);
-            }
-        }
-
-        // Sorts validators based on whether or not they are 'required'. We want to run
-        // 'required' validators first so that we get the best possible error message.
-        private class ValidatorOrderComparer : IComparer<IModelValidator>
-        {
-            public static readonly ValidatorOrderComparer Instance = new ValidatorOrderComparer();
-
-            public int Compare(IModelValidator x, IModelValidator y)
-            {
-                var xScore = x.IsRequired ? 0 : 1;
-                var yScore = y.IsRequired ? 0 : 1;
-                return xScore.CompareTo(yScore);
             }
         }
     }
--- a/src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsModelValidatorProvider.cs
+++ b/src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsModelValidatorProvider.cs
@@ -1,9 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
 using System.ComponentModel.DataAnnotations;
-using System.Linq;
 using System.Reflection;
 using Microsoft.Extensions.Localization;
 using Microsoft.Extensions.OptionsModel;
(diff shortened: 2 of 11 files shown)


src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultModelValidatorProvider.cs
```diff
@@ -21,7 +21,16 @@ public void GetValidators(ModelValidatorProviderContext context)
                 var validator = context.ValidatorMetadata[i] as IModelValidator;
                 if (validator != null)
                 {
-                    context.Validators.Add(validator);
+                    // Inserts validators based on whether or not they are 'required'. We want to run
+                    // 'required' validators first so that we get the best possible error message.
+                    if (validator.IsRequired)
+                    {
+                        context.Validators.Insert(0, validator);
```
pranavkm: Should this attempt to keep the order of validators relative to the order in which validator providers are listed? i.e.

`Provider1` -> `A (Required), B`
`Provider2` -> `C, D (Required)`

Result: `D, A, B, C`

Is `A, D, B, C` the more appropriate ordering in this case?

ajaybhargavb: I was aware that it will not maintain the order. But I didn't think that maintaining the order was important.

rynowak: I don't think it's important.

> Don't sort validators by IModelValidator.IsRequired. **We should remove this property and have the validator provider put validators in the correct order.** This can avoid duplicating the list and doing the sort using Linq.

Let's just remove `IsRequired`

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsModelValidatorProvider.cs
```diff
@@ -45,7 +45,18 @@ public void GetValidators(ModelValidatorProviderContext context)
 
             foreach (var attribute in context.ValidatorMetadata.OfType<ValidationAttribute>())
```
rynowak: This is `Linq` YO.

I'd expect to see a `using` removed in each of these files :+1:

rynowak:
Please include before/after allocation profiles when you update. Use the bug as a hint of what to include.

rynowak:
> Have the ValidationVisitor cache the ModelValidatorProviderContext - this would mean that we'd allocate it once per request instead of N-times (N is the number of nodes we need to validate). We could do something more exotic in the future if needed. Reducing the number of MVPC instances would also allow us to reuse lists and arrays that they in-turn allocate.

I don't see this part addressed here...

ajaybhargavb:
:up: :date: 

Allocation Profile:

Before
image



After
image

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/ValidationVisitor.cs
```diff
@@ -255,11 +255,21 @@ private bool VisitChildren(IValidationStrategy strategy)
             return isValid;
         }
 
-        private IList<IModelValidator> GetValidators(ModelMetadata metadata)
+        private IList<IModelValidator> GetValidators()
         {
-            var context = new ModelValidatorProviderContext(metadata);
-            _validatorProvider.GetValidators(context);
-            return context.Validators.OrderBy(v => v, ValidatorOrderComparer.Instance).ToList();
+            if (_context == null)
```
rynowak: Add a comment here that we are reusing this perfs

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsModelValidatorProvider.cs
```diff
@@ -43,9 +41,26 @@ public void GetValidators(ModelValidatorProviderContext context)
                     _stringLocalizerFactory);
             }
 
-            foreach (var attribute in context.ValidatorMetadata.OfType<ValidationAttribute>())
+            foreach (var item in context.ValidatorMetadata)
```
rynowak: Is this an `IList`? Use a `for` not a `foreach`

src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsModelValidatorProvider.cs
```diff
@@ -43,9 +41,26 @@ public void GetValidators(ModelValidatorProviderContext context)
...
+                // Inserts validators based on whether or not they are 'required'. We want to run
+                // 'required' validators first so that we get the best possible error message.
+                if (attribute is RequiredAttribute)
+                {
+                    context.Validators.Insert(0, validator);
+                }
+                else
+                {
+                    context.Validators.Add(validator);
+                }
```
rynowak: Lets have a unit test for this new ordering behavior

rynowak:
:watch:

ajaybhargavb:
:up: :date:

test/Microsoft.AspNet.Mvc.DataAnnotations.Test/DataAnnotationsModelValidatorProviderTest.cs
```diff
@@ -39,6 +39,29 @@ public void GetValidators_ReturnsValidatorForIValidatableObject()
...
+                typeof(ClassWithProperty),
+                "PropertyWithMultipleValidationAttributes");
+
+            var providerContext = new ModelValidatorProviderContext(metadata);
+
+            // Act
+            provider.GetValidators(providerContext);
+
+            // Assert
+            Assert.Equal(4, providerContext.Validators.Count);
```
pranavkm:
```C#
Assert.Collection(providerContext.Validators,
  validator => 
  {
      var dataAnnotationsValidator = Assert.IsType<DataAnnotationsModelValidator>(validator);
      Assert.IsType<CustomRequiredAttribute2>(dataAnnotationsValidator);
  },
 {
      var dataAnnotationsValidator = Assert.IsType<DataAnnotationsModelValidator>(validator);
      Assert.IsType<CustomRequiredAttribute1>(dataAnnotationsValidator);
  },
  ...
  );
```

Verify this works on Mono since this depends on reflection order so internal ordering might be platform dependent.

ajaybhargavb: You are right. Travis failed because of this.

ajaybhargavb: Updated this test to check if it is assignable from type `RequiredAttribute`. No need to check for specific types as we do not care about the exact order. Just need to verify if the required attributes show up first.

ajaybhargavb:
:up: :date:

test/Microsoft.AspNet.Mvc.DataAnnotations.Test/DataAnnotationsModelValidatorProviderTest.cs
```diff
@@ -39,6 +39,27 @@ public void GetValidators_ReturnsValidatorForIValidatableObject()
...
+            Assert.IsAssignableFrom<RequiredAttribute>((providerContext.Validators[0] as DataAnnotationsModelValidator).Attribute);
+            Assert.IsAssignableFrom<RequiredAttribute>((providerContext.Validators[1] as DataAnnotationsModelValidator).Attribute);
```
rynowak: Get used to using a c-style cast in these kind of cases. `((DataAnnotationsModelValidator)providerContext.Validators[1]).Attribute` instead of  `(providerContext.Validators[1] as DataAnnotationsModelValidator).Attribute`

It's more semantically correct, and fail at the case instead with a mysterious nullref later.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/ValidationVisitor.cs
```diff
@@ -255,11 +255,22 @@ private bool VisitChildren(IValidationStrategy strategy)
...
+            {
+                _context = new ModelValidatorProviderContext(_metadata);
+            }
+            else
+            {
+                // Reusing the context to improve performance.
```
rynowak: You can do better than this comment. You need to mention that without reusing the context we'd allocate a list + context for every model property that we validate.

rynowak:
:shipit:  after fixing small comments

test/Microsoft.AspNet.Mvc.DataAnnotations.Test/DataAnnotationsModelValidatorProviderTest.cs
```diff
@@ -39,6 +39,27 @@ public void GetValidators_ReturnsValidatorForIValidatableObject()
...
+                "PropertyWithMultipleValidationAttributes");
+
+            var providerContext = new ModelValidatorProviderContext(metadata);
+
+            // Act
+            provider.GetValidators(providerContext);
+
+            // Assert
+            Assert.Equal(4, providerContext.Validators.Count);
+            Assert.IsAssignableFrom<RequiredAttribute>(((DataAnnotationsModelValidator) providerContext.Validators[0]).Attribute);
```
pranavkm: Spacing

pranavkm:
:shipit:

PR #3490, opened 2015-11-06
Make Cors filters run before any other authorization filters

kichalla:
@rynowak

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Cors/CorsAuthorizationFilter.cs
+++ b/src/Microsoft.AspNet.Mvc.Cors/CorsAuthorizationFilter.cs
@@ -41,7 +41,9 @@ namespace Microsoft.AspNet.Mvc.Cors
         {
             get
             {
-                return int.MaxValue - 100;
+                // Since clients' preflight requests would not have data to authenticate requests, this
+                // filter must run before any other authorization filters.
+                return int.MinValue + 100;
             }
         }
 
--- a/test/WebSites/CorsWebSite/Startup.cs
+++ b/test/WebSites/CorsWebSite/Startup.cs
@@ -47,6 +47,26 @@ namespace CorsWebSite
                                .WithMethods("PUT", "POST")
                                .WithExposedHeaders("exposed1", "exposed2");
                     });
+
+                options.AddPolicy(
+                    "AllowAll",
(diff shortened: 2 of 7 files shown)


src/Microsoft.AspNet.Mvc.Cors/CorsAuthorizationFilter.cs
```diff
@@ -41,7 +41,9 @@ public int Order
         {
             get
             {
-                return int.MaxValue - 100;
+                // Since clients' preflight requests would not have data to authenticate requests, this
+                // filter must run before any other authorization filters.
+                return int.MinValue + 100;
```
pranavkm: Isn't this going to overflow?

kichalla: Its `min` (-2,147,483,648) and not `max`

davidfowl:
There's no better way to implement this? What about a resource filter?

kichalla:
@davidfowl Authorization filters run before the resource filters

rynowak:
> There's no better way to implement this? 

This is what order is for bro. Both of these are policies that tell you if you're allowed to access the action.

davidfowl:
@rynowak then what about using an order relative to the thing you're trying to go before?

test/WebSites/CorsWebSite/Startup.cs
```diff
@@ -47,6 +47,26 @@ public void ConfigureServices(IServiceCollection services)
                                .WithMethods("PUT", "POST")
                                .WithExposedHeaders("exposed1", "exposed2");
                     });
+
+                options.AddPolicy(
+                    "AllowAll",
+                    builder =>
+                    {
+                        builder.AllowCredentials()
```
pranavkm: Use VS default formatting.

src/Microsoft.AspNet.Mvc.Cors/CorsAuthorizationFilter.cs
pranavkm: :+1:

kichalla:
ping @rynowak @pranavkm

pranavkm:
:shipit:

PR #3492, opened 2015-11-06
React to aspnet/Razor#561.

NTaylorMullen:
- `TagHelper` directives no longer have quotes.
- Regenerated test output.


NTaylorMullen:
Just an FYI.

rynowak:
:heart_eyes: does this mean we have intellisense for these now?

NTaylorMullen:
lol @rynowak nopppe, It'd be incredibly difficult to add IntelliSense for these bits.

PR #3496, opened 2015-11-09
Implement DefaultEditorTemplates and DefaultDisplayTemplates as service

genusP:
I write mvc library for change generated html.
I need to change the html generated for the model, analog IHtmlGenerator


dougbu:
@genusP what is the exact use case for this PR? Could you please file an issue providing additional details, perhaps a sample of what you're attempting? We usually start from issues and move to PRs after discussion.

Separately, if we go forward with this approach, please have a look at the [Engineering Guidelines](https://github.com/aspnet/Home/wiki/Engineering-guidelines), the style of the surrounding code, and the failing tests.

Eilon:
@genusP yeah please open a discussion for this. I'm not sure this is a design we would go with because it limits the available templates to generate. I think that if we do something like this it would be more general purpose and allow templates for any type to be created.

Eilon:
Closing this because this isn't a design we want to go with.

PR #3513, opened 2015-11-10
Learn the one wierd trick that really works to shed pounds of allocations

rynowak:
Doctors **hate** him

This textwriter needs to inherit HtmlTextWriter and the
StringCollectionTextWriter needs to have the right conditional test.

This allows us to 'pass-through' any IHtmlContent instances without
writing out intermediate strings.

Here's some allocation data from 3000 requests to a simple page doing model binding and using TagHelpers to render an HTML form. 

**Before**
image

**After**
image

This change is worth about 10mb based on current numbers, and isn't a design change, it was already supposed to do this :laughing:

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor/TagHelperContentWrapperTextWriter.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/TagHelperContentWrapperTextWriter.cs
@@ -2,16 +2,17 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.IO;
 using System.Text;
+using Microsoft.AspNet.Html.Abstractions;
+using Microsoft.AspNet.Mvc.ViewFeatures;
 using Microsoft.AspNet.Razor.TagHelpers;
 
 namespace Microsoft.AspNet.Mvc.Razor
 {
     /// <summary>
-    /// <see cref="TextWriter"/> implementation which writes to a <see cref="TagHelperContent"/> instance.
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Mvc.Razor/TagHelperContentWrapperTextWriter.cs
```diff
@@ -11,7 +12,7 @@ namespace Microsoft.AspNet.Mvc.Razor
     /// <summary>
     /// <see cref="TextWriter"/> implementation which writes to a <see cref="TagHelperContent"/> instance.
```
NTaylorMullen: Need to update doc comment.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/StringCollectionTextWriter.cs
```diff
@@ -183,10 +183,10 @@ public override Task WriteLineAsync()
         /// <param name="encoder">The <see cref="HtmlEncoder"/> to encode the copied/written content.</param>
         public void CopyTo(TextWriter writer, HtmlEncoder encoder)
```
NTaylorMullen: Need to update doc comment.

NTaylorMullen:
:watch: on doc comments since they're failing the travis/appveyor builds.

kichalla:
Nice!

rynowak:
Updated YO

NTaylorMullen:
:shipit:

PR #3525, opened 2015-11-11
Add `launchSettings.json` to `.gitignore`.

NTaylorMullen:
With new VS a `launchSettings.json` is added to every project. Assuming this is an intentional change we should be ignoring it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,4 @@ nuget.exe
 *.sln.ide
 node_modules
 **/[Cc]ompiler/[Rr]esources/**/*.js
+*launchSettings.json


rynowak:
Shouldn't this go into every repo? Also, this is going to be irritating

NTaylorMullen:
Yup, should def go to every repo. Thought I'd try it out here first.

pranavkm:
@rynowak we could update individual repos as and when we touch them.

pranavkm:
:shipit:

rynowak:
> @rynowak we could update individual repos as and when we touch them

That often ends up being annoying. Isn't there a way to update universe and push this everywhere?

barrytang:
Wait. This is a bug. @BillHiebert fixed it already and will show up in next tooling build. Don't change anything on your side.

BillHiebert:
Correct. I checked in a fix this afternoon.

NTaylorMullen:
> Assuming this is an intentional change we should be ignoring it.

:smile: I'll back out the change once we get upgraded to new tooling.

Issue #3527, opened 2015-11-11
Localization of CompareAttribute should be localized


Anderman:
The 
`_stringLocalizer[Attribute.ErrorMessage,displayName,
                                     GetOtherPropertyDisplayName(context)`
Is missing
remarks
* To be consistent: Displayname and OtherPropertyName should also be localized 

setup test
```
        [Compare(nameof(Password))]
        [Display(Name = "Email")]
        public string Email { get; set; }

        [Compare(nameof(Password), ErrorMessage = "{0} en {1} komen niet overeen(NL)")]
        [Display(Name = "PwPassword", ResourceType = typeof(loc))]
```

NTaylorMullen:
This overlaps a bit with https://github.com/aspnet/Mvc/issues/3518

Anderman:
Yeah, I totally agree that display name should be localized by the new _stringlocalizer

Eilon:
Looks like this was fixed along with all the other attributes, here's the current code: https://github.com/aspnet/Mvc/blob/3393ba43c2028af7048d45178c56802adfe5a80f/src/Microsoft.AspNet.Mvc.DataAnnotations/CompareAttributeAdapter.cs#L77-L84

PR #3546, opened 2015-11-12
Commit Localization MaxLengthAttribute and compareAttribute

Anderman:
I made a fix for aspnet-5 localization for ErrorMEssage in the MaxLengthAttribute and compareAttribute.

See issue https://github.com/aspnet/Mvc/issues/3527 and all other related issues

### DIFF
--- a/src/Microsoft.AspNet.Mvc.DataAnnotations/CompareAttributeAdapter.cs
+++ b/src/Microsoft.AspNet.Mvc.DataAnnotations/CompareAttributeAdapter.cs
@@ -12,7 +12,7 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
     public class CompareAttributeAdapter : DataAnnotationsClientModelValidator<CompareAttribute>
     {
         public CompareAttributeAdapter(CompareAttribute attribute, IStringLocalizer stringLocalizer)
-            : base(new CompareAttributeWrapper(attribute), stringLocalizer)
+            : base(new CompareAttributeWrapper(attribute, stringLocalizer), stringLocalizer)
         {
             if (attribute == null)
             {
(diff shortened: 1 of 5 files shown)


Anderman:
@dougbu Is this the right way todo this?. I can make a fix for other ErrorMessage localization

dougbu:
@Anderman this looks like a solid improvement over what we have today. Main question is about the higher-level design i.e. `params` versus adding the following or some other approach.
``` c#
protected virtual LocalizedString GetLocalizedErrorMessage(string displayName)
{
    return _stringLocalizer[Attribute.ErrorMessage, displayName];
}   
```

@kirthik what fits best with how you've done localization elsewhere?

Separately, should fix `DataAnnotationsModelValidator`s at the same time as the `DataAnnotationsClientModelValidator<TAttribute>`s.

Anderman:
@kirthik the `[EmailAddress]` attribute set Attribute.ErrorMessage. The code use this to chose between    the `_stringLocalizer[]` and the default `Attribute.FormatErrormessage()`

A workaround could be something like this
```
CustomErrorMessageSet = attribute.ErrorMessage != (new DataTypeAttribute(attribute.DataType)).ErrorMessage;
```

Any suggestions?

Anderman:
I found the problem. It is in `System.ComponentModel.DataAnnotations/DataAnnotations/EmailAddressAttribute.cs`

This line was set 28 days ago. But it should not be set at all 
`DefaultErrorMessage = DataAnnotationsResources.EmailAddressAttribute_Invalid;`

kirthik:
@Anderman : You implemented one way of solving this issue. @dougbu suggested another way. I am working on a prototype to address this issue. I will let you know once we have a finalized design.

Anderman:
Oke fine, I created just a issue about the [emailadressAttribute](https://github.com/dotnet/corefx/issues/4481)

Issue #3334, opened 2015-10-16
Consider NoContent action result method on Controller


PradeepKadubandi:
This came up while I and Ryan were briefly discussing code to be generated for Web API scaffolding, it may be handy to define NoContent action result method on Controller class.

danilodeven:
This is already implemented @PradeepKadubandi
```c#
    public IActionResult index() {
        return new NoContentResult();
    }
```
Is this what you are looking for?

Eilon:
I would imagine this is about having a helper factory method on the `Controller` base class, just like there is for many other result types.

ryanbrandenburg:
Fixed in #3547

PR #3547, opened 2015-11-12
Add NoContent factory method to Controller

ryanbrandenburg:
As per #3334, adding a NoContent factory method to Controller.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
@@ -508,6 +508,17 @@ namespace Microsoft.AspNet.Mvc
             return result;
         }
 
+
+        /// <summary>
+        /// Creates a <see cref="NoContentResult"/> object that produces an empty No Content (204) response.
+        /// </summary>
+        /// <returns>The created <see cref="NoContentResult"/> object for the response.</returns>
+        [NonAction]
+        public virtual NoContentResult NoContent()
+        {
+            return new NoContentResult();
+        }
+
         /// <summary>
         /// Creates a <see cref="JsonResult"/> object that serializes the specified <paramref name="data"/> object
         /// to JSON.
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ControllerTest.cs
```diff
@@ -1056,6 +1056,20 @@ public void Controller_View_WithParameterViewNameAndViewModel_SetsResultViewName
         }
 
         [Fact]
+        public void Controller_NoContent()
+        {
+            // Arrange
+            var controller = new TestableController();
+
+            // Act
+            var actualContentResult = controller.NoContent();
```
rynowak: I don't like this name.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ControllerTest.cs
```diff
@@ -1056,6 +1056,20 @@ public void Controller_View_WithParameterViewNameAndViewModel_SetsResultViewName
...
+
+            // Assert
+            Assert.IsType<NoContentResult>(actualContentResult);
```
rynowak: This isn't really needed. The compiler already did this.

rynowak:
:shipit:

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
```diff
@@ -508,6 +508,12 @@ public virtual ContentResult Content(string content, MediaTypeHeaderValue conten
             return result;
         }
 
+        [NonAction]
+        public virtual NoContentResult NoContent()
```
pranavkm: This needs docs.

Eilon: Yeah feel free to steal from MSDN from Web API or MVC or wherever we might have had a similar API before.

ryanbrandenburg:
:up: :date:

pranavkm:
:shipit:

Issue #3502, opened 2015-11-09
ViewResultExecutor logs partial view messages for regular views


pranavkm:
These should be about finding a view, not a partial:

* https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/ViewResultExecutor.cs#L88
* https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/ViewResultExecutor.cs#L105

ryanbrandenburg:
Fixed in #3548

PR #3548, opened 2015-11-12
Log correct message for Found/NotFound views

ryanbrandenburg:
As per #3502 making sure we use the right messages for when we do or don't find a view.


rynowak:
:shipit:

src/Microsoft.AspNet.Mvc.ViewFeatures/Logging/ViewResultExecutorLoggerExtensions.cs
```diff
@@ -10,6 +11,8 @@ namespace Microsoft.AspNet.Mvc.Logging
     public static class ViewResultExecutorLoggerExtensions
     {
         private static Action<ILogger, string, Exception> _viewResultExecuting;
+        private static Action<ILogger, string, Exception> _viewFound;
```
pranavkm: `readonly`

pranavkm:
:shipit:

PR #3559, opened 2015-11-14
Removes unnecessary null check

### LINKED ISSUE #3558: Potential invalid operation exception?
Just scanning some code out of curiosity when I spotted [this](https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/FileResult.cs#L74).

If you're specifying it's a required service then line 74 doesn't need the `?.`. 
Otherwise, if that was changed to `GetService` then you could have a null reference exception [here](https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/FileResult.cs#L91).

herecydev:
Resolves #3558

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/FileResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/FileResult.cs
@@ -71,7 +71,7 @@ namespace Microsoft.AspNet.Mvc
             }
 
             var loggerFactory = context.HttpContext.RequestServices.GetRequiredService<ILoggerFactory>();
-            var logger = loggerFactory?.CreateLogger<FileResult>();
+            var logger = loggerFactory.CreateLogger<FileResult>();
 
             var response = context.HttpContext.Response;
             response.ContentType = ContentType.ToString();
--- a/test/Microsoft.AspNet.Mvc.Core.Test/FileResultTest.cs
+++ b/test/Microsoft.AspNet.Mvc.Core.Test/FileResultTest.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.IO;
 using System.Net.Mime;
 using System.Threading.Tasks;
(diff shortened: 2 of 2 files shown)


src/Microsoft.AspNet.Mvc.Core/FileResult.cs
```diff
@@ -88,7 +88,7 @@ public override Task ExecuteResultAsync(ActionContext context)
                 context.HttpContext.Response.Headers[HeaderNames.ContentDisposition] = contentDisposition.ToString();
             }
             
-            logger.FileResultExecuting(FileDownloadName);
+            logger?.FileResultExecuting(FileDownloadName);
```
dougbu: If `CreateLogger()` can return `null`, many classes in the MVC repo will fall over. So I hope this isn't for a real scenario.

herecydev: Well it's up to you, my thoughts were if you allow the client to bind up an ILoggerFactory then this could happen. We could always extract the checks into some generic util class that can return either the ILogger<T> or throw an obvious error message about an ILoggerFactory returning a null logger.

davidfowl: Remove the `?.` CreateLogger should never return null.

herecydev:
At work at the moment, will apply changes later for you.

herecydev:
@davidfowl @Eilon Should be cleaned up now. I've tried to follow the current test style also.

test/Microsoft.AspNet.Mvc.Core.Test/FileResultTest.cs
```diff
@@ -120,6 +121,42 @@ public void Constructor_SetsContentType()
...
+            httpContext.RequestServices = services.BuildServiceProvider();
+
+            var actionContext = CreateActionContext(httpContext);
+            var result = new EmptyFileResult("application/my-type");
+
+            // Act
+            await result.ExecuteResultAsync(actionContext);
+
+            // Assert
+            Assert.True(loggerSink.Writes.Count == 1);
```
Eilon: Use `Assert.Equals(expected, actual);`

test/Microsoft.AspNet.Mvc.Core.Test/FileResultTest.cs
```diff
@@ -120,6 +121,42 @@ public void Constructor_SetsContentType()
...
+        public void ExecuteResultAsync_ThrowsException_IfCannotResolveLoggerFactory()
+        {
+            // Arrange
+            var httpContext = new DefaultHttpContext();
+            httpContext.RequestServices = new ServiceCollection().BuildServiceProvider();
+            var actionContext = CreateActionContext(httpContext);
+            var result = new EmptyFileResult("application/my-type");
+
+            // Act
+            Action act = () => result.ExecuteResultAsync(actionContext).Wait();
```
Eilon: Can do this inline in the `Assert.Throws<...>()`. Kind of weird to define an Action for this. Then just change to comment to `// Act & Assert`.

herecydev: Hopefully that's better? Apologies, I'm not used to using xunit's assertions.

Eilon:
Looks much nicer. @pranavkm ?

Eilon:
@herecydev BTW you'll want to squash the commits down to one single commit as well.

pranavkm:
I'll merge it in.

herecydev:
@Eilon Should be merged into a single commit now.

pranavkm:
Merged in https://github.com/aspnet/Mvc/commit/db94a8ed19e626b4e094cc283abe0dfe4f39871f

Issue #3550, opened 2015-11-12
Pool JsonSerializer instances


rynowak:
See data here:

![image](https://cloud.githubusercontent.com/assets/1430011/11124519/514e15ca-891a-11e5-82e6-88a4db8a59a9.png)

See comment from @JamesNK here: https://github.com/aspnet/Mvc/issues/3542#issuecomment-156200726

rynowak:
Partial fix here: e75eebbed000541f05dda8e9f524b445ba0722ff for output formatters. 

Fixing input will take some doing.

JamesNK:
Keep in mine that JsonSerializerSettings is mutable. Changing configuration on it without updating the JsonSerializer could confuse some users.

rynowak:
Load balancing this to a Pranav

pranavkm:
Merged in cf9221df073467f0e6bae910f65a6957cd29523e

PR #3564, opened 2015-11-16
Pool JsonSerializer for Output formatting

rynowak:
Partial fix for #3550


rynowak:
The fix for `JsonInputFormatter` isn't so simple. Will follow up on that separately.

kichalla:
:shipit:

PR #3567, opened 2015-11-16
[Fixes #3503] Removing formatters on a given type only works on collections in MVC options

### LINKED ISSUE #3503: Removing formatters on a given type only works on collections in MVC options
Currently a formatter can be removed by using the [`RemoveType`](https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/Formatters/FormatterCollection.cs#L19) method on the formatters collection on MVC options 

Example: `options.OutputFormatters.RemoveType<StringOutputFormatter>()`

We have collection of formatters exposed through other areas of the framework like [`ResourceExecutingContext`](https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Abstractions/Filters/ResourceExecutingContext.cs#L35) where this method cannot be used. An example scenario here is a user trying to have a per-controller specific list of formatters, which he can achieve using  a resource filter

Should we have the formatter collections on MVC options to be just `List<T>` and have an extension method on `IList<T>` which can achieve this scenario? I know this is very generic and need not apply to formatters only

Example:
```
public static class MyExtensions
{
    public static void RemoveType<T>(this IList<T> list, Type itemType)
    {
        var itemTypeInfo = itemType.GetTypeInfo();
        for (var i = list.Count - 1; i >= 0; i--)
        {
            var item = list[i];
            if (itemTypeInfo.IsAssignableFrom(item.GetType().GetTypeInfo()))
            {
                list.RemoveAt(i);
            }
        }
    }
}
```

One observation...`FormatterCollection` is under `Mvc.Formatters` namespace which makes discovering `RemoveType<>` hard

kichalla:
@rynowak

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Abstractions/Formatters/FormatterCollection.cs
+++ b/src/Microsoft.AspNet.Mvc.Abstractions/Formatters/FormatterCollection.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Collections.Generic;
 using System.Collections.ObjectModel;
 
 namespace Microsoft.AspNet.Mvc.Formatters
@@ -11,6 +12,23 @@ namespace Microsoft.AspNet.Mvc.Formatters
     /// <typeparam name="TFormatter">The type of formatters in the collection.</typeparam>
     public class FormatterCollection<TFormatter> : Collection<TFormatter>
     {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="FormatterCollection{TFormatter}"/> class that is empty.
+        /// </summary>
+        public FormatterCollection()
+            : base()
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="FormatterCollection{TFormatter}"/> class
+        /// as a wrapper for the specified list.
+        /// </summary>
+        /// <param name="list">The list that is wrapped by the new collection.</param>
+        public FormatterCollection(IList<TFormatter> list)
+            : base(list)
+        {
+        }
 
         /// <summary>
         /// Removes all formatters of the specified type.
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNet.Mvc.Abstractions/Formatters/FormatterCollection.cs
```diff
@@ -11,6 +12,23 @@ namespace Microsoft.AspNet.Mvc.Formatters
     /// <typeparam name="TFormatter">The type of formatters in the collection.</typeparam>
     public class FormatterCollection<TFormatter> : Collection<TFormatter>
```
kichalla: Just FYI...I moved this type from Mvc.Core to Mvc.Abstractions

test/WebSites/FiltersWebSite/Controllers/SpecificFormattersController.cs
```diff
@@ -0,0 +1,72 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNet.Mvc;
+using Microsoft.AspNet.Mvc.Filters;
+using Microsoft.AspNet.Mvc.Formatters;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
+
+namespace FiltersWebSite
```
rynowak: Can you put this in `BasicWebSite` instead? We might mothball this site. This is a reasonable test, and I don't want it to get lost in the scrum.

test/WebSites/FiltersWebSite/Controllers/SpecificFormattersController.cs
```diff
@@ -0,0 +1,72 @@
...
+            if (!ModelState.IsValid)
+            {
+                return HttpBadRequest(ModelState);
+            }
+
+            return Ok(person);
+        }
+
+        [HttpPost("noncamelcase")]
+        public IActionResult Post2([FromBody] Person person)
```
rynowak: Better names for these?

src/Microsoft.AspNet.Mvc.Abstractions/Formatters/FormatterCollection.cs
rynowak: Seems alright.

rynowak:
:watch:

test/WebSites/FiltersWebSite/Controllers/SpecificFormattersController.cs
kichalla: Sure will do

kichalla:
:up: :date:

rynowak:
:shipit:

kichalla:
982c7abea8fc540142dc6566353508f42c4dc352

PR #3568, opened 2015-11-16
Limit [FromServices] to apply only to parameters

### LINKED ISSUE #3507: Limit [FromServices] to apply only to parameters
This should help reduce a lot of the confusion around this feature.

It currently applies to parameters *and* properties. The problem with properties is that it appears to many people that it can be applied to any property of any object. But in reality it works only with properties on objects that are model bound, which is controller properties, and sub-properties of types used in MVC action method parameters.

By limiting `[FromServices]` to only parameters, that will restrict the usage to one that will always work: top-level action method parameters.

cc @rynowak

pranavkm:
Fixes #3507


Eilon:
:shipit:

src/Microsoft.AspNet.Mvc.Core/FromServicesAttribute.cs
```diff
@@ -7,25 +7,9 @@
...
-    /// {
-    ///     [FromServices]
-    ///     public ILocationService LocationService { get; set; }
-    ///
-    ///     public void Update()
-    ///     {
-    ///         LocationService.Update();
-    ///     }
-    /// }
-    /// </code>
```
rynowak: Why kill the example?

test/Microsoft.AspNet.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
```diff
@@ -268,9 +268,6 @@ private class Order2
         private class Person2
         {
             public string Name { get; set; }
-
-            [FromServices]
-            public IActionBindingContextAccessor BindingContext { get; set; }
         }
 
         [Fact]
```
rynowak: This test isn't valid any more. This should be updated to use `[FromBody]` or something similar. If `[FromBody]` is already covered then this is good enough.

test/Microsoft.AspNet.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
```diff
@@ -1489,9 +1484,6 @@ private class Person9
         {
             [FromBody]
             public Address1 Address { get; set; }
-
-            [FromServices]
-            public IActionBindingContextAccessor BindingContext { get; set; }
```
rynowak: I'm fine with removing this from this scenario.

test/WebSites/ModelBindingWebSite/Controllers/VehicleController.cs
```diff
@@ -59,7 +59,6 @@ public IActionResult UpdateDealerVehicle(VehicleWithDealerViewModel model)
                 return PartialView("UpdateVehicle", model);
             }
 
-            model.Update();
```
rynowak: ???

pranavkm: The code basically did nothing and sort of tested that services could be property injected.

src/Microsoft.AspNet.Mvc.Core/FromServicesAttribute.cs
pranavkm: Property injection doesn't work any more. The other example talks about parameter injection and I've left that alone.

rynowak:
:shipit:

Issue #3152, opened 2015-09-17
IViewComponentHelper should return IHtmlContent not HtmlString


rynowak:
This API needs an update since we added IHtmlContent.

There's a few other things we should review here as well as part of API reviews.

rynowak:
The same is probably true of `IJsonHelper`

Eilon:
Slightly related to https://github.com/aspnet/Mvc/issues/3123.

dougbu:
b3c68de

PR #3570, opened 2015-11-17
Change `IViewComponentHelper` methods to return `IHtmlContent`

dougbu:
- #3152


dougbu:
/cc @rynowak 

No tests had to change. Any tests I should add?

rynowak:
I don't think so. We've got good functional tests in this area, and if something was broken, it would be broken at an integration point. :shipit:

dougbu:
b3c68de

PR #3581, opened 2015-11-18
[Design] Replace ActionBindingContext with ControllerContext

rynowak:
This change is still a work in progress, but is step one of dropping the use of CallContext from our code.

The general idea is to create a combined `ActionContext` + `ActionBindingContext` that matches what Controllers need to have access to. So once you enter the invoker, you get 'upgraded' to a `ControllerContext`.

Extensibility points that specifically deal with controllers (ControllerFactory, ArgumentBinding) are updated to use `ControllerContext`. Extensibility points that are more general (ViewEngine, ModelBinding) still use `ActionContext`.


src/Microsoft.AspNet.Mvc.Abstractions/ControllerContext.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+        }
+
+        public IList<IModelBinder> ModelBinders { get; set; }
+
+        public IList<IValueProvider> ValueProviders { get; set; }
+
+        public IList<IInputFormatter> InputFormatters { get; set; }
+
+        public IList<IModelValidatorProvider> ValidatorProviders { get; set; }
```
rynowak: `ActionBindingContext` represented all of these as composites. I think that's kinda dumb, so I undid it.

This had an impact on a lot of the `TryUpdateModel` tests, which were only valid assuming that some mock they created was the top-level model binder. Mostly I just fixed those to do what they did before, but without that assumption.

src/Microsoft.AspNet.Mvc.Abstractions/Filters/ResourceExecutingContext.cs
```diff
@@ -29,12 +29,6 @@ public ResourceExecutingContext(ActionContext actionContext, IList<IFilterMetada
         /// </summary>
         public virtual FormatterCollection<IInputFormatter> InputFormatters { get; set; }
 
-        /// <summary>
-        /// Gets or sets the list of <see cref="IOutputFormatter"/> instances used to format the response.
-        /// </summary>
-        public virtual FormatterCollection<IOutputFormatter> OutputFormatters { get; set; }
```
rynowak: As discussed, removing this from ResourceFilters. All these scenarios still work with result filters.

src/Microsoft.AspNet.Mvc.Abstractions/ControllerContext.cs
```diff
@@ -0,0 +1,31 @@
...
+        public ControllerContext(ActionContext context)
+            : base(context)
+        {
+        }
+
+        public IList<IModelBinder> ModelBinders { get; set; }
+
+        public IList<IValueProvider> ValueProviders { get; set; }
+
+        public IList<IInputFormatter> InputFormatters { get; set; }
```
rynowak: This should be `FormatterCollection`

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/IValueProviderFactory.cs
```diff
@@ -10,10 +10,10 @@ public interface IValueProviderFactory
         /// <summary>
         /// Gets a <see cref="IValueProvider"/> with values from the current request.
         /// </summary>
-        /// <param name="context">The <see cref="ValueProviderFactoryContext"/>.</param>
+        /// <param name="context">The <see cref="ActionContext"/>.</param>
         /// <returns>
         /// A <see cref="Task"/> that when completed will yield a <see cref="IValueProvider"/> instance or <c>null</c>.
         /// </returns>
-        Task<IValueProvider> GetValueProviderAsync(ValueProviderFactoryContext context);
+        Task<IValueProvider> GetValueProviderAsync(ActionContext context);
```
rynowak: This was a holdover from previous iterations where modelbinding was a separate package.

src/Microsoft.AspNet.Mvc.Core/Controllers/ControllerActionInvoker.cs
```diff
@@ -30,12 +29,10 @@ public class ControllerActionInvoker : FilterActionInvoker
             IControllerFactory controllerFactory,
             ControllerActionDescriptor descriptor,
             IReadOnlyList<IInputFormatter> inputFormatters,
-            IReadOnlyList<IOutputFormatter> outputFormatters,
             IControllerActionArgumentBinder controllerActionArgumentBinder,
             IReadOnlyList<IModelBinder> modelBinders,
             IReadOnlyList<IModelValidatorProvider> modelValidatorProviders,
             IReadOnlyList<IValueProviderFactory> valueProviderFactories,
-            IActionBindingContextAccessor actionBindingContextAccessor,
```
rynowak: :wave:

src/Microsoft.AspNet.Mvc.Abstractions/ControllerContext.cs
```diff
@@ -0,0 +1,31 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Microsoft.AspNet.Mvc.Formatters;
+using Microsoft.AspNet.Mvc.ModelBinding;
+using Microsoft.AspNet.Mvc.ModelBinding.Validation;
+
+namespace Microsoft.AspNet.Mvc
+{
+    public class ControllerContext : ActionContext
```
rynowak: Two thoughts.

I considered adding a `new` property here for `ControllerActionDescriptor` so you don't need to cast if you want to see things like the type and methodinfo.

Doing that would pull this out of abstractions, and into core. Which seems like a good idea anyway. It's a detail of how controller + its invoker work.

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
```diff
@@ -102,21 +102,24 @@ public ModelStateDictionary ModelState
...
         /// <remarks>
         /// <see cref="Controllers.IControllerActivator"/> activates this property while activating controllers.
         ///  If user code directly instantiates a controller, the getter returns an empty
-        /// <see cref="Mvc.ActionContext"/>.
+        /// <see cref="Mvc.ControllerContext"/>.
         /// </remarks>
-        [ActionContext]
-        public ActionContext ActionContext
+        [ControllerContext]
+        public ControllerContext ControllerContext
```
rynowak: The main impact of this change is to unit tests, or where someone wrote unnecessarily verbose code like `ActionContext.HttpContext`.

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
```diff
@@ -27,7 +27,7 @@ namespace Microsoft.AspNet.Mvc
     /// </summary>
     public abstract class Controller : IActionFilter, IAsyncActionFilter, IDisposable
```
rynowak: /cc @Eilon

src/Microsoft.AspNet.Mvc.WebApiCompatShim/ApiController.cs
```diff
@@ -31,15 +31,14 @@ public abstract class ApiController : IDisposable
         /// Gets the action context.
         /// </summary>
         /// <remarks>The setter is intended for unit testing purposes only.</remarks>
-        [ActionContext]
-        public ActionContext ActionContext { get; set; }
+        public ActionContext ActionContext => ControllerContext;
```
rynowak: I need to check this for WebAPI compat.

src/Microsoft.AspNet.Mvc.Core/Controllers/DefaultControllerPropertyActivator.cs
```diff
@@ -12,58 +12,43 @@ namespace Microsoft.AspNet.Mvc.Controllers
...
             }
         }
 
-        private PropertyActivator<Contexts>[] GetPropertiesToActivate(Type type)
+        private PropertyActivator<ControllerContext>[] GetPropertiesToActivate(Type type)
         {
-            IEnumerable<PropertyActivator<Contexts>> activators;
-            activators = PropertyActivator<Contexts>.GetPropertiesToActivate(
+            IEnumerable<PropertyActivator<ControllerContext>> activators;
+            activators = PropertyActivator<ControllerContext>.GetPropertiesToActivate(
```
pranavkm: Super unrelated but could we potentially simplify this code? There's really no good reason for a controller type to have more than one `ControllerContext`.

rynowak: Probably

src/Microsoft.AspNet.Mvc.Abstractions/ControllerContext.cs
pranavkm: That sounds nice.

pranavkm: Docs for the type + properties.

src/Microsoft.AspNet.Mvc.Core/Controllers/DefaultControllerActionArgumentBinder.cs
```diff
@@ -55,13 +49,13 @@ public class DefaultControllerActionArgumentBinder : IControllerActionArgumentBi
                 throw new ArgumentNullException(nameof(controller));
             }
 
-            var actionDescriptor = actionContext.ActionDescriptor as ControllerActionDescriptor;
+            var actionDescriptor = context.ActionDescriptor as ControllerActionDescriptor;
```
pranavkm: Follow up to remove the cast once you `new` this property.

src/Microsoft.AspNet.Mvc.Core/Controllers/DefaultControllerFactory.cs
```diff
@@ -47,20 +47,20 @@ protected IControllerActivator ControllerActivator
...
         {
-            if (actionContext == null)
+            if (context == null)
             {
-                throw new ArgumentNullException(nameof(actionContext));
+                throw new ArgumentNullException(nameof(context));
             }
 
-            var actionDescriptor = actionContext.ActionDescriptor as ControllerActionDescriptor;
+            var actionDescriptor = context.ActionDescriptor as ControllerActionDescriptor;
```
pranavkm: Remove when `context.ActionDescriptor` has the right type.

src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs
```diff
@@ -441,20 +408,20 @@ private Task InvokeAllResourceFiltersAsync()
...
-                        ActionContext.RouteData.Values);
+                    Context.InputFormatters = _resourceExecutingContext.InputFormatters;
+                    Context.ModelBinders = _resourceExecutingContext.ModelBinders;
+                    Context.ValidatorProviders = _resourceExecutingContext.ValidatorProviders;
 
-                    ActionBindingContext.ValueProvider = await CompositeValueProvider.CreateAsync(
-                        _resourceExecutingContext.ValueProviderFactories,
-                        valueProviderFactoryContext);
+                    var valueProviders = new List<IValueProvider>();
+                    foreach (var factory in _resourceExecutingContext.ValueProviderFactories)
```
pranavkm: Could this be replaced with a `for` loop?

src/Microsoft.AspNet.Mvc.Core/Controllers/IControllerActionArgumentBinder.cs
```diff
@@ -16,12 +16,8 @@ public interface IControllerActionArgumentBinder
...
         /// </summary>
-        /// <param name="context">The action context assoicated with the current action.</param>
-        /// <param name="bindingContext">The <see cref="ActionBindingContext"/>.</param>
+        /// <param name="context">The <see cref="ControllerContext"/> associated with the current action.</param>
         /// <param name="controller">The controller object which contains the action.</param>
-        Task<IDictionary<string, object>> BindActionArgumentsAsync(
-            ActionContext context,
-            ActionBindingContext bindingContext,
-            object controller);
+        Task<IDictionary<string, object>> BindActionArgumentsAsync(ControllerContext context, object controller);
```
pranavkm: Could you fix the summary for this method:

> Returns a dictionary of representing the parameter-argument name-value pairs,

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
```diff
@@ -102,21 +102,24 @@ public ModelStateDictionary ModelState
         }
 
         /// <summary>
-        /// Gets or sets the <see cref="Mvc.ActionContext"/> object.
+        /// Gets or sets the <see cref="Mvc.ControllerContext"/>.
         /// </summary>
         /// <remarks>
         /// <see cref="Controllers.IControllerActivator"/> activates this property while activating controllers.
         ///  If user code directly instantiates a controller, the getter returns an empty
```
pranavkm: Unrelated, but this comment has a leading space.

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
```diff
@@ -237,9 +234,7 @@ public ViewDataDictionary ViewData
                 if (_viewData == null)
                 {
                     // This should run only for the controller unit test scenarios
-                    _viewData =
-                        new ViewDataDictionary(new EmptyModelMetadataProvider(),
-                                                ActionContext?.ModelState ?? new ModelStateDictionary());
+                    _viewData = new ViewDataDictionary(new EmptyModelMetadataProvider(), ControllerContext.ModelState);
```
pranavkm: Why did this change?

pranavkm: Can `ControllerContext` be null?

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
```diff
@@ -1407,15 +1402,7 @@ public virtual void OnActionExecuted(ActionExecutedContext context)
...
-            if (BindingContext == null)
-            {
-                var message = Resources.FormatPropertyOfTypeCannotBeNull(
-                    nameof(BindingContext),
-                    typeof(Controller).FullName);
-                throw new InvalidOperationException(message);
-            }
-
-            return TryUpdateModelAsync(model, prefix, BindingContext.ValueProvider);
+            return TryUpdateModelAsync(model, prefix, new CompositeValueProvider(ControllerContext.ValueProviders));
```
pranavkm: Should this repurpose the `BindingContext == null` exceptions for `ControllerContext == null`?

src/Microsoft.AspNet.Mvc.WebApiCompatShim/ApiController.cs
```diff
@@ -31,15 +31,14 @@ public abstract class ApiController : IDisposable
         /// Gets the action context.
         /// </summary>
         /// <remarks>The setter is intended for unit testing purposes only.</remarks>
```
pranavkm: Remove this

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ControllerTest.cs
```diff
@@ -1421,13 +1407,19 @@ public static IEnumerable<object[]> RedirectTestData
...
+                .Returns(true);
+
             var binder = new Mock<IModelBinder>();
-            var valueProvider = Mock.Of<IValueProvider>();
             binder.Setup(b => b.BindModelAsync(It.IsAny<ModelBindingContext>()))
                   .Callback((ModelBindingContext context) =>
                   {
-                      Assert.Equal(prefix, context.ModelName);
-                      Assert.Same(valueProvider, context.ValueProvider);
+                      System.Diagnostics.Debugger.Break();
```
pranavkm: Remove

src/Microsoft.AspNet.Mvc.Core/Controllers/DefaultControllerPropertyActivator.cs
rynowak: Actually the reason not to simplify this is that `[ActionContext]` is still a thing.

src/Microsoft.AspNet.Mvc.Core/Controllers/FilterActionInvoker.cs
rynowak: Yeah, you're right, this should be.

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
Eilon: Check immediately above - it creates a new one if it's null.

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
Eilon: Can't be null.

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
pranavkm: Ok. We should remove all the null-safe accessors in properties in that case

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
```diff
@@ -42,7 +42,7 @@ public IServiceProvider Resolver
         {
             get
             {
-                return ActionContext?.HttpContext?.RequestServices;
+                return ControllerContext?.HttpContext?.RequestServices;
```
pranavkm: `ControllerContext` is guaranteed to be non-null. Remove the null safe accessor for these properties.

src/Microsoft.AspNet.Mvc.Core/Controllers/DefaultControllerActionArgumentBinder.cs
```diff
@@ -229,18 +221,16 @@ private void ActivateProperties(object controller, Type containerType, Dictionar
...
-                ModelBinder = bindingContext.ModelBinder,
-                ValidatorProvider = bindingContext.ValidatorProvider,
+                InputFormatters = context.InputFormatters,
+                ModelBinder = new CompositeModelBinder(context.ModelBinders),
+                ValidatorProvider = new CompositeModelValidatorProvider(context.ValidatorProviders),
                 MetadataProvider = _modelMetadataProvider,
-                HttpContext = actionContext.HttpContext,
-                ValueProvider = bindingContext.ValueProvider,
+                HttpContext = context.HttpContext,
+                ValueProvider = new CompositeValueProvider(context.ValueProviders),
```
Eilon: Not sure if it makes sense, but instead of creating all these composite things does it make more sense to change the data type that's passed around into an `IList<Foo>`? Or does a CompositeFoo ultimately need to be created at some point for each of these anyway.

(This comment applies to several places throughout this change, not just here.)

src/Microsoft.AspNet.Mvc.WebApiCompatShim/ApiController.cs
rynowak: WebAPI provided `ActionContext`, so leaving this.

src/Microsoft.AspNet.Mvc.Core/Controllers/DefaultControllerFactory.cs
rynowak: Removed

src/Microsoft.AspNet.Mvc.Core/Controllers/IControllerActionArgumentBinder.cs
rynowak: Fixed

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
rynowak: :open_mouth:

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
rynowak: That's the nice part of just having one context. The unit testing is really simple now.

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
rynowak: Fixed a bunch of these.

src/Microsoft.AspNet.Mvc.WebApiCompatShim/ApiController.cs
rynowak: Fixed

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ControllerTest.cs
rynowak: :+1:

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
rynowak: Fixed

src/Microsoft.AspNet.Mvc.Core/Controllers/DefaultControllerActionArgumentBinder.cs
rynowak: The composite needs to be created at some point. I chose that boundary to be the static methods on `ModelBindingHelper` and the creation of `OperationBindingContext`. 

If anyone wants to make another suggestion, make your case :laughing:

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ControllerTest.cs
```diff
@@ -1797,34 +1779,41 @@ public void TempData_CanSetAndGetValues()
...
+            var controllerContext = new ControllerContext()
             {
-                ModelBinder = binder,
-                ValueProvider = provider,
+                HttpContext = httpContext,
                 InputFormatters = new List<IInputFormatter>(),
-                ValidatorProvider = new DataAnnotationsModelValidatorProvider(
-                    options: null,
-                    stringLocalizerFactory: null)
+                ModelBinders = new List<IModelBinder>()
```
Eilon: Just test code, but would this work:

```c#
var ct = new CT()
{
    ...
    ModelBinders = new[] { binder },
    ValueProviders = new[] { valueProvider },
};
```

Etc.?

Is the compiler smart enough to figure all the covariant stuff out?

rynowak: The type of the array matches the type of the items inside `{ ... }` if they all match exactly. This would work in the cases you're suggesting. I usually avoid this pattern because it breaks if you add another thing that isn't typed exactly the same. 

It don't believe it tries to generate a 'bottom' type when inferring the value of `new [] { ... }`

OK
```
controller.ControllerContext.ModelBinders = new []{ (IModelBinder)foo, (IModelBinder)bar, };
```

Un-OK
```
controller.ControllerContext.ModelBinders = new []{ (MyModelBinder1)foo, (MyModelBinder2)bar, };
```

Eilon:
The test changes aren't as scary as I would've thought they'd be. Very nice.

src/Microsoft.AspNet.Mvc.Core/Controllers/DefaultControllerActionArgumentBinder.cs
Eilon: Naw, just curiosity. This is fine.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ControllerTest.cs
Eilon: Yeah I was talking about these specific code snippets where there's only one type in the array, and due to `IList<T>` covariance and array implementing `IList<T>` that it might just work. No biggie.

rynowak:
Updated this with responses to feedback so far.

src/Microsoft.AspNet.Mvc.Core/ControllerContext.cs
```diff
@@ -0,0 +1,154 @@
...
+    /// </summary>
+    public class ControllerContext : ActionContext
+    {
+        private FormatterCollection<IInputFormatter> _inputFormatters;
+        private IList<IModelBinder> _modelBinders;
+        private IList<IModelValidatorProvider> _validatorProviders;
+        private IList<IValueProvider> _valueProviders;
+
+        /// <summary>
+        /// Creates a new <see cref="ControllerContext"/>. This constructor should only be used for
```
pranavkm: The unit test only comment should probably go to `remarks`.

src/Microsoft.AspNet.Mvc.Core/ControllerContext.cs
```diff
@@ -0,0 +1,154 @@
...
+            {
+                throw new ArgumentException(Resources.FormatActionDescriptorMustBeBasedOnControllerAction(
+                    typeof(ControllerActionDescriptor)),
+                    nameof(context));
+            }
+        }
+
+        public new ControllerActionDescriptor ActionDescriptor
+        {
+            get { return (ControllerActionDescriptor)base.ActionDescriptor; }
```
pranavkm: Should this account for `ActionContext.ActionDescriptor` being independently settable?

src/Microsoft.AspNet.Mvc.Core/ControllerContext.cs
```diff
@@ -0,0 +1,154 @@
...
+        {
+            if (!(context.ActionDescriptor is ControllerActionDescriptor))
+            {
+                throw new ArgumentException(Resources.FormatActionDescriptorMustBeBasedOnControllerAction(
+                    typeof(ControllerActionDescriptor)),
+                    nameof(context));
+            }
+        }
+
+        public new ControllerActionDescriptor ActionDescriptor
```
pranavkm: Docs

src/Microsoft.AspNet.Mvc.Core/ControllerContext.cs
rynowak: What would that look like? It won't be supported.

src/Microsoft.AspNet.Mvc.Core/ControllerContext.cs
rynowak: Sure

src/Microsoft.AspNet.Mvc.WebApiCompatShim/ApiController.cs
```diff
@@ -30,16 +30,14 @@ public abstract class ApiController : IDisposable
         /// <summary>
         /// Gets the action context.
```
pranavkm: > Gets the `<see cref="ActionContext" />`.

Eilon:
Anymore design work needed on this? Looks :shipit: to me.

rynowak:
I'm just waiting for :shipit: from @pranavkm :+1:

src/Microsoft.AspNet.Mvc.Core/ControllerContext.cs
rynowak: Fixed

src/Microsoft.AspNet.Mvc.WebApiCompatShim/ApiController.cs
rynowak: Fixed

test/Microsoft.AspNet.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
```diff
@@ -1351,6 +1351,7 @@ private class Person9
 
             var model = Assert.IsType<Order9>(modelBindingResult.Model);
             Assert.NotNull(model.Customer);
+
```
pranavkm: Undo

test/WebSites/ApplicationModelWebSite/Controllers/ActionModelController.cs
```diff
@@ -15,7 +15,7 @@ public class ActionModelController : Controller
         [ActionName2("ActionName")]
         public string GetActionName()
         {
-            var actionDescriptor = (ControllerActionDescriptor)ActionContext.ActionDescriptor;
+            var actionDescriptor = (ControllerActionDescriptor)ControllerContext.ActionDescriptor;
```
pranavkm: Do you need these casts? Same for the other controller

pranavkm:
:shipit:

PR #3582, opened 2015-11-19
React to aspnet/Razor#358.

NTaylorMullen:
- Removed `WriteTagHelperAsync` methods from `RazorPage`.
- Moved `WriteTagHelperAsync` tests into Razor since `TagHelperOutput` is now an `IHtmlContent`.
- Updated code generation test files.

Razor PR: https://github.com/aspnet/Razor/pull/615

aspnet/Razor#358


src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperContentExtensions.cs
```diff
@@ -51,7 +52,18 @@ public static class TagHelperContentExtensions
```
NTaylorMullen: Realllly dislike this extension method. I'd like to nuke the method and inline this behavior in the `Script` and `Link` `TagHelper`s (method is super misleading as is).

Copied the below uglyness (`StringWriter` stuff) from `RazorPage` because it doesn't need to be first in the page anymore.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -314,106 +314,6 @@ public TagHelperContent EndTagHelperWritingScope()
...
-        /// Writes the content of a specified <paramref name="tagHelperExecutionContext"/> to the specified
-        /// <paramref name="writer"/>.
-        /// </summary>
-        /// <param name="writer">The <see cref="TextWriter"/> instance to write to.</param>
-        /// <param name="tagHelperExecutionContext">The execution context containing the content.</param>
-        /// <returns>
-        /// A <see cref="Task"/> that on completion writes the <paramref name="tagHelperExecutionContext"/> content
-        /// to the <paramref name="writer"/>.
-        /// </returns>
-        public async Task WriteTagHelperToAsync(
```
NTaylorMullen: This stuff is now handled by `TagHelperOutput.WriteTo`

test/Microsoft.AspNet.Mvc.Razor.Test/RazorPageTest.cs
```diff
@@ -1237,767 +1237,6 @@ public static TheoryData WriteAttributeData
             Assert.Equal("Hello world", HtmlContentUtilities.HtmlContentToString(((IHtmlContent)buffer[0])));
         }
 
-        public static TheoryData<TagHelperOutput, string> WriteTagHelper_InputData
```
NTaylorMullen: All these bits got transitioned into the Razor tests.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -483,28 +375,6 @@ public virtual void WriteTo(TextWriter writer, object value)
             var htmlContent = value as IHtmlContent;
             if (htmlContent != null)
             {
-                if (escapeQuotes)
```
rynowak: So nice to see this become more isolated

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperContentExtensions.cs
rynowak: I've never seen this before. This is pretty ugly.

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperContentExtensions.cs
```diff
@@ -51,7 +52,18 @@ public static class TagHelperContentExtensions
...
-                RazorPage.WriteTo(writer, encoder, value, escapeQuotes: true);
+                using (var stringWriter = new StringWriter())
+                {
+                    RazorPage.WriteTo(stringWriter, encoder, value);
+
+                    // In this case the text likely came directly from the Razor source. Since the original string is
+                    // an attribute value that may have been quoted with single quotes, must handle any double quotes
+                    // in the value. Writing the value out surrounded by double quotes.
+                    //
+                    // This is really not optimal from a perf point of view, but it's the best we can do for right now.
```
dougbu: Remove this line. Same [reasons](https://github.com/aspnet/Razor/pull/615#discussion-diff-45298662) as the comment in aspnet/Razor#615.

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperContentExtensions.cs
dougbu: Why yes, it's purpose is ugly. That's one of the reasons I didn't want to make the method `public` when I wrote the original version.

Might be possible to avoid duplicating the ugliness if the `<link>` and `<script>` tag helpers created the additional elements using a `TagHelperOutput` instance per added element. That's possible with the new Razor code but may also be ugly.

dougbu: That said, this class is getting uglier and the Razor changes may provide a way to get rid of it. Just not sure whether rewriting those two tag helpers should be done in this PR versus just after.

dougbu:
:watch: for `TagHelperContentExtensions` discussion.

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperContentExtensions.cs
NTaylorMullen: Don't want to transition `Link` and `Script` tags to use `TagHelperOutput` since writing out their content as is is as performant as we could ever hope for :smile:. So if we stick with the writing the tags inline I'd prefer to nuke the method and move the logic into the `Script`/`Link` `TagHelper`s

NTaylorMullen:
https://github.com/aspnet/Mvc/pull/3582#discussion_r45382548

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperContentExtensions.cs
dougbu: :ok: I guess. I'll see if de-weird-ing `LinkTagHelper` and `ScriptTagHelper` fits into one of my assigned "don't encode 'til absolutely necessary" bugs.

dougbu:
:shipit:

Issue #3587, opened 2015-11-19
`CompilationFailuresAreListedByErrorPageMiddleware` functional tests fail on Mac


dougbu:
See last two failures in [Travis job 3915.2](https://travis-ci.org/aspnet/Mvc/jobs/92083777). Looks as if `CompilationFailedException`s are propagating though the `ErrorPageMiddleware`. But without debugging further, can't be sure where the problem lies. Need to investigate and file the appropriate CoreFx, DNX or `ErrorPageMiddleware` issue or resolve in this repo.

dougbu:
Nice, @Eilon must be buying me a Mac :smiley_cat:

Eilon:
Use the Mac you borrowed :wink:

ryanbrandenburg:
As @pranavkm said, it looks like this one's fixed. Closing.

PR #3588, opened 2015-11-19
Add temporary workarounds for aspnet/External#50 and aspnet/Mvc#3587

dougbu:


dougbu:
FYI only since these are xplat test-only changes. Will merge immediately.

dougbu:
79d7412

PR #3590, opened 2015-11-20
React to hosting changes

pakrym:
Fixes build
@ryanbrandenburg  @rynowak


test/Microsoft.AspNet.Mvc.FunctionalTests/MvcTestFixture.cs
```diff
@@ -110,7 +110,7 @@ protected virtual void AddAdditionalServices(IServiceCollection services)
                     new TestApplicationEnvironment(applicationEnvironment, applicationName, applicationRoot));
 
                 var hostingEnvironment = new HostingEnvironment();
-                hostingEnvironment.Initialize(applicationRoot, new WebHostOptions());
+                hostingEnvironment.Initialize(applicationRoot, new WebHostOptions(), null);
```
dougbu: nit: please name this parameter so we know what it is w/o hovering

dougbu:
:shipit: (actually don't wait for my squirrel; just fix the broken build)

Issue #3586, opened 2015-11-19
TryValidateModel throws exception when unit tested


ivaylokenov:
When TryValidateModel is called in a controller, and then the controller is unit tested, the method throws exception. Tried with both beta8 and rc1. Project.json is the default one with referenced Microsoft.AspNet.Mvc, Microsoft.AspNet.Http and Xunit. In ASP.NET Web API 2, this was quite easy to test. Sample code:

Controller: 
```c#
public class HomeController : Controller
    {
        [HttpPost]
        public IActionResult Index(SomeModel model)
        {
            this.TryValidateModel(model);

            if (this.ModelState.IsValid)
            {
                return Ok();
            }

            return HttpBadRequest();
        }
    }
```

Model:
```c#
public class SomeModel
    {
        [Required]
        public int Int { get; set; }

        [Required]
        [StringLength(20, MinimumLength = 5)]
        public string String { get; set; }
    }
```

Unit test:
```c#
[Fact]
        public void TryValidateModelTest()
        {
            // GetController is the same method used in the MVC source code tests
            var controller = GetController(new GenericModelBinder(), new CompositeValueProvider());

            controller.Index(new SomeModel()); // throws null reference exception, test fails

            Assert.False(controller.ModelState.IsValid);
        }
```

Stacktrace of the failed test:
```
Test Name:	ControllersTest.TryValidateModelTest
Test FullName:	TestApp.Tests.ControllersTest.TryValidateModelTest
Test Source:	c:\users\admin\documents\visual studio 2015\Projects\WebApplication2\src\ClassLibrary1\Class1.cs : line 48
Test Outcome:	Failed
Test Duration:	0:00:00.075

Result StackTrace:	
at Microsoft.AspNet.Mvc.ModelBinding.Validation.DataAnnotationsModelValidatorProvider.GetValidators(ModelValidatorProviderContext context)
   at Microsoft.AspNet.Mvc.ModelBinding.Validation.ValidationVisitor.GetValidators(ModelMetadata metadata)
   at Microsoft.AspNet.Mvc.ModelBinding.Validation.ValidationVisitor.ValidateNode()
   at Microsoft.AspNet.Mvc.ModelBinding.Validation.ValidationVisitor.VisitSimpleType()
   at Microsoft.AspNet.Mvc.ModelBinding.Validation.ValidationVisitor.Visit(ModelMetadata metadata, String key, Object model)
[...]

PR #3591, opened 2015-11-20
Ease unit testing of `Controller.TryValidateModel()`

dougbu:
- #3586
- reverse conditions to access `_options.Value` only when test has set up an `IStringLocalizerFactory`

### DIFF
--- a/test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ControllerTest.cs
+++ b/test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ControllerTest.cs
@@ -1771,6 +1771,22 @@ namespace Microsoft.AspNet.Mvc.Test
         }
 
         [Fact]
+        public void TryValidateModel_Succeeds_WithoutValidatorMetadata()
+        {
+            // Arrange
+            var model = new TryValidateModelModel();
+            var binder = new GenericModelBinder();
+            var valueProvider = new CompositeValueProvider();
+            var controller = GetController(binder, valueProvider);
+
+            // Act
+            var result = controller.TryValidateModel(model);
+
+            // Assert
+            Assert.True(controller.ModelState.IsValid);
+        }
+
+        [Fact]
         public void TryValidateModelEmptyBindingContextThrowsException()
         {
             // Arrange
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ControllerTest.cs
```diff
@@ -1771,6 +1771,22 @@ public void TryValidateModelWithInvalidModelNoPrefix_ReturnsFalse()
...
+        public void TryValidateModel_Succeeds_WithoutValidatorMetadata()
+        {
+            // Arrange
+            var model = new TryValidateModelModel();
+            var binder = new GenericModelBinder();
+            var valueProvider = new CompositeValueProvider();
+            var controller = GetController(binder, valueProvider);
+
+            // Act
+            var result = controller.TryValidateModel(model);
```
dougbu: Test blew up here with same stack trace as #3586 before my change.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ControllerTest.cs
```diff
@@ -1771,6 +1771,22 @@ public void TryValidateModelWithInvalidModelNoPrefix_ReturnsFalse()
         }
 
         [Fact]
+        public void TryValidateModel_Succeeds_WithoutValidatorMetadata()
+        {
+            // Arrange
+            var model = new TryValidateModelModel();
+            var binder = new GenericModelBinder();
```
rynowak: Why this model binder in particular? I'd expect model binding to do nothing / not be called.

rynowak:
:shipit:

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ControllerTest.cs
dougbu: Very `true` that `ActionBindingContext` is fine w/ setting `ModelBinder` to `null` and the binder won't be used in this scenario. But I'm matching the OP's code in #3586.

The `IValueProvider` will also not be used.

The only real difference between this test and `TryValidateModelWithValidModel_ReturnsTrue` is we're using a real `IModelValidatorProvider` here. That test succeeded because the mock provider provided an empty array and never reached the product code I changed.

Are you suggesting I switch to `GetController(binder: null, provider: null)` since that's the minimal setup needed to hit the original issue?

dougbu:
6297822

PR #3594, opened 2015-11-20
Removed unnecessary call in Controller.TryValidateModel

ivaylokenov:
In Controller.TryValidateModel MetadataProvider.GetModelExplorerForType is called and its return value is not used afterwards. I could not find any particular reason for the call and all tests pass after removing it.


rynowak:
This is some vestigial code that it looks like we never removed! Good catch.

rynowak:
:shipit: - I'll take care of merging this.

rynowak:
Merged 535fdf3356f4237ff8dfc0fdaa71842f5a525a63 thanks!

Issue #3428, opened 2015-10-28
Make `ModelBinderAttribute.BindingSource` setter `protected`


genusP:
For this code
```
public IActionResult Add(
            string entity, 
            [ModelBinder(BinderType = typeof(EntityViewModelBinder), BindingSource = BindingSource.Form)]
            object viewModel
            )
        {
...
}
```

Result:
Error	CS0655	'BindingSource' is not a valid named attribute argument because it is not a valid attribute parameter type	WebApplication2.DNX 4.5.1

.NET Framework Clr version 4.0.30319.42000   |   DNX x86 version 1.0.0-beta8-15858   |   Microsoft.AspNet.Hosting version 1.0.0-beta8-15756   |   Windows 7.0

danroth27:
You need to derive your own attribute to specify the BindingSource.

We should make this clearer by making the BindingSource setter protected.

geniusP:
I derive my attribute:
```
[AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct | AttributeTargets.Enum | AttributeTargets.Property | AttributeTargets.Parameter, AllowMultiple = false, Inherited = true)]
    public class EntityViewModelBinderAttribute:ModelBinderAttribute
    {
        public EntityViewModelBinderAttribute()//:base()
        {
            BinderType = typeof(EntityViewModelBinder);
            BindingSource = Microsoft.AspNet.Mvc.ModelBinding.BindingSource.ModelBinding;
        }
    }
```
And set this attribute to parameter:
```
public IActionResult Add(
            string entity, 
            [EntityViewModelBinderAttribute]
            object viewModel
            )
        {
```
Result. My binder not calling. Why?

dougbu:
MVC itself never subclasses `[ModelBinder]`. Instead `[FromBody]` and similar attributes implement `IBindingSourceMetadata` themselves. Wondering about removing the setter for `ModelBinderAttribute.BindingSource` entirely. Getter would remain as-is, indicating whether the `BinderType` had been set (making it a `BindingSource.Custom` binder).

@Eilon / @danroth27 / @rynowak thoughts?

Eilon:
Why not just make it protected?

dougbu:
> Why not just make it `protected`?

We have a simple way to create an attribute that specifies the `BindingSource` e.g.
``` c#
    [AttributeUsage(AttributeTargets.Parameter | AttributeTargets.Property, AllowMultiple = false, Inherited = true)]
[...]

PR #3597, opened 2015-11-20
Make `ModelBinderAttribute.BindingSource` setter `protected`

dougbu:
- #3428
- the `public` setter was not useful when this class is used as an attribute
- make property `virtual` to support other override patterns
- update existing test to use both override patterns

### DIFF
--- a/test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/Metadata/ModelBinderAttributeTest.cs
+++ b/test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/Metadata/ModelBinderAttributeTest.cs
@@ -1,7 +1,6 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
 using Xunit;
 
 namespace Microsoft.AspNet.Mvc.ModelBinding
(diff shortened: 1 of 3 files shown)


test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/Metadata/ModelBinderAttributeTest.cs
```diff
@@ -49,5 +46,12 @@ public void BinderType_SettingBindingSource_OverridesDefaultCustomBindingSource(
             // Assert
             Assert.Equal(BindingSource.Query, source);
         }
+
+        private class FromQueryModelBinderAttribute : ModelBinderAttribute
+        {
+            // Not the perfect way to override this property since setting its value (possible when not using the class
+            // as an attribute) has no effect. base.BindingSource && BindingSource.Query may be right for some cases.
```
rynowak: Not sure what relevance this comment has outside the scope of this PR.

rynowak:
:shipit:

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/Metadata/ModelBinderAttributeTest.cs
dougbu: Yeah, meant to say "Reviewers: ..." since users aren't going to look at this as an example of what they should or shouldn't do. Planned to remove.

dougbu:
45ff0a2

PR #3600, opened 2015-11-21
Removing PageInstrumentation

### LINKED ISSUE #3497: Remove PageInstrumentation
Follow up to https://github.com/aspnet/Mvc/issues/3281 - once BrowserLink is updated to use DiagnosticSource, we need to remove PageInstrumentation and its use from Mvc.

pranavkm:
Fixes #3497


davidfowl:
:shipit:

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -43,25 +41,13 @@ public abstract class RazorPage : IRazorPage
...
-            {
-                if (ViewContext == null)
-                {
-                    return null;
-                }
-
-                return ViewContext.HttpContext;
-            }
-        }
+        public HttpContext Context => ViewContext?.HttpContext;
```
dougbu: Why these unrelated style changes? Whitespace updates sure. But pure style changes usually just results in churn.

src/Microsoft.AspNet.PageExecutionInstrumentation.Interfaces/IBufferedTextWriter.cs
```diff
@@ -1,32 +0,0 @@
...
-
-using System.IO;
-using System.Threading.Tasks;
-
-namespace Microsoft.AspNet.Mvc.Razor
-{
-    /// <summary>
-    /// Specifies the contracts for a <see cref="TextWriter"/> that buffers its content.
-    /// </summary>
-    public interface IBufferedTextWriter
```
dougbu: How does removing `IBufferedTextWriter` fit into this PR?

dougbu: Put another way, why was this `interface` in this project to begin with?

test/WebSites/RazorPageExecutionInstrumentationWebSite/Startup.cs
```diff
@@ -22,24 +20,28 @@ public void ConfigureServices(IServiceCollection services)
...
-            services.AddScoped<TestPageExecutionListenerFeature, TestPageExecutionListenerFeature>();
-            services.AddScoped<IHoldInstrumentationData>(
-                provider => provider.GetRequiredService<TestPageExecutionListenerFeature>().Holder);
         }
 
         public void Configure(IApplicationBuilder app)
         {
+            var listener = new RazorPageDiagnosticListener();
+            var diagnosticSource = app.ApplicationServices.GetRequiredService<DiagnosticListener>();
+            diagnosticSource.SubscribeWithAdapter(listener);
```
dougbu: This setup looks suspect because the only registered `DiagnosticListener` is a singleton. That implies this site can be used once and only once, undermining the text fixtures we use.

Also odd because the `diagnosticSource` has type `DiagnosticListener`.

test/WebSites/RazorPageExecutionInstrumentationWebSite/project.json
```diff
@@ -3,13 +3,14 @@
...
-        "Microsoft.AspNet.StaticFiles": "1.0.0-*"
-    },
+  "dependencies": {
+    "Microsoft.AspNet.Mvc": "6.0.0-*",
+    "Microsoft.AspNet.Mvc.TestConfiguration": "1.0.0",
+    "Microsoft.AspNet.Server.Kestrel": "1.0.0-*",
+    "Microsoft.AspNet.Server.WebListener": "1.0.0-*",
+    "Microsoft.AspNet.StaticFiles": "1.0.0-*",
+    "Microsoft.Extensions.DiagnosticAdapter": "1.0.0-*"
+  },
```
dougbu: Reformat rest of the file to match.

test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/RazorPageExecutionInstrumentationWebSite.Home.ViewWithPartial.html
```diff
@@ -5,23 +5,23 @@
...
+/Views/_ViewStart.cshtml: Non-literal at 96 contains 16 characters.
+/Views/_ViewStart.cshtml: Literal at 112 contains 2 characters.
+/Views/Home/ViewWithPartial.cshtml: Literal at 0 contains 27 characters.
+/Views/Home/ViewWithPartial.cshtml: Non-literal at 28 contains 39 characters.
+/Views/Home/_PartialView.cshtml: Literal at 31 contains 2 characters.
+/Views/Home/_PartialView.cshtml: Literal at 33 contains 8 characters.
+/Views/Home/_PartialView.cshtml: Non-literal at 41 contains 4 characters.
+/Views/Home/_PartialView.cshtml: Literal at 45 contains 1 characters.
+/Views/Home/_PartialView.cshtml: Literal at 46 contains 20 characters.
+/Views/Home/ViewWithPartial.cshtml: Literal at 67 contains 2 characters.
```
dougbu: `_PartialView.cshtml` -> `ViewWithPartial.cshtml` doesn't look right. Is something wrong w/ the new infrastructure?

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
pranavkm: :+1: I'll merge it in as a separate commit.

src/Microsoft.AspNet.PageExecutionInstrumentation.Interfaces/IBufferedTextWriter.cs
pranavkm: It was used as part of a contract with PageInstrumentation

test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/RazorPageExecutionInstrumentationWebSite.Home.ViewWithPartial.html
pranavkm: The existing code was incorrect - we don't have 67 characters in `_PartialView`. Additionally the line mappings for `_Layout` were also incorrect. I didn't spend time investigating why it was incorrect.

test/WebSites/RazorPageExecutionInstrumentationWebSite/Startup.cs
pranavkm: We don't have more than one tests here - so it really doesn't matter. We could fix it if we ever add more. `DiagnosticListener` derives from `DiagnosticSource` and it would have been confusing to have two `listeners` here.

pranavkm:
:Up: :date:

test/WebSites/RazorPageExecutionInstrumentationWebSite/Startup.cs
dougbu: It does matter because users look at these sites for inspiration. If we can't set up a per-request diagnostic sink (as I was able to do w/ the old approach), something is wrong here. Suggest we find a way to solve this before pushing this PR.

test/WebSites/RazorPageExecutionInstrumentationWebSite/Startup.cs
```diff
@@ -22,24 +20,28 @@ public void ConfigureServices(IServiceCollection services)
...
+            app.Use(async (context, next) =>
             {
-                var listenerFeature = context.RequestServices.GetRequiredService<TestPageExecutionListenerFeature>();
-                context.Features.Set<IPageExecutionListenerFeature>(listenerFeature);
-
-                return next();
+                // Add MVC to the request pipeline
+                await next();
+
+                using (var writer = new HttpResponseStreamWriter(context.Response.Body, Encoding.UTF8))
```
dougbu: Suggestion from @lodejard: Give MVC `TextWriter.Null` and have `RazorPageDiagnosticListener` write directly to the response as the events occur. No reason to cache the events anywhere.

test/WebSites/RazorPageExecutionInstrumentationWebSite/Startup.cs
dougbu: My suggestion is to do the least you need to make the test work, not to register something globally. For example could place your own source and listener into a feature within your middleware callback. Or could register what you need as `Scoped` in DI.

dougbu:
:watch:

pranavkm:
:up: :date:

test/WebSites/RazorPageExecutionInstrumentationWebSite/Startup.cs
dougbu: Nice. The `HttpContext.Items` approach means the listener has no state :smiley_cat:

dougbu:
:shipit:

PR #3602, opened 2015-11-21
Remove per-instance string.Concat

rynowak:
This is allocating per request, and at 500kb / 3000 req this is definitely something we want to fix.

image

### DIFF
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
@@ -35,6 +35,8 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
     public class LinkTagHelper : UrlResolutionTagHelper
     {
         private static readonly string Namespace = typeof(LinkTagHelper).Namespace;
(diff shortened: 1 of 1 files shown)


dougbu:
:shipit: though I'm surprised the compiler doesn't do better. Nice we can worry about 186 unneeded bytes per request.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -35,6 +35,8 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
     public class LinkTagHelper : UrlResolutionTagHelper
     {
         private static readonly string Namespace = typeof(LinkTagHelper).Namespace;
+        private static readonly string FallbackJavaScriptResourceName = 
+            Namespace + ".compiler.resources.LinkTagHelper_FallbackJavaScript.js";
```
pranavkm: Could you just inline `Namespace`. It seems like such a poor variable \ field name.

rynowak: Sure

PR #3613, opened 2015-11-23
Fixing local parameter usage

sebastienros:
The argument should be used instead of the property.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
@@ -177,7 +177,7 @@ namespace Microsoft.AspNet.Mvc.Razor
             {
                 for (var i = 0; i < viewStartPages.Count; i++)
                 {
-                    var viewStart = ViewStartPages[i];
+                    var viewStart = viewStartPages[i];
                     context.ExecutingFilePath = viewStart.Path;
 
                     // Copy the layout value from the previous view start (if any) to the current.


rynowak:
What's the bug?

sebastienros:
It's using a mix of the argument parameter and the class one. In this case I assume the goal is to use the argument, or there is no point in passing it. I'll show you tomorrow if you don't understand, look closely at the case of the variable.

rynowak:
I understand the change you made to the code. Is there a bug here?

The change should probably refactor the code further.

dougbu:
The parameter is passed only because the single caller needs that parameter to determine whether or not `_ViewStart.cshtml` files are relevant. I.e. the method isn't called unless `ViewStartPages == viewStartPages`.

Could certainly change the calling method to instead have a `invokeViewStarts` parameter and use `ViewStartPages` (the property) everywhere. But there's no bug here.

Eilon:
@sebastienros @rynowak @dougbu can you guys figure out what the full fix is for this and update the PR?

dougbu:
My suggestion is (in a bit more detail):
- use `ViewStartPages` throughout `RazorView` e.g. remove `viewStartPages` parameter from `RenderViewStartsAsync()`.
- change `RenderPageAsync()` to instead have a `bool invokeViewStarts` parameter
- call `RenderPageAsync()` with `invokeViewStarts: true` from `RenderAsync()` and `invokeViewStarts: false` from `RenderLayoutAsync()`

sebastienros:
Nothing to add on this PR, I just wanted to point out that the wrong variable name was used. Feel free to close this PR and open a new issue if there is a more general problem with this class.

Eilon:
@dougbu can you do the deed?

dougbu:
:ok:

Issue #3571, opened 2015-11-17
Remove use of `WebUtility.HtmlEncode()` and `...HtmlDecode()`


dougbu:
`ContentViewComponentResult` uses both `System.Net.Http.WebUtility.HtmlEncode()` and `...HtmlDecode()`. Looks like a dusty legacy corner.

Use `System.Text.Encodings.Web.HtmlEncoder.Encode()` instead of `WebUtility.HtmlEncode()`. And if we don't yet have HTML decoders, perhaps the `ContentViewComponentResult` pattern should be rethought.

davidfowl:
Yah

rynowak:
We should just delete the constructor that takes `HtmlString`. You can return `HtmlString` directly from a VC so it's not really important.

dougbu:
eb70b1c

dougbu:
@danroth27 @Eilon the fix for this bug also added support for view components returning `IHtmlContent` (not just `HtmlString`). Should this `task` thus be considered a (fixed) `bug` or perhaps even an `enhancement`?

Eilon:
Doubt it matters.

PR #3615, opened 2015-11-23
Remove `WebUtility` use and handle `IHtmlContent` return values from view components

dougbu:
- #3571 and part of #3123
- split `ContentViewComponentResult` in two
- add `HtmlEncoder` to `ViewComponentContext`
- remove use of `WebUtility.HtmlEncode()` and `HtmlDecode()`

nits: remove unused `using`s in files I had open


dougbu:
Performance wasn't the goal of this change but avoiding `string` allocations when HTML encoding was not necessary did have some benefit. Numbers below are for 3000 requests and a simple site with one view component that returns a short `string`.

Before:
before

After:
after

rynowak:
@dougbu - have you puzzled out there the reduction in allocations came from? The total number indicates about .7mb was shaved off, the difference in byte[] was about .5mb. 

byte[] is generally the 'experimental error' - it's variable based on the amount of server load, and the amount of output written (until kestrel finishes the SocketOutput work).

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponent.cs
```diff
@@ -201,15 +201,15 @@ public ICompositeViewEngine ViewEngine
         /// Returns a result which will render HTML encoded text.
         /// </summary>
         /// <param name="content">The content, will be HTML encoded before output.</param>
-        /// <returns>A <see cref="ContentViewComponentResult"/>.</returns>
-        public ContentViewComponentResult Content(string content)
+        /// <returns>A <see cref="StringViewComponentResult"/>.</returns>
+        public StringViewComponentResult Content(string content)
```
rynowak: Why? This is meant to be an analouge of `ContentResult`

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentHelper.cs
```diff
@@ -229,7 +238,7 @@ private ViewComponentDescriptor SelectComponent(Type componentType)
                 throw new ArgumentNullException(nameof(descriptor));
             }
 
-            var context = new ViewComponentContext(descriptor, arguments, _viewContext, writer);
+            var context = new ViewComponentContext(descriptor, arguments, _htmlEncoder, _viewContext, writer);
```
rynowak: Should we make the encoder part of the `ViewContext` instead? It seems to show up everywhere that's used.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponent.cs
rynowak: I like the cleanup of this class, but not the rename.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentHelper.cs
dougbu: Let's consider this later, once other aspects of HTML encoding settle.

rynowak: Agreed, I just want to pose the question.

dougbu:
> have you puzzled out there the reduction in allocations came from?

Nah. `WebUtility` looks suspect but I'm not going to dive deeper.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponent.cs
dougbu: Restored old name for the `string` result.

dougbu:
:up::date:

rynowak:
:shipit:

dougbu:
eb70b1c

PR #3625, opened 2015-11-23
Bump Razor to dotnet5.5

pranavkm:
cc @Eilon


Kagamine:
why bump to 5.5?

davidfowl:
The dependency on System.Runtime.Loader required dotnet5.5

PR #3626, opened 2015-11-24
Log exceptions in JsonInputFormatter

### LINKED ISSUE #3557: JsonInputFormatter doesn't log errors
I am on MVC 6.0.0-beta8 and I had a JSON deserialization failure. It didn't log any errors but instead it just didn't bind anything and populated the model state. This is the response I had:

```json
51
{"startedAt":["The input was not valid."],"endedAt":["The input was not valid."]}
0


```

When I tried the same deserialization in isolation, I got the below error from `JsonConvert.DeserializeObject<T>`:

> An unhandled exception of type 'NodaTime.Utility.InvalidNodaDataException' occurred in Newtonsoft.Json.dll

> Additional information: Unexpected token parsing Instant. Expected String, got Date.

I would expect to see this as a Debug level log so that I wouldn't need to dig deep to see what is going wrong. Would be also useful to log more info along with the error like what is the value, etc.

ryanbrandenburg:
Fixes #3557.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Formatters.Json/Internal/MvcJsonMvcOptionsSetup.cs
+++ b/src/Microsoft.AspNet.Mvc.Formatters.Json/Internal/MvcJsonMvcOptionsSetup.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.OptionsModel;
 using Microsoft.Net.Http.Headers;
 using Newtonsoft.Json;
@@ -10,18 +11,23 @@ namespace Microsoft.AspNet.Mvc.Formatters.Json.Internal
 {
     public class MvcJsonMvcOptionsSetup : ConfigureOptions<MvcOptions>
     {
-        public MvcJsonMvcOptionsSetup(IOptions<MvcJsonOptions> jsonOptions)
-            : base((_) => ConfigureMvc(_, jsonOptions.Value.SerializerSettings))
+        public MvcJsonMvcOptionsSetup(ILoggerFactory loggerFactory, IOptions<MvcJsonOptions> jsonOptions)
+            : base((_) => ConfigureMvc(_, jsonOptions.Value.SerializerSettings, loggerFactory))
         {
         }
 
-        public static void ConfigureMvc(MvcOptions options, JsonSerializerSettings serializerSettings)
+        public static void ConfigureMvc(
+            MvcOptions options,
+            JsonSerializerSettings serializerSettings,
+            ILoggerFactory loggerFactory)
         {
+            var jsonInputLogger = loggerFactory.CreateLogger<JsonInputFormatter>();
+            var jsonInputPatchLogger = loggerFactory.CreateLogger<JsonPatchInputFormatter>();
+
             options.OutputFormatters.Add(new JsonOutputFormatter(serializerSettings));
-
-            options.InputFormatters.Add(new JsonInputFormatter(serializerSettings));
-            options.InputFormatters.Add(new JsonPatchInputFormatter(serializerSettings));
-
+            options.InputFormatters.Add(new JsonInputFormatter(jsonInputLogger, serializerSettings));
+            options.InputFormatters.Add(new JsonPatchInputFormatter(jsonInputPatchLogger, serializerSettings));
+            
             options.FormatterMappings.SetMediaTypeMappingForFormat("json", MediaTypeHeaderValue.Parse("application/json"));
 
             options.ValidationExcludeFilters.Add(typeof(JToken));
--- a/src/Microsoft.AspNet.Mvc.Formatters.Json/JsonInputFormatter.cs
+++ b/src/Microsoft.AspNet.Mvc.Formatters.Json/JsonInputFormatter.cs
@@ -2,12 +2,11 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.IO;
-using System.Text;
 using System.Threading.Tasks;
-using Microsoft.AspNet.Mvc.Infrastructure;
+using Microsoft.AspNet.Mvc.Formatters.Json.Logging;
 using Microsoft.AspNet.Mvc.Internal;
 using Microsoft.AspNet.Mvc.ModelBinding;
+using Microsoft.Extensions.Logging;
 using Newtonsoft.Json;
 
 namespace Microsoft.AspNet.Mvc.Formatters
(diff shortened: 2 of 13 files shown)


src/Microsoft.AspNet.Mvc.Formatters.Json/JsonInputFormatter.cs
```diff
@@ -72,6 +72,9 @@ public override Task<InputFormatterResult> ReadRequestBodyAsync(InputFormatterCo
                 return InputFormatterResult.FailureAsync();
             }
 
+            var loggerFactory = context.HttpContext.RequestServices.GetRequiredService<ILoggerFactory>();
+            var logger = loggerFactory.CreateLogger<FileResult>();
```
rynowak: ???

rynowak: We should be passing this in the constructor.

src/Microsoft.AspNet.Mvc.Formatters.Json/Logging/JsonInputFormatterLoggerExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace Microsoft.AspNet.Mvc.Formatters.Json.Logging
+{
+    internal static class JsonInputFormatterLoggerExtensions
+    {
+        private static readonly Action<ILogger, string, Exception> _jsonInputFormatterCrashed;
+
+        static JsonInputFormatterLoggerExtensions()
+        {
+            _jsonInputFormatterCrashed = LoggerMessage.Define<string>(
+                LogLevel.Debug,
```
rynowak: Use verbose. I'm not aware of anywhere we've used debug, and there's an ongoing debate about what exactly it means.

rynowak: Actually let's talk to @lodejard about the right level here. I'm not sure this topic has been figured out fully.

davidfowl: Debug log level is going away.https://github.com/aspnet/Logging/issues/299

Eilon: We were just going to re-order Debug, I believe, not make it go away.

Eilon: Though I guess the jury is still out.

ryanbrandenburg: In person conversation with @lodejard says we're going for 'Verbose'.

test/Microsoft.AspNet.Mvc.Formatters.Json.Test/JsonInputFormatterTest.cs
```diff
@@ -187,6 +195,13 @@ public static IEnumerable<object[]> JsonFormatterReadSimpleTypesData
             Assert.True(result.HasError);
             Assert.Equal("The supplied value is invalid for Byte.", modelState["[2]"].Errors[0].ErrorMessage);
             Assert.Null(modelState["[2]"].Errors[0].Exception);
+            Mock.Get<ILogger>(logger).Verify(l => l.Log(
+                LogLevel.Verbose,
+                It.IsAny<int>(),
+                It.IsAny<object>(),
+                It.IsAny<Exception>(),
+                It.IsAny<Func<object, Exception, string>>()),
+            Times.Once);
```
Eilon: This formatting hurts my eyes. What is this testing, anyway?

test/Microsoft.AspNet.Mvc.Formatters.Json.Test/JsonOutputFormatterTests.cs
```diff
@@ -284,6 +286,13 @@ public void Constructor_UsesSerializerSettings()
             return encoding;
         }
 
+        private static ILogger GetLogger()
+        {
+            var logger = new Mock<ILogger>();
```
Eilon: BTW we do have a TestLogger in the Logger.Testing project. Would that help here?

test/Microsoft.AspNet.Mvc.Formatters.Json.Test/JsonInputFormatterTest.cs
ryanbrandenburg: That when we have an error coming out of the formatter we log the error in addition to returning it. I suppose it's probably mostly paranoia on my to double check that the logging happened, I I'll remove it.

src/Microsoft.AspNet.Mvc.Formatters.Json/Logging/JsonInputFormatterLoggerExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+    internal static class JsonInputFormatterLoggerExtensions
+    {
+        private static readonly Action<ILogger, string, Exception> _jsonInputFormatterCrashed;
+
+        static JsonInputFormatterLoggerExtensions()
+        {
+            _jsonInputFormatterCrashed = LoggerMessage.Define<string>(
+                LogLevel.Verbose,
+                1,
+                "Json input formatting failed, exception was '{0}'.");
```
kichalla: @Eilon might consider this as a robotic message :smile: ...probably something like 'an exception occurred while attempting to deserialize the request body'? Also you can remove the `{0}` as the exception detail is always provided as a parameter to a logger...example: https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs#L87

src/Microsoft.AspNet.Mvc.Formatters.Json/Internal/MvcJsonMvcOptionsSetup.cs
```diff
@@ -10,18 +11,23 @@ namespace Microsoft.AspNet.Mvc.Formatters.Json.Internal
...
+        public static void ConfigureMvc(
+            MvcOptions options,
+            JsonSerializerSettings serializerSettings,
+            ILoggerFactory loggerFactory)
         {
-            options.OutputFormatters.Add(new JsonOutputFormatter(serializerSettings));
-
-            options.InputFormatters.Add(new JsonInputFormatter(serializerSettings));
-            options.InputFormatters.Add(new JsonPatchInputFormatter(serializerSettings));
+            var inputLogger = loggerFactory.CreateLogger<JsonInputFormatter>();
```
kichalla: nit: better variable names

src/Microsoft.AspNet.Mvc.Formatters.Json/Logging/JsonInputFormatterLoggerExtensions.cs
ryanbrandenburg: Ah, good point on the exception. I figured the message might not pass muster but couldn't come up with a better one mostly because I was trying to shoe-horn the exception in there.

test/Microsoft.AspNet.Mvc.Formatters.Json.Test/JsonInputFormatterTest.cs
Eilon: Maybe just fix the indentation then? The `Times.Once` can't be on the same line as `Mock.Get`.

Try something like:

```c#
Mock
    .Get<ILogger>(logger)
    .Verify(l => l.Log(  
        LogLevel.Verbose,  
        It.IsAny<int>(),  
        It.IsAny<object>(),  
        It.IsAny<Exception>(),  
        It.IsAny<Func<object, Exception, string>>()),  
    Times.Once);  
```

src/Microsoft.AspNet.Mvc.Formatters.Json/Logging/JsonInputFormatterLoggerExtensions.cs
Eilon: Well, for one, `Json` needs to be `JSON`. But aside from that, when logging an exception does there need to be a `{0}` in the message or will the system automatically append the exception details to the log anyway? I forget how the logging exception formatting stuff works.

But back to the error message itself, how about: `The JSON input formatter threw the following exception: {0}` (and if the `{0}` ends up not being needed, say `... threw an exception.`)

ryanbrandenburg:
:up: :date:

Eilon:
:shipit:

PR #3634, opened 2015-11-24
MVC companion to aspnet/Routing#238

rynowak:
This is a companion change to make the RouteValueDictionary type more
promient. Using the concrete type here allows to avoid allocations in a
lot of common scenarios.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/RedirectToActionResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/RedirectToActionResult.cs
@@ -2,10 +2,10 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
 using Microsoft.AspNet.Mvc.Core;
 using Microsoft.AspNet.Mvc.Logging;
 using Microsoft.AspNet.Mvc.ViewFeatures;
+using Microsoft.AspNet.Routing;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 
@@ -16,7 +16,7 @@ namespace Microsoft.AspNet.Mvc
         public RedirectToActionResult(
             string actionName,
             string controllerName,
-            IDictionary<string, object> routeValues)
+            object routeValues)
             : this(actionName, controllerName, routeValues, permanent: false)
         {
         }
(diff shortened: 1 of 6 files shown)


rynowak:
See https://github.com/aspnet/Routing/pull/238/ for details

src/Microsoft.AspNet.Mvc.Core/RedirectToActionResult.cs
```diff
@@ -24,12 +24,12 @@ public class RedirectToActionResult : ActionResult, IKeepTempDataResult
         public RedirectToActionResult(
             string actionName,
             string controllerName,
-            IDictionary<string, object> routeValues,
+            object routeValues,
```
rynowak: This was the only one of these that wasn't using `object`

src/Microsoft.AspNet.Mvc.Core/RedirectToActionResult.cs
```diff
@@ -39,7 +39,7 @@ public class RedirectToActionResult : ActionResult, IKeepTempDataResult
 
         public string ControllerName { get; set; }
 
-        public IDictionary<string, object> RouteValues { get; set; }
+        public RouteValueDictionary RouteValues { get; set; }
```
pranavkm: If you're feel overtly enthusiastic, copy the description from the other `*Result`s

src/Microsoft.AspNet.Mvc.Core/Routing/UrlHelper.cs
```diff
@@ -46,7 +46,7 @@ public virtual string Action(UrlActionContext actionContext)
                 throw new ArgumentNullException(nameof(actionContext));
             }
 
-            var valuesDictionary = PropertyHelper.ObjectToDictionary(actionContext.Values);
```
pranavkm: Could we delete `ObjectToDictionary` or are there other places its still being used?

pranavkm:
:shipit:

src/Microsoft.AspNet.Mvc.Core/Routing/UrlHelper.cs
rynowak: I'm sure there are other places

src/Microsoft.AspNet.Mvc.Core/RedirectToActionResult.cs
rynowak: Sure

src/Microsoft.AspNet.Mvc.Core/Routing/UrlHelper.cs
dougbu: For example `HtmlHelper` uses `ObjectToDictionary` a lot.

PR #3635, opened 2015-11-24
Couple more parts of #3123

dougbu:
- update `IJsonHelper` to use `IHtmlContent`
- update `IHtmlGenerator.GenerateAntiforgery()` and related bits to use `IHtmlContentBuilder`

nit: quiet `initialization` target in `MvcSandbox` project

### DIFF
--- a/samples/MvcSandbox/package.json
+++ b/samples/MvcSandbox/package.json
@@ -1,4 +1,5 @@
 {
   "name": "ASP.NET",
+  "private":  true,
   "version": "0.0.0"
 }
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/IJsonHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/IJsonHelper.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.AspNet.Html.Abstractions;
 using Newtonsoft.Json;
 
 namespace Microsoft.AspNet.Mvc.Rendering
@@ -15,7 +16,7 @@ namespace Microsoft.AspNet.Mvc.Rendering
         /// </summary>
         /// <param name="value">The value to serialize as JSON.</param>
         /// <returns>A new <see cref="HtmlString"/> containing the serialized JSON.</returns>
-        HtmlString Serialize(object value);
+        IHtmlContent Serialize(object value);
 
         /// <summary>
         /// Returns serialized JSON for the <paramref name="value"/>.
@@ -23,6 +24,6 @@ namespace Microsoft.AspNet.Mvc.Rendering
         /// <param name="value">The value to serialize as JSON.</param>
         /// <param name="serializerSettings">The <see cref="JsonSerializerSettings"/> to be used by the serializer.</param>
         /// <returns>A new <see cref="HtmlString"/> containing the serialized JSON.</returns>
-        HtmlString Serialize(object value, JsonSerializerSettings serializerSettings);
+        IHtmlContent Serialize(object value, JsonSerializerSettings serializerSettings);
     }
 }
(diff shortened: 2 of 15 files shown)


dougbu:
This change gets rid of one long string per request, plus a few odds and ends. E.g. when using HTML helpers, including one `@Html.AntiForgeryToken()`... (Savings when using the `<form>` tag helper are similar.)

Before:
html 0

After:
html 1

samples/MvcSandbox/package.json
```diff
@@ -1,4 +1,5 @@
 {
   "name": "ASP.NET",
+  "private":  true,
```
dougbu: @ryanbrandenburg I didn't need to force this change into a commit. Why isn't the `.gitignore` file you added working as expected?

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/IJsonHelper.cs
```diff
@@ -15,7 +16,7 @@ public interface IJsonHelper
         /// </summary>
         /// <param name="value">The value to serialize as JSON.</param>
         /// <returns>A new <see cref="HtmlString"/> containing the serialized JSON.</returns>
```
pranavkm: Need to be `IHtmlContent`. Same for `Serialize`

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
```diff
@@ -125,15 +125,19 @@ public string FormatValue(object value, string format)
         }
 
         /// <inheritdoc />
-        public virtual IHtmlContent GenerateAntiforgery(ViewContext viewContext)
+        public virtual void GenerateAntiforgery(ViewContext viewContext, IHtmlContentBuilder content)
```
rynowak: This is part of why I think the antiforgery part of this change isn't ideal.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/HtmlHelper.cs
```diff
@@ -260,8 +260,10 @@ public virtual void Contextualize(ViewContext viewContext)
         /// <inheritdoc />
         public IHtmlContent AntiForgeryToken()
         {
-            var html = _htmlGenerator.GenerateAntiforgery(ViewContext);
-            return html ?? HtmlString.Empty;
+            var content = new HtmlContentBuilder();
+            _htmlGenerator.GenerateAntiforgery(ViewContext, content);
+
+            return content;
```
rynowak: This is part of why I think the antiforgery part of this change isn't ideal.

samples/MvcSandbox/package.json
ryanbrandenburg: Ahh, second line there: https://git-scm.com/docs/gitignore gitignore won't save you from changes to an existing file. Not sure what the correct thing is here, I'm not sure "--assume-unchanged" works across repos?

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/HtmlHelper.cs
dougbu: The tag helpers have a builder pattern -- get a `TagHelperOutput` and muck w/ it. The HTML helpers mostly have a factory pattern -- they return `IHtmlContent`s. Neither is ideal for `IHtmlGenerator` but that's not really relevant here since this method is unusual and can't return a `TagBuilder`.

In any case either pattern works. I prefer this pattern here because it doesn't force an unnecessary adapter in the tag helper case. Since it doesn't match the rest of `IHtmlGenerator` and you disagree, I'll switch.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/IJsonHelper.cs
dougbu: :ok:

dougbu:
:up::date:

pranavkm:
:shipit:

rynowak:
:shipit:

dougbu:
- d84587a for `package.json` nit
- 7bb0a1a for #3123 part 2 of 5 or so (`IJsonHelper`)
- aspnet/Antiforgery@6a9b38d and 6e299d6 for #3123 part 3 of 5 or so (`IAntiforgery`)

PR #3637, opened 2015-11-25
Use string.IsNullOrEmpty instead of string.IsNullOrWhitespace

### LINKED ISSUE #3593: ModelMetadata.ConvertEmptyStringToNull meaning in codebase
Hi, I saw that [simple type model binder](https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/ModelBinding/SimpleTypeModelBinder.cs#L39) use `ModelMetadata.ConvertEmptyStringToNull` to null out empty **and white space** string values, like this:
```C#
if (bindingContext.ModelMetadata.ConvertEmptyStringToNull &&
    string.IsNullOrWhiteSpace(modelAsString))
{
    model = null;
}
```

Is using `string.IsNullOrWhiteSpace(modelAsString)` correct here?
Because MVC5 nulls out empty values too, but strings containing only white space are left in tact.

pranavkm:
Fixes #3593

### DIFF
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
@@ -93,20 +93,25 @@ namespace Microsoft.AspNet.Mvc.TagHelpers.Internal
             var presentAttributes = new List<string>();
             var missingAttributes = new List<string>();
 
-            foreach (var attribute in requiredAttributes)
+            foreach (var requiredAttribute in requiredAttributes)
             {
-                if (!context.AllAttributes.ContainsName(attribute) ||
-                    context.AllAttributes[attribute] == null ||
-                    (context.AllAttributes[attribute].Value is string &&
-                    string.IsNullOrWhiteSpace(context.AllAttributes[attribute].Value as string)))
+                IReadOnlyTagHelperAttribute attribute;
+                if (!context.AllAttributes.TryGetAttribute(requiredAttribute, out attribute))
                 {
-                    // Missing attribute!
-                    missingAttributes.Add(attribute);
+                    // Missing attribute.
+                    missingAttributes.Add(requiredAttribute);
+                    continue;
                 }
-                else
+
+                var valueAsString = attribute.Value as string;
+                if (valueAsString != null && string.IsNullOrEmpty(valueAsString))
                 {
-                    presentAttributes.Add(attribute);
+                    // Treat minimized attributes or attributes with empty values as missing.
+                    missingAttributes.Add(requiredAttribute);
+                    continue;
                 }
+
+                presentAttributes.Add(requiredAttribute);
             }
 
             return new PresentMissingAttributes { Present = presentAttributes, Missing = missingAttributes };
--- a/src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
@@ -123,7 +123,7 @@ namespace Microsoft.AspNet.Mvc.Formatters
             if (contentType != null)
             {
                 var charset = contentType.Charset;
-                if (!string.IsNullOrWhiteSpace(charset))
+                if (!string.IsNullOrEmpty(charset))
                 {
                     foreach (var supportedEncoding in SupportedEncodings)
                     {
--- a/src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
@@ -254,7 +254,7 @@ namespace Microsoft.AspNet.Mvc.Formatters
                 for (var i = 0; i < acceptValues.Count; i++)
                 {
                     var charset = acceptValues[i].Value;
-                    if (!string.IsNullOrWhiteSpace(charset))
+                    if (!string.IsNullOrEmpty(charset))
                     {
                         for (var j = 0; j < SupportedEncodings.Count; j++)
                         {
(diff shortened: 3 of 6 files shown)


pranavkm:
There are two other places which use `string.IsNullOrWhitespace`:

1. ModelBindingHelper.ConvertSimpleType. It's used as a check before we call into `TypeConverter` to convert strings to primitives. Passing in a whitespace string here causes an array out of bounds exception from the `TypeConverter`.
2. `EnvironmentTagHelper` which has tests for tabs and spaces.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
```diff
@@ -93,20 +93,26 @@ public static class AttributeMatcher
             var presentAttributes = new List<string>();
             var missingAttributes = new List<string>();
 
-            foreach (var attribute in requiredAttributes)
+            foreach (var requiredAttribute in requiredAttributes)
             {
-                if (!context.AllAttributes.ContainsName(attribute) ||
```
pranavkm: This was kinda hard to read.

dougbu: Still messy. Also deserves comments about minimized attributes and `attributeName=""` being considered missing. (I don't remember why this works that way.)

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
```diff
@@ -93,20 +93,26 @@ public static class AttributeMatcher
...
-                    context.AllAttributes[attribute] == null ||
-                    (context.AllAttributes[attribute].Value is string &&
-                    string.IsNullOrWhiteSpace(context.AllAttributes[attribute].Value as string)))
+                IReadOnlyTagHelperAttribute attribute;
+                if (!context.AllAttributes.TryGetAttribute(requiredAttribute, out attribute) ||
+                    attribute == null)
```
dougbu: Can `attribute` be `null` when `TryGetAttribute` succeeds? Looks redundant but I may be forgetting something.

dougbu:
:watch: for questions

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
pranavkm: It used to check for nulls in the past.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
pranavkm: > Still messy

Do you have suggestions on what it should look like?

dougbu: Comments should help.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
dougbu: Right but wasn't that because `AllAttributes` returns `null` if the given `attribute` name isn't found?

dougbu: Put another way, none of the methods in `TagHelperExecutionContext` can add `null` to the source for that class' `AttAttributes` collection. And that's the source for this collection.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
Eilon: This change looks fine to me. What's the problem?

dougbu: My objections are mainly about the unexplained interpretation of `attributeName=""` as a "missing" attribute and (on the flip side) the lack of special handling for minimized attributes. One possible improvement would be to check `TryGetAttribute()` and call it a day. Another would be to find reasons for the other special cases and add comments explaining why they exist.

Eilon: Are you asking about some existing issue, or something related to the code change here?

pranavkm:
:up: :date:

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
dougbu: The code change made something more visible that existed before. Could certainly deal w/ in a follow-up issue. Good thought @Eilon.

Separately I remembered some of the history here: We were originally thinking about required attribute as a subset of bound attributes. The code definitely also pre-dated handling of minimized attributes. Enough has changed since for the checks (now at [line 107](https://github.com/aspnet/Mvc/pull/3637/files#diff-a8ba4397ef9cafc42bb0d852329baa66R107)) to be irrelevant and often-confusing esp. for attributes that are required but not bound.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
```diff
@@ -93,20 +93,25 @@ public static class AttributeMatcher
...
+                    missingAttributes.Add(requiredAttribute);
+                    continue;
                 }
-                else
+
+                var valueAsString = attribute.Value as string;
+                if (valueAsString != null && string.IsNullOrEmpty(valueAsString))
                 {
-                    presentAttributes.Add(attribute);
+                    // Treat minimized attributes or attributes with empty values as missing.
```
dougbu: I misspoke earlier and this does not handle minimized attributes -- they're treated as present. This is just about an empty string value. Put another way, the only `true` case is `string.Equals(attribute.Value, string,Empty)`.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
dougbu: @pranavkm could you please file the follow-up?

dougbu:
:shipit: w/ one corrected code comment

Issue #3636, opened 2015-11-25
Add extension methods for easily mapping area routes


rynowak:
### The problem of Areas

We're planning on removing *magic link generation* from MVC. This is the feature that causes MVC to only generate a link when we can guarantee that said URL will match an action. We're removing it because it's complicated and doesn't seem to add too much value - users are surprised by the behavior, and we don't want to make it throw.

However, we need to solve the problem that *magic link generation* solves. Namely generating links to areas....

#### Example 1:
```
app.UseMvc(routes =>
{
    routes.MapRoute("Admin/{controller}/{action}/{id?}", defaults: new { area = "Admin" });
    routes.MapRoute("{controller}/{action}/{id?}");
}

@Url.Action("Index", "Home"); // Generates /Admin/Home/Index
@Url.Action("Index", "Home", new { area = "Admin" }); // Generates /Admin/Home/Index
```

#### Example 2:
```
app.UseMvc(routes =>
{
    routes.MapRoute("{controller}/{action}/{id?}");
    routes.MapRoute("Admin/{controller}/{action}/{id?}", defaults: new { area = "Admin" });
}

@Url.Action("Index", "Home"); // Generates /Home/Index
@Url.Action("Index", "Home", new { area = "Admin" }); // Generates /Home/Index?area=Admin
```

In either of these cases, this is not desirable.. You can't use ordering to resolve the problem, because there's no ordering of routes that puts the 'most specific' first. MVC5 solved this problem by **manually filtering the route collection**. This is not a solution that we're willing to pursue in MVC6.

### Something that works

```
app.UseMvc(routes =>
{
    routes.MapRoute(
        "Admin/{controller}/{action}/{id?}",
        defaults: new { area = "Admin" },
        constraints: new { area = "Admin" });

    routes.MapRoute("{controller}/{action}/{id?}");
}

@Url.Action("Index", "Home"); // Generates /Home/Index
@Url.Action("Index", "Home", new { area = "Admin" }); // Generates /Admin/Home/Index
```

**Note that the area route needs to come first in this case** as it's more specific.

### Additionally

We want to provide something simpler that does this for you in the common case:

```
app.UseMvc(routes =>
{
    routes.MapAreaRoute("Admin", "Admin/{controller}/{action}/{id?}"});
    routes.MapRoute("{controller}/{action}/{id?}");
}

@Url.Action("Index", "Home"); // Generates /Home/Index
[...]

PR #3639, opened 2015-11-25
Remove Magic Link Generation

### LINKED ISSUE #3512: RC2 Linking to invalid controller/action stops request
Runtime: dnx-coreclr-darwin-x64.1.0.0-rc2-16128 (also impacts mono version)
Environment: OSX
MVC/Other deps : 6.0.0-rc2-16334 (6.0.0-*)

Previously if you added a `@Html.ActionLink("Home", "Index", "Home")` where the Controller/Action didnt exist, the link would be rendered empty without error.  With this update any links to invalid actions throws the following error:
```
fail: Microsoft.AspNet.Diagnostics.DeveloperExceptionPageMiddleware[0]
      An unhandled exception has occurred while executing the request
      System.NullReferenceException: Object reference not set to an instance of an object
        at System.Text.Encodings.Web.TextEncoder.Encode (System.IO.TextWriter output, System.String value) <0x10e5b2380 + 0x00018> in <filename unknown>:0 
        at Microsoft.AspNet.Mvc.Rendering.TagBuilder.AppendAttributes (System.IO.TextWriter writer, System.Text.Encodings.Web.HtmlEncoder encoder) <0x10e5d7980 + 0x001a1> in <filename unknown>:0 
        at Microsoft.AspNet.Mvc.Rendering.TagBuilder.WriteTo (System.IO.TextWriter writer, System.Text.Encodings.Web.HtmlEncoder encoder) <0x10e5d7580 + 0x00179> in <filename unknown>:0 
        at Microsoft.AspNet.Mvc.ViewFeatures.StringCollectionTextWriter+StringCollectionTextWriterContent.WriteTo (System.IO.TextWriter writer, System.Text.Encodings.Web.HtmlEncoder encoder) <0x10e5b6f10 + 0x001c0> in <filename unknown>:0 
        at Microsoft.AspNet.Mvc.ViewFeatures.StringCollectionTextWriter.CopyTo (System.IO.TextWriter writer, System.Text.Encodings.Web.HtmlEncoder encoder) <0x10e5c99f0 + 0x001b3> in <filename unknown>:0 
        at Microsoft.AspNet.Mvc.ViewFeatures.StringCollectionTextWriter.CopyToAsync (System.IO.TextWriter writer, System.Text.Encodings.Web.HtmlEncoder encoder) <0x10e5c99b0 + 0x0001e> in <filename unknown>:0 
        at (wrapper remoting-invoke-with-check) Microsoft.AspNet.Mvc.ViewFeatures.StringCollectionTextWriter:CopyToAsync (System.IO.TextWriter,System.Text.Encodings.Web.HtmlEncoder)
        at Microsoft.AspNet.Mvc.Razor.RazorTextWriter.CopyToAsync (System.IO.TextWriter writer) <0x10e5c9680 + 0x00068> in <filename unknown>:0 
        at Microsoft.AspNet.Mvc.Razor.RazorView+<RenderLayoutAsync>d__20.MoveNext () <0x10e512000 + 0x008d7> in <filename unknown>:0 
...
```

This error interupts the page rather than totally fails it, so the browser gets a `Net::ERR_INCOMPLETE_CHUNKED_ENCODING` (chrome) with a partially rendered page.

rynowak:
This change resolves #3512 and #3636 by removing 'magic' link generation
and adding an extension method to add routes to areas correctly using the new
pattern. This is pretty much exactly the same as how MapWebApiRoute works.

For site authors, we recommend adding area-specific routes in a way that
includes a default AND constraint for the area. Put your most specific
(for link generation) routes FIRST.

Ex:
```
  routes.MapRoute(
     "admin_route"
      "Admin/{controller}/{action}/{id?}",
      defaults: new { area = "Admin" },
      constraints: new { area = "Admin" });
```

We're providing a new extension method that does this for you.
```
  routes.MapAreaRoute("admin_route", "Admin", "Admin/{controller}/{action}/{id?}");
```

The bulk of the changes here are to tests that unwittingly relied on the
old behavior.


src/Microsoft.AspNet.Mvc.Core/Infrastructure/MvcRouteHandler.cs
```diff
@@ -32,11 +32,7 @@ public VirtualPathData GetVirtualPath(VirtualPathContext context)
                 throw new ArgumentNullException(nameof(context));
             }
 
-            EnsureServices(context.Context);
-
-            // The contract of this method is to check that the values coming in from the route are valid;
-            // that they match an existing action, setting IsBound = true if the values are OK.
-            context.IsBound = _actionSelector.HasValidAction(context);
+            context.IsBound = true;
```
rynowak: This feature will be removed wholesale in a future routing change.

test/Microsoft.AspNet.Mvc.FunctionalTests/InlineConstraintTests.cs
```diff
@@ -408,7 +408,8 @@ public static IEnumerable<object[]> QueryParameters
                         "InlineConstraints_Products",
                         "GetProductById",
                         "id",
-                        "sdsd", ""
+                        "sdsd",
+                        "/area-exists/InlineConstraints_Products/GetProductById?id=sdsd"
```
rynowak: This route is now "greedy". I don't see much reason to try and rationalize this behavior, these tests still test what they did before - that the constraint in question doesn't allow that route to match with invalid values.

test/Microsoft.AspNet.Mvc.FunctionalTests/RoutingTests.cs
```diff
@@ -903,7 +903,7 @@ public RoutingTests(MvcTestFixture<RoutingWebSite.Startup> fixture)
         }
 
         [Fact]
-        public async Task ConventionalRoutedAction_InArea_ImplicitLeaveArea()
+        public async Task ConventionalRoutedAction_InArea_StaysInArea()
```
rynowak: Intentional behavior change.

test/Microsoft.AspNet.Mvc.FunctionalTests/RoutingTests.cs
```diff
@@ -985,7 +985,7 @@ public RoutingTests(MvcTestFixture<RoutingWebSite.Startup> fixture)
         }
 
         [Fact]
-        public async Task AttributeRoutedAction_InArea_ImplicitLeaveArea()
+        public async Task AttributeRoutedAction_InArea_StaysInArea_ActionDoesntExist()
```
rynowak: Intentional behavior change.

src/Microsoft.AspNet.Mvc.Core/Infrastructure/MvcRouteHandler.cs
```diff
@@ -32,11 +32,7 @@ public VirtualPathData GetVirtualPath(VirtualPathContext context)
                 throw new ArgumentNullException(nameof(context));
             }
 
-            EnsureServices(context.Context);
```
Eilon: Is this gone because this was needed for action selection, which is no longer used?

rynowak: Correct.

test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.Index.Encoded.html
```diff
@@ -29,7 +29,7 @@
         </a>
     </div>
     <div>
-        <a href="">
+        <a href="HtmlEncode[[unkonwn://localhost/HtmlGeneration_Product/List]]">
```
Eilon: Is this a konwn unkonwn or an unkonwn unknown?

test/WebSites/HtmlGenerationWebSite/Views/HtmlGeneration_Home/Index.cshtml
```diff
@@ -35,7 +35,7 @@
         </a>
     </div>
     <div>
-        <a asp-controller="Product" asp-protocol="unkonwn" asp-action="List">
+        <a asp-controller="HtmlGeneration_Product" asp-protocol="unkonwn" asp-action="List">
```
Eilon: Also unknown.

test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.Index.Encoded.html
rynowak: You go to war with the unkonwn you have

src/Microsoft.AspNet.Mvc.Core/Builder/RouteBuilderExtensions.cs
```diff
@@ -0,0 +1,136 @@
...
+
+using System;
+using Microsoft.AspNet.Routing;
+
+namespace Microsoft.AspNet.Builder
+{
+    /// <summary>
+    /// Extension methods for <see cref="IRouteBuilder"/>.
+    /// </summary>
+    public static class RouteBuilderExtensions
```
Eilon: Would like to rename this to something like `MvcAreaRouteBuilderExtensions` so it's pretty specific (because it's in a generic namespace). Or can remove `Area` from the name if we might add more non-area stuff here.

rynowak: sure

Eilon:
:shipit: but yeah some of those tests look wonky to me. I agree it's fine to leave because they do capture some interesting behaviors.

rynowak:
> some of those tests look wonky to me

A lot of these tests really only touch this feature by accident. That's the trouble with these big html-generation/baseline style tests. We ran into similar issues when we started fixing bugs in modelbinding. Tests for unrelated features include bad behavior and verified bad data :+1:

Eilon:
Yup.

src/Microsoft.AspNet.Mvc.Core/Builder/RouteBuilderExtensions.cs
```diff
@@ -0,0 +1,136 @@
...
+    {
+        /// <summary>
+        /// Adds a route to the <see cref="IRouteBuilder"/> with the given MVC area with the specified
+        /// name and template.
```
dougbu: Suggest:
- The `<summary>`s should vary to list all that's specified.
- Should use `<paramref>`.

So here `... specified <paramref name="name"/>, <paramref name="areaName"/> and <paramref name="template">.`

Separately did you intend to create an overload that only had `areaName` and `template` parameters i.e. used a default route `name`?

src/Microsoft.AspNet.Mvc.Core/Builder/RouteBuilderExtensions.cs
```diff
@@ -0,0 +1,136 @@
...
+                defaultsDictionary.Add("area", areaName);
+            }
+
+            var constraintsDictionary = new RouteValueDictionary(constraints);
+            if (!constraintsDictionary.ContainsKey("area"))
+            {
+                constraintsDictionary.Add("area", areaName);
+            }
+
+            return routeBuilder;
```
dougbu: Something is missing here: Not using the `routeBuilder`, created dictionaries, or the remaining parameters.

src/Microsoft.AspNet.Mvc.Core/Builder/RouteBuilderExtensions.cs
```diff
@@ -0,0 +1,136 @@
...
+            object defaults,
+            object constraints,
+            object dataTokens)
+        {
+            if (routeBuilder == null)
+            {
+                throw new ArgumentNullException(nameof(routeBuilder));
+            }
+
+            var defaultsDictionary = new RouteValueDictionary(defaults);
```
dougbu: Why no `null` or `string.IsNullOrEmpty()` check of `name`?

src/Microsoft.AspNet.Mvc.Core/Builder/RouteBuilderExtensions.cs
```diff
@@ -0,0 +1,136 @@
...
+    /// </summary>
+    public static class RouteBuilderExtensions
+    {
+        /// <summary>
+        /// Adds a route to the <see cref="IRouteBuilder"/> with the given MVC area with the specified
+        /// name and template.
+        /// </summary>
+        /// <param name="routeBuilder">The <see cref="IRouteBuilder"/> to add the route to.</param>
+        /// <param name="name">The name of the route.</param>
+        /// <param name="areaName">The MVC area name.</param>
```
dougbu: I seem to remember `null` and empty area names have specific meanings. If so, would be useful to mention the semantics in these descriptions.

test/Microsoft.AspNet.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.Index.Encoded.html
```diff
@@ -39,32 +39,32 @@
...
     <div>
-        <a href="">Non-existent Action</a>
+        <a href="HtmlEncode[[/Order/NonExistentAction]]">Non-existent Action</a>
     </div>
     <div>
         <a id="Id" href="HtmlEncode[[http://somewhere/]]">Some Where</a>
     </div>
     <div>
-        <a href="">
+        <a href="HtmlEncode[[unknown://localhost/NoControll#fragment]]">
```
dougbu: Disappointed about the inconsistency in the unkonwn you have.

src/Microsoft.AspNet.Mvc.Core/Builder/RouteBuilderExtensions.cs
dougbu: Also suggest using these methods in the functional test sites or unit tests. Ignored in all tests at the moment.

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.Core/Builder/RouteBuilderExtensions.cs
rynowak: > Separately did you intend to create an overload that only had areaName and template parameters i.e. used a default route name?

No, `MapRoute` always has a positional parameter for route name, even though it's not required.

rynowak: Added these details to the docs. Agreed that it seems dumb to list name, areaName and template, but not other things.

src/Microsoft.AspNet.Mvc.Core/Builder/RouteBuilderExtensions.cs
rynowak: :+1:

src/Microsoft.AspNet.Mvc.Core/Builder/RouteBuilderExtensions.cs
rynowak: It's allowed to be null or empty.

rynowak: We should check `areaName` though.

src/Microsoft.AspNet.Mvc.Core/Builder/RouteBuilderExtensions.cs
rynowak: It only has a meaning in link generation. You can't define a null or empty area.

rynowak:
Updated and added some unit tests. There aren't any good places I could find to use these in functional tests, because our tests all use the 'new' style `{area}/{controller}...` route.

test/WebSites/HtmlGenerationWebSite/Views/HtmlGeneration_Home/Index.cshtml
dougbu: The source of all that's unkonwn.

test/Microsoft.AspNet.Mvc.Core.Test/Builder/MvcAreaRouteBuilderExtensionsTest.cs
```diff
@@ -0,0 +1,258 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Moq;
+using Xunit;
+
+namespace Microsoft.AspNet.Builder
+{
+    public class MvcAreaRouteBuilderExtensionsTest
+    {
+        [Fact]
+        public void MapAreaRoute_Simple()
```
dougbu: Shouldn't the test name say something about expectations e.g. `MapAreaRoute_Simple_DoesTheNeedful`?

dougbu:
:shipit: after correcting your spelling.

test/Microsoft.AspNet.Mvc.Core.Test/Builder/MvcAreaRouteBuilderExtensionsTest.cs
rynowak: > if you have nothing to say say nothing

All of these tests verify the most obvious possible thing. I don't see any reason to make the names more verbose.

PR #3642, opened 2015-11-25
Remove IHttpContextAccessor from TempData

rynowak:
This change removes the dependency of TempData on the IHttpContextAccessor
by creating an ITempDataDictionaryFactory abstraction. In general, no one
will replace the factory, it's just indirection.

This allows us to drop our dependency on IHttpContextAccessor, and move it
to the functional tests where we specifically depend on it.

The bulk of code churn here is to update tests that use TempData.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/DependencyInjection/MvcViewFeaturesMvcCoreBuilderExtensions.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/DependencyInjection/MvcViewFeaturesMvcCoreBuilderExtensions.cs
@@ -133,12 +133,10 @@ namespace Microsoft.Extensions.DependencyInjection
             //
             // Temp Data
             //
-            // Holds per-request data so it should be scoped
-            services.TryAddScoped<ITempDataDictionary, TempDataDictionary>();
-            services.TryAddScoped<SaveTempDataFilter>();
-
             // This does caching so it should stay singleton
             services.TryAddSingleton<ITempDataProvider, SessionStateTempDataProvider>();
+            services.TryAddSingleton<ITempDataDictionaryFactory, TempDataDictionaryFactory>();
+            services.TryAddSingleton<SaveTempDataFilter>();
         }
     }
 }
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/ITempDataDictionaryFactory.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/ITempDataDictionaryFactory.cs
@@ -13,12 +13,12 @@ namespace Microsoft.AspNet.Mvc.ViewFeatures
     {
         /// <summary>
         /// Gets or creates an <see cref="ITempDataDictionary"/> instance for the request associated with the
-        /// given <see cref="HttpContext"/>.
+        /// given <paramref name="context"/>.
         /// </summary>
         /// <param name="context">The <see cref="HttpContext"/>.</param>
         /// <returns>
         /// An <see cref="ITempDataDictionary"/> instance for the request associated with the given
-        /// <see cref="HttpContext"/>.
+        /// <paramref name="context"/>.
         /// </returns>
         ITempDataDictionary GetTempData(HttpContext context);
     }
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/ITempDataDictionaryFactory.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/ITempDataDictionaryFactory.cs
@@ -1,0 +1,25 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNet.Http;
+
+namespace Microsoft.AspNet.Mvc.ViewFeatures
+{
+    /// <summary>
+    /// A factory which provides access to an <see cref="ITempDataDictionary"/> instance
+    /// for a request.
+    /// </summary>
+    public interface ITempDataDictionaryFactory
+    {
+        /// <summary>
+        /// Gets or creates an <see cref="ITempDataDictionary"/> instance for the request associated with the
+        /// given <see cref="HttpContext"/>.
+        /// </summary>
+        /// <param name="context">The <see cref="HttpContext"/>.</param>
+        /// <returns>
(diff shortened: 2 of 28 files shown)


rynowak:
@davidfowl each your :heart: out

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/TempDataDictionaryFactory.cs
```diff
@@ -0,0 +1,55 @@
...
+            {
+                throw new ArgumentNullException(nameof(provider));
+            }
+
+            _provider = provider;
+        }
+
+        /// <inheritdoc />
+        public ITempDataDictionary GetTempData(HttpContext context)
+        {
```
davidfowl: Is it really a factory if it's get not create?

davidfowl:
I like it. So we generally have a few ways to do "per request storage":

1. Context.items
2. Scoped service
3. Async local state

We should write up a summary of the pros and cons of each and check to see where we have other dependencies like that.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/TempDataDictionaryFactory.cs
rynowak: I think so. We already have `ITempDataProvider` or I would have called it that.

rynowak:
4. Feature Collection

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/ITempDataDictionaryFactory.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    /// <summary>
+    /// A factory which provides access to an <see cref="ITempDataDictionary"/> instance
+    /// for a request.
+    /// </summary>
+    public interface ITempDataDictionaryFactory
+    {
+        /// <summary>
+        /// Gets or creates an <see cref="ITempDataDictionary"/> instance for the request associated with the
+        /// given <see cref="HttpContext"/>.
```
dougbu: Our doc comments normally refer to given or specified `<paramref/>`s and leave type information to the `<param/>` descriptions. Here that would be `... given <paramref name="context"/>.`

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/TempDataDictionaryFactory.cs
```diff
@@ -0,0 +1,55 @@
...
+using Microsoft.AspNet.Http;
+
+namespace Microsoft.AspNet.Mvc.ViewFeatures
+{
+    /// <summary>
+    /// A default implementation of <see cref="ITempDataDictionaryFactory"/>.
+    /// </summary>
+    public class TempDataDictionaryFactory : ITempDataDictionaryFactory
+    {
+        private static readonly string Key = typeof(ITempDataDictionary).FullName;
```
dougbu: Why not use `typeof(ITempDataDictionary)` as the key?

test/WebSites/RequestServicesWebSite/Startup.cs
```diff
@@ -14,6 +16,7 @@ public void ConfigureServices(IServiceCollection services)
             services.AddMvc();
 
             services.AddScoped<RequestIdService>();
+            services.AddSingleton<IHttpContextAccessor, HttpContextAccessor>();
```
dougbu: Is it the plan to remove `IHttpContextAccessor` entirely? Or will the Hosting repo (for example) register it for customer use?

dougbu:
:watch: mostly for questions

test/WebSites/RequestServicesWebSite/Startup.cs
davidfowl: User code can add it if needed

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/ITempDataDictionaryFactory.cs
rynowak: This is news to me.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/TempDataDictionaryFactory.cs
rynowak: Sure :+1:

test/WebSites/RequestServicesWebSite/Startup.cs
rynowak: We're going to remove our dependency on it by default. The user can add it back if desired.

rynowak:
Updated

src/Microsoft.AspNet.Mvc.ViewFeatures/DependencyInjection/MvcViewFeaturesMvcCoreBuilderExtensions.cs
```diff
@@ -133,12 +133,10 @@ internal static void AddViewServices(IServiceCollection services)
...
             // Temp Data
             //
-            // Holds per-request data so it should be scoped
-            services.TryAddScoped<ITempDataDictionary, TempDataDictionary>();
-            services.TryAddScoped<SaveTempDataFilter>();
-
             // This does caching so it should stay singleton
             services.TryAddSingleton<ITempDataProvider, SessionStateTempDataProvider>();
+            services.TryAddSingleton<ITempDataDictionaryFactory, TempDataDictionaryFactory>();
+            services.TryAddSingleton<SaveTempDataFilter>();
```
dougbu: Comment reads like it might apply to these two lines. Suggest an additional blank line to help. Or add a comment saying these two depend on `ITempDataProvider` directly or indirectly.

dougbu:
:shipit: (but use fewer `if`s :trollface:)

src/Microsoft.AspNet.Mvc.ViewFeatures/DependencyInjection/MvcViewFeaturesMvcCoreBuilderExtensions.cs
rynowak: It says 'this' not 'these'

rynowak: I'll add a comment saying that those other two have no particular hopes or dreams.

PR #3644, opened 2015-11-25
Remove FormatFilter dependency on IActionContextAccessor

rynowak:
This change simplifies IFormatFilter's API and removes the dependency on
IActionContext accessor.

The old API for IFormatFilter required computing state based on the
current request as part of the constructor, which in turn implied the use
of a context accessor. This isn't really needed. I didn't preserve caching of
the 'format' as that seems like an early optimization.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/DependencyInjection/MvcCoreMvcCoreBuilderExtensions.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/DependencyInjection/MvcCoreMvcCoreBuilderExtensions.cs
@@ -62,7 +62,7 @@ namespace Microsoft.Extensions.DependencyInjection
         // Internal for testing.
         internal static void AddFormatterMappingsServices(IServiceCollection services)
         {
-            services.TryAddTransient<FormatFilter, FormatFilter>();
+            services.TryAddSingleton<FormatFilter, FormatFilter>();
         }
 
         public static IMvcCoreBuilder AddAuthorization(this IMvcCoreBuilder builder)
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNet.Mvc.Core/Formatters/FormatFilter.cs
```diff
@@ -65,13 +61,15 @@ public void OnResourceExecuting(ResourceExecutingContext context)
...
-            if (!IsActive)
+            var format = GetFormat(context);
+            if (format == null)
             {
                 // no format specified by user, so the filter is muted
                 return;
             }
 
-            if (ContentType == null)
+            var contentType = _options.FormatterMappings.GetMediaTypeMappingForFormat(format);
```
rynowak: Note the change here `ContentType` -> `contentType`, which is why the lambda below might make you do a double-take.

src/Microsoft.AspNet.Mvc.Core/DependencyInjection/MvcCoreMvcCoreBuilderExtensions.cs
```diff
@@ -62,7 +62,7 @@ public static IMvcCoreBuilder AddFormatterMappings(this IMvcCoreBuilder builder)
         // Internal for testing.
         internal static void AddFormatterMappingsServices(IServiceCollection services)
         {
-            services.TryAddTransient<FormatFilter, FormatFilter>();
+            services.TryAddSingleton<FormatFilter, FormatFilter>();
```
Eilon: Curious - why is this service not registered as the `IFormatFilter` interface? #n00b

rynowak: `FormatFilter` is a service that is an implementation detail of `FormatFilterAttribute`. `IFormatFilter` is for items that are contributors to the 'format filter' protocol.

Eilon:
:shipit:

PR #3645, opened 2015-11-25
Adds ActionContext to OperationBindingContext

rynowak:
This change makes it possible to access the ActionContext/ActionDescriptor
from inside of modelbinding.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/OperationBindingContext.cs
+++ b/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/OperationBindingContext.cs
@@ -15,9 +15,14 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
     public class OperationBindingContext
     {
         /// <summary>
-        /// Gets or sets the <see cref="HttpContext"/> for the current request.
+        /// Gets or sets the <see cref="Mvc.ActionContext"/> for the current request.
         /// </summary>
-        public HttpContext HttpContext { get; set; }
+        public ActionContext ActionContext { get; set; }
+
+        /// <summary>
+        /// Gets the <see cref="Http.HttpContext"/> for the current request.
+        /// </summary>
+        public HttpContext HttpContext => ActionContext.HttpContext;
 
         /// <summary>
         /// Gets or sets the set of <see cref="IInputFormatter"/> instances associated with this context.
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/ModelBindingHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/ModelBindingHelper.cs
@@ -11,7 +11,6 @@ using System.Linq.Expressions;
 using System.Reflection;
 using System.Runtime.ExceptionServices;
 using System.Threading.Tasks;
-using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Mvc.Core;
 using Microsoft.AspNet.Mvc.Formatters;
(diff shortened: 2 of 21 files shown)


src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/OperationBindingContext.cs
```diff
@@ -15,9 +15,14 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
...
     {
         /// <summary>
-        /// Gets or sets the <see cref="HttpContext"/> for the current request.
+        /// Gets or sets the <see cref="Mvc.ActionContext"/> for the current request.
         /// </summary>
-        public HttpContext HttpContext { get; set; }
+        public ActionContext ActionContext { get; set; }
+
+        /// <summary>
+        /// Gets the <see cref="HttpContext"/> for the current request.
```
dougbu: Is `HttpContext` ambiguous like `ActionContext` is above?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/ModelBindingHelper.cs
```diff
@@ -29,7 +28,7 @@ public static class ModelBindingHelper
         /// <param name="model">The model instance to update and validate.</param>
         /// <param name="prefix">The prefix to use when looking up values in the <paramref name="valueProvider"/>.
         /// </param>
-        /// <param name="httpContext">The <see cref="HttpContext"/> for the current executing request.</param>
+        /// <param name="actionContext">The <see cref="ActionContext"/> for the current executing request.</param>
```
dougbu: While here please add 3 more `TryUpdateModelAsync()` overloads :trollface:

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/OperationBindingContext.cs
rynowak: Yeah, I bet you're right. I'll fix that too. It's something I tend to notice when it autocompletes but isn't the right colour.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/ModelBindingHelperTest.cs
```diff
@@ -46,7 +47,7 @@ public static TheoryData<ModelBindingResult> UnsuccessfulModelBindingData
             var result = await ModelBindingHelper.TryUpdateModelAsync(
                 model,
                 string.Empty,
-                Mock.Of<HttpContext>(),
+                new ActionContext() { HttpContext = new DefaultHttpContext() },
```
dougbu: With this change in, will we have any remaining `Mock<HttpContext>`s? If yes, why?

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/OperationBindingContext.cs
dougbu: Thanks very much for spelling colour correctly :maple_leaf:

dougbu:
:shipit:

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/ModelBindingHelperTest.cs
rynowak: There are mock model binders in this class. Lots of `Mock<HttpContext>` in other places

Issue #3599, opened 2015-11-21
Do a pass for Linq and boxed enumerators in TagHelpers


NTaylorMullen:
_From @rynowak on November 12, 2015 0:44_

We should just do a general pass of TagHelpers code for Linq use and boxing foreach'es

![image](https://cloud.githubusercontent.com/assets/1430011/11107507/79d5be08-8893-11e5-87d2-73c15e182588.png)

Some of these are coming from MVC/`HtmlGenerator` as well.

_Copied from original issue: aspnet/Razor#605_

NTaylorMullen:
490fcf1ab4d416ee6cc140797c2377e352677bda

Issue #3617, opened 2015-11-23
Remove use of `String.Split` in Environment taghelper


rynowak:
![image](https://cloud.githubusercontent.com/assets/1430011/11342562/d75775f8-91bc-11e5-8717-5d40e81d6b4f.png)

Use `StringSegment` or similar

PR #3646, opened 2015-11-25
Remove the use of `Linq` and unneeded enumerators from Mvc `TagHelper`s.

NTaylorMullen:
- Prior to this change we were using `Linq` throughout our `TagHelper`s which resulted excess allocations.
- Also went through our data structures and changed some from `Enumerable<T>` to `IList<T>` where we were just not exposing the more accessible types (list).
- Changed several locations of `foreach` to `for` to remove the enumerator allocation.

**Before:**
image

**After:**
image

**Result:** Roughly a 2.5% reduction in overall allocations.

#3599

### DIFF
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/CacheTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/CacheTagHelper.cs
@@ -3,11 +3,9 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Security.Cryptography;
 using System.Text;
 using System.Threading.Tasks;
-using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Mvc.Rendering;
 using Microsoft.AspNet.Mvc.ViewFeatures;
 using Microsoft.AspNet.Razor.TagHelpers;
@@ -101,7 +99,7 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
 
         /// <summary>
         /// Gets or sets a value that determines if the cached result is to be varied by the Identity for the logged in
-        /// <see cref="HttpContext.User"/>.
+        /// <see cref="Http.HttpContext.User"/>.
         /// </summary>
         [HtmlAttributeName(VaryByUserAttributeName)]
         public bool VaryByUser { get; set; }
@@ -261,18 +259,19 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
                        .Append(keyName)
                        .Append("(");
 
-                var tokenFound = false;
-                foreach (var item in Tokenize(value))
-                {
-                    tokenFound = true;
+                var values = Tokenize(value);
 
+                // Perf: Avoid allocating enumerator
+                for (var i = 0; i < values.Count; i++)
+                {
+                    var item = values[i];
                     builder.Append(item)
                            .Append(CacheKeyTokenSeparator)
                            .Append(sourceCollection[item])
                            .Append(CacheKeyTokenSeparator);
                 }
 
-                if (tokenFound)
+                if (values.Count > 0)
                 {
                     // Remove the trailing separator
                     builder.Length -= CacheKeyTokenSeparator.Length;
@@ -282,12 +281,12 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
             }
         }
 
-        private static void AddStringCollectionKey<T>(
+        private static void AddStringCollectionKey<TSourceCollection>(
             StringBuilder builder,
             string keyName,
             string value,
-            T sourceCollection,
-            Func<T, string, StringValues> accessor)
+            TSourceCollection sourceCollection,
+            Func<TSourceCollection, string, StringValues> accessor)
         {
             if (!string.IsNullOrEmpty(value))
             {
@@ -296,10 +295,12 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
                        .Append(keyName)
                        .Append("(");
 
-                var tokenFound = false;
-                foreach (var item in Tokenize(value))
+                var values = Tokenize(value);
+
+                // Perf: Avoid allocating enumerator
+                for (var i = 0; i < values.Count; i++)
                 {
-                    tokenFound = true;
+                    var item = values[i];
 
                     builder.Append(item)
                            .Append(CacheKeyTokenSeparator)
@@ -307,7 +308,7 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
                            .Append(CacheKeyTokenSeparator);
                 }
 
-                if (tokenFound)
+                if (values.Count > 0)
                 {
                     // Remove the trailing separator
                     builder.Length -= CacheKeyTokenSeparator.Length;
@@ -327,8 +328,10 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
                        .Append(nameof(VaryByRoute))
                        .Append("(");
 
-                foreach (var route in Tokenize(VaryByRoute))
+                var varyByRoutes = Tokenize(VaryByRoute);
+                for (var i = 0; i < varyByRoutes.Count; i++)
                 {
+                    var route = varyByRoutes[i];
                     tokenFound = true;
 
                     builder.Append(route)
@@ -346,11 +349,27 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
             }
         }
 
-        private static IEnumerable<string> Tokenize(string value)
+        private static IList<string> Tokenize(string value)
         {
-            return value.Split(AttributeSeparator, StringSplitOptions.RemoveEmptyEntries)
-                        .Select(token => token.Trim())
-                        .Where(token => token.Length > 0);
+            var values = value.Split(AttributeSeparator, StringSplitOptions.RemoveEmptyEntries);
+            if (values.Length == 0)
+            {
+                return values;
+            }
+
+            var trimmedValues = new List<string>();
+
+            for (var i = 0; i < values.Length; i++)
+            {
+                var trimmedValue = values[i].Trim();
+
+                if (trimmedValue.Length > 0)
+                {
+                    trimmedValues.Add(trimmedValue);
+                }
+            }
+
+            return trimmedValues;
         }
     }
 }
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/EnvironmentTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/EnvironmentTagHelper.cs
@@ -2,7 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Linq;
+using System.Collections.Generic;
 using Microsoft.AspNet.Hosting;
 using Microsoft.AspNet.Razor.TagHelpers;
 
@@ -67,28 +67,39 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
                 return;
             }
 
-            var environments = Names.Split(NameSeparator, StringSplitOptions.RemoveEmptyEntries)
-                                    .Where(name => !string.IsNullOrWhiteSpace(name));
-
-            if (!environments.Any())
-            {
-                // Names contains only commas or empty entries, do nothing
-                return;
-            }
-
             var currentEnvironmentName = HostingEnvironment.EnvironmentName?.Trim();
-
-            if (string.IsNullOrWhiteSpace(currentEnvironmentName))
+            if (string.IsNullOrEmpty(currentEnvironmentName))
             {
                 // No current environment name, do nothing
                 return;
             }
 
-            if (environments.Any(name =>
-                string.Equals(name.Trim(), currentEnvironmentName, StringComparison.OrdinalIgnoreCase)))
+            var values = Names.Split(NameSeparator, StringSplitOptions.RemoveEmptyEntries);
+            var environments = new List<string>();
+
+            for (var i = 0; i < values.Length; i++)
             {
-                // Matching environment name found, do nothing
+                var trimmedValue = values[i].Trim();
+
+                if (trimmedValue.Length > 0)
+                {
+                    environments.Add(trimmedValue);
+                }
+            }
+
+            if (environments.Count == 0)
+            {
+                // Names contains only commas or empty entries, do nothing
                 return;
+            }
+
+            for (var i = 0; i < environments.Count; i++)
+            {
+                if (string.Equals(environments[i], currentEnvironmentName, StringComparison.OrdinalIgnoreCase))
+                {
+                    // Matching environment name found, do nothing
+                    return;
+                }
             }
 
             // No matching environment name found, suppress all output
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Mvc.ModelBinding;
 using Microsoft.AspNet.Mvc.Rendering;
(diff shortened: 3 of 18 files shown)


src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
```diff
@@ -74,7 +74,7 @@ public GlobbingUrlBuilder(IFileProvider fileProvider, IMemoryCache cache, PathSt
         /// <param name="includePattern">The file globbing include pattern.</param>
         /// <param name="excludePattern">The file globbing exclude pattern.</param>
         /// <returns>The list of URLs</returns>
-        public virtual IEnumerable<string> BuildUrlList(string staticUrl, string includePattern, string excludePattern)
+        public virtual ICollection<string> BuildUrlList(string staticUrl, string includePattern, string excludePattern)
```
NTaylorMullen: Only reason going `ICollection<string>` here is to avoid `.Any()`. This is technically an `Internal` type though so I thought it was warranted.

rynowak: ewww `ICollection<T>` :stuck_out_tongue_closed_eyes: - I think it's fine for a public API, just don't be surprised if you change it again later.

Do you then do a `foreach` over the collection?

NTaylorMullen: Yup :cry:

src/Microsoft.AspNet.Mvc.TagHelpers/CacheTagHelper.cs
```diff
@@ -346,11 +351,22 @@ private void AddVaryByRouteKey(StringBuilder builder)
             }
         }
 
-        private static IEnumerable<string> Tokenize(string value)
+        private static IList<string> Tokenize(string value)
         {
-            return value.Split(AttributeSeparator, StringSplitOptions.RemoveEmptyEntries)
-                        .Select(token => token.Trim())
-                        .Where(token => token.Length > 0);
+            var values = value.Split(AttributeSeparator, StringSplitOptions.RemoveEmptyEntries);
```
rynowak: Does this happen every request? We really don't want to do `string.Split`

Can you log an issue to follow up on this guy? splitting and trimming a comma-separated list seems like it should be first class and handled by codegen.

src/Microsoft.AspNet.Mvc.TagHelpers/EnvironmentTagHelper.cs
```diff
@@ -67,25 +67,33 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
...
+
+            if (environments.Count == 0)
+            {
+                // Names contains only commas or empty entries, do nothing
                 return;
             }
 
-            if (environments.Any(name =>
-                string.Equals(name.Trim(), currentEnvironmentName, StringComparison.OrdinalIgnoreCase)))
+            if (environments.Contains(currentEnvironmentName))
```
rynowak: This could be much easier done with a tokenizer.

basically `string.IndexOf(currentEnvironmentName)` and then check the surrounding characters for whitespace/comma

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
```diff
@@ -113,9 +113,11 @@ private IEnumerable<string> ExpandGlobbedUrl(string include, string exclude)
...
                 {
                     var options = new MemoryCacheEntryOptions();
-                    foreach (var pattern in includePatterns)
+
+                    // Perf: Avoid allocating enumerator
+                    for (var i = 0; i < includePatterns.Length; i++)
                     {
-                        var changeToken = FileProvider.Watch(pattern);
+                        var changeToken = FileProvider.Watch(includePatterns[i]);
                         options.AddExpirationToken(changeToken);
```
rynowak: Is this really perf-sensitive code?

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/ModeMatchResult.cs
```diff
@@ -57,17 +55,39 @@ public class ModeMatchResult<TMode>
...
+
+                        if (presentIsPartialOnlyMatch)
+                        {
+                            partialOnlyMatches.Add(PartialMatches[i]);
+                            break;
+                        }
+                    }
+                }
+
+                logger.TagHelperPartialMatches(uniqueId, viewPath, partialOnlyMatches);
```
rynowak: Can we move all of this ugliness into the extension method?

rynowak:
BTW - say you're working on like the script tag. It seems totally fine to just leave a comment in one place saying that we're optimizing **all** of that code to reduce allocs, rather than copy-pasting the same comment 20 times in the body.

Follow your :heart:

rynowak:
:watch:

src/Microsoft.AspNet.Mvc.TagHelpers/CacheTagHelper.cs
NTaylorMullen: Sadly, yes it does. As for an issue, already had you covered: https://github.com/aspnet/Razor/issues/623

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
NTaylorMullen: Only if the cache continually expires. So most of the time, no. Didn't feel like it reduced the readability of the code by changing it though. Maybe remove the comment?

src/Microsoft.AspNet.Mvc.TagHelpers/EnvironmentTagHelper.cs
NTaylorMullen: Sure, doesn't sound like this issue though. That sounds more like [this](https://github.com/aspnet/Mvc/issues/3617)

src/Microsoft.AspNet.Mvc.TagHelpers/CacheTagHelper.cs
```diff
@@ -346,11 +351,22 @@ private void AddVaryByRouteKey(StringBuilder builder)
...
+            var trimmedValues = new List<string>();
```
dougbu: Why allocate this `List` when `values.Length == 0`?

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
rynowak: I'd remove the comment if it's not really perf-sensitive code. The point about for/foreach I don't really mind either way. I don't think a `for` makes the code worse.

src/Microsoft.AspNet.Mvc.TagHelpers/EnvironmentTagHelper.cs
```diff
@@ -67,25 +67,33 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
                 return;
             }
 
-            var environments = Names.Split(NameSeparator, StringSplitOptions.RemoveEmptyEntries)
-                                    .Where(name => !string.IsNullOrWhiteSpace(name));
-
-            if (!environments.Any())
+            var currentEnvironmentName = HostingEnvironment.EnvironmentName?.Trim();
+            if (string.IsNullOrWhiteSpace(currentEnvironmentName))
```
dougbu: How can a `Trim`ed value contain just whitespace? Use `IsNullOrEmpty()` instead.

src/Microsoft.AspNet.Mvc.TagHelpers/EnvironmentTagHelper.cs
```diff
@@ -67,25 +67,33 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
...
             {
-                // Names contains only commas or empty entries, do nothing
+                // No current environment name, do nothing
                 return;
             }
 
-            var currentEnvironmentName = HostingEnvironment.EnvironmentName?.Trim();
+            var values = Names.Split(NameSeparator, StringSplitOptions.RemoveEmptyEntries);
+            var environments = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```
dougbu: Why worry much about extra runtime checks if the user is silly enough to repeat environment names in the attribute?

src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
```diff
@@ -389,15 +390,14 @@ private string GetInputType(ModelExplorer modelExplorer, out string inputTypeHin
...
-                modelExplorer.Metadata.DataTypeName,
-            };
+                yield return modelExplorer.Metadata.TemplateHint;
+            }
 
-            foreach (string inputTypeHint in inputTypeHints.Where(s => !string.IsNullOrEmpty(s)))
+            if (!string.IsNullOrEmpty(modelExplorer.Metadata.DataTypeName))
             {
-                yield return inputTypeHint;
+                yield return modelExplorer.Metadata.DataTypeName;
```
dougbu: How expensive is this state machine? Would we be better off returning a `List<string>`?

NTaylorMullen: It has the potential to bail out early in the call site so `yield return` made more sense.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
```diff
@@ -23,7 +22,7 @@ public static class AttributeMatcher
         /// <returns>The <see cref="ModeMatchResult{TMode}"/>.</returns>
         public static ModeMatchResult<TMode> DetermineMode<TMode>(
             TagHelperContext context,
-            IEnumerable<ModeAttributes<TMode>> modeInfos)
+            ModeAttributes<TMode>[] modeInfos)
```
dougbu: Would something less specific such as `IReadOnlyList<ModelAttributes>` really be that much more expensive? This looks bizarre.

src/Microsoft.AspNet.Mvc.TagHelpers/EnvironmentTagHelper.cs
NTaylorMullen: :+1:

src/Microsoft.AspNet.Mvc.TagHelpers/EnvironmentTagHelper.cs
NTaylorMullen: :+1:

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
```diff
@@ -114,9 +120,9 @@ public static class AttributeMatcher
 
         private class PresentMissingAttributes
         {
-            public IEnumerable<string> Present { get; set; }
+            public IList<string> Present { get; set; }
 
-            public IEnumerable<string> Missing { get; set; }
+            public IList<string> Missing { get; set; }
```
dougbu: FYI only: Could use `List<string>` and get a `struct` enumerator for free.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/ModeMatchResult.cs
```diff
@@ -57,17 +55,39 @@ public class ModeMatchResult<TMode>
...
+                            }
+                        }
+
+                        if (presentIsPartialOnlyMatch)
+                        {
+                            partialOnlyMatches.Add(PartialMatches[i]);
+                            break;
+                        }
+                    }
+                }
```
dougbu: This is some complex machinery for logging. Could we log a bit less information and save 20 lines?

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
NTaylorMullen: Sure.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
NTaylorMullen: :+1:

NTaylorMullen: Note: We don't currently use the enumerator for this type anywhere but since it's a `private class` exposing it as `List<string>` will make future changes easier

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/ModeMatchResult.cs
NTaylorMullen: Agree that it's semi-complex. Why change though?

src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
dougbu: That's not always `true` in part because these state machines aren't tiny. Will need to measure both approaches if this one shows up high when profiling a page with 10 or so `<input>` tag helpers.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/ModeMatchResult.cs
dougbu: Same reason @rynowak wanted to hide this in the extension method -- it's ugly.

src/Microsoft.AspNet.Mvc.TagHelpers/CacheTagHelper.cs
```diff
@@ -261,18 +261,19 @@ internal MemoryCacheEntryOptions GetMemoryCacheEntryOptions(IEntryLink entryLink
                        .Append(keyName)
```
dougbu: What still needs `System.Linq` in this `class`?

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -244,7 +244,7 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
```
dougbu: What still needs `System.Linq` in this `class`?

dougbu:
`SelectTagHelper` and `TagHelperOutputExtensions` haven't changed in this PR. Have you confirmed they only use `Linq` for `Enumerable.Empty<T>` and `.ToArray()`?

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
dougbu: What future changes? I expect future simplifications and removals, not additions.

NTaylorMullen: Didn't have anything in mind, there's just no other value to changing it to `List<string>` from `IList<string>`

src/Microsoft.AspNet.Mvc.TagHelpers/InputTagHelper.cs
NTaylorMullen: Looked through the dotmemory profile and it didn't show it's head so going to leave it as is.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/ModeMatchResult.cs
NTaylorMullen: Nah, i'll move tit to an extension method but not going to change behavior.

src/Microsoft.AspNet.Mvc.TagHelpers/CacheTagHelper.cs
NTaylorMullen: nada, removed it.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
NTaylorMullen: A `.ToArray()`

NTaylorMullen:
> SelectTagHelper and TagHelperOutputExtensions haven't changed in this PR. Have you confirmed they only use Linq for Enumerable.Empty<T> and .ToArray()?

Yup.

NTaylorMullen:
:up::date:

rynowak:
:shipit: from me

dougbu:
:shipit:

PR #3647, opened 2015-11-25
[Design] Refactor RazorPage to write to IHtmlContent

### LINKED ISSUE #3030: Refactor IRazorPage to write output to a IHtmlContent instead of TextWriter
The interface contract for a `IRazorPage` \ `RazorPage` is to write to the `TextWriter` instance exposed via `ViewContext.Writer`. In the default case (i.e. the user isn't using a custom `IRazorView`), `ViewContext.Writer` is an instance of `RazorTextWriter` which essentially has the same sort of contract \ behavior as `IHtmlContent`.

We could instead change `IRazorPage` to expose its results as an `IHtmlContent` instead. This change would allow us

a) We can follow the same append pattern we get with appending one `IHtmlContent` to another used through the remainder of the application.

b) Remove the contract that we give special treatment to `TextWriter` instances that implement `IBufferedTextWriter`. Also remove `IBufferedTextWriter`.

Related discussion: Since page instrumentation \ BrowserLink will be affected by this interface change, we can follow up with @jodavis to figure out if they can rely on EventNotification instead our current interface based contract.

pranavkm:
Fixes #3030

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor/HelperResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/HelperResult.cs
@@ -2,9 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.IO;
-using System.Text.Encodings.Web;
-using System.Threading.Tasks;
 using Microsoft.AspNet.Html.Abstractions;
 
 namespace Microsoft.AspNet.Mvc.Razor
@@ -12,53 +9,33 @@ namespace Microsoft.AspNet.Mvc.Razor
     /// <summary>
     /// Represents a deferred write operation in a <see cref="RazorPage"/>.
     /// </summary>
-    public class HelperResult : IHtmlContent
+    public class HelperResult
     {
-        private readonly Func<TextWriter, Task> _asyncAction;
-
         /// <summary>
         /// Creates a new instance of <see cref="HelperResult"/>.
         /// </summary>
-        /// <param name="asyncAction">The asynchronous delegate to invoke when
+        /// <param name="renderAction">The asynchronous delegate to invoke when
         /// <see cref="WriteTo(TextWriter, HtmlEncoder)"/> is called.</param>
         /// <remarks>Calls to <see cref="WriteTo(TextWriter, HtmlEncoder)"/> result in a blocking invocation of
-        /// <paramref name="asyncAction"/>.</remarks>
-        public HelperResult(Func<TextWriter, Task> asyncAction)
+        /// <paramref name="renderAction"/>.</remarks>
+        public HelperResult(RenderAsyncDelegate renderAction)
         {
-            if (asyncAction == null)
+            if (renderAction == null)
             {
-                throw new ArgumentNullException(nameof(asyncAction));
+                throw new ArgumentNullException(nameof(renderAction));
             }
 
-            _asyncAction = asyncAction;
+            RenderAction = renderAction;
         }
 
         /// <summary>
         /// Gets the asynchronous delegate to invoke when <see cref="WriteTo(TextWriter, HtmlEncoder)"/> is called.
         /// </summary>
-        public Func<TextWriter, Task> WriteAction
+        public RenderAsyncDelegate RenderAction { get; }
+
+        public void WriteTo(IHtmlContentBuilder content)
         {
-            get { return _asyncAction; }
-        }
-
-        /// <summary>
-        /// Method invoked to produce content from the <see cref="HelperResult"/>.
-        /// </summary>
-        /// <param name="writer">The <see cref="TextWriter"/> instance to write to.</param>
-        /// <param name="encoder">The <see cref="HtmlEncoder"/> to encode the content.</param>
-        public virtual void WriteTo(TextWriter writer, HtmlEncoder encoder)
-        {
-            if (writer == null)
-            {
-                throw new ArgumentNullException(nameof(writer));
-            }
-
-            if (encoder == null)
-            {
-                throw new ArgumentNullException(nameof(encoder));
-            }
-
-            _asyncAction(writer).GetAwaiter().GetResult();
+            RenderAction(content).GetAwaiter().GetResult();
         }
     }
 }
--- a/src/Microsoft.AspNet.Mvc.Razor/IRazorPage.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/IRazorPage.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Threading.Tasks;
+using Microsoft.AspNet.Html.Abstractions;
 using Microsoft.AspNet.Mvc.Rendering;
 
 namespace Microsoft.AspNet.Mvc.Razor
@@ -22,7 +23,7 @@ namespace Microsoft.AspNet.Mvc.Razor
         /// <summary>
         /// Gets or sets the action invoked to render the body.
         /// </summary>
-        Func<TextWriter, Task> RenderBodyDelegateAsync { get; set; }
+        IHtmlContent BodyContent { get; set; }
 
         /// <summary>
         /// Gets or sets a flag that determines if the layout of this page is being rendered.
@@ -67,5 +68,7 @@ namespace Microsoft.AspNet.Mvc.Razor
         /// <exception cref="InvalidOperationException">if one or more sections were not rendered or if no sections were
         /// defined and the body was not rendered.</exception>
         void EnsureRenderedBodyOrSections();
+
+        IHtmlContentBuilder Output { get; set; }
     }
 }
--- a/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.IO;
 using System.Linq;
 using System.Security.Claims;
 using System.Text.Encodings.Web;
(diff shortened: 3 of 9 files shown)


src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -333,44 +288,37 @@ public static void WriteTo(TextWriter writer, HtmlEncoder encoder, object value)
...
-                    // TextWriter.
-                    htmlTextWriter.Write(htmlContent);
-                }
 
-                return;
+                // This special case allows us to keep buffering as IHtmlContent until we get to the 'final'
+                // IHtmlContentBuilder.
+                contentBuilder.Append(htmlContent);
+            }
+            else if (value is HelperResult)
```
pranavkm: The `HelperResult` delegate changed from `Func<TextWriter, Task>`  to `Func<IHtmlContentBuilder, Task>` - which made it difficult to continue treating it as a `IHtmlContent`.

src/Microsoft.AspNet.Mvc.Razor/RazorTextWriter.cs
```diff
@@ -181,13 +54,9 @@ public override Task WriteLineAsync()
         /// </summary>
         public override void Flush()
         {
-            if (IsBuffering)
-            {
-                IsBuffering = false;
-                TargetWriter = UnbufferedWriter;
```
pranavkm: This used to swtich from buffering to unbuffered when you Flushed. It was hard to keep that behavior with the current flow. That said, I'm not sure if that behavior is useful or expected.

src/Microsoft.AspNet.Mvc.Razor/HelperResult.cs
```diff
@@ -12,53 +9,33 @@ namespace Microsoft.AspNet.Mvc.Razor
...
         /// Creates a new instance of <see cref="HelperResult"/>.
         /// </summary>
-        /// <param name="asyncAction">The asynchronous delegate to invoke when
+        /// <param name="renderAction">The asynchronous delegate to invoke when
         /// <see cref="WriteTo(TextWriter, HtmlEncoder)"/> is called.</param>
         /// <remarks>Calls to <see cref="WriteTo(TextWriter, HtmlEncoder)"/> result in a blocking invocation of
-        /// <paramref name="asyncAction"/>.</remarks>
-        public HelperResult(Func<TextWriter, Task> asyncAction)
+        /// <paramref name="renderAction"/>.</remarks>
+        public HelperResult(RenderAsyncDelegate renderAction)
```
rynowak: :+1:

src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
```diff
@@ -93,22 +95,30 @@ public string Path
                 throw new ArgumentNullException(nameof(context));
             }
 
-            var bodyWriter = await RenderPageAsync(RazorPage, context, ViewStartPages);
-            await RenderLayoutAsync(context, bodyWriter);
+            var oldWriter = context.Writer;
+            var razorTextWriter = new RazorTextWriter(context.Writer, _htmlEncoder);
```
pranavkm: We have to keep this around for things that continue writing to `ViewContext.Output` (like `Html.BeginForm`)

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
```diff
@@ -166,15 +168,24 @@ protected void ProcessUrlAttribute(string attributeName, TagHelperOutput output)
                     }
                     else
                     {
-                        var htmlStringValue = attribute.Value as HtmlString;
-                        if (htmlStringValue != null &&
-                            TryResolveUrl(
-                                htmlStringValue.ToString(),
+                        var htmlContent = attribute.Value as IHtmlContent;
```
pranavkm: Changed from `HtmlString` -> `IHtmlContent`

src/Microsoft.AspNet.Mvc.TagHelpers/TagHelperContentExtensions.cs
```diff
@@ -50,18 +51,19 @@ public static class TagHelperContentExtensions
...
             }
 
-            using (var writer = new TagHelperContentWrapperTextWriter(encoding, content))
+            using (var stringWriter = new StringWriter())
             {
-                using (var stringWriter = new StringWriter())
-                {
-                    RazorPage.WriteTo(stringWriter, encoder, value);
+                var htmlContentBuilder = new HtmlContentBuilder();
+                RazorPage.WriteToStatic(htmlContentBuilder, value);
```
pranavkm: Needs a better name. There's a virtual method with the same name.

src/Microsoft.AspNet.Mvc.Razor/IRazorPage.cs
```diff
@@ -67,5 +68,7 @@ public interface IRazorPage
         /// <exception cref="InvalidOperationException">if one or more sections were not rendered or if no sections were
         /// defined and the body was not rendered.</exception>
         void EnsureRenderedBodyOrSections();
+
+        IHtmlContentBuilder Output { get; set; }
```
rynowak: What's the difference here between `BodyContent` and `Output`.

In the earlier code, I know that `RenderBodyDelegateAsync()` would render the body. What triggers rendering of the body now?

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -28,21 +27,21 @@ namespace Microsoft.AspNet.Mvc.Razor
...
+        private readonly Stack<IHtmlContentBuilder> _contentBuilderScopes;
+        private IHtmlContentBuilder _originalWriter;
         private IUrlHelper _urlHelper;
         private ITagHelperActivator _tagHelperActivator;
         private ITypeActivatorCache _typeActivatorCache;
         private bool _renderedBody;
         private AttributeInfo _attributeInfo;
         private TagHelperAttributeInfo _tagHelperAttributeInfo;
-        private StringCollectionTextWriter _valueBuffer;
+        private HtmlContentBuilder _valueBuffer;
```
rynowak: This is going to get re-specialized when we integrate pooling here. No reason to block you, but FYI.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -72,19 +71,22 @@ public RazorPage()
...
-                if (ViewContext == null)
+                if (_contentBuilderScopes.Count > 0)
                 {
-                    var message = Resources.FormatViewContextMustBeSet("ViewContext", "Output");
-                    throw new InvalidOperationException(message);
+                    return _contentBuilderScopes.Peek();
                 }
 
-                return ViewContext.Writer;
+                return _originalWriter;
```
rynowak: You should push the original writer onto the stack rather than special casing.

src/Microsoft.AspNet.Mvc.Razor/IRazorPage.cs
pranavkm: The delegate used to be invoked from `IRazorPage.RenderBody()`. We simply return the `BodyContent` from it since there's no async work involved.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -193,82 +204,48 @@ private ITypeActivatorCache TypeActivatorCache
...
+                throw new ArgumentNullException(nameof(contentBuilder));
             }
 
             // We need to replace the ViewContext's Writer to ensure that all content (including content written
             // from HTML helpers) is redirected.
-            ViewContext.Writer = writer;
 
-            _writerScopes.Push(ViewContext.Writer);
+            _contentBuilderScopes.Push(contentBuilder);
+            RazorTextWriter.Content = contentBuilder;
```
rynowak: So in the new world we swap out the content on the writer? interesting. You have to be careful doing things like this because TextWriters buffer.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -193,82 +204,48 @@ private ITypeActivatorCache TypeActivatorCache
...
         }
 
         /// <summary>
         /// Ends the current writing scope that was started by calling <see cref="StartTagHelperWritingScope"/>.
         /// </summary>
-        /// <returns>The <see cref="TextWriter"/> that contains the content written to the <see cref="Output"/> or
+        /// <returns>The <see cref="IHtmlContentBuilder"/> that contains the content written to the <see cref="Output"/> or
```
rynowak: `TagHelperContent`

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -281,48 +258,26 @@ public virtual void Write(object value)
...
         /// <param name="value">The <see cref="object"/> to write.</param>
         /// <remarks>
         /// <paramref name="value"/>s of type <see cref="IHtmlContent"/> are written using
-        /// <see cref="IHtmlContent.WriteTo(TextWriter, HtmlEncoder)"/>.
+        /// <see cref="IHtmlContent.WriteTo(IHtmlContentBuilder, HtmlEncoder)"/>.
         /// For all other types, the encoded result of <see cref="object.ToString"/> is written to the
-        /// <paramref name="writer"/>.
+        /// <paramref name="contentBuilder"/>.
         /// </remarks>
-        public virtual void WriteTo(TextWriter writer, object value)
```
rynowak: I don't really see the reason for these changes. Let's discuss Monday.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
pranavkm: We could. There are pieces of code that look at the stack to determine if they're inside a tag helper. Just makes them slightly obtuse now (`if (_contentBuilderScopes.Count > 1) // 1 is fine since it's the default writer`)

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
pranavkm: Yeah ignore the docs for now. I just did a find and replace so it's quite messed up.

src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
dougbu: Why not make `ViewContext.Output` an `IHtmlContentBuilder` as well?

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
dougbu: Changing this is part of #3123 and unrelated to anything else you're doing in this PR.

src/Microsoft.AspNet.Mvc.Razor/HelperResult.cs
davidfowl: Why?

pranavkm:
We decided not to #3030 for the time being. I'll pull in changes around removing the `CopyTo` semantics in `RazorTextWriter` in a separate PR.

PR #3648, opened 2015-11-26
Adds ActionContext to Validation contexts

rynowak:
This change makes it possible to access the ActionContext/ActionDescriptor
from inside of validators and client validators.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Validation/ClientModelValidationContext.cs
+++ b/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Validation/ClientModelValidationContext.cs
@@ -8,10 +8,15 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
     public class ClientModelValidationContext
     {
         public ClientModelValidationContext(
+            ActionContext actionContext,
             ModelMetadata metadata,
-            IModelMetadataProvider metadataProvider,
-            IServiceProvider requestServices)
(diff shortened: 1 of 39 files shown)


src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Validation/ClientModelValidationContext.cs
```diff
@@ -22,20 +27,14 @@ public class ClientModelValidationContext
...
-            {
-                throw new ArgumentNullException(nameof(requestServices));
-            }
-
             ModelMetadata = metadata;
             MetadataProvider = metadataProvider;
-            RequestServices = requestServices;
         }
 
+        public ActionContext ActionContext { get; }
```
rynowak: We had services here in lieu of the `ActionContext` in the past. Now that this is here it seems silly to keep services as a top-level property.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/Validation/DefaultObjectValidatorTests.cs
```diff
@@ -21,7 +21,8 @@ public void Validate_SimpleValueType_Valid_WithPrefix()
         {
             // Arrange
             var validatorProvider = CreateValidatorProvider();
```
rynowak: Everything after this point should be find + replace due to the API changes.

dougbu:
:shipit:

PR #3652, opened 2015-11-26
Remove IActionContextAccessor from UrlHelper

rynowak:
This change removes the IActionContextAccessor as a dependency of
UrlHelper, and shifts UrlHelper to use a factory pattern. Consumers of
IUrlHelper should create an instance using the factory when needed.

This is the last part of MVC that has a dependency on IActionContext
accessor. As part of this change we no longer register it by default, and
treat it as an optional component.


src/Microsoft.AspNet.Mvc.Core/Infrastructure/MvcRouteHandler.cs
```diff
@@ -128,31 +134,23 @@ private Task InvokeActionAsync(RouteContext context, ActionDescriptor actionDesc
...
-                _actionContextAccessor = context.RequestServices.GetRequiredService<IActionContextAccessor>();
+                return;
             }
 
-            if (_actionInvokerFactory == null)
-            {
-                _actionInvokerFactory = context.RequestServices.GetRequiredService<IActionInvokerFactory>();
-            }
+            // The IActionContextAccessor is optional. We want to avoid the overhead of using CallContext
+            // if possible.
```
dougbu: Is the perf issue w/ `CallContext` on read or write (or both)? If only reading is a major problem, could register the service unconditionally and just not read it anywhere in our code.

src/Microsoft.AspNet.Mvc.Core/RedirectToRouteResult.cs
```diff
@@ -77,7 +78,14 @@ public override void ExecuteResult(ActionContext context)
 
         private IUrlHelper GetUrlHelper(ActionContext context)
```
dougbu: To confirm: None of these `GetUrlHelper()` methods are ever called more than once?

src/Microsoft.AspNet.Mvc.Core/Routing/IUrlHelperFactory.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Mvc.Routing
+{
+    public interface IUrlHelperFactory
+    {
+        IUrlHelper Create(ActionContext context);
```
dougbu: Docu commenties

src/Microsoft.AspNet.Mvc.Core/Routing/UrlHelper.cs
```diff
@@ -17,22 +17,27 @@ namespace Microsoft.AspNet.Mvc.Routing
...
+                throw new ArgumentNullException(nameof(actionContext));
+            }
+            
+            ActionContext = actionContext;
         }
 
         protected RouteValueDictionary AmbientValues => ActionContext.RouteData.Values;
 
-        protected ActionContext ActionContext => _actionContextAccessor.ActionContext;
+        protected ActionContext ActionContext { get; set; }
```
dougbu: Value required in the ctor. So why is this settable?

src/Microsoft.AspNet.Mvc.Core/Routing/UrlHelperFactory.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Mvc.Routing
+{
+    public class UrlHelperFactory : IUrlHelperFactory
+    {
+        public IUrlHelper Create(ActionContext context)
```
dougbu: `/// <inheritdoc />`

src/Microsoft.AspNet.Mvc.Razor/RazorPageActivator.cs
```diff
@@ -111,20 +112,29 @@ private PropertyActivator<ViewContext> CreateActivateInfo(PropertyInfo property)
...
                 valueAccessor = context => context.ViewData;
             }
+            else if (typeof(IUrlHelper).IsAssignableFrom(property.PropertyType))
+            {
+                valueAccessor = context =>
+                {
+                    var serviceProvider = context.HttpContext.RequestServices;
+                    var factory = (IUrlHelperFactory)serviceProvider.GetRequiredService(typeof(IUrlHelperFactory));
+                    return factory.Create(context);
+                };
```
dougbu: This doesn't feel right as part of this PR. Have we received requests for `[RazorInject] IUrlHelper UrlHelper { get; set; }`?

src/Microsoft.AspNet.Mvc.TagHelpers/AnchorTagHelper.cs
```diff
@@ -158,6 +165,7 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
                 if (Route == null)
                 {
                     tagBuilder = Generator.GenerateActionLink(
+                        viewContext: ViewContext,
```
dougbu: Don't really need to name this parameter. The rest are just named for the `string.Empty` case.

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
```diff
@@ -166,7 +167,7 @@ public IUrlHelper Url
             {
                 if (_url == null)
                 {
-                    _url = Resolver?.GetRequiredService<IUrlHelper>();
+                    _url = Resolver?.GetRequiredService<IUrlHelperFactory>()?.Create(ControllerContext);
```
dougbu: Huh? Combo of `GetRequiredService` and the null-propagating operator implies a whole new physics.

src/Microsoft.AspNet.Mvc.Core/Routing/IUrlHelperFactory.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Mvc.Routing
```
dougbu: This class is fairly visible, not just an implementation detail. Should be in `Microsoft.AspNet.Mvc` with `IUrlHelper`.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponent.cs
```diff
@@ -108,7 +109,7 @@ public IUrlHelper Url
                 {
                     // May be null in unit-testing scenarios.
                     var services = ViewComponentContext.ViewContext?.HttpContext?.RequestServices;
-                    _url = services?.GetRequiredService<IUrlHelper>();
+                    _url = services?.GetRequiredService<IUrlHelperFactory>()?.Create(ViewComponentContext.ViewContext);
```
dougbu: Sorry @rynowak no @pranavkm points awarded when the `?` isn't useful.

dougbu: :ok: I'll give you 1/2 points for the necessary first '?'.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/IHtmlGenerator.cs
```diff
@@ -22,6 +22,7 @@ public interface IHtmlGenerator
         string FormatValue(object value, string format);
 
         TagBuilder GenerateActionLink(
```
dougbu: Time to invoke the "you touched it; you document it" rule? Perhaps but at most for the two methods you've touched.

src/Microsoft.AspNet.Mvc.WebApiCompatShim/ApiController.cs
```diff
@@ -155,7 +156,7 @@ public IUrlHelper Url
             {
                 if (_urlHelper == null)
                 {
-                    _urlHelper = Context?.RequestServices?.GetRequiredService<IUrlHelper>();
+                    _urlHelper = Context?.RequestServices?.GetRequiredService<IUrlHelperFactory>().Create(ActionContext);
```
dougbu: What, now `RequestServices` can be `null` in a non-`null` `HttpContext`?

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.ViewFeatures/RemoteAttribute.cs
```diff
@@ -199,7 +199,9 @@ protected virtual string GetUrl(ClientModelValidationContext context)
             }
 
             var services = context.ActionContext.HttpContext.RequestServices;
-            var urlHelper = services.GetRequiredService<IUrlHelper>();
+            var factory = services.GetRequiredService<IUrlHelperFactory>();
+            var urlHelper =factory.Create(context.ActionContext);
```
NTaylorMullen: =f

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
```diff
@@ -261,8 +270,8 @@ public virtual IHtmlContent GenerateAntiforgery(ViewContext viewContext)
                 throw new ArgumentNullException(nameof(viewContext));
             }
 
-            var action =
-                _urlHelper.RouteUrl(routeName, values: routeValues, protocol: null, host: null, fragment: null);
+            var urlHelper = _urlHelperFactory.Create(viewContext);
```
NTaylorMullen: Is there something wrong with us just lazily initializing `urlHelper` on first call so repeated calls don't create N url helpers?

NTaylorMullen: Hmm, I see. because we don't capture `ViewContext` on the `DefaultHtmlGenerator`. Seems odd since we typically will have 1 `DefaultHtmlGenerator` per request.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
```diff
@@ -106,6 +107,7 @@ public string FormatValue(object value, string format)
 
         /// <inheritdoc />
         public virtual TagBuilder GenerateActionLink(
+            ViewContext viewContext,
```
NTaylorMullen: Everything that calls this should already have access to a constructed `UrlHelper` what do you think about just passing in the `UrlHelper` here and in `GenerateRouteLink`? Will save unneeded constructions of the object.

NTaylorMullen:
Seems good. Only concern I have is the use in the HtmlGenerator bits (left comments)

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
dougbu: > we typically will have 1 `DefaultHtmlGenerator` per view context.

Huh? This class is registered as scoped in DI and, with @rynowak's changes, can become a singleton.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
dougbu: > Everything that calls this should already have access to a constructed `UrlHelper`

Everything? Perhaps most of the tag helper call sites. But where is this `IUrlHelper` instance in `HtmlHelper`?

src/Microsoft.AspNet.Mvc.Core/Infrastructure/MvcRouteHandler.cs
rynowak: The perf issue is with using it at all. We're going to same way here as hosting went with `IHttpContextAccessor`.

src/Microsoft.AspNet.Mvc.Core/RedirectToRouteResult.cs
rynowak: I don't expect so. Why does it matter?

src/Microsoft.AspNet.Mvc.Core/Routing/IUrlHelperFactory.cs
rynowak: :+1:

src/Microsoft.AspNet.Mvc.Core/Routing/UrlHelper.cs
rynowak: Yeah, good point. This should be `{ get; }`

src/Microsoft.AspNet.Mvc.Razor/RazorPageActivator.cs
rynowak: This is already part of RazorPage by default. This is necessary to keep it working.

src/Microsoft.AspNet.Mvc.TagHelpers/AnchorTagHelper.cs
rynowak: sure

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
rynowak: I'm just removing this chaining across the board, it seems sloppy.

src/Microsoft.AspNet.Mvc.Core/Routing/IUrlHelperFactory.cs
rynowak: It was always intended to put `IUrlHelper` in .`Routing`. Starting a side-discussion about whether or not to move that.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
rynowak: perf is perf. If we find a scenario where this has an impact, we'll do something.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
rynowak: I don't see a reason to change this. A lot of stuff has access to view context here. It seems like really **all** of these should from a design POV.

rynowak:
Updated

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
```diff
@@ -166,7 +167,8 @@ public IUrlHelper Url
             {
                 if (_url == null)
                 {
-                    _url = Resolver?.GetRequiredService<IUrlHelper>();
+                    var factory = Resolver.GetRequiredService<IUrlHelperFactory>();
+                    _url = factory?.GetUrlHelper(ControllerContext);
```
NTaylorMullen: You nuked the null checks on all other locations like this. Why's this one different?

NTaylorMullen:
:ship:

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
rynowak: Yeah this needs a null check on factory.

src/Microsoft.AspNet.Mvc.Core/RedirectToRouteResult.cs
dougbu: I was curious whether `GetRequiredService()` could be expensive enough and called frequently enough that this code should update the `UrlHelper` property. My expectation was that would be useless because the result is executed just once.

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
dougbu: Why does this need a `null` check on `factory`? Using `GetRequiredService()`.

dougbu: Or can `Resolver` still be `null`?

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponent.cs
```diff
@@ -107,8 +108,9 @@ public IUrlHelper Url
                 if (_url == null)
                 {
                     // May be null in unit-testing scenarios.
-                    var services = ViewComponentContext.ViewContext?.HttpContext?.RequestServices;
-                    _url = services?.GetRequiredService<IUrlHelper>();
+                    var services = ViewComponentContext.ViewContext?.HttpContext.RequestServices;
+                    var factory = services?.GetRequiredService<IUrlHelperFactory>();
+                    _url = factory.GetUrlHelper(ViewComponentContext.ViewContext);
```
dougbu: `factory` can be `null` here.

src/Microsoft.AspNet.Mvc.ViewFeatures/Controller.cs
rynowak: `Resolver` can be null.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/IHtmlGenerator.cs
```diff
@@ -2,7 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Collections.Generic;
-using Microsoft.AspNet.Html.Abstractions;
+using Microsoft.AspNet.Html;
```
dougbu: Think you need to rebase. Fortunately have a Coherent build with the update packages now.

rynowak:
Updated again

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/IHtmlGenerator.cs
rynowak: Yheah, haven't rebased yet.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/IHtmlGenerator.cs
```diff
@@ -21,7 +21,31 @@ public interface IHtmlGenerator
...
+        /// created using <see cref="object"/> initializer syntax. Alternatively, an
+        /// <see cref="IDictionary{string, object}"/> instance containing the route parameters.
+        /// </param>
+        /// <param name="htmlAttributes">
+        /// An <see cref="object"/> that contains the HTML attributes for the element. Alternatively, an
+        /// <see cref="IDictionary{string, object}"/> instance containing the HTML attributes.
+        /// </param>
+        /// <returns>
+        /// A <see cref="TagBuilder"/> instance for the &lt;a&gt; element.
+        /// </returns>
```
dougbu: Thx :fireworks:

test/WebSites/ControllersFromServicesClassLibrary/ControllerWithConstructorInjection.cs
```diff
@@ -3,36 +3,22 @@
 
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Mvc;
-using Microsoft.AspNet.Mvc.Routing;
 
 namespace ControllersFromServicesClassLibrary
 {
-    public class ConstructorInjectionController
+    public class ConstructorInjectionController : Controller
```
dougbu: Wasn't this intended to test a POCO controller?

dougbu:
:shipit: my only remaining concern is https://github.com/aspnet/Mvc/pull/3652#discussion-diff-46496548

test/WebSites/ControllersFromServicesClassLibrary/ControllerWithConstructorInjection.cs
rynowak: It appears to be for testing constructor injection.

rynowak: I'll just put it back for funzies since we're not sure.

PR #3661, opened 2015-11-29
Move logic from `HtmlLocalizer` to `LocalizedHtmlString`

dougbu:
- part of #3123 (4 of 5 or so)
- `LocalizedHtmlString` should not subclass `HtmlString`; now implements `IHtmlContent`

nit: Remove duplicate tests reported while testing these fixes in VS.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Localization/HtmlLocalizer.cs
+++ b/src/Microsoft.AspNet.Mvc.Localization/HtmlLocalizer.cs
@@ -141,7 +141,7 @@ namespace Microsoft.AspNet.Mvc.Localization
         /// </summary>
         /// <param name="result">The <see cref="LocalizedString"/>.</param>
         protected virtual LocalizedHtmlString ToHtmlString(LocalizedString result) =>
-            new LocalizedHtmlString(result.Name, result.Value, result.ResourceNotFound, arguments: null);
+            new LocalizedHtmlString(result.Name, result.Value, result.ResourceNotFound);
 
         protected virtual LocalizedHtmlString ToHtmlString(LocalizedString result, object[] arguments) =>
             new LocalizedHtmlString(result.Name, result.Value, result.ResourceNotFound, arguments);
--- a/src/Microsoft.AspNet.Mvc.Localization/HtmlLocalizer.cs
+++ b/src/Microsoft.AspNet.Mvc.Localization/HtmlLocalizer.cs
@@ -4,8 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
-using System.Text;
-using System.Text.Encodings.Web;
 using Microsoft.Extensions.Localization;
 
 namespace Microsoft.AspNet.Mvc.Localization
@@ -17,27 +15,19 @@ namespace Microsoft.AspNet.Mvc.Localization
     public class HtmlLocalizer : IHtmlLocalizer
     {
         private IStringLocalizer _localizer;
-        private readonly HtmlEncoder _encoder;
 
         /// <summary>
         /// Creates a new <see cref="HtmlLocalizer"/>.
         /// </summary>
         /// <param name="localizer">The <see cref="IStringLocalizer"/> to read strings from.</param>
-        /// <param name="encoder">The <see cref="HtmlEncoder"/>.</param>
-        public HtmlLocalizer(IStringLocalizer localizer, HtmlEncoder encoder)
+        public HtmlLocalizer(IStringLocalizer localizer)
         {
             if (localizer == null)
             {
                 throw new ArgumentNullException(nameof(localizer));
             }
 
-            if (encoder == null)
-            {
-                throw new ArgumentNullException(nameof(encoder));
-            }
-
             _localizer = localizer;
-            _encoder = encoder;
         }
 
         /// <inheritdoc />
@@ -80,7 +70,7 @@ namespace Microsoft.AspNet.Mvc.Localization
                 throw new ArgumentNullException(nameof(culture));
             }
 
-            return new HtmlLocalizer(_localizer.WithCulture(culture), _encoder);
+            return new HtmlLocalizer(_localizer.WithCulture(culture));
         }
 
         /// <summary>
@@ -95,7 +85,7 @@ namespace Microsoft.AspNet.Mvc.Localization
                 throw new ArgumentNullException(nameof(culture));
             }
 
-            return new HtmlLocalizer(_localizer.WithCulture(culture), _encoder);
+            return new HtmlLocalizer(_localizer.WithCulture(culture));
         }
 
         /// <inheritdoc />
@@ -143,9 +133,7 @@ namespace Microsoft.AspNet.Mvc.Localization
                 throw new ArgumentNullException(nameof(key));
             }
 
-            var stringValue = _localizer[key].Value;
-
-            return ToHtmlString(new LocalizedString(key, EncodeArguments(stringValue, arguments)));
+            return ToHtmlString(_localizer.GetString(key), arguments);
         }
 
         /// <summary>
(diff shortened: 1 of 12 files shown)


dougbu:
Changes save a couple of large-ish `string`s and `char[]` arrays per request in an HTML helper scenario using an `IViewLocalizer`:
html 0
becomes
html 1

dougbu:
:up::date: for rebase

rynowak:
Nice!

rynowak:
Unrelated: What is that `System.Reflection.Assembl`?

src/Microsoft.AspNet.Mvc.Localization/HtmlLocalizer.cs
```diff
@@ -153,120 +141,9 @@ public virtual LocalizedHtmlString Html(string key, params object[] arguments)
...
         protected virtual LocalizedHtmlString ToHtmlString(LocalizedString result) =>
-            new LocalizedHtmlString(result.Name, result.Value, result.ResourceNotFound);
-
-        /// <summary>
-        /// Encodes the arguments based on the object type.
-        /// </summary>
-        /// <param name="resourceString">The resourceString whose arguments need to be encoded.</param>
-        /// <param name="arguments">The array of objects to encode.</param>
-        /// <returns>The string with encoded arguments.</returns>
-        protected virtual string EncodeArguments(string resourceString, object[] arguments)
```
rynowak: We may need to bring this back for perf reasons if localization requires it. Also, object-pooling for stringbuilders...

src/Microsoft.AspNet.Mvc.Localization/HtmlLocalizerFactory.cs
```diff
@@ -13,27 +12,19 @@ namespace Microsoft.AspNet.Mvc.Localization
...
         private readonly IStringLocalizerFactory _factory;
-        private readonly HtmlEncoder _encoder;
 
         /// <summary>
         /// Creates a new <see cref="HtmlLocalizer"/>.
         /// </summary>
         /// <param name="localizerFactory">The <see cref="IStringLocalizerFactory"/>.</param>
-        /// <param name="encoder">The <see cref="HtmlEncoder"/>.</param>
-        public HtmlLocalizerFactory(IStringLocalizerFactory localizerFactory, HtmlEncoder encoder)
+        public HtmlLocalizerFactory(IStringLocalizerFactory localizerFactory)
```
rynowak: :heart_eyes_cat:

src/Microsoft.AspNet.Mvc.Localization/IHtmlLocalizer.cs
```diff
@@ -19,6 +19,9 @@ public interface IHtmlLocalizer : IStringLocalizer
         /// <returns>A culture-specific <see cref="IHtmlLocalizer"/>.</returns>
         new IHtmlLocalizer WithCulture(CultureInfo culture);
 
+        // Reviewers: I suggest the Html() methods should just return IHtmlContent because the specialized information
+        // in LocalizedHtmlString is unused. Any objections?
```
rynowak: I don't feel strongly about it. If we're going to provide LocalizedHtmlString as an abstraction, then it can be nice for unit testing if that's a factor.

rynowak: It's obviously not required to make the API work

src/Microsoft.AspNet.Mvc.Localization/IViewLocalizer.cs
```diff
@@ -4,7 +4,8 @@
 namespace Microsoft.AspNet.Mvc.Localization
 {
     /// <summary>
-    /// A service that provides localized strings for views.
+    /// A service that provides localized strings for views. Keys are interpreted relative to the path of the executing
+    /// view.
```
rynowak: > Keys are interpreted relative to the path of the executing view.

IMO this doesn't provide enough information to be useful.

src/Microsoft.AspNet.Mvc.Localization/LocalizedHtmlString.cs
```diff
@@ -26,13 +27,28 @@ public LocalizedHtmlString(string key, string value)
         /// <param name="key">The name of the string resource.</param>
         /// <param name="value">The string resource.</param>
         /// <param name="isResourceNotFound">A flag that indicates if the resource is not found.</param>
-        public LocalizedHtmlString(string key, string value, bool isResourceNotFound)
-            : base(value)
+        /// <param name="arguments">The values to format the <paramref name="value"/> with.</param>
+        public LocalizedHtmlString(string key, string value, bool isResourceNotFound, params object[] arguments)
```
rynowak: You should still provide the constructor without `params` here. The array is created by the callee.

src/Microsoft.AspNet.Mvc.Localization/LocalizedHtmlString.cs
```diff
@@ -41,11 +57,145 @@ public LocalizedHtmlString(string key, string value, bool isResourceNotFound)
...
+            HtmlEncoder encoder,
+            string resourceString,
+            object[] arguments)
+        {
+            var position = 0;
+            var length = resourceString.Length;
+            StringBuilder tokenBuffer = null;
+            var isToken = false;
+
+            while (position < length)
```
rynowak: OK good you did keep this code :+1:

src/Microsoft.AspNet.Mvc.Localization/LocalizedHtmlString.cs
```diff
@@ -26,13 +27,28 @@ public LocalizedHtmlString(string key, string value)
...
+            }
+
             Key = key;
+            Value = value;
             IsResourceNotFound = isResourceNotFound;
+            _arguments = arguments ?? EmptyArguments;
         }
 
+        // Reviewers: The following properties, especially Key and IsResourceNotFound seem to exist only for
+        // consistency with LocalizedString but they're unused. Should they be removed?
```
rynowak: I'm not the person to answer this, it's probably @DamianEdwards. My guess is that `Key` is there for debugging/testing, and `IsResourceNotFound` is vestigial and based on a different behavior for what to do if a resource is missing.

rynowak: Or, the `Key` was used to throw a meaningful error message when `IsResourceNotFound` was true.

rynowak:
:shipit:

src/Microsoft.AspNet.Mvc.Localization/IHtmlLocalizerOfT.cs
```diff
@@ -4,9 +4,10 @@
 namespace Microsoft.AspNet.Mvc.Localization
 {
     /// <summary>
-    /// An <see cref="IHtmlLocalizer"/> that provides localized HTML content.
+    /// An <see cref="IHtmlLocalizer"/> that provides localized HTML content. Keys are interpreted relative to the
+    /// <see cref="System.Type.FullName"/> of <typeparamref name="TResource"/>.
     /// </summary>
-    /// <typeparam name = "TResource"> The <see cref="System.Type"/> to scope the resource names.</typeparam>
+    /// <typeparam name = "TResource">The <see cref="System.Type"/> to scope the resource names.</typeparam>
```
pranavkm: > `<typeparam name="TResource">`

src/Microsoft.AspNet.Mvc.Localization/LocalizedHtmlString.cs
```diff
@@ -41,11 +57,145 @@ public LocalizedHtmlString(string key, string value, bool isResourceNotFound)
...
+            {
+                FormatValue(writer, encoder, Value, _arguments);
+            }
+            else
+            {
+                htmlTextWriter.Write(this);
+            }
+        }
+
+        private void FormatValue(
```
pranavkm: Could this be static?

pranavkm: Is this copied as is from HtmlLocalizer? (does this need to be reviewed?)

src/Microsoft.AspNet.Mvc.Localization/LocalizedHtmlString.cs
```diff
@@ -41,11 +57,145 @@ public LocalizedHtmlString(string key, string value, bool isResourceNotFound)
...
+            }
+            else
+            {
+                writer.Write(value);
+            }
+        }
+
+        private void AppendCurly(
+            bool isToken,
+            char curly,
```
pranavkm: `char braces`?

src/Microsoft.AspNet.Mvc.Localization/LocalizedHtmlString.cs
```diff
@@ -1,24 +1,25 @@
...
     /// </summary>
-    public class LocalizedHtmlString : HtmlString
+    public class LocalizedHtmlString : IHtmlContent
     {
-        /// <summary>
-        /// Creates an instance of <see cref="LocalizedHtmlString"/>.
-        /// </summary>
-        /// <param name="key">The name of the string resource.</param>
-        /// <param name="value">The string resource.</param>
-        public LocalizedHtmlString(string key, string value)
```
pranavkm: Where did this ctor go?

src/Microsoft.AspNet.Mvc.Localization/LocalizedHtmlString.cs
dougbu: Maybe `brace` but only if you feel strongly. The `char` is always `'{'` or `'}'`, not one of those new-fangled square or angular (Angular:tm: ?) thingies or (horrors) a parenthesis.

src/Microsoft.AspNet.Mvc.Localization/LocalizedHtmlString.cs
dougbu: No, the callee (there's just one) explicitly passes `null`. I can add the ctor back but it'll be for style and not to save an array allocation.

src/Microsoft.AspNet.Mvc.Localization/LocalizedHtmlString.cs
pranavkm: Not super vested. `curly` reminds me of a comedy act rather than the serious business that is localization.

dougbu:
> Unrelated: What is that `System.Reflection.Assembl`?

`System.Reflection.AssemblyName`

Most of the allocations are internal to `ResourceManager` and trace back through `Microsoft.Extensions.Localization.ResourceManagerStringLocalizer.GetStringSafely(...)`. `ResourceManagerStringLocalizer` has a cache of _missing_ resources (well, manifests) but that's per-instance. Not sure why `ResourceManagerStringLocalizerFactory` (a singleton) doesn't own that cache nor why only `ResourceManagerStringLocalizer.GetAllStrings()` uses the factory's resource string cache.

The few allocations outside `GetStringSafely()` are shown below.
assemblyname

DamianEdwards:
We should investigate making that cache Global

dougbu:
> We should investigate making that cache Global

The resource names cache _is_ global. The problem is it's not used except in the `GetAllStrings()` method and nothing uses that method.

DamianEdwards:
I meant the missing manifest cache, but the meta point is we should cache whatever makes sense.

src/Microsoft.AspNet.Mvc.Localization/IHtmlLocalizer.cs
dougbu: Decided the additional cleanup I mentioned in these "Reviewers" comments can happen later. Will file an issue in the Localization repo.

dougbu: #3681 and aspnet/Localization#158.

src/Microsoft.AspNet.Mvc.Localization/LocalizedHtmlString.cs
dougbu: Totally, completely and entirely not used. #3681 and aspnet/Localization#158.

src/Microsoft.AspNet.Mvc.Localization/IViewLocalizer.cs
dougbu: Removed

src/Microsoft.AspNet.Mvc.Localization/IHtmlLocalizerOfT.cs
dougbu: :white_check_mark:

src/Microsoft.AspNet.Mvc.Localization/LocalizedHtmlString.cs
dougbu: Mostly copied from `HtmlLocalizer` but old code used a second `StringBuilder` rather than the current `TextWriter`. Also added `AppendCurlyBrace()`.

src/Microsoft.AspNet.Mvc.Localization/LocalizedHtmlString.cs
dougbu: Renamed method and the parameter.

src/Microsoft.AspNet.Mvc.Localization/LocalizedHtmlString.cs
dougbu: Never mind, I brought it back.

src/Microsoft.AspNet.Mvc.Localization/LocalizedHtmlString.cs
dougbu: Oh, and I made the `private` methods `static` as well.

dougbu:
:up::date: also rebased. /cc @pranavkm

rynowak:
:shipit:

pranavkm:
:shipit:

dougbu:
42f3e76

PR #3664, opened 2015-11-30
Removing CopyTo from RazorTextWriter

pranavkm:


pranavkm:
cc @rynowak

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -249,26 +249,26 @@ public TagHelperContent EndTagHelperWritingScope()
...
                 var stringCollectionTextWriter = writer as StringCollectionTextWriter;
                 if (stringCollectionTextWriter != null)
                 {
-                    stringCollectionTextWriter.CopyTo(tagHelperContentWrapperTextWriter, HtmlEncoder);
+                    tagHelperContent.Append(stringCollectionTextWriter.Content);
                 }
                 else
                 {
-                    tagHelperContentWrapperTextWriter.Write(writer.ToString());
+                    tagHelperContent.Append(writer.ToString());
```
rynowak: I'm not a fan of this case. Calling `ToString` on a `TextWriter` only does the right thing in a few specific cases. Should this throw instead? I don't think we let the user plug in an arbitrary textwriter do we?

src/Microsoft.AspNet.Mvc.Razor/IRazorPage.cs
```diff
@@ -20,9 +19,9 @@ public interface IRazorPage
         ViewContext ViewContext { get; set; }
 
         /// <summary>
-        /// Gets or sets the action invoked to render the body.
+        /// Gets or sets the body content.
         /// </summary>
-        Func<TextWriter, Task> RenderBodyDelegateAsync { get; set; }
+        IHtmlContent BodyContent { get; set; }
```
rynowak: Should this be a method instead? What's the point of this property?

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
pranavkm: Yeah - it probably expects a `StringWriter`. Let me figure out what scenarios causes it to land here.

src/Microsoft.AspNet.Mvc.Razor/IRazorPage.cs
pranavkm: This property is the buffered content that gets passed to a layout page, not super sure what it would look like if we were to turn it into a method.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
dougbu: `ViewContext.Writer` has a public setter and users are probably doing all kinda strange things deep in a partial view. Fortunately at the moment the worst case is an ugly type name showing up in the generated HTML.

src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
```diff
@@ -203,7 +203,7 @@ private Task RenderPageCoreAsync(IRazorPage page, ViewContext context)
                 // in the layout.
                 previousPage.IsLayoutBeingRendered = true;
                 layoutPage.PreviousSectionWriters = previousPage.SectionWriters;
-                layoutPage.RenderBodyDelegateAsync = bodyWriter.CopyToAsync;
+                layoutPage.BodyContent = bodyWriter.Buffer;
```
dougbu: So the body is treated even less like a section than before?

pranavkm: I guess?

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
pranavkm: I'll file a bug for this.

pranavkm:
bump?

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
dougbu: Agree on the bug but this line should use `AppendHtml()`. Don't want to double-encode the `string`.

dougbu: And do we have tests demonstrating the `StringWriter` case doesn't lead to double-encoding?

pranavkm: The code hasn't changed apart from not using the `CopyTo` semantics of writer. `Append` would be equivalent to `Write`.

test/Microsoft.AspNet.Mvc.Razor.Test/RazorPageTest.cs
```diff
@@ -311,7 +311,7 @@ public void IsSectionDefined_ThrowsIfPreviousSectionWritersIsNotRegistered()
             {
                 { "baz", _nullRenderAsyncDelegate }
             };
-            page.RenderBodyDelegateAsync = CreateAsyncWriteDelegate("body-content");
+            page.BodyContent = new HtmlEncodedString("body-content");
```
dougbu: Why not `HtmlString`? We don't use `HtmlEncodedString` anywhere in the MVC repo.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
dougbu: No, `IHtmlContentBuilder` implementations encode when `Append()` is called while `TextWriter`s do not encode when `Write()` is called.

dougbu:
:watch:

test/Microsoft.AspNet.Mvc.Razor.Test/RazorPageTest.cs
pranavkm: Ok

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
pranavkm: Ok.

pranavkm:
:up: :date:

dougbu:
:shipit: if @rynowak is :ok: with the `BodyContent` property now.

PR #3667, opened 2015-12-01
Dispose FileWatcher in tests

BrennanConroy:
@dougbu

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor.Host/IMvcRazorHost.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor.Host/IMvcRazorHost.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.IO;
 using Microsoft.AspNet.Razor.CodeGenerators;
 
@@ -9,7 +10,7 @@ namespace Microsoft.AspNet.Mvc.Razor
     /// <summary>
     /// Specifies the contracts for a Razor host that parses Razor files and generates C# code.
     /// </summary>
-    public interface IMvcRazorHost
+    public interface IMvcRazorHost : IDisposable
     {
         /// <summary>
         /// Parses and generates the contents of a Razor file represented by <paramref name="inputStream"/>.
--- a/src/Microsoft.AspNet.Mvc.Razor/Precompilation/RazorPreCompiler.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/Precompilation/RazorPreCompiler.cs
@@ -277,36 +277,38 @@ namespace Microsoft.AspNet.Mvc.Razor.Precompilation
 
             using (var stream = fileInfo.FileInfo.CreateReadStream())
             {
-                var host = GetRazorHost();
-                var results = host.GenerateCode(fileInfo.RelativePath, stream);
-
-                if (results.Success)
+                using (var host = GetRazorHost())
                 {
-                    var syntaxTree = SyntaxTreeGenerator.Generate(
-                        results.GeneratedCode,
-                        fileInfo.FileInfo.PhysicalPath,
-                        CompilationSettings);
-                    var fullTypeName = results.GetMainClassName(host, syntaxTree);
+                    var results = host.GenerateCode(fileInfo.RelativePath, stream);
 
-                    if (fullTypeName != null)
+                    if (results.Success)
                     {
-                        var razorFileInfo = new RazorFileInfo
+                        var syntaxTree = SyntaxTreeGenerator.Generate(
+                            results.GeneratedCode,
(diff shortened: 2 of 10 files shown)


test/Microsoft.AspNet.Mvc.Core.Test/VirtualFileResultTest.cs
```diff
@@ -262,21 +262,23 @@ public void Constructor_SetsContentTypeAndParameters()
         {
             // Arrange
             // Point the IFileProvider root to a different subfolder
-            var fileProvider = new PhysicalFileProvider(Path.GetFullPath("./Properties"));
-            var filePathResult = new VirtualFileResult(path, "text/plain")
+            using (var fileProvider = new PhysicalFileProvider(Path.GetFullPath("./Properties")))
```
pranavkm: We might be able to do away with not using a `PhysicalFileProvider` here. It looks like all it cares about is to get a `NotFound` file result which we could mock rather than use a real file.

dougbu:
Good start. But shouldn't we be more worried about the `PhysicalFileProvider` instances MVC product code creates e.g. in `RazorViewEngineOptionsSetup`, `MvcRazorHost` and `RazorPreCompileModule`? Or the `PhysicalFileProvider` instance per `TestServer` that `IHostingEnvironment.WebRootFileProvider` brings in?

BrennanConroy:
That would be great, but how do you propose disposing those?

dougbu:
Some of them are used within a single method e.g. the `PhysicalFileProvider` created in `RazorPreCompileModule` should last only 'til the end of `BeforeCompile()`. `MvcRazorHost` and `IChunkTreeCache` could be `IDisposable` and then at least `RazorPreCompiler` could clean them up. Plus `MvcRazorHost` is registered as transient in DI. Doesn't that lead to automatic disposal if it implements `IDisposable`?

All of that leaves only the `PhysicalFileProvider` we bury in `RazorViewEngineOptions`. Not sure whether it would matter if `RazorViewEngineOptions` were `IDisposable`. But worth testing if it could be cleaned up when xUnit disposes of `MvcTestFixture` instances and therefore `TestServer`s.

BrennanConroy:
The ```PhysicalFileProvider``` is passed to RazorPreCompiler and placed inside the BeforeCompileContext which was passed to BeforeCompile. So we cannot dispose it without possibly causing issues somewhere else.

I can look into making the other things disposable though and the DI should deal with the disposable correctly.

dougbu:
> The `PhysicalFileProvider` is passed to `RazorPreCompiler`

The `RazorPreCompiler` is very short-lived. I meant updating `RazorPreCompileModule` like:
``` c#
    using (var fileProvider = new PhysicalFileProvider(context.ProjectContext.ProjectDirectory))
    {
        var viewCompiler = new RazorPreCompiler(
            context,
            fileProvider,
            memoryCache)
        {
            GenerateSymbols = GenerateSymbols
        };

        viewCompiler.CompileViews();
    }
```

In turn `RazorPreCompiler` could wrap its `MvcRazorHost` instance in a `using`.

BrennanConroy:
Added some disposes

test/Microsoft.AspNet.Mvc.Core.Test/VirtualFileResultTest.cs
dougbu: @pranavkm I'm not too worried about this somewhat-unrelated suggestion. Are you?

src/Microsoft.AspNet.Mvc.Razor/Precompilation/RazorPreCompiler.cs
```diff
@@ -276,8 +276,8 @@ protected virtual PrecompilationCacheEntry GetCacheEntry(RelativeFileInfo fileIn
             }
 
             using (var stream = fileInfo.FileInfo.CreateReadStream())
+            using (var host = GetRazorHost())
```
dougbu: Add curly braces for the outer `using` and nest this appropriately.

dougbu:
:shipit: once @pranavkm has answered on the open question. LGTM

test/Microsoft.AspNet.Mvc.Core.Test/VirtualFileResultTest.cs
pranavkm: Nope. We can change it as a follow up.

src/Microsoft.AspNet.Mvc.Razor.Host/IMvcRazorHost.cs
```diff
@@ -9,7 +10,7 @@ namespace Microsoft.AspNet.Mvc.Razor
     /// <summary>
     /// Specifies the contracts for a Razor host that parses Razor files and generates C# code.
     /// </summary>
-    public interface IMvcRazorHost
+    public interface IMvcRazorHost : IDisposable
```
pranavkm: cc @ToddGrun for tooling.

pranavkm:
:shipit:

PR #3673, opened 2015-12-02
Refactor razor backing buffer

pranavkm:


src/Microsoft.AspNet.Mvc.Razor/Buffer/RazorBuffer.cs
```diff
@@ -0,0 +1,131 @@
...
+
+            var chunk = BufferChunks[BufferChunks.Count - 1];
+            chunk.Data.Array[chunk.Data.Offset + _currentIndex] = new RazorValue(encoded, needsEncoding: false);
+            _currentIndex = (_currentIndex + 1) % chunk.Data.Count;
+            return this;
+        }
+
+        public IHtmlContentBuilder Clear()
+        {
+            throw new NotImplementedException();
```
pranavkm: I should fix this

pranavkm:
This uses a modified https://github.com/rynowak/mvc-sandbox/tree/master/workloads/BigViews (modified to use a Layout and fewer items so that `SelectListItem` doesn't dominate the results)

For 3000 requests, this dropped allocations by about 12 MB (~5%). The comparisons aren't fully equivalent because some types disappeared entirely which is tricky to compare.

**Before**
before


**After**
after

src/Microsoft.AspNet.Mvc.Razor/Buffer/RazorBufferChunk.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNet.Mvc.Razor.Buffer
+{
+    public struct RazorBufferChunk
```
pranavkm: Would be nice to not use the term `Chunk` here - `RazorBufferBlock`?

pranavkm:
cc @rynowak

src/Microsoft.AspNet.Mvc.Razor/Buffer/MemoryPoolRazorBufferSource.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            LeasedArraySegment<RazorValue> segment = null;
+
+            try
+            {
+                segment = _pool.Lease(1024);
+                _leased.Add(segment);
+                return new RazorBufferChunk(segment.Data);
+            }
+            catch when (segment != null)
```
dougbu: :100: Pranav Points :stars:

src/Microsoft.AspNet.Mvc.Razor/Buffer/MemoryPoolRazorBufferSource.cs
```diff
@@ -0,0 +1,57 @@
...
+        public MemoryPoolRazorBufferSource(IArraySegmentPool<RazorValue> pool)
+        {
+            _pool = pool;
+
+            _leased = new List<LeasedArraySegment<RazorValue>>();
+        }
+
+        public RazorBufferChunk GetChunk()
+        {
+            LeasedArraySegment<RazorValue> segment = null;
```
dougbu: Shouldn't this method `throw` if `_disposed`?

src/Microsoft.AspNet.Mvc.Razor/Buffer/RazorBuffer.cs
```diff
@@ -0,0 +1,131 @@
...
+namespace Microsoft.AspNet.Mvc.Razor.Buffer
+{
+    [DebuggerDisplay("{DebuggerToString()}")]
+    public class RazorBuffer : IHtmlContentBuilder
+    {
+        private readonly IRazorBufferSource _bufferSource;
+        private readonly string _name;
+        private int _currentIndex;
+
+        public RazorBuffer(IRazorBufferSource scope, string name)
```
dougbu: `scope`?

src/Microsoft.AspNet.Mvc.Razor/Buffer/RazorBufferChunk.cs
dougbu: Razor overuses both `Block` and `Chunk`. Suggest `RazorBufferSegment`.

src/Microsoft.AspNet.Mvc.Razor/Buffer/RazorBufferTextWriter.cs
```diff
@@ -0,0 +1,166 @@
...
+
+namespace Microsoft.AspNet.Mvc.Razor.Buffer
+{
+    /// <summary>
+    /// A <see cref="HtmlTextWriter"/> that stores individual write operations as a sequence of
+    /// <see cref="string"/> and <see cref="IHtmlContent"/> instances.
+    /// </summary>
+    /// <remarks>
+    /// This is primarily designed to avoid creating large in-memory strings.
+    /// Refer to https://aspnetwebstack.codeplex.com/workitem/585 for more details.
```
dougbu: Ick CodePlex. Suggest we need a GitHub issue tracking this work or use aspnet/Razor#614.

src/Microsoft.AspNet.Mvc.Razor/Buffer/RazorBufferTextWriter.cs
```diff
@@ -0,0 +1,166 @@
...
+    /// </remarks>
+    public class RazorBufferTextWriter : HtmlTextWriter
+    {
+        private const int MaxCharToStringLength = 1024;
+
+        private readonly Encoding _encoding;
+        private readonly RazorBuffer _buffer;
+
+        public RazorBufferTextWriter(RazorBuffer buffer, Encoding encoding)
```
dougbu: Why not `IHtmlContentBuilder buffer`?

src/Microsoft.AspNet.Mvc.Razor/Buffer/RazorBufferTextWriter.cs
```diff
@@ -0,0 +1,166 @@
...
+            if (buffer == null)
+            {
+                throw new ArgumentNullException(nameof(buffer));
+            }
+
+            if (index < 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(index));
+            }
+            if (count < 0 || (buffer.Length - index < count))
```
dougbu: Why not `index + count > buffer.Length`? Bit easier to read.

src/Microsoft.AspNet.Mvc.Razor/Buffer/RazorBufferTextWriter.cs
```diff
@@ -0,0 +1,166 @@
...
+
+        /// <inheritdoc />
+        public override void Write(string value)
+        {
+            if (string.IsNullOrEmpty(value))
+            {
+                return;
+            }
+
+            _buffer.AppendHtml(value);
```
dougbu: If `value` is a 1KB (or more) string, should this split it up and allow the long string to be GC'd earlier?

src/Microsoft.AspNet.Mvc.Razor/Buffer/RazorValue.cs
```diff
@@ -0,0 +1,61 @@
...
+            }
+
+            var htmlContentValue = Value as IHtmlContent;
+            if (htmlContentValue != null)
+            {
+                htmlContentValue.WriteTo(writer, encoder);
+                return;
+            }
+
+            Debug.Fail("We shouldn't get here.");
```
dougbu: `Debug.Assert(Value is IHtmlContent)` earlier and make the cast unconditional.

src/Microsoft.AspNet.Mvc.Razor/Buffer/RazorValue.cs
```diff
@@ -0,0 +1,61 @@
...
+        public RazorValue(string encoded, bool needsEncoding)
+        {
+            Value = encoded;
+            _needsEncoding = needsEncoding;
+        }
+
+        public RazorValue(IHtmlContent content)
+        {
+            Value = content;
+            _needsEncoding = false;
```
dougbu: `true` is closer to the truth.

src/Microsoft.AspNet.Mvc.Razor/RazorTextWriter.cs
```diff
@@ -28,14 +24,14 @@ public class RazorTextWriter : HtmlTextWriter
         /// </summary>
         /// <param name="unbufferedWriter">The <see cref="TextWriter"/> to write output to when this instance
         /// is no longer buffering.</param>
-        /// <param name="encoding">The character <see cref="Encoding"/> in which the output is written.</param>
+        /// <param name="buffer">The <see cref="RazorBuffer"/>.</param>
         /// <param name="encoder">The HTML encoder.</param>
-        public RazorTextWriter(TextWriter unbufferedWriter, Encoding encoding, HtmlEncoder encoder)
+        public RazorTextWriter(TextWriter unbufferedWriter, RazorBuffer buffer, HtmlEncoder encoder)
```
dougbu: Again, why not `IHtmlContentBuilder`?

src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
```diff
@@ -93,6 +97,7 @@ public string Path
                 throw new ArgumentNullException(nameof(context));
             }
 
+            _bufferSource = context.HttpContext.RequestServices.GetRequiredService<IRazorBufferSource>();
```
dougbu: `RazorPageEngine` should pass this into the ctor.

dougbu:
Few comments but looks like a good start.

src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
pranavkm: The service is scoped to a request but view engines are singletons. So that wouldn't work.

src/Microsoft.AspNet.Mvc.Razor/RazorTextWriter.cs
pranavkm: The `RazorBufferTextWriter` deals with `RazorBuffer` instances.

src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
dougbu: Then this can't be used safely on multiple threads. Need to pass the `IRazorBufferSource` around everywhere.

src/Microsoft.AspNet.Mvc.Razor/RazorTextWriter.cs
dougbu: Right, I asked "why" there too.

src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
pranavkm: > Then this can't be used safely on multiple threads.

Could you explain the scenario where this fails?

dougbu: - Thread A grabs its buffer source and sets `_bufferSource`.
- Thread B grabs its buffer source and sets `_bufferSource`.
- Thread A gets around to an invocation of `RenderPageAsync()`, perhaps for a layout.

src/Microsoft.AspNet.Mvc.Razor/Buffer/MemoryPoolRazorBufferSource.cs
rynowak: Yes, that would be nice to help detect incorrect usage.

src/Microsoft.AspNet.Mvc.Razor/Buffer/MemoryPoolRazorBufferSource.cs
```diff
@@ -0,0 +1,57 @@
...
+
+            try
+            {
+                segment = _pool.Lease(1024);
+                _leased.Add(segment);
+                return new RazorBufferChunk(segment.Data);
+            }
+            catch when (segment != null)
+            {
+                segment.Owner.Return(segment);
```
rynowak: You could have a double-Reture here if `RazorBufferChunk.ctor()` throws

src/Microsoft.AspNet.Mvc.Razor/Buffer/IRazorBufferSource.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Mvc.Razor.Buffer
+{
+    public interface IRazorBufferSource
```
rynowak: Personally I like `IRazorBufferScope` a little more.

src/Microsoft.AspNet.Mvc.Razor/Buffer/RazorBuffer.cs
```diff
@@ -0,0 +1,131 @@
...
+            if (scope == null)
+            {
+                throw new ArgumentNullException(nameof(scope));
+            }
+
+            _bufferSource = scope;
+            _name = name;
+        }
+
+        public List<RazorBufferChunk> BufferChunks { get; private set; }
```
rynowak: `IList` or private

src/Microsoft.AspNet.Mvc.Razor/Buffer/RazorBuffer.cs
rynowak: Why?

rynowak: I would just change it to `NotSupportedException`

src/Microsoft.AspNet.Mvc.Razor/Buffer/RazorBuffer.cs
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        public void WriteTo(TextWriter writer, HtmlEncoder encoder)
+        {
+            if (BufferChunks == null)
+            {
+                return;
+            }
+
+            for (var i = 0; i < BufferChunks.Count; i++)
```
rynowak: We should special case `HtmlTextWriter` here

src/Microsoft.AspNet.Mvc.Razor/Buffer/RazorBufferChunk.cs
rynowak: :+1: for segment

src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
pranavkm: I'm still not following this - why is the same view instance being executed from multiple threads?

src/Microsoft.AspNet.Mvc.Razor/Buffer/RazorBufferTextWriter.cs
rynowak: I don't think we need this reference. It's a pretty prevalent concept in our system.

src/Microsoft.AspNet.Mvc.Razor/RazorTextWriter.cs
pranavkm: Yeah, doesn't seem like it needs to use the concrete type.

src/Microsoft.AspNet.Mvc.Razor/Buffer/RazorBufferTextWriter.cs
rynowak: That sounds like a big rathole. I'd suggest we wait for evidence that's a good idea.

Generally the long strings come from static razor content, so they aren't eligible for GC.

src/Microsoft.AspNet.Mvc.Razor/Buffer/RazorValue.cs
```diff
@@ -0,0 +1,61 @@
...
+using System.Text.Encodings.Web;
+using Microsoft.AspNet.Html.Abstractions;
+
+namespace Microsoft.AspNet.Mvc.Razor.Buffer
+{
+    public struct RazorValue
+    {
+        private readonly bool _needsEncoding;
+
+        public RazorValue(string encoded, bool needsEncoding)
```
rynowak: We should just get rid of `needsEncoding` here and wrap the strings that **don't** need encoding. This was an experiment that was I was doing, and I'd prefer to continue it after implementing the buffer chunks.

src/Microsoft.AspNet.Mvc.Razor/Buffer/RazorBufferTextWriter.cs
pranavkm: I'm not sure if there's value in doing it without knowing if \ how we end up with large strings here.

src/Microsoft.AspNet.Mvc.Razor/Buffer/RazorBufferTextWriter.cs
pranavkm: Sure

src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
rynowak: How are two different threads executing a razor view **at the same time**

src/Microsoft.AspNet.Mvc.Razor/Buffer/RazorBufferChunk.cs
pranavkm: :+1:

src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
dougbu: Ah, right. `RazorViewEngine` doesn't cache or reuse `RazorView` instances. My apologies.

pranavkm:
:up: :date:

src/Microsoft.AspNet.Mvc.Razor/Buffer/RazorBuffer.cs
```diff
@@ -0,0 +1,158 @@
...
+
+        /// <inheritdoc />
+        public void WriteTo(TextWriter writer, HtmlEncoder encoder)
+        {
+            if (BufferChunks == null)
+            {
+                return;
+            }
+
+            var htmlTextWriter = writer as HtmlTextWriter;
```
rynowak: if the writer is an `HtmlTextWriter` just insert `this` and return

rynowak:
Can get we some numbers for BasicViews also? (Both with and without TagHelpers). I want to see how many chunks get created for each request.

pranavkm:
TagHelpers continue to use non-buffer pooled `StringCollectionTextWriter` so we're not going to see much improvement there. I'll send a follow up PR to address that (and combine `HtmlTextWriterWrapper` and `StringCollectionTextWriter`). 

**Before**
before

**After**
after

rynowak:
This is still a really nice improvement for the basic scenario :+1: Excelsior!

pranavkm:
:up: :date:

src/Microsoft.AspNet.Mvc.Razor/Buffer/RazorBuffer.cs
pranavkm: Making the suggested change exposed a potential issue. We end up with an instance of `RazorBuffer` inside of a `CacheTagHelper` and we cannot keep the `RazorBuffer` around between requests.

rynowak: Yeah, cachetag needs to use a stringwriter.

pranavkm:
:up: :date:

src/Microsoft.AspNet.Mvc.Razor/Buffer/MemoryPoolRazorBufferScope.cs
```diff
@@ -0,0 +1,82 @@
...
+                segment = _pool.Lease(SegmentSize);
+                _leased.Add(segment);
+            }
+            catch when (segment != null)
+            {
+                segment.Owner.Return(segment);
+                throw;
+            }
+
+            return new RazorBufferSegment(segment.Data);
```
rynowak: This will leak if the constructor throws. This isn't really important, because the pool that the CLR team is implementing with tolerate leaks in this case. Likelihood of happening = 0 / 100

rynowak:
:shipit: !

PR #3675, opened 2015-12-02
Clean up remaining `HtmlString` usage

dougbu:
- #3122 and #3123 (5 of 5)
- avoid concatenating values and wrapping them in an `HtmlString`
  - simplify `LinkTagHelper`, `RazorPage` and `TagHelperContentExtensions`
  - add some special cases in `UrlResolutionTagHelper`
- add a few special cases to avoid `StringWriter` use when value is an `HtmlString`
- move `HtmlTextWriter` optimizations a bit lower and add missing checks for that type

nits:
- make `null` checks of `HtmlTextWriter`s consistent
- correct comments that incorrectly mention `HtmlString`s
- update comments in `JavaScriptStringArrayEncoder`

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
@@ -335,15 +335,15 @@ namespace Microsoft.AspNet.Mvc.Razor
             if (htmlContent != null)
             {
                 var htmlTextWriter = writer as HtmlTextWriter;
-                if (htmlTextWriter != null)
+                if (htmlTextWriter == null)
+                {
+                    htmlContent.WriteTo(writer, encoder);
+                }
+                else
                 {
                     // This special case allows us to keep buffering as IHtmlContent until we get to the 'final'
                     // TextWriter.
                     htmlTextWriter.Write(htmlContent);
-                }
-                else
-                {
-                    htmlContent.WriteTo(writer, encoder);
                 }
 
                 return;
--- a/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
@@ -335,15 +335,15 @@ namespace Microsoft.AspNet.Mvc.Razor
             if (htmlContent != null)
             {
                 var htmlTextWriter = writer as HtmlTextWriter;
-                if (htmlTextWriter == null)
-                {
-                    htmlContent.WriteTo(writer, encoder);
-                }
-                else
+                if (htmlTextWriter != null)
                 {
                     // This special case allows us to keep buffering as IHtmlContent until we get to the 'final'
                     // TextWriter.
                     htmlTextWriter.Write(htmlContent);
+                }
+                else
+                {
+                    htmlContent.WriteTo(writer, encoder);
                 }
 
                 return;
@@ -602,22 +602,10 @@ namespace Microsoft.AspNet.Mvc.Razor
         {
             if (!_tagHelperAttributeInfo.Suppressed)
             {
-                HtmlString htmlString;
-
-                if (_valueBuffer != null)
-                {
-                    using (var stringWriter = new StringWriter())
-                    {
-                        _valueBuffer.Content.WriteTo(stringWriter, HtmlEncoder);
-                        htmlString = new HtmlString(stringWriter.ToString());
-                    }
-                }
-                else
-                {
-                    htmlString = HtmlString.Empty;
-                }
-
-                executionContext.AddHtmlAttribute(_tagHelperAttributeInfo.Name, htmlString);
+                executionContext.AddHtmlAttribute(
+                    _tagHelperAttributeInfo.Name,
+                    _valueBuffer?.Content ?? HtmlString.Empty);
+                _valueBuffer = null;
             }
         }
 
@@ -843,7 +831,7 @@ namespace Microsoft.AspNet.Mvc.Razor
         /// content to the <see cref="HttpResponse.Body"/>.
         /// </summary>
         /// <returns>A <see cref="Task{HtmlString}"/> that represents the asynchronous flush operation and on
-        /// completion returns a <see cref="HtmlString.Empty"/>.</returns>
+        /// completion returns <see cref="HtmlString.Empty"/>.</returns>
         /// <remarks>The value returned is a token value that allows FlushAsync to work directly in an HTML
         /// section. However the value does not represent the rendered content.
         /// This method also writes out headers, so any modifications to headers must be done before
@@ -936,7 +924,7 @@ namespace Microsoft.AspNet.Mvc.Razor
         /// <summary>
         /// Sets antiforgery cookie and X-Frame-Options header on the response.
         /// </summary>
-        /// <returns>A <see cref="HtmlString"/> that returns a <see cref="HtmlString.Empty"/>.</returns>
+        /// <returns><see cref="HtmlString.Empty"/>.</returns>
         /// <remarks> Call this method to send antiforgery cookie token and X-Frame-Options header to client
         /// before <see cref="FlushAsync"/> flushes the headers. </remarks>
         public virtual HtmlString SetAntiforgeryCookieAndHeader()
--- a/src/Microsoft.AspNet.Mvc.Razor/RazorTextWriter.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/RazorTextWriter.cs
@@ -101,13 +101,13 @@ namespace Microsoft.AspNet.Mvc.Razor
         public override void Write(IHtmlContent value)
         {
             var htmlTextWriter = TargetWriter as HtmlTextWriter;
-            if (htmlTextWriter != null)
+            if (htmlTextWriter == null)
             {
-                htmlTextWriter.Write(value);
+                value.WriteTo(TargetWriter, HtmlEncoder);
             }
             else
             {
-                value.WriteTo(TargetWriter, HtmlEncoder);
+                htmlTextWriter.Write(value);
             }
         }
 
--- a/src/Microsoft.AspNet.Mvc.Razor/RazorTextWriter.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/RazorTextWriter.cs
@@ -101,13 +101,13 @@ namespace Microsoft.AspNet.Mvc.Razor
         public override void Write(IHtmlContent value)
         {
             var htmlTextWriter = TargetWriter as HtmlTextWriter;
-            if (htmlTextWriter == null)
+            if (htmlTextWriter != null)
             {
-                value.WriteTo(TargetWriter, HtmlEncoder);
+                htmlTextWriter.Write(value);
             }
             else
             {
-                htmlTextWriter.Write(value);
+                value.WriteTo(TargetWriter, HtmlEncoder);
             }
         }
 
--- a/src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
@@ -4,8 +4,10 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel;
+using System.IO;
 using System.Reflection;
 using System.Text.Encodings.Web;
+using Microsoft.AspNet.Html;
 using Microsoft.AspNet.Mvc.Rendering;
 using Microsoft.AspNet.Mvc.Routing;
 using Microsoft.AspNet.Mvc.ViewFeatures;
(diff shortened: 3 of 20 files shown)


src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -355,13 +356,12 @@ private void BuildFallbackBlock(TagHelperContent builder)
...
-                            JavaScriptResources.GetEmbeddedJavaScript(FallbackJavaScriptResourceName),
-                            JavaScriptEncoder.Encode(FallbackTestProperty),
-                            JavaScriptEncoder.Encode(FallbackTestValue),
-                            JavaScriptStringArrayEncoder.Encode(JavaScriptEncoder, fallbackHrefs)))
+                    .AppendFormat(
+                        CultureInfo.InvariantCulture,
+                        JavaScriptResources.GetEmbeddedJavaScript(FallbackJavaScriptResourceName),
+                        new HtmlString(JavaScriptEncoder.Encode(FallbackTestProperty)),
+                        new HtmlString(JavaScriptEncoder.Encode(FallbackTestValue)),
+                        new HtmlString(JavaScriptStringArrayEncoder.Encode(JavaScriptEncoder, fallbackHrefs)))
```
dougbu: This overall change is basically a wash on @rynowak's BasicViews test site with `asp-fallback-*` attributes added to `<link>` and `<script>` elements. However this simplification alone is negative. Higher numbers in the comparison below all tie directly to the following block of code. I'm leaning toward checking this in and then fixing up the new HtmlAbstractions repo, special-casing `HtmlEncodedString` there and changing `HtmlString` to inherit from `HtmlEncodedString` here. Any objections?
``` c#
public string Format(string format, object arg, IFormatProvider formatProvider)
{
    // This is the case we need to special case. We trust the IHtmlContent instance to do the
    // right thing with encoding.
    var htmlContent = arg as IHtmlContent;
    if (htmlContent != null)
    {
        using (var writer = new StringWriter())
        {
            htmlContent.WriteTo(writer, _encoder);
            return writer.ToString();
        }
    }
    ...
}
```

tag 0
tag 1

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -334,16 +334,16 @@ public static void WriteTo(TextWriter writer, HtmlEncoder encoder, object value)
...
+                if (htmlTextWriter != null)
                 {
                     // This special case allows us to keep buffering as IHtmlContent until we get to the 'final'
                     // TextWriter.
                     htmlTextWriter.Write(htmlContent);
                 }
+                else
+                {
+                    htmlContent.WriteTo(writer, encoder);
+                }
```
rynowak: Why this change? This is just moving deck chairs around

src/Microsoft.AspNet.Mvc.Razor/RazorTextWriter.cs
```diff
@@ -101,13 +101,13 @@ public override void Write(string value)
...
+            if (htmlTextWriter != null)
             {
-                value.WriteTo(TargetWriter, HtmlEncoder);
+                htmlTextWriter.Write(value);
             }
             else
             {
-                htmlTextWriter.Write(value);
+                value.WriteTo(TargetWriter, HtmlEncoder);
             }
```
rynowak: Why?

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
```diff
@@ -154,26 +156,45 @@ protected void ProcessUrlAttribute(string attributeName, TagHelperOutput output)
...
+
+                            IHtmlContent resolvedUrl;
+                            if (TryResolveUrl(stringValue, resolvedUrl: out resolvedUrl))
+                            {
+                                attribute.Value = resolvedUrl;
+                            }
+                            else if (htmlString == null)
+                            {
+                                // Not a ~/ URL. Just avoid re-encoding the attribute value later.
+                                attribute.Value = new HtmlString(stringValue);
```
rynowak: The common case here is that the url isn't a ~url. What type is the `attribute.Value` going to be if the attribute only contains static text?

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
rynowak: My suggestion is that we **never** use `AppendFormat`. 

I don't think the changes you're suggesting are bad, but it will be simpler for us in the long run to just avoid it.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/TagBuilder.cs
```diff
@@ -252,6 +251,15 @@ public void MergeAttribute(string key, string value, bool replaceExisting)
         /// <inheritdoc />
         public void WriteTo(TextWriter writer, HtmlEncoder encoder)
         {
+            var htmlTextWriter = writer as HtmlTextWriter;
+            if (htmlTextWriter != null)
+            {
+                // As a perf optimization, we can buffer this output rather than writing it
+                // out character by character.
+                htmlTextWriter.Write(this);
+                return;
```
rynowak: It would be great to see some impact for these kind of changes. What's the effect of this?

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/HtmlContentViewComponentResult.cs
```diff
@@ -47,13 +47,13 @@ public void Execute(ViewComponentContext context)
...
+            if (htmlWriter != null)
             {
-                EncodedContent.WriteTo(context.Writer, context.HtmlEncoder);
+                htmlWriter.Write(EncodedContent);
             }
             else
             {
-                htmlWriter.Write(EncodedContent);
+                EncodedContent.WriteTo(context.Writer, context.HtmlEncoder);
             }
```
rynowak: Why?

rynowak:
:watch:

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
rynowak: If you want to follow up on this issue one way or another in a separate PR that's fine by me. Either way, a .Format-based API will never be as good as just calling .Append a bunch of times

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
dougbu: Yup, exactly. I was looking at the code special-casing `HtmlTextWriter` and the inconsistencies started to get confusing. This also moves the `var` closer to its use. But I can revert if you dislike.

src/Microsoft.AspNet.Mvc.Razor/RazorTextWriter.cs
dougbu: As above. Also mentioned in the PR / commit description: "make `null` checks of `HtmlTextWriters` consistent".

rynowak: But why? This doesn't make the code better, it's just a stylistic choice.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/JavaScriptStringArrayEncoder.cs
```diff
@@ -1,35 +1,38 @@
...
     /// </summary>
     public static class JavaScriptStringArrayEncoder
     {
         /// <summary>
-        /// Encodes a .NET string array for safe use as a JavaScript array literal, including inline in an HTML file.
+        /// Encodes a <c>string[]</c> for safe use as a JavaScript array literal in many contexts, including
+        /// inline in an HTML file.
         /// </summary>
         public static string Encode(JavaScriptEncoder encoder, string[] values)
+        /// <param name="encoder">
```
pranavkm: Looks like these are misplaced. Also why are we documenting types in the `Internal` namespace?

src/Microsoft.AspNet.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
dougbu: Since this helper specifically targets known URL-holding attributes, I'm not sure not matching is the "common case". Regardless `attribute.Value` is usually set from a `StringCollectionTextWriter.Content` property and therefore has type `StringCollectionTextWriterContent` (at least today). In that case, saving just a `string` allocation and not re-encoding.

This is more important when the attribute value contains a C# expression of course.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
dougbu: Tried calling `.Append()` a bunch of times and it didn't result in a significant improvement. The main issue appeared to be the need to sub-string the JavaScript resource -- the file needs to be valid for minification to work. (As-is `JavaScriptResources.GetEmbeddedJavaScript()` calls `.Replace()` which allocates a similar-length `string`.) I test again...

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/TagBuilder.cs
dougbu: This saves
1. a `StringWriter`
2. a `StringBuilder` that backs the `StringWriter`
3. a `char[]` that backs the `StringBuilder`
4. a `string` that gets allocated when `StringWriter.ToString()` is called.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/HtmlContentViewComponentResult.cs
dougbu: Again "make `null` checks of `HtmlTextWriters` consistent".

dougbu:
:up::date:

dougbu:
#### New numbers
Saving about 5.25 MB in a simple ("BasicViews") tag helper scenario.

##### Before
tag 0
##### After
tag 2

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/JavaScriptResources.cs
```diff
@@ -38,7 +38,8 @@ public static string GetEmbeddedJavaScript(string resourceName)
                 // Load the JavaScript from embedded resource
                 using (var resourceStream = getManifestResourceStream(key))
                 {
-                    Debug.Assert(resourceStream != null, "Embedded resource missing. Ensure 'prebuild' script has run.");
+                    Debug.Assert(resourceStream != null,
+                        "Embedded resource missing. Ensure 'prebuild' script has run.");
```
dougbu: I'll move first parameter to its own line (:sheep:ish).

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/JavaScriptResources.cs
```diff
@@ -49,14 +50,11 @@ public static string GetEmbeddedJavaScript(string resourceName)
...
+
         private static string PrepareFormatString(string input)
         {
-            // Replace unescaped/escaped chars with their equivalent
-            return input.Replace("{", "{{")
-                        .Replace("}", "}}")
-                        .Replace("[[[", "{")
-                        .Replace("]]]", "}");
+            // Remove final ");". Those characters are in the file only to allow minification.
+            return input.Substring(0, input.Length - 2);
```
dougbu: FYI this class is `static` and has a `ConcurrentDictionary` which will cache this substring.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -350,16 +349,17 @@ private void BuildFallbackBlock(TagHelperContent builder)
...
-                        string.Format(
-                            CultureInfo.InvariantCulture,
-                            JavaScriptResources.GetEmbeddedJavaScript(FallbackJavaScriptResourceName),
-                            JavaScriptEncoder.Encode(FallbackTestProperty),
-                            JavaScriptEncoder.Encode(FallbackTestValue),
-                            JavaScriptStringArrayEncoder.Encode(JavaScriptEncoder, fallbackHrefs)))
-                    .AppendHtml("</script>");
+                    .AppendHtml(JavaScriptResources.GetEmbeddedJavaScript(FallbackJavaScriptResourceName))
+                    .AppendHtml("\"")
+                    .AppendHtml(JavaScriptEncoder.Encode(FallbackTestProperty))
```
dougbu: If these intermediate encoded `string`s turn out to be a significant cost, could create an `IHtmlContent` implementation that's similar to `StringHtmlContent` but ignores the `encoder` passed to `WriteTo()` and instead encodes using a `JavaScriptEncoder` passed to the constructor. Likely a useful follow-up but not something I'm doing today.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/JavaScriptResources.cs
rynowak: Commented on the commit about this and it looks like you already answered :+1:

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
rynowak: Yeah, let's revisit only if necessary.

rynowak:
:shipit: from me

dougbu:
1f76b3c

PR #3680, opened 2015-12-02
Add extensibility point to Razor compilation

pakrym:
@davidfowl  @pranavkm

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationContext.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationContext.cs
@@ -1,0 +1,33 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.CodeAnalysis.CSharp;
+
+namespace Microsoft.AspNet.Mvc.Razor.Compilation
+{
+    /// <summary>
+    /// Context object used to pass information about the current Razor page compilation.
+    /// </summary>
+    public class RoslynCompilationContext
+    {
+        /// <summary>
+        /// Constructs a new instance of the <see cref="RoslynCompilationContext"/> type.
+        /// </summary>
+        /// <param name="compilation"><see cref="CSharpCompilation"/> to be set to <see cref="Compilation"/> property.</param>
+        public RoslynCompilationContext(CSharpCompilation compilation)
+        {
+            if (compilation == null)
+            {
+                throw new ArgumentNullException(nameof(compilation));
+            }
+
+            Compilation = compilation;
+        }
+
+        /// <summary>
+        /// Gets or sets the <see cref="CSharpCompilation"/> used for current source file compilation.
+        /// </summary>
+        public CSharpCompilation Compilation { get; set; }
+    }
+}
--- a/src/Microsoft.AspNet.Mvc.Razor/RazorViewEngineOptions.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/RazorViewEngineOptions.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using Microsoft.AspNet.FileProviders;
+using Microsoft.AspNet.Mvc.Razor.Compilation;
 
 namespace Microsoft.AspNet.Mvc.Razor
 {
@@ -15,6 +16,8 @@ namespace Microsoft.AspNet.Mvc.Razor
         private IFileProvider _fileProvider;
 
         private string _configuration;
+
+        private Action<RoslynCompilationContext> _compilationCallback = c => { };
 
         /// <summary>
         /// Get a <see cref="IList{IViewLocationExpander}"/> used by the <see cref="RazorViewEngine"/>.
@@ -63,6 +66,24 @@ namespace Microsoft.AspNet.Mvc.Razor
                     throw new ArgumentException(Resources.ArgumentCannotBeNullOrEmpty, nameof(value));
                 }
                 _configuration = value;
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets the callback that is used to customize Razor compilation
+        /// to change compilation settings you can update <see cref="RoslynCompilationContext.Compilation"/> property.
+        /// Customizations made here would not reflect in tooling (Intellisense).
+        /// </summary>
+        public Action<RoslynCompilationContext> CompilationCallback
+        {
+            get { return _compilationCallback; }
+            set
+            {
+                if (value == null)
+                {
+                    throw new ArgumentNullException(nameof(value));
+                }
+                _compilationCallback = value;
             }
         }
     }
--- a/src/Microsoft.AspNet.Mvc.Razor/RazorViewEngineOptionsSetup.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/RazorViewEngineOptionsSetup.cs
@@ -4,6 +4,8 @@
 using Microsoft.AspNet.FileProviders;
 using Microsoft.AspNet.Hosting;
 using Microsoft.AspNet.Mvc.Razor;
+using Microsoft.AspNet.Mvc.Razor.Compilation;
+using Microsoft.Extensions.CompilationAbstractions;
 using Microsoft.Extensions.OptionsModel;
 using Microsoft.Extensions.PlatformAbstractions;
 
@@ -19,13 +21,13 @@ namespace Microsoft.AspNet.Mvc
         /// </summary>
         /// <param name="applicationEnvironment"><see cref="IApplicationEnvironment"/> for the application.</param>
         /// <param name="hostingEnvironment"><see cref="IHostingEnvironment"/> for the application.</param>
-        public RazorViewEngineOptionsSetup(IApplicationEnvironment applicationEnvironment, IHostingEnvironment hostingEnvironment)
+        public RazorViewEngineOptionsSetup(IApplicationEnvironment applicationEnvironment,
+            IHostingEnvironment hostingEnvironment)
             : base(options => ConfigureRazor(options, applicationEnvironment, hostingEnvironment))
         {
         }
 
-        private static void ConfigureRazor(
-            RazorViewEngineOptions razorOptions,
+        private static void ConfigureRazor(RazorViewEngineOptions razorOptions,
             IApplicationEnvironment applicationEnvironment,
             IHostingEnvironment hostingEnvironment)
         {
--- a/test/Microsoft.AspNet.Mvc.Razor.Test/Compilation/RoslynCompilationServiceTest.cs
+++ b/test/Microsoft.AspNet.Mvc.Razor.Test/Compilation/RoslynCompilationServiceTest.cs
@@ -362,6 +362,45 @@ public class NotRazorPrefixType {}";
                 });
         }
 
+        [Fact]
+        public void Compile_RunsCallback()
+        {
+            var content = "public class MyTestType  {}";
+            var applicationEnvironment = PlatformServices.Default.Application;
+            var libraryExporter = CompilationServices.Default.LibraryExporter;
+
+            var compilerOptionsProvider = new Mock<ICompilerOptionsProvider>();
+            compilerOptionsProvider
+                .Setup(p => p.GetCompilerOptions(
+                    applicationEnvironment.ApplicationName,
+                    applicationEnvironment.RuntimeFramework,
+                    ConfigurationName))
+                .Returns(new CompilerOptions());
+
+            RoslynCompilationContext usedCompilation = null;
+            var mvcRazorHost = new Mock<IMvcRazorHost>();
+            mvcRazorHost.SetupGet(m => m.MainClassNamePrefix)
+                        .Returns(string.Empty);
+
+            var compilationService = new RoslynCompilationService(
+                applicationEnvironment,
+                libraryExporter,
+                compilerOptionsProvider.Object,
+                mvcRazorHost.Object,
+                GetOptions(callback: c => usedCompilation = c));
+
+            var relativeFileInfo = new RelativeFileInfo(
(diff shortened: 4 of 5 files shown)


pakrym:
I wasn't able to move `CompilationFailure` because `ICompilationException` depends on it.

pranavkm:
Could we move `ICompilationException` to Diagnostics?

pakrym:
Dnx would need to reference it, so we'll get circular dependency.

pranavkm:
> Dnx would need to reference it

This would be a `ICompilationException` under a different namespace. I'll file a bug for this since it looks like it's solely a contract between Mvc compilation and diagnostics.

Eilon:
@NTaylorMullen can you also take a look at this when you have a chance?

pakrym:
:up: :date:

src/Microsoft.AspNet.Mvc.Razor/Compilation/DnxRoslynCompilationService.cs
```diff
@@ -0,0 +1,166 @@
...
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.Dnx.Compilation.CSharp;
+using Microsoft.Extensions.CompilationAbstractions;
+using Microsoft.Extensions.PlatformAbstractions;
+using Microsoft.Extensions.OptionsModel;
+using System.Runtime.Versioning;
+
+namespace Microsoft.AspNet.Mvc.Razor.Compilation
+{
+    public class RoslynCompilationContext
```
NTaylorMullen: Move to its own file.

src/Microsoft.AspNet.Mvc.Razor/Compilation/DnxRoslynCompilationService.cs
```diff
@@ -0,0 +1,166 @@
...
+    {
+        public RoslynCompilationContext(CSharpCompilation compilation)
+        {
+            Compilation = compilation;
```
NTaylorMullen: Assuming this can't be `null`. Should `null` check it.

NTaylorMullen:
Talked to @pranavkm. He wants to wait until you talk with David as to if this is the right approach.

Will revisit then.

src/Microsoft.AspNet.Mvc.Razor/Compilation/DnxRoslynCompilationService.cs
```diff
@@ -0,0 +1,155 @@
...
+        /// <param name="environment">The environment for the executing application.</param>
+        /// <param name="loaderAccessor">
+        /// The accessor for the <see cref="IAssemblyLoadContext"/> used to load compiled assemblies.
+        /// </param>
+        /// <param name="libraryManager">The library manager that provides export and reference information.</param>
+        /// <param name="compilerOptionsProvider">
+        /// The <see cref="ICompilerOptionsProvider"/> that provides Roslyn compilation settings.
+        /// </param>
+        /// <param name="host">The <see cref="IMvcRazorHost"/> that was used to generate the code.</param>
+        public DnxCompilationCallback(
```
davidfowl: Why does this need to be extracted? All we need to do is add the hook and call it. This logic doesn't need to be moved into another class.

pakrym: Because i want to isolate library manger and `ICompilerOptionsProvider` from RoslynCompilationServices so they are not required or i still would need to write some kind of shims for them to make MVC work on dotnet

pranavkm: With the hook, you should be able to use a minimal stub (a no-op `ILibraryExporter`) and that should suffice, wouldn't it?

pakrym: We need to decide on goal of this refactoring. I think of it as allow user to use mvc in `dotnet` without dnx without stubing non existent interfaces.

pranavkm: I thought we weren't refactoring. It was just an extensibility point.

pakrym: I thought we were making extensibility point and forwarding all dnx compilation stuff through it, so we can replace it easily.

pakrym:
:up: :date:

src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationContext.cs
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.CodeAnalysis.CSharp;
+
+namespace Microsoft.AspNet.Mvc.Razor.Compilation
+{
+    /// <summary>
+    /// Class used to pass <see cref="CSharpCompilation"/> to compilation callback
```
pranavkm: > Context object used to pass information about the current Razor page compilation.

src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationContext.cs
```diff
@@ -0,0 +1,33 @@
...
+            if (compilation == null)
+            {
+                throw new ArgumentNullException(nameof(compilation));
+            }
+
+            Compilation = compilation;
+        }
+
+        /// <summary>
+        /// <see cref="CSharpCompilation"/> used for current source file compilation
```
pranavkm: > Gets or sets the ...

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngineOptions.cs
```diff
@@ -65,5 +68,25 @@ public string Configuration
...
+        /// </summary>
+        /// <remarks>
+        /// At sturtup, this is initialized to empty lambda <code>(c => {})</code>.
+        /// </remarks>
+        public Action<RoslynCompilationContext> CompilationCallback
+        {
+            get { return _compilationCallback; }
+            set
+            {
+                if (value == null)
```
pranavkm: Think it should be ok for this to be null since it starts off as being null.

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngineOptions.cs
```diff
@@ -65,5 +68,25 @@ public string Configuration
...
         }
+
+        /// <summary>
+        /// Gets or sets callback that is ran before emiting assembly with razor view type
+        /// to change compilation settings you can update <see cref="RoslynCompilationContext.Compilation"/> property
+        /// </summary>
+        /// <remarks>
+        /// At sturtup, this is initialized to empty lambda <code>(c => {})</code>.
+        /// </remarks>
+        public Action<RoslynCompilationContext> CompilationCallback
```
pranavkm: Thoughts on a better name?

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngineOptions.cs
```diff
@@ -65,5 +68,25 @@ public string Configuration
                 _configuration = value;
             }
         }
+
+        /// <summary>
+        /// Gets or sets callback that is ran before emiting assembly with razor view type
```
pranavkm: > Gets or sets the callback that is used to customize Razor compilation?

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngineOptions.cs
```diff
@@ -65,5 +68,25 @@ public string Configuration
...
+        /// to change compilation settings you can update <see cref="RoslynCompilationContext.Compilation"/> property
+        /// </summary>
+        /// <remarks>
+        /// At sturtup, this is initialized to empty lambda <code>(c => {})</code>.
```
pranavkm: `startup`

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngineOptionsSetup.cs
```diff
@@ -38,6 +40,7 @@ public RazorViewEngineOptionsSetup(IApplicationEnvironment applicationEnvironmen
             {
                 razorOptions.Configuration = "Release";
             }
+            razorOptions.CompilationCallback = c => { };
```
pranavkm: You could do this inline if it's meant to be non-null.

test/Microsoft.AspNet.Mvc.Razor.Test/Compilation/RoslynCompilationServiceTest.cs
```diff
@@ -373,12 +412,13 @@ private static DiagnosticDescriptor GetDiagnosticDescriptor(string messageFormat
                 isEnabledByDefault: true);
         }
 
-        private static IOptions<RazorViewEngineOptions> GetOptions(IFileProvider fileProvider = null)
+        private static IOptions<RazorViewEngineOptions> GetOptions(IFileProvider fileProvider = null, Action<RoslynCompilationContext> callback = null)
```
pranavkm: Split these across two lines (we use a 120 char limit)

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngineOptions.cs
pranavkm: Nevermind. Didn't see that this got instantiated to a value in the other type.

pranavkm:
:watch:

pakrym:
:up: :date:

src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationContext.cs
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.CodeAnalysis.CSharp;
+
+namespace Microsoft.AspNet.Mvc.Razor.Compilation
+{
+    /// <summary>
+    /// Context object used to pass information about the current Razor page compilation
```
NTaylorMullen: For consistency, end your doc comments with periods everywhere.

src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationContext.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// </summary>
+    public class RoslynCompilationContext
+    {
+        /// <summary>
+        /// Constructs a new instance of <see cref="RoslynCompilationContext"/> type
```
NTaylorMullen: `... instance of the...`

src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationContext.cs
```diff
@@ -0,0 +1,33 @@
...
+                throw new ArgumentNullException(nameof(compilation));
+            }
+
+            Compilation = compilation;
+        }
+
+        /// <summary>
+        /// Gets or sets the <see cref="CSharpCompilation"/> used for current source file compilation
+        /// </summary>
+        public CSharpCompilation Compilation { get; set; }
```
NTaylorMullen: Why do we need to have the Compilation settable?

NTaylorMullen: Ahh, so the compilation callback. Gotcha.

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngineOptions.cs
```diff
@@ -65,5 +68,25 @@ public string Configuration
                 _configuration = value;
             }
         }
+
+        /// <summary>
+        /// Gets or sets the callback that is used to customize Razor compilation
+        /// to change compilation settings you can update <see cref="RoslynCompilationContext.Compilation"/> property
+        /// </summary>
+        /// <remarks>
+        /// At startup, this is initialized to empty lambda <code>(c => {})</code>.
```
NTaylorMullen: No need for the remarks. You cover this in the summary by stating it's used to `customize`. `Callback` also indicates that it doesn't do anything.

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngineOptions.cs
```diff
@@ -65,5 +68,22 @@ public string Configuration
                 _configuration = value;
             }
         }
+
+        /// <summary>
+        /// Gets or sets the callback that is used to customize Razor compilation
```
NTaylorMullen: Should definitely add a comment stating that customizations here will not reflect in tooling. It'd normally go in remarks but users wouldn't be able to see it in IntelliSense.

NTaylorMullen:
:shipit: once you've addressed [this](https://github.com/aspnet/Mvc/pull/3680#discussion_r46601960).

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngineOptions.cs
pakrym: `/// Customizations made here would not reflect in tooling (Intellisense)` sound ok?

NTaylorMullen: :+1:

src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationContext.cs
Eilon: It's not for consistency, it's for correctness! These are English sentences

PR #3687, opened 2015-12-04
Use Routing.Abstractions

rynowak:
FYI only


src/Microsoft.AspNet.Mvc.Abstractions/project.json
```diff
@@ -26,7 +26,8 @@
     "Microsoft.Extensions.PropertyHelper.Sources": {
       "version": "1.0.0-*",
       "type": "build"
-    }
+    },
+    "Microsoft.Net.Http.Headers": "1.0.0-*"
```
rynowak: This is a leaner dependency set for Mvc.Abstractions

Issue #3633, opened 2015-11-24
Compilation big rocks moving to dotnet CLI


davidfowl:
Now that we don't have the DNX providing compilation services, we need to figure out how to handle a few things:

- [x] Getting compilation options
- [x] Getting reference assemblies
- [x] Remove dependency on `Microsoft.Extensions.CompilationAbstractions`

/cc @anurse @pakrym

pakrym:
And target framework too

davidfowl:
@pakrym target framework will come from IApplicationEnvironment. Once we have the CLR feature then it'll just work.

pakrym:
I was able to run MVC with dotnet cli by:
1. Overriding `IAssemblyProvider` to `StaticAssemblyProvider` with entry assembly added (At this point it was able to find controllers) 
2. Adding naive implementation of `ICompilerOptionsProvider`
3. Adding naive implementation of `ILibraryExporter`
3. Overriding `IApplicationEnvironment` and returning not null `RuntimeFramework`

rynowak:
nice!

pakrym:
Compilation options provider shouldn't be a big problem, we can infer most of settings at runtime from `IHostringEnvironment` and `IRuntimeEnvironment`

pakrym:
For `ILibraryExporter` we can Embed `project.json` into assembly (something like "embedProjectMetadata" compilation option) and  and add API to read from it, so MVC can implement its own logic on top

davidfowl:
Well that one is a bit tricky. The dotnet cli would need to do that and it's bit weird to always do it. Also you need the lock file not just the project.json file. Also you need access to the reference assemblies, which aren't always available unless we make them available.

pranavkm:
@pakrym wouldn't we need the project.json to look at things like LanguageVersion, Defines, WarningAsErrors?

pakrym:
Or we hardcode them to some default values :)

pranavkm:
At that point, isn't it much simpler to just have an alternate way of configuring views (e.g. via options). It also separates oddities like figuring out which properties to exclude (code signing, `emitEntryPoint`) from the primary project and plays nicely with binaries containing embedded cshtml. @davidfowl doesn't particularly like this solution though.

davidfowl:
I think we can have everything on razor options. All we're talking about is defaults. Let's start by treating the razor view engine as a separate compilation system that can be fully configured in Startup. Then we figure out where the defaults come from.

[...]

PR #3693, opened 2015-12-06
Remove dependency on ICompilationOptionsProvider

davidfowl:
- Removed Configuration from RazorViewEngineOptions
- Added ParseOptions and CompilationOptions to RazorViewEngineOptions

#3633

/cc @pranavkm @pakrym 

There's a broken test that shows flowing compilation options from the project file. I can update the functional test project to support passing setting the right preprocessor symbols at runtime.

EDIT:

Update the test

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationService.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationService.cs
@@ -38,7 +38,6 @@ namespace Microsoft.AspNet.Mvc.Razor.Compilation
         private readonly IFileProvider _fileProvider;
         private readonly Lazy<List<MetadataReference>> _applicationReferences;
         private readonly string _classPrefix;
-        private readonly string _configuration;
         private readonly Action<RoslynCompilationContext> _compilationCallback;
         private readonly CSharpParseOptions _parseOptions;
         private readonly CSharpCompilationOptions _compilationOptions;
--- a/src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationService.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationService.cs
@@ -70,8 +70,8 @@ namespace Microsoft.AspNet.Mvc.Razor.Compilation
             _fileProvider = optionsAccessor.Value.FileProvider;
             _classPrefix = host.MainClassNamePrefix;
             _compilationCallback = optionsAccessor.Value.CompilationCallback;
-            _parseOptions = optionsAccessor.Value.ParseOptions ?? new CSharpParseOptions(LanguageVersion.CSharp6);
-            _compilationOptions = optionsAccessor.Value.CompilationOptions ?? new CSharpCompilationOptions(OutputKind.DynamicallyLinkedLibrary);
+            _parseOptions = optionsAccessor.Value.ParseOptions;
+            _compilationOptions = optionsAccessor.Value.CompilationOptions;
 
 
 #if DOTNET5_5
--- a/src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationService.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationService.cs
@@ -35,12 +35,13 @@ namespace Microsoft.AspNet.Mvc.Razor.Compilation
 
         private readonly ILibraryExporter _libraryExporter;
         private readonly IApplicationEnvironment _environment;
-        private readonly ICompilerOptionsProvider _compilerOptionsProvider;
         private readonly IFileProvider _fileProvider;
         private readonly Lazy<List<MetadataReference>> _applicationReferences;
         private readonly string _classPrefix;
         private readonly string _configuration;
-        private Action<RoslynCompilationContext> _compilationCallback;
+        private readonly Action<RoslynCompilationContext> _compilationCallback;
+        private readonly CSharpParseOptions _parseOptions;
+        private readonly CSharpCompilationOptions _compilationOptions;
 
 #if DOTNET5_5
         private readonly RazorLoadContext _razorLoadContext;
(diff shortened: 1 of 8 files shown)


src/Microsoft.AspNet.Mvc.Razor/RazorViewEngineOptions.cs
```diff
@@ -15,7 +16,9 @@ public class RazorViewEngineOptions
     {
         private IFileProvider _fileProvider;
 
-        private string _configuration;
+        private CSharpParseOptions _parseOptions = new CSharpParseOptions(LanguageVersion.CSharp6);
```
pranavkm: `readonly`

pranavkm: nevermnd. It's settable.

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngineOptions.cs
```diff
@@ -49,42 +52,54 @@ public IFileProvider FileProvider
...
         }
 
         /// <summary>
-        /// Gets or sets the callback that is used to customize Razor compilation
-        /// to change compilation settings you can update <see cref="RoslynCompilationContext.Compilation"/> property.
-        /// Customizations made here would not reflect in tooling (Intellisense).
+        /// Gets or sets the <see cref="CSharpParseOptions"/> options used by Razor view compilation.
         /// </summary>
-        public Action<RoslynCompilationContext> CompilationCallback
+        public CSharpParseOptions ParseOptions
```
pranavkm: Granted we can make this work, but is there benefit in letting users customize this? Razor parser had to be modified to switch to C# 6, so changing the language version here might not be sufficient.

src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationService.cs
```diff
@@ -258,13 +254,16 @@ private List<MetadataReference> GetApplicationReferences()
...
-                // Microsoft.Extensions.Runtime.Roslyn/RoslynCompiler.cs#L164
-                // We don't want to take a dependency on the Roslyn bit directly since it pulls in more dependencies
-                // than the view engine needs (Microsoft.Extensions.Runtime) for example
-                references.Add(ConvertMetadataReference(metadataReference));
+                foreach (var metadataReference in export.MetadataReferences)
+                {
+                    // Taken from https://github.com/aspnet/KRuntime/blob/757ba9bfdf80bd6277e715d6375969a7f44370ee/src/...
+                    // Microsoft.Extensions.Runtime.Roslyn/RoslynCompiler.cs#L164
+                    // We don't want to take a dependency on the Roslyn bit directly since it pulls in more dependencies
+                    // than the view engine needs (Microsoft.Extensions.Runtime) for example
```
rynowak: Is this comment still relevant?

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngineOptions.cs
pakrym: Defines are also there

davidfowl: Defines and language version

src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationService.cs
davidfowl: sorta, with renames

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngineOptions.cs
pranavkm: Are those interesting though? Do we expect people to use `#ifdef` in a view?

davidfowl: Don't see why that matters? What if they do? This basically exposes everything that we use as part of the compilation process.

We at least have a single test that uses it :smile:

davidfowl:
:alien:

pakrym:
Precompilation gets out of sync but i guess we don't care for it much at the moment?

davidfowl:
No we don't care.

davidfowl:
Plus the only way to make it never go out of sync is to only have it be configurable via a file.

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngineOptionsSetup.cs
```diff
@@ -22,7 +22,7 @@ public class RazorViewEngineOptionsSetup : ConfigureOptions<RazorViewEngineOptio
         /// <param name="applicationEnvironment"><see cref="IApplicationEnvironment"/> for the application.</param>
         /// <param name="hostingEnvironment"><see cref="IHostingEnvironment"/> for the application.</param>
         public RazorViewEngineOptionsSetup(IApplicationEnvironment applicationEnvironment,
-            IHostingEnvironment hostingEnvironment)
+                                           IHostingEnvironment hostingEnvironment)
```
pakrym: nit: default VS whitespace

pakrym:
LGTM

src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationService.cs
```diff
@@ -258,13 +254,16 @@ private List<MetadataReference> GetApplicationReferences()
             }
 
             var export = _libraryExporter.GetAllExports(_environment.ApplicationName);
-            foreach (var metadataReference in export.MetadataReferences)
+            if (export != null)
```
NTaylorMullen: Curious: Why'd we need to add this?

src/Microsoft.AspNet.Mvc.Razor/Compilation/SyntaxTreeGenerator.cs
```diff
@@ -17,6 +17,19 @@ public static class SyntaxTreeGenerator
             string path,
             CompilationSettings compilationSettings)
```
NTaylorMullen: Looking at this now, it doesn't really make sense to have a method that doesn't do any compilation take in compilation settings. It'd probably be best removing this overload and just rely on the parseoptions variations.

davidfowl: Eventually

src/Microsoft.AspNet.Mvc.Razor/Compilation/RoslynCompilationService.cs
davidfowl: Because it can be null.

NTaylorMullen: It couldn't before?

Issue #3670, opened 2015-12-01
Add test coverage for Flags enum binding in integration tests


kichalla:
As part of task https://github.com/aspnet/Mvc/issues/3624, the ValueProvidersWebSite is being deleted. Currently it covers some scenarios related to binding to Flags enum based parameters...I checked to see if there is existing coverage for this in integration tests but couldn't find any. So this task is to keep track of it.

rynowak:
This is also a good candidate for the unit tests of `SimpleTypeModelBinder`

kichalla:
ceaa9a92514bf4af87b5a0fae0a6c7e74c23b900

PR #3700, opened 2015-12-07
Add test coverage for Flags enum binding

kichalla:
Fixes: Add test coverage for Flags enum binding in integration tests #3670

@pranavkm @rynowak

### DIFF
--- a/test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/SimpleTypeModelBinderTest.cs
+++ b/test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/SimpleTypeModelBinderTest.cs
@@ -189,6 +189,32 @@ namespace Microsoft.AspNet.Mvc.ModelBinding.Test
             Assert.True(bindingContext.ModelState.ContainsKey("theModelName"));
         }
 
+        [Theory]
+        [InlineData("0", 0)]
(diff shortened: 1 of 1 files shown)


kichalla:
@dougbu

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/SimpleTypeModelBinderTest.cs
```diff
@@ -189,6 +189,31 @@ public static TheoryData<Type> ConvertableTypeData
...
+            };
+
+            var binder = new SimpleTypeModelBinder();
+
+            // Act
+            var result = await binder.BindModelAsync(bindingContext);
+
+            // Assert
+            Assert.True(result.IsModelSet);
+            Assert.Equal(expected, result.Model.ToString());
```
rynowak: Do we need to tostring here? can't we just compare to the enum value? (Pass in expected as an int and cast it)

kichalla: :+1:

kichalla:
:up: :date:

rynowak:
:shipit:

Issue #3676, opened 2015-12-02
Model binding sometimes gives NREs


Tragetaschen:
I'm not sure what I'm seeing here, but my code definitely hasn't been reached yet.

I have the following Action defined on a controller:
```csharp
[HttpPost("UploadForm")]
public async Task<IActionResult> UploadForm([FromBody] JObject json)
```

With rc1-final on Linux, my application can reach two states
- Every request to this method succeeds
- Every request to this method fails with
```
      An unhandled exception has occurred while executing the request
      System.NullReferenceException: Object reference not set to an instance of an object
        at Microsoft.AspNet.Mvc.ModelBinding.Validation.DefaultComplexObjectValidationStrategy+Enumerator.MoveNext () [0x00000] in <filename unknown>:0
        at Microsoft.AspNet.Mvc.ModelBinding.Validation.ValidationVisitor.VisitChildren (IValidationStrategy strategy) [0x00000] in <filename unknown>:0
        at Microsoft.AspNet.Mvc.ModelBinding.Validation.ValidationVisitor.VisitComplexType () [0x00000] in <filename unknown>:0
        at Microsoft.AspNet.Mvc.ModelBinding.Validation.ValidationVisitor.Visit (Microsoft.AspNet.Mvc.ModelBinding.ModelMetadata metadata, System.String key, System.Object model) [0x00000] in <filename unknown>:0
        at Microsoft.AspNet.Mvc.ModelBinding.Validation.ValidationVisitor.VisitChildren (IValidationStrategy strategy) [0x00000] in <filename unknown>:0
        at Microsoft.AspNet.Mvc.ModelBinding.Validation.ValidationVisitor.VisitEnumerableType () [0x00000] in <filename unknown>:0
        at Microsoft.AspNet.Mvc.ModelBinding.Validation.ValidationVisitor.Visit (Microsoft.AspNet.Mvc.ModelBinding.ModelMetadata metadata, System.String key, System.Object model) [0x00000] in <filename unknown>:0
        at Microsoft.AspNet.Mvc.ModelBinding.Validation.ValidationVisitor.Validate (Microsoft.AspNet.Mvc.ModelBinding.ModelMetadata metadata, System.String key, System.Object model) [0x00000] in <filename unknown>:0
[...]

PR #3707, opened 2015-12-08
Partial fix for #3676 - fix race in ElementMetadata

rynowak:
The accessor for ElementMetadata can sometimes return null when
concurrently accessed. This change solves this issue by precomputing all
type-related facets of ModelMetadata. This also adds a ElementType
property to ModelMetadata to maintain the current separation of concerns
as ModelMetadata is unaware of the metadata provider.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelMetadata.cs
+++ b/src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelMetadata.cs
@@ -18,8 +18,6 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
     [DebuggerDisplay("{DebuggerToString(),nq}")]
     public abstract class ModelMetadata
     {
-        private bool? _isComplexType;
-
         /// <summary>
         /// The default value of <see cref="ModelMetadata.Order"/>.
         /// </summary>
@@ -32,6 +30,8 @@ namespace Microsoft.AspNet.Mvc.ModelBinding
         protected ModelMetadata(ModelMetadataIdentity identity)
         {
             Identity = identity;
+
+            InitializeTypeInformation();
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelMetadata.cs
```diff
@@ -415,6 +368,47 @@ public string GetDisplayName()
...
+            IsReferenceOrNullableType = !ModelType.GetTypeInfo().IsValueType || IsNullableValueType;
+            UnderlyingOrModelType = Nullable.GetUnderlyingType(ModelType) ?? ModelType;
+
+            var collectionType = ClosedGenericMatcher.ExtractGenericInterface(ModelType, typeof(ICollection<>));
+            IsCollectionType = collectionType != null;
+
+            if (ModelType == typeof(string) || !typeof(IEnumerable).IsAssignableFrom(ModelType))
+            {
+                IsEnumerableType = false;
+                ElementType = null;
```
dougbu: Why redo defaults? Prefer leaving this block empty.

dougbu:
A simpler fix for the race part of #3676 would move a single line to the end of the `ElementMetadata` getter. Need to improve the description to cover what you're actually doing: Setting all `Type`-related properties in `ModelMetadata` up front, removing lazy initialization of a half-dozen or so properties. I don't object to the change but it is only "inspired by #3676".

dougbu:
:shipit: with broader commit description

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelMetadata.cs
rynowak: Sure

rynowak:
>  This change solves this issue by precomputing all type-related facets of ModelMetadata.

Already in the commit description. I'll update it to be clearer.

PR #3715, opened 2015-12-09
Move buffer types to ViewFeatures

pranavkm:
Use buffer pooling in more places


pranavkm:
cc @rynowak

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/HtmlHelperOfT.cs
```diff
@@ -21,35 +22,19 @@ public class HtmlHelper<TModel> : HtmlHelper, IHtmlHelper<TModel>
...
             JavaScriptEncoder javaScriptEncoder)
-            : base(htmlGenerator, viewEngine, metadataProvider, htmlEncoder, urlEncoder, javaScriptEncoder)
+            : base(
+            htmlGenerator,
+            viewEngine,
+            metadataProvider,
+            bufferScope,
+            htmlEncoder,
+            urlEncoder,
+            javaScriptEncoder)
```
rynowak: Indent

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/DefaultTemplatesUtilities.cs
```diff
@@ -276,14 +264,22 @@ public static HtmlHelper<TModel> GetHtmlHelper<TModel>(TModel model, IModelMetad
...
+
+            var serviceProvider = new ServiceCollection()
+               .AddSingleton(viewEngine)
+               .AddSingleton(urlHelperFactory.Object)
+               .AddSingleton(Mock.Of<IViewComponentHelper>())
+               .AddSingleton(innerHelper)
+               .AddSingleton<IViewBufferScope, TestViewBufferScope>()
+               .BuildServiceProvider();
+
+            httpContext.RequestServices = serviceProvider;
```
rynowak: :clap:

rynowak:
:shipit:

rynowak:
But let's do a follow up to implement chunking before merging this.

pranavkm:
Updated

src/Microsoft.AspNet.Mvc.ViewFeatures/Buffer/MemoryPoolViewBufferScope.cs
```diff
@@ -0,0 +1,80 @@
...
+using System.Collections.Generic;
+
+namespace Microsoft.AspNet.Mvc.ViewFeatures.Buffer
+{
+    /// <summary>
+    /// A <see cref="IViewBufferScope"/> that uses pooled memory.
+    /// </summary>
+    public class MemoryPoolViewBufferScope : IViewBufferScope, IDisposable
+    {
+        public static readonly int SegmentSize = 512;
```
pranavkm: I dropped the buffer size since this it resulted in lower allocations.

src/Microsoft.AspNet.Mvc.ViewFeatures/Buffer/MemoryPoolViewBufferScope.cs
```diff
@@ -0,0 +1,80 @@
...
+            if (_leased == null)
+            {
+                _leased = new List<ViewBufferValue[]>(1);
+            }
+
+            ViewBufferValue[] segment = null;
+
+            try
+            {
+                segment = _pool.Rent(SegmentSize);
```
pranavkm: not super sure on this - it says Rent(minimumBufferSize). Does this mean we could end up with a larger buffer here and we'd have to do sub-chunking?

src/Microsoft.AspNet.Mvc.ViewFeatures/DependencyInjection/MvcViewFeaturesMvcCoreBuilderExtensions.cs
```diff
@@ -136,6 +139,11 @@ internal static void AddViewServices(IServiceCollection services)
             // These are stateless so their lifetime isn't really important.
             services.TryAddSingleton<ITempDataDictionaryFactory, TempDataDictionaryFactory>();
             services.TryAddSingleton<SaveTempDataFilter>();
+
+            services.TryAddSingleton(ArrayPool<ViewBufferValue>.Create(MemoryPoolViewBufferScope.SegmentSize, 50));
```
pranavkm: The default array is pretty crazy long: `private const int DefaultArrayLength = 1048576;` based on reflection. Is the idea that you're supposed to perform sub chunking on the one array that is pooled?

pranavkm:
Also allocations: This accounted for about ~1% in allocations

Before (Total 130998 kb)
before

After (Total 129348 kb)
after

rynowak:
That's a good result for this size of benchmark.

src/Microsoft.AspNet.Mvc.ViewFeatures/DependencyInjection/MvcViewFeaturesMvcCoreBuilderExtensions.cs
rynowak: That's the default **maximum** that will be retained by the pool. You can ask for any size you want, the pool will create multiple sizes.

src/Microsoft.AspNet.Mvc.ViewFeatures/Buffer/MemoryPoolViewBufferScope.cs
rynowak: It means you get at-least N. Nothing requires you to do chunking.

src/Microsoft.AspNet.Mvc.ViewFeatures/Buffer/MemoryPoolViewBufferScope.cs
rynowak: Yeah, we could go pretty low with this. I'd say let's check in this design change, and then experiment with tuning this.

src/Microsoft.AspNet.Mvc.ViewFeatures/Buffer/MemoryPoolViewBufferScope.cs
rynowak: We probably don't need to implement chunking since the pool will go pretty low if we need it to. Default minimum is 16.

src/Microsoft.AspNet.Mvc.ViewFeatures/DependencyInjection/MvcViewFeaturesMvcCoreBuilderExtensions.cs
```diff
@@ -136,6 +139,11 @@ internal static void AddViewServices(IServiceCollection services)
             // These are stateless so their lifetime isn't really important.
             services.TryAddSingleton<ITempDataDictionaryFactory, TempDataDictionaryFactory>();
             services.TryAddSingleton<SaveTempDataFilter>();
+
+            services.TryAddSingleton(ArrayPool<ViewBufferValue>.Shared);
+            // This manages buffer references for the lifetime of a single request. This must be scoped to prevent
+            // memory leaks.
+            services.TryAddScoped<IViewBufferScope, MemoryPoolViewBufferScope>();
```
rynowak: It won't result in a leak with the default pool, it just gives us more optimal perf.

rynowak:
:shipit: again

src/Microsoft.AspNet.Mvc.ViewFeatures/DependencyInjection/MvcViewFeaturesMvcCoreBuilderExtensions.cs
pranavkm: Ok. I'll get rid of the comment.

PR #3719, opened 2015-12-10
Tweak CacheTagHelper sample to use a token provider service for

### LINKED ISSUE #3573: CacheTagHelper cache invalidation outside view
While we're still in beta, I wanted to ask if there will be a way to manually remove items from the cache like the old RemoveOutputCacheItem method.  Although the question below is related to OutputCache it does show how frustrating it can be to invalidate a cache without knowning the key.

http://stackoverflow.com/questions/5326230/mvc3-outputcache-removeoutputcacheitem-renderaction

In Mvc/src/Microsoft.AspNet.Mvc.TagHelpers/CacheTagHelper.cs I see that context.UniqueId is used for part of the key, but it's created at view compilation time.  Would it be possible to allow a custom key to be provided as an optional attribute which could be used instead?

It may not be necessary as there may be workarounds, but I wanted to float the idea in case this scenario had not already been considered.

pranavkm:
cache expiration.

Fixes #3573

### DIFF
--- a/test/WebSites/HtmlGenerationWebSite/CancellableChangeToken.cs
+++ b/test/WebSites/HtmlGenerationWebSite/CancellableChangeToken.cs
@@ -1,0 +1,20 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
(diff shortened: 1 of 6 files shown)


pranavkm:
cc @lodejard

PR #3729, opened 2015-12-10
Update NuGet.config

tuespetre:
Make it very straightforward to do an initial build. Example: I had an `AspNetVNext` package source configured in my global NuGet.config file, and I had also placed an entry for it under `disabledPackageSources`. When I cloned the Mvc repository and tried to build, I was getting errors that dependencies could not be resolved. Either adding explicit `remove` entries to the `disabledPackageSources` element in this file or changing the key in this file from `AspNetVNext` to `AspNetCiDev` would prevent that from happening; the approach I took here was to just completely tighten the local NuGet configuration to only use these two feeds and clear all disabled package source entries.

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -1,7 +1,11 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
-    <add key="AspNetVNext" value="https://www.myget.org/F/aspnetcidev/api/v3/index.json" />
+    <clear />
+    <add key="AspNetCiDev" value="https://www.myget.org/F/aspnetcidev/api/v3/index.json" />
     <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
   </packageSources>
+  <disabledPackageSources>
+    <clear />
+  </disabledPackageSources>
 </configuration>


dougbu:
@tuespetre MVC developers often use `$env:appdata\NuGet\NuGet.config` to override sources without needing to edit a checked-in file. The latest ASP.NET 5 project templates don't include `NuGet.config` files. Why move in the other direction?

Separately please open an issue for discussion. If you'd like to contribute code, the [up-for-grabs bugs label](https://github.com/aspnet/Mvc/issues?q=is%3Aopen+is%3Aissue+label%3Aup-for-grabs) usually applies to more than one approved issue.

davidfowl:
@tuespetre We don't make changes like this to individual repos. It needs to be done across all repositories.

@pranavkm Would this break our CI build?

pranavkm:
Yeah, we use `AspNetVNext` to do the feed hijacking on the feed (as also a bunch of scripts that replace the value during branching). Clearing it would be ok although I'm not sure if it would affect other dev's workflows.

tuespetre:
@davidfowl:

> We don't make changes like this to individual repos. It needs to be done across all repositories.

That makes sense.

@dougbu:

> Separately please open an issue for discussion.

Sorry, I thought it to be small enough to just discuss within the PR. I will do that from now on. I now realize that I probably wasted build server time.

pranavkm:
> I had an AspNetVNext package source configured in my global NuGet.config file, and I had also placed an entry for it under disabledPackageSources.

I think NuGet fixed the issue over hierarchical priorties in recent releases. Updating dnu to do this \ using NuGet to restore would resolve this.

Issue #3094, opened 2015-09-05
Improve controller authoring experience for AddMvcCore


rynowak:
See this comment thread: https://github.com/aspnet/Mvc/issues/3083#issuecomment-137599872

Right now it's confusing if you do `AddMvcCore()` and then try to use `Controller`. You'll get an exception about `ITempData` not being found in the service provider.

rynowak:
a2393f21be55c5c02ea67df55407aa7585121173

PR #3734, opened 2015-12-10
Adds ControllerBase to Mvc.Core

rynowak:
This change adds a base class for controllers to Mvc.Core that can be used
without a dependency on views, json, etc.

Most of the functionality on Controller moves to ControllerBase. I kept
the IActionFilter and IDisposable functionality on Controller since it's
not really a fit with the 'minimal' philosophy.


rynowak:
Fixes #3094

src/Microsoft.AspNet.Mvc.Core/ControllerBase.cs
```diff
@@ -0,0 +1,1451 @@
...
+using Microsoft.AspNet.Mvc.ModelBinding.Validation;
+using Microsoft.AspNet.Mvc.Routing;
+using Microsoft.AspNet.Routing;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNet.Mvc
+{
+    /// <summary>
+    /// Base class for an MVC controller.
```
Eilon: Need to have a different description from `Controller`. In fact, should update both of them to say something unique about what each one is (and perhaps isn't).

Eilon:
For some reason GitHub isn't showing the whole diff, but I assume that 99.9% of this is cut & paste from Controller to ControllerBase.

rynowak:
> 23 additions, 1,458 deletions not shown

That's what github has to say about the changes to `Controller.cs`. It's all just cut-paste

Eilon:
Sure :shipit: after fixing comment. But I'd like to see the updated comments first :smile:

rynowak:
Updated

PR #3737, opened 2015-12-11
Fix localization tests to test only MVC specific services

kichalla:
@pranavkm looks like the CI finally caught up and these tests are failing

cc @DamianEdwards

### DIFF
--- a/samples/MvcSandbox/wwwroot/_references.js
+++ b/samples/MvcSandbox/wwwroot/_references.js
@@ -1,3 +1,5 @@
 ﻿/// <autosync enabled="true" />
 /// <reference path="../gulpfile.js" />
 /// <reference path="js/site.js" />
+/// <reference path="lib/jquery/dist/jquery.js" />
+/// <reference path="lib/jquery-validation-unobtrusive/jquery.validate.unobtrusive.js" />
--- a/test/Microsoft.AspNet.Mvc.Localization.Test/Internal/MvcLocalizationServiceCollectionExtensionsTest.cs
+++ b/test/Microsoft.AspNet.Mvc.Localization.Test/Internal/MvcLocalizationServiceCollectionExtensionsTest.cs
@@ -24,7 +24,7 @@ namespace Microsoft.AspNet.Mvc.Localization.Internal
             var collection = new ServiceCollection();
 
             // Act
-            MvcLocalizationServices.AddLocalizationServices(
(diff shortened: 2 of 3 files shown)


samples/MvcSandbox/wwwroot/_references.js
```diff
@@ -1,3 +1,5 @@
 ﻿/// <autosync enabled="true" />
 /// <reference path="../gulpfile.js" />
 /// <reference path="js/site.js" />
+/// <reference path="lib/jquery/dist/jquery.js" />
```
kichalla: will revert these

test/Microsoft.AspNet.Mvc.Localization.Test/Internal/MvcLocalizationServiceCollectionExtensionsTest.cs
```diff
@@ -53,23 +53,6 @@ public void AddLocalizationServices_AddsNeededServices()
                     Assert.Equal(typeof(IViewLocalizer), service.ServiceType);
                     Assert.Equal(typeof(ViewLocalizer), service.ImplementationType);
                     Assert.Equal(ServiceLifetime.Transient, service.Lifetime);
-                },
-                service =>
```
kichalla: These should be verified in the Localization repo tests.

DamianEdwards:
Does this explain why the tests in the Localization repo are also failing?

DamianEdwards:
Looks like this changed caused the Localization break: https://github.com/aspnet/Options/commit/ec13249a35bbf1d4a8a381a9697965cc1208fe82

kichalla:
@DamianEdwards These tests are trying to verify the services which the `AddOptions` extension would also add. I think these tests should not verify the services that `AddOptions` is adding as it can change anytime and this is the reason they are currently failing due to an addition of a service `OptionsMonitor`

DamianEdwards:
Yeah agreed. Now you say that, I should probably change how I fixed the Localization tests to match that idea.

kichalla:
@DamianEdwards Can I get a signoff so that I can push the changes in?

dougbu:
@kichalla why are you waiting to fix a build break that's affecting the CI? In any case :shipit:

kichalla:
Sure, pushed the change now f2bb90fa55cd18acd8fde77737e8d29804fe66ec

PR #3744, opened 2015-12-11
Enhance javascript fallback-test

nguymin4:
If there is error while evaluating the test then use the fallback source instead of throwing error.


pranavkm:
Why not make the try-catch part of the evaluation? Or create an expression that doesn't throw? This seems unnecessary.

nguymin4:
At the moment I did it like this
```<script src="~/js/site.min.js" asp-file-version="true" 
				asp-fallback-src="~/js/site.js"
				asp-fallback-test="(function(){try {return angular.module('myApp')} catch(e) {return false}}())">```
But then it's repetitive and not easy to read. But ok I understand your point.

pranavkm:
cc @DamianEdwards

Eilon:
Yeah I don't think we want the framework to hide "bad code". Bad code should produce errors.

PR #3747, opened 2015-12-11
Last part of #3676

rynowak:
Uses the correct IEnumerable<> in validation strategies


src/Microsoft.AspNet.Mvc.Core/ModelBinding/Validation/DefaultCollectionValidationStrategy.cs
```diff
@@ -116,7 +130,7 @@ public void Dispose()
 
             public void Reset()
             {
-                throw new NotImplementedException();
+                _enumerator.Reset();
```
dougbu: bonus :clap:

dougbu:
:shipit:

PR #3758, opened 2015-12-14
React to breaking in IRouteConstraint

rynowak:


rynowak:
FYI only

src/Microsoft.AspNet.Mvc.Core/Routing/KnownRouteValueConstraint.cs
```diff
@@ -43,20 +42,20 @@ public class KnownRouteValueConstraint : IRouteConstraint
...
-                                                                    StringComparison.OrdinalIgnoreCase));
-
-                    return match;
+                    foreach (var existingValue in allValues)
+                    {
+                        if (string.Equals(value, existingValue, StringComparison.OrdinalIgnoreCase))
+                        {
+                            return true;
+                        }
+                    }
```
rynowak: This is linq on a pretty hot path. We haven't been testing with this constraint, but it's definitely worth fixing.

PR #3761, opened 2015-12-14
[Fixes #3705] Bring back render partial

### LINKED ISSUE #3705: Html Helper missing RenderPartial
Mvc/src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/IHtmlHelper.cs

Does not contain RenderPartial.

I was upgrading my mvc5 project to mvc6 and came across this issue.

In the above file we have

Partial
PartialAsync
RenderPartialAsync

So I dont think its by design since if render methods were removed/depreciated I would also expect RenderPartialAsyc to be missing.

Even if we should be using a new methodology eg  @await Html.PartialAsync ?  its going to cause pain for upgraders.  What is the official line.  I am quite happy to find and replace but it seems unclear to me which way to go,

Thanks

Mike

javiercn:


src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/HtmlHelperPartialExtensions.cs
```diff
@@ -241,6 +241,94 @@ public static class HtmlHelperPartialExtensions
         /// <param name="partialViewName">
         /// The name of the partial view used to create the HTML markup. Must not be <c>null</c>.
         /// </param>
+        /// <remarks>
+        /// In this context, "renders" means the method writes its output using <see cref="ViewContext.Writer"/>.
```
pranavkm: Do we really need this remark? Might want to merge it into the summary perhaps.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/HtmlHelperPartialExtensions.cs
```diff
@@ -241,6 +241,94 @@ public static class HtmlHelperPartialExtensions
         /// <param name="partialViewName">
         /// The name of the partial view used to create the HTML markup. Must not be <c>null</c>.
```
pranavkm: Feels weird to call out the non-nullability of parameters. We don't do that elsewhere.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperPartialExtensionsTest.cs
```diff
@@ -141,6 +141,79 @@ public class HtmlHelperPartialExtensionsTest
...
+        [MemberData(nameof(RenderPartialExtensionMethods))]
+        public void RenderPartialMethods_DoesNotWrapThrownException(
+            Action<IHtmlHelper> partialMethod,
+            object unusedModel,
+            ViewDataDictionary unusedViewData)
+        {
+            // Arrange
+            var expected = new InvalidOperationException();
+            var helper = new Mock<IHtmlHelper>();
+            helper.Setup(h => h.RenderPartialAsync("test", It.IsAny<object>(), It.IsAny<ViewDataDictionary>()))
```
pranavkm: Use VS default indentation (i.e. just hit enter and let VS indent at 4 spaces)

test/WebSites/RazorWebSite/Controllers/PartialViewEngineController.cs
```diff
@@ -42,5 +42,15 @@ public IActionResult PartialWithModel()
             };
             return PartialView(model);
         }
+
+        public IActionResult RenderPartialWithModel()
```
pranavkm: Probably don't need functional tests for this if unit test covers this.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/HtmlHelperPartialExtensions.cs
javiercn: It's the same comment than in the Async version. I just removed the <returns> section of the comment. Everything else seemed to apply

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/HtmlHelperPartialExtensions.cs
javiercn: It's the same comment than in the Async version. I don't have a strong opinion about it, but I rather hear from more people before changing it.

test/WebSites/RazorWebSite/Controllers/PartialViewEngineController.cs
javiercn: I think it's good to have a functional test specifically because we block with GetAwaiter().GetResult(), and I believe that that has the potential to be host dependent.

pranavkm: All the functional tests run under an in-memory host. Running it should be identical to a unit test. We went through an exercise of trimming functional tests so it really doesn't make to add coverage for trivial scenarios.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/HtmlHelperPartialExtensions.cs
Eilon: The comments are a little weird but don't seem to do much harm. I wouldn't go out of the way to either add these types of comments nor remove them; if they're there, they can stay.

test/WebSites/RazorWebSite/Controllers/PartialViewEngineController.cs
dougbu: +1 with what @pranavkm said. `GetAwaiter().GetResult()` isn't likely to be host-dependent. As long as the unit tests run under both full CLR and .NET Core, you're covered.

dougbu:
Suggest rebasing to see if this PR hits any xplat issues. Right now the CI builds are failing because these commits are out-of-date.

dougbu:
:shipit: once CI builds pass, minus the functional test bit.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperPartialExtensionsTest.cs
javiercn: @pranavkm Not sure what you mean by this, the other test method in the class looks exactly the same as this one.

pranavkm: I think the `Callback` is indented to align with `Setup`. Instead, just hit tab. We've been fixing existing code where we went out of the way to keep the `.`s aligned.

pranavkm: See https://github.com/aspnet/Mvc/pull/3761/files#diff-c44a4239a9811abb5bf641f6210344c3R205

Issue #3495, opened 2015-11-08
Expose ViewResult.Model


davidfowl:
For code portability from MVC 5:

https://msdn.microsoft.com/en-us/library/system.web.mvc.viewresult(v=vs.118).aspx

Muchiachio:
:+1: really get's in a way when porting tests.

NTaylorMullen:
Is this still needed? You can do `viewResult.ViewData.Model`. By adding an accessor to `ViewResult` we'd just be adding a 1-off shortcut.

pranavkm:
@NTaylorMullen , isn't this a compat requirement?

NTaylorMullen:
Ya. I just don't see the value in bringing it forward when you can just do `.ViewData.Model`.

Muchiachio:
This doesn't look much, but testing model values returned from the controller actions is already long enough :smile:
```C#
ControllerModel actual = ((controller.Edit(controllerModel) as ViewResult).ViewData.Model as ControllerModel);
```

NTaylorMullen:
29ea696c5ae4f0fbea66ae576147dca9f922446e

PR #3764, opened 2015-12-14
Add `Model` assessor to `ViewResult` for MVC5 portability.

NTaylorMullen:
#3495

### DIFF
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewResult.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewResult.cs
@@ -29,6 +29,11 @@ namespace Microsoft.AspNet.Mvc
         public string ViewName { get; set; }
 
         /// <summary>
+        /// Gets the view data model.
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Mvc.ViewFeatures/ViewResult.cs
```diff
@@ -29,6 +29,11 @@ public class ViewResult : ActionResult
         public string ViewName { get; set; }
 
         /// <summary>
+        /// Gets the view data model.
+        /// </summary>
+        public object Model => ViewData.Model;
```
Eilon: Do we want `ViewData?.Model`?

NTaylorMullen: Yup!

pranavkm:
:shipit:

NTaylorMullen:
:up: :date:

Eilon:
:shipit:

Issue #3245, opened 2015-10-01
MVC Controller Response - Wrong ContentType


Tratcher:
_From @Fosol on October 1, 2015 19:13_

The default response ContentType of an endpoint that returns a struct value (i.e. int) is "application/json".  However the body contains invalid json as it only has the int value.

If you attempt to set the Response.ContentType = "text/plain" it ignores you and sets it to "application/json".

If you attempt to set the Response.ContentType = "text/plain" through the Response.OnStarting method it results in a body containing gibberish.

How do you set the Content Type of the response if you are returning a struct value type?

_Copied from original issue: aspnet/HttpAbstractions#424_

Tratcher:
_From @kichalla on October 1, 2015 19:19_

@Fosol what content type do you want to send it as? text/plain?...can you check if you return the int as a string you should see the response as text/plain...otherwise you can use the `ContentResult` to set the content and content-type the way you want

Tratcher:
_From @Fosol on October 1, 2015 19:56_

Here's an example endpoint.


        [HttpGet]
        [Route("add/{value1}/{value2}")]
        public int Add(int value1, int value2)
{
            return value1 + value2;
}

The default response ContentType: application/json.  But the body only has the number and therefore is invalid json.

If you return a string it does by default become text/plain.

Tratcher:
_From @kichalla on October 1, 2015 20:9_

do you want to send the data in json format? if yes, you would need to have that value as a proeprty on an object...example: if your return type is `dynamic`, you can return something like `return new { result = value1 + value 2}`...the response would be in json and `{ "result":30 }`

Tratcher:
_From @Fosol on October 1, 2015 20:41_

You mention the ContentResult object.  Does this only work if you are returning something like a ActionResult?

Fosol:
This is a bug.

Struct value types should not be returned as application/json by default.

danroth27:
A number is actually valid JSON based on the latest RFC (http://www.ietf.org/rfc/rfc7159.txt):

> JSON-text = ws value ws
> value = false / null / true / object / array / number / string

In MVC generally if you return an object that isn't an IActionResult then we default to using an ObjectResult to format the response based on content negotiation. So the behavior you are seeing is _mostly_ expected.

Except, you should be able to explicitly set the content type and not have it get stomped on by the selected formatter. The fact that your content type is overwritten is a bug.

Fosol:
Thanks!  That was informative.

[...]

PR #3767, opened 2015-12-15
MVC Controller Response - Wrong ContentType #3245

kichalla:
@rynowak

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/Infrastructure/ObjectResultExecutor.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Infrastructure/ObjectResultExecutor.cs
@@ -48,7 +48,7 @@ namespace Microsoft.AspNet.Mvc.Infrastructure
             Logger = loggerFactory.CreateLogger<ObjectResultExecutor>();
             WriterFactory = writerFactory.CreateWriter;
         }
-        
+
         /// <summary>
         /// Gets the <see cref="ILogger"/>.
         /// </summary>
@@ -89,8 +89,6 @@ namespace Microsoft.AspNet.Mvc.Infrastructure
                 throw new ArgumentNullException(nameof(result));
             }
 
-            ValidateContentTypes(result.ContentTypes);
-
             var formatters = result.Formatters;
             if (formatters == null || formatters.Count == 0)
             {
@@ -108,7 +106,7 @@ namespace Microsoft.AspNet.Mvc.Infrastructure
                 WriterFactory,
                 objectType,
                 result.Value);
-            
+
             var selectedFormatter = SelectFormatter(formatterContext, result.ContentTypes, formatters);
             if (selectedFormatter == null)
             {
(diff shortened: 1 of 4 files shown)


test/WebSites/FormatterWebSite/Controllers/StreamController.cs
```diff
@@ -37,7 +37,7 @@ public Stream MemoryStreamWithContentTypeFromProducesWithMultipleValues()
 
         [HttpGet]
         [Produces("text/plain")]
-        public Stream MemoryStreamOverridesContentTypeWithProduces()
+        public Stream MemoryStreamOverridesProducesContentTypeWithResponseContentType()
         {
             // Produces will set a ContentType on the implicit ObjectResult and
```
kichalla: will change this comment

src/Microsoft.AspNet.Mvc.Core/Infrastructure/ObjectResultExecutor.cs
```diff
@@ -159,6 +157,15 @@ public virtual Task ExecuteAsync(ActionContext context, ObjectResult result)
...
             }
 
+            var responseContentType = formatterContext.HttpContext.Response.ContentType;
+            if (!string.IsNullOrEmpty(responseContentType))
+            {
+                contentTypes.Clear();
+                contentTypes.Add(MediaTypeHeaderValue.Parse(responseContentType));
+            }
+
+            ValidateContentTypes(contentTypes);
```
rynowak: This whole block should really be inside of `ExecuteAsync` (where you took this code from). If I'm overriding this method, I want to **use** what was passed in, not ignore it.

src/Microsoft.AspNet.Mvc.Core/Infrastructure/ObjectResultExecutor.cs
```diff
@@ -159,6 +157,15 @@ public virtual Task ExecuteAsync(ActionContext context, ObjectResult result)
                 throw new ArgumentNullException(nameof(formatters));
             }
 
+            var responseContentType = formatterContext.HttpContext.Response.ContentType;
+            if (!string.IsNullOrEmpty(responseContentType))
+            {
+                contentTypes.Clear();
+                contentTypes.Add(MediaTypeHeaderValue.Parse(responseContentType));
+            }
```
rynowak: Why does `Response.ContentType` take precedence over something set on the object result? Let's get agreement on the right priority, and add some comments here to that effect. Let's discuss with @lodejard when he's around.

rynowak:
:watch: - let's discuss briefly with Lou. The code changes are good in general, I'm just not sure the priority is right.

kichalla:
:up: :date:

rynowak:
:shipit:

kichalla:
d77655fb7334e4fdde99797b9e2579649aa72aeb

Issue #3746, opened 2015-12-11
Allow IValidatableObject implementors and ValidationAttribute derivatives to access the IServiceProvider


tuespetre:
Because of the removal of the ability to use the FromServices attribute on model properties, and considering the current lack of support for constructors or dependency injection of any sort on models, there is no longer a way for models implementing the IValidatableObject interface to make use of services for validation purposes.

I would like to propose a very simple idea to fill in that gap: have the model binding system pass the IServiceProvider into the constructor of the ValidationContext during validation.

- [See where it would be done for IValidatableObject here.](https://github.com/aspnet/Mvc/blob/f7a211c09531f145329d76fe71d6676d5c4f3685/src/Microsoft.AspNet.Mvc.DataAnnotations/ValidatableObjectAdapter.cs#L32)
- [See where it would be done for ValidationAttribute here.](https://github.com/aspnet/Mvc/blob/3393ba43c2028af7048d45178c56802adfe5a80f/src/Microsoft.AspNet.Mvc.DataAnnotations/DataAnnotationsModelValidator.cs#L85)

Since ASP.NET 5 and the ValidationContext use the exact same IServiceProvider interface, developers would even be able to make use of the extension methods then (GetRequiredService, etc.)

If this is agreeable I would be happy to do the required work. Of course if anyone proposes an alternative solution to allow models or attributes to use services for validation purposes, I would be glad to hear it.

tuespetre:
@rynowak resolved this with https://github.com/aspnet/Mvc/commit/328f4d648b8d921d2e9e9e3849d8afa88d56694b in response to #3770

PR #3770, opened 2015-12-15
Allow ValidationAttributes and IValidatableObjects to resolve service…

tuespetre:
Allow ValidationAttributes and IValidatableObjects to resolve services from the RequestServices provider by injecting it into the ValidationContext. 

See: #3746


Eilon:
@rynowak assigning to you because I think you had some thoughts on this.

rynowak:
Yeah, this code change seems good as-is :shipit: 

The concern that I have here is re: perf of the `ValidationContext` when you pass in the service provider. I'm going to try out this change and verify that it doesn't blow up our allocations.

rynowak:
328f4d648b8d921d2e9e9e3849d8afa88d56694b

rynowak:
Thanks!

PR #3772, opened 2015-12-15
Change `HtmlString` to inherit from `HtmlEncodedString`

dougbu:
- part of aspnet/HtmlAbstractions#5 fix

### DIFF
--- a/test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperTest.cs
+++ b/test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperTest.cs
@@ -125,7 +125,7 @@ namespace Microsoft.AspNet.Mvc.Rendering
             {
                 return new TheoryData<string, string>
                 {
-                    { null, null },
+                    { null, string.Empty },
                     { string.Empty, string.Empty },
                     { "<\">", "<\">" },
                     { "<br />", "<br />" },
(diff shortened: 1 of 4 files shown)


test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperTest.cs
```diff
@@ -125,7 +125,7 @@ public class HtmlHelperTest
             {
                 return new TheoryData<string, string>
                 {
-                    { null, null },
+                    { null, string.Empty },
```
dougbu: Was just strange that `HtmlString.ToString()` could previously return `null`.

dougbu:
FYI slight improvement in memory performance with simple view that uses a view component that uses `AppendFormat()`. Most of the change comes from reducing the `StringWriter` count by one per request.
Before:
before
After:
after

dougbu:
Note CI builds cannot succeed until aspnet/HtmlAbstractions#6 is in. Have tested combination locally of course.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperTest.cs
rynowak: agreed, very strange

dougbu:
:up::date: per offline discussion.

@rynowak did you intend to sign off?

rynowak:
:shipit: 

Ya sorry, I had intended to sign off yesterday, I got distracted by something distracting.

dougbu:
3d82255

PR #3773, opened 2015-12-16
Modify IViewComponentHelper to remove method selection ambiguity.

### LINKED ISSUE #612: Loosen ViewComponent invoke restrictions
@rynowak and I talked a bit about this in person and I found it very unappealing how in today's world we don't provide a default signature for Invoke/InvokeAsync which makes building ViewComponents inconvenient.

Potential fixes to this could be allowing 
```C#
Component.InvokeAsync("typeName", "methodName", myparams)
```
Or some derivative. Even possibly:
```C#
Component.Invoke<MyParamObject>("typeName", new MyParamObject(...))
```
Which would allow us to have a virtual Invoke/InvokeAsync signature.

Of course these are just ideas thrown around and neither may be the right solution to the problem.  We should discuss making this API simpler and easier to get started with to prevent users from writing code that never gets invoked.

pranavkm:
Fixes #612


test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewComponents/ViewComponentMethodSelectorTest.cs
```diff
@@ -1,282 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
```
pranavkm: Ignore this for now. I'll revive this.

test/WebSites/RazorWebSite/Views/ViewEngine/ViewWithRelativePath.cshtml
```diff
@@ -12,4 +12,4 @@
 }
 ViewWithRelativePath-content
 <partial>@await Html.PartialAsync("../Shared/_PartialThatSetsTitle.cshtml")</partial>
-@await Component.InvokeAsync("ComponentWithRelativePath", person)
\ No newline at end of file
+@await Component.InvokeAsync("ComponentWithRelativePath", new { person })
```
pranavkm: This.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentInvoker.cs
```diff
@@ -179,11 +157,13 @@ public IViewComponentResult InvokeSyncCore(MethodInfo method, ViewComponentConte
             {
                 _diagnosticSource.BeforeViewComponent(context, component);
                 _logger.ViewComponentExecuting(context);
-
+                var indexedArguments = ControllerActionExecutor.PrepareArguments(
```
pranavkm: Should we throw here if any values are null?

src/Microsoft.AspNet.Mvc.ViewFeatures/IViewComponentHelper.cs
```diff
@@ -9,20 +9,12 @@ namespace Microsoft.AspNet.Mvc
...
-        void RenderInvoke(Type componentType, params object[] args);
-
-        Task<IHtmlContent> InvokeAsync(string name, params object[] args);
-
-        Task<IHtmlContent> InvokeAsync(Type componentType, params object[] args);
-
-        Task RenderInvokeAsync(string name, params object[] args);
-
-        Task RenderInvokeAsync(Type componentType, params object[] args);
+        Task RenderInvokeAsync(Type componentType, object arguments);
```
rynowak: https://github.com/aspnet/Mvc/issues/612#issuecomment-159103197

No more `RenderInvokeAsync

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponentHelperExtensions.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
```
rynowak: license and docs yea.

YEAHIKNOWITSADESIGNPRANDIDONTCARE

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentInvoker.cs
rynowak: We should make sure to support default values as well. 

I don't know what you mean? why would null be invalid?

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/ViewComponentContext.cs
```diff
@@ -87,7 +81,7 @@ public ViewComponentContext()
         /// <remarks>
         /// The property setter is provided for unit test purposes only.
         /// </remarks>
-        public object[] Arguments { get; set; }
+        public object Arguments { get; set; }
```
rynowak: This should be a dictionary

rynowak: See ActionFilterContext

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/ViewComponentMethodSelector.cs
```diff
@@ -63,36 +62,5 @@ public static MethodInfo FindSyncMethod(TypeInfo componentType, object[] args)
 
             return method;
         }
-
-        private static MethodInfo GetMethod(TypeInfo componentType, object[] args, string methodName)
```
rynowak: :clap:

test/WebSites/RazorWebSite/Views/ViewEngine/ViewWithRelativePath.cshtml
rynowak: :heart_eyes_cat:

rynowak:
This will sooo need an announcement

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentInvoker.cs
pranavkm: In the event an argument does not have a value?

Component.InvokeAsync("Foo", new { a = "some value" }) -->  InvokeAsync(object a, object b)

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/ViewComponentContext.cs
pranavkm: Ok

src/Microsoft.AspNet.Mvc.ViewFeatures/IViewComponentHelper.cs
pranavkm: Ahh. It sounded like we were getting rid of the synchronous variants.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentInvoker.cs
rynowak: >In the event an argument does not have a value?

Yes. The default value needs to get plugged in at that point

src/Microsoft.AspNet.Mvc.ViewFeatures/IViewComponentHelper.cs
rynowak: Yeah, both of those. We're simplifying it to the minimum needed.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewComponents/ViewComponentMethodSelectorTest.cs
pranavkm: The code formerly referred to as `ViewComponentMethodSelector` is part of the descriptor provider now.

pranavkm:
:up: :date:

test/WebSites/RazorWebSite/Views/ViewEngine/ViewWithRelativePath.cshtml
pranavkm: Not for this PR, but we could try to be clever in this case and attempt to make it work if the user does

`@await Component.InvokeAsync("ComponentWithRelativePath", person)` (similarly with a params array). The user experience for single parameters is a lot worse than what it used to be before.

src/Microsoft.AspNet.Mvc.ViewFeatures/IViewComponentHelper.cs
```diff
@@ -7,22 +7,27 @@
...
-
-        Task<IHtmlContent> InvokeAsync(string name, params object[] args);
-
-        Task<IHtmlContent> InvokeAsync(Type componentType, params object[] args);
-
-        Task RenderInvokeAsync(string name, params object[] args);
-
-        Task RenderInvokeAsync(Type componentType, params object[] args);
+        /// <summary>
+        /// Invokes a View Component with the specified <paramref name="name"/>.
```
pranavkm: Is the casing for this `View Component` or `view component`? I saw some docs earlier which used the former and stuck with it.

src/Microsoft.AspNet.Mvc.ViewFeatures/Logging/ViewComponentResultLoggerExtensions.cs
```diff
@@ -18,32 +19,35 @@ static ViewComponentResultLoggerExtensions()
...
-                {
-                    formattedArguments[i] = Convert.ToString(arguments[i]);
-                }
-
+                var formattedArguments = GetFormattedArguments(arguments);
                 _viewComponentResultExecuting(logger, viewComponentType.Name, formattedArguments, null);
             }
         }
+
+        private static string[] GetFormattedArguments(object arguments)
```
pranavkm: @rynowak should we get rid of this (i.e. remove the arguments from logging?)

src/Microsoft.AspNet.Mvc.ViewFeatures/IViewComponentHelper.cs
```diff
@@ -7,22 +7,27 @@
...
+        /// </summary>
+        /// <param name="name">The name of the View Component.</param>
+        /// <param name="arguments">Arguments to be passed to tbe invoked View Component method.</param>
```
NTaylorMullen: `the`

src/Microsoft.AspNet.Mvc.ViewFeatures/IViewComponentHelper.cs
```diff
@@ -7,22 +7,27 @@
...
+        /// <returns>A <see cref="Task"/> that on completion returns the rendered <see cref="IHtmlContent" />.
+        /// </returns>
+        Task<IHtmlContent> InvokeAsync(string name, object arguments);
+
+        /// <summary>
+        /// Invokes a View Component of type <paramref name="componentType" />.
+        /// </summary>
+        /// <param name="componentType">The View Component <see cref="Type"/>.</param>
+        /// <param name="arguments">Arguments to be passed to tbe invoked View Component method.</param>
```
NTaylorMullen: `the`

src/Microsoft.AspNet.Mvc.ViewFeatures/Logging/DefaultViewComponentInvokerLoggerExtensions.cs
```diff
@@ -40,10 +40,10 @@ public static void ViewComponentExecuting(this ILogger logger, ViewComponentCont
         public static void ViewComponentExecuted(
             this ILogger logger,
             ViewComponentContext context,
-            int startTime,
+            DateTime startTime,
             object result)
         {
-            var elapsed = new TimeSpan(Environment.TickCount - startTime);
+            var elapsed = DateTime.UtcNow - startTime;
```
NTaylorMullen: Do we care about accuracy here? Doing date time comparisons to compute elapsed times isn't very accurate.

src/Microsoft.AspNet.Mvc.ViewFeatures/Logging/ViewComponentResultLoggerExtensions.cs
```diff
@@ -18,32 +19,35 @@ static ViewComponentResultLoggerExtensions()
...
+        {
+            var argumentDictionary = PropertyHelper.ObjectToDictionary(arguments);
+            var formattedArguments = new string[argumentDictionary.Count];
+            var i = 0;
+            foreach (var item in argumentDictionary)
+            {
+                formattedArguments[i++] = Convert.ToString(item.Value);
```
NTaylorMullen: I know we didn't do this before but we don't care about the item keys at all?

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/ViewComponentHelperExtensions.cs
```diff
@@ -7,48 +7,76 @@
...
         }
 
-        public static Task<IHtmlContent> InvokeAsync<TComponent>(
-            this IViewComponentHelper helper,
-            params object[] args)
+        /// <summary>
+        /// Invokes a View Component of type <typeparam name="TComponent"/>.
+        /// </summary>
+        /// <param name="helper">The <see cref="IViewComponentHelper"/>.</param>
+        /// <param name="arguments">Arguments to be passed to tbe invoked View Component method.</param>
```
NTaylorMullen: `the`

src/Microsoft.AspNet.Mvc.ViewFeatures/Logging/DefaultViewComponentInvokerLoggerExtensions.cs
pranavkm: It's better than using `Environment.TickCounts` according to http://stackoverflow.com/a/8865560. I think the reason it doesn't use a `Stopwatch` is to avoid the allocation for logging.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/ViewComponentContext.cs
```diff
@@ -48,11 +48,6 @@ public ViewComponentContext()
                 throw new ArgumentNullException(nameof(viewComponentDescriptor));
             }
 
-            if (arguments == null)
```
pranavkm: Undo

src/Microsoft.AspNet.Mvc.ViewFeatures/Resources.resx
```diff
@@ -137,10 +137,7 @@
     <value>An invoker could not be created for the view component '{0}'.</value>
   </data>
   <data name="ViewComponent_CannotFindMethod" xml:space="preserve">
-    <value>Could not find an '{0}' method matching the parameters.</value>
-  </data>
-  <data name="ViewComponent_CannotFindMethod_WithFallback" xml:space="preserve">
-    <value>Could not find an '{0}' or '{1}' method matching the parameters.</value>
+    <value>Could not find an '{0}' or '{1}' method for the view component '{2}'.</value>
```
NTaylorMullen: Should stay consistent with elsewhere when saying `view component` vs. `View Component`

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentDescriptorProvider.cs
```diff
@@ -14,6 +16,8 @@ namespace Microsoft.AspNet.Mvc.ViewComponents
     /// </summary>
     public class DefaultViewComponentDescriptorProvider : IViewComponentDescriptorProvider
     {
+        private const string AsyncMethodName = "InvokeAsync";
+        private const string SyncMethodName = "Invoke";
```
NTaylorMullen: So it's always `@Component.InvokeAsync` but you can write your `ViewComponent` with `Invoke` OR `InvokeAsync`? Interesting.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentDescriptorProvider.cs
```diff
@@ -63,16 +67,69 @@ protected virtual bool IsViewComponentType(TypeInfo typeInfo)
...
             };
 
             return candidate;
         }
+
+        private static MethodInfo FindMethod(Type componentType)
+        {
+            var componentName = componentType.FullName;
+            var methods = componentType.GetMethods(BindingFlags.Public | BindingFlags.Instance)
+                .Where(m => string.Equals(m.Name, AsyncMethodName, StringComparison.Ordinal) ||
```
NTaylorMullen: `m` => `method` who cares about long line here imo.

src/Microsoft.AspNet.Mvc.ViewFeatures/Resources.resx
```diff
@@ -122,13 +122,13 @@
...
   <data name="ViewComponent_AsyncMethod_ShouldReturnTask" xml:space="preserve">
-    <value>The async view component method '{0}' should be declared to return Task&lt;T&gt;.</value>
+    <value>Method '{0}' of view component '{1}' should be declared to return {2}&lt;T&gt;.</value>
   </data>
   <data name="ViewComponent_MustReturnValue" xml:space="preserve">
     <value>A view component must return a non-null value.</value>
   </data>
   <data name="ViewComponent_SyncMethod_ShouldReturnValue" xml:space="preserve">
-    <value>The view component method '{0}' should be declared to return a value.</value>
+    <value>Method '{0}' of view component '{1}' should be declared to return a value.</value>
```
NTaylorMullen: Could add a comment saying an alternative (if you don't want to return a value) is to inject services.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentInvoker.cs
```diff
@@ -55,60 +65,33 @@ public class DefaultViewComponentInvoker : IViewComponentInvoker
...
             if (context == null)
             {
                 throw new ArgumentNullException(nameof(context));
             }
 
-            var method = ViewComponentMethodSelector.FindSyncMethod(
-                context.ViewComponentDescriptor.Type.GetTypeInfo(),
-                context.Arguments);
-            if (method == null)
+            var methodInfo = context.ViewComponentDescriptor?.MethodInfo;
```
NTaylorMullen: `ViewComponentDescriptor` can be `null`?

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/ViewComponentContext.cs
NTaylorMullen: Arguments still can't be `null` can it? The `PropertyHelper.ObjectToDictionary` should always at least create a new dictionary for you.

test/WebSites/RazorWebSite/Views/ViewEngine/ViewWithRelativePath.cshtml
NTaylorMullen: One thing, tooling sucks with this doesn't it? Aka second you type `{` does it properly auto-complete the braces? If not we should file a tooling bug. I think I recall it being difficult to type out in the Razor editor.

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/ViewComponentContext.cs
pranavkm: Yeah, it can't be null anymore.

src/Microsoft.AspNet.Mvc.ViewFeatures/IViewComponentHelper.cs
rynowak: I think I've generally done lower case.

src/Microsoft.AspNet.Mvc.ViewFeatures/Logging/ViewComponentResultLoggerExtensions.cs
rynowak: No

src/Microsoft.AspNet.Mvc.ViewFeatures/Logging/DefaultViewComponentInvokerLoggerExtensions.cs
rynowak: We don't care about accuracy beyond the ms level. What we're doing here is intentional

src/Microsoft.AspNet.Mvc.ViewFeatures/Logging/ViewComponentResultLoggerExtensions.cs
rynowak: As long as they are in the correct order, we'd want the keys... It looks like they are not in the correct order.

src/Microsoft.AspNet.Mvc.ViewFeatures/Resources.resx
rynowak: I don't need this comment. `void` and `Task` are specifically not allowed for VCs

src/Microsoft.AspNet.Mvc.ViewFeatures/IViewComponentHelper.cs
rynowak: Yeah let's just standardize on lower case

src/Microsoft.AspNet.Mvc.ViewFeatures/Microsoft.AspNet.Mvc.ViewFeatures.xproj
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <VisualStudioVersion Condition="'$(VisualStudioVersion)' == ''">14.0</VisualStudioVersion>
     <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
+    <DisabledCustomTools>.resx</DisabledCustomTools>
```
rynowak: :heart_eyes_cat:

src/Microsoft.AspNet.Mvc.ViewFeatures/Logging/DefaultViewComponentInvokerLoggerExtensions.cs
rynowak: Wait, I just noticed the change. If you're going to make this change, do it more globablly and not part of this PR. We have 3-4 different places we do the same thing in code.

rynowak:
:watch:

src/Microsoft.AspNet.Mvc.ViewFeatures/Logging/DefaultViewComponentInvokerLoggerExtensions.cs
pranavkm: Okie dokie.

src/Microsoft.AspNet.Mvc.ViewFeatures/Logging/ViewComponentResultLoggerExtensions.cs
pranavkm: We don't have the descriptor at this point to figure out what the order is. Are there any other places that log a dictionary?

src/Microsoft.AspNet.Mvc.ViewFeatures/Resources.resx
pranavkm: :+1: to @rynowak. You can't have a ViewComponent Invoke method that returns nothing. (which is pretty much what the exception message says)

src/Microsoft.AspNet.Mvc.ViewFeatures/Resources.resx
pranavkm: `view component` it is.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentInvoker.cs
pranavkm: It's settable and we only verify if the context is non null at the start of the method.

test/WebSites/RazorWebSite/Views/ViewEngine/ViewWithRelativePath.cshtml
pranavkm: Tooling works fine.

pranavkm:
Updated

NTaylorMullen:
:shipit:

src/Microsoft.AspNet.Mvc.ViewFeatures/Logging/ViewComponentResultLoggerExtensions.cs
rynowak: Look at what we do for actions. It might be ok to log keys + values.

pranavkm:
@rynowak moved the logging to the invoker where we have ordered arguments.

Issue #3215, opened 2015-09-28
Extend from PR #3209 to cover other framework-provided messages


dougbu:
PR does not cover all errors for which the model binding system provides messages added to the `ModelStateDictionary`.

For example, comments from design phase of that PR:
``` c#
        // See use of Resources.Common_ValueNotValidForProperty in ValidationHelpers.
        // <summary>
        // Fallback error message HTML and tag helpers display when a property is invalid but the
        // <see cref="ModelError"/>s have <c>null</c> <see cref="ModelError.ErrorMessage"/>s.
        // </summary>
        // <value>Default <see cref="string"/> is "The value '{0}' is invalid.".</value>
        ////Func<string> ValueInvalid_UnknownErrorResource { get; }

        // See how Resources.ModelError_InvalidValue_MessageWithModelValue is used
        // <summary>
        // Replacement <see cref="ModelError.ErrorMessage"/> used in <see cref="ModelError"/> when
        // <see cref="ModelError.Exception"/> is of type <see cref="FormatException"/> and value is known i.e. when
        // the bound value could not be converted and the "replacer" has the attempted value.
        // </summary>
        // <value>Default <see cref="string"/> is "The value '{0}' is not valid for {1}.".</value>
        ////Func<string> ValueInvalid_WithValueResource { get; }

        // See how Resources.ModelError_InvalidValue_GenericMessage is used.
        // <summary>
        // Replacement <see cref="ModelError.ErrorMessage"/> used in <see cref="ModelError"/> when
        // <see cref="ModelError.Exception"/> is of type <see cref="FormatException"/> and value is unknown i.e. when
        // the bound value could not be converted and the "replacer" does not have the attempted value.
        // </summary>
        // <value>Default <see cref="string"/> is "The supplied value is invalid for {0}.".</value>
        ////Func<string> ValueInvalid_WithoutValueResource { get; }
```

Also see

- [PR comment](https://github.com/aspnet/Mvc/pull/3209#discussion-diff-40490318):
> One way to avoid this framework-provided string entirely would be to use the next two `Func`s to ensure `ModelError`s never have a `null` or empty `ErrorMessage`. Then `DefaultHtmlGenerator` and `ValidationHelpers` would never need to fill a gap the model binding system left.

- [PR comment](https://github.com/aspnet/Mvc/pull/3209#discussion-diff-40490391):
[...]

PR #3775, opened 2015-12-16
Extend from PR #3209 to cover other framework-provided messages #3215

kichalla:
@rynowak


src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Metadata/IModelBindingMessageProvider.cs
```diff
@@ -30,5 +30,28 @@ public interface IModelBindingMessageProvider
         /// </summary>
         /// <value>Default <see cref="string"/> is "The value '{0}' is invalid.".</value>
         Func<string, string> ValueMustNotBeNullAccessor { get; }
+
+        /// <summary>
+        /// Fallback error message HTML and tag helpers display when a property is invalid but the
+        /// <see cref="ModelError"/>s have <c>null</c> <see cref="ModelError.ErrorMessage"/>s.
+        /// </summary>
+        /// <value>Default <see cref="string"/> is "The value '{0}' is invalid.".</value>
+        Func<string, string> ValueInvalid_UnknownErrorResource { get; }
```
kichalla: Name suggestions?

dougbu:
@kichalla the "framework-provided messages" listed in #3215 are _examples_. Not sure if there's more but two other cases are:
- https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs#L144
- https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/ModelBinding/BodyModelBinder.cs#L89

kichalla:
:up: :date:

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Metadata/IModelBindingMessageProvider.cs
rynowak: Follow the naming conventions of the other ones. Drop the underscore, end with 'Accessor' instead of 'Resource'

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Metadata/IModelBindingMessageProvider.cs
```diff
@@ -30,5 +30,46 @@ public interface IModelBindingMessageProvider
...
+
+        /// <summary>
+        /// Replacement <see cref="ModelError.ErrorMessage"/> used in <see cref="ModelError"/> when
```
rynowak: `Error message the model binding system adds when `.... rather than what you have

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Metadata/IModelBindingMessageProvider.cs
```diff
@@ -30,5 +30,46 @@ public interface IModelBindingMessageProvider
...
+        /// <see cref="ModelError.Exception"/> is of type <see cref="FormatException"/> and value is known i.e. when
+        /// the bound value could not be converted and the "replacer" has the attempted value.
```
rynowak: what's a replacer?

src/Microsoft.AspNet.Mvc.Core/Properties/AssemblyInfo.cs
```diff
@@ -5,6 +5,7 @@
 using System.Resources;
 using System.Runtime.CompilerServices;
 
+[assembly: InternalsVisibleTo("Microsoft.AspNet.Mvc.Abstractions.Test, PublicKey=0024000004800000940000000602000000240000525341310004000001000100f33a29044fa9d740c9b3213a93e57c84b472c84e0b8a0e1ae48e67a9f8f6de9d5f7f3d52ac23e48ac51801f1dc950abe901da34d2a9e3baadb141a17c77ef3c565dd5ee5054b91cf63bb3c6ab83f72ab3aafe93d0fc3c2348b764fafb0b1c0733de51459aeab46580384bf9d74c4e28164b7cde247f891ba07891c9d872ad2bb")]
 [assembly: InternalsVisibleTo("Microsoft.AspNet.Mvc.Core.Test, PublicKey=0024000004800000940000000602000000240000525341310004000001000100f33a29044fa9d740c9b3213a93e57c84b472c84e0b8a0e1ae48e67a9f8f6de9d5f7f3d52ac23e48ac51801f1dc950abe901da34d2a9e3baadb141a17c77ef3c565dd5ee5054b91cf63bb3c6ab83f72ab3aafe93d0fc3c2348b764fafb0b1c0733de51459aeab46580384bf9d74c4e28164b7cde247f891ba07891c9d872ad2bb")]
```
rynowak: What's this needed for?

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Metadata/IModelBindingMessageProvider.cs
```diff
@@ -30,5 +30,46 @@ public interface IModelBindingMessageProvider
...
+
+        /// <summary>
+        /// Fallback error message used when deserialization of request body fails due to no supported encodings
+        /// found on the input formatter.
+        /// </summary>
+        /// <value>
+        /// Default <see cref="string"/> is "No encoding found for input formatter '{0}'. There must be at
+        /// least one supported encoding registered in order for the formatter to read content.".
+        /// </value>
+        Func<string, string> NoEncodingFoundOnInputFormatter { get; }
```
rynowak: I'm not sure we should be putting messages on here that don't map to a `ModelMetadata`. This needs more discussion

kichalla: I have reverted this change now (also unlike other messages which are user facing or which are displayed to the user, the messages by `NoEncodingFoundOnInputFormatter` and `unsupportedcontenttype` are consumed by api clients.)

kichalla:
:up: :date:

src/Microsoft.AspNet.Mvc.ViewFeatures/Internal/ValidationHelpers.cs
```diff
@@ -23,7 +27,7 @@ public static string GetUserErrorMessageOrDefault(ModelError modelError, ModelSt
             }
 
             var attemptedValue = entry.AttemptedValue ?? "null";
-            return Resources.FormatCommon_ValueNotValidForProperty(attemptedValue);
+            return messageProvider.InvalidValueWithUnknownModelErrorAccessor(attemptedValue);
```
rynowak: This is more of what I was talking about. This isn't tied to a specific model metadata, and so it doesn't fit into the current design.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Metadata/IModelBindingMessageProvider.cs
dougbu: This decision goes against the intent of the bug. We want users to be able to override all framework-provided messages that can be displayed on a page. Why arbitrarily leave out some errors, especially those that occur during model binding? (Side note: I could understand a need to generalize the `IModelBindingMessageProvider` name or to move the model metadata provider property from to `DisplayMetadata`.)

> don't map to a `ModelMetadata`

We have a `ModelMetadata` instance for the property the system is attempting to bind.

> the messages by `NoEncodingFoundOnInputFormatter` and `unsupportedcontenttype` are consumed by api clients.

This ignores AJAX scenarios.

src/Microsoft.AspNet.Mvc.ViewFeatures/Internal/ValidationHelpers.cs
dougbu: Taking this one out represents a regression from MVC 5, where users can override this exact message.

If we're painting ourselves into a corner using `ModelMetadata` to carry these messages around, let's go back to one setting per message system-wide. (The per-property overrides supported in MVC 6 have no equivalent in earlier versions.) That is, just hang everything off `MvcOptions` and pass the `IModelBindingMessageProvider` into `[Try]AddModelError()` in addition to the `ModelMetadata` (with similar changes elsewhere).

src/Microsoft.AspNet.Mvc.Core/Resources.resx
```diff
@@ -325,4 +325,13 @@
...
   </data>
+  <data name="ModelError_InvalidValue_GenericMessage" xml:space="preserve">
+    <value>The supplied value is invalid for {0}.</value>
+  </data>
+  <data name="ModelError_InvalidValue_MessageWithModelValue" xml:space="preserve">
+    <value>The value '{0}' is not valid for {1}.</value>
+  </data>
+  <data name="Common_ValueNotValidForProperty" xml:space="preserve">
+    <value>The value '{0}' is invalid.</value>
+  </data>
```
dougbu: Well at least one aspect of fixing the whole bug remains :trollface:

kichalla: @dougbu I did not get it. Can you elaborate?

dougbu: See my other recent comments on deleted code. The commits you reverted are necessary to fix this issue. @rynowak's comment would be handled by removing the use of `ModelMetadata`.

In particular this resource isn't used because you unfortunately removed the additions requiring it.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Metadata/IModelBindingMessageProvider.cs
```diff
@@ -30,5 +30,19 @@ public interface IModelBindingMessageProvider
         /// </summary>
         /// <value>Default <see cref="string"/> is "The value '{0}' is invalid.".</value>
         Func<string, string> ValueMustNotBeNullAccessor { get; }
+
+        /// <summary>
+        /// Error message the model binding system adds when <see cref="ModelError.Exception"/> is of type
+        /// <see cref="FormatException"/> and value is known.
+        /// </summary>
+        /// <value>Default <see cref="string"/> is "The value '{0}' is not valid for {1}.".</value>
+        Func<string, string, string> InvalidValueWithKnownSuppliedValueAccessor { get; }
```
rynowak: What about `ValueIsInvalidAccessor` for this one and `SuppliedValueIsInvalidAccessor` for the other?

dougbu:
@kichalla open conversations [here](https://github.com/aspnet/Mvc/pull/3775#issuecomment-165257359), [here](https://github.com/aspnet/Mvc/pull/3775#discussion-diff-47978338) and [here](https://github.com/aspnet/Mvc/pull/3775#discussion-diff-48051533) are background to [this](https://github.com/aspnet/Mvc/pull/3775#discussion-diff-48065744):
> See my other recent comments on deleted code. The commits you reverted are necessary to fix this issue. @rynowak's comment would be handled by removing the use of `ModelMetadata`.

> In particular this resource isn't used because you unfortunately removed the additions requiring it.

The general points are:
- Driver here is completing the feature -- giving users the ability to override framework-provided messages that we add to the `ModelStateDictionary`. Issue #3215 covered user-controlled framework-provided error messages in MVC 5 that PR #3209 and Issue #2969 hadn't already covered. In https://github.com/aspnet/Mvc/pull/3775#issuecomment-165257359 I mentioned two other error messages that should be similarly handled.
- This feature is not necessarily tied to `ModelMetadata` at all. `ModelMetadata` connection was added when I originally did PR #3209 because we _thought_ per-property or per-type error messages would prove useful. But we don't have any concrete uses of this feature so far. Of course the `ModelMetadata` link also helped because instances were already passed into a few important methods.

For more MVC 5 background, have a look at how `GetValueInvalidResource()` and `GetValueRequiredResource()` are used in https://aspnetwebstack.codeplex.com/SourceControl/latest#src/System.Web.Mvc/DefaultModelBinder.cs or how `GetInvalidPropertyValueResource()` is used in https://aspnetwebstack.codeplex.com/SourceControl/latest#src/System.Web.Mvc/Html/ValidationExtensions.cs  More generally search MVC 5 for uses of `HttpContextBase.GetGlobalResourceObject()`.

dougbu:
Had an offline conversation about #3215. I'll start from @kichalla's `kiran/movemessages-to-messageprovider` branch but in a new PR. Plan is to leave the `Resources.InputFormatterNoEncoding` and `Resources.UnsupportedContentType` messages alone but do all three listed in #3215.

PR #3781, opened 2015-12-17
Fix IModelBinder docs

rynowak:


src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/IModelBinder.cs
```diff
@@ -6,22 +6,26 @@
 namespace Microsoft.AspNet.Mvc.ModelBinding
 {
     /// <summary>
-    /// Interface for model binding.
+    /// Defines an interface for model binders.
```
DamianEdwards: Standard for .NET API docs of interfaces usually takes the form "Represents a ..", so in this case e.g. "Represents an object that performs model binding operations.", or some such.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/IModelBinder.cs
```diff
@@ -6,22 +6,26 @@
...
         /// <summary>
-        /// Async function to bind to a particular model.
+        /// Attempts to bind a model.
         /// </summary>
-        /// <param name="bindingContext">The binding context which has the object to be bound.</param>
-        /// <returns>A Task which on completion returns a <see cref="ModelBindingResult"/> which represents the result
-        /// of the model binding process.
+        /// <param name="bindingContext">The <see cref="ModelBindingContext"/>.</param>
+        /// <returns>
+        /// <para>
```
pranavkm: Does VS intellisense look ok if the first entry is a `para`?

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/IModelBinder.cs
rynowak: Represents is literally the worst

DamianEdwards: Being inconsistent in our language is worse IMO.

rynowak: So lets be consistent and not use it. Why keep doing something that's awful

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/IModelBinder.cs
rynowak: sure

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/IModelBinder.cs
DamianEdwards: Wow, thanks for volunteering to do the sweep of our code base to ensure we use "Defines" everywhere! :trollface:

Seriously though, you're right. "Defines" is nicer, and we should use it consistently, but we currently don't.

PR #3784, opened 2015-12-17
Smarter antiforgery

rynowak:
Adds the concept of an IAntiforgeryPolicy marker interface as well as
the ability to overide policy with a 'closer' filter.

Adds a new [IgnoreAntiforgeryToken] attribute for overriding a scoped
antiforgery policy.

Adds a new [AutoValidateAntiforgeryToken] attribute (good name tbd) for
applying an application-wide antiforgery token. The idea is that you can
configure this as a global filter if your site is acting as a pure
browser-based or 1st party SPA. This new attribute only validates the
token for unsafe HTTP methods, so you can apply it broadly.


rynowak:
@Eilon - feedback on name?

src/Microsoft.AspNet.Mvc.ViewFeatures/AutoValidateAntiforgeryTokenAttribute.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = false, Inherited = true)]
+    public class AutoValidateAntiforgeryTokenAttribute : Attribute, IFilterFactory, IOrderedFilter
+    {
+        public int Order { get; set; }
+
+        public IFilterMetadata CreateInstance(IServiceProvider serviceProvider)
+        {
+            var antiforgery = serviceProvider.GetRequiredService<IAntiforgery>();
+            return new ValidateAntiforgeryTokenAuthorizationFilter(antiforgery);
```
Eilon: Should this be using the Auto one?

src/Microsoft.AspNet.Mvc.ViewFeatures/AutoValidateAntiforgeryTokenAuthorizationFilter.cs
```diff
@@ -0,0 +1,74 @@
...
+            var method = context.HttpContext.Request.Method;
+            if (string.Equals("GET", method, StringComparison.OrdinalIgnoreCase) ||
+                string.Equals("HEAD", method, StringComparison.OrdinalIgnoreCase) ||
+                string.Equals("TRACE", method, StringComparison.OrdinalIgnoreCase) ||
+                string.Equals("OPTIONS", method, StringComparison.OrdinalIgnoreCase))
+            {
+                return false;
+            }
+
+            // Assume anything we know isn't safe is unsafe.
```
Eilon: If we *know* it then it isn't an assumption...

src/Microsoft.AspNet.Mvc.ViewFeatures/AutoValidateAntiforgeryTokenAttribute.cs
```diff
@@ -0,0 +1,22 @@
...
+
+using System;
+using Microsoft.AspNet.Antiforgery;
+using Microsoft.AspNet.Mvc.Filters;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.AspNet.Mvc.ViewFeatures
+{
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = false, Inherited = true)]
+    public class AutoValidateAntiforgeryTokenAttribute : Attribute, IFilterFactory, IOrderedFilter
```
Eilon: Yeah the name doesn't say what's "auto" about it... but I'm not sure how to capture the idea of HTTP-method-based filtering in the name...

rynowak:
Updated. Design tag removed. 

Still need good names. /cc @blowdart

src/Microsoft.AspNet.Mvc.ViewFeatures/AutoValidateAntiforgeryTokenAttribute.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNet.Mvc.Filters;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.AspNet.Mvc.ViewFeatures
```
rynowak: This should be in `.Mvc`

src/Microsoft.AspNet.Mvc.ViewFeatures/IgnoreAntiforgeryTokenAttribute.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNet.Mvc.Filters;
+
+namespace Microsoft.AspNet.Mvc.ViewFeatures
+{
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = false, Inherited = true)]
+    public class IgnoreAntiforgeryTokenAttribute : Attribute, IAntiforgeryPolicy, IOrderedFilter
```
rynowak: Will add docs

src/Microsoft.AspNet.Mvc.ViewFeatures/AutoValidateAntiforgeryTokenAuthorizationFilter.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+            // Determine if this instance is the 'effective' antiforgery policy.
+            for (var i = filters.Count - 1; i >= 0; i--)
+            {
+                var filter = filters[i];
+                if (filter is IAntiforgeryPolicy)
+                {
+                    return object.ReferenceEquals(this, filters[i]);
+                }
+            }
```
rynowak: I had some discussion with @lodejard about formalizing this pattern. We had some neat ideas, but nothing I want to rush out and do right now.

src/Microsoft.AspNet.Mvc.ViewFeatures/AutoValidateAntiforgeryTokenAttribute.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    /// <summary>
+    /// An attribute that causes validation of antiforgery tokens for all unsafe HTTP methods.
```
Eilon: Not sure I like the word "unsafe", especially considering it doesn't define what "unsafe" is. Can it instead just list the exact methods it checks?

src/Microsoft.AspNet.Mvc.ViewFeatures/AutoValidateAntiforgeryTokenAttribute.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// <remarks>
+    /// <see cref="AutoValidateAntiforgeryTokenAttribute"/> can be applied at as a global filter to trigger
+    /// validation of antiforgery tokens by default for an application. Use
+    /// <see cref="IgnoreAntiforgeryTokenAttribute"/> to suppress validation of the antiforgery token for
+    /// a controller or action.
+    /// </remarks>
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = false, Inherited = true)]
+    public class AutoValidateAntiforgeryTokenAttribute : Attribute, IFilterFactory, IOrderedFilter
+    {
+        /// <inheridoc />
```
Eilon: Tpyo in the XML tag (in other places too)

src/Microsoft.AspNet.Mvc.ViewFeatures/AutoValidateAntiforgeryTokenAuthorizationFilter.cs
Eilon: My earlier comment still applies... the code comment here seems highly illogical to me. Or actually, too logical.

src/Microsoft.AspNet.Mvc.ViewFeatures/IgnoreAntiforgeryTokenAttribute.cs
Eilon: Docs

Eilon: Weird, didn't see your comment...

src/Microsoft.AspNet.Mvc.ViewFeatures/AutoValidateAntiforgeryTokenAttribute.cs
rynowak: Fixed

src/Microsoft.AspNet.Mvc.ViewFeatures/IgnoreAntiforgeryTokenAttribute.cs
rynowak: Added

src/Microsoft.AspNet.Mvc.ViewFeatures/AutoValidateAntiforgeryTokenAttribute.cs
rynowak: http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html "Safe" is defined here. I'm using "unsafe" to mean the negation.

> Can it instead just list the exact methods it checks?

HTTP methods are arbitrary text yo. We can specify the ones we **don't** check.

src/Microsoft.AspNet.Mvc.ViewFeatures/AutoValidateAntiforgeryTokenAttribute.cs
rynowak: gah.

src/Microsoft.AspNet.Mvc.ViewFeatures/Internal/AutoValidateAntiforgeryTokenAuthorizationFilter.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        private bool IsClosestAntiforgeryPolicy(IList<IFilterMetadata> filters)
+        {
+            // Determine if this instance is the 'effective' antiforgery policy.
+            for (var i = filters.Count - 1; i >= 0; i--)
+            {
+                var filter = filters[i];
+                if (filter is IAntiforgeryPolicy)
+                {
+                    return object.ReferenceEquals(this, filters[i]);
```
pranavkm: `return object.ReferenceEquals(this, filter);`

src/Microsoft.AspNet.Mvc.ViewFeatures/Internal/ValidateAntiforgeryTokenAuthorizationFilter.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        private bool IsClosestAntiforgeryPolicy(IList<IFilterMetadata> filters)
+        {
+            // Determine if this instance is the 'effective' antiforgery policy.
+            for (var i = filters.Count - 1; i >= 0; i--)
+            {
+                var filter = filters[i];
+                if (filter is IAntiforgeryPolicy)
+                {
+                    return object.ReferenceEquals(this, filters[i]);
```
pranavkm: `return object.ReferenceEquals(this, filters[i]);`

src/Microsoft.AspNet.Mvc.ViewFeatures/Internal/AutoValidateAntiforgeryTokenAuthorizationFilter.cs
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        public Task OnAuthorizationAsync(AuthorizationContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            if (IsUnsafeHttpMethod(context) && IsClosestAntiforgeryPolicy(context.Filters))
```
pranavkm: Could we derive from the other type and move this check to `ShouldValidateRequest`? The `Auto` behavior is a special case of validation so this would be a reasonable candidate for type hierarchy.

rynowak: Yeah, I did htat.

pranavkm:
:shipit:

tinchou:
In case anyone else arrives here searching how to use `AutoValidateAntiforgeryTokenAttribute`, here's how you do it:

```csharp
services.AddMvc(options =>
{
    options.Filters.Add(new AutoValidateAntiforgeryTokenAttribute());
});
```

PR #3786, opened 2015-12-18
Use system.buffers for our reader/writer

rynowak:
/cc @pranavkm


src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
```diff
@@ -90,59 +92,71 @@ public HttpResponseStreamWriter(Stream stream, Encoding encoding, int bufferSize
...
+            _charPool = charPool;
+
+            try
+            {
+                _charBuffer = charPool.Rent(_charBufferSize);
+
+                var requiredLength = encoding.GetMaxByteCount(bufferSize) - encoding.GetPreamble().Length;
+                _byteBuffer = bytePool.Rent(requiredLength);
+            }
+            catch when (_charBuffer != null)
```
pranavkm: 100 points

src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
```diff
@@ -90,59 +92,71 @@ public HttpResponseStreamWriter(Stream stream, Encoding encoding, int bufferSize
...
 
             _encoder = encoding.GetEncoder();
-            _byteBuffer = leasedByteBuffer.Data;
-            _charBuffer = leasedCharBuffer.Data;
+            _bytePool = bytePool;
+            _charPool = charPool;
+
+            try
+            {
+                _charBuffer = charPool.Rent(_charBufferSize);
```
pranavkm: Could this be moved out of the try block? Is there a scenario where the assignment would fail. Ran into this when I was working with the view buffer changes.

pranavkm:
:shipit:

src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
rynowak: yeah, that should be fine.

PR #3787, opened 2015-12-18
Use System.Buffers for JSON.NET

rynowak:
/cc @pranavkm


src/Microsoft.AspNet.Mvc.Formatters.Json/Internal/JsonBufferPool.cs
```diff
@@ -0,0 +1,40 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Buffers;
+using Newtonsoft.Json;
+
+namespace Microsoft.AspNet.Mvc.Formatters.Json.Internal
+{
+    public class JsonBufferPool<T> : IJsonBufferPool<T>
```
pranavkm: Is there a benefit to making it generic as opposed to it being `IJsonBufferPool<char>`?

src/Microsoft.AspNet.Mvc.Formatters.Json/Internal/MvcJsonMvcOptionsSetup.cs
```diff
@@ -12,22 +13,26 @@ namespace Microsoft.AspNet.Mvc.Formatters.Json.Internal
 {
     public class MvcJsonMvcOptionsSetup : ConfigureOptions<MvcOptions>
     {
-        public MvcJsonMvcOptionsSetup(ILoggerFactory loggerFactory, IOptions<MvcJsonOptions> jsonOptions)
-            : base((_) => ConfigureMvc(_, jsonOptions.Value.SerializerSettings, loggerFactory))
+        public MvcJsonMvcOptionsSetup(
+            ILoggerFactory loggerFactory,
+            IOptions<MvcJsonOptions> jsonOptions,
+            ArrayPool<char> charPool)
+            : base((_) => ConfigureMvc(_, jsonOptions.Value.SerializerSettings, loggerFactory, charPool))
```
pranavkm: `_` => `options` since we use it.

src/Microsoft.AspNet.Mvc.Formatters.Json/Internal/JsonBufferPool.cs
```diff
@@ -0,0 +1,40 @@
...
+
+            _inner = inner;
+        }
+
+        public T[] RentBuffer(int minSize)
+        {
+            return _inner.Rent(minSize);
+        }
+
+        public void ReturnBuffer(ref T[] buffer)
```
pranavkm: Where does this get called?

pranavkm:
:watch: for figuring out what the story is for returning the buffers

src/Microsoft.AspNet.Mvc.Formatters.Json/Internal/JsonBufferPool.cs
rynowak: This is called by JSON.NET. They didn't want to take a dependency on System.Buffers.

src/Microsoft.AspNet.Mvc.Formatters.Json/Internal/JsonBufferPool.cs
rynowak: No reason really, just kinda following along since everything else here is generic.

src/Microsoft.AspNet.Mvc.Formatters.Json/Internal/MvcJsonMvcOptionsSetup.cs
rynowak: Sure

rynowak:
I'm not sure what you're not sure about.

pranavkm:
Not sure about this bit: https://github.com/aspnet/Mvc/pull/3787#discussion_r47988700. :shipit:

rynowak:
I answered you BRO. They call that.

JamesNK:
FYI I'm thinking about changing the interface name to IArrayPool when I release this.

rynowak:
> FYI I'm thinking about changing the interface name to IArrayPool when I release this.

Cool, thanks for the heads up. Whenever you have a release of this available, we're planning to bump up our dependency to the latest and merge this change.

image

Compared to our previous conversation and benchmark, notice `char[]` and `byte[]` are super low. This change cuts out 6+mb of `char[]` that was being allocated. :+1:

JamesNK:
Nice. And half the memory allocated compared to a month ago - https://cloud.githubusercontent.com/assets/1430011/11124519/514e15ca-891a-11e5-82e6-88a4db8a59a9.png

JamesNK:
If you are looking for more memory savings there is an optimization in Json.NET that avoids allocating property name strings if the JSON property string exactly matches the JsonProperty.PropertyName. Because you're passing in camel case property names they don't match, e.g. name != Name. If you use a CamelCasePropertyContractResolver that should save string allocations.

rynowak:
> Because you're passing in camel case property names they don't match, e.g. "name" in JSON != "Name" from model. If you use a CamelCasePropertyContractResolver that should save string allocations.

Very cool! I'll add this to my benchmark and rerun.

rynowak:
> And half the memory allocated compared to a month ago:

There's a few big rocks here, some of which are improvements in how I've been measuring.

Measure on CoreCLR. Lots of people have been filing bugs and fixing them on perf, but that doesn't affect desktop (4.6) right away, whereas CoreCLR gets better every day. In this particular benchmark, CoreCLR is about 3mb better (10%) for allocated bytes. This is partly due to improvements to the runtime, and partial due to overhead CoreCLR doesn't have to pay (`LogicalCallContext`).

Get a better tool for load testing. https://www.npmjs.com/package/loadtest I was using ApacheBench in the past. The main problem with AB is that it's HTTP1.0, which means a connection per-request when using chunked-encoding - this isn't a realistic connection/traffic ratio and really acts like more of a stress test for Kestrel (`FrameRequestHeaders`). Additionally, loadtest has throttling. By selecting an RPS value that I can sustain with the profiler on, I can avoid growth of various caches and pools perturbing the measurements (see the delta in `byte[]`).

===============

Some things we're actually doing better in the code:

Stop using `CallContext`/`AsyncLocal<T>` for core framework state. This is a killer for allocations. (`HashTable`).

Starting to optimize routing / link generation. I believe that earlier measurement was allocating 11 `RouteValueDictionary` instances per request. Currently we're at 8. I want to get to 2 (if possible). This requires a lot of finagling, as the cause of these extras is defensive copies the framework does to avoid messing up 'your' data. We're starting to do pooling inside link generation, which should resolve a lot of these.

I'll have another PR this week doing object pooling on `JsonSerializer` instances.

We're also iterating really hard right now on optimizing content negotiation.

I've hinted at it a few other places, but my goal right now is to get this scenario down to 25mb (8.3kb/request).

rynowak:
@JamesNK - can you think of a way we could pool `JsonTextReader` / `JsonTextWriter`? I know that we're changing the underlying reader/writer/stream per request.

JamesNK:
If you can reset the underlying TextReader or TextWriter then you can reuse them today. You will need to set ReadMultipleContent (or something like that) on JsonTextReader but otherwise it should just work. Although if you did that then you would need to throw the JsonTextReader/JsonTextWriter if there was an exception because they would be in an invalid state.

JamesNK:
Are there any examples of a Stream or reader-like type that has a public reset method?

rynowak:
> Are there any examples of a Stream or reader-like type that has a public reset method?

I'm not aware or one. It would be us adding it to our type, or creating a resettable wrapper. It's not something I was planning to try any time soon, just wondering.

rynowak:
@JamesNK - any eta for when you might do a stable release of 8.0?

rynowak:
:tada: https://www.nuget.org/packages/Newtonsoft.Json/ :tada: 

Awesome! Thanks

JamesNK:
About 4 hours ago :grin: 

http://james.newtonking.com/archive/2015/12/20/json-net-8-0-release-1-allocations-and-bug-fixes

src/Microsoft.AspNet.Mvc.Formatters.Json/Internal/JsonBufferPool.cs
JamesNK: Should this class be internal?

rynowak:
This is updated now for `8.0.1`. I'll plan to merge tomorrow morning if there's no other feedback.

src/Microsoft.AspNet.Mvc.Formatters.Json/Internal/JsonArrayPool.cs
```diff
@@ -0,0 +1,39 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Buffers;
+using Newtonsoft.Json;
+
+namespace Microsoft.AspNet.Mvc.Formatters.Json.Internal
+{
+    public class JsonArrayPool<T> : IArrayPool<T>
```
JamesNK: Should this be internal?

rynowak: The `.Internal` namespace is that for us... Hidden from most users, and subject to change/disappearance in future releases.

We don't generally make things `internal` without a good reason.

rynowak:
@JamesNK - I measured after this change with and without the camel case contract resolver. Pretty rad!

**With CamelCase**
image


**With Default**
image

rynowak:
0a9804056e87120a21907552c50836ff038d44e7

Issue #3723, opened 2015-12-10
Render placeholder tag with Display data annotation


rustd:
Copying from UserVoice site https://aspnet.uservoice.com/forums/41201-asp-net-mvc/suggestions/3397091-render-placeholder-tag-with-display-data-annotatio

Support Display(Prompt="...") data annotation in order to render HTML5 placeholder tag with this value

Eilon:
@rynowak assigning to you to track the PR at https://github.com/aspnet/Mvc/pull/3788

rynowak:
8b726ddc0ce208b41a07194af7052dbe14eed99f

PR #3788, opened 2015-12-18
Display annotation prompt placeholder

tuespetre:
Addresses #3723

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
@@ -433,6 +433,20 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding.Metadata
         }
 
         /// <inheritdoc />
+        public override string Placeholder
+        {
+            get
+            {
+                if (DisplayMetadata.Placeholder == null)
+                {
+                    return null;
+                }
+
+                return DisplayMetadata.Placeholder();
+            }
+        }
+
+        /// <inheritdoc />
         public override ModelPropertyCollection Properties
         {
             get
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
@@ -26,6 +26,10 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
         private static readonly MethodInfo ConvertEnumFromStringMethod =
             typeof(DefaultHtmlGenerator).GetTypeInfo().GetDeclaredMethod(nameof(ConvertEnumFromString));
 
+        // See: (http://www.w3.org/TR/html5/forms.html#the-input-element)
+        private static readonly string[] _placeholderInputTypes =
+            new[] { "text", "search", "url", "tel", "email", "password", "number" };
+
         private readonly IAntiforgery _antiforgery;
         private readonly IClientModelValidatorProvider _clientModelValidatorProvider;
         private readonly IModelMetadataProvider _metadataProvider;
@@ -1159,11 +1163,23 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
                     nameof(expression));
             }
 
+            var inputTypeString = GetInputTypeString(inputType);
             var tagBuilder = new TagBuilder("input");
             tagBuilder.TagRenderMode = TagRenderMode.SelfClosing;
             tagBuilder.MergeAttributes(htmlAttributes);
-            tagBuilder.MergeAttribute("type", GetInputTypeString(inputType));
+            tagBuilder.MergeAttribute("type", inputTypeString);
             tagBuilder.MergeAttribute("name", fullName, replaceExisting: true);
+
+            var suppliedTypeString = tagBuilder.Attributes["type"];
+            if (_placeholderInputTypes.Contains(suppliedTypeString))
+            {
+                modelExplorer = modelExplorer ?? ExpressionMetadataProvider.FromStringExpression(expression, viewContext.ViewData, _metadataProvider);
+                var placeholder = modelExplorer.Metadata.Placeholder;
+                if (!string.IsNullOrEmpty(placeholder))
+                {
+                    tagBuilder.MergeAttribute("placeholder", placeholder);
+                }
+            }
 
             var valueParameter = FormatValue(value, format);
             var usedModelState = false;
--- a/test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperTextBoxTest.cs
+++ b/test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperTextBoxTest.cs
@@ -1,0 +1,64 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Mvc.TestCommon;
+using System.ComponentModel.DataAnnotations;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Mvc.Rendering
+{
+    public class HtmlHelperTextBoxTest
+    {
+        [Theory]
+        [InlineData("text")]
+        [InlineData("search")]
+        [InlineData("url")]
+        [InlineData("tel")]
+        [InlineData("email")]
+        [InlineData("number")]
+        public void TextBoxFor_GeneratesPlaceholderAttribute_WhenDisplayAttributePromptIsSetAndTypeIsValid(string type)
+        {
+            // Arrange            
+            var model = new TextBoxModel();
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper(model);
+
+            // Act
+            var result = HtmlContentUtilities.HtmlContentToString(helper.TextBoxFor(m => m.Property1, new { type }));
+
+            // Assert
+            Assert.True(result.Contains(@"placeholder=""HtmlEncode[[placeholder]]"""));
+        }
+
+        [Theory]
+        [InlineData("hidden")]
+        [InlineData("date")]
+        [InlineData("time")]
+        [InlineData("range")]
+        [InlineData("color")]
+        [InlineData("checkbox")]
+        [InlineData("radio")]
+        [InlineData("submit")]
+        [InlineData("reset")]
+        [InlineData("button")]
+        [InlineData("image")]
+        [InlineData("file")]
+        public void TextBoxFor_DoesNotGeneratePlaceholderAttribute_WhenDisplayAttributePromptIsSetAndTypeIsInvalid(string type)
+        {
+            // Arrange            
+            var model = new TextBoxModel();
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper(model);
+
+            // Act
+            var result = HtmlContentUtilities.HtmlContentToString(helper.TextBoxFor(m => m.Property1, new { type }));
+
+            // Assert
+            Assert.False(result.Contains(@"placeholder=""HtmlEncode[[placeholder]]"""));
+        }
+
+        private class TextBoxModel
+        {
+            [Display(Prompt = "placeholder")]
+            public string Property1 { get; set; }
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelMetadata.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelMetadata.cs
@@ -259,6 +259,11 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
         public abstract int Order { get; }
 
         /// <summary>
+        /// Gets the text to display as a placeholder value for an editor.
+        /// </summary>
+        public abstract string Placeholder { get; }
+
+        /// <summary>
         /// Gets the text to display when the model is <c>null</c>.
         /// </summary>
         public abstract string NullDisplayText { get; }
--- a/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Metadata/DisplayMetadata.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Metadata/DisplayMetadata.cs
@@ -116,6 +116,12 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding.Metadata
         public int Order { get; set; } = 10000;
 
         /// <summary>
+        /// Gets or sets a delegate which is used to get a value for the
+        /// model's placeholder text. See <see cref="ModelMetadata.Placeholder"/>.
+        /// </summary>
+        public Func<string> Placeholder { get; set; }
+
+        /// <summary>
         /// Gets or sets a value indicating whether or not to include in the model value in display.
         /// See <see cref="ModelMetadata.ShowForDisplay"/>
         /// </summary>
--- a/src/Microsoft.AspNetCore.Mvc.DataAnnotations/Internal/DataAnnotationsMetadataProvider.cs
+++ b/src/Microsoft.AspNetCore.Mvc.DataAnnotations/Internal/DataAnnotationsMetadataProvider.cs
@@ -84,6 +84,7 @@ namespace Microsoft.AspNetCore.Mvc.DataAnnotations.Internal
             if (displayAttribute != null)
             {
                 displayMetadata.Description = () => displayAttribute.GetDescription();
+                displayMetadata.Placeholder = () => displayAttribute.GetPrompt();
             }
 
             // DisplayFormatString
--- a/test/Microsoft.AspNetCore.Mvc.Abstractions.Test/ModelBinding/ModelMetadataTest.cs
+++ b/test/Microsoft.AspNetCore.Mvc.Abstractions.Test/ModelBinding/ModelMetadataTest.cs
@@ -520,6 +520,14 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
                 }
             }
 
+            public override string Placeholder
+            {
+                get
+                {
+                    throw new NotImplementedException();
+                }
+            }
+
             public override ModelPropertyCollection Properties
             {
                 get
--- a/test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/Metadata/DefaultModelMetadataTest.cs
+++ b/test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/Metadata/DefaultModelMetadataTest.cs
@@ -62,6 +62,7 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding.Metadata
             Assert.Null(metadata.NullDisplayText);
             Assert.Null(metadata.TemplateHint);
             Assert.Null(metadata.SimpleDisplayProperty);
+            Assert.Null(metadata.Placeholder);
 
             Assert.Equal(10000, ModelMetadata.DefaultOrder);
             Assert.Equal(ModelMetadata.DefaultOrder, metadata.Order);
--- a/test/Microsoft.AspNetCore.Mvc.DataAnnotations.Test/Internal/DataAnnotationsMetadataProviderTest.cs
+++ b/test/Microsoft.AspNetCore.Mvc.DataAnnotations.Test/Internal/DataAnnotationsMetadataProviderTest.cs
@@ -26,6 +26,7 @@ namespace Microsoft.AspNetCore.Mvc.DataAnnotations.Internal
                     { new DisplayAttribute() { Description = "d" }, d => d.Description(), "d" },
                     { new DisplayAttribute() { Name = "DN" }, d => d.DisplayName(), "DN" },
                     { new DisplayAttribute() { Order = 3 }, d => d.Order, 3 },
+                    { new DisplayAttribute() { Prompt = "Enter Value" }, d => d.Placeholder(), "Enter Value" },
 
                     { new DisplayColumnAttribute("Property"), d => d.SimpleDisplayProperty, "Property" },
 
--- a/test/Microsoft.AspNetCore.Mvc.DataAnnotations.Test/Internal/ModelMetadataProviderTest.cs
+++ b/test/Microsoft.AspNetCore.Mvc.DataAnnotations.Test/Internal/ModelMetadataProviderTest.cs
@@ -195,6 +195,9 @@ namespace Microsoft.AspNetCore.Mvc.DataAnnotations.Internal
                         new DisplayAttribute { Name = "value" }, metadata => metadata.DisplayName
                     },
                     {
+                        new DisplayAttribute { Prompt = "value" }, metadata => metadata.Placeholder
+                    },
+                    {
                         new DisplayFormatAttribute { DataFormatString = "value" },
                         metadata => metadata.DisplayFormatString
                     },
@@ -420,6 +423,22 @@ namespace Microsoft.AspNetCore.Mvc.DataAnnotations.Internal
 
             // Assert
             Assert.Equal("description", result);
+        }
+
+        [Fact]
+        public void DisplayAttribute_PromptAsPlaceholder()
+        {
+            // Arrange
+            var display = new DisplayAttribute() { Prompt = "prompt" };
+            var provider = CreateProvider(new[] { display });
+
+            var metadata = provider.GetMetadataForType(typeof(string));
+
+            // Act
+            var result = metadata.Placeholder;
+
+            // Assert
+            Assert.Equal("prompt", result);
         }
 
         [Fact]
--- a/test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperPasswordTest.cs
+++ b/test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperPasswordTest.cs
@@ -399,6 +399,21 @@ namespace Microsoft.AspNetCore.Mvc.Rendering
                 HtmlContentUtilities.HtmlContentToString(passwordResult));
         }
 
+        [Fact]
+        public void PasswordFor_GeneratesPlaceholderAttribute_WhenDisplayAttributePromptIsSet()
+        {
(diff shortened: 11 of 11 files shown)


src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DisplayMetadata.cs
```diff
@@ -116,6 +116,12 @@ public class DisplayMetadata
         public int Order { get; set; } = 10000;
 
         /// <summary>
+        /// Gets or sets the placeholder.
```
Eilon: Recommend stating that this is a *delegate* and not the placeholder itself. I recommend using the text from the `Description` attribute because it's the same pattern. The comment might also want to say "a delegate that gets the placeholder *text*".

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
```diff
@@ -25,6 +25,9 @@ public class DefaultHtmlGenerator : IHtmlGenerator
         private const string HiddenListItem = @"<li style=""display:none""></li>";
         private static readonly MethodInfo ConvertEnumFromStringMethod =
             typeof(DefaultHtmlGenerator).GetTypeInfo().GetDeclaredMethod(nameof(ConvertEnumFromString));
+        // See: HTML5 spec, 4.10.5
```
Eilon: Would be nice to link to the W3C spec if possible.

Eilon: (To the specific section.)

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperTextBoxTest.cs
```diff
@@ -0,0 +1,47 @@
...
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Linq.Expressions;
+using Microsoft.AspNet.Mvc.ModelBinding;
+using Microsoft.AspNet.Mvc.TestCommon;
+using Microsoft.AspNet.Mvc.ViewFeatures;
+using Microsoft.AspNet.Testing;
+using Xunit;
+using Microsoft.AspNet.Mvc.ModelBinding.Metadata;
```
Eilon: Sort

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperPasswordTest.cs
```diff
@@ -10,6 +10,7 @@
 using Microsoft.AspNet.Mvc.ViewFeatures;
 using Microsoft.AspNet.Testing;
 using Xunit;
+using Microsoft.AspNet.Mvc.ModelBinding.Metadata;
```
Eilon: Sort

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperTextBoxTest.cs
```diff
@@ -0,0 +1,47 @@
...
+
+namespace Microsoft.AspNet.Mvc.Rendering
+{
+    public class HtmlHelperTextBoxTest
```
Eilon: Nice test! Would be good to also add negative tests: That is, make sure that even with a `Prompt` set, that "incompatible" input types (ones not in the list you created) do *not* get placeholder text.

Eilon:
@tuespetre thanks for the PR! Apologies for the delay - it's holiday season here and only about 10% of the people are at the office :smile:

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
```diff
@@ -432,6 +432,20 @@ public override int Order
...
+        public override string Placeholder
+        {
+            get
+            {
+                if (DisplayMetadata.Placeholder == null)
+                {
+                    return null;
+                }
+
+                return DisplayMetadata.Placeholder();
```
rynowak: nit: This can be simplified to `DisplayMetadata.Placeholder?.Invoke()` if you like.

rynowak: Would be nice to add a comment here that this shouldn't be cached because it's localizable.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
```diff
@@ -1156,12 +1159,23 @@ internal static object GetModelStateValue(ViewContext viewContext, string key, T
...
             tagBuilder.MergeAttributes(htmlAttributes);
-            tagBuilder.MergeAttribute("type", GetInputTypeString(inputType));
+            tagBuilder.MergeAttribute("type", inputTypeString);
             tagBuilder.MergeAttribute("name", fullName, replaceExisting: true);
 
+            if (_placeholderInputTypes.Contains(inputTypeString))
+            {
+                modelExplorer = modelExplorer ?? ExpressionMetadataProvider.FromStringExpression(expression, viewContext.ViewData, _metadataProvider);
+
+                if (modelExplorer.Metadata.Placeholder != null)
```
rynowak: We should cache this in a local since it's calling into arbitrary user code.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperTextBoxTest.cs
```diff
@@ -0,0 +1,47 @@
...
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper(model);
+
+            // Act
+            var result = helper.TextBoxFor(m => m.Property1, new { type });
+
+            // Assert
+            Assert.Equal(string.Format(expectedResult, type), HtmlContentUtilities.HtmlContentToString(result));
+        }
+
+        public class TextBoxModel
```
rynowak: private

tuespetre:
Thanks for the feedback! Between getting out of the holidays and getting over being sick, I am finally ready to address these points. I'll get to them as soon as I catch up on what I've missed here at my work.

Eilon:
@tuespetre thanks, looking forward to it.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
tuespetre: When I was messing around with my own micro ORM I found that `Invoke()` was slower than a simple `()` invocation; I don't know if that's changed at all so I'll leave it 'expanded' but I will add the comment. Thanks!

tuespetre:
My checklist:

- [ ] https://github.com/aspnet/Mvc/pull/3788#discussion_r48496142
- [ ] https://github.com/aspnet/Mvc/pull/3788#discussion_r48496163
- [ ] https://github.com/aspnet/Mvc/pull/3788#discussion_r48496229
- [ ] https://github.com/aspnet/Mvc/pull/3788#discussion_r48496239
- [ ] https://github.com/aspnet/Mvc/pull/3788#discussion_r48496277
- [ ] https://github.com/aspnet/Mvc/pull/3788#discussion_r48496380
- [ ] https://github.com/aspnet/Mvc/pull/3788#discussion_r48499078
- [ ] https://github.com/aspnet/Mvc/pull/3788#discussion_r48499131

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
tuespetre: If you mean the ExpressionMetadataProvider bit, I think I'll leave that one alone as I see there are four total such usages in this class and I'm not sure what the team would like to do with that.

tuespetre:
Changes made and pushed up. Thanks again

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
```diff
@@ -1154,12 +1157,52 @@ internal static object GetModelStateValue(ViewContext viewContext, string key, T
...
+
+            if (inputType == InputType.Text || inputType == InputType.Password)
+            {
+                object suppliedTypeString;
+                if (htmlAttributes != null &&
+                    htmlAttributes.TryGetValue("type", out suppliedTypeString) &&
+                    suppliedTypeString != null &&
+                    _placeholderInputTypes.Contains(suppliedTypeString))
+                {
+                }
```
rynowak: @tuespetre - what goes here?

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
```diff
@@ -1154,12 +1157,52 @@ internal static object GetModelStateValue(ViewContext viewContext, string key, T
...
+                    }
+                    else if (inputTypeString == "text")
+                    {
+                        goto case InputType.Password;
+                    }
+                    break;
+
+                case InputType.Password:
+                    modelExplorer = modelExplorer ?? ExpressionMetadataProvider.FromStringExpression(expression, viewContext.ViewData, _metadataProvider);
+                    if (modelExplorer.Metadata.Placeholder != null)
```
rynowak: What I meant in [this comment](https://github.com/aspnet/Mvc/pull/3788#discussion_r48499078) is that calling `.Placeholder` invokes some user code which might call back into localization, pull this out into a variable.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
```diff
@@ -1154,12 +1157,52 @@ internal static object GetModelStateValue(ViewContext viewContext, string key, T
...
+                        suppliedTypeString != null)
+                    {
+                        if (_placeholderInputTypes.Contains(suppliedTypeString))
+                        {
+                            goto case InputType.Password;
+                        }
+                    }
+                    else if (inputTypeString == "text")
+                    {
+                        goto case InputType.Password;
```
rynowak: This is a really strange way to formulate this logic. It looks like this is in the middle of a refactor and not quite finished.

 I'd expect to see something like:

```
var canIncludePlaceholder false;

switch
{
case InputType.Text:
  if (...) // conditions for text
  {
    canIncludePlaceholder = true;
  }
  break;
case InputType.Text:
  if (...) // conditions for password
  {
    canIncludePlaceholder = true;
  }
  break;
}

if (canIncludePlaceholder)
{
   // Add the placeholder
}
```

rynowak:
:watch:

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
tuespetre: Very true. I want to say was trying to avoid declaring a local but with all I've learned in the past week about MSIL I can now see why that was silly.

- Derek Gray

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
tuespetre: Isn't that what was desired though? To not cache the result of invoking `Description()`, `DisplayName()`, and `Placeholder()`? Or should the `DefaultModelMetadata` invoke those at most one time and cache them in private fields?

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
tuespetre: Well, apparently I left my head there. Thank you! :fearful:

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
```diff
@@ -32,6 +32,12 @@ public class DefaultModelMetadata : ModelMetadata
         private ModelPropertyCollection _properties;
         private bool? _validateChildren;
         private ReadOnlyCollection<object> _validatorMetadata;
+        private bool _invokedDescription = false;
```
Eilon: BTW don't need to set any of these default values.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
```diff
@@ -32,6 +32,12 @@ public class DefaultModelMetadata : ModelMetadata
...
+        private string _cachedDescription = null;
+        private bool _invokedDisplayName = false;
+        private string _cachedDisplayName = null;
+        private bool _invokedPlaceholder = false;
+        private string _cachedPlaceholder = null;
```
dougbu: None of these values should be cached. They are culture-specific and `ModelMetadata` instances have very long lives.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
```diff
@@ -25,6 +25,9 @@ public class DefaultHtmlGenerator : IHtmlGenerator
         private const string HiddenListItem = @"<li style=""display:none""></li>";
         private static readonly MethodInfo ConvertEnumFromStringMethod =
             typeof(DefaultHtmlGenerator).GetTypeInfo().GetDeclaredMethod(nameof(ConvertEnumFromString));
+        // See: (http://www.w3.org/TR/html5/forms.html#the-input-element)
```
dougbu: Minor but a blank line before this comment should make it more readable.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
```diff
@@ -1159,12 +1162,47 @@ internal static object GetModelStateValue(ViewContext viewContext, string key, T
...
+            var inputTypeString = GetInputTypeString(inputType);
             var tagBuilder = new TagBuilder("input");
             tagBuilder.TagRenderMode = TagRenderMode.SelfClosing;
             tagBuilder.MergeAttributes(htmlAttributes);
-            tagBuilder.MergeAttribute("type", GetInputTypeString(inputType));
+            tagBuilder.MergeAttribute("type", inputTypeString);
             tagBuilder.MergeAttribute("name", fullName, replaceExisting: true);
 
+            var tryIncludingPlaceholder = false;
+            switch (inputType)
```
dougbu: Suggest merging this into the existing `switch` on `inputType`.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
tuespetre: Ah, ok. I wasn't aware that they have long lives. I suppose now I am just not sure of what @rynowak means about storing `Placeholder` in a local.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
dougbu: Actually, the only guaranteed thing about the `inputType` is that it controls where `value` is found. Instead of all of this, do the following after the existing `switch`:
``` c#
    string typeString = tagBuilder.Attributes["type"];
    if (_placeholderInputTypes.Contains(suppliedTypeString))
    {
        modelExplorer = modelExplorer ??
            ExpressionMetadataProvider.FromStringExpression(expression, viewContext.ViewData, _metadataProvider);
        if (modelExplorer.Metadata.Placeholder != null)
        {
            tagBuilder.MergeAttribute("placeholder", modelExplorer.Metadata.Placeholder);
        }
    }
```

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperTextBoxTest.cs
```diff
@@ -0,0 +1,50 @@
...
+        [Theory]
+        [InlineData("text", InputWithPlaceholderAttribute)]
+        [InlineData("search", InputWithPlaceholderAttribute)]
+        [InlineData("url", InputWithPlaceholderAttribute)]
+        [InlineData("tel", InputWithPlaceholderAttribute)]
+        [InlineData("email", InputWithPlaceholderAttribute)]
+        [InlineData("number", InputWithPlaceholderAttribute)]
+        [InlineData("date", InputWithoutPlaceholderAttribute)]
+        [InlineData("color", InputWithoutPlaceholderAttribute)]
+        public void TextBoxFor_GeneratesPlaceholderAttribute_WhenDisplayAttributePromptIsSet(string type, string expectedResult)
```
dougbu: Might be slightly more clear if split into two `[Theory]`s, one that checks output with the placeholder and one without.

dougbu:
:watch:

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
tuespetre: Nevermind, I totally see it now. Sorry @rynowak!

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperTextBoxTest.cs
tuespetre: Agreed. Will do that and make sure the `InlineData` covers all of the relevant `type` attribute values for `input` between both theories.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
tuespetre: Just verified that works. Thanks @dougbu

tuespetre:
I've completed those items as of yesterday; just pushed again to resolve the conflict

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
```diff
@@ -1159,12 +1163,24 @@ internal static object GetModelStateValue(ViewContext viewContext, string key, T
...
+            if (_placeholderInputTypes.Contains(suppliedTypeString))
+            {
+                modelExplorer = modelExplorer ?? ExpressionMetadataProvider.FromStringExpression(expression, viewContext.ViewData, _metadataProvider);
+                var placeholder = modelExplorer.Metadata.Placeholder;
+                if (placeholder != null)
+                {
+                    tagBuilder.MergeAttribute("placeholder", placeholder);
+                }
+            }
+
```
rynowak: This is 100% better. :fireworks: :100: :fireworks:

rynowak:
To address this comment: https://github.com/aspnet/Mvc/pull/3788#discussion_r50894328

Don't sweat stuff like this, we optimize for readability/maintainability. If there's a need to follow a specific pattern for perf, we'll let you know.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
rynowak: @dougbu is this right or is the `switch` here needed?

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
```diff
@@ -203,12 +203,12 @@ public override string Description
...
             {
-                if (DisplayMetadata.Description == null)
+                if (DisplayMetadata.Description != null)
                 {
-                    return null;
+                    return DisplayMetadata.Description();
                 }
 
-                return DisplayMetadata.Description();
+                return null;
```
dougbu: We tend to leave existing style alone -- except in the case of bizarre and hard to maintain line wrapping. Why did this and `DisplayName` change at all?

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
dougbu: @rynowak no need for a `switch`.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
```diff
@@ -1159,12 +1163,24 @@ internal static object GetModelStateValue(ViewContext viewContext, string key, T
...
-            tagBuilder.MergeAttribute("type", GetInputTypeString(inputType));
+            tagBuilder.MergeAttribute("type", inputTypeString);
             tagBuilder.MergeAttribute("name", fullName, replaceExisting: true);
 
+            var suppliedTypeString = tagBuilder.Attributes["type"];
+            if (_placeholderInputTypes.Contains(suppliedTypeString))
+            {
+                modelExplorer = modelExplorer ?? ExpressionMetadataProvider.FromStringExpression(expression, viewContext.ViewData, _metadataProvider);
+                var placeholder = modelExplorer.Metadata.Placeholder;
+                if (placeholder != null)
```
dougbu: @tuespetre shouldn't this be `string.IsNullOrEmpty(placeholder)`? I.e. when is it useful to generate `placeholder=""`?

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperTextBoxTest.cs
```diff
@@ -0,0 +1,69 @@
...
+        [InlineData("time")]
+        [InlineData("range")]
+        [InlineData("color")]
+        [InlineData("checkbox")]
+        [InlineData("radio")]
+        [InlineData("submit")]
+        [InlineData("reset")]
+        [InlineData("button")]
+        // Skipping these two because they won't match the constant string,
+        // only because their 'value' attribute also does not get set.
```
dougbu: Is this really a comment about `_placeholderInputTypes` in `DefaultHtmlGenerator`? If yes, should be moved.

Either way, I'm not quite sure what this sentence is saying. Could you rephrase a bit?

dougbu:
:shipit: my comments are minor though I would like to hear from @tuespetre on https://github.com/aspnet/Mvc/pull/3788#discussion-diff-51012982.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadata.cs
tuespetre: Sorry, I didn't even consider that. They changed because I was reversing the mistake I made about caching the values.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
tuespetre: Good catch

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperTextBoxTest.cs
tuespetre: I will amend this comment with my proper anglish skillz. :wink:

tuespetre: I take that back; I'm going to amend the tests to simply check for whether the result contains the placeholder attribute rather than comparing against a formatted constant.

rynowak: Why? These tests seem fine as is

tuespetre: @rynowak that way the tests will cover every possible `type` attribute value defined for `<input>` by the HTML5 spec.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperTextBoxTest.cs
```diff
@@ -0,0 +1,64 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Mvc.TestCommon;
+using System.ComponentModel.DataAnnotations;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Mvc.Rendering
+{
+    public class HtmlHelperTextBoxTest
```
dougbu: @tuespetre not :100:% required: Please add similar test methods to `InputTagHelperTest`? It uses the same code paths through `DefaultHtmlGenerator` and passes in an `htmlAttributes` argument that includes the `type` which will really be generated. Definitely do not need to double-check _all_ of these values.

dougbu:
Thanx @tuespetre

rynowak:
Merged, thanks! 8b726ddc0ce208b41a07194af7052dbe14eed99f

PR #3790, opened 2015-12-18
Remove TextWriter.ToString from RazorPage

### LINKED ISSUE #3668: Remove TextWriter.ToString from RazorPage
Based on discussion here - https://github.com/aspnet/Mvc/pull/3664#discussion_r46205297

There's code in RazorPage (https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs#L267) where we invoke `ToString` on a `TextWriter` instance. This is likely to succeed only with specific kinds of `TextWriter`s such as `StringWriter`. We should remove this behavior.

pranavkm:
Fixes #3668


pranavkm:
cc @NTaylorMullen

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -209,30 +209,15 @@ private IViewBufferScope BufferScope
...
-        }
-
-        /// <summary>
-        /// Starts a new writing scope with the given <paramref name="writer"/>.
-        /// </summary>
-        /// <remarks>
-        /// All writes to the <see cref="Output"/> or <see cref="ViewContext.Writer"/> after calling this method will
-        /// be buffered until <see cref="EndTagHelperWritingScope"/> is called.
-        /// </remarks>
-        public void StartTagHelperWritingScope(TextWriter writer)
```
pranavkm: @NTaylorMullen it doesn't look like we use this method (none of the functional tests rely on this). Is there a scenario I'm missing that isn't covered by the tests?

src/Microsoft.AspNet.Mvc.Razor/TagHelperContentWrapperTextWriter.cs
```diff
@@ -1,78 +0,0 @@
...
-using System.Text;
-using Microsoft.AspNet.Html;
-using Microsoft.AspNet.Razor.TagHelpers;
-
-namespace Microsoft.AspNet.Mvc.Razor
-{
-    /// <summary>
-    /// <see cref="HtmlTextWriter"/> implementation which writes to a <see cref="TagHelperContent"/> instance.
-    /// </summary>
-    public class TagHelperContentWrapperTextWriter : HtmlTextWriter
```
pranavkm: Should've removed it in an earlier PR, but this is entirely unused now.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
NTaylorMullen: Nope, we used to use it then removed the usage. Looks like we just didn't clean this up.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -267,24 +253,7 @@ public TagHelperContent EndTagHelperWritingScope()
             }
 
             var tagHelperContent = new DefaultTagHelperContent();
```
pranavkm: Need to update docs.

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -267,24 +253,7 @@ public TagHelperContent EndTagHelperWritingScope()
...
-                {
-                    tagHelperContent.Append(htmlContentTextWriter.ContentBuilder);
-                }
-                else
-                {
-                    tagHelperContent.AppendHtml(writer.ToString());
-                }
-            }
-
+            tagHelperContent.Append(writer.ContentBuilder);
```
pranavkm: Might be a negligibly small optimization but would be nice if we could pass the ContentBuilder into the ctor of the `DefaultTagHelperContent`. This would help avoid the backing `List<IHtmlContent>` from resizes.

NTaylorMullen:
:shipit:

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
NTaylorMullen: If it still appended each item in the buffer then I could definitely see a copy ctor being valuable.

pranavkm: Yeah, it would allocate a List of size 1 and append it. Might be pretty marginal

PR #3796, opened 2015-12-19
Simplifying MvcTestFixture

pranavkm:


pranavkm:
cc @davidfowl \ @JunTaoLuo

test/Microsoft.AspNet.Mvc.FunctionalTests/MvcTestFixture.cs
```diff
@@ -89,6 +57,7 @@ protected virtual void AddAdditionalServices(IServiceCollection services)
             // points to the root folder of the test project.
             // To compensate, we need to calculate the correct project path and override the application
             // environment value so that components like the view engine work properly in the context of the test.
+            var startupAssembly = typeof(TStartup).GetTypeInfo().Assembly;
             var applicationName = startupAssembly.GetName().Name;
             var library = libraryManager.GetLibrary(applicationName);
             var applicationRoot = Path.GetDirectoryName(library.Path);
```
davidfowl: Do you need to reset the `hostingEnvironment` still? Doesn't it rebase itself on the new app base?

pranavkm: Haven't checked. Let me verify.

pranavkm: Nope. Removed

test/Microsoft.AspNet.Mvc.FunctionalTests/MvcTestFixture.cs
```diff
@@ -2,60 +2,34 @@
...
             // localization i.e. RequestLocalizationMiddleware behavior. Ensure the saved culture
             // (DefaultRequestCulture) is consistent regardless of system configuration or personal preferences.
             using (new CultureReplacer())
             {
                 var builder = new WebApplicationBuilder()
-                    .Configure(configureApplication)
-                    .ConfigureServices(
-                        services => InitializeServices(startupTypeInfo.Assembly, services, configureServices));
+                    .ConfigureServices(InitializeServices)
+                    .UseStartup(typeof(TStartup));
```
davidfowl: nit: `UseStartup<TStartup>()`?

test/Microsoft.AspNet.Mvc.FunctionalTests/MvcTestFixture.cs
```diff
@@ -98,21 +66,11 @@ protected virtual void AddAdditionalServices(IServiceCollection services)
...
             assemblyProvider.CandidateAssemblies.Add(startupAssembly);
             services.AddSingleton<IAssemblyProvider>(assemblyProvider);
-
-            AddAdditionalServices(services);
-
-            if (configureServices != null)
-            {
-                configureServices(services);
-            }
+            services.AddSingleton<ILoggerFactory>(NullLoggerFactory.Instance);
```
davidfowl: Why?

davidfowl:
Looks good besides CI failure?

test/Microsoft.AspNet.Mvc.FunctionalTests/MvcTestFixture.cs
pranavkm: I end up having to annotate `TStartup : new()`. It's less code this way :)

rynowak:
@pranavkm - this is super nice cleanup.

The other part of this item... what else would a **user** need to do to integration test their site? What tasks is the fixture currently performing that the hosting builder is not performing? (Ignore things like test logger and test encoder, which our users probably don't care about by default).

davidfowl:
@rynowak setting the app base is the main one. If we override that by handling it at the `IApplicationEnvironment` layer then `AppContext.BaseDirectory` won't work but maybe that's fine and we don't care?

test/Microsoft.AspNet.Mvc.FunctionalTests/MvcTestFixture.cs
pranavkm: This is a bug in the `TestHost` implementation. Some apps register a ConsoleLogger which results in a null ref here: https://github.com/aspnet/Hosting/blob/0f0999161d5dfb12feba90f26bbc05795fdf00b3/src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs#L144

```C#
System.NullReferenceException : Object reference not set to an instance of an object.
Stack Trace:
   at Microsoft.AspNet.Http.Internal.DefaultHttpRequest.get_Protocol()
   at Microsoft.AspNet.Hosting.Internal.HostingLoggerExtensions.HostingRequestStarting.ToString()
   at Microsoft.AspNet.Hosting.Internal.HostingLoggerExtensions.HostingRequestStarting.<>c.<.cctor>b__7_0(Object state, Exception exception)
   at Microsoft.Extensions.Logging.Console.ConsoleLogger.Log(LogLevel logLevel, Int32 eventId, Object state, Exception exception, Func`3 formatter)
   at Microsoft.Extensions.Logging.Logger.Log(LogLevel logLevel, Int32 eventId, Object state, Exception exception, Func`3 formatter)
```
I'll file a bug for it.

pranavkm:
The StaticAssemblyProvider bit is also something a user would need to setup. If we change Mvc to not perform controller discovery and instead use a list provided at startup, this would go away.

> AppContext.BaseDirectory won't work but maybe that's fine and we don't care?

That seems fine to me.

Issue #3718, opened 2015-12-10
ViewLocalizer should look for resource name with the application name prefix when LocalizationOptions.ResourcesPath is not set


DamianEdwards:
Like #3717

When `LocalizationOptions.ResourcesPath` is not set, `ViewLocalizer` should prepend the view execution path with the application name.

Let's also trim off the ".cshtml" file extension as having to include that in the resx file name is just cruel.

e.g. where `View.ExecutingFilePath` is `MyApp/Views/Home/Index.cshtml` and `LocalizationOptions.ResourcesPath` is not set, `ViewLocalizer` would look for a resource with base name "MyApp.Views.Home.Index".

DamianEdwards:
Let's also trim off the ".cshtml" file extension as having to include that in the resx file name is just cruel.

Issue #2767, opened 2015-06-30
[Localization]: Fix ViewLocalizer so it works for layouts too


kirthik:
`ViewLocalizer` cannot be used on layout pages since the logic in `Contextualize` method reads current view to get the resources file. Since `ViewContext.View` property does not point to layout,  layout specific resources are not localized. Today we have to use `IStringLocalizer` for this scenario.

We should come up with a solution where `ViewLocalizer` works for layouts too.

danroth27:
@damian

Eilon:
@DamianEdwards do we want this for beta8? If so, we need to discuss a design.

dougbu:
Hit this issue while experimenting. For the same reason, localization does not work as expected in `_ViewStart.cshtml` files e.g. when setting a default `ViewData["Message"]` value. This breaks componentization and means resources must be repeated for all top-level views.

Suggest `ViewLocalizer` [line 101](https://github.com/aspnet/Mvc/blob/490fcf1ab4d416ee6cc140797c2377e352677bda/src/Microsoft.AspNet.Mvc.Localization/ViewLocalizer.cs#L101) should use `viewContext.ExecutingFilePath` and not the current `viewContext.View.Path`. The first changes much more frequently.

Eilon:
Can you clarify? Do you mean that _ViewStart is picking up resources from the real view that's running? If so, I agree that's wrong.

dougbu:
> `_ViewStart` [and `_Layout`] is picking up resources from the real view that's running?

Yes, that's what I and @kirthik meant. Fortunately one design (thinking of the Needs Design label) is a one-line fix.

DamianEdwards:
ViewLocalizer should use ViewContext.ExecutingFilePath when constructing the resource base name

PR #3799, opened 2015-12-21
Improve usability of ViewLocalizer

DamianEdwards:
Changes to `ViewLocalizer` resource lookup semantics:
- Now uses `ViewContext.ExecutingFilePath` before `ViewContext.View.Path` for determining base name
- Trims any file extension from the base name
- Always prepends the application name (note the `ResourceManagerStringLocalizerFactory`'s new behavior WRT the application name prefix makes this all work well)
- Updated the loc sample app to match (also used by the functional tests)
- Added unit tests to cover this behavior
- #3718
- #2767

@ryanbrandenburg @rynowak


src/Microsoft.AspNet.Mvc.Localization/IViewLocalizer.cs
```diff
@@ -4,7 +4,7 @@
 namespace Microsoft.AspNet.Mvc.Localization
 {
     /// <summary>
-    /// A service that provides localized strings for views.
+    /// Represents a type that provides HTML-aware localization for views.
```
ryanbrandenburg: Is this our usual wording? Seems kinda off.

src/Microsoft.AspNet.Mvc.Localization/ViewLocalizer.cs
```diff
@@ -93,12 +96,30 @@ public void Contextualize(ViewContext viewContext)
...
+
+            // Trim the file extension from the end of the path
+            if (!string.IsNullOrEmpty(path) && Path.HasExtension(path))
             {
-                baseName = baseName.Substring(1);
+                var extension = Path.GetExtension(path);
+                path = path.Substring(0, path.Length - extension.Length);
             }
 
+            Debug.Assert(!string.IsNullOrEmpty(path), "Couldn't determine a path for the view");
```
ryanbrandenburg: Just a reminder that this is here, fine to leave it but don't know if it's intentional.

DamianEdwards: It was intentional.

src/Microsoft.AspNet.Mvc.Localization/IViewLocalizer.cs
DamianEdwards: This is standard .NET API reference docs language, e.g. https://msdn.microsoft.com/en-us/library/s4ys34ea(v=vs.110).aspx

src/Microsoft.AspNet.Mvc.Localization/ViewLocalizer.cs
```diff
@@ -93,12 +96,30 @@ public void Contextualize(ViewContext viewContext)
...
 
-            var baseName = viewContext.View.Path.Replace('/', '.').Replace('\\', '.');
-            if (baseName.StartsWith(".", StringComparison.OrdinalIgnoreCase))
+            // Given a view path "/Views/Home/Index.cshtml" we want a baseName like "MyApplication.Views.Home.Index"
+
+            var path = viewContext.ExecutingFilePath;
+
+            if (string.IsNullOrEmpty(path))
+            {
+                path = viewContext.View.Path;
```
ryanbrandenburg: Love this change, having the "cshtml" in there always struck me as super weird.

ryanbrandenburg:
:shipit: from me since you addressed all my concerns.

src/Microsoft.AspNet.Mvc.Localization/ViewLocalizer.cs
```diff
@@ -93,12 +96,30 @@ public void Contextualize(ViewContext viewContext)
...
+            }
+
+            // Trim the file extension from the end of the path
+            if (!string.IsNullOrEmpty(path) && Path.HasExtension(path))
```
pranavkm: Might want to consider using a `StringBuilder` to do these replaces \ substrings if it happens in ordinary cases.

DamianEdwards: Too late. Log an issue?

pranavkm: Sure. Depends on how common these cases are. Although my guess is nearly all view paths start with a leading slash and have an extension so it's not uncommon.

PR #3800, opened 2015-12-22
Make ViewLocalizer base name generation more efficient:

DamianEdwards:
- Because @pranavkm made me do it
- And it is actually faster and less allocating than the old code (~25% faster)
- Added some more test cases

@pranavkm @ryanbrandenburg


pranavkm:
:shipit:

ryanbrandenburg:
Yay for performance, :shipit:

PR #3801, opened 2015-12-22
Modify FileVersionProvider to cache missing file info.

### LINKED ISSUE #3765: Script and Link tags hit the disk on every request
Path normalization done in the file provider is extremely expensive in terms of allocations.

This is about 3.7mb allocated over 3000 requests
image

In context, this is about 2.5% of the total allocated bytes

pranavkm:
Fixes #3765

### DIFF
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
@@ -81,16 +81,16 @@ namespace Microsoft.AspNet.Mvc.TagHelpers.Internal
             if (!_cache.TryGetValue(path, out value))
             {
                 var cacheEntryOptions = new MemoryCacheEntryOptions();
-                var fileInfo = _fileProvider.GetFileInfo(resolvedPath);
                 cacheEntryOptions.AddExpirationToken(_fileProvider.Watch(resolvedPath));
+                var fileInfo = _fileProvider.GetFileInfo(resolvedPath);
 
                 if (!fileInfo.Exists &&
                     _requestPathBase.HasValue &&
                     resolvedPath.StartsWith(_requestPathBase.Value, StringComparison.OrdinalIgnoreCase))
                 {
-                    resolvedPath = resolvedPath.Substring(_requestPathBase.Value.Length);
-                    fileInfo = _fileProvider.GetFileInfo(resolvedPath);
-                    cacheEntryOptions.AddExpirationToken(_fileProvider.Watch(resolvedPath));
+                    var requestPathBaseRelativePath = resolvedPath.Substring(_requestPathBase.Value.Length);
+                    cacheEntryOptions.AddExpirationToken(_fileProvider.Watch(requestPathBaseRelativePath));
+                    fileInfo = _fileProvider.GetFileInfo(requestPathBaseRelativePath);
                 }
 
                 if (fileInfo.Exists)
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
@@ -16,6 +16,7 @@ namespace Microsoft.AspNet.Mvc.TagHelpers.Internal
     public class FileVersionProvider
     {
         private const string VersionKey = "v";
+        private static readonly char[] QueryStringAndFragmentTokens = new [] { '?', '#' };
         private readonly IFileProvider _fileProvider;
         private readonly IMemoryCache _cache;
         private readonly PathString _requestPathBase;
@@ -24,8 +25,8 @@ namespace Microsoft.AspNet.Mvc.TagHelpers.Internal
         /// Creates a new instance of <see cref="FileVersionProvider"/>.
         /// </summary>
         /// <param name="fileProvider">The file provider to get and watch files.</param>
-        /// <param name="applicationName">Name of the application.</param>
         /// <param name="cache"><see cref="IMemoryCache"/> where versioned urls of files are cached.</param>
+        /// <param name="requestPathBase">The base path for the current HTTP request.</param>
         public FileVersionProvider(
             IFileProvider fileProvider,
             IMemoryCache cache,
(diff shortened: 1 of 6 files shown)


pranavkm:
cc @rynowak \ @ajaybhargavb

pranavkm:
cc @NTaylorMullen (since @ajaybhargavb is OOF today)

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
```diff
@@ -76,35 +77,39 @@ public string AddFileVersionToPath(string path)
                 return path;
             }
 
-            var fileInfo = _fileProvider.GetFileInfo(resolvedPath);
-            if (!fileInfo.Exists)
+            string value;
+            if (!_cache.TryGetValue(path, out value))
```
pranavkm: This isn't new behavior, but this cache instance is shared with the user. We should use a better key than this to avoid collisions.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/Internal/FileVersionProviderTest.cs
```diff
@@ -38,54 +37,188 @@ public void AddsVersionToFiles_WhenCacheIsAbsent(string filePath, string expecte
...
+                pathStartsWithAppName: false,
+                fileDoesNotExist: true);
+            var fileVersionProvider = new FileVersionProvider(
+                fileProvider,
+                new MemoryCache(new MemoryCacheOptions()),
+                GetRequestPathBase());
+            var mockFileProvider = Mock.Get(fileProvider);
+
+            // Act 1
+            var result = fileVersionProvider.AddFileVersionToPath(path);
```
ajaybhargavb: Isn't it enough just to call this twice and do the `Times.Once()` asserts once? Do we need two acts and asserts?

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
ajaybhargavb: :+1:

ajaybhargavb:
:watch: for one small clarification

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
NTaylorMullen: We could hash the query string/fragment bits.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
```diff
@@ -76,35 +77,39 @@ public string AddFileVersionToPath(string path)
...
             {
-                if (_requestPathBase.HasValue &&
+                var cacheEntryOptions = new MemoryCacheEntryOptions();
+                var fileInfo = _fileProvider.GetFileInfo(resolvedPath);
+                cacheEntryOptions.AddExpirationToken(_fileProvider.Watch(resolvedPath));
```
NTaylorMullen: Move up one line to be next to the `cacheEntryOptions`.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
```diff
@@ -76,35 +77,39 @@ public string AddFileVersionToPath(string path)
...
+
+                if (!fileInfo.Exists &&
+                    _requestPathBase.HasValue &&
                     resolvedPath.StartsWith(_requestPathBase.Value, StringComparison.OrdinalIgnoreCase))
                 {
                     resolvedPath = resolvedPath.Substring(_requestPathBase.Value.Length);
                     fileInfo = _fileProvider.GetFileInfo(resolvedPath);
+                    cacheEntryOptions.AddExpirationToken(_fileProvider.Watch(resolvedPath));
```
NTaylorMullen: Up one, confused me at firsti because I thought you were doing the same thing twice. WHen in reality you're recaclculating the resolvedPath.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
```diff
@@ -76,35 +77,39 @@ public string AddFileVersionToPath(string path)
...
                 }
-            }
 
-            string value;
-            if (!_cache.TryGetValue(path, out value))
-            {
-                value = QueryHelpers.AddQueryString(path, VersionKey, GetHashForFile(fileInfo));
-                var cacheEntryOptions = new MemoryCacheEntryOptions().AddExpirationToken(_fileProvider.Watch(resolvedPath));
-                _cache.Set(path, value, cacheEntryOptions);
+                value = _cache.Set<string>(path, value, cacheEntryOptions);
```
NTaylorMullen: Hmm, do we want to care about thread safety? Could end up with a lotttt of extra file watchers, that is unless the `_fileProvider` caches them.

NTaylorMullen:
:watch:

test/Microsoft.AspNet.Mvc.TagHelpers.Test/Internal/FileVersionProviderTest.cs
pranavkm: I guess. Just a preference to separate out the two separate actions

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
pranavkm: Or use a type with a comparer. Makes it more reliable than strings.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileVersionProvider.cs
pranavkm: IChangeTokens created by `PhysicalFileSystem` are cached. Is that what you were concerned about?

NTaylorMullen: Ya that was my concern.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/Internal/FileVersionProviderTest.cs
pranavkm: Plus it also doesn't look like an accidental line copy.

pranavkm:
Updated

NTaylorMullen:
:shipit:

PR #3821, opened 2015-12-29
FormatFilter overrides content type set explicitly by the developer

kichalla:
@rynowak

### DIFF
--- a/samples/FormatFilterSample.Web/Startup.cs
+++ b/samples/FormatFilterSample.Web/Startup.cs
@@ -15,7 +15,7 @@ namespace FormatFilterSample.Web
         // Set up application services
         public void ConfigureServices(IServiceCollection services)
         {
-            services.AddMvc(options =>
+            var mvcBuilder = services.AddMvc(options =>
             {
                 var formatFilter = new FormatFilterAttribute();
                 options.Filters.Add(formatFilter);
@@ -28,6 +28,8 @@ namespace FormatFilterSample.Web
                     "custom",
                     MediaTypeHeaderValue.Parse("application/custom"));
             });
+
+            mvcBuilder.AddXmlDataContractSerializerFormatters();
         }
 
         public void Configure(IApplicationBuilder app)
--- a/src/Microsoft.AspNet.Mvc.Core/Formatters/FormatFilter.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Formatters/FormatFilter.cs
@@ -122,12 +122,22 @@ namespace Microsoft.AspNet.Mvc.Formatters
             }
 
             var objectResult = context.Result as ObjectResult;
-            if (objectResult != null)
+            if (objectResult == null)
             {
-                var contentType = _options.FormatterMappings.GetMediaTypeMappingForFormat(format);
-                objectResult.ContentTypes.Clear();
-                objectResult.ContentTypes.Add(contentType);
+                return;
             }
+
+            // If the action sets a single content type, then its takes precedence over the user
+            // supplied content type based on format mapping.
+            if ((objectResult.ContentTypes != null && objectResult.ContentTypes.Count == 1) ||
+                !string.IsNullOrEmpty(context.HttpContext.Response.ContentType))
+            {
+                return;
+            }
+
+            var contentType = _options.FormatterMappings.GetMediaTypeMappingForFormat(format);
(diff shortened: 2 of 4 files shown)


samples/FormatFilterSample.Web/Startup.cs
```diff
@@ -23,6 +23,7 @@ public void ConfigureServices(IServiceCollection services)
                 var customFormatter = new CustomFormatter("application/custom");
                 options.OutputFormatters.Add(customFormatter);
                 options.OutputFormatters.RemoveType<StringOutputFormatter>();
+                options.OutputFormatters.Add(new XmlDataContractSerializerOutputFormatter());
```
rynowak: Use the extension method on `IMvcBuilder` rather than just adding the formatter

src/Microsoft.AspNet.Mvc.Core/Formatters/FormatFilter.cs
```diff
@@ -122,12 +122,22 @@ public void OnResultExecuting(ResultExecutingContext context)
...
+            if (objectResult == null)
             {
-                var contentType = _options.FormatterMappings.GetMediaTypeMappingForFormat(format);
-                objectResult.ContentTypes.Clear();
-                objectResult.ContentTypes.Add(contentType);
+                return;
             }
+
+            // If the developer is setting a content type explicitly, then its takes precedence over the user
+            // supplied content type based on format mapping.
```
rynowak: Grammar is wonky here.

rynowak: I would say `If the action set single a content type`.

What about `[Produces(...)]` with a single content type? Do you expect that to behave the same way as if the developer set `response.ContentType`?

samples/FormatFilterSample.Web/Controllers/FormatFilterController.cs
```diff
@@ -21,5 +22,21 @@ public Product ProducesMethod(int id)
...
+            result.ContentTypes.Add(new MediaTypeHeaderValue("text/xml"));
+            return result;
+        }
+
+        [Route("[controller]/[action]/{id}.{format}")]
+        public Product ExplicitContentTypeOnResponse(int id)
+        {
+            Response.ContentType = "text/xml";
+
+            return new Product() { SampleInt = id };
```
rynowak: Why would any of this be in the sample? Why would you want to do something like this?

kichalla: Good point. This should not be in a sample. This scenario is covered in unit test here so we should be good to remove it.

samples/FormatFilterSample.Web/Startup.cs
kichalla: :+1:

src/Microsoft.AspNet.Mvc.Core/Formatters/FormatFilter.cs
kichalla: > What about [Produces(...)] with a single content type? Do you expect that to behave the same way as if the developer set response.ContentType?

In this case the user would see a 404 Not Found if he is requesting a content type other than the supported content type on Produces

kichalla:
:up: :date:

samples/FormatFilterSample.Web/Controllers/FormatFilterController.cs
```diff
@@ -21,5 +22,13 @@ public Product ProducesMethod(int id)
         {
             return new Product() { SampleInt = id };
         }
+
+        [Route("[controller]/[action]/{id}.{format}")]
+        public IActionResult ExplicitContentTypeOnObjectResult(int id)
+        {
+            var result = new ObjectResult(new Product() { SampleInt = id });
+            result.ContentTypes.Add(new MediaTypeHeaderValue("text/xml"));
+            return result;
```
rynowak: Why is this still here?

This isn't code you'd ever write.

src/Microsoft.AspNet.Mvc.Core/Formatters/FormatFilter.cs
rynowak: This comment still hasn't been cleaned up

rynowak:
:shipit:  after fixing minor comments

PR #3824, opened 2015-12-29
Use DateTime.UtcNow instead of Environment.TickCount

pranavkm:


pranavkm:
cc @rynowak

rynowak:
:shipit:

src/Microsoft.AspNet.Mvc.Core/Infrastructure/MvcRouteHandler.cs
```diff
@@ -87,7 +87,7 @@ public VirtualPathData GetVirtualPath(VirtualPathContext context)
                 {
                     _logger.ExecutingAction(actionDescriptor);
 
-                    var startTime = Environment.TickCount;
+                    var startTime = DateTime.UtcNow;
```
PinpointTownes: Curious: what's the performance impact of this change?

`DateTime.UtcNow` is known to be rather slow, which might not be appropriate in such a critical code path.
Maybe it would be better to only call `DateTime.UtcNow` after a quick `logger.IsEnabled` check?

pranavkm: Yeah, seems like this change is pretty unnecessary. I'll go ahead and close the PR.

PinpointTownes: IMHO, avoid using `Environment.TickCount` or `DateTime.UtcNow` for profiling is a really good thing, as they don't offer enough precision for that (IIRC, `Environment.TickCount`'s resolution is between 10 and 15 ms).

`Stopwatch` is probably the best candidate for profiling.

pranavkm: I think the intent was to avoid allocating the `StopWatch` instance.

PinpointTownes: Ideally, a `Stopwatch` would only be allocated after an `IsEnabled(LogLevel.Information)` check, to avoid instantiating it abusively (optionally, I guess `Stopwatch` instances could be pooled). IMHO, profiling should be a pay-to-play scenario.

mj1856: You don't need to create an instance of `Stopwatch`.  Just do something like:

```csharp
long t1 = Stopwatch.GetTimestamp();
// ...
long t2 = Stopwatch.GetTimestamp();

long ticks = t2 - t1; // stopwatch ticks, not datetime ticks
int ms = (int) ((double) ticks / Stopwatch.Frequency) * 1000;
```

pranavkm: This was addressed as part of https://github.com/aspnet/Mvc/commit/c125b4e59badf551c85bbc8d6ae8cb133fe6de7c

mj1856: Great! Looks good.

PR #3828, opened 2015-12-29
Make RazorViewEngineOptions.FileProvider a list instead of a single item

### LINKED ISSUE #3806: Make RazorViewEngineOptions.FileProvider a list instead of a single item
See comment here: https://github.com/aspnet/Mvc/issues/3750#issuecomment-164474888

This has become a pretty common task for building anything modular. It would be nice if we just make `FileProvider` into a list so it was obvious that you can do this.

pranavkm:
Fixes #3806

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor/RazorViewEngineOptions.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/RazorViewEngineOptions.cs
@@ -14,12 +14,9 @@ namespace Microsoft.AspNet.Mvc.Razor
     /// </summary>
     public class RazorViewEngineOptions
     {
-        private IFileProvider _fileProvider;
-
         private CSharpParseOptions _parseOptions = new CSharpParseOptions(LanguageVersion.CSharp6);
-
-        private CSharpCompilationOptions _compilationOptions = new CSharpCompilationOptions(CodeAnalysis.OutputKind.DynamicallyLinkedLibrary);
-
+        private CSharpCompilationOptions _compilationOptions =
+            new CSharpCompilationOptions(CodeAnalysis.OutputKind.DynamicallyLinkedLibrary);
         private Action<RoslynCompilationContext> _compilationCallback = c => { };
 
         /// <summary>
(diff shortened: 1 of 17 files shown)


pranavkm:
cc @rynowak

src/Microsoft.AspNet.Mvc.Razor/RazorViewEngineOptions.cs
```diff
@@ -97,9 +88,31 @@ public CSharpCompilationOptions CompilationOptions
...
                 _compilationOptions = value;
             }
         }
 
+        /// <summary>
+        /// Gets the composite result of <see cref="FileProviders"/>.
+        /// </summary>
+        /// <returns>The <see cref="IFileProvider"/> instance if exactly one instance is registered, otherwise
+        /// a <see cref="CompositeFileProvider"/> that contains all registered instances.</returns>
+        public IFileProvider GetCompositeFileProvider()
```
rynowak: Should we just move this to a facade service? I don't think this is horrid, just kinda ugly to put on options (which the user will see).

pranavkm: Ok. It seemed a bit easy to get away with sticking it on options. I can move it to a service if you feel strongly about it

rynowak: I'm just asking the question. I think either one is a bit cumbersome. Let me ask @lodejard if he's around today.

davidfowl: This is a weird thing to expose

pranavkm: Fine. I'll make it a service.

davidfowl: Internal namespace?

pranavkm: That seems worse.

pranavkm:
:up: :date:

rynowak:
:shipit:

PR #3830, opened 2015-12-29
Rename DictionaryBasedValueProvider to RouteValueProvider

### LINKED ISSUE #3629: Simplify RouteValueProvider/RouteValueProviderFactory
We should rename https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/ModelBinding/DictionaryBasedValueProvider.cs to `RouteValueProvider`. Our other VPs have distinct class names.

We should also add a `protected set` `CultureInfo` property to all of our VPFs so that scenarios like [this one](https://github.com/aspnet/Routing/issues/237) are easier to address.

pranavkm:
Fixes #3629

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/ModelBinding/JQueryFormValueProvider.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ModelBinding/JQueryFormValueProvider.cs
@@ -9,7 +9,7 @@ using Microsoft.Extensions.Primitives;
 namespace Microsoft.AspNet.Mvc.ModelBinding
 {
     /// <summary>
-    /// An <see cref="IValueProvider"/> for form data stored in an <see cref="IDictionary{string, string[]}"/>.
+    /// An <see cref="IValueProvider"/> for jQuery formatted form data.
     /// </summary>
     public class JQueryFormValueProvider : BindingSourceValueProvider, IEnumerableValueProvider
     {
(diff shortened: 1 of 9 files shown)


pranavkm:
cc @rynowak \ @Eilon

src/Microsoft.AspNet.Mvc.Core/ModelBinding/JQueryFormValueProvider.cs
```diff
@@ -9,7 +9,7 @@
 namespace Microsoft.AspNet.Mvc.ModelBinding
 {
     /// <summary>
-    /// An <see cref="IValueProvider"/> for form data stored in an <see cref="IDictionary{string, string[]}"/>.
+    /// An <see cref="IValueProvider"/> for JQuery formatted form data.
```
Eilon: `jQuery` (that's how they case it)

pranavkm: https://msdn.microsoft.com/en-us/library/system.web.mvc.jqueryformvalueprovider(v=vs.118).aspx

> The JQuery Form Value provider is used to handle JQuery formatted data in request Forms.

Eilon: Believe it or not, MSDN is wrong :smile:

pranavkm:
Updated

Eilon:
:shipit:

rynowak:
:shipit:

PR #3838, opened 2015-12-31
Remove long form Dispose

### LINKED ISSUE #3834: Get rid of 'long form' dispose on MvcForm
MvcForm has the 'long form' dispose implementation that's only really needed if you use native resources and need finalization. We can remove this.

I did a search and there are 3-4 other places that include `GC.SuppressFinalize` include (gasp) Controller and ApiController. None of these have finalizers.

pranavkm:
Fixes #3834

### DIFF
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/MvcForm.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/MvcForm.cs
@@ -8,11 +8,18 @@ using Microsoft.Extensions.DependencyInjection;
 
 namespace Microsoft.AspNet.Mvc.Rendering
 {
+    /// <summary>
+    /// Represents an HTML form element in an MVC view.
+    /// </summary>
     public class MvcForm : IDisposable
     {
         private readonly ViewContext _viewContext;
         private bool _disposed;
 
+        /// <summary>
+        /// Initializes a new instance of <see cref="MvcForm"/>.
+        /// </summary>
+        /// <param name="viewContext">The <see cref="ViewContext"/>.</param>
         public MvcForm(ViewContext viewContext)
         {
             if (viewContext == null)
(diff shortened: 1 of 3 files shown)


pranavkm:
cc @rynowak

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/MvcForm.cs
```diff
@@ -8,11 +8,18 @@
 
 namespace Microsoft.AspNet.Mvc.Rendering
 {
+    /// <summary>
+    /// Represents an HTML form element in an MVC view.
```
dougbu: Ick "represents". This object represents nothing at all. It's an implementation detail used to ensure we generate one and only one `</form>` end tag.

pranavkm: I thought @DamianEdwards wanted us to "Represents ..".

rynowak: This **does** actually represent an HTML form element.

rynowak:
:shipit:

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/MvcForm.cs
dougbu: Huh?

DamianEdwards: I'm mostly concerned with us being consistent, which we aren't, at all. The "Represents..." wording has traditionally been used for .NET API docs.

pranavkm: Ok, I guess we've not been using that term, so I'll keep up with that tradition.

> An HTML form element in an MVC view.

PR #3839, opened 2015-12-31
Replace MediaTypeWithQualityHeaderValue with StringSegment in content negotiation and API Explorer

javiercn:
Replaces MediaTypeWithQualityHeaderValue with StringSegment in content negotiation and API Explorer for StringSegment to improve memory allocation though the stack and be more permissive about media types though our stack.


src/Microsoft.AspNet.Mvc.ApiExplorer/ApiResponseFormat.cs
```diff
@@ -19,6 +19,6 @@ public class ApiResponseFormat
         /// <summary>
         /// The media type of the response.
         /// </summary>
-        public MediaTypeHeaderValue MediaType { get; set; }
+        public StringSegment MediaType { get; set; }
```
rynowak: Lets make this just a string. We're not worried about allocations here.

src/Microsoft.AspNet.Mvc.Core/ApiExplorer/IApiResponseFormatMetadataProvider.cs
```diff
@@ -28,8 +29,8 @@ public interface IApiResponseFormatMetadataProvider
         /// The <see cref="Type"/> for which the supported content types are desired.
         /// </param>
         /// <returns>Content types which are supported by the <see cref="Formatters.IOutputFormatter"/>.</returns>
-        IReadOnlyList<MediaTypeHeaderValue> GetSupportedContentTypes(
-            MediaTypeHeaderValue contentType,
+        IReadOnlyList<StringSegment> GetSupportedContentTypes(
+            StringSegment contentType,
```
rynowak: Same for this. Let's make it `string`.

src/Microsoft.AspNet.Mvc.Core/ApiExplorer/IApiResponseMetadataProvider.cs
```diff
@@ -20,6 +20,6 @@ public interface IApiResponseMetadataProvider
         /// <summary>
         /// Configures a collection of allowed content types which can be produced by the action.
         /// </summary>
-        void SetContentTypes(IList<MediaTypeHeaderValue> contentTypes);
+        void SetContentTypes(IList<StringSegment> contentTypes);
```
rynowak: These can be `string`

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
```diff
@@ -0,0 +1,179 @@
...
+using System.Globalization;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNet.Mvc.Formatters
+{
+    public static class AcceptHeaderParser
+    {
+        private static readonly StringSegment QualityParameter = new StringSegment("q");
+
+        public static IList<MediaTypeSegmentWithQuality> ParseAcceptHeader(IList<string> acceptHeaders)
```
rynowak: Should the input to this be a `StringValues`?

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
```diff
@@ -0,0 +1,179 @@
...
+            }
+
+            if (parsedValues == null)
+            {
+                throw new ArgumentNullException(nameof(parsedValues));
+            }
+
+            // If a parser returns an empty list, it means there was no value, but that's valid (e.g. "Accept: ").
+            // The caller can ignore the value.
+            if (acceptHeaders == null)
```
rynowak: I don't see anything getting 'returned' here

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
```diff
@@ -0,0 +1,179 @@
...
+            {
+                return;
+            }
+
+            foreach (var value in acceptHeaders)
```
rynowak: This allocates, **and** you're already using an explicit `index`. Just make it a for loop.

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
```diff
@@ -0,0 +1,179 @@
...
+            {
+                int index = 0;
```
rynowak: var

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
```diff
@@ -0,0 +1,179 @@
...
+
+                while (!string.IsNullOrEmpty(value) && index < value.Length)
+                {
+                    MediaTypeSegmentWithQuality output;
+                    if (TryParseValue(value, ref index, out output))
+                    {
+                        // The entry may not contain an actual value, like " , "
```
rynowak: Is this actually the case? I think you're over-generalizing.

rynowak: What's an example of the input on the wire that would result in this?

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
```diff
@@ -0,0 +1,179 @@
...
+                        if (output.MediaType.HasValue)
+                        {
+                            parsedValues.Add(output);
+                        }
+                    }
+                    else
+                    {
+                        throw new FormatException(string.Format(CultureInfo.InvariantCulture, "Invalid values '{0}'.",
+                            value.Substring(index)));
```
rynowak: I'm not 100% convinced we want to throw on an invalid accept value... what's the behavior today?

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
```diff
@@ -0,0 +1,179 @@
...
+            var current = GetNextNonEmptyOrWhitespaceIndex(value, index, out separatorFound);
+
+            if (separatorFound && !true)
+            {
+                return false; // leading separators not allowed if we don't support multiple values.
+            }
+
+            if (current == value.Length)
+            {
+                if (true)
```
rynowak: Cool story bro

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
```diff
@@ -0,0 +1,179 @@
...
+            //  Accept: text/xml; q=1
+            //  Accept:
+            //  Accept: text/plain; q=0.2
+            if (string.IsNullOrEmpty(value) || (index == value.Length))
+            {
+                return true;
+            }
+
+            var separatorFound = false;
+            var current = GetNextNonEmptyOrWhitespaceIndex(value, index, out separatorFound);
```
rynowak: `EatWhitespace`?

rynowak: What is current, is it an index? Should we just pass index by ref?

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
```diff
@@ -0,0 +1,179 @@
...
+
+            if (length == 0)
+            {
+                return false;
+            }
+
+            current = current + length;
+            current = GetNextNonEmptyOrWhitespaceIndex(value, current, out separatorFound);
+
+            // If we support multiple values and we've not reached the end of the string, then we must have a separator.
```
rynowak: `If we support multiple values` I have no idea what that means

rynowak: Use examples

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
```diff
@@ -0,0 +1,179 @@
...
+            return true;
+        }
+
+        private static int GetNextNonEmptyOrWhitespaceIndex(
+            string input,
+            int startIndex,
+            out bool separatorFound)
+        {
+            Contract.Requires(input != null);
+            Contract.Requires(startIndex <= input.Length); // it's OK if index == value.Length.
```
rynowak: We don't use these in our code. Use `Debug.Assert`

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
```diff
@@ -0,0 +1,179 @@
...
+            separatorFound = false;
+            var current = startIndex + HttpTokenParsingRules.GetWhitespaceLength(input, startIndex);
+
+            if ((current == input.Length) || (input[current] != ','))
+            {
+                return current;
+            }
+
+            // If we have a separator, skip the separator and all following whitespaces. If we support
+            // empty values, continue until the current character is neither a separator nor a whitespace.
```
rynowak: `If we support empty values` - again, no idea what that means

rynowak: What are `separators`?

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
```diff
@@ -0,0 +1,179 @@
...
+            return current;
+        }
+
+        private static int GetMediaTypeWithQualityLength(
+            string input,
+            int start,
+            out MediaTypeSegmentWithQuality result)
+        {
+            result = default(MediaTypeSegmentWithQuality);
+            var enumerator = new MediaTypeParser(input, start, length: null).GetEnumerator();
```
rynowak: Is this a media type parser or a media type parameter parser?

rynowak: I'm not convinced that this is super valuable as an abstraction vs a `TryGetNextParameter` method

src/Microsoft.AspNet.Mvc.Core/Formatters/FormatFilter.cs
```diff
@@ -78,7 +79,7 @@ public void OnResourceExecuting(ResourceExecutingContext context)
 
             // Determine media types this action supports.
             var responseTypeFilters = context.Filters.OfType<IApiResponseMetadataProvider>();
-            var supportedMediaTypes = new List<MediaTypeHeaderValue>();
+            var supportedMediaTypes = new List<StringSegment>();
```
rynowak: these can be string too

src/Microsoft.AspNet.Mvc.Core/Formatters/HttpParseResult.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Mvc.Formatters
+{
+    internal enum HttpParseResult
```
rynowak: Nope. No internal

src/Microsoft.AspNet.Mvc.Core/Formatters/HttpTokenParsingRules.cs
```diff
@@ -0,0 +1,273 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics.Contracts;
+using System.Globalization;
+using System.Text;
+
+namespace Microsoft.AspNet.Mvc.Formatters
+{
+    internal static class HttpTokenParsingRules
```
rynowak: Was this copied from the headers code?

rynowak: Is this the minimal set of methods we call into?

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
```diff
@@ -0,0 +1,301 @@
...
+    {
+        /// <summary>
+        /// Determines if the <paramref name="subset" /> media type is a subset of the <paramref name="set" /> media type
+        /// without taking into account the quality parameter.
+        /// </summary>
+        /// <param name="set">The more general media type.</param>
+        /// <param name="subset">The more specific media type.</param>
+        /// <returns><code>true</code> if <paramref name="set" />is a more general media type than <paramref name="subset"/>
+        /// except for the quality parameter; otherwise <code>false</code>.</returns>
+        public static bool IsSubsetOf(StringSegment set, StringSegment subset)
```
rynowak: `IsCompatibleWith`?

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
```diff
@@ -0,0 +1,301 @@
...
+        }
+
+        /// <summary>
+        /// Determines if two media types are equal except for the quality parameter.
+        /// </summary>
+        /// <param name="left">The first media type.</param>
+        /// <param name="right">The second media type.</param>
+        /// <returns><code>true</code> if <paramref name="set" />is equal to <paramref name="subset"/>
+        /// except for quality parameter; otherwise <code>false</code>.</returns>
+        public static bool AreEqual(StringSegment left, StringSegment right)
```
rynowak: `AreEquivalent`

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComponent.cs
```diff
@@ -0,0 +1,138 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Internal;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNet.Mvc.Formatters
+{
+    /// <summary>
+    /// Represents a component within a media type.
```
rynowak: What's a `component`?

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComponent.cs
```diff
@@ -0,0 +1,138 @@
...
+
+            return obj is MediaTypeComponent && Equals((MediaTypeComponent)obj);
+        }
+
+        /// <inheritdoc />
+        public override int GetHashCode()
+        {
+            HashCodeCombiner hashCode = HashCodeCombiner.Start();
+            hashCode.Add(Name.Value);
+            hashCode.Add(Value.Value);
```
rynowak: These allocate.

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeParser.cs
```diff
@@ -0,0 +1,438 @@
...
+
+            /// <inheritdoc />
+            public MediaTypeComponent Current
+            {
+                get
+                {
+                    if (_parsingStatus == ParsingStatus.NotStarted)
+                    {
+                        throw new InvalidOperationException("Enumeration has not started. Call MoveNext.");
+                    }
```
rynowak: This is technically undefined behavior, so you don't need to throw here.

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeSegmentWithQuality.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    /// <summary>
+    /// Represents a media type with its associated quality.
+    /// </summary>
+    public struct MediaTypeSegmentWithQuality
+    {
+        /// <summary>
+        /// Gets or sets the media type of this <see cref="MediaTypeSegmentWithQuality"/>.
+        /// </summary>
+        public StringSegment MediaType { get; set; }
```
rynowak: No mutable structs. Not even once.

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
```diff
@@ -37,15 +39,41 @@ protected OutputFormatter()
         public IList<Encoding> SupportedEncodings { get; }
 
         /// <summary>
-        /// Gets the mutable collection of <see cref="MediaTypeHeaderValue"/> elements supported by
-        /// this <see cref="OutputFormatter"/>.
+        /// Gets the mutable collection of <see cref="StringSegment"/> elements that represent the media types
```
rynowak: Why are these stringsegments?

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
```diff
@@ -37,15 +39,41 @@ protected OutputFormatter()
...
+                    {
+                        cache.Add(mediaType, CreateMediaTypeWithEncoding(mediaType, Encoding.UTF8));
+                    }
+
+                    // Safe race condition, worst case scenario we initialize the field multiple times with dictionaries containing
+                    // the same values.
+                    _outputMediaTypeCache = cache;
+                }
+
+                return _outputMediaTypeCache;
```
rynowak: Why is this part of this PR?

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
```diff
@@ -246,6 +275,83 @@ public virtual void WriteResponseHeaders(OutputFormatterWriteContext context)
...
+        /// <summary>
+        /// Adds or replaces the charset parameter in a given <paramref name="mediaType"/> with the
+        /// given <paramref name="encoding"/>.
+        /// </summary>
+        /// <param name="mediaType">The <see cref="StringSegment"/> with the media type.</param>
+        /// <param name="encoding">
+        /// The <see cref="Encoding"/> to add or replace in the <paramref name="mediaType"/>.
+        /// </param>
+        /// <returns>The mediaType with the given encoding.</returns>
+        protected StringSegment AddOrOverrideEncoding(StringSegment mediaType, Encoding encoding)
```
rynowak: `GetMediaTypeWithCharset`?

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
```diff
@@ -246,6 +275,83 @@ public virtual void WriteResponseHeaders(OutputFormatterWriteContext context)
...
+        {
+            var charset = GetCharset(mediaType);
+            if (charset.Equals("utf-8", StringComparison.OrdinalIgnoreCase))
+            {
+                // This is an optimization for utf-8 that prevents the Substring caused by
+                // charset.Value
+                return Encoding.UTF8;
+            }
+
+            return charset.HasValue ? Encoding.GetEncoding(charset.Value) : null;
```
rynowak: :heart_eyes_cat:

src/Microsoft.AspNet.Mvc.Core/Infrastructure/ObjectResultExecutor.cs
```diff
@@ -185,14 +187,14 @@ public virtual Task ExecuteAsync(ActionContext context, ObjectResult result)
...
 
+            var mediaTypes = GetMediaTypes(contentTypes, request);
             IOutputFormatter selectedFormatter = null;
-            if (contentTypes == null || contentTypes.Count == 0)
+            if (contentTypes.Count == 0)
             {
                 // Check if we have enough information to do content-negotiation, otherwise get the first formatter
                 // which can write the type. Let the formatter choose the Content-Type.
-                if (acceptValues == null || acceptValues.Count == 0)
+                if (!(mediaTypes.Count() > 0))
```
rynowak: ????

rynowak: Don't use `.Count()`

src/Microsoft.AspNet.Mvc.Core/Infrastructure/ObjectResultExecutor.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
```
rynowak: No. Absolutely not.

src/Microsoft.AspNet.Mvc.Core/Infrastructure/ObjectResultExecutor.cs
```diff
@@ -270,6 +252,69 @@ public virtual Task ExecuteAsync(ActionContext context, ObjectResult result)
             return selectedFormatter;
         }
 
+        private IList<MediaTypeSegmentWithQuality> GetMediaTypes(
+            IList<StringSegment> contentTypes,
+            HttpRequest request)
+        {
+            // Will fix the list with object pooling
```
rynowak: No TODOs in code

src/Microsoft.AspNet.Mvc.Core/Infrastructure/ObjectResultExecutor.cs
```diff
@@ -270,6 +252,69 @@ public virtual Task ExecuteAsync(ActionContext context, ObjectResult result)
...
+            for (int i = 0; i < acceptableMediaTypes.Count; i++)
+            {
+                if (MediaTypeComparisons.IsSubsetOf(mediaType, acceptableMediaTypes[i]))
+                {
+                    return true;
+                }
+            }
+
+            return false;
+        }
```
rynowak: This is way less clear with all these little helper methods that have 2 statements

src/Microsoft.AspNet.Mvc.Core/Infrastructure/ObjectResultExecutor.cs
```diff
@@ -297,7 +342,7 @@ public virtual Task ExecuteAsync(ActionContext context, ObjectResult result)
 
             foreach (var formatter in formatters)
             {
-                formatterContext.ContentType = null;
+                formatterContext.ContentType = new StringSegment(); // Equivalent to null MediaType
```
rynowak: So?

src/Microsoft.AspNet.Mvc.Core/Infrastructure/ObjectResultExecutor.cs
```diff
@@ -321,10 +366,10 @@ public virtual Task ExecuteAsync(ActionContext context, ObjectResult result)
         /// <returns>
         /// The selected <see cref="IOutputFormatter"/> or <c>null</c> if no formatter can write the response.
         /// </returns>
-        protected virtual IOutputFormatter SelectFormatterUsingSortedAcceptHeaders(
+        protected virtual IOutputFormatter SelectFormatterUsingAcceptHeader(
```
rynowak: Are they not sorted anymore?

src/Microsoft.AspNet.Mvc.Core/Infrastructure/ObjectResultExecutor.cs
```diff
@@ -321,10 +366,10 @@ public virtual Task ExecuteAsync(ActionContext context, ObjectResult result)
...
             OutputFormatterWriteContext formatterContext,
             IList<IOutputFormatter> formatters,
-            IList<MediaTypeHeaderValue> sortedAcceptHeaders)
+            IList<MediaTypeSegmentWithQuality> acceptHeader)
```
rynowak: why the change?

src/Microsoft.AspNet.Mvc.Core/Infrastructure/ObjectResultExecutor.cs
```diff
@@ -336,16 +381,16 @@ public virtual Task ExecuteAsync(ActionContext context, ObjectResult result)
...
 
-            if (sortedAcceptHeaders == null)
+            if (acceptHeader == null)
             {
-                throw new ArgumentNullException(nameof(sortedAcceptHeaders));
+                throw new ArgumentNullException(nameof(acceptHeader));
             }
 
-            foreach (var contentType in sortedAcceptHeaders)
+            foreach (var mediaTpe in acceptHeader)
```
rynowak: typo

src/Microsoft.AspNet.Mvc.Core/Logging/ObjectResultExecutorLoggerExtensions.cs
```diff
@@ -40,7 +41,7 @@ static ObjectResultExecutorLoggerExtensions()
                 LogLevel.Debug,
                 4,
                 "No information found on request to perform content negotiation.");
-            _noFormatterFromNegotiation = LoggerMessage.Define<IEnumerable<MediaTypeHeaderValue>>(
+            _noFormatterFromNegotiation = LoggerMessage.Define<IEnumerable<MediaTypeSegmentWithQuality>>(
```
rynowak: `MediaTypeSegmentWithQuality` does this `ToString` nicely?

src/Microsoft.AspNet.Mvc.Core/MediaTypeCollection.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using System.Collections.ObjectModel;
+using Microsoft.Extensions.Primitives;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNet.Mvc
```
rynowak: Nope.

rynowak: `.Formatters`

src/Microsoft.AspNet.Mvc.Core/ProducesAttribute.cs
```diff
@@ -74,12 +75,12 @@ public override void OnResultExecuting(ResultExecutingContext context)
...
+        private MediaTypeCollection GetContentTypes(string firstArg, string[] args)
         {
             var completeArgs = new List<string>();
             completeArgs.Add(firstArg);
             completeArgs.AddRange(args);
-            var contentTypes = new List<MediaTypeHeaderValue>();
+            var contentTypes = new MediaTypeCollection();
             foreach (var arg in completeArgs)
             {
                 var contentType = MediaTypeHeaderValue.Parse(arg);
```
rynowak: Should we be using a new parsing function for this?

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
```diff
@@ -0,0 +1,179 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.Contracts;
+using System.Globalization;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNet.Mvc.Formatters
+{
+    public static class AcceptHeaderParser
```
rynowak: There's not really any code sharing between this and the media type parser... should there be?

javiercn: AcceptHeaderParser uses MediaTypeParser internally, but mainly accept header parser deals with parsing a comma separated list of values.

src/Microsoft.AspNet.Mvc.Core/ProducesAttribute.cs
javiercn: This is actually 
var contentType = new StringSegment(arg);
if (MediaTypeComparisons.MatchesAllSubtypes(contentType)||
    MediaTypeComparisons.MatchesAllTypes(contentType))

src/Microsoft.AspNet.Mvc.Core/Logging/ObjectResultExecutorLoggerExtensions.cs
javiercn: Yes it does, it will print the media type.

src/Microsoft.AspNet.Mvc.Core/Infrastructure/ObjectResultExecutor.cs
javiercn: This is due to the change of the parsing code, moving from MediaTypeSegmentWithQuality to StringSegment (ignoring the quality after we already sorted it) implies creating a new list, which I don't want to do.

src/Microsoft.AspNet.Mvc.Core/Infrastructure/ObjectResultExecutor.cs
javiercn: This is a mistake, will revert the name

src/Microsoft.AspNet.Mvc.Core/Infrastructure/ObjectResultExecutor.cs
javiercn: I'll remove the comment

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
javiercn: I like more AddOrOverrideEncoding/Charset as it clearly implies what the method does, but will change it for GetMediaTypeWithCharset if you feel strongly about it.

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
javiercn: I can break this into a separate PR

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
javiercn: Will change <see ... StringSegment for media type.

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeSegmentWithQuality.cs
javiercn: Fair enough

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComponent.cs
javiercn: Yes they do, but we don't use MediaTypeComponent in a way that involves getting it's hash code, and if in the future we do, we can easily change the implementation.

For what is worth, I'm happy just doing return 1 which satisfies the GetHashCode contract and  causes a degenerate list in the worst case scenario. The only reason GetHashCode is overriden is because it needs to be done for equality

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComponent.cs
javiercn: I can be more verbose and explain that the components are type, subtype and the parameters of the media type. Is that what you want?

src/Microsoft.AspNet.Mvc.Core/Formatters/HttpTokenParsingRules.cs
javiercn: Yes and yes, I just renamed the class as the methods left are more specific, but can undo the renaming

src/Microsoft.AspNet.Mvc.Core/Formatters/HttpParseResult.cs
javiercn: I'm keeping this internal as it's only used in HttpTokenParsingRules, we can change later if we want to

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
javiercn: This is verbatim from the parsing code in HttpAbstractions, I'm not changing any of this

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
javiercn: Will change, I removed those in most of the code, but i might have miss some

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
javiercn: This is verbatim from the parsing code in HttpAbstractions, let's not change the scenarios for parsing the accept header in this PR

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
javiercn: This is from the parsing code in HttpAbstractions, we can do further refactoring if we want later, I personally despise passing a parameter by ref unless absolutely necessary. Keeping this as a pure function is much better and clearer from my point of view.

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
javiercn: Missed some of this, this used to be SupportsMultipleValues which in our case is always true.

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
javiercn: This is the behavior today

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
javiercn: I haven't changed any of this from HttpAbstractions

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
javiercn: I don't think so, StringValues is an implementation detail, it's better if the parser works with the minimal abstraction it requires to work with, there's no foreseeable advantage of passing a more specific type.

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
javiercn: removed the comment. It was leftovers

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
javiercn: will change

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
javiercn: No, current represents the current index while we are advancing and we don't want to pass index by ref as it's already by ref in that method and we don't want to change the index until we know for sure what situation we are in

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
javiercn: This is not a media type parameter parser, in order to calculate the length of the media type, we iterate over all it's components using the enumerator, at the end, the enumerator contains the offset within the string where the media type ends

src/Microsoft.AspNet.Mvc.Core/Infrastructure/ObjectResultExecutor.cs
javiercn: removed

src/Microsoft.AspNet.Mvc.Core/Infrastructure/ObjectResultExecutor.cs
javiercn: Changed to .Count

javiercn:
Merged individual PRs

PR #3844, opened 2015-12-31
Change PhysicalFileResult and VirtualFileResult to use SendFileAsync

davidfowl:
- SendFileAsync does the proper fallback to stream copy when the feature
isn't available. Take advantage of it in MVC. There are plans to use the
buffer pool as part of the stream copy so MVC will get that benefit for
free.
- Left CopyToAsync in SendFileAsync when the file doesn't have a
physical path

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/PhysicalFileResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/PhysicalFileResult.cs
@@ -3,10 +3,8 @@
 
 using System;
 using System.IO;
-using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Http;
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Mvc.Core/PhysicalFileResult.cs
```diff
@@ -18,7 +16,6 @@ namespace Microsoft.AspNet.Mvc
     /// </summary>
     public class PhysicalFileResult : FileResult
     {
-        private const int DefaultBufferSize = 0x1000;
```
rynowak: :clap:

rynowak:
:shipit:

PR #3845, opened 2016-01-01
Optimize MvcRouteHandler

rynowak:
- Check MVC services once at startup
- Make action selector sync

We've never really had a scenario for the action selector being async, it
just ended up that way. None of our extensibility here lets you do
anything async without replacing it wholesale, which we don't
recommend.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/Infrastructure/IActionSelector.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Infrastructure/IActionSelector.cs
@@ -1,7 +1,6 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System.Threading.Tasks;
 using Microsoft.AspNet.Mvc.Abstractions;
 using Microsoft.AspNet.Routing;
 
@@ -9,6 +8,6 @@ namespace Microsoft.AspNet.Mvc.Infrastructure
 {
     public interface IActionSelector
     {
-        Task<ActionDescriptor> SelectAsync(RouteContext context);
+        ActionDescriptor SelectAsync(RouteContext context);
     }
 }
--- a/src/Microsoft.AspNet.Mvc.Core/Infrastructure/IActionSelector.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Infrastructure/IActionSelector.cs
@@ -8,6 +8,6 @@ namespace Microsoft.AspNet.Mvc.Infrastructure
 {
     public interface IActionSelector
     {
-        ActionDescriptor SelectAsync(RouteContext context);
+        ActionDescriptor Select(RouteContext context);
     }
 }
--- a/src/Microsoft.AspNet.Mvc.Core/Infrastructure/MvcRouteHandler.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Infrastructure/MvcRouteHandler.cs
@@ -38,25 +38,20 @@ namespace Microsoft.AspNet.Mvc.Infrastructure
             return null;
         }
 
-        public async Task RouteAsync(RouteContext context)
+        public Task RouteAsync(RouteContext context)
         {
             if (context == null)
             {
                 throw new ArgumentNullException(nameof(context));
             }
 
-            var services = context.HttpContext.RequestServices;
-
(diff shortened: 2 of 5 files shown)


rynowak:
Here's the data that I'm working with:

image

We can get rid of a state machine here by making MvcRouteHandler just be synchronous (and `Task`-returning.

src/Microsoft.AspNet.Mvc.Core/Infrastructure/IActionSelector.cs
```diff
@@ -9,6 +8,6 @@ namespace Microsoft.AspNet.Mvc.Infrastructure
 {
     public interface IActionSelector
     {
-        Task<ActionDescriptor> SelectAsync(RouteContext context);
+        ActionDescriptor SelectAsync(RouteContext context);
```
Tragetaschen: Shouldn't this lose the `Async` suffix now as well?

Eilon: :+1:

Eilon: And same for all the other methods that are now sync.

src/Microsoft.AspNet.Mvc.Core/Infrastructure/MvcRouteHandler.cs
```diff
@@ -68,12 +63,13 @@ public VirtualPathData GetVirtualPath(VirtualPathContext context)
                         context.RouteData.Values.Add(kvp.Key, kvp.Value);
                     }
                 }
-            }
 
-            // Removing RouteGroup from RouteValues to simulate the result of conventional routing
-            context.RouteData.Values.Remove(TreeRouter.RouteGroupKey);
+                // Removing RouteGroup from RouteValues to simulate the result of conventional routing
+                context.RouteData.Values.Remove(TreeRouter.RouteGroupKey);
```
Eilon: Any particular reason this moved to inside the `if()`? Is it not needed in other cases?

rynowak: Yes, it's an attribute-routing-only kinda thing.

src/Microsoft.AspNet.Mvc.Core/Infrastructure/IActionSelector.cs
rynowak: YEAH

src/Microsoft.AspNet.Mvc.Core/Infrastructure/IActionSelector.cs
```diff
@@ -9,6 +8,6 @@ namespace Microsoft.AspNet.Mvc.Infrastructure
 {
     public interface IActionSelector
     {
-        Task<ActionDescriptor> SelectAsync(RouteContext context);
+        ActionDescriptor Select(RouteContext context);
```
pranavkm: docs while you're at it?

src/Microsoft.AspNet.Mvc.Core/Infrastructure/MvcRouteHandler.cs
```diff
@@ -121,15 +117,22 @@ private void EnsureServices(HttpContext context)
                 return;
             }
 
+            var services = context.RequestServices;
+
+            // Verify if AddMvc was done before calling UseMvc
+            // We use the MvcMarkerService to make sure if all the services were added.
+            MvcServicesHelper.ThrowIfMvcNotRegistered(services);
+
+
```
pranavkm: extra space.

test/Microsoft.AspNet.Mvc.Core.Test/Infrastructure/DefaultActionSelectorTests.cs
```diff
@@ -375,9 +374,9 @@ public class DefaultActionSelectorTests
             context.RouteData.Values.Add("action", "Buy");
 
             // Act
-            var ex = await Assert.ThrowsAsync<AmbiguousActionException>(async () =>
+            var ex = Assert.Throws<AmbiguousActionException>(() =>
```
pranavkm: ` var ex = Assert.Throws<AmbiguousActionException>(() => selector.Select(context));`

pranavkm:
:shipit:

PR #3853, opened 2016-01-03
Mock `HttpContext`, not `DefaultHttpContext`

dougbu:
- build break
- recent HttpAbstractions changes made `DefaultHttpContext` harder to mock (would need `CallBase=true`)


dougbu:
FYI only

dougbu:
568b061

PR #3855, opened 2016-01-04
Change `ViewLocalizerTest` to use `ExecutingFilePath` to determine base name

dougbu:
- follow-up to #2767
- resources are now associated with specific page, not the `IView` when the `ViewContext` was created

nit: remove unused `Resources` class


dougbu:
Closing with prejudice. @DamianEdwards already added a test covering the mainline `ViewLocalizer` path determination.

PR #3857, opened 2016-01-04
Add convenience overloads for AddControllersAsServices

khellang:
What the title says :dancers: 

This adds one overload for adding assemblies, taking a `TAssembly`, which will add a single assembly based on the type argument.

It also adds a `params` overload for both assemblies and types :sparkles:

Is that something you'd be interested in?

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/DependencyInjection/MvcCoreMvcBuilderExtensions.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/DependencyInjection/MvcCoreMvcBuilderExtensions.cs
@@ -74,11 +74,6 @@ namespace Microsoft.Extensions.DependencyInjection
                 throw new ArgumentNullException(nameof(builder));
             }
 
-            if (controllerTypes == null)
-            {
-                throw new ArgumentNullException(nameof(controllerTypes));
-            }
-
             ControllersAsServices.AddControllersAsServices(builder.Services, controllerTypes);
             return builder;
         }
@@ -97,11 +92,6 @@ namespace Microsoft.Extensions.DependencyInjection
             if (builder == null)
             {
                 throw new ArgumentNullException(nameof(builder));
-            }
-
-            if (controllerAssemblies == null)
-            {
-                throw new ArgumentNullException(nameof(controllerAssemblies));
             }
 
             ControllersAsServices.AddControllersAsServices(builder.Services, controllerAssemblies);
--- a/src/Microsoft.AspNet.Mvc.Core/DependencyInjection/MvcCoreMvcBuilderExtensions.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/DependencyInjection/MvcCoreMvcBuilderExtensions.cs
@@ -7,6 +7,7 @@ using System.Reflection;
 using Microsoft.AspNet.Mvc;
 using Microsoft.AspNet.Mvc.Formatters;
 using Microsoft.AspNet.Mvc.Internal;
+using System.Linq;
 
 namespace Microsoft.Extensions.DependencyInjection
 {
@@ -66,6 +67,20 @@ namespace Microsoft.Extensions.DependencyInjection
         /// <returns>The <see cref="IMvcBuilder"/>.</returns>
         public static IMvcBuilder AddControllersAsServices(
            this IMvcBuilder builder,
+           params Type[] controllerTypes)
+        {
+            return builder.AddControllersAsServices(controllerTypes.AsEnumerable());
+        }
+
+        /// <summary>
+        /// Register the specified <paramref name="controllerTypes"/> as services and as a source for controller
+        /// discovery.
+        /// </summary>
+        /// <param name="builder">The <see cref="IMvcBuilder"/>.</param>
+        /// <param name="controllerTypes">A sequence of controller <see cref="Type"/>s to register.</param>
+        /// <returns>The <see cref="IMvcBuilder"/>.</returns>
+        public static IMvcBuilder AddControllersAsServices(
+           this IMvcBuilder builder,
            IEnumerable<Type> controllerTypes)
         {
             if (builder == null)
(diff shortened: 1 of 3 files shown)


khellang:
I'm not a fan of the copy-pasta between the `IMvcBuilder` and `IMvcCoreBuilder` extensions. Why does this separation exist? Can't `IMvcBuilder` implement `IMvcCoreBuilder`?

rynowak:
Yes, we're definitely interested is in this. I'll let others chime in with opinions but seeing code like
```
mvc.AddControllersAsServices<HomeController>()
```

throws me. Do it mean add **HomeController** as a service? That's why I don't like it. (and yes, I've seen this before and I am familiar with the pattern).


> I'm not a fan of the copy-pasta between the IMvcBuilder and IMvcCoreBuilder extensions. Why does this separation exist? Can't IMvcBuilder implement IMvcCoreBuilder?

Such things are necessary. `IMvcBuilder` isn't a superset, and you might have implementations that differ in what features are available or required across both experiences.

khellang:
> I'll let others chime in with opinions but seeing code like [...] throws me.

Yes, me too, actually. I just stuck with the name for now. Do you want to see it removed, or just renamed? I've see the `FromAssemblyOf<T>` naming around often. Would that make sense?

rynowak:
If we can come up with a good name, it would be nice to have, otherwise remove it. I'll ask around.

```
.AddControllersAsServicesFromAssemblyOf<HomeController>() 
```

Seems just as verbose as
```
.AddControllersAsServices(typeof(HomeController).Assembly);
```

khellang:
> `.AddControllersAsServicesFromAssemblyOf<HomeController>()` seems just as verbose as `.AddControllersAsServices(typeof(HomeController).Assembly);`

Yes. Yes it does. But the former will be taken care of by IntelliSense in many cases, and the latter will also need `.GetTypeInfo()` before `.Assembly` :smile:

danroth27:
The behavior of specifying a single example type in order to specify an entire assembly feels weird and I think that's why there really isn't a good name for this method. I don't think we should have a method with this behavior - it will just cause confusion.

khellang:
@danroth27 It's a really common pattern across all kinds of libs, so it can't be *that* weird :wink: I even think Entity Framework have (had?) the same kind of methods to "scan" for types in assemblies. Anyway, just say the word and it's gone :sparkles: 

BTW, can any of you shed some light on why these methods exist in the first place? I just stumbled over it when debugging https://github.com/structuremap/structuremap.dnx/issues/6 and found it weird that they didn't exist.

Not having controllers as "services" by default sounds like a really weird default IMO. Especially when you involve a 3rd party container (which is encouraged, right?).

danroth27:
@khellang Can you point me to some examples where this pattern is used? Maybe we can borrow an existing naming convention.

khellang:
- StructureMap has `AssemblyContainingType<T>` (see [source](https://github.com/structuremap/structuremap/blob/da0a94750ba4a1d0f91520e44d2caeec24bff7f3/src/StructureMap.Shared/Graph/AssemblyScanner.cs#L73-L76))
 - Fluent NHibernate has just `AssemblyOf<T>` (see [source](https://github.com/jagregory/fluent-nhibernate/blob/6302b200c349ca5c5c61ad5f829fbf14b9ba123b/src/FluentNHibernate/Automapping/AutoMap.cs#L17-L20))
 - EF Contrib (I thought this was actually pulled into EF proper, but I couldn't find it) has `AddFromAssemblyOf<T>` (see [source](https://ef6contrib.codeplex.com/SourceControl/latest#src/EntityFramework.Contrib/ModelConfiguration/Configuration/ConfigurationRegistrarExtensions.cs))
 - Ninject has `FromAssemblyContaining<T>` (see [source](https://github.com/ninject/Ninject.Extensions.Conventions/blob/d0f2acc2d0ab17e4fdda9698ccc407413a17b56e/src/Ninject.Extensions.Conventions/BindingBuilder/ConventionSyntax.From.cs#L73-L76))
 - FluentSitemap has `FromAsemblyOf<T>` (see [source](https://github.com/khalidabuhakmeh/FluentSitemap/blob/a18308af0b9c93f7f64ce3c0d6823bfd8029ad46/FluentSitemap.Core/SitemapConfigurator.cs#L295-L299))
 - Scrutor has `FromAssemblyOf<T>` (see [source](https://github.com/khellang/Scrutor/blob/2545b16747b81f42c889ec57415e3acc8018f9c0/src/Scrutor/LifetimeSelector.cs#L22-L25))

These are just some that I knew from the top of my head. I'm sure there are others.

danroth27:
OK, despite it's verbosity, `AddControllersAsServicesFromAssemblyOf<HomeController>()` does seem like the logical choice. @rynowak chime in if you and @pranavkm come up with something better.

rynowak:
I haven't come up with anything better. I'm equally happy with `AddControllersAsServicesFromAssemblyOf` or just leaving it out.

pranavkm:
An alternative might be to have a builder for AddControllersAsServices

```C#
IControllerAsAssemblyBuilder AddControllersAsServices(params Type[] controllerTypes)

IControllerAsAssemblyBuilder
{
   AddType();
}
```
We could bake the `FromAssemblyContaining(Type type)` method in or have DI systems provide it as an extension method.

rynowak:
@Eilon  - any thoughts on https://github.com/aspnet/Mvc/pull/3857#issuecomment-169057146 and resulting discussion?

I think that's pretty much the one thing left to discuss here. 

My opinion: leave it out :laughing:

Eilon:
@rynowak yeah leave out for now. Too awkward.

rynowak:
@khellang - any argument from you if I remove the `TAssembly` overloads and merge this bad boy?

khellang:
@rynowak Nope, but that single method was actually the method I was after, and the reason I sent this. Too bad naming is so hard :stuck_out_tongue_closed_eyes: Want me to strip it quickly?

rynowak:
sure

khellang:
Done :+1.

rynowak:
6737a1f243a35e7daed119c40ed6b026382d8fae

Issue #3361, opened 2015-10-20
Model binding order


brockallen:
In MVC 5 when multiple parameters with the same names are submitted the order of precedence was: 1) form, 2) route, and then 3) query string. In MVC 6 (beta8) it seems the order is: 1) route, 2) query, and then 3) form. 

Is this difference conscious? If so, can you provide some insight into the rationale for the change? If not, should it be the same as MVC5?

pranavkm:
Pretty certain the current behavior is entirely accidental (https://github.com/aspnet/Mvc/commit/574ce6de). It was appended without giving any thought to the precedence.

brockallen:
Perhaps some thought should go into it?

pranavkm:
:+1:

Eilon:
Yeah this still looks wrong:
https://github.com/aspnet/Mvc/blob/aab051a20f7a9c6ffe68d2219a9f736d3bc15898/src/Microsoft.AspNet.Mvc.Core/Internal/MvcCoreMvcOptionsSetup.cs#L53-L57

```c#
// Set up ValueProviders 
options.ValueProviderFactories.Add(new RouteValueProviderFactory()); 
options.ValueProviderFactories.Add(new QueryStringValueProviderFactory()); 
options.ValueProviderFactories.Add(new FormValueProviderFactory()); 
options.ValueProviderFactories.Add(new JQueryFormValueProviderFactory()); 
```

@rynowak any idea why the order changed? Just accident?

FYI the MVC 5 order can be seen here: http://aspnetwebstack.codeplex.com/SourceControl/latest#src/System.Web.Mvc/ValueProviderFactories.cs

```c#
            new ChildActionValueProviderFactory(),
            new FormValueProviderFactory(),
            new JsonValueProviderFactory(),
            new RouteDataValueProviderFactory(),
            new QueryStringValueProviderFactory(),
            new HttpFileCollectionValueProviderFactory(),
            new JQueryFormValueProviderFactory()
```

rynowak:
No idea. I feel like this is one of those things that we've reviewed for compatibility multiple times without  issue. Not sure what the deal is.

brockallen:
One does sometimes code to the order, and it's certainly a [breaking] change in behavior from prior MVCs.

Eilon:
Yeah the specific order is super important, and I can't think of any reason to deviate from MVC 5's order.

PR #3858, opened 2016-01-04
Changed the value provider preference to be backwards compatible

ajaybhargavb:
Issue - #3361 

@pranavkm

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/Internal/MvcCoreMvcOptionsSetup.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Internal/MvcCoreMvcOptionsSetup.cs
@@ -51,9 +51,9 @@ namespace Microsoft.AspNet.Mvc.Internal
             options.OutputFormatters.Add(new StreamOutputFormatter());
 
             // Set up ValueProviders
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Mvc.Core/Internal/MvcCoreMvcOptionsSetup.cs
```diff
@@ -51,9 +51,9 @@ public static void ConfigureMvc(MvcOptions options, IHttpRequestStreamReaderFact
             options.OutputFormatters.Add(new StreamOutputFormatter());
 
             // Set up ValueProviders
+            options.ValueProviderFactories.Add(new FormValueProviderFactory());
```
pranavkm: Could we add a test for this since the ordering is pretty important?

pranavkm: Looks like there is one - https://github.com/aspnet/Mvc/blob/dev/test/Microsoft.AspNet.Mvc.Test/MvcOptionsSetupTest.cs#L71 - shouldn't it fail?

ajaybhargavb:
Updated.

pranavkm:
:shipit:

PR #3862, opened 2016-01-05
[Issue3683 Part 1] Public interface changes to support allocation free content negotiation

javiercn:
Public interface changes with stub implementation.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.ApiExplorer/ApiResponseFormat.cs
+++ b/src/Microsoft.AspNet.Mvc.ApiExplorer/ApiResponseFormat.cs
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using Microsoft.AspNet.Mvc.Formatters;
-using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNet.Mvc.ApiExplorer
 {
@@ -19,6 +18,6 @@ namespace Microsoft.AspNet.Mvc.ApiExplorer
         /// <summary>
         /// The media type of the response.
         /// </summary>
-        public MediaTypeHeaderValue MediaType { get; set; }
+        public string MediaType { get; set; }
     }
 }
--- a/src/Microsoft.AspNet.Mvc.ApiExplorer/DefaultApiDescriptionProvider.cs
+++ b/src/Microsoft.AspNet.Mvc.ApiExplorer/DefaultApiDescriptionProvider.cs
@@ -15,7 +15,6 @@ using Microsoft.AspNet.Routing;
 using Microsoft.AspNet.Routing.Template;
 using Microsoft.Extensions.Internal;
 using Microsoft.Extensions.Options;
-using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNet.Mvc.ApiExplorer
 {
@@ -311,7 +310,7 @@ namespace Microsoft.AspNet.Mvc.ApiExplorer
 
             // Walk through all 'filter' attributes in order, and allow each one to see or override
             // the results of the previous ones. This is similar to the execution path for content-negotiation.
-            var contentTypes = new List<MediaTypeHeaderValue>();
+            var contentTypes = new MediaTypeCollection();
             if (responseMetadataAttributes != null)
             {
                 foreach (var metadataAttribute in responseMetadataAttributes)
@@ -322,7 +321,7 @@ namespace Microsoft.AspNet.Mvc.ApiExplorer
 
             if (contentTypes.Count == 0)
             {
-                contentTypes.Add(null);
+                contentTypes.Add((string)null);
             }
 
             foreach (var contentType in contentTypes)
--- a/src/Microsoft.AspNet.Mvc.Core/ApiExplorer/IApiResponseFormatMetadataProvider.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ApiExplorer/IApiResponseFormatMetadataProvider.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNet.Mvc.ApiExplorer
 {
@@ -28,8 +27,8 @@ namespace Microsoft.AspNet.Mvc.ApiExplorer
         /// The <see cref="Type"/> for which the supported content types are desired.
         /// </param>
         /// <returns>Content types which are supported by the <see cref="Formatters.IOutputFormatter"/>.</returns>
-        IReadOnlyList<MediaTypeHeaderValue> GetSupportedContentTypes(
-            MediaTypeHeaderValue contentType,
+        IReadOnlyList<string> GetSupportedContentTypes(
+            string contentType,
             Type objectType);
     }
 }
--- a/src/Microsoft.AspNet.Mvc.Core/ConsumesAttribute.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ConsumesAttribute.cs
@@ -8,6 +8,8 @@ using Microsoft.AspNet.Mvc.Abstractions;
 using Microsoft.AspNet.Mvc.ActionConstraints;
 using Microsoft.AspNet.Mvc.Core;
 using Microsoft.AspNet.Mvc.Filters;
+using Microsoft.AspNet.Mvc.Formatters;
+using Microsoft.Extensions.Primitives;
 using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNet.Mvc
@@ -30,6 +32,15 @@ namespace Microsoft.AspNet.Mvc
                 throw new ArgumentNullException(nameof(contentType));
             }
 
+            // We want to ensure that the given provided content types are valid values, so
+            // we validate them using the semantics of MediaTypeHeaderValue.
+            MediaTypeHeaderValue.Parse(contentType);
+
+            for (var i = 0; i < otherContentTypes.Length; i++)
+            {
+                MediaTypeHeaderValue.Parse(otherContentTypes[i]);
+            }
+
             ContentTypes = GetContentTypes(contentType, otherContentTypes);
         }
 
@@ -39,7 +50,7 @@ namespace Microsoft.AspNet.Mvc
         int IActionConstraint.Order { get; } = ConsumesActionConstraintOrder;
 
         /// <inheritdoc />
-        public IList<MediaTypeHeaderValue> ContentTypes { get; set; }
+        public MediaTypeCollection ContentTypes { get; set; }
 
         /// <inheritdoc />
         public void OnResourceExecuting(ResourceExecutingContext context)
@@ -53,13 +64,12 @@ namespace Microsoft.AspNet.Mvc
             // Ignore all other filters. This is to ensure we have a overriding behavior.
             if (IsApplicable(context.ActionDescriptor))
             {
-                MediaTypeHeaderValue requestContentType = null;
-                MediaTypeHeaderValue.TryParse(context.HttpContext.Request.ContentType, out requestContentType);
+                var requestContentType = context.HttpContext.Request.ContentType;
 
                 // Confirm the request's content type is more specific than a media type this action supports e.g. OK
                 // if client sent "text/plain" data and this action supports "text/*".
                 if (requestContentType != null &&
-                    !ContentTypes.Any(contentType => requestContentType.IsSubsetOf(contentType)))
+                    !ContentTypes.Any(contentType => MediaTypeComparisons.IsSubsetOf(contentType, requestContentType)))
                 {
                     context.Result = new UnsupportedMediaTypeResult();
                 }
@@ -86,8 +96,7 @@ namespace Microsoft.AspNet.Mvc
                 return true;
             }
 
-            MediaTypeHeaderValue requestContentType = null;
-            MediaTypeHeaderValue.TryParse(context.RouteContext.HttpContext.Request.ContentType, out requestContentType);
+            var requestContentType = context.RouteContext.HttpContext.Request.ContentType;
 
             // If the request content type is null we need to act like pass through.
             // In case there is a single candidate with a constraint it should be selected.
@@ -104,7 +113,7 @@ namespace Microsoft.AspNet.Mvc
 
             // Confirm the request's content type is more specific than a media type this action supports e.g. OK
             // if client sent "text/plain" data and this action supports "text/*".
-            if (ContentTypes.Any(contentType => requestContentType.IsSubsetOf(contentType)))
+            if (ContentTypes.Any(contentType => MediaTypeComparisons.IsSubsetOf(contentType, requestContentType)))
             {
                 return true;
             }
@@ -164,22 +173,22 @@ namespace Microsoft.AspNet.Mvc
 
         }
 
-        private List<MediaTypeHeaderValue> GetContentTypes(string firstArg, string[] args)
+        private MediaTypeCollection GetContentTypes(string firstArg, string[] args)
         {
             var completeArgs = new List<string>();
             completeArgs.Add(firstArg);
             completeArgs.AddRange(args);
-            var contentTypes = new List<MediaTypeHeaderValue>();
+            var contentTypes = new MediaTypeCollection();
             foreach (var arg in completeArgs)
             {
-                var contentType = MediaTypeHeaderValue.Parse(arg);
-                if (contentType.MatchesAllSubTypes || contentType.MatchesAllTypes)
+                if (MediaTypeComparisons.MatchesAllSubtypes(arg) ||
+                    MediaTypeComparisons.MatchesAllTypes(arg))
                 {
                     throw new InvalidOperationException(
                         Resources.FormatMatchAllContentTypeIsNotAllowed(arg));
                 }
 
-                contentTypes.Add(contentType);
+                contentTypes.Add(arg);
             }
 
             return contentTypes;
--- a/src/Microsoft.AspNet.Mvc.Core/ContentResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ContentResult.cs
@@ -16,10 +16,10 @@ namespace Microsoft.AspNet.Mvc
 {
     public class ContentResult : ActionResult
     {
-        private readonly MediaTypeHeaderValue DefaultContentType = new MediaTypeHeaderValue("text/plain")
+        private readonly string DefaultContentType = new MediaTypeHeaderValue("text/plain")
         {
             Encoding = Encoding.UTF8
-        };
+        }.ToString();
 
         /// <summary>
         /// Gets or set the content representing the body of the response.
@@ -27,9 +27,9 @@ namespace Microsoft.AspNet.Mvc
         public string Content { get; set; }
 
         /// <summary>
-        /// Gets or sets the <see cref="MediaTypeHeaderValue"/> representing the Content-Type header of the response.
+        /// Gets or sets the Content-Type header for the response.
         /// </summary>
-        public MediaTypeHeaderValue ContentType { get; set; }
+        public string ContentType { get; set; }
 
         /// <summary>
         /// Gets or sets the HTTP status code.
--- a/src/Microsoft.AspNet.Mvc.Core/ControllerBase.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/ControllerBase.cs
@@ -271,7 +271,7 @@ namespace Microsoft.AspNet.Mvc
             var result = new ContentResult
             {
                 Content = content,
-                ContentType = contentType
+                ContentType = contentType?.ToString()
             };
 
             return result;
--- a/src/Microsoft.AspNet.Mvc.Core/FileContentResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/FileContentResult.cs
@@ -31,10 +31,6 @@ namespace Microsoft.AspNet.Mvc
             {
                 throw new ArgumentNullException(nameof(fileContents));
             }
-            if (contentType == null)
-            {
-                throw new ArgumentNullException(nameof(contentType));
-            }
         }
 
         /// <summary>
@@ -45,16 +41,11 @@ namespace Microsoft.AspNet.Mvc
         /// <param name="fileContents">The bytes that represent the file contents.</param>
         /// <param name="contentType">The Content-Type header of the response.</param>
         public FileContentResult(byte[] fileContents, MediaTypeHeaderValue contentType)
-            : base(contentType)
+            : base(contentType?.ToString())
         {
             if (fileContents == null)
             {
                 throw new ArgumentNullException(nameof(fileContents));
-            }
-
-            if (contentType == null)
-            {
-                throw new ArgumentNullException(nameof(contentType));
             }
 
             FileContents = fileContents;
--- a/src/Microsoft.AspNet.Mvc.Core/FileResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/FileResult.cs
@@ -7,6 +7,7 @@ using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Mvc.Logging;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Primitives;
 using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNet.Mvc
@@ -24,7 +25,7 @@ namespace Microsoft.AspNet.Mvc
         /// the provided <paramref name="contentType"/>.
         /// </summary>
         /// <param name="contentType">The Content-Type header of the response.</param>
-        protected FileResult(MediaTypeHeaderValue contentType)
+        protected FileResult(string contentType)
         {
             if (contentType == null)
             {
@@ -35,9 +36,9 @@ namespace Microsoft.AspNet.Mvc
         }
 
         /// <summary>
-        /// Gets the <see cref="MediaTypeHeaderValue"/> representing the Content-Type header of the response.
+        /// Gets the Content-Type header for the response.
         /// </summary>
-        public MediaTypeHeaderValue ContentType { get; }
+        public string ContentType { get; }
 
         /// <summary>
         /// Gets the file name that will be used in the Content-Disposition header of the response.
--- a/src/Microsoft.AspNet.Mvc.Core/FileStreamResult.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/FileStreamResult.cs
@@ -42,16 +42,11 @@ namespace Microsoft.AspNet.Mvc
         /// <param name="fileStream">The stream with the file.</param>
         /// <param name="contentType">The Content-Type header of the response.</param>
         public FileStreamResult(Stream fileStream, MediaTypeHeaderValue contentType)
-            : base(contentType)
+            : base(contentType?.ToString())
         {
             if (fileStream == null)
             {
                 throw new ArgumentNullException(nameof(fileStream));
-            }
-
-            if (contentType == null)
-            {
-                throw new ArgumentNullException(nameof(contentType));
             }
 
             FileStream = fileStream;
--- a/src/Microsoft.AspNet.Mvc.Core/Formatters/FormatterMappings.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Formatters/FormatterMappings.cs
@@ -4,24 +4,46 @@
 using System;
 using System.Collections.Generic;
 using Microsoft.AspNet.Mvc.Core;
+using Microsoft.Extensions.Primitives;
 using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNet.Mvc.Formatters
 {
     /// <summary>
-    /// Used to specify mapping between the URL Format and corresponding <see cref="MediaTypeHeaderValue"/>.
+    /// Used to specify mapping between the URL Format and corresponding media type.
     /// </summary>
     public class FormatterMappings
     {
-        private readonly Dictionary<string, MediaTypeHeaderValue> _map =
-            new Dictionary<string, MediaTypeHeaderValue>(StringComparer.OrdinalIgnoreCase);
+        private readonly Dictionary<string, string> _map =
+            new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
 
         /// <summary>
-        /// Sets mapping for the format to specified <see cref="MediaTypeHeaderValue"/>. 
-        /// If the format already exists, the <see cref="MediaTypeHeaderValue"/> will be overwritten with the new value.
+        /// Sets mapping for the format to specified media type. 
+        /// If the format already exists, the media type will be overwritten with the new value.
         /// </summary>
         /// <param name="format">The format value.</param>
-        /// <param name="contentType">The <see cref="MediaTypeHeaderValue"/> for the format value.</param>
+        /// <param name="contentType">The media type for the format value.</param>
+        public void SetMediaTypeMappingForFormat(string format, string contentType)
+        {
+            if (format == null)
+            {
+                throw new ArgumentNullException(nameof(format));
+            }
+
+            if (contentType == null)
+            {
+                throw new ArgumentNullException(nameof(contentType));
+            }
+
+            SetMediaTypeMappingForFormat(format, MediaTypeHeaderValue.Parse(contentType));
+        }
+
+        /// <summary>
+        /// Sets mapping for the format to specified media type. 
+        /// If the format already exists, the media type will be overwritten with the new value.
+        /// </summary>
+        /// <param name="format">The format value.</param>
+        /// <param name="contentType">The media type for the format value.</param>
         public void SetMediaTypeMappingForFormat(string format, MediaTypeHeaderValue contentType)
         {
             if (format == null)
@@ -36,31 +58,34 @@ namespace Microsoft.AspNet.Mvc.Formatters
 
             ValidateContentType(contentType);
             format = RemovePeriodIfPresent(format);
-            _map[format] = contentType.CopyAsReadOnly();
+            _map[format] = contentType.ToString();
         }
 
         /// <summary>
-        /// Gets <see cref="MediaTypeHeaderValue"/> for the specified format.
+        /// Gets the media type for the specified format.
         /// </summary>
         /// <param name="format">The format value.</param>
-        /// <returns>The <see cref="MediaTypeHeaderValue"/> for input format.</returns>
-        public MediaTypeHeaderValue GetMediaTypeMappingForFormat(string format)
+        /// <returns>The media type for input format.</returns>
+        public string GetMediaTypeMappingForFormat(string format)
         {
-            if (format == null)
+            if (string.IsNullOrEmpty(format))
             {
-                throw new ArgumentNullException(nameof(format));
+                var message = Resources.FormatFormatFormatterMappings_GetMediaTypeMappingForFormat_InvalidFormat(
+                    nameof(format));
+
+                throw new ArgumentException(message, nameof(format));
             }
 
             format = RemovePeriodIfPresent(format);
 
-            MediaTypeHeaderValue value = null;
+            string value = null;
             _map.TryGetValue(format, out value);
 
             return value;
         }
 
         /// <summary>
-        /// Clears the <see cref="MediaTypeHeaderValue"/> mapping for the format.
+        /// Clears the media type mapping for the format.
         /// </summary>
         /// <param name="format">The format value.</param>
         /// <returns><c>true</c> if the format is successfully found and cleared; otherwise, <c>false</c>.</returns>
--- a/src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
@@ -8,6 +8,7 @@ using System.Reflection;
 using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Mvc.Core;
+using Microsoft.Extensions.Primitives;
 using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNet.Mvc.Formatters
@@ -37,10 +38,10 @@ namespace Microsoft.AspNet.Mvc.Formatters
         public IList<Encoding> SupportedEncodings { get; } = new List<Encoding>();
 
         /// <summary>
-        /// Gets the mutable collection of <see cref="MediaTypeHeaderValue"/> elements supported by
+        /// Gets the mutable collection of media type elements supported by
         /// this <see cref="InputFormatter"/>.
         /// </summary>
-        public IList<MediaTypeHeaderValue> SupportedMediaTypes { get; } = new List<MediaTypeHeaderValue>();
+        public MediaTypeCollection SupportedMediaTypes { get; } = new MediaTypeCollection();
 
         protected object GetDefaultValueForType(Type modelType)
         {
@@ -61,8 +62,7 @@ namespace Microsoft.AspNet.Mvc.Formatters
             }
 
             var contentType = context.HttpContext.Request.ContentType;
-            MediaTypeHeaderValue requestContentType;
-            if (!MediaTypeHeaderValue.TryParse(contentType, out requestContentType))
+            if (string.IsNullOrEmpty(contentType))
             {
                 return false;
             }
@@ -71,7 +71,7 @@ namespace Microsoft.AspNet.Mvc.Formatters
             // client sent "text/plain" data and this formatter supports "text/*".
             return SupportedMediaTypes.Any(supportedMediaType =>
             {
-                return requestContentType.IsSubsetOf(supportedMediaType);
+                return MediaTypeComparisons.IsSubsetOf(supportedMediaType, contentType);
             });
         }
 
@@ -107,27 +107,28 @@ namespace Microsoft.AspNet.Mvc.Formatters
 
         /// <summary>
         /// Returns an <see cref="Encoding"/> based on <paramref name="context"/>'s
-        /// <see cref="MediaTypeHeaderValue.Charset"/>.
+        /// character set.
         /// </summary>
         /// <param name="context">The <see cref="InputFormatterContext"/>.</param>
         /// <returns>
         /// An <see cref="Encoding"/> based on <paramref name="context"/>'s
-        /// <see cref="MediaTypeHeaderValue.Charset"/>. <c>null</c> if no supported encoding was found.
+        /// character set. <c>null</c> if no supported encoding was found.
         /// </returns>
         protected Encoding SelectCharacterEncoding(InputFormatterContext context)
         {
             var request = context.HttpContext.Request;
 
-            MediaTypeHeaderValue contentType;
-            MediaTypeHeaderValue.TryParse(request.ContentType, out contentType);
-            if (contentType != null)
+            if (request.ContentType != null)
             {
-                var charset = contentType.Charset;
-                if (!string.IsNullOrEmpty(charset))
+                var encoding = MediaTypeEncoding.GetEncoding(request.ContentType);
+                if (encoding != null)
                 {
                     foreach (var supportedEncoding in SupportedEncodings)
                     {
-                        if (string.Equals(charset, supportedEncoding.WebName, StringComparison.OrdinalIgnoreCase))
+                        if (string.Equals(
+                            encoding.WebName,
+                            supportedEncoding.WebName,
+                            StringComparison.OrdinalIgnoreCase))
                         {
                             return supportedEncoding;
                         }
--- a/src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeCollection.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeCollection.cs
@@ -1,0 +1,73 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.ObjectModel;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNet.Mvc.Formatters
+{
+    /// <summary>
+    /// A collection of media types.
+    /// </summary>
+    public class MediaTypeCollection : Collection<string>
+    {
+        /// <summary>
+        /// Initializes a new instance of <see cref="MediaTypeCollection"/>.
+        /// </summary>
+        public MediaTypeCollection()
+        {
+        }
+
+        /// <summary>
+        /// Adds an object to the end of the <see cref="MediaTypeCollection"/>. 
+        /// </summary>
+        /// <param name="item">The media type to be added to the end of the <see cref="MediaTypeCollection"/>.</param>
+        public void Add(MediaTypeHeaderValue item)
+        {
+            if (item == null)
+            {
+                throw new ArgumentNullException(nameof(item));
+            }
+
+            Add(item.ToString());
+        }
+
+        /// <summary>
+        /// Inserts an element into the <see cref="MediaTypeCollection"/> at the specified index.
+        /// </summary>
+        /// <param name="index">The zero-based index at which <paramref name="item"/> should be inserted.</param>
+        /// <param name="item">The media type to insert.</param>
+        public void Insert(int index, MediaTypeHeaderValue item)
+        {
+            if (index < 0 || index > Count)
+            {
+                throw new ArgumentOutOfRangeException(nameof(index));
+            }
+
+            if (item == null)
+            {
+                throw new ArgumentNullException(nameof(item));
+            }
+
+            Insert(index, item.ToString());
+        }
+
+        /// <summary>
+        /// Removes the first occurrence of a specific media type from the <see cref="MediaTypeCollection"/>.
+        /// </summary>
+        /// <param name="item"></param>
+        /// <returns><code>true</code> if <paramref name="item"/> is successfully removed; otherwise, <code>false</code>.
+        /// This method also returns <code>false</code> if <paramref name="item"/> was not found in the original 
+        /// <see cref="MediaTypeCollection"/>.</returns>
+        public bool Remove(MediaTypeHeaderValue item)
+        {
+            if (item == null)
+            {
+                throw new ArgumentNullException(nameof(item));
+            }
+
+            return Remove(item.ToString());
+        }
+    }
+}
--- a/src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
@@ -1,0 +1,120 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Primitives;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNet.Mvc.Formatters
+{
+    /// <summary>
+    /// Different types of tests against media type values.
+    /// </summary>
+    public static class MediaTypeComparisons
+    {
+        /// <summary>
+        /// Determines if the <paramref name="subset" /> media type is a subset of the <paramref name="set" /> media type
+        /// without taking into account the quality parameter.
+        /// </summary>
+        /// <param name="set">The more general media type.</param>
+        /// <param name="subset">The more specific media type.</param>
+        /// <returns><code>true</code> if <paramref name="set" /> is a more general media type than <paramref name="subset"/>;
+        /// otherwise <code>false</code>.</returns>
+        public static bool IsSubsetOf(StringSegment set, string subset)
+        {
+            return IsSubsetOf(set, new StringSegment(subset));
+        }
+
+        /// <summary>
+        /// Determines if the <paramref name="subset" /> media type is a subset of the <paramref name="set" /> media type
+        /// without taking into account the quality parameter.
+        /// </summary>
+        /// <param name="set">The more general media type.</param>
+        /// <param name="subset">The more specific media type.</param>
+        /// <returns><code>true</code> if <paramref name="set" /> is a more general media type than <paramref name="subset"/>;
+        /// otherwise <code>false</code>.</returns>
+        public static bool IsSubsetOf(string set, string subset)
+        {
+            return IsSubsetOf(new StringSegment(set), new StringSegment(subset));
+        }
+
+        /// <summary>
+        /// Determines if the <paramref name="subset" /> media type is a subset of the <paramref name="set" /> media type.
+        /// Two media types are compatible if one is a subset of the other ignoring any charset
+        /// parameter.
+        /// </summary>
+        /// <param name="set">The more general media type.</param>
+        /// <param name="subset">The more specific media type.</param>
+        /// <param name="ignoreQuality">Whether or not we should skip checking the quality parameter.</param>
+        /// <returns><code>true</code> if <paramref name="set" /> is a more general media type than <paramref name="subset"/>;
+        /// otherwise <code>false</code>.</returns>
+        public static bool IsSubsetOf(StringSegment set, StringSegment subset)
+        {
+            if (!set.HasValue || !subset.HasValue)
+            {
+                return false;
+            }
+
+            MediaTypeHeaderValue setMediaType;
+            MediaTypeHeaderValue subSetMediaType;
+
+            return MediaTypeHeaderValue.TryParse(set.Value, out setMediaType) &&
+                MediaTypeHeaderValue.TryParse(subset.Value, out subSetMediaType) &&
+                subSetMediaType.IsSubsetOf(setMediaType);
+        }
+
+        /// <summary>
+        /// Determines if the type of a given <paramref name="mediaType" /> matches all types, E.g, */*.
+        /// </summary>
+        /// <param name="mediaType">The media type to check</param>
+        /// <returns><code>true</code> if the <paramref name="mediaType" /> matches all subtypes; otherwise <code>false</code>.</returns>
+        public static bool MatchesAllTypes(string mediaType)
+        {
+            return MatchesAllTypes(new StringSegment(mediaType));
+        }
+
+        /// <summary>
+        /// Determines if the type of a given <paramref name="mediaType" /> matches all types, E.g, */*.
+        /// </summary>
+        /// <param name="mediaType">The media type to check</param>
+        /// <returns><code>true</code> if the <paramref name="mediaType" /> matches all subtypes; otherwise <code>false</code>.</returns>
+        public static bool MatchesAllTypes(StringSegment mediaType)
+        {
+            if (!mediaType.HasValue)
+            {
+                return false;
+            }
+
+            MediaTypeHeaderValue parsedMediaType;
+            return MediaTypeHeaderValue.TryParse(mediaType.Value, out parsedMediaType) &&
+                parsedMediaType.MatchesAllTypes;
+        }
+
+        /// <summary>
+        /// Determines if the given <paramref name="mediaType" /> matches all subtypes, E.g, text/*.
+        /// </summary>
+        /// <param name="mediaType">The media type to check</param>
+        /// <returns><code>true</code> if the <paramref name="mediaType" /> matches all subtypes; otherwise <code>false</code>.</returns>
+        public static bool MatchesAllSubtypes(string mediaType)
+        {
+            return MatchesAllSubtypes(new StringSegment(mediaType));
+        }
+
+        /// <summary>
+        /// Determines if the given <paramref name="mediaType" /> matches all subtypes, E.g, text/*.
+        /// </summary>
+        /// <param name="mediaType">The media type to check</param>
+        /// <returns><code>true</code> if the <paramref name="mediaType" /> matches all subtypes; otherwise <code>false</code>.</returns>
+        public static bool MatchesAllSubtypes(StringSegment mediaType)
+        {
+            if (!mediaType.HasValue)
+            {
+                return false;
+            }
+
+            MediaTypeHeaderValue parsedMediaType;
+            return MediaTypeHeaderValue.TryParse(mediaType.Value, out parsedMediaType) &&
+                parsedMediaType.MatchesAllSubTypes;
+        }
+    }
+}
--- a/src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeEncoding.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeEncoding.cs
@@ -1,0 +1,99 @@
+﻿using System;
+using System.Text;
+using Microsoft.Extensions.Primitives;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNet.Mvc.Formatters
+{
+    /// <summary>
+    /// A set of operations to manipulate the encoding of a media type value.
+    /// </summary>
+    public class MediaTypeEncoding
+    {
+        /// <summary>
+        /// Gets the <see cref="Encoding"/> for the given <see cref="mediaType"/> if it exists.
+        /// </summary>
+        /// <param name="mediaType">The media type from which to get the charset parameter.</param>
+        /// <returns>The <see cref="Encoding"/> of the media type if it exists; otherwise <code>null</code>.</returns>
+        public static Encoding GetEncoding(StringSegment mediaType)
+        {
+            var charset = GetCharsetParameter(mediaType);
+            return GetEncodingFromCharset(charset);
+        }
+
+        /// <summary>
+        /// Gets the <see cref="Encoding"/> for the given <see cref="mediaType"/> if it exists.
+        /// </summary>
+        /// <param name="mediaType">The media type from which to get the charset parameter.</param>
+        /// <returns>The <see cref="Encoding"/> of the media type if it exists or a <see cref="StringSegment"/> without value if not.</returns>
+        public static Encoding GetEncoding(string mediaType)
+        {
+            var charset = GetCharsetParameter(new StringSegment(mediaType));
+            return GetEncodingFromCharset(charset);
+        }
+
+        /// <summary>
+        /// Gets the charset parameter of the given <paramref name="mediaType"/> if it exists.
+        /// </summary>
+        /// <param name="mediaType">The media type from which to get the charset parameter.</param>
+        /// <returns>The charset of the media type if it exists or a <see cref="StringSegment"/> without value if not.</returns>
+        public static StringSegment GetCharsetParameter(StringSegment mediaType)
+        {
+            MediaTypeHeaderValue parsedMediaType;
+            if (MediaTypeHeaderValue.TryParse(mediaType.Value, out parsedMediaType))
+            {
+                return new StringSegment(parsedMediaType.Charset);
+            }
+            return new StringSegment();
+        }
+
+        /// <summary>
+        /// Replaces the encoding of the given <paramref name="mediaType"/> with the provided
+        /// <paramref name="encoding"/>.
+        /// </summary>
+        /// <param name="mediaType">The media type whose encoding will be replaced.</param>
+        /// <param name="encoding">The encoding that will replace the encoding in the <paramref name="mediaType"/></param>
+        /// <returns>A media type with the replaced encoding.</returns>
+        public static string ReplaceEncoding(string mediaType, Encoding encoding)
+        {
+            return ReplaceEncoding(new StringSegment(mediaType), encoding);
+        }
+
+        /// <summary>
+        /// Replaces the encoding of the given <paramref name="mediaType"/> with the provided
+        /// <paramref name="encoding"/>.
+        /// </summary>
+        /// <param name="mediaType">The media type whose encoding will be replaced.</param>
+        /// <param name="encoding">The encoding that will replace the encoding in the <paramref name="mediaType"/></param>
+        /// <returns>A media type with the replaced encoding.</returns>
+        public static string ReplaceEncoding(StringSegment mediaType, Encoding encoding)
+        {
+            var parsedMediaType = MediaTypeHeaderValue.Parse(mediaType.Value);
+            parsedMediaType.Encoding = encoding;
+
+            return parsedMediaType.ToString();
+        }
+
+        private static Encoding GetEncodingFromCharset(StringSegment charset)
+        {
+            if (charset.Equals("utf-8", StringComparison.OrdinalIgnoreCase))
+            {
+                // This is an optimization for utf-8 that prevents the Substring caused by
+                // charset.Value
+                return Encoding.UTF8;
+            }
+
+            try
+            {
+                // charset.Value might be an invalid encoding name as in charset=invalid.
+                // For that reason, we catch the exception thrown by Encoding.GetEncoding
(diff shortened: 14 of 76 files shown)


javiercn:
@rynowak

src/Microsoft.AspNet.Mvc.ApiExplorer/ApiResponseFormat.cs
```diff
@@ -2,7 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using Microsoft.AspNet.Mvc.Formatters;
-using Microsoft.Net.Http.Headers;
+using Microsoft.Extensions.Primitives;
```
dougbu: Why is `Microsoft.Extensions.Primitives` needed to use a `string`?

src/Microsoft.AspNet.Mvc.ApiExplorer/DefaultApiDescriptionProvider.cs
```diff
@@ -332,7 +333,9 @@ private string GetRelativePath(RouteTemplate parsedTemplate)
                     var responseFormatMetadataProvider = formatter as IApiResponseFormatMetadataProvider;
                     if (responseFormatMetadataProvider != null)
                     {
-                        var supportedTypes = responseFormatMetadataProvider.GetSupportedContentTypes(contentType, type);
+                        var supportedTypes = responseFormatMetadataProvider.GetSupportedContentTypes(
+                            contentType,
+                            type);
```
dougbu: Revert. Line easier to read with 121 `char`s.

src/Microsoft.AspNet.Mvc.Core/ApiExplorer/IApiResponseFormatMetadataProvider.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using Microsoft.Extensions.Primitives;
```
dougbu: Why?

src/Microsoft.AspNet.Mvc.Core/ConsumesAttribute.cs
```diff
@@ -30,6 +32,14 @@ public ConsumesAttribute(string contentType, params string[] otherContentTypes)
                 throw new ArgumentNullException(nameof(contentType));
             }
 
+            // For validation purposes, preserves existing behavior.
```
dougbu: Is this a comment to reviewers? Won't tell a developer anything tomorrow.

src/Microsoft.AspNet.Mvc.ApiExplorer/DefaultApiDescriptionProvider.cs
```diff
@@ -322,7 +323,7 @@ private string GetRelativePath(RouteTemplate parsedTemplate)
 
             if (contentTypes.Count == 0)
             {
-                contentTypes.Add(null);
+                contentTypes.Add((string)null);
```
davidfowl: Why is this needed?

src/Microsoft.AspNet.Mvc.Core/ConsumesAttribute.cs
rynowak: Yeah, this comment could be improved.

src/Microsoft.AspNet.Mvc.Core/ConsumesAttribute.cs
```diff
@@ -30,6 +32,14 @@ public ConsumesAttribute(string contentType, params string[] otherContentTypes)
...
+            MediaTypeHeaderValue.Parse(contentType);
+
+            for (int i = 0; i < otherContentTypes.Length; i++)
```
rynowak: var

src/Microsoft.AspNet.Mvc.Core/ConsumesAttribute.cs
```diff
@@ -53,13 +63,12 @@ public void OnResourceExecuting(ResourceExecutingContext context)
...
             {
-                MediaTypeHeaderValue requestContentType = null;
-                MediaTypeHeaderValue.TryParse(context.HttpContext.Request.ContentType, out requestContentType);
+                var requestContentType = context.HttpContext.Request.ContentType;
 
                 // Confirm the request's content type is more specific than a media type this action supports e.g. OK
                 // if client sent "text/plain" data and this action supports "text/*".
                 if (requestContentType != null &&
-                    !ContentTypes.Any(contentType => requestContentType.IsSubsetOf(contentType)))
+                    !ContentTypes.Any(contentType => MediaTypeComparisons.IsSubsetOf(contentType, requestContentType)))
```
rynowak: You've been awared 16 RyanPoints for this.

src/Microsoft.AspNet.Mvc.Core/ControllerBase.cs
```diff
@@ -266,7 +260,7 @@ public virtual ContentResult Content(string content, string contentType, Encodin
         /// <param name="contentType">The content type (MIME type).</param>
         /// <returns>The created <see cref="ContentResult"/> object for the response.</returns>
         [NonAction]
-        public virtual ContentResult Content(string content, MediaTypeHeaderValue contentType)
```
rynowak: Revert the API changes.

rynowak: We want to keep this overload

src/Microsoft.AspNet.Mvc.Core/FileContentResult.cs
```diff
@@ -25,26 +26,6 @@ public class FileContentResult : FileResult
...
-        }
-
-        /// <summary>
-        /// Creates a new <see cref="FileContentResult"/> instance with
-        /// the provided <paramref name="fileContents"/> and the
-        /// provided <paramref name="contentType"/>.
-        /// </summary>
-        /// <param name="fileContents">The bytes that represent the file contents.</param>
-        /// <param name="contentType">The Content-Type header of the response.</param>
-        public FileContentResult(byte[] fileContents, MediaTypeHeaderValue contentType)
```
rynowak: Revert

src/Microsoft.AspNet.Mvc.Core/FileStreamResult.cs
```diff
@@ -30,18 +31,6 @@ public class FileStreamResult : FileResult
...
-        }
-
-        /// <summary>
-        /// Creates a new <see cref="FileStreamResult"/> instance with
-        /// the provided <paramref name="fileStream"/> and the
-        /// provided <paramref name="contentType"/>.
-        /// </summary>
-        /// <param name="fileStream">The stream with the file.</param>
-        /// <param name="contentType">The Content-Type header of the response.</param>
-        public FileStreamResult(Stream fileStream, MediaTypeHeaderValue contentType)
```
rynowak: Revert

src/Microsoft.AspNet.Mvc.Core/Formatters/FormatterMappings.cs
```diff
@@ -4,24 +4,25 @@
...
         /// <summary>
-        /// Sets mapping for the format to specified <see cref="MediaTypeHeaderValue"/>. 
-        /// If the format already exists, the <see cref="MediaTypeHeaderValue"/> will be overwritten with the new value.
+        /// Sets mapping for the format to specified media type. 
+        /// If the format already exists, the media type will be overwritten with the new value.
         /// </summary>
         /// <param name="format">The format value.</param>
-        /// <param name="contentType">The <see cref="MediaTypeHeaderValue"/> for the format value.</param>
+        /// <param name="contentType">The media type for the format value.</param>
         public void SetMediaTypeMappingForFormat(string format, MediaTypeHeaderValue contentType)
```
rynowak: Add a `Set...(string, string)` if not present

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeCollection.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using System.Collections.ObjectModel;
+using Microsoft.Extensions.Primitives;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNet.Mvc.Formatters
+{
+    public class MediaTypeCollection : Collection<string>
```
rynowak: Docs

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeCollection.cs
```diff
@@ -0,0 +1,44 @@
...
+        {
+            if (item == null)
+            {
+                throw new ArgumentNullException(nameof(item));
+            }
+
+            Add(item.ToString());
+        }
+
+        public void Insert(int index, MediaTypeHeaderValue item)
```
rynowak: Check index as well

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeCollection.cs
```diff
@@ -0,0 +1,44 @@
...
+        {
+            if (item == null)
+            {
+                throw new ArgumentNullException(nameof(item));
+            }
+
+            Insert(index, item.ToString());
+        }
+
+        public void Remove(MediaTypeHeaderValue item)
```
rynowak: Does remove return a `bool`?

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
```diff
@@ -0,0 +1,292 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Primitives;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNet.Mvc.Formatters
+{
+    public static class MediaTypeComparisons
```
rynowak: Docs

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
```diff
@@ -0,0 +1,292 @@
...
+        /// Determines if the <paramref name="subset" /> media type is a subset of the <paramref name="set" /> media type.
+        /// Two media types are compatible if one is a subset of the other ignoring any charset
+        /// parameter.
+        /// </summary>
+        /// <param name="set">The more general media type.</param>
+        /// <param name="subset">The more specific media type.</param>
+        /// <param name="ignoreQuality">Whether or not we should skip checking the quality parameter.</param>
+        /// <returns><code>true</code> if <paramref name="set" />is a more general media type than <paramref name="subset"/>;
+        /// otherwise <code>false</code>.</returns>
+        public static bool IsSubsetOf(StringSegment set, StringSegment subset, bool ignoreQuality)
```
rynowak: `ignoreQuality` What do we need this for?

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
```diff
@@ -0,0 +1,292 @@
...
+        {
+            if (!set.HasValue || !subset.HasValue)
+            {
+                return false;
+            }
+
+            try
+            {
+                var setMt = MediaTypeHeaderValue.Parse(set.Value);
+                var subSetMt = MediaTypeHeaderValue.Parse(subset.Value);
```
rynowak: Names are horrid

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
```diff
@@ -0,0 +1,292 @@
...
+
+                if (!ignoreQuality && setMt.Quality != subSetMt.Quality)
+                {
+                    return false;
+                }
+
+                return subSetMt.IsSubsetOf(setMt);
+            }
+            catch (Exception)
```
rynowak: Why?

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
```diff
@@ -0,0 +1,292 @@
...
+        /// <summary>
+        /// Determines if two media types are compatible.
+        /// Two media types are compatible if one is a subset of the other ignoring any charset
+        /// parameter.
+        /// </summary>
+        /// <param name="set">The more general media type.</param>
+        /// <param name="subset">The more specific media type.</param>
+        /// <returns><code>true</code> if <paramref name="set" />is a more general mediaType than <paramref name="subset"/>
+        /// except for the charset and quality parameters; otherwise <code>false</code>.</returns>
+        public static bool IsCompatible(StringSegment set, StringSegment subset)
```
rynowak: What do we need these for?

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
```diff
@@ -0,0 +1,292 @@
...
+
+        /// <summary>
+        /// Determines if two media types are equal except for the quality parameter.
+        /// </summary>
+        /// <param name="left">The first media type.</param>
+        /// <param name="right">The second media type.</param>
+        /// <param name="ignoreQuality">Whether or not we should avoid comparing the quality parameter.</param>
+        /// <returns><code>true</code> if <paramref name="set" />is equal to <paramref name="subset"/>;
+        /// otherwise <code>false</code>.</returns>
+        public static bool AreEquivalent(StringSegment left, StringSegment right, bool ignoreQuality)
```
rynowak: What do we need these for?

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeEncoding.cs
```diff
@@ -0,0 +1,91 @@
+﻿using System;
+using System.Text;
+using Microsoft.Extensions.Primitives;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNet.Mvc.Formatters
+{
+    public class MediaTypeEncoding
```
rynowak: static. docs

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeEncoding.cs
```diff
@@ -0,0 +1,91 @@
...
+
+            }
+            catch (Exception)
+            {
+
+                return new StringSegment();
+            }
+        }
+
+        public static string ReplaceEncoding(string mediaType, Encoding encoding)
```
rynowak: Docs

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeEncoding.cs
```diff
@@ -0,0 +1,91 @@
...
+            if (charset.Equals("utf-8", StringComparison.OrdinalIgnoreCase))
+            {
+                // This is an optimization for utf-8 that prevents the Substring caused by
+                // charset.Value
+                return Encoding.UTF8;
+            }
+
+            try
+            {
+                return charset.HasValue ? Encoding.GetEncoding(charset.Value) : null;
```
rynowak: This is unfortunate. Leave a comment

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeSegmentWithQuality.cs
```diff
@@ -0,0 +1,38 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Diagnostics;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNet.Mvc
```
rynowak: Nope.

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeSegmentWithQuality.cs
```diff
@@ -0,0 +1,38 @@
...
+    public struct MediaTypeSegmentWithQuality
+    {
+        public MediaTypeSegmentWithQuality(StringSegment mediaType, double quality)
+        {
+            MediaType = mediaType;
+            Quality = quality;
+        }
+
+        /// <summary>
+        /// Gets or sets the media type of this <see cref="MediaTypeSegmentWithQuality"/>.
```
rynowak: gets

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeSegmentWithQuality.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// </summary>
+        public StringSegment MediaType { get; }
+
+        /// <summary>
+        /// Gets or sets the quality of this <see cref="MediaTypeSegmentWithQuality"/>.
```
rynowak: gets

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeSegmentWithQuality.cs
```diff
@@ -0,0 +1,38 @@
...
+
+using System.Diagnostics;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNet.Mvc
+{
+    /// <summary>
+    /// Represents a media type with its associated quality.
+    /// </summary>
+    [DebuggerDisplay("{MediaType}")]
```
rynowak: You don't need this since you did `ToString`

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeSegmentWithQuality.cs
```diff
@@ -0,0 +1,38 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Diagnostics;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNet.Mvc
+{
+    /// <summary>
+    /// Represents a media type with its associated quality.
```
rynowak: Bleh

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeSegmentWithQuality.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// </summary>
+    [DebuggerDisplay("{MediaType}")]
+    public struct MediaTypeSegmentWithQuality
+    {
+        public MediaTypeSegmentWithQuality(StringSegment mediaType, double quality)
```
rynowak: Docs

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
```diff
@@ -18,7 +19,8 @@ namespace Microsoft.AspNet.Mvc.Formatters
     public abstract class OutputFormatter : IOutputFormatter, IApiResponseFormatMetadataProvider
     {
         // using a field so we can return it as both IList and IReadOnlyList
```
rynowak: Comment is stale

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
```diff
@@ -37,10 +39,10 @@ protected OutputFormatter()
         public IList<Encoding> SupportedEncodings { get; }
 
         /// <summary>
-        /// Gets the mutable collection of <see cref="MediaTypeHeaderValue"/> elements supported by
-        /// this <see cref="OutputFormatter"/>.
+        /// Gets the mutable collection of media type elements that represent the media types
+        /// supported by this <see cref="OutputFormatter"/>.
         /// </summary>
```
rynowak: It was better before

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
```diff
@@ -37,10 +39,10 @@ protected OutputFormatter()
...
-        public IList<MediaTypeHeaderValue> SupportedMediaTypes
+        public MediaTypeCollection SupportedMediaTypes
         {
             get { return _supportedMediaTypes; }
```
rynowak: No need for explicit backing field now. #PranavPoints

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
```diff
@@ -114,15 +116,18 @@ public virtual Encoding SelectCharacterEncoding(OutputFormatterWriteContext cont
                 return encoding;
             }
 
-            var charset = context.ContentType?.Charset;
-            if (charset != null)
+            if (context.ContentType.HasValue)
             {
-                for (var i = 0; i < SupportedEncodings.Count; i++)
+                var contentTypeEncoding = MediaTypeEncoding.GetEncoding(context.ContentType);
```
rynowak: Since `SupportedEncodings` is a thing... would it be better to just return the charset here?

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
```diff
@@ -246,6 +250,33 @@ public virtual void WriteResponseHeaders(OutputFormatterWriteContext context)
...
+            else if (mediaTypeEncoding == null)
+            {
+                return CreateMediaTypeWithEncoding(mediaType, encoding);
+            }
+            else
+            {
+                // This can happen if the user has overriden SelectCharacterEncoding
+                return MediaTypeEncoding.ReplaceEncoding(mediaType, encoding);
+            }
+        }
```
rynowak: This seems like something that should be part of the `MediaTypeEncoding` class

src/Microsoft.AspNet.Mvc.Core/Infrastructure/ObjectResultExecutor.cs
```diff
@@ -270,6 +251,58 @@ public virtual Task ExecuteAsync(ActionContext context, ObjectResult result)
             return selectedFormatter;
         }
 
+        private List<MediaTypeSegmentWithQuality> GetMediaTypes(
+            MediaTypeCollection contentTypes,
+            HttpRequest request)
+        {
+            var result = new List<MediaTypeSegmentWithQuality>();
+            var parsedHeaders = request.GetTypedHeaders().Accept;
+            for (int i = 0; i < parsedHeaders?.Count; i++)
```
rynowak: var

src/Microsoft.AspNet.Mvc.Core/Infrastructure/ObjectResultExecutor.cs
```diff
@@ -324,7 +357,7 @@ public virtual Task ExecuteAsync(ActionContext context, ObjectResult result)
         protected virtual IOutputFormatter SelectFormatterUsingSortedAcceptHeaders(
             OutputFormatterWriteContext formatterContext,
             IList<IOutputFormatter> formatters,
-            IList<MediaTypeHeaderValue> sortedAcceptHeaders)
+            IList<MediaTypeSegmentWithQuality> acceptHeader)
```
rynowak: This was better before

src/Microsoft.AspNet.Mvc.Core/Infrastructure/ObjectResultExecutor.cs
```diff
@@ -336,16 +369,16 @@ public virtual Task ExecuteAsync(ActionContext context, ObjectResult result)
...
+            if (acceptHeader == null)
             {
-                throw new ArgumentNullException(nameof(sortedAcceptHeaders));
+                throw new ArgumentNullException(nameof(acceptHeader));
             }
 
-            foreach (var contentType in sortedAcceptHeaders)
+            foreach (var mediaType in acceptHeader)
             {
+                formatterContext.ContentType = mediaType.MediaType;
```
rynowak: This was better before

src/Microsoft.AspNet.Mvc.Core/PhysicalFileResult.cs
```diff
@@ -28,25 +29,6 @@ public class PhysicalFileResult : FileResult
...
-            }
-        }
-
-        /// <summary>
-        /// Creates a new <see cref="PhysicalFileResult"/> instance with
-        /// the provided <paramref name="fileName"/> and the provided <paramref name="contentType"/>.
-        /// </summary>
-        /// <param name="fileName">The path to the file. The path must be an absolute path.</param>
-        /// <param name="contentType">The Content-Type header of the response.</param>
-        public PhysicalFileResult(string fileName, MediaTypeHeaderValue contentType)
```
rynowak: Revert

src/Microsoft.AspNet.Mvc.Core/ProducesAttribute.cs
```diff
@@ -46,12 +47,20 @@ public ProducesAttribute(string contentType, params string[] additionalContentTy
                 throw new ArgumentNullException(nameof(contentType));
             }
 
+            // For validation purposes, preserves existing behavior.
```
rynowak: Make this comment better

src/Microsoft.AspNet.Mvc.Core/VirtualFileResult.cs
```diff
@@ -25,32 +26,13 @@ public class VirtualFileResult : FileResult
...
-        }
-
-        /// <summary>
         /// Creates a new <see cref="VirtualFileResult"/> instance with
         /// the provided <paramref name="fileName"/> and the
         /// provided <paramref name="contentType"/>.
         /// </summary>
         /// <param name="fileName">The path to the file. The path must be relative/virtual.</param>
         /// <param name="contentType">The Content-Type header of the response.</param>
-        public VirtualFileResult(string fileName, MediaTypeHeaderValue contentType)
```
rynowak: Revert

src/Microsoft.AspNet.Mvc.Core/project.json
```diff
@@ -30,6 +30,10 @@
       "version": "1.0.0-*",
       "type": "build"
     },
+    "Microsoft.Extensions.HashCodeCombiner.Sources": {
+      "version": "1.0.0-*",
+      "type": "build"
+    },
```
rynowak: Is this used anywhere?

src/Microsoft.AspNet.Mvc.Formatters.Xml/XmlDataContractSerializerOutputFormatter.cs
```diff
@@ -186,8 +187,8 @@ public override Task WriteResponseBodyAsync(OutputFormatterWriteContext context)
                 throw new ArgumentNullException(nameof(context));
             }
 
-            var writerSettings = WriterSettings.Clone();
-            writerSettings.Encoding = context.ContentType?.Encoding ?? Encoding.UTF8;
+            var writerSettings = WriterSettings.Clone();;
```
rynowak: :astonished:

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponentResult.cs
```diff
@@ -33,7 +33,7 @@ public class ViewComponentResult : ActionResult
         /// <summary>
         /// Gets or sets the <see cref="MediaTypeHeaderValue"/> representing the Content-Type header of the response.
```
rynowak: Update

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/ViewExecutor.cs
```diff
@@ -25,10 +25,7 @@ public class ViewExecutor
         /// <summary>
         /// The default content-type header value for views, <c>text/html; charset=utf8</c>.
```
rynowak: Comment isn't 100% right :trollface:

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewResult.cs
```diff
@@ -51,9 +51,9 @@ public class ViewResult : ActionResult
         public IViewEngine ViewEngine { get; set; }
 
         /// <summary>
-        /// Gets or sets the <see cref="MediaTypeHeaderValue"/> representing the Content-Type header of the response.
+        /// Gets or sets the media type representing the Content-Type header of the response.
         /// </summary>
```
rynowak: Comment could be improved

test/Microsoft.AspNet.Mvc.Core.Test/ContentResultTest.cs
```diff
@@ -41,7 +42,10 @@ public class ContentResultTest
             await contentResult.ExecuteResultAsync(actionContext);
 
             // Assert
-            Assert.Equal("text/plain; charset=utf-7", httpContext.Response.ContentType);
+            Assert.True(MediaTypeComparisons.AreEquivalent(
+                "text/plain; charset=utf-7",
+                httpContext.Response.ContentType,
+                ignoreQuality: false));
```
rynowak: Why does this need to change? Aren't they exactly equal?

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
dougbu: @javiercn +1 to what @rynowak said. You're porting `MediaTypeHeaderValue` code with the signature `public bool IsSubsetOf(MediaTypeHeaderValue otherMediaType)`. There's no `ignoreQuality` equivalent -- effectively it's always `true`. So what would call this new method with `ignoreQuality: false`?

Separately you aren't correctly ignoring the `q` parameter. `IsSubsetOf()` should ignore the `q` parameter and _all following parameters_. See [this block](https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs#L288-L293).

test/Microsoft.AspNet.Mvc.Core.Test/ContentResultTest.cs
```diff
@@ -41,7 +42,10 @@ public class ContentResultTest
             await contentResult.ExecuteResultAsync(actionContext);
 
             // Assert
-            Assert.Equal("text/plain; charset=utf-7", httpContext.Response.ContentType);
+            Assert.True(MediaTypeComparisons.AreEquivalent(
```
dougbu: This is not readable and `Assert.True` should be the last resort. Rather than adding all these `AreEquivalent()` overloads to the _product_, suggest a `MediaTypeAssertExtensions` class containing for example `AreEquivalent(this Assert assert, string, string, bool)` overloads. Best if that extensions class can live in a single project but use `Microsoft.AspNet.Mvc.TestCommon` if mandatory. Or do the `ExceptionAssert` / `TypeAssert` / `XmlAssert` thing and provide test helpers in a well-named `static` class.

rynowak: Or even better (IMO) - just revert all of this and use string equality.

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
javiercn: now that we have only strings passing around an invalid media type can make it's way here, so in that case we just return false

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
javiercn: removed

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
rynowak: So just throw?

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
javiercn: moved to test common, only used for testing

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
rynowak: Also, you can use `TryParse` here if that's the intent.

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
javiercn: Removed the quality overload, will change the implementation on part 3 to skip anything after q

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
javiercn: All this body goes away in the next PR

src/Microsoft.AspNet.Mvc.Core/Formatters/FormatterMappings.cs
javiercn: done

src/Microsoft.AspNet.Mvc.Core/ConsumesAttribute.cs
javiercn: Changed with             // We want to ensure that the given provided content types are valid values, so
            // we validate them using the semantics of MediaTypeHeaderValue.

src/Microsoft.AspNet.Mvc.Core/ConsumesAttribute.cs
javiercn: Is that good or bad

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeCollection.cs
javiercn: will do

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
javiercn: will do

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeCollection.cs
javiercn: changed

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeCollection.cs
javiercn: done

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeEncoding.cs
javiercn: Can you explain yourself?

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
javiercn: I don't see how this comment is stale, but I'm removing it

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeEncoding.cs
rynowak: There's no way to `TryGetEncoding` in the BCL, which is unfortunate. If our only alternative is to call something that might throw, we should leave a comment.

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
javiercn: Don't understand this comment, you mean in MediaTypeEncoding?

src/Microsoft.AspNet.Mvc.Core/Infrastructure/ObjectResultExecutor.cs
javiercn: reverted

src/Microsoft.AspNet.Mvc.Core/PhysicalFileResult.cs
javiercn: will do

src/Microsoft.AspNet.Mvc.Core/ProducesAttribute.cs
javiercn: changed

src/Microsoft.AspNet.Mvc.Core/project.json
javiercn: It's leftover from breaking the PR into parts

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewResult.cs
javiercn: how

test/Microsoft.AspNet.Mvc.Core.Test/ContentResultTest.cs
javiercn: String equality is fragile, we don't care about non significant whitespaces. I move the class to TestCommon, I'll rename it and follow the TypeAssert ... pattern

src/Microsoft.AspNet.Mvc.Core/project.json
rynowak: So take it out?

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/ViewExecutor.cs
rynowak: utf8 -> utf-8

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewResult.cs
rynowak: `the media type representing the Content-Type header of the response`

I would just say "the Content-Type header for the response"

src/Microsoft.AspNet.Mvc.Core/ContentResult.cs
```diff
@@ -27,9 +27,9 @@ public class ContentResult : ActionResult
         public string Content { get; set; }
 
         /// <summary>
-        /// Gets or sets the <see cref="MediaTypeHeaderValue"/> representing the Content-Type header of the response.
+        /// Gets or sets the media type representing the Content-Type header of the response.
         /// </summary>
```
rynowak: Yeah let's fix all of these. `Gets or sets the Content-Type header of the response.`

src/Microsoft.AspNet.Mvc.Core/FileResult.cs
```diff
@@ -35,9 +36,9 @@ protected FileResult(MediaTypeHeaderValue contentType)
         }
 
         /// <summary>
-        /// Gets the <see cref="MediaTypeHeaderValue"/> representing the Content-Type header of the response.
+        /// Gets the media type representing the Content-Type header of the response.
         /// </summary>
-        public MediaTypeHeaderValue ContentType { get; }
+        public string ContentType { get; }
```
rynowak: `Gets the Content-Type header of the response.`

src/Microsoft.AspNet.Mvc.Core/Formatters/FormatterMappings.cs
```diff
@@ -36,15 +58,15 @@ public void SetMediaTypeMappingForFormat(string format, MediaTypeHeaderValue con
 
             ValidateContentType(contentType);
             format = RemovePeriodIfPresent(format);
-            _map[format] = contentType.CopyAsReadOnly();
+            _map[format] = contentType.ToString();
         }
 
         /// <summary>
-        /// Gets <see cref="MediaTypeHeaderValue"/> for the specified format.
+        /// Gets media type for the specified format.
```
rynowak: Gets **the** media type...

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeCollection.cs
```diff
@@ -0,0 +1,73 @@
...
+
+using System;
+using System.Collections.ObjectModel;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNet.Mvc.Formatters
+{
+    /// <summary>
+    /// Represents a collection of media types.
+    /// </summary>
```
rynowak: Drop `Represents`

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
```diff
@@ -0,0 +1,136 @@
...
+
+using System;
+using Microsoft.Extensions.Primitives;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNet.Mvc.Formatters
+{
+    /// <summary>
+    /// Represents different types of tests against media type values.
+    /// </summary>
```
rynowak: Drop `represents`

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
```diff
@@ -0,0 +1,136 @@
...
+    public static class MediaTypeComparisons
+    {
+        /// <summary>
+        /// Determines if the <paramref name="subset" /> media type is a subset of the <paramref name="set" /> media type
+        /// without taking into account the quality parameter.
+        /// </summary>
+        /// <param name="set">The more general media type.</param>
+        /// <param name="subset">The more specific media type.</param>
+        /// <returns><code>true</code> if <paramref name="set" />is a more general media type than <paramref name="subset"/>
```
rynowak: space after `<paramref name="set" />`

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
```diff
@@ -0,0 +1,136 @@
...
+        /// except for the quality parameter; otherwise <code>false</code>.</returns>
```
rynowak: `except for the quality parameter` is this accurate?

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
```diff
@@ -0,0 +1,136 @@
...
+                var setMediaType = MediaTypeHeaderValue.Parse(set.Value);
+                var subSetMediaType = MediaTypeHeaderValue.Parse(subset.Value);
+
+                return subSetMediaType.IsSubsetOf(setMediaType);
+            }
+            catch (Exception)
+            {
+
+                return false;
+            }
```
rynowak: Use TryParse instead.

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
```diff
@@ -0,0 +1,136 @@
...
+            if (!mediaType.HasValue)
+            {
+                return false;
+            }
+
+            try
+            {
+                return MediaTypeHeaderValue.Parse(mediaType.Value).MatchesAllSubTypes;
+            }
+            catch (Exception)
```
rynowak: TryParse

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeEncoding.cs
```diff
@@ -0,0 +1,111 @@
+﻿using System;
+using System.Text;
+using Microsoft.Extensions.Primitives;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNet.Mvc.Formatters
+{
+    /// <summary>
+    /// Represents a set of operations to manipulate the encoding of a media type value.
```
rynowak: No `Represents`

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
```diff
@@ -0,0 +1,136 @@
...
+            {
+
+                return false;
+            }
+        }
+
+        /// <summary>
+        /// Determines if the type of a given <paramref name="mediaType" /> matches all types, E.g, */*.
+        /// </summary>
+        /// <param name="mediaType">The mediaType to check</param>
```
rynowak: Fix casing/spacing of mediaType in all of these.

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeEncoding.cs
```diff
@@ -0,0 +1,111 @@
...
+    /// </summary>
+    public class MediaTypeEncoding
+    {
+        /// <summary>
+        /// Gets the <see cref="Encoding"/> for the given <see cref="mediaType"/> if it exists.
+        /// </summary>
+        /// <param name="mediaType">The mediaType from which to get the charset parameter.</param>
```
rynowak: Fix casing/spacing of mediaType in all of these.

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeEncoding.cs
```diff
@@ -0,0 +1,111 @@
...
+        /// <summary>
+        /// Gets the charset parameter of the given <paramref name="mediaType"/> if it exists.
+        /// </summary>
+        /// <param name="mediaType">The mediaType from which to get the charset parameter.</param>
+        /// <returns>The charset of the mediaType if it exists or a <see cref="StringSegment"/> without value if not.</returns>
+        public static StringSegment GetCharset(StringSegment mediaType)
+        {
+            try
+            {
+                var mt = MediaTypeHeaderValue.Parse(mediaType.Value);
```
rynowak: pick a better name and use TryParse

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeEncoding.cs
```diff
@@ -0,0 +1,111 @@
...
+        /// <param name="encoding">The encoding that will replace the encoding in the <paramref name="mediaType"/></param>
+        /// <returns>A media type with the replaced encoding.</returns>
+        public static string ReplaceEncoding(StringSegment mediaType, Encoding encoding)
+        {
+            var charset = GetCharset(mediaType);
+
+            if (GetEncodingFromCharset(charset) == encoding)
+            {
+                return mediaType.Value;
+            }
```
rynowak: Why do we need this check at all?

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeSegmentWithQuality.cs
```diff
@@ -0,0 +1,41 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNet.Mvc.Formatters
+{
+    /// <summary>
+    /// Represents a media type with its associated quality.
```
rynowak: Drop `represents`

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeSegmentWithQuality.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// </summary>
+    public struct MediaTypeSegmentWithQuality
+    {
+        /// <summary>
+        /// Initializes an instance of <see cref="MediaTypeSegmentWithQuality"/>.
+        /// </summary>
+        /// <param name="mediaType">The <see cref="StringSegment"/> containing the media type.</param>
+        /// <param name="quality">The quality parameter of the media type or 1 in case it does not exist.</param>
```
rynowak: `in the case....`

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
```diff
@@ -215,7 +212,8 @@ public Task WriteAsync(OutputFormatterWriteContext context)
             if (selectedEncoding != null)
             {
                 // Override the content type value even if one already existed.
-                selectedMediaType.Encoding = selectedEncoding;
+                var mediaTypeWithCharset = GetMediaTypeWithCharset(selectedMediaType.Value, selectedEncoding);
+                selectedMediaType = new StringSegment(mediaTypeWithCharset);
```
rynowak: Isn't this just `ReplaceEncoding`?

src/Microsoft.AspNet.Mvc.Core/Internal/ResponseContentTypeHelper.cs
```diff
@@ -5,6 +5,8 @@
 using System.Text;
 using Microsoft.Net.Http.Headers;
 using Microsoft.AspNet.Http;
+using Microsoft.Extensions.Primitives;
+using Microsoft.AspNet.Mvc.Formatters;
```
rynowak: Sort

src/Microsoft.AspNet.Mvc.Formatters.Json/JsonOutputFormatter.cs
```diff
@@ -140,7 +140,7 @@ public override Task WriteResponseBodyAsync(OutputFormatterWriteContext context)
             }
 
             var response = context.HttpContext.Response;
-            var selectedEncoding = context.ContentType?.Encoding ?? Encoding.UTF8;
+            var selectedEncoding = MediaTypeEncoding.GetEncoding(context.ContentType) ?? Encoding.UTF8;
```
rynowak: Will this be ok if `context.ContentType` is null?

src/Microsoft.AspNet.Mvc.ViewFeatures/PartialViewResult.cs
```diff
@@ -46,9 +46,9 @@ public class PartialViewResult : ActionResult
         public IViewEngine ViewEngine { get; set; }
 
         /// <summary>
-        /// Gets or sets the <see cref="MediaTypeHeaderValue"/> representing the Content-Type header of the response.
+        /// Gets or sets the media type representing the Content-Type header of the response.
```
rynowak: Lets do a find + replace for this comment

test/Microsoft.AspNet.Mvc.Core.Test/Formatters/StreamOutputFormatterTest.cs
```diff
@@ -21,7 +22,7 @@ public void CanWriteResult_ReturnsTrue_ForStreams(Type type, string contentType)
         {
             // Arrange
             var formatter = new StreamOutputFormatter();
-            var contentTypeHeader = contentType == null ? null : new MediaTypeHeaderValue(contentType);
+            StringSegment contentTypeHeader = contentType == null ? new StringSegment() : new StringSegment(contentType);
```
rynowak: var

test/Microsoft.AspNet.Mvc.Core.Test/MediaTypeCollectionTest.cs
```diff
@@ -0,0 +1,62 @@
...
+            {
+                MediaTypeHeaderValue.Parse("text/plain"),
+                MediaTypeHeaderValue.Parse("text/xml")
+            };
+
+            // Act
+            collection.Remove(MediaTypeHeaderValue.Parse("text/xml"));
+
+            // Assert
+            Assert.DoesNotContain("application/json;charset=utf-16", collection);
```
rynowak: I don't think this test is doing what you want.

test/Microsoft.AspNet.Mvc.TestCommon/MediaTypeAssert.cs
```diff
@@ -0,0 +1,59 @@
...
+                var rightMt = MediaTypeHeaderValue.Parse(right.Value);
+
+                if (!leftMt.Equals(rightMt))
+                {
+                    throw new EqualException(left.ToString(), right.ToString());
+                }
+            }
+            catch (Exception)
+            {
+                throw new EqualException(left.ToString(), right.ToString());
```
rynowak: What's the exception here since you did tryparse?

test/Microsoft.AspNet.Mvc.TestCommon/MediaTypeAssert.cs
```diff
@@ -0,0 +1,59 @@
...
+            else if (!left.HasValue || !right.HasValue)
+            {
+                throw new EqualException(left.ToString(), right.ToString());
+            }
+
+            MediaTypeHeaderValue temp;
+            if (!MediaTypeHeaderValue.TryParse(left.Value, out temp) &&
+                !MediaTypeHeaderValue.TryParse(right.Value, out temp))
+            {
+                return;
```
rynowak: So if either fails to parse they are equal?

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewFeatures/ViewResultExecutorTest.cs
```diff
@@ -277,7 +278,7 @@ public void FindView_WritesDiagnostic_ViewNotFound()
             // Check if the original instance provided by the user has not changed.
             // Since we do not have access to the new instance created within the view executor,
             // check if at least the content is the same.
-            Assert.Null(contentType.Encoding);
+            Assert.Null(MediaTypeEncoding.GetEncoding(contentType));
```
rynowak: We can take out this check IMO. It's not really proving anything

rynowak:
:watch: for small comments. Looking mostly done

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
javiercn: It'll make more sense with the second part of this change

src/Microsoft.AspNet.Mvc.Core/project.json
javiercn: it'll come back in part 3, it's just in the wrong place, but doesn't hurt

rynowak: Nah. Remove this and put it back when it's used. What if you don't end up using it after we review part 3? Then this just sits here like a turd. :hankey:

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeEncoding.cs
javiercn: It's just an optimization in case you pass the same encoding as the one on the media type

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
javiercn: This will do the cache lookup in part 2 of the PR

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeEncoding.cs
rynowak: Given that this isn't the final version of this code, this is premature. You might end up optimizing this in a totally different way. Leave it out until then.

src/Microsoft.AspNet.Mvc.Formatters.Json/JsonOutputFormatter.cs
javiercn: Yes

test/Microsoft.AspNet.Mvc.Core.Test/MediaTypeCollectionTest.cs
javiercn: fixed

rynowak:
:shipit:

src/Microsoft.AspNet.Mvc.Core/ConsumesAttribute.cs
```diff
@@ -53,13 +64,12 @@ public void OnResourceExecuting(ResourceExecutingContext context)
             // Ignore all other filters. This is to ensure we have a overriding behavior.
             if (IsApplicable(context.ActionDescriptor))
             {
-                MediaTypeHeaderValue requestContentType = null;
-                MediaTypeHeaderValue.TryParse(context.HttpContext.Request.ContentType, out requestContentType);
+                var requestContentType = context.HttpContext.Request.ContentType;
 
                 // Confirm the request's content type is more specific than a media type this action supports e.g. OK
                 // if client sent "text/plain" data and this action supports "text/*".
                 if (requestContentType != null &&
```
dougbu: Shouldn't this be `string.IsNullOrEmpty(requestContentType)` now?

src/Microsoft.AspNet.Mvc.Core/ConsumesAttribute.cs
```diff
@@ -30,6 +32,15 @@ public ConsumesAttribute(string contentType, params string[] otherContentTypes)
...
             }
 
+            // We want to ensure that the given provided content types are valid values, so
+            // we validate them using the semantics of MediaTypeHeaderValue.
+            MediaTypeHeaderValue.Parse(contentType);
+
+            for (var i = 0; i < otherContentTypes.Length; i++)
+            {
+                MediaTypeHeaderValue.Parse(otherContentTypes[i]);
+            }
```
dougbu: Previous code ignored invalid media types. Why is it now correct to throw `FormatException` in that case?

src/Microsoft.AspNet.Mvc.Core/FileContentResult.cs
```diff
@@ -45,7 +45,7 @@ public FileContentResult(byte[] fileContents, string contentType)
         /// <param name="fileContents">The bytes that represent the file contents.</param>
         /// <param name="contentType">The Content-Type header of the response.</param>
         public FileContentResult(byte[] fileContents, MediaTypeHeaderValue contentType)
-            : base(contentType)
+            : base(contentType.ToString())
```
dougbu: Is `contentType` guaranteed to be non-`null`?

src/Microsoft.AspNet.Mvc.Core/FileStreamResult.cs
```diff
@@ -42,7 +42,7 @@ public FileStreamResult(Stream fileStream, string contentType)
         /// <param name="fileStream">The stream with the file.</param>
         /// <param name="contentType">The Content-Type header of the response.</param>
         public FileStreamResult(Stream fileStream, MediaTypeHeaderValue contentType)
-            : base(contentType)
+            : base(contentType.ToString())
```
dougbu: Is `contentType` guaranteed to be non-`null`?

src/Microsoft.AspNet.Mvc.Core/Formatters/FormatterMappings.cs
```diff
@@ -36,15 +58,15 @@ public void SetMediaTypeMappingForFormat(string format, MediaTypeHeaderValue con
...
-        /// Gets <see cref="MediaTypeHeaderValue"/> for the specified format.
+        /// Gets the media type for the specified format.
         /// </summary>
         /// <param name="format">The format value.</param>
-        /// <returns>The <see cref="MediaTypeHeaderValue"/> for input format.</returns>
-        public MediaTypeHeaderValue GetMediaTypeMappingForFormat(string format)
+        /// <returns>The media type for input format.</returns>
+        public string GetMediaTypeMappingForFormat(string format)
         {
             if (format == null)
```
dougbu: Should this be `string.IsNullOrEmpty(format)`?

src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
```diff
@@ -107,27 +107,25 @@ public virtual Task<InputFormatterResult> ReadAsync(InputFormatterContext contex
...
             {
-                var charset = contentType.Charset;
-                if (!string.IsNullOrEmpty(charset))
+                var encoding = MediaTypeEncoding.GetEncoding(request.ContentType);
+                if (encoding != null)
                 {
                     foreach (var supportedEncoding in SupportedEncodings)
                     {
-                        if (string.Equals(charset, supportedEncoding.WebName, StringComparison.OrdinalIgnoreCase))
+                        if (encoding.WebName.Equals(supportedEncoding.WebName, StringComparison.OrdinalIgnoreCase))
```
dougbu: Why switch _away_ from using the `static` method? We use `string.Equals()` almost everywhere.

src/Microsoft.AspNet.Mvc.Core/PhysicalFileResult.cs
```diff
@@ -47,7 +47,7 @@ public PhysicalFileResult(string fileName, string contentType)
         /// <param name="fileName">The path to the file. The path must be an absolute path.</param>
         /// <param name="contentType">The Content-Type header of the response.</param>
         public PhysicalFileResult(string fileName, MediaTypeHeaderValue contentType)
-            : base(contentType)
+            : base(contentType.ToString())
```
dougbu: Is `contentType` guaranteed to be non-`null`?

src/Microsoft.AspNet.Mvc.Core/VirtualFileResult.cs
```diff
@@ -51,7 +51,7 @@ public VirtualFileResult(string fileName, string contentType)
         /// <param name="fileName">The path to the file. The path must be relative/virtual.</param>
         /// <param name="contentType">The Content-Type header of the response.</param>
         public VirtualFileResult(string fileName, MediaTypeHeaderValue contentType)
-            : base(contentType)
+            : base(contentType.ToString())
```
dougbu: Is `contentType` guaranteed to be non-`null`?

src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
javiercn: Didn't know that was the case, I can change it, even though it's perfectly safe to use in this case (see encoding != null above)

src/Microsoft.AspNet.Mvc.Core/ConsumesAttribute.cs
javiercn: No, in fact, I could remove requestContentType != null as MediaTypeComparisons.IsSubsetOf will do the right thing. The check for null is just a shortcut to not even try in that situation

Also, requestContentType is not a property that a user will normally set (they would need to be doing some extensibility work to do so) so I don't think is worth the change

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeEncoding.cs
```diff
@@ -0,0 +1,99 @@
...
+        /// <summary>
+        /// Replaces the encoding of the given <paramref name="mediaType"/> with the provided
+        /// <paramref name="encoding"/>.
+        /// </summary>
+        /// <param name="mediaType">The media type whose encoding will be replaced.</param>
+        /// <param name="encoding">The encoding that will replace the encoding in the <paramref name="mediaType"/></param>
+        /// <returns>A media type with the replaced encoding.</returns>
+        public static string ReplaceEncoding(StringSegment mediaType, Encoding encoding)
+        {
+            var parsedMediaType = MediaTypeHeaderValue.Parse(mediaType.Value);
```
dougbu: Why doesn't this return `null` if `mediaType` fails to parse? Doesn't look like `throw`ing is correct for all callers.

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.Core/ConsumesAttribute.cs
javiercn: Previous code didn't ignore invalid media types. See the tests
https://github.com/aspnet/Mvc/blob/dev/test/Microsoft.AspNet.Mvc.Core.Test/ConsumesAttributeTests.cs

src/Microsoft.AspNet.Mvc.Core/FileContentResult.cs
javiercn: will change to ?.ToString()

src/Microsoft.AspNet.Mvc.Core/ConsumesAttribute.cs
dougbu: Ah, sorry. I missed the `Parse()` that used to be on [line 175](https://github.com/aspnet/Mvc/pull/3862/files#diff-1b5675f1b083c83b1f98dd3e87cf08f6L175).

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeEncoding.cs
javiercn: It's fine, if you are passing an invalid media type you are doing something wrong, better to have a clear exception than return null, and get a null ref later that the user would need to track

src/Microsoft.AspNet.Mvc.Core/ConsumesAttribute.cs
dougbu: Agree there's no downside to leaving this code as-is because you've made `IsSubsetOf()` resilient.

----

FYI
>  `requestContentType` is not a property that a user will normally set

Huh? That variable comes directly from the HTTP request's headers. And that header is set in almost-all POST requests.

src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
```diff
@@ -107,27 +107,27 @@ public virtual Task<InputFormatterResult> ReadAsync(InputFormatterContext contex
...
-                if (!string.IsNullOrEmpty(charset))
+                var encoding = MediaTypeEncoding.GetEncoding(request.ContentType);
+                if (encoding != null)
                 {
                     foreach (var supportedEncoding in SupportedEncodings)
                     {
-                        if (string.Equals(charset, supportedEncoding.WebName, StringComparison.OrdinalIgnoreCase))
+                        if (string.Equals(encoding.WebName,
+                            supportedEncoding.WebName,
+                            StringComparison.OrdinalIgnoreCase))
```
dougbu: 1. If wrapping, always put _every_ argument on its own line.
2. Why is this wrapped? Looks shorter than 120 characters even with the "g.WebName" bit added.

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
```diff
@@ -0,0 +1,120 @@
...
+        /// otherwise <code>false</code>.</returns>
+        public static bool IsSubsetOf(StringSegment set, StringSegment subset)
+        {
+            if (!set.HasValue || !subset.HasValue)
+            {
+                return false;
+            }
+
+            MediaTypeHeaderValue setMediaType = null;
+            MediaTypeHeaderValue subSetMediaType = null;
```
dougbu: Don't initialize `out` arguments.

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
```diff
@@ -0,0 +1,120 @@
...
+        /// <param name="mediaType">The media type to check</param>
+        /// <returns><code>true</code> if the <paramref name="mediaType" /> matches all subtypes; otherwise <code>false</code>.</returns>
+        public static bool MatchesAllTypes(StringSegment mediaType)
+        {
+            if (!mediaType.HasValue)
+            {
+                return false;
+            }
+
+            MediaTypeHeaderValue parsedMediaType = null;
```
dougbu: Don't initialize `out` arguments.

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
```diff
@@ -0,0 +1,120 @@
...
+        /// <param name="mediaType">The media type to check</param>
+        /// <returns><code>true</code> if the <paramref name="mediaType" /> matches all subtypes; otherwise <code>false</code>.</returns>
+        public static bool MatchesAllSubtypes(StringSegment mediaType)
+        {
+            if (!mediaType.HasValue)
+            {
+                return false;
+            }
+
+            MediaTypeHeaderValue parsedMediaType = null;
```
dougbu: Don't initialize `out` arguments.

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
dougbu: For this one, `subSetMediaType` _will_ need the init.

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeEncoding.cs
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        /// <summary>
+        /// Gets the charset parameter of the given <paramref name="mediaType"/> if it exists.
+        /// </summary>
+        /// <param name="mediaType">The media type from which to get the charset parameter.</param>
+        /// <returns>The charset of the media type if it exists or a <see cref="StringSegment"/> without value if not.</returns>
+        public static StringSegment GetCharsetParameter(StringSegment mediaType)
+        {
+            MediaTypeHeaderValue parsedMediaType = null;
```
dougbu: ditto

test/Microsoft.AspNet.Mvc.Core.Test/Formatters/FormatterMappingsTest.cs
```diff
@@ -11,32 +13,43 @@ namespace Microsoft.AspNet.Mvc.Formatters
...
+        [InlineData(null)]
+        [InlineData("")]
+        public void FormatterMappings_GetMediaTypeMappingForFormat_ThrowsForInvalidFormats(string format)
+        {
+            // Arrange
+            var options = new FormatterMappings();
+
+            // Act & Assert
+            Assert.Throws<ArgumentException>("format", () => options.GetMediaTypeMappingForFormat(format));
+        }
```
dougbu: +1 useful addition

dougbu:
:shipit: with a couple of small removals.

----

In the future, to make it much easier on your reviewers,
- Please do not squash commits, especially the latest one(s) addressing comments, before getting the sign-offs you're looking for.
- Please get into the :up::date: habit.

src/Microsoft.AspNet.Mvc.Core/ConsumesAttribute.cs
javiercn: What I meant is that user in the server code will not do request.Headers["Content-Type"] = "", that the content type typically comes in from the request from the client and that it's not normal that a client sends Content-Type: (which i don't even know if would result in Content-Type: "" or Content-Type: null or our current stack

src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
javiercn: 124 chars, I fixed the wrapping.

javiercn:
@dougbu updated

dougbu:
@javiercn what am I supposed to look at? I've signed off already and you squashed the commits so it's very difficult to confirm anything except that you changed lines I commented on.

javiercn:
@dougbu you don't have to, just wanted you to know in case you wanted to.

javiercn:
Merged as part of the Part 3 pull request

PR #3863, opened 2016-01-05
[Issue3683 Part 2] Cache Content-Type string for UTF-8

javiercn:


javiercn:
@rynowak

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeSegmentWithQuality.cs
```diff
@@ -9,7 +8,6 @@ namespace Microsoft.AspNet.Mvc
     /// <summary>
     /// Represents a media type with its associated quality.
     /// </summary>
-    [DebuggerDisplay("{MediaType}")]
```
javiercn: This is just left over from breaking the PR, It will go away when I join all three parts and squash them

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
```diff
@@ -361,6 +395,25 @@ private IList<StringWithQualityHeaderValue> Sort(IList<StringWithQualityHeaderVa
...
         }
 
+        private string GetFromCacheIfCompatible(string mediaType, Encoding encoding)
+        {
+            if (encoding != Encoding.UTF8)
+            {
+                return null;
+            }
+
+            foreach (var supportedMediaType in SupportedMediaTypes)
```
rynowak: use `for` if this is an `IList<>`

rynowak: Why is `GetMediaTypeWithCharset` looking at `SupportedMediaTypes`? We already know it's a 'valid' media-type, we just need to set the charset.

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
```diff
@@ -268,12 +294,20 @@ protected string GetMediaTypeWithCharset(string mediaType, Encoding encoding)
...
+                {
+                    return CreateMediaTypeWithEncoding(mediaType, encoding);
+                }
             }
             else
             {
                 // This can happen if the user has overriden SelectCharacterEncoding
-                return MediaTypeEncoding.ReplaceEncoding(mediaType, encoding);
+                var result = GetFromCacheIfCompatible(mediaType, encoding);
+                return result != null ? result : MediaTypeEncoding.ReplaceEncoding(mediaType, encoding);
```
rynowak: I think your decomposition of these cases is wrong.

Both the `else if` and `else` handle cases where the value comes from the cache. Both the `else if` and `else` handle cases where the string needs to change.

test/Microsoft.AspNet.Mvc.Core.Test/Formatters/OutputFormatterTests.cs
```diff
@@ -101,6 +101,55 @@ public static IEnumerable<object[]> SelectResponseCharacterEncodingData
...
+            formatter.Object.SupportedMediaTypes.Add("application/json");
+
+            // Act
+            formatter.Object.WriteAsync(formatterContext);
+
+            // Assert
+            Assert.True(MediaTypeComparisons.AreEquivalent(
+                mediaTypeCache["application/json"],
+                formatterContext.ContentType, 
+                ignoreQuality: false));
```
rynowak: This doesn't seem useful. Just use string equality.

rynowak:
:watch:

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
```diff
@@ -41,6 +43,32 @@ protected OutputFormatter()
         public MediaTypeCollection SupportedMediaTypes { get; }
 
         /// <summary>
+        /// Gets a cache with of the <see cref="SupportedMediaTypes"/> with utf-8 encoding
+        /// that this formatter uses when it writes the Content-Type header of the response.
+        /// </summary>
+        protected virtual IDictionary<string, string> OutputMediaTypeCache
```
rynowak: Why protected virtual?

rynowak: I would just leave this private.

test/Microsoft.AspNet.Mvc.Core.Test/Formatters/OutputFormatterTests.cs
```diff
@@ -101,6 +102,48 @@ public static IEnumerable<object[]> SelectResponseCharacterEncodingData
...
+                {
+                    "application/json",
+                    "application/json; charset=utf-8"
+                }
+            };
+
+            formatter
+                .Protected()
+                .SetupGet<IDictionary<string, string>>("OutputMediaTypeCache")
+                .Returns(mediaTypeCache);
```
rynowak: You can write the same tests without complicated mockery by just verifying that it returns the same string twice.

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
```diff
@@ -256,17 +284,18 @@ public virtual void WriteResponseHeaders(OutputFormatterWriteContext context)
         protected string GetMediaTypeWithCharset(string mediaType, Encoding encoding)
         {
             var mediaTypeEncoding = MediaTypeEncoding.GetEncoding(mediaType);
-            if (mediaTypeEncoding == encoding)
+            if (mediaTypeEncoding != null &&
+                mediaTypeEncoding.WebName.Equals(encoding.WebName, StringComparison.OrdinalIgnoreCase))
```
rynowak: I'd also like to see a test for this case

rynowak:
:watch:

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
javiercn: Id' like people that inherit from OutputFormatter to be able to reuse the cache if they need to, however, I've followed your comment and made it private.

In Web API there were a few components that maintained private caches and the fact that those caches were private made it impossible for them to properly derive from the class.

The reason I like it to be protected is because if it's important enough for us to have a cache for these types for performance reasons, it will be good for someone else to be aware of its existence, which is what protected does here.

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
javiercn: I changed the test for the cache to avoid using mocks, this is covered now by that test

rynowak: This is the case where the accept header has a charset is it not? I don't see a test for it. I see a test for the UTF-8 cache optimization.

rynowak:
:watch:

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
javiercn: I added the test and fixed the underlying bug, good catch

javiercn:
@rynowak updated

test/Microsoft.AspNet.Mvc.Core.Test/Formatters/OutputFormatterTests.cs
```diff
@@ -102,6 +103,62 @@ public static IEnumerable<object[]> SelectResponseCharacterEncodingData
...
+            };
+
+            formatter.SupportedMediaTypes.Add("application/json");
+            formatter.SupportedEncodings.Add(Encoding.UTF8);
+
+            // Act
+            formatter.WriteAsync(formatterContext);
+
+            // Assert
+            Assert.Equal(new StringSegment("application/json; charset=utf-8"), formatterContext.ContentType);
```
rynowak: You still haven't added the case that I'm asking for.

ContentType: `"application/json; charset=utf-16"` - supported encoding: UTF-16

Check that the string you get out is the same as the one you put in. You took the time to handle this case specially, so add a test for it :+1:

test/Microsoft.AspNet.Mvc.Core.Test/Formatters/OutputFormatterTests.cs
```diff
@@ -102,6 +103,62 @@ public static IEnumerable<object[]> SelectResponseCharacterEncodingData
...
+                new DefaultHttpContext(),
+                new TestHttpResponseStreamWriterFactory().CreateWriter,
+                objectType: null,
+                @object: null)
+            {
+                ContentType = new StringSegment("application/json; charset=utf-7"),
+            };
+
+            formatter.SupportedMediaTypes.Add("application/json");
+            formatter.SupportedEncodings.Add(Encoding.UTF8);
```
rynowak: This test is a little misleading without adding UTF7 to supported encodings.

rynowak:
:watch:

test/Microsoft.AspNet.Mvc.Core.Test/Formatters/OutputFormatterTests.cs
javiercn: Actually, I need to do it this way to force the code path I'm looking for. Maybe that's an indication that we should rethink the scenario, but I think the scenario is valid, given that formatters are not selected using the charset, once the formatter has been selected, if it doesn't support the charset on the content type, it will pick the first charset it supports.

javiercn:
@rynowak updated

rynowak:
:shipit:

javiercn:
Merged as part of the Part 3 pull request

PR #3864, opened 2016-01-05
[Issue3683 Part 3] Replace stub with real algorithms

javiercn:


javiercn:
@rynowak

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
```diff
@@ -0,0 +1,165 @@
...
+using System.Globalization;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNet.Mvc.Formatters
+{
+    public static class AcceptHeaderParser
+    {
+        private static readonly StringSegment QualityParameter = new StringSegment("q");
+
+        public static IList<MediaTypeSegmentWithQuality> ParseAcceptHeader(IList<string> acceptHeaders)
```
rynowak: Should this be `StringValues`?

rynowak: I guess it's fine to leave it general

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
```diff
@@ -0,0 +1,165 @@
...
+            }
+
+            if (parsedValues == null)
+            {
+                throw new ArgumentNullException(nameof(parsedValues));
+            }
+
+            for (var i = 0; i < acceptHeaders.Count; i++)
+            {
+                var index = 0;
```
rynowak: should this be `charIndex`? or something a little more distinct from `i`?

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
```diff
@@ -0,0 +1,165 @@
...
+                var value = acceptHeaders[i];
+
+                while (!string.IsNullOrEmpty(value) && index < value.Length)
+                {
+                    MediaTypeSegmentWithQuality output;
+                    if (TryParseValue(value, ref index, out output))
+                    {
+                        // The entry may not contain an actual value, like " , "
```
rynowak: This example would be clearer if you wrote it like: `Accept: application/json, , */*`

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
```diff
@@ -0,0 +1,165 @@
...
+                        if (output.MediaType.HasValue)
+                        {
+                            parsedValues.Add(output);
+                        }
+                    }
+                    else
+                    {
+                        throw new FormatException(string.Format(CultureInfo.InvariantCulture, "Invalid values '{0}'.",
+                            value.Substring(index)));
```
rynowak: This is a #YOLObreak - put them each on their own line. 

We need to provide more info here and use a resource.

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
```diff
@@ -0,0 +1,165 @@
...
+
+        private static bool TryParseValue(string value, ref int index, out MediaTypeSegmentWithQuality parsedValue)
+        {
+            parsedValue = default(MediaTypeSegmentWithQuality);
+
+            // If multiple values are supported (i.e. list of values), then accept an empty string: The header may
+            // be added multiple times to the request/response message. E.g.
+            //  Accept: text/xml; q=1
+            //  Accept:
+            //  Accept: text/plain; q=0.2
```
rynowak: Give an example that really shows what you're talking about by `multiple values are supported`. This example is the opposite.

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
```diff
@@ -0,0 +1,165 @@
...
+                index = currentIndex;
+                return true;
+            }
+
+            MediaTypeSegmentWithQuality result;
+            var length = GetMediaTypeWithQualityLength(value, currentIndex, out result);
+
+            if (length == 0)
+            {
+                return false;
```
rynowak: You handle 'empty' in `ParseAcceptHeader` as if this should return true. What's the deal?

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComparisons.cs
```diff
@@ -55,12 +55,59 @@ public static bool IsSubsetOf(StringSegment set, StringSegment subset)
...
-            return MediaTypeHeaderValue.TryParse(set.Value, out setMediaType) &&
-                MediaTypeHeaderValue.TryParse(subset.Value, out subSetMediaType) &&
-                subSetMediaType.IsSubsetOf(setMediaType);
+            var setEnumerator = setParser.GetEnumerator();
+            var subSetEnumerator = subSetParser.GetEnumerator();
+
+            if (!(setEnumerator.MoveNext() && subSetEnumerator.MoveNext()))
+                {
+                    return false;
+                }
```
rynowak: Fix indent

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComponent.cs
```diff
@@ -0,0 +1,138 @@
...
+    {
+        /// <summary>
+        /// The name of the type in a media type when represented as a <see cref="MediaTypeComponent"/>.
+        /// </summary>
+        public static readonly StringSegment Type = new StringSegment("type");
+
+        /// <summary>
+        /// The name of the subtype in a media type when represented as a <see cref="MediaTypeComponent"/>.
+        /// </summary>
+        public static readonly StringSegment Subtype = new StringSegment("subtype");
```
rynowak: I think it's really strange to represent the 'type' and 'subtype' as part of an Enumerator. Every media type has them, why not simplify this and simplify your life?

rynowak: Once you do that, this class becomes basically KeyValuePair<> and just seems superfluous

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeEncoding.cs
```diff
@@ -39,12 +38,21 @@ public static Encoding GetEncoding(string mediaType)
...
+            var parameter = parser.GetParameter("charset");
+
+            if (parameter.HasValue)
+            {
+                return parameter.Value.Value;
+            }
+            else
             {
-                return new StringSegment(parsedMediaType.Charset);
+                return new StringSegment(buffer: null);
```
rynowak: this can just be `new StringSegment()` right?

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeEncoding.cs
```diff
@@ -68,10 +76,25 @@ public static string ReplaceEncoding(string mediaType, Encoding encoding)
         /// <returns>A media type with the replaced encoding.</returns>
         public static string ReplaceEncoding(StringSegment mediaType, Encoding encoding)
         {
-            var parsedMediaType = MediaTypeHeaderValue.Parse(mediaType.Value);
-            parsedMediaType.Encoding = encoding;
+            var charset = GetCharsetParameter(mediaType);
+
+            if (charset != null &&
```
rynowak: It can't be null, it's a value type

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeEncoding.cs
```diff
@@ -68,10 +76,25 @@ public static string ReplaceEncoding(string mediaType, Encoding encoding)
...
+            {
+                return CreateMediaTypeWithEncoding(mediaType, encoding);
+            }
+
+            var charsetOffset = charset.Offset - mediaType.Offset;
+            var restOffset = charsetOffset + charset.Length;
+            var mediaTypeString = mediaType.Substring(0, charsetOffset);
+            var restString = mediaType.Substring(restOffset, mediaType.Length - restOffset);
+
+            return $"{mediaTypeString}{encoding.WebName}{restString}";
```
rynowak: Use a `StringBuilder` and append ranges of the string to it.

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
```diff
@@ -283,21 +283,22 @@ public virtual void WriteResponseHeaders(OutputFormatterWriteContext context)
...
-                {
-                    return OutputMediaTypeCache[mediaType];
-                }
+                throw new ArgumentNullException(nameof(encoding));
+            }
 
-                return MediaTypeEncoding.ReplaceEncoding(mediaType, encoding);
+            if (encoding.WebName.Equals(Encoding.UTF8.WebName, StringComparison.OrdinalIgnoreCase))
+            {
+                return OutputMediaTypeCache[mediaType];
```
rynowak: Should we just make the cache part of `ReplaceEncoding`?

rynowak:
:watch:

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
javiercn: Multiple values refers to the Accept header appearing multiple times on the request response,
Accept: text/xml
Accept: application/json

is equivalent to
Accept: text/xml, application/json

I've changed the comment to clarify that it means multiple Accept headers appearing on the request

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
javiercn: by the time we get here, we know the string is not null or empty, that means it contains something, we tried to parse it and we couldn't, so for that reason we return false (the header contained something like Accept: json or Accept: application/xml, json so they are invalid and we fail parsing

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeEncoding.cs
javiercn: Yes, I can change it if you want, but I feel like new StringSegment(buffer: null) represents the intent more clearly, I don't like to use default constructors in structs because it's indistinguishable that you initialized the type from it being initialized by default.

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
javiercn: Id rather keep it here so that anyone extending OutputFormatter can take advantage of it rather than just ourselves, in its current form it is just an implementation detail.

javiercn:
@rynowak updated

src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs
rynowak: That doesn't really mean anything. This code would call that code. I'm fine with leaving it here, but let's not pretend that it enables :sparkler: **Magic.** :tophat:

src/Microsoft.AspNet.Mvc.Core/ConsumesAttribute.cs
```diff
@@ -68,14 +68,27 @@ public void OnResourceExecuting(ResourceExecutingContext context)
...
+                if (requestContentType != null && !IsSubsetOfAnyContentType(requestContentType))
                 {
                     context.Result = new UnsupportedMediaTypeResult();
                 }
             }
         }
 
+        private bool IsSubsetOfAnyContentType(string requestMediaType)
+        {
+            MediaType parsedRequestMediaType = new MediaType(requestMediaType);
```
rynowak: var

src/Microsoft.AspNet.Mvc.Core/ConsumesAttribute.cs
```diff
@@ -68,14 +68,27 @@ public void OnResourceExecuting(ResourceExecutingContext context)
...
+            for (int i = 0; i < ContentTypes.Count; i++)
```
rynowak: var

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
```diff
@@ -0,0 +1,154 @@
...
+
+            // The accept header may be added multiple times to the request/response message. E.g.
+            // Accept: text/xml; q=1
+            // Accept:
+            // Accept: text/plain; q=0.2
+            // In this case, do not fail parsing in case one of the values is the empty string.
+            if (string.IsNullOrEmpty(value) || (index == value.Length))
+            {
+                return true;
+            }
```
rynowak: This is why I was objecting to the factoring earlier :laughing: - I expected `true` here to mean, 'has value', but it actually means 'is valid'.

src/Microsoft.AspNet.Mvc.Core/Formatters/FormatFilter.cs
```diff
@@ -89,13 +89,28 @@ public void OnResourceExecuting(ResourceExecutingContext context)
...
                 {
                     context.Result = new HttpNotFoundResult();
                 }
             }
         }
 
+        private bool IsSuperSetOfAnySupportedMediaType(string contentType, MediaTypeCollection supportedMediaTypes)
+        {
+            var parsedContentType = new MediaType(contentType);
+            for (int i = 0; i < supportedMediaTypes.Count; i++)
```
rynowak: var

src/Microsoft.AspNet.Mvc.Core/Formatters/HttpTokenParsingRules.cs
```diff
@@ -0,0 +1,273 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics.Contracts;
+using System.Globalization;
+using System.Text;
+
+namespace Microsoft.AspNet.Mvc.Formatters
+{
+    internal static class HttpTokenParsingRules
```
rynowak: Not reviewing this since it's copy-pasta

src/Microsoft.AspNet.Mvc.Core/Formatters/HttpParseResult.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Mvc.Formatters
+{
+    internal enum HttpParseResult
```
rynowak: Not reviewing this since it's copy-pasta

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaType.cs
```diff
@@ -0,0 +1,678 @@
...
+    /// A media type value.
+    /// </summary>
+    public struct MediaType
+    {
+        private static readonly StringSegment QualityParameter = new StringSegment("q");
+
+        private readonly string _mediaType;
+        private readonly int _offset;
+        private readonly int? _length;
+        private Enumerator _parameterEnumerator;
```
rynowak: Why do you cache the enumerator, this is just bloat.

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaType.cs
```diff
@@ -0,0 +1,678 @@
...
+            {
+                return;
+            }
+            Type = enumerator.Current.Value;
+
+            if (!enumerator.MoveNext())
+            {
+                return;
+            }
+            SubType = enumerator.Current.Value;
```
rynowak: I still really don't get this at all. Why is the parameter enumerator doing the parsing of the type/subtype.

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaType.cs
```diff
@@ -0,0 +1,678 @@
...
+        /// <summary>
+        /// Determines whether the current <see cref="MediaType"/> is a subset of the <paramref name="set"/> <see cref="MediaType"/>.
+        /// </summary>
+        /// <param name="set">The set <see cref="MediaType"/>.</param>
+        /// <returns>
+        /// <code>true</code> if this <see cref="MediaType"/> is a subset of <paramref name="set"/>; otherwise<code>false</code>.
+        /// </returns>
+        public bool IsSubsetOf(MediaType set)
+        {
+            return HasValue && set.HasValue &&
```
rynowak: Should we throw here instead of checking HasValue? It's not even a valid question to ask if it's not a valid media type.

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaType.cs
```diff
@@ -0,0 +1,678 @@
...
+            while (enumerator.MoveNext())
+            {
+                if (enumerator.Current.HasName(QualityParameter))
+                {
+                    quality = double.Parse(
+                        enumerator.Current.Value.Value, NumberStyles.AllowDecimalPoint,
+                        NumberFormatInfo.InvariantInfo);
+                }
+            }
+            if (parsedMediaType.HasValue)
```
rynowak: Shouldn't we check this first?

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaType.cs
```diff
@@ -0,0 +1,678 @@
...
+        }
+
+        private struct Enumerator : IEnumerator<MediaTypeComponent>
+        {
+            private ParsingStatus _parsingStatus;
+            private string _mediaType;
+            private int _initialOffset;
+            private int _currentOffset;
+            private int? _length;
+            private MediaTypeComponent _current;
```
rynowak: This is a really big enumerator. It would be nice to trime this down.

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaType.cs
```diff
@@ -0,0 +1,678 @@
...
+                _parsingStatus = ParsingStatus.NotStarted;
+                _mediaType = mediaType;
+                _initialOffset = offset;
+                _length = length;
+                _currentOffset = _initialOffset;
+                _current = default(MediaTypeComponent);
+            }
+
+            /// <inheritdoc />
+            public MediaTypeComponent Current
```
rynowak: Just use KeyValuePair

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComponent.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Internal;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNet.Mvc.Formatters
+{
+}
```
rynowak: ???

src/Microsoft.AspNet.Mvc.Core/Infrastructure/ObjectResultExecutor.cs
```diff
@@ -256,19 +256,11 @@ public virtual Task ExecuteAsync(ActionContext context, ObjectResult result)
...
-                    new StringSegment(parsedHeaders[i].ToString()),
-                    parsedHeaders[i].Quality ?? 1.0));
-            }
-
-            for (var i = 0; i < result.Count; i++)
-            {
-                if (!RespectBrowserAcceptHeader &&
-                    MediaTypeComparisons.MatchesAllTypes(result[i].MediaType) &&
-                    MediaTypeComparisons.MatchesAllSubtypes(result[i].MediaType))
+                var mediaType = new MediaType(result[i].MediaType);
```
rynowak: Since we parse these into `MediaTypeSegmentWithQuality` and then convert them to MediaType, is it worth just adding a quality property to `MediaType` and calling it good?

rynowak: It's an idea, I don't feel super strongly about it

src/Microsoft.AspNet.Mvc.Core/Internal/ResponseContentTypeHelper.cs
```diff
@@ -37,14 +37,16 @@ public static class ResponseContentTypeHelper
         {
             Debug.Assert(defaultContentType != null);
 
-            var defaultContentTypeEncoding = MediaTypeEncoding.GetEncoding(defaultContentType);
+            var parsedDefaultContentType = new MediaType(defaultContentType);
+            var defaultContentTypeEncoding = parsedDefaultContentType.Encoding;
```
rynowak: This pattern seems repeated a lot, should we just have a static?

test/Microsoft.AspNet.Mvc.Core.Test/Formatters/MediaTypeTest.cs
```diff
@@ -0,0 +1,153 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+using Microsoft.Extensions.Primitives;
+using Xunit;
+
+namespace Microsoft.AspNet.Mvc.Formatters
+{
+    public class MediaTypeTests
```
rynowak: `MediaTypeTest`

rynowak:
:watch:

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaType.cs
javiercn: The enumerator does the parsing of the whole thing, it first parses the type, then the subtype, and then each of the parameters in the media type, it's not a parameter enumerator, it's a "media type enumerator"

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaType.cs
javiercn: Id prefer to keep the private struct, it has semantics about how values are compared and makes the signatures way less verbose (MediaTypeComponent) vs KeyValuePair<StringSegment,StringSegment>

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaTypeComponent.cs
javiercn: sorry, left over, will delete

src/Microsoft.AspNet.Mvc.Core/Internal/ResponseContentTypeHelper.cs
javiercn: I'll add the static

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaType.cs
rynowak: But why? This would be much more straightforward if you pulled it out.

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaType.cs
rynowak: Ok, then can we delete the stuff like the equality and hashcode? are those used?

src/Microsoft.AspNet.Mvc.Core/Infrastructure/ObjectResultExecutor.cs
javiercn: MediaType is a heavy struct, if we have a list of MediaType it will show up in the profile in a negative way, so I think it's best to have a separate type to represent just the minimum amount of information that we care about at this stage and use MediaType locally only for doing comparisons (like is subset of, matches all types, etc)

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaType.cs
javiercn: The equality is used as part of IsSubsetOf, Id rather keep all comparison related code within the MediaTypeComponent struct, rather than having to specify how comparisons are made in each call site. What is your concern?

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaType.cs
javiercn: This is not passed around through the stack, nor is MediaType, now we keep any of those two types in a list or something like that, is that your concern? I don't think there's an easy way to trim this down, at least none that I can come up with, do you have a suggestion that doesn't involve heavily reworking the current implementation?

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaType.cs
javiercn: Sorry, I was doing it wrong, i forgot I was working with a struct. I need to check enumerator.ParsingFailed not parsedMediaType.HasValue

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaType.cs
javiercn: It's not, methods like GetParameter don't need to reparse type and subtype before they can search for the parameter, they can start at the right position.

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaType.cs
javiercn: Would it be clear if instead of this being an "Enumerator" was a "State machine"? in the end is what it is. I can make it MediaTypeParsingState instead of the Enumerator, if that clarifies things.

javiercn:
@rynowak updated.

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaType.cs
javiercn: I think both things are correct, what advantage does throwing here provide? IMO it's ok to say that if you don't give me two valid MediaTypes IsSubsetOf is going to be false all the time, in the current usage on our code, it would imply that a formatter doesn't get selected which I think it's the right behavior.

rynowak: > I think both things are correct, what advantage does throwing here provide?

Throwing lets you know if you're in an unexpected case. Generally if you electrify the fences you need to take steps to avoid touching them. This can either be good or bad. Generally throwing is the safer choice if you don't understand 100% of the scenarios.


> IMO it's ok to say that if you don't give me two valid MediaTypes IsSubsetOf is going to be false all the time, in the current usage on our code, it would imply that a formatter doesn't get selected which I think it's the right behavior.

I'm not 100% convinced about this. What you're describing is a silent failure.

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaType.cs
rynowak: No, it's super clear what the code is doing.

This is needlessly complicated by parsing multiple different types of syntax within the same method, which is what **requires** a state machine. If you parse the the Type/SubType in the constructor of MediaType and then only parse parameters in the Enumerator you remove a bunch of coupling and can drop a few of these fields (and that enum).

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaType.cs
rynowak: `This is not passed around through the stack, nor is MediaType`

I don't believe you.

My main concern is still the coupling here between the enumerator and then media-type class.

javiercn: Sorry, I meant to say that this is not passed through the stack, what I mean by this, is that all uses of the MediaType struct are within a method implementation and that we don't pass them as arguments to other methods nor we return them from any method.

As a solution, I can remove the _parser field and just do the parsing from scratch all the time, although I think it's unnecessary and that we should measure to justify it. The enumerator is at this point an implementation detail of the MediaType struct

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaType.cs
javiercn: @rynowak I think that if we allow invalid media types through the system is better that we are just resilient about them, if we were to throw an exception for an invalid media type wouldn't it be better to have it throw it as soon as possible rather than later in the execution?

The implication of returning false on IsSubsetOf with an invalid media type is that you'll get a 406 (haven't verified this, but is my expectation) which I think is totally valid and easy enough for a user/client to discover. The scenarios I can foresee for this are:

Client sends bad media type in accept header. I would expect here Content-Negotiation to fail or just return whatever predetermined format the server has configured, (like json). It's up to the server.

Server sets the content-type to be an invalid media type. If we throw, then we would need to know what to do with the exception or the user would have to handle it, if we don't you would likely get a 406 with the appropriate error message or you would get a predetermined format like json.

javiercn:
@rynowak updated

rynowak:
BTW - by continuing to amend your commit before we're done reviewing, it's causing me to keep going the same code again and again, increasing the chances I'm going to start nitpicking on something I ignored before.

src/Microsoft.AspNet.Mvc.Core/Formatters/AcceptHeaderParser.cs
```diff
@@ -0,0 +1,154 @@
...
+            return current;
+        }
+
+        private static int GetMediaTypeWithQualityLength(
+            string input,
+            int start,
+            out MediaTypeSegmentWithQuality result)
+        {
+
+            result = MediaType.CreateMediaTypeSegmentWithQuality(input, start);
```
rynowak: extra blank line

src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
```diff
@@ -69,10 +69,21 @@ public virtual bool CanRead(InputFormatterContext context)
 
             // Confirm the request's content type is more specific than a media type this formatter supports e.g. OK if
             // client sent "text/plain" data and this formatter supports "text/*".
-            return SupportedMediaTypes.Any(supportedMediaType =>
+            return IsSubsetOfAnySupportedContentType(contentType);
+        }
+
+        private bool IsSubsetOfAnySupportedContentType(string contentType)
+        {
+            MediaType parsedContentType = new MediaType(contentType);
```
rynowak: var

src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
```diff
@@ -69,10 +69,21 @@ public virtual bool CanRead(InputFormatterContext context)
...
+            for (int i = 0; i < SupportedMediaTypes.Count; i++)
```
rynowak: var

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaType.cs
```diff
@@ -0,0 +1,550 @@
...
+            }
+
+            if (length != null && offset + length > mediaType.Length)
+            {
+                throw new ArgumentOutOfRangeException(nameof(length));
+            }
+
+            _parameterParser = default(MediaTypeParameterParser);
+            Type = new StringSegment();
+            SubType = new StringSegment();
```
rynowak: Move these down to the cases where they have an effect

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaType.cs
```diff
@@ -0,0 +1,550 @@
...
+            if (subtypeLength == 0)
+            {
+                return;
+            }
+            else
+            {
+                SubType = subType;
+            }
+
+            _parameterParser = new MediaTypeParameterParser(mediaType, offset + typeLength + subtypeLength, length);
```
rynowak: Why do you need to construct this eagerly and keep it around as a member?

rynowak: is `offset + typeLength + subtypeLength` really correct? Can any whitespace appear in between `Type / SubType`

You could just use `subType.Offset + subType.Length` which should handle all of these concerns provided that it was parsed correctly.

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaType.cs
```diff
@@ -0,0 +1,550 @@
...
+
+            return parameterFound;
+        }
+
+        private struct MediaTypeParameterParser
+        {
+            private string _mediaType;
+            private int? _length;
+
+            public MediaTypeParameterParser(string mediaType, int offset, int? length)
```
rynowak: Should we call `mediaType` `buffer` or something similar? It's really not required to be a single media type, it's a string that contains a media type.

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaType.cs
```diff
@@ -0,0 +1,550 @@
...
+
+            public bool ParseNextParameter(out MediaTypeParameter result)
+            {
+                if (_mediaType == null)
+                {
+                    result = default(MediaTypeParameter);
+                    return false;
+                }
+
+                int parameterLength = GetParameterLength(_mediaType, CurrentOffset, out result);
```
rynowak: var

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaType.cs
```diff
@@ -0,0 +1,550 @@
...
+            {
+                return;
+            }
+            else
+            {
+                Type = type;
+            }
+
+            StringSegment subType;
+            var subtypeLength = GetSubtypeLength(mediaType, offset + typeLength, out subType);
```
rynowak: is it `SubType` or `Subtype`?

rynowak:

src/Microsoft.AspNet.Mvc.Core/Formatters/MediaType.cs
javiercn: I basically need to keep the buffer, the offset and the length stored somewhere, I could add it to MediaType and then pass it back to the parameter parser, but I think it's just easier to construct a new parameter parser instance and keep it in media type itself.

rynowak: Ok that seems fine, what about the other issues?

javiercn: offset + typeLength + subtypeLenght is correct, GetXXXLength method will return how many characters the parser has advanced, including tokens like '/' and trivia like whitespaces, this accounts for situations like application / json which we parse successfully (as we did before this change)

javiercn:
@rynowak upated

rynowak:
:shipit: again

javiercn:
Merged

PR #3865, opened 2016-01-05
Optimize allocations in Script/Link TagHelpers

### LINKED ISSUE #3618: Optimize allocations in Script/Link TagHelpers
image

The `AttributeMatcher` is doing a lot of expensive allocations per-request. Just the `List<>` created here is well over 1% of allocations for the page.

pranavkm:
Fixes #3618


pranavkm:
Lazy creating the `List<>` results in reduction of ~1.4 Mb of allocations (total allocations for the site were 82 Mb to start with):

**Initial**
initial

**Lazy initialization**
lazyloaded-list

An additional change would be to avoid calculating partially matched attributes unless we have logging enabled. It results in a further 1Mb reduction when logging is disabled:

**This goes away**
conditional2

NTaylorMullen:
How expensive are the Partial and PartiallyMatched attribute dictionaries in `ModeMatchResult` [here](https://github.com/aspnet/Mvc/blob/47a3aee2c152b44bb75075e1f7494ba2124c82fa/src/Microsoft.AspNet.Mvc.TagHelpers/Internal/ModeMatchResult.cs)? Looking at the `Link`/`Script` tag helpers now we don't really utilize them at all other than for logging.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -237,7 +237,9 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
             // not function properly.
             Href = output.Attributes[HrefAttributeName]?.Value as string;
 
-            var modeResult = AttributeMatcher.DetermineMode(context, ModeDetails);
+            var determinePartiallyMatchedAttributes = Logger.IsEnabled(
+                ModeMatchResultLoggerExtensions.PartiallyMatchedAttributesLogLevel);
+            var modeResult = AttributeMatcher.DetermineMode(context, ModeDetails, determinePartiallyMatchedAttributes);
             Logger.TagHelperModeMatchResult(modeResult, context.UniqueId, ViewContext.View.Path, this);
```
rynowak: Is this really important to have now that we've finished building these taghelpers? It would be nice to remove all of the code complexity if it's not needed.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
```diff
@@ -23,6 +23,22 @@ public static class AttributeMatcher
...
+        /// <typeparam name="TMode">The type representing the <see cref="ITagHelper" />'s modes.</typeparam>
+        /// <param name="context">The <see cref="TagHelperContext"/>.</param>
+        /// <param name="modeInfos">The modes and their required attributes.</param>
+        /// <param name="determinePartiallyMatchedAttributes">A value that determines if partially matched
+        /// attributes are calculated.</param>
+        /// <returns>The <see cref="ModeMatchResult{TMode}"/>.</returns>
+        public static ModeMatchResult<TMode> DetermineMode<TMode>(
+            TagHelperContext context,
+            IReadOnlyList<ModeAttributes<TMode>> modeInfos,
+            bool determinePartiallyMatchedAttributes)
```
rynowak: This whole thing could really probably be reimagined as a few bitsets right?

rynowak: `modeInfos` provides a definite ordering of the attributes.

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
pranavkm: I'll run it by @DamianEdwards

DamianEdwards: The `Logger` call? That was intended to aid developers when debugging why a Tag Helper didn't run the way they expected, e.g. missing an argument. Without it, it's pretty hard to know why a Tag Helper didn't "work".

rynowak: So assuming that you're using *our complicated attribute thing* and you *happen to look at logging* then this will give you a hint? Why is this better than a runtime exception if we know that what you're doing is wrong?

Think about what the experience would be like if we had time to build the tooling features we wanted here - would this be a compile error? warning? green squiggle? I don't think logging is a great surrogate for the experience that we really want.

I just want to make sure it's a good investment before we invest a lot of time into fixing the perf of this code.

DamianEdwards: Your logic makes me :cry: 

We had said we didn't want Tag Helpers to throw (they're not controls) as a general rule. We were going to use the logging system to raise usage warnings to the dev, but this is as far as we got (we had specced up a logging Tag Helper that would output the warnings to the console to the browser, but it never got done).

I'm OK to remove all this for now, but don't start throwing exceptions, just no-op.

pranavkm: I was looking at an alternative where we would flow `ILogger` to `AttributeMatcher` and collate data if logging is enabled. If the logging is something we care about, I could keep it around (albeit in a messier form).

pranavkm:
Removing the logging in entirety results in an additional ~ 2.5% reduction in allocation from the previous conditional changes: 
conditional3 - string arrays
conditional3 - string arrays - stack

cleanup

pranavkm:
Up dated

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
```diff
@@ -13,16 +13,20 @@ namespace Microsoft.AspNet.Mvc.TagHelpers.Internal
...
         /// </summary>
         /// <typeparam name="TMode">The type representing the <see cref="ITagHelper" />'s modes.</typeparam>
         /// <param name="context">The <see cref="TagHelperContext"/>.</param>
         /// <param name="modeInfos">The modes and their required attributes.</param>
-        /// <returns>The <see cref="ModeMatchResult{TMode}"/>.</returns>
-        public static ModeMatchResult<TMode> DetermineMode<TMode>(
+        /// <param name="compare">A comparer delegate.</param>
+        /// <param name="result">The resulting most effective mode.</param>
+        /// <returns><c>true</c> if a result was found, otherwise <c>false</c>.</returns>
+        public static bool  TryDetermineMode<TMode>(
```
NTaylorMullen: extra space

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
```diff
@@ -13,16 +13,20 @@ namespace Microsoft.AspNet.Mvc.TagHelpers.Internal
...
+        /// all their required attributes present.
         /// </summary>
         /// <typeparam name="TMode">The type representing the <see cref="ITagHelper" />'s modes.</typeparam>
         /// <param name="context">The <see cref="TagHelperContext"/>.</param>
         /// <param name="modeInfos">The modes and their required attributes.</param>
-        /// <returns>The <see cref="ModeMatchResult{TMode}"/>.</returns>
-        public static ModeMatchResult<TMode> DetermineMode<TMode>(
+        /// <param name="compare">A comparer delegate.</param>
+        /// <param name="result">The resulting most effective mode.</param>
+        /// <returns><c>true</c> if a result was found, otherwise <c>false</c>.</returns>
```
NTaylorMullen: `if a mode was determined`

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
```diff
@@ -13,16 +13,20 @@ namespace Microsoft.AspNet.Mvc.TagHelpers.Internal
...
+        public static bool  TryDetermineMode<TMode>(
             TagHelperContext context,
-            IReadOnlyList<ModeAttributes<TMode>> modeInfos)
+            IReadOnlyList<ModeAttributes<TMode>> modeInfos,
+            Func<TMode, TMode, int> compare,
+            out TMode result)
         {
             if (context == null)
```
NTaylorMullen: Should be consistent with the added `compare` param, either no `null` checks or all `null` checks (since they can't be null). At first we had the `null` checks because it was public API for a while I believe.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
```diff
@@ -13,16 +13,20 @@ namespace Microsoft.AspNet.Mvc.TagHelpers.Internal
...
-        /// <returns>The <see cref="ModeMatchResult{TMode}"/>.</returns>
-        public static ModeMatchResult<TMode> DetermineMode<TMode>(
+        /// <param name="compare">A comparer delegate.</param>
+        /// <param name="result">The resulting most effective mode.</param>
+        /// <returns><c>true</c> if a result was found, otherwise <c>false</c>.</returns>
+        public static bool  TryDetermineMode<TMode>(
             TagHelperContext context,
-            IReadOnlyList<ModeAttributes<TMode>> modeInfos)
+            IReadOnlyList<ModeAttributes<TMode>> modeInfos,
+            Func<TMode, TMode, int> compare,
```
NTaylorMullen: Whyyyy C# Whyyyyyyy. I want where `T : enum` darnit!

NTaylorMullen:
:shipit:

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
pranavkm: The alternative was to say `where TMode : IComparable` but that would result in boxing.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/AttributeMatcher.cs
pranavkm: I'll add the null check.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/ModeAttributesOfT.cs
```diff
@@ -9,14 +9,20 @@ namespace Microsoft.AspNet.Mvc.TagHelpers.Internal
     /// <typeparam name="TMode">The type representing the <see cref="AspNet.Razor.TagHelpers.ITagHelper"/>'s mode.</typeparam>
     public class ModeAttributes<TMode>
     {
+        public ModeAttributes(TMode mode, string[] attributes)
```
rynowak: Docs

rynowak:
@pranavkm - where are the list allocations coming from in this code?

pranavkm:
environmenttaghelper

The source of List allocations.

Issue #3825, opened 2015-12-29
Func<object,object> as Model not working anymore


andrew-vandenbrink:
PaperMaterial.cshtml
```
@model Func<object, object>
@{

}
<paper-material>@Model(null)</paper-material>
```

PaperMaterialViewComponent.cs
```
public class PaperMaterialViewComponent : ViewComponent
    {       
        public IViewComponentResult Invoke(Func<object,object> content)
        {
            return View("~/Components/PaperMaterial", content);
        }
    }
```

Calling it this way not working
```
@Component.Invoke("PaperMaterial",
                        @<strong>Is there any way 
to put html or script as parameter within view component?</strong>)
```

This is the exception
![image](https://cloud.githubusercontent.com/assets/12871379/12034995/3c33fa2c-ae38-11e5-819a-9f80383a428d.png)

NTaylorMullen:
2b83dbb52e7fba7612859a8954556475ef947760

PR #3877, opened 2016-01-06
Make `ExpressionRewriter` more resilient to unconvertible types.

NTaylorMullen:
#3825

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Razor/Compilation/ExpressionRewriter.cs
+++ b/src/Microsoft.AspNet.Mvc.Razor/Compilation/ExpressionRewriter.cs
@@ -35,7 +35,7 @@ namespace Microsoft.AspNet.Mvc.Razor.Compilation
         {
             if (IsInsideClass)
             {
-                // Avoid recursing into nested classes. 
+                // Avoid recursing into nested classes.
                 return node;
             }
 
@@ -101,7 +101,12 @@ namespace Microsoft.AspNet.Mvc.Razor.Compilation
             //      }
             //
             var type = SemanticModel.GetTypeInfo(node);
-            if (type.ConvertedType.Name != typeof(Expression).Name &&
+
+            // Due to an anomaly where Roslyn (depending on code sample) may finish compilation without diagnostic
+            // errors (this code path does not execute when diagnostic errors are present) we need to validate that
+            // the ConvertedType was determined/is not null.
+            if (type.ConvertedType == null ||
+                type.ConvertedType.Name != typeof(Expression).Name &&
                 type.ConvertedType.ContainingNamespace.Name != typeof(Expression).Namespace)
             {
                 return node;
@@ -129,7 +134,7 @@ namespace Microsoft.AspNet.Mvc.Razor.Compilation
             SimpleLambdaExpressionSyntax node,
             IdentifierNameSyntax memberAccess)
         {
-            // We want to make the new span 
+            // We want to make the new span
             var originalSpan = node.GetLocation().GetMappedLineSpan();
 
             // Start by collecting all the trivia 'inside' the expression - we need to tack that on the end, but
--- a/test/Microsoft.AspNet.Mvc.Razor.Test/Compilation/ExpressionRewriterTest.cs
+++ b/test/Microsoft.AspNet.Mvc.Razor.Test/Compilation/ExpressionRewriterTest.cs
@@ -4,8 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Linq.Expressions;
-using System.Reflection;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
@@ -18,6 +16,46 @@ namespace Microsoft.AspNet.Mvc.Razor.Compilation
 {
     public class ExpressionRewriterTest
     {
+        [Fact]
+        public void ExpressionRewriter_DoesNotThrowsOnUnknownTypes()
+        {
+            // Arrange
+            var source = @"
+using System.Threading.Tasks;
+using Microsoft.AspNet.Mvc;
+using Microsoft.AspNet.Mvc.Razor;
+
+public class ExamplePage : RazorPage
+{
+    public IViewComponentHelper Component { get; set; }
+
+    public override async Task ExecuteAsync()
+    {
+        Write(
+            await Component.InvokeAsync(
+                ""SomeComponent"",
+                item => new HelperResult((__razor_template_writer) => WriteLiteralTo(__razor_template_writer, ""Hello World""))));
+        }
+    }
+";
+            var tree = CSharpSyntaxTree.ParseText(source);
+
+            // Allow errors here because of an anomaly where Roslyn (depending on code sample) will finish compilation
+            // without diagnostic errors. This test case replicates that scenario by allowing a semantic model with
+            // errors to be visited by the expression rewriter to validate unexpected exceptions aren't thrown.
+            // Error created: "Cannot convert lambda expression to type 'object' because it is not a delegate type."
+            var compilation = Compile(tree, allowErrors: true);
+            var semanticModel = compilation.GetSemanticModel(tree, ignoreAccessibility: true);
+            var rewriter = new ExpressionRewriter(semanticModel);
+            var root = tree.GetRoot();
+
+            // Act
+            var result = rewriter.Visit(root);
+
+            // Assert
+            Assert.True(root.IsEquivalentTo(result));
+        }
+
         [Fact]
         public void ExpressionRewriter_CanRewriteExpression_IdentityExpression()
         {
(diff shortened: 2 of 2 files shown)


test/Microsoft.AspNet.Mvc.Razor.Test/Compilation/ExpressionRewriterTest.cs
```diff
@@ -472,10 +503,13 @@ private CSharpCompilation Compile(SyntaxTree tree)
...
 
-            var diagnostics = compilation.GetDiagnostics();
-            if (diagnostics.Length > 0)
+            if (!allowErrors)
             {
-                Assert.False(true, string.Join(Environment.NewLine, diagnostics));
+                var diagnostics = compilation.GetDiagnostics();
+                if (diagnostics.Length > 0)
+                {
+                    Assert.False(true, string.Join(Environment.NewLine, diagnostics));
```
dougbu: Just `Assert.False(diagnostics.Length == 0, string.Join(Environment.NewLine, diagnostics));`

NTaylorMullen: Didn't touch this code, will change it though.

test/Microsoft.AspNet.Mvc.Razor.Test/Compilation/ExpressionRewriterTest.cs
```diff
@@ -19,6 +19,37 @@ namespace Microsoft.AspNet.Mvc.Razor.Compilation
...
+    {
+        Write(
+            await Component.InvokeAsync(
+                ""SomeComponent"",
+                item => new HelperResult((__razor_template_writer) => WriteLiteralTo(__razor_template_writer, ""Hello World""))));
+        }
+    }
+";
+            var tree = CSharpSyntaxTree.ParseText(source);
+            var compilation = Compile(tree, allowErrors: true);
```
dougbu: Exactly how does the system get into a state where it ignores C# compilation errors? That is, how is the `ExpressionRewriter` reached after compilation fails?

I'm wondering whether there's a more important underlying bug that this PR partially works around.

dougbu: Also deserves a comment about the expected error here.

NTaylorMullen: Ryan and I poked around this yesterday and it's truly an anomaly. When attempting to replicate the user who reported this bugs scenario by fabricating a near similar example you end up running into compilation errors prior to ever hitting the `ExpressionRewriter`. However, attempting the users exact code side by side failed after compile. 

After a significant time investment into trying to determine why Roslyn was parsing/compiling the way it was for the users code but not ours (we tried to duplicate without `ViewComponents`) we threw in the towel and decided making the `ExpressionRewriter` more resilient was an appropriate fix.

Can add a comment explaining this background.

dougbu: :+1:

NTaylorMullen:
:up: :date:

test/Microsoft.AspNet.Mvc.Razor.Test/Compilation/ExpressionRewriterTest.cs
```diff
@@ -19,6 +19,41 @@ namespace Microsoft.AspNet.Mvc.Razor.Compilation
...
+            var tree = CSharpSyntaxTree.ParseText(source);
+
+            // Allow errors here because of an anomaly where Roslyn (depending on code sample) will finish compilation
+            // without diagnostic errors. This test case replicates that scenario by allowing a semantic model with
+            // errors to be visited by the expression rewriter to validate unexpected exceptions aren't thrown.
+            var compilation = Compile(tree, allowErrors: true);
+            var semanticModel = compilation.GetSemanticModel(tree, ignoreAccessibility: true);
+            var rewriter = new ExpressionRewriter(semanticModel);
+
+            // Act - This should not throw, if the type cannot be resolved for the lambda it should no-op.
```
dougbu: Might be improved slightly if it's possible to assert that the rewriter no-oped. Would
``` c#
// still in Arrange
var root = tree.GetRoot();

// Act
var result = rewriter.Visit(root);

// Assert
Assert.Equal(root, result);
```

Put another way, do Linq expressions nodes provide useful `Equals()` implementations? If not is there a solid `SyntaxNodeComparator` available?

test/Microsoft.AspNet.Mvc.Razor.Test/Compilation/ExpressionRewriterTest.cs
```diff
@@ -19,6 +19,41 @@ namespace Microsoft.AspNet.Mvc.Razor.Compilation
...
+                ""SomeComponent"",
+                item => new HelperResult((__razor_template_writer) => WriteLiteralTo(__razor_template_writer, ""Hello World""))));
+        }
+    }
+";
+            var tree = CSharpSyntaxTree.ParseText(source);
+
+            // Allow errors here because of an anomaly where Roslyn (depending on code sample) will finish compilation
+            // without diagnostic errors. This test case replicates that scenario by allowing a semantic model with
+            // errors to be visited by the expression rewriter to validate unexpected exceptions aren't thrown.
```
dougbu: Probably want similar comments in the product code about the Roslyn anomaly.

Curious: What's the actual error?

test/Microsoft.AspNet.Mvc.Razor.Test/Compilation/ExpressionRewriterTest.cs
dougbu: `node.IsEquivalentTo(SyntaxNode)` might do it. Can't see for sure w/o cranking up ILSpy or similar.

dougbu:
:shipit: though I'd prefer a test that confirms no operation.

test/Microsoft.AspNet.Mvc.Razor.Test/Compilation/ExpressionRewriterTest.cs
NTaylorMullen: Looks like I'm able to take the `IsEquivalentTo` approach. :+1:

test/Microsoft.AspNet.Mvc.Razor.Test/Compilation/ExpressionRewriterTest.cs
NTaylorMullen: What happens in the reported users scenario is you get here with 0 diagnostic errors and then you get a null ref explosion when attempting to access `ConvertedType` because it's `null`.

dougbu: Understand that and agree it shouldn't happen. My question was about the error you're causing _here_.

NTaylorMullen: Oh, this is in result to me not being able to 100% replicate the users code (without writing an entire website with view components etc.) that causes the anomaly of the compiler not creating diagnostic errors.

src/Microsoft.AspNet.Mvc.Razor/Compilation/ExpressionRewriter.cs
```diff
@@ -101,7 +101,12 @@ public override SyntaxNode VisitSimpleLambdaExpression(SimpleLambdaExpressionSyn
             //      }
             //
             var type = SemanticModel.GetTypeInfo(node);
-            if (type.ConvertedType.Name != typeof(Expression).Name &&
+
+            // Due to an anomaly where Roslyn (depending on code sample) may finish compilation without diagnostic
+            // errors (this code path does not execute when diagnostic errors are present) we need to validate that
+            // the ConvertedType was determined/is not null.
```
dougbu: Looks good

test/Microsoft.AspNet.Mvc.Razor.Test/Compilation/ExpressionRewriterTest.cs
dougbu: Thanks for the answer but that wasn't what I was looking for. What's the diagnostic error that this test ignores? The general point is the code snippet contains an error but nothing says what that error is. My guess is it's that no class named `SomeComponent` exists but it's only a guess, not something the test makes easy to see.

rynowak:
:shipit:

test/Microsoft.AspNet.Mvc.Razor.Test/Compilation/ExpressionRewriterTest.cs
NTaylorMullen: `Cannot convert lambda expression to type `object` because it is not a delegate type`

dougbu: Excellent. Could you mention that in the test code? Much appreciated.

Issue #750, opened 2014-07-04
Make UrlHelper methods virtual


rynowak:
@kulmugdha moved from aspnet/Routing#80

On the routing blog http://blogs.msdn.com/b/webdev/archive/2014/06/20/asp-net-vnext-routing-overview.aspx, a user expressed that he wants to slightly modify the generated URL on its way out. If we don't make the methods virtual, he has to do either of following. 
1. implement IUrlHelper from scratch
2. Implement IUrlHelper and wrap an instance of UrlHelper in the implementation. 
3. Extend UrlHelper and use Explicit Interface Implementation

None of these is very optimal.

PR #3878, opened 2016-01-06
Modify UrlHelper to use a single StringBuilder

pranavkm:


pranavkm:
The change replaces the virtual `GeneratePathFromRoute` with two overloads that let you modify the generation of the path component and the url in entirety. The goals of https://github.com/aspnet/Mvc/issues/750 aren't affected by this change.

src/Microsoft.AspNet.Mvc.Core/Routing/UrlHelper.cs
```diff
@@ -110,27 +105,63 @@ public virtual string RouteUrl(UrlRouteContext routeContext)
...
 
             var valuesDictionary = new RouteValueDictionary(routeContext.Values);
+            var virtualPathData = GetVirtualPathData(routeContext.RouteName, valuesDictionary);
+            return GenerateUrl(routeContext.Protocol, routeContext.Host, virtualPathData, routeContext.Fragment);
+        }
 
-            var path = GeneratePathFromRoute(routeContext.RouteName, valuesDictionary);
-            if (path == null)
+        /// <inheritdoc />
+        public virtual string Content(string contentPath)
```
dougbu: Hard to see what's changing with this method move in the middle. Could you move it back during the PR and fix it's position when we're done?

dougbu: ... or checkin a quick reordering commit first.

src/Microsoft.AspNet.Mvc.Core/Routing/UrlHelper.cs
```diff
@@ -138,33 +169,60 @@ protected virtual string GeneratePathFromRoute(string routeName, RouteValueDicti
...
+                if (builder.Length == 0 || builder[0] != '/')
+                {
+                    builder.Insert(0, '/');
+                }
+            }
+
+            return builder.ToString();
+        }
+
+        internal void AppendPathAndFragment(StringBuilder builder, VirtualPathData pathData, string fragment)
```
dougbu: Why is this `internal`?

pranavkm:
cc @rynowak

src/Microsoft.AspNet.Mvc.Core/Routing/UrlHelper.cs
pranavkm: Tests.

src/Microsoft.AspNet.Mvc.Core/Routing/UrlHelper.cs
pranavkm: Sure, I can do that

dougbu:
:watch: for more readable diffs

pranavkm:
Updated

src/Microsoft.AspNet.Mvc.Core/Routing/UrlHelper.cs
```diff
@@ -177,12 +162,16 @@ protected virtual string GeneratePathFromRoute(string routeName, RouteValueDicti
                     }
 
                     builder.Append(pathData.VirtualPath);
```
dougbu: Slightly unrelated to your change but the `VirtualPath` handling is doing about the same things in two completely different ways. Suggest making the `if` and `else` blocks more consistent though of course `VirtualPath.Length == 0` isn't quite the same w/ and w/o a `PathBase`.

src/Microsoft.AspNet.Mvc.Core/Routing/UrlHelper.cs
```diff
@@ -213,34 +202,47 @@ public virtual string Link(string routeName, object values)
...
-            Debug.Assert(path != null);
-            var url = path;
-            if (!string.IsNullOrEmpty(fragment))
+            if (pathData == null)
             {
-                url += "#" + fragment;
+                return null;
             }
 
+            // VirtualPathData.VirtualPath returns string.Empty for null.
```
dougbu: Do you mean "... instead of null"?

src/Microsoft.AspNet.Mvc.Core/Routing/UrlHelper.cs
```diff
@@ -110,61 +105,51 @@ public virtual string RouteUrl(UrlRouteContext routeContext)
...
-            if (pathData == null)
-            {
-                return null;
-            }
+            return Router.GetVirtualPath(context);
+        }
 
-            // VirtualPathData.VirtualPath returns string.Empty for null.
-            Debug.Assert(pathData.VirtualPath != null);
+        internal void AppendPathAndFragment(StringBuilder builder, VirtualPathData pathData, string fragment)
```
dougbu: Pls add the usual `// internal for testing` reminder for future generations.

dougbu:
Is there a bug or a perf win behind this PR? I expect some benefit but am not sure whether it's enough for an API change here.

Separately, get approval for the API change. `protected` methods are effectively `public`.

dougbu:
:watch: for slight consistency cleanup

pranavkm:
Updated - @Eilon is ok with the API change.

src/Microsoft.AspNet.Mvc.Core/Routing/UrlHelper.cs
```diff
@@ -213,34 +203,47 @@ public virtual string Link(string routeName, object values)
...
 
-        private string GenerateUrl(string protocol, string host, string path, string fragment)
+        /// <summary>
+        /// Generates the url using the specified components.
+        /// </summary>
+        /// <param name="protocol">The protocol.</param>
+        /// <param name="host">The host.</param>
+        /// <param name="pathData">The <see cref="VirtualPathData"/>.</param>
+        /// <param name="fragment">The URL fragment.</param>
+        /// <returns>The generated url.</returns>
```
rynowak: This should be URL everywhere #EilonPoints

rynowak:
:shipit:

src/Microsoft.AspNet.Mvc.Core/Routing/UrlHelper.cs
dougbu: Thx for making the `if` block consistent w/ the code here.

dougbu:
:shipit:

PR #3884, opened 2016-01-07
Return BadRequest response when antiforgery token validation fails

ajaybhargavb:
Issue - https://github.com/aspnet/Antiforgery/issues/28

@pranavkm @rynowak

### DIFF
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/Logging/ValidateAntiforgeryTokenAuthorizationFilterLoggerExtensions.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/Logging/ValidateAntiforgeryTokenAuthorizationFilterLoggerExtensions.cs
@@ -1,0 +1,26 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNet.Mvc.Logging
+{
+    public static class ValidateAntiforgeryTokenAuthorizationFilterLoggerExtensions
+    {
+        private static readonly Action<ILogger, string, Exception> _antiforgeryTokenInvalid;
+
+        static ValidateAntiforgeryTokenAuthorizationFilterLoggerExtensions()
+        {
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNet.Mvc.ViewFeatures/Logging/ValidateAntiforgeryTokenAuthorizationFilterLoggerExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+namespace Microsoft.AspNet.Mvc.Logging
+{
+    public static class ValidateAntiforgeryTokenAuthorizationFilterLoggerExtensions
+    {
+        private static readonly Action<ILogger, string, Exception> _antiforgeryTokenInvalid;
+
+        static ValidateAntiforgeryTokenAuthorizationFilterLoggerExtensions()
+        {
+            _antiforgeryTokenInvalid = LoggerMessage.Define<string>(
+                LogLevel.Error,
```
ajaybhargavb: Not sure if this should be `Error`.

rynowak: Info IMO. I think that's the level we use for authentication failures.

src/Microsoft.AspNet.Mvc.ViewFeatures/Logging/ValidateAntiforgeryTokenAuthorizationFilterLoggerExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+                1,
+                "Antiforgery token validation failed. {Message}");
+        }
+
+        public static void AntiforgeryTokenInvalid(this ILogger logger, string message, Exception exception)
+        {
+            if (logger.IsEnabled(LogLevel.Error))
```
rynowak: This check isn't needed unless you're doing something complex/expensive in here. Just skip it.

rynowak:
Tests? I'd really expect this to impact existing functional tests. If we don't have unit test coverage, then add it. :watch:

ajaybhargavb:
:up: :date:

rynowak:
:shipit:

PR #3885, opened 2016-01-07
Fix #3501 - Issues with Flush

### LINKED ISSUE #3501: Issues with Flush in our output
### Background

Recall that a `TextWriter` has its own internal buffer. When talking about flushing, there are two phases:
- Flushing the writer's buffer - does a `Write` to the underlying stream
- Flushing the underlying stream

Without doing the former when the writer is disposed, you would *lose* the text that's been encoded, but not written.

But, our writer only does the former when disposed. We don't want to flush the underlying stream because that would cause content to chunk.

However in Razor calling `Flush()` on the page will call flush on the writer and underlying stream - this is **feature** that we provide so that the server can start sending content eagerly by leveraging chunking.

So.
- `Dispose` -> flush buffer
- `Flush` -> flush buffer + flush underlying stream

### The problem
We added a `FlushAsync` call to `ViewExecutor` here: https://github.com/aspnet/Mvc/commit/6a0a24481a07fc34fd44bd410bfa57d7fe9d1c61

The purpose of this change was to make sure that the call to `writer.Dispose` doesn't do a blocking `Write` to flush its internal buffer. This was showing up in profiles as a perf issue for kestrel.

The issue is that this should lead to chunking because we're flushing the underlying stream, instead of just flushing our underlying buffer. We'd like to have a way to flush the buffer, but not the stream through a direct API call, but the problem is that this type should really really be exposed as a `TextWriter`.

There are more places we'd like to apply this optimization, provided that we can do it without flushing the underlying stream.

rynowak:
Calling `Flush[Async]()` on the writer will NOT flush the stream.
Calling `Flush[Async]()` in Razor will flush both the writer and the stream.

Our normal flow will be to flush the writer, but not the stream. This
avoids chunking, but allows us to do a `WriteAsync` on the stream as part of
the call to `FlushAsync`. This is done to avoid a synchronous write due to
`Dispose` calling `Flush` on the writer, which needs to call `Write `on the
stream.

See issue for extensive background.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
@@ -272,7 +272,7 @@ namespace Microsoft.AspNet.Mvc
                 throw new ObjectDisposedException("stream");
             }
 
-            FlushInternal(true, true);
+            FlushInternal();
         }
 
         public override Task FlushAsync()
@@ -282,18 +282,16 @@ namespace Microsoft.AspNet.Mvc
                 throw new ObjectDisposedException("stream");
             }
 
-            return FlushInternalAsync(flushStream: true, flushEncoder: true);
+            return FlushInternalAsync();
         }
 
-        // Do not flush the stream on Dispose, as this will cause response to be
-        // sent in chunked encoding in case of Helios.
         protected override void Dispose(bool disposing)
         {
             if (disposing && _stream != null)
             {
                 try
                 {
-                    FlushInternal(flushStream: false, flushEncoder: true);
+                    FlushInternal();
                 }
                 finally
                 {
@@ -314,7 +312,9 @@ namespace Microsoft.AspNet.Mvc
             }
         }
 
-        private void FlushInternal(bool flushStream = false, bool flushEncoder = false)
+        // Note: our FlushInternal method does NOT flush the underlying stream. This would result in
+        // chunking.
+        private void FlushInternal()
         {
             if (_charBufferCount == 0)
             {
@@ -327,7 +327,7 @@ namespace Microsoft.AspNet.Mvc
                 _charBufferCount,
                 _byteBuffer,
                 0,
-                flushEncoder);
+                flush: true);
 
             if (count > 0)
             {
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
```diff
@@ -335,14 +335,11 @@ private void FlushInternal(bool flushStream = false, bool flushEncoder = false)
...
 
             _charBufferCount = 0;
-
-            if (flushStream)
-            {
-                _stream.Flush();
-            }
         }
 
-        private async Task FlushInternalAsync(bool flushStream = false, bool flushEncoder = false)
```
rynowak: `flushEncoder` is always true, so I just removed it.

src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
```diff
@@ -355,7 +352,7 @@ private void FlushInternal(bool flushStream = false, bool flushEncoder = false)
                 _charBufferCount,
                 _byteBuffer,
                 0,
-                flushEncoder);
+                flush: true);
```
pranavkm: Should this be false?

src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs
```diff
@@ -812,8 +812,8 @@ public Task<HtmlString> RenderSectionAsync(string name, bool required)
         }
 
         /// <summary>
-        /// Invokes <see cref="TextWriter.FlushAsync"/> on <see cref="Output"/> writing out any buffered
-        /// content to the <see cref="HttpResponse.Body"/>.
+        /// Invokes <see cref="TextWriter.FlushAsync"/> on <see cref="Output"/> and <see cref="Stream.FlushAsync"/>
+        /// on the response stream, writing out any buffered content to the <see cref="HttpResponse.Body"/>.
```
pranavkm: on the response body

src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
pranavkm: Oh nevermind. This is the encoder.

pranavkm:
:shipit:

PR #3892, opened 2016-01-08
Fixes #3818 - Support Consumes in ApiExplorer

### LINKED ISSUE #3818: Add support for [Consumes] to ApiDescription
I'm thinking something like this

```
public IList<ApiRequestFormat> SupportedRequestFormats { get; set; }
```

```
public class ApiRequestFormat
{
    public MediaTypeHeaderValue ContentType { get; set; }

    public IInputFormatter Formatter { get; set; }
}
```

rynowak:
This change adds a list of ApiRequestFormat objects to ApiDescription
object which include the content type and formatter for each supported
content type which can be understood by the action.

Computation is aware of the [Consumes] attribute via the
IApiRequestMetadataProvider metadata interface, and aware of Input
Formatters via the new IApiRequestFormatMetadataProvider interface.

This algorithm is essentially the same as what we do for
produces/output-formatters. We iterate the filters and ask them what
content types they think are supported. Then we cross check that list with
the formatters, and ask them which from that list are supported. If no
[Consumes] filters are used, the formatters will include everything they
support by default.

This feature and data is only available when an action has a [FromBody]
parameter, which will naturally exclude actions that handle GET or DELETE
and don't process the body.

### DIFF
--- a/src/Microsoft.AspNet.Mvc.ApiExplorer/ApiDescription.cs
+++ b/src/Microsoft.AspNet.Mvc.ApiExplorer/ApiDescription.cs
@@ -14,47 +14,37 @@ namespace Microsoft.AspNet.Mvc.ApiExplorer
     public class ApiDescription
     {
         /// <summary>
-        /// Creates a new instance of <see cref="ApiDescription"/>.
-        /// </summary>
-        public ApiDescription()
-        {
-            Properties = new Dictionary<object, object>();
-            ParameterDescriptions = new List<ApiParameterDescription>();
-            SupportedResponseFormats = new List<ApiResponseFormat>();
-        }
-
-        /// <summary>
-        /// The <see cref="ActionDescriptor"/> for this api.
+        /// Gets or sets <see cref="ActionDescriptor"/> for this api.
         /// </summary>
         public ActionDescriptor ActionDescriptor { get; set; }
 
         /// <summary>
-        /// The group name for this api.
+        /// Gets or sets group name for this api.
         /// </summary>
         public string GroupName { get; set; }
 
         /// <summary>
-        /// The supported HTTP method for this api, or null if all HTTP methods are supported.
+        /// Gets or sets the supported HTTP method for this api, or null if all HTTP methods are supported.
         /// </summary>
         public string HttpMethod { get; set; }
 
         /// <summary>
-        /// The list of <see cref="ApiParameterDescription"/> for this api.
+        /// Gets a list of <see cref="ApiParameterDescription"/> for this api.
         /// </summary>
-        public IList<ApiParameterDescription> ParameterDescriptions { get; private set; }
+        public IList<ApiParameterDescription> ParameterDescriptions { get; } = new List<ApiParameterDescription>();
 
         /// <summary>
-        /// Stores arbitrary metadata properties associated with the <see cref="ApiDescription"/>.
+        /// Gets arbitrary metadata properties associated with the <see cref="ApiDescription"/>.
         /// </summary>
-        public IDictionary<object, object> Properties { get; private set; }
+        public IDictionary<object, object> Properties { get; } = new Dictionary<object, object>();
 
         /// <summary>
-        /// The relative url path template (relative to application root) for this api.
+        /// Gets or sets relative url path template (relative to application root) for this api.
         /// </summary>
         public string RelativePath { get; set; }
 
         /// <summary>
-        /// The <see cref="ModelMetadata"/> for the <see cref="ResponseType"/> or null.
+        /// Gets or sets <see cref="ModelMetadata"/> for the <see cref="ResponseType"/> or null.
         /// </summary>
         /// <remarks>
         /// Will be null if <see cref="ResponseType"/> is null.
@@ -62,7 +52,7 @@ namespace Microsoft.AspNet.Mvc.ApiExplorer
         public ModelMetadata ResponseModelMetadata { get; set; }
 
         /// <summary>
-        /// The CLR data type of the response or null.
+        /// Gets or sets the CLR data type of the response or null.
         /// </summary>
         /// <remarks>
         /// Will be null if the action returns no response, or if the response type is unclear. Use
@@ -71,12 +61,21 @@ namespace Microsoft.AspNet.Mvc.ApiExplorer
         public Type ResponseType { get; set; }
 
         /// <summary>
-        /// A list of possible formats for a response.
+        /// Gets the list of possible formats for a response.
         /// </summary>
         /// <remarks>
         /// Will be empty if the action returns no response, or if the response type is unclear. Use
         /// <c>ProducesAttribute</c> on an action method to specify a response type.
         /// </remarks>
-        public IList<ApiResponseFormat> SupportedResponseFormats { get; private set; }
+        public IList<ApiRequestFormat> SupportedRequestFormats { get; } = new List<ApiRequestFormat>();
+
+        /// <summary>
+        /// Gets the list of possible formats for a response.
+        /// </summary>
+        /// <remarks>
+        /// Will be empty if the action returns no response, or if the response type is unclear. Use
+        /// <c>ProducesAttribute</c> on an action method to specify a response type.
+        /// </remarks>
+        public IList<ApiResponseFormat> SupportedResponseFormats { get; } = new List<ApiResponseFormat>();
     }
 }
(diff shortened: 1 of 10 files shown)


src/Microsoft.AspNet.Mvc.ApiExplorer/ApiDescription.cs
```diff
@@ -77,6 +69,6 @@ public ApiDescription()
         /// Will be empty if the action returns no response, or if the response type is unclear. Use
         /// <c>ProducesAttribute</c> on an action method to specify a response type.
         /// </remarks>
-        public IList<ApiResponseFormat> SupportedResponseFormats { get; private set; }
+        public IList<ApiResponseFormat> SupportedResponseFormats { get; } = new List<ApiResponseFormat>();
```
rynowak: #PRANAVPOINTS

src/Microsoft.AspNet.Mvc.ApiExplorer/DefaultApiDescriptionProvider.cs
```diff
@@ -302,6 +313,56 @@ private string GetRelativePath(RouteTemplate parsedTemplate)
...
+                                    Formatter = formatter,
+                                    MediaType = supportedType,
+                                });
+                            }
+                        }
+                    }
+                }
+            }
+
+            return results;
```
rynowak: This code is pretty much a mirror of existing `GetResponseFormats`

rynowak:
@javiercn - I'll have to rebase this on top your change once it goes in.

src/Microsoft.AspNet.Mvc.Core/ApiExplorer/IApiRequestMetadataProvider.cs
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNet.Mvc.ApiExplorer
+{
+    /// <summary>
+    /// Provides a a set of possible content types than can be consumed by the action.
```
javiercn: Remove duplicated 'a'

src/Microsoft.AspNet.Mvc.Core/ConsumesAttribute.cs
```diff
@@ -184,5 +189,15 @@ private List<MediaTypeHeaderValue> GetContentTypes(string firstArg, string[] arg
 
             return contentTypes;
         }
+
+        /// <inheritdoc />
+        public void SetContentTypes(IList<MediaTypeHeaderValue> contentTypes)
+        {
+            contentTypes.Clear();
```
javiercn: Is this the right behavior? Won't this mean that 
```C#
[Consumes("application/json")]
[Consumes("application/xml")]
```
will result in ```[Consumes("application/json")]``` being overriden/ignored by ```[Consumes("application/xml")]```?

src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
```diff
@@ -147,5 +148,41 @@ protected Encoding SelectCharacterEncoding(InputFormatterContext context)
...
+            if (contentType == null)
+            {
+                // If contentType is null, then any type we support is valid.
+                return SupportedMediaTypes.Count > 0 ? new List<MediaTypeHeaderValue>(SupportedMediaTypes) : null;
+            }
+            else
+            {
+                List<MediaTypeHeaderValue> mediaTypes = null;
+
+                // Confirm this formatter supports a more specific media type than requested e.g. OK if "text/*"
```
javiercn: This comment reads a bit weird for an InputFormatter, I'm not sure what requested means here, when I think of Input, i don't think of the user "requesting" anything, but "specifying" on the Content-Type, maybe I'm missing something here

src/Microsoft.AspNet.Mvc.ApiExplorer/DefaultApiDescriptionProvider.cs
```diff
@@ -302,6 +313,56 @@ private string GetRelativePath(RouteTemplate parsedTemplate)
             return string.Join("/", segments);
         }
 
+        private IReadOnlyList<ApiRequestFormat> GetRequestFormats(
+            ControllerActionDescriptor action,
+            IApiRequestMetadataProvider[] requestMetadataAttributes,
+            Type type)
+        {
+            var results = new List<ApiRequestFormat>();
```
javiercn: Move the declaration closer to it's first usage

You can also return null instead of the empty list in case there are no results, in other areas of the code we do that and it only adds an if clause to check for null in https://github.com/aspnet/Mvc/pull/3892/files#diff-0e55710732ebb46d94a4706b0b9c24b7R143

test/Microsoft.AspNet.Mvc.ApiExplorer.Test/DefaultApiDescriptionProviderTest.cs
```diff
@@ -470,13 +470,12 @@ public void GetApiDescription_IncludesResponseFormats()
...
             var description = Assert.Single(descriptions);
-            Assert.Equal(4, description.SupportedResponseFormats.Count);
-
-            var formats = description.SupportedResponseFormats;
-            Assert.Single(formats, f => f.MediaType.ToString() == "text/json");
-            Assert.Single(formats, f => f.MediaType.ToString() == "application/json");
-            Assert.Single(formats, f => f.MediaType.ToString() == "text/xml");
-            Assert.Single(formats, f => f.MediaType.ToString() == "application/xml");
+            Assert.Collection(
+                description.SupportedResponseFormats.OrderBy(f => f.MediaType.ToString()),
```
javiercn: Is Order important here? Are you just changing the way we do the assertions here? I personally think it reads better the older way, although it might be a bit more verbose.

src/Microsoft.AspNet.Mvc.Core/ConsumesAttribute.cs
rynowak: `[Consumes]` doesn't allow multiple attributes. `[Consumes("application/json", "application/xml")]` is how you would write it, and gives the desired behavior.

src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs
rynowak: Sure, will update comment. I think it was copy-pasted, as was most of this code.

test/Microsoft.AspNet.Mvc.ApiExplorer.Test/DefaultApiDescriptionProviderTest.cs
rynowak: Order isn't in the product code, just ordering here to make the tests less fragile.

src/Microsoft.AspNet.Mvc.ApiExplorer/DefaultApiDescriptionProvider.cs
rynowak: I'll move this down.

It's going to complicate the code more in both places to return null, and this isn't a perf-sensitive code path.

src/Microsoft.AspNet.Mvc.Core/ApiExplorer/IApiRequestMetadataProvider.cs
rynowak: :+1:

src/Microsoft.AspNet.Mvc.Core/ConsumesAttribute.cs
javiercn: Yeah, I just realized that

src/Microsoft.AspNet.Mvc.ApiExplorer/DefaultApiDescriptionProvider.cs
javiercn: Fair enough, my comment was more due to the fact that in other parts of the code base we return null instead of creating a list (specially on public extensibility points) but I'm happy with returning an empty list to. I'm all for avoiding null checks where possible, but we do it in that way on other parts of the product, hence the comment.

javiercn:
It looks good to me, :shipit:

PR #3894, opened 2016-01-08
Remove use of `String.Split` in Environment taghelper

### LINKED ISSUE #3617: Remove use of `String.Split` in Environment taghelper

pranavkm:
Fixes #3617

### DIFF
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/EnvironmentTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/EnvironmentTagHelper.cs
@@ -2,9 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
 using Microsoft.AspNet.Hosting;
 using Microsoft.AspNet.Razor.TagHelpers;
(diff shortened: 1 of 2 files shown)


pranavkm:
## Before
envtaghelper - before
envtaghelper - before1

## After
envtaghelper - after

This accounts for 670k (~1.2% of allocations) over 3000 requests. This was tested using TagHelperSample.Web

pranavkm:
cc @dougbu \ @NTaylorMullen

src/Microsoft.AspNet.Mvc.TagHelpers/EnvironmentTagHelper.cs
```diff
@@ -74,36 +68,28 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
                 return;
```
dougbu: Not really related to this PR but this `return` is for a weird case. Very odd that `<environment Names="dev">content</environment>` writes out `content` when `EnvironmentName` is `string.Empty` but not when the property has any other value (but `"dev"`).

dougbu:
:shipit:

NTaylorMullen:
:shipit:

PR #3896, opened 2016-01-09
[Prototype] Reduce allocations in GlobbingUrlBuilder

pranavkm:


src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
```diff
@@ -75,61 +82,76 @@ public GlobbingUrlBuilder(IFileProvider fileProvider, IMemoryCache cache, PathSt
...
             {
-                return Enumerable.Empty<string>();
+                return EmptyList;
             }
-
-            var includePatterns = include.Split(PatternSeparator, StringSplitOptions.RemoveEmptyEntries);
-            var excludePatterns = exclude?.Split(PatternSeparator, StringSplitOptions.RemoveEmptyEntries);
-
-            if (includePatterns.Length == 0)
+            else if (Cache == null)
```
pranavkm: In actual code, this is never null. I want to remove it in the next iteration but it causes a bunch of diff in the test. Sending this as a design PR to vet the changes independent of the clean up.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
```diff
@@ -75,61 +82,76 @@ public GlobbingUrlBuilder(IFileProvider fileProvider, IMemoryCache cache, PathSt
         /// <returns>The list of URLs</returns>
         public virtual ICollection<string> BuildUrlList(string staticUrl, string includePattern, string excludePattern)
         {
-            var urls = new HashSet<string>(StringComparer.Ordinal);
```
pranavkm: HashSet has no guarantee of ordering (which we expect from this result). It just happened to keep the order in which items were inserted.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
```diff
@@ -75,61 +82,76 @@ public GlobbingUrlBuilder(IFileProvider fileProvider, IMemoryCache cache, PathSt
         /// <returns>The list of URLs</returns>
         public virtual ICollection<string> BuildUrlList(string staticUrl, string includePattern, string excludePattern)
```
pranavkm: I'll change this to `IList` and touch up any consuming code.

pranavkm:
cc @rynowak

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
```diff
@@ -75,61 +82,76 @@ public GlobbingUrlBuilder(IFileProvider fileProvider, IMemoryCache cache, PathSt
...
-            // Add the statically declared url if present
+            var urls = new List<string>();
             if (staticUrl != null)
             {
                 urls.Add(staticUrl);
             }
 
             // Add urls that match the globbing patterns specified
-            var matchedUrls = ExpandGlobbedUrl(includePattern, excludePattern);
-            urls.UnionWith(matchedUrls);
```
rynowak: I'm worried that you're losing some semantics here.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
```diff
@@ -189,41 +212,80 @@ public int Compare(string x, string y)
...
                 }
 
-                var xSegments = xNoExt.Split(new[] { '/' }, StringSplitOptions.RemoveEmptyEntries);
-                var ySegments = yNoExt.Split(new[] { '/' }, StringSplitOptions.RemoveEmptyEntries);
+                var xNoExt = xExtIndex >= 0 ? x.Substring(0, xExtIndex) : x;
+                var yNoExt = yExtIndex >= 0 ? y.Substring(0, yExtIndex) : y;
 
-                if (xSegments.Length != ySegments.Length)
+                var result = 0;
+                var xTokenizer = new StringTokenizer(xNoExt, '/').GetEnumerator();
```
rynowak: This is really `xEnumerator`

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
```diff
@@ -150,16 +173,16 @@ private IEnumerable<string> FindFiles(string[] includePatterns, string[] exclude
...
-        private string ResolveMatchedPath(FilePatternMatch matchedPath)
-        {
-            // Resolve the path to site root
-            var relativePath = new PathString("/" + matchedPath.Path);
-            return RequestPathBase.Add(relativePath).ToString();
+            matchedUrls.Sort(DefaultPathComparer);
+            return matchedUrls;
         }
 
         private class PathComparer : IComparer<string>
```
rynowak: Everything going on here, looks good (in terms of design).

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
```diff
@@ -75,61 +82,76 @@ public GlobbingUrlBuilder(IFileProvider fileProvider, IMemoryCache cache, PathSt
...
             {
-                var cacheKey = $"{nameof(GlobbingUrlBuilder)}-inc:{include}-exc:{exclude}";
-                IEnumerable<string> files;
+                var cacheKey = new GlobbingUrlKey(include, exclude);
+                List<string> files;
                 if (!Cache.TryGetValue(cacheKey, out files))
                 {
-                    var options = new MemoryCacheEntryOptions();
+                    var includePatterns = include.Split(PatternSeparator, StringSplitOptions.RemoveEmptyEntries);
+                    if (includePatterns.Length == 0)
```
rynowak: Is the idea that we don't care about making this fast because it's cached?

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
```diff
@@ -75,61 +82,76 @@ public GlobbingUrlBuilder(IFileProvider fileProvider, IMemoryCache cache, PathSt
...
+                }
 
-            if (Cache != null)
+                return FindFiles(includePatterns, exclude);
+            }
+            else
             {
-                var cacheKey = $"{nameof(GlobbingUrlBuilder)}-inc:{include}-exc:{exclude}";
-                IEnumerable<string> files;
+                var cacheKey = new GlobbingUrlKey(include, exclude);
```
rynowak: Don't taghelpers provide a per-call-site uniqueId? Should that be our general practice -- use that as our key?

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
```diff
@@ -75,61 +82,76 @@ public GlobbingUrlBuilder(IFileProvider fileProvider, IMemoryCache cache, PathSt
...
                 }
+
                 return files;
             }
-
-            return FindFiles(includePatterns, excludePatterns);
         }
 
-        private IEnumerable<string> FindFiles(string[] includePatterns, string[] excludePatterns)
+        private List<string> FindFiles(string[] includePatterns, string exclude)
```
rynowak: `includePatterns` could probably be a `StringSegment.Enumerator`

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
```diff
@@ -150,16 +173,16 @@ private IEnumerable<string> FindFiles(string[] includePatterns, string[] exclude
             }
 
             var matches = matcher.Execute(_baseGlobbingDirectory);
+            var matchedUrls = new List<string>();
+            foreach (var matchedPath in matches.Files)
+            {
+                // Resolve the path to site root
+                var relativePath = new PathString("/" + matchedPath.Path);
+                matchedUrls.Add(RequestPathBase.Add(relativePath).ToString());
```
rynowak: Should we do `basepath + '/' + matchedPath.Path` instead? This is 2 concats as-written.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
```diff
@@ -150,16 +173,16 @@ private IEnumerable<string> FindFiles(string[] includePatterns, string[] exclude
...
-            return matches.Files.Select(ResolveMatchedPath)
-                .OrderBy(path => path, DefaultPathComparer);
-        }
-
-        private string ResolveMatchedPath(FilePatternMatch matchedPath)
-        {
-            // Resolve the path to site root
-            var relativePath = new PathString("/" + matchedPath.Path);
-            return RequestPathBase.Add(relativePath).ToString();
+            matchedUrls.Sort(DefaultPathComparer);
```
rynowak: I thought we were going to do the binary-search trick?

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
```diff
@@ -75,61 +82,76 @@ public GlobbingUrlBuilder(IFileProvider fileProvider, IMemoryCache cache, PathSt
         /// <returns>The list of URLs</returns>
         public virtual ICollection<string> BuildUrlList(string staticUrl, string includePattern, string excludePattern)
         {
-            var urls = new HashSet<string>(StringComparer.Ordinal);
 
-            // Add the statically declared url if present
+            var urls = new List<string>();
```
rynowak: You could pass this list into `ExpandGlobbedUrl` and save a list allocation

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
```diff
@@ -75,61 +82,76 @@ public GlobbingUrlBuilder(IFileProvider fileProvider, IMemoryCache cache, PathSt
...
+            string lastAddedEntry = null;
+            for (var i = 0; i < globbedUrls.Count; i++)
+            {
+                if (!string.Equals(staticUrl, globbedUrls[i], StringComparison.Ordinal) &&
+                    !string.Equals(lastAddedEntry, globbedUrls[i], StringComparison.Ordinal))
+                {
+                    lastAddedEntry = globbedUrls[i];
+                    urls.Add(lastAddedEntry);
+                }
+            }
```
rynowak: If you pass in `urls` to `ExpandGlobbedUrl` you can just do an insertion-sort.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
```diff
@@ -244,5 +306,25 @@ private static string TrimLeadingTildeSlash(string value)
...
+            public string Include { get; }
+
+            public string Exclude { get; }
+
+            public bool Equals(GlobbingUrlKey other)
+            {
+                return string.Equals(Include, other.Include, StringComparison.Ordinal) &&
+                    string.Equals(Exclude, other.Exclude, StringComparison.Ordinal);
+
+            }
```
rynowak: Do we need to implement hashcode?

pranavkm: Yup.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
pranavkm: `PathString` url encodes url segments as part of ToString(). We'd have to re implement that which seems like more hassle than necessary.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
pranavkm: The unique id doesn't capture the value of patterns which are the things to vary on

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
pranavkm: It made things a bit more gnarly - but I've updated it to remove the String.Split in entirety.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
pranavkm: Ran this by @dougbu, it feels like the code wanted distinct values in an order preserving collection. HashSet happened to preserve the order but this formalizes it.

dougbu: Put another way, @pranavkm fixed a bug in that the code is now reliably doing what the comments said it was doing. He also moved some of the code around but that didn't change the intended semantics.

PR #3897, opened 2016-01-09
Invert TimestampToTicks division

### LINKED ISSUE #3849: Logged action execution time is wrong
Action in RC2 app:

        public IActionResult Test()
        {
            System.Threading.Thread.Sleep(500);
            return Content("Test view");
        }

produces log:
```
info: Microsoft.AspNet.Mvc.Infrastructure.MvcRouteHandler[2]
      Executed action Asp5WebApp.Controllers.HomeController.Test in 0.05ms
info: Microsoft.AspNet.Hosting.Internal.WebApplication[2]
      Request finished in 515ms 200 text/plain; charset=utf-8
```

Issue similar to https://github.com/aspnet/Hosting/issues/478

benaadams:
Division to do the conversion was backwards. Also use const [`TimeSpan.TicksPerSecond`](https://msdn.microsoft.com/en-us/library/system.timespan.tickspersecond(v=vs.110).aspx) rather than numeric 10,000,000.0

Resolves #3849

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/Logging/MvcRouteHandlerLoggerExtensions.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Logging/MvcRouteHandlerLoggerExtensions.cs
@@ -11,7 +11,7 @@ namespace Microsoft.AspNet.Mvc.Logging
 {
     internal static class MvcRouteHandlerLoggerExtensions
     {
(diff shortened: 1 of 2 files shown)


pranavkm:
Merged in https://github.com/aspnet/Mvc/commit/63a7d50dd14d3e747d323fa4422268cbcfcb909d

PR #3902, opened 2016-01-11
Fix incorrect usage of flush on encoder

rynowak:
/cc @pranavkm

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
@@ -143,7 +143,7 @@ namespace Microsoft.AspNet.Mvc
 
             if (_charBufferCount == _charBufferSize)
             {
-                FlushInternal();
+                FlushInternal(flushEncoder: false);
             }
 
             _charBuffer[_charBufferCount] = value;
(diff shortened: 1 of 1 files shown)


pranavkm:
:shipit:

src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
```diff
@@ -282,7 +282,7 @@ public override Task FlushAsync()
                 throw new ObjectDisposedException("stream");
             }
 
-            return FlushInternalAsync();
+            return FlushInternalAsync(flushEncoder: true);
```
dougbu: Why `true` in either `Flush()` or `FlushAsync()`? This writer can certainly be used after either of those calls.

dougbu:
:watch: for one question

src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs
rynowak: >This writer can certainly be used after either of those calls.

That's not really related though.

flushEncoder has to do with the behavior of surrogate pairs.

The encoder buffers the first `char` if it's a surrogate pair in expectation of seeing the second part of the pair. Since we want to write everything buffered when you call `Flush()` it should be true in that case. 

In the other cases, we might have more output, so it's technically more correct to leave it false.

This likely has no impact on anything since we always call flush. I accidentally changed this behavior about a week ago so I'm putting it back to what it was.

dougbu: :ok: thx

dougbu:
:shipit:

PR #3910, opened 2016-01-12
Filter cache

rynowak:
**Before**
image

**After**
image

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Abstractions/Filters/FilterItem.cs
+++ b/src/Microsoft.AspNet.Mvc.Abstractions/Filters/FilterItem.cs
@@ -6,10 +6,19 @@ using System.Diagnostics;
 
 namespace Microsoft.AspNet.Mvc.Filters
 {
-    // Used to flow filters back from the FilterProviderContext
+    /// <summary>
+    /// Used to associate executable filters with <see cref="IFilterMetadata"/> instances
+    /// as part of <see cref="FilterProviderContext"/>. An <see cref="IFilterProvider"/> should
+    /// inspect <see cref="FilterProviderContext.Results"/> and set <see cref="Filter"/> and
+    /// <see cref="IsReusable"/> as appropriate.
+    /// </summary>
     [DebuggerDisplay("FilterItem: {Filter}")]
     public class FilterItem
     {
+        /// <summary>
+        /// Creates a new <see cref="FilterItem"/>.
+        /// </summary>
+        /// <param name="descriptor">The <see cref="FilterDescriptor"/>.</param>
         public FilterItem(FilterDescriptor descriptor)
         {
             if (descriptor == null)
@@ -20,6 +29,11 @@ namespace Microsoft.AspNet.Mvc.Filters
             Descriptor = descriptor;
         }
 
+        /// <summary>
+        /// Creates a new <see cref="FilterItem"/>.
+        /// </summary>
+        /// <param name="descriptor">The <see cref="FilterDescriptor"/>.</param>
+        /// <param name="filter"></param>
         public FilterItem(FilterDescriptor descriptor, IFilterMetadata filter)
             : this(descriptor)
         {
@@ -31,8 +45,19 @@ namespace Microsoft.AspNet.Mvc.Filters
             Filter = filter;
         }
 
-        public FilterDescriptor Descriptor { get; set; }
+        /// <summary>
+        /// Gets the <see cref="FilterDescriptor"/> containing the filter metadata.
+        /// </summary>
+        public FilterDescriptor Descriptor { get; }
 
+        /// <summary>
+        /// Gets or sets the executable <see cref="IFilterMetadata"/> associated with <see cref="Descriptor"/>.
+        /// </summary>
         public IFilterMetadata Filter { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether or not <see cref="Filter"/> can be reused across requests.
+        /// </summary>
+        public bool IsReusable { get; set; }
     }
 }
--- a/src/Microsoft.AspNet.Mvc.Core/Controllers/ControllerActionInvoker.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Controllers/ControllerActionInvoker.cs
@@ -9,6 +9,7 @@ using System.Threading.Tasks;
 using Microsoft.AspNet.Mvc.Core;
 using Microsoft.AspNet.Mvc.Filters;
 using Microsoft.AspNet.Mvc.Formatters;
+using Microsoft.AspNet.Mvc.Internal;
 using Microsoft.AspNet.Mvc.Logging;
 using Microsoft.AspNet.Mvc.ModelBinding;
 using Microsoft.AspNet.Mvc.ModelBinding.Validation;
@@ -25,11 +26,11 @@ namespace Microsoft.AspNet.Mvc.Controllers
 
         public ControllerActionInvoker(
             ActionContext actionContext,
-            IReadOnlyList<IFilterProvider> filterProviders,
+            FilterCache filterCache,
             IControllerFactory controllerFactory,
             ControllerActionDescriptor descriptor,
             IReadOnlyList<IInputFormatter> inputFormatters,
-            IControllerActionArgumentBinder controllerActionArgumentBinder,
+            IControllerActionArgumentBinder argumentBinder,
             IReadOnlyList<IModelBinder> modelBinders,
             IReadOnlyList<IModelValidatorProvider> modelValidatorProviders,
             IReadOnlyList<IValueProviderFactory> valueProviderFactories,
@@ -38,7 +39,7 @@ namespace Microsoft.AspNet.Mvc.Controllers
             int maxModelValidationErrors)
             : base(
                   actionContext,
-                  filterProviders,
+                  filterCache,
                   inputFormatters,
                   modelBinders,
                   modelValidatorProviders,
@@ -47,16 +48,6 @@ namespace Microsoft.AspNet.Mvc.Controllers
                   diagnosticSource,
                   maxModelValidationErrors)
         {
-            if (actionContext == null)
-            {
-                throw new ArgumentNullException(nameof(actionContext));
-            }
-
-            if (filterProviders == null)
-            {
-                throw new ArgumentNullException(nameof(filterProviders));
-            }
-
             if (controllerFactory == null)
             {
                 throw new ArgumentNullException(nameof(controllerFactory));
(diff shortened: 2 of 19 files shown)


src/Microsoft.AspNet.Mvc.Core/Controllers/ControllerActionInvoker.cs
```diff
@@ -67,50 +58,21 @@ public class ControllerActionInvoker : FilterActionInvoker
...
-
-            if (logger == null)
-            {
-                throw new ArgumentNullException(nameof(logger));
-            }
-
-            if (diagnosticSource == null)
-            {
-                throw new ArgumentNullException(nameof(diagnosticSource));
-            }
```
rynowak: These are all validated by the base class.

src/Microsoft.AspNet.Mvc.Abstractions/Filters/FilterItem.cs
```diff
@@ -20,19 +20,10 @@ public FilterItem(FilterDescriptor descriptor)
...
-            : this(descriptor)
-        {
-            if (filter == null)
-            {
-                throw new ArgumentNullException(nameof(filter));
-            }
-
-            Filter = filter;
-        }
-
```
rynowak: Only used in 1 test.

src/Microsoft.AspNet.Mvc.Abstractions/Filters/FilterItem.cs
```diff
@@ -20,19 +20,10 @@ public FilterItem(FilterDescriptor descriptor)
...
-        public FilterDescriptor Descriptor { get; set; }
+        public FilterDescriptor Descriptor { get; }
 
         public IFilterMetadata Filter { get; set; }
+
+        public bool ShouldCache { get; set; }
```
pranavkm: `IsReusable`?

src/Microsoft.AspNet.Mvc.Core/Controllers/ControllerActionInvoker.cs
```diff
@@ -67,50 +58,21 @@ public class ControllerActionInvoker : FilterActionInvoker
...
 
             if (descriptor.MethodInfo == null)
             {
                 throw new ArgumentException(
-                    Resources.FormatPropertyOfTypeCannotBeNull("MethodInfo",
-                                                               typeof(ControllerActionDescriptor)),
+                    Resources.FormatPropertyOfTypeCannotBeNull(
+                        "MethodInfo",
+                        typeof(ControllerActionDescriptor)),
                     "descriptor");
```
pranavkm: `nameof(descriptor)`

src/Microsoft.AspNet.Mvc.Core/Internal/FilterCache.cs
```diff
@@ -0,0 +1,211 @@
...
+            }
+        }
+
+        public IFilterMetadata[] GetFilters(ActionContext actionContext)
+        {
+            var cache = CurrentCache;
+            var actionDescriptor = actionContext.ActionDescriptor;
+
+            CacheEntry entry;
+            if (cache.Entries.TryGetValue(actionDescriptor, out entry))
```
pranavkm: What happens this method is invoked an outdated (old version) `ActionDescriptor`?

src/Microsoft.AspNet.Mvc.Core/Internal/FilterCache.cs
```diff
@@ -0,0 +1,211 @@
...
+
+            for (var i = _filterProviders.Length - 1; i >= 0; i--)
+            {
+                _filterProviders[i].OnProvidersExecuted(context);
+            }
+        }
+
+        private IFilterMetadata[] ExtractFilters(List<FilterItem> items)
+        {
+            var count = 0;
```
pranavkm: Couldn't we use `List<T>.RemoveAll` and cache the predicate?

src/Microsoft.AspNet.Mvc.Core/Internal/FilterCache.cs
rynowak: This is safe because the action descriptors use reference equality. You'll cache values for the 'old' action descriptor, but they won't be associated with the 'new' action descriptor (which would be invalid). 

Effectively, when the ad-cache is updated, we call `Clear()`, but it's just to prevent unbounded growth. If we never cleared the cache it would still be correct.

src/Microsoft.AspNet.Mvc.Core/Internal/FilterCache.cs
```diff
@@ -0,0 +1,211 @@
...
+                return entry.Filters;
+            }
+
+            var items = new List<FilterItem>(entry.Items);
+            for (var i = 0; i < entry.Items.Length; i++)
+            {
+                var item = entry.Items[i];
+                if (item.ShouldCache)
+                {
+                    items.Add(item);
```
pranavkm: This seems weird - isn't this adding to the `List` twice (since it's already been added as part of the ctor?)

src/Microsoft.AspNet.Mvc.Core/Internal/FilterCache.cs
rynowak: We need to turn the `List<FilterItem>` into a `IFilterMetadata`, so there's no value in manipulating the list, it would just be extra work

src/Microsoft.AspNet.Mvc.Core/Internal/FilterCache.cs
rynowak: Yeah, you're right, that should be `entry.Items.Count`

src/Microsoft.AspNet.Mvc.Core/Internal/FilterCache.cs
```diff
@@ -0,0 +1,211 @@
...
+        {
+            _collectionProvider = collectionProvider;
+            _filterProviders = filterProviders.OrderBy(item => item.Order).ToArray();
+        }
+
+        private InnerCache CurrentCache
+        {
+            get
+            {
+                var current = Volatile.Read(ref _currentCache);
```
pranavkm: Is this piece of code equivalent to declaring `_currentCache` as `volatile`?

src/Microsoft.AspNet.Mvc.Core/Internal/FilterCache.cs
```diff
@@ -0,0 +1,211 @@
...
+                return GetFiltersFromEntry(entry, actionContext);
+            }
+
+            var items = new List<FilterItem>(actionDescriptor.FilterDescriptors.Count);
+            for (var i = 0; i < actionDescriptor.FilterDescriptors.Count; i++)
+            {
+                items.Add(new FilterItem(actionDescriptor.FilterDescriptors[i]));
+            }
+
+            entry = new CacheEntry(items);
```
pranavkm: Could we make this part of the `if (allFiltersCached)` condition? It isn't immediately used

src/Microsoft.AspNet.Mvc.Core/Internal/FilterCache.cs
```diff
@@ -0,0 +1,211 @@
...
+
+        private struct CacheEntry
+        {
+            public CacheEntry(IFilterMetadata[] filters)
+            {
+                Filters = filters;
+                Items = null;
+            }
+
+            public CacheEntry(IList<FilterItem> items)
```
pranavkm: This could be a `List<T>`

pranavkm:
:watch:

src/Microsoft.AspNet.Mvc.Core/Internal/FilterCache.cs
rynowak: Yeah, I can switch this if it's going to be easier to grok. I read this http://joeduffyblog.com/2010/12/04/sayonara-volatile/ not too long ago, which lead me to prefer this style. I can see it's causing you pain though.

src/Microsoft.AspNet.Mvc.Core/Internal/FilterCache.cs
rynowak: sure

src/Microsoft.AspNet.Mvc.Abstractions/Filters/FilterItem.cs
rynowak: meh, I think this this is a little clearer about what it does.

src/Microsoft.AspNet.Mvc.Core/Controllers/ControllerActionInvoker.cs
rynowak: Sure

src/Microsoft.AspNet.Mvc.Core/Internal/FilterCache.cs
rynowak: Yeah, this is good. :+1: this is left over from an earlier factoring

rynowak:
Updated

src/Microsoft.AspNet.Mvc.Core/Internal/FilterCache.cs
rynowak: Diff isn't showing it for some reason but this was changed to initialize `items` with a capacity

src/Microsoft.AspNet.Mvc.Abstractions/Filters/FilterItem.cs
pranavkm: It's been used in IHttpHandler and folks might be aware of what it means. Not super vested in this name.

pranavkm: Also docs. @ajaybhargavb has to add dummy docs to everything as part of the cli transition, so lets ease his burden :)

src/Microsoft.AspNet.Mvc.Core/Internal/FilterCache.cs
```diff
@@ -0,0 +1,214 @@
...
+            public InnerCache(int version)
+            {
+                Version = version;
+            }
+
+            public ConcurrentDictionary<ActionDescriptor, CacheEntry> Entries { get; } = 
+                new ConcurrentDictionary<ActionDescriptor, CacheEntry>();
+
+            public int Version { get; }
+
```
pranavkm: Spacing

src/Microsoft.AspNet.Mvc.Cors/CorsAuthorizationFilterFactory.cs
```diff
@@ -34,6 +34,9 @@ public int Order
             }
         }
 
+        /// <inheritdoc />
+        public bool ShouldCache => true;
```
pranavkm: `public bool ShouldCache { get; } = true;` (to be consistent with everything else)

test/Microsoft.AspNet.Mvc.Core.Test/Internal/FilterCacheTest.cs
```diff
@@ -0,0 +1,157 @@
...
+                filters1,
+                f => Assert.NotSame(action.FilterDescriptors[0].Filter, f), // Created by factory
+                f => Assert.Same(action.FilterDescriptors[1].Filter, f)); // Copied by provider
+
+            // Act - 2
+            var filters2 = cache.GetFilters(context);
+
+            Assert.Same(filters1, filters2);
+
+            Assert.Collection(
```
pranavkm: `Assert.Equal(filters1, filters2);` is equivalent but this seems to convey the idea better.

test/Microsoft.AspNet.Mvc.Core.Test/Internal/FilterCacheTest.cs
```diff
@@ -0,0 +1,157 @@
...
+                filters2,
+                f => Assert.Same(filters1[0], f), // Cached
+                f => Assert.Same(filters1[1], f)); // Cached
+        }
+
+        [Fact]
+        public void GetFilters_CachesSomeFilters()
```
pranavkm: `GetFilters_DoesNotCacheFiltersWithShouldCacheFalse` or something along those lines?

pranavkm:
+ Docs for types you edited.

pranavkm:
:shipit:

src/Microsoft.AspNet.Mvc.Cors/CorsAuthorizationFilterFactory.cs
rynowak: Nope, this is always cachable.

pranavkm: https://github.com/aspnet/Mvc/pull/3910/files#diff-c6cea4c77c2eec6adaf1e9469190828eR28

src/Microsoft.AspNet.Mvc.Abstractions/Filters/FilterItem.cs
rynowak: Ok, you convinced me. It's `IsReusable`.

test/Microsoft.AspNet.Mvc.Core.Test/Internal/FilterCacheTest.cs
rynowak: Sure

PR #3914, opened 2016-01-12
Htmlhelper unittests

ajaybhargavb:
Partial fix for issue #453 
Will send another PR for the remaining extensions.
@dougbu

### DIFF
--- a/test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperCheckboxTest.cs
+++ b/test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperCheckboxTest.cs
@@ -649,6 +649,73 @@ namespace Microsoft.AspNetCore.Mvc.Rendering
             Assert.Equal(expected, HtmlContentUtilities.HtmlContentToString(html));
         }
 
+        [Fact]
+        public void Checkbox_UsesSpecifiedExpression()
+        {
+            // Arrange
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper();
+
+            // Act
+            var checkboxResult = helper.CheckBox("Property1");
+
+            // Assert
+            Assert.Equal(
+                "<input id=\"HtmlEncode[[Property1]]\" name=\"HtmlEncode[[Property1]]\" type=\"HtmlEncode[[checkbox]]\" value=\"HtmlEncode[[true]]\" />" +
+                "<input name=\"HtmlEncode[[Property1]]\" type=\"HtmlEncode[[hidden]]\" value=\"HtmlEncode[[false]]\" />",
+                HtmlContentUtilities.HtmlContentToString(checkboxResult));
+        }
+
+        [Fact]
+        public void Checkbox_UsesSpecifiedIsChecked()
+        {
+            // Arrange
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper();
+
+            // Act
+            var checkboxResult = helper.CheckBox("Property1", isChecked: true);
+
+            // Assert
+            Assert.Equal(
+                "<input checked=\"HtmlEncode[[checked]]\" id=\"HtmlEncode[[Property1]]\" name=\"HtmlEncode[[Property1]]\" type=\"HtmlEncode[[checkbox]]\" value=\"HtmlEncode[[true]]\" />" +
+                "<input name=\"HtmlEncode[[Property1]]\" type=\"HtmlEncode[[hidden]]\" value=\"HtmlEncode[[false]]\" />",
+                HtmlContentUtilities.HtmlContentToString(checkboxResult));
+        }
+
+        [Fact]
+        public void Checkbox_UsesSpecifiedIsCheckedRegardlessOfExpressionValue()
+        {
+            // Arrange
+            var metadataProvider = new EmptyModelMetadataProvider();
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper(new ViewDataDictionary<TestModel>(metadataProvider));
+            helper.ViewContext.ClientValidationEnabled = false;
+            helper.ViewData.Model = new TestModel { Property1 = true };
+
+            // Act
+            var checkboxResult = helper.CheckBox("Property1", isChecked: false);
+
+            // Assert
+            Assert.Equal(
+                "<input id=\"HtmlEncode[[Property1]]\" name=\"HtmlEncode[[Property1]]\" type=\"HtmlEncode[[checkbox]]\" value=\"HtmlEncode[[true]]\" />" +
+                "<input name=\"HtmlEncode[[Property1]]\" type=\"HtmlEncode[[hidden]]\" value=\"HtmlEncode[[false]]\" />",
+                HtmlContentUtilities.HtmlContentToString(checkboxResult));
+        }
+
+        [Fact]
+        public void Checkbox_UsesSpecifiedHtmlAttributes()
+        {
+            // Arrange
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper();
+
+            // Act
+            var checkboxResult = helper.CheckBox("Property1", htmlAttributes: new { attr = "value" });
+
+            // Assert
+            Assert.Equal(
+                "<input attr=\"HtmlEncode[[value]]\" id=\"HtmlEncode[[Property1]]\" name=\"HtmlEncode[[Property1]]\" type=\"HtmlEncode[[checkbox]]\" value=\"HtmlEncode[[true]]\" />" +
+                "<input name=\"HtmlEncode[[Property1]]\" type=\"HtmlEncode[[hidden]]\" value=\"HtmlEncode[[false]]\" />",
+                HtmlContentUtilities.HtmlContentToString(checkboxResult));
+        }
+
         private static ViewDataDictionary<TestModel> GetTestModelViewData()
         {
             return new ViewDataDictionary<TestModel>(new EmptyModelMetadataProvider())
--- a/test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperHiddenTest.cs
+++ b/test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperHiddenTest.cs
@@ -835,6 +835,60 @@ namespace Microsoft.AspNetCore.Mvc.Rendering
             Assert.Equal(expected, HtmlContentUtilities.HtmlContentToString(result));
         }
 
+        [Fact]
+        public void Hidden_UsesSpecifiedExpression()
+        {
+            // Arrange
+            var metadataProvider = new EmptyModelMetadataProvider();
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper(new ViewDataDictionary<TestModel>(metadataProvider));
+            helper.ViewContext.ClientValidationEnabled = false;
+            helper.ViewData.Model = new TestModel { Property1 = "propValue" };
+
+            // Act
+            var hiddenResult = helper.Hidden("Property1");
+
+            // Assert
+            Assert.Equal(
+                "<input id=\"HtmlEncode[[Property1]]\" name=\"HtmlEncode[[Property1]]\" type=\"HtmlEncode[[hidden]]\" value=\"HtmlEncode[[propValue]]\" />",
+                HtmlContentUtilities.HtmlContentToString(hiddenResult));
+        }
+
+        [Fact]
+        public void HiddenFor_UsesSpecifiedExpression()
+        {
+            // Arrange
+            var metadataProvider = new EmptyModelMetadataProvider();
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper(new ViewDataDictionary<TestModel>(metadataProvider));
+            helper.ViewContext.ClientValidationEnabled = false;
+            helper.ViewData.Model = new TestModel { Property1 = "propValue" };
+
+            // Act
+            var hiddenForResult = helper.HiddenFor(m => m.Property1);
+
+            // Assert
+            Assert.Equal(
+                "<input id=\"HtmlEncode[[Property1]]\" name=\"HtmlEncode[[Property1]]\" type=\"HtmlEncode[[hidden]]\" value=\"HtmlEncode[[propValue]]\" />",
+                HtmlContentUtilities.HtmlContentToString(hiddenForResult));
+        }
+
+        [Fact]
+        public void Hidden_UsesSpecifiedValue()
+        {
+            // Arrange
+            var metadataProvider = new EmptyModelMetadataProvider();
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper(new ViewDataDictionary<TestModel>(metadataProvider));
+            helper.ViewContext.ClientValidationEnabled = false;
+            helper.ViewData.Model = new TestModel { Property1 = "propValue" };
+
+            // Act
+            var hiddenResult = helper.Hidden("Property1", value: "myvalue");
+
+            // Assert
+            Assert.Equal(
+                "<input id=\"HtmlEncode[[Property1]]\" name=\"HtmlEncode[[Property1]]\" type=\"HtmlEncode[[hidden]]\" value=\"HtmlEncode[[myvalue]]\" />",
+                HtmlContentUtilities.HtmlContentToString(hiddenResult));
+        }
+
         private static ViewDataDictionary<HiddenModel> GetViewDataWithNullModelAndNonNullViewData()
         {
             return new ViewDataDictionary<HiddenModel>(new EmptyModelMetadataProvider())
@@ -898,6 +952,11 @@ namespace Microsoft.AspNetCore.Mvc.Rendering
             public string Property5 { get; set; }
 
             public List<string> Property6 { get; } = new List<string>();
+        }
+
+        private class TestModel
+        {
+            public string Property1 { get; set; }
         }
     }
 }
--- a/test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperLabelExtensionsTest.cs
+++ b/test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperLabelExtensionsTest.cs
@@ -252,6 +252,71 @@ namespace Microsoft.AspNetCore.Mvc.Core
             Assert.Equal("<label for=\"HtmlEncode[[unknownKey]]\">HtmlEncode[[unknownKey]]</label>", HtmlContentUtilities.HtmlContentToString(result));
         }
 
+        [Fact]
+        public void Label_UsesSpecifiedLabelText()
+        {
+            // Arrange
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper();
+
+            // Act
+            var labelResult = helper.Label("Property1", labelText: "Hello");
+
+            // Assert
+            Assert.Equal("<label for=\"HtmlEncode[[Property1]]\">HtmlEncode[[Hello]]</label>", HtmlContentUtilities.HtmlContentToString(labelResult));
+        }
+
+        [Fact]
+        public void LabelFor_UsesSpecifiedLabelText()
+        {
+            // Arrange
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper();
+
+            // Act
+            var labelForResult = helper.LabelFor(m => m.Property1, labelText: "Hello");
+
+            // Assert
+            Assert.Equal("<label for=\"HtmlEncode[[Property1]]\">HtmlEncode[[Hello]]</label>", HtmlContentUtilities.HtmlContentToString(labelForResult));
+        }
+
+        [Fact]
+        public void LabelForModel_UsesSpecifiedLabelText()
+        {
+            // Arrange
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper();
+
+            // Act
+            var labelForModelResult = helper.LabelForModel(labelText: "Hello");
+
+            // Assert
+            Assert.Equal("<label for=\"\">HtmlEncode[[Hello]]</label>", HtmlContentUtilities.HtmlContentToString(labelForModelResult));
+        }
+
+        [Fact]
+        public void LabelFor_DisplaysSpecifiedHtmlAttributes()
+        {
+            // Arrange
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper();
+
+            // Act
+            var labelForResult = helper.LabelFor(m => m.Property1, htmlAttributes: new { attr="value" });
+
+            // Assert
+            Assert.Equal("<label attr=\"HtmlEncode[[value]]\" for=\"HtmlEncode[[Property1]]\">HtmlEncode[[Property1]]</label>", HtmlContentUtilities.HtmlContentToString(labelForResult));
+        }
+
+        [Fact]
+        public void LabelForModel_DisplaysSpecifiedHtmlAttributes()
+        {
+            // Arrange
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper();
+
+            // Act
+            var labelForModelResult = helper.LabelForModel(labelText: "Hello", htmlAttributes: new { attr = "value" });
+
+            // Assert
+            Assert.Equal("<label attr=\"HtmlEncode[[value]]\" for=\"\">HtmlEncode[[Hello]]</label>", HtmlContentUtilities.HtmlContentToString(labelForModelResult));
+        }
+
         private sealed class InnerClass
         {
             public int Id { get; set; }
--- a/test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperPasswordTest.cs
+++ b/test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperPasswordTest.cs
@@ -345,6 +345,60 @@ namespace Microsoft.AspNetCore.Mvc.Rendering
             Assert.Equal(expected, HtmlContentUtilities.HtmlContentToString(result));
         }
 
+        [Fact]
+        public void Password_UsesSpecifiedExpressionForNames_IgnoresExpressionValue()
+        {
+            // Arrange
+            var metadataProvider = new EmptyModelMetadataProvider();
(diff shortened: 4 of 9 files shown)


test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperInputExtensionsTest.cs
```diff
@@ -0,0 +1,324 @@
...
+        }
+
+        [Fact]
+        public void CheckboxHelpers_UsesSpecifiedIsChecked()
+        {
+            // Arrange
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper();
+
+            // Act
+            var checkboxResult = helper.CheckBox("Property1", true);
```
Eilon: Consider named param for this (considering that's the whole purpose of this test).

Eilon:
BTW are these tests ported from the old code, or did you write brand new ones?

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperInputExtensionsTest.cs
```diff
@@ -0,0 +1,324 @@
...
+using Microsoft.AspNet.Mvc.Rendering;
+using Microsoft.AspNet.Mvc.TestCommon;
+using Xunit;
+
+namespace Microsoft.AspNet.Mvc.Core
+{
+    /// <summary>
+    /// Test the <see cref="HtmlHelperInputExtensions" /> class.
+    /// </summary>
+    public class HtmlHelperInputExtensionsTest
```
dougbu: Given the other test classes in the folder, this grouping is quite odd. Why aren't you adding `CheckBoxExtensionsTest`, `HiddenExtensionsTest` and so on?

ajaybhargavb:
I wrote these tests. These tests covered the untested extension methods in `HtmlHelperInputExtensions`

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperInputExtensionsTest.cs
ajaybhargavb: When we discussed yesterday, I thought you said its better to add a corresponding test class for every extension class.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperInputExtensionsTest.cs
```diff
@@ -0,0 +1,324 @@
...
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper();
+
+            // Act
+            var checkboxResult = helper.CheckBox("Property1");
+
+            // Assert
+            Assert.Equal(
+                "<input id=\"HtmlEncode[[Property1]]\" name=\"HtmlEncode[[Property1]]\" type=\"HtmlEncode[[checkbox]]\" value=\"HtmlEncode[[true]]\" />" +
+                "<input name=\"HtmlEncode[[Property1]]\" type=\"HtmlEncode[[hidden]]\" value=\"HtmlEncode[[false]]\" />",
+                HtmlContentUtilities.HtmlContentToString(checkboxResult));
```
dougbu: Not a great approach for testing extension methods. As we discussed offline, we have some tests in this folder following the pattern you're using here. But it's better to clearly separate tests of the actual `IHtmlHelper` and `IHtmlHelper<T>` methods from tests of the extensions.

To do that, suggest you either mock `IHtmlGenerator` (see `HtmlHelperFormExtensionsTest`) or `IHtmlHelper` / `IHtmlHelper<T>` (see `HtmlHelperPartialExtensionsTest`). I prefer the second because it's more straightforward.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperInputExtensionsTest.cs
dougbu: Hmm, meant "extension _method_". See existing `HtmlHelperHiddenTest` and `HtmlHelperPasswordTest` classes.

dougbu:
:watch:

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperInputExtensionsTest.cs
Eilon: If you mock the HTML generator, what would the output be here? I think it's nice in the unit tests to see what the rendered HTML will look like.

dougbu: @Eilon I agree in general but not for these tests. As-is, they aren't unit tests and are instead testing `DefaultHtmlGenerator`, `HtmlHelper`, `HtmlHelper<T>` as well as `HtmlHelperInputExtensions`. So I'd like to see lots of checks of the rendered HTML in unit tests of `HtmlHelper` and `HtmlHelper<T>`.

Could take that to an extreme and mock `IHtmlGenerator` when testing `HtmlHelper`. But we don't have many tests like that today and `IHtmlGenerator` is something of an implementation detail for HTML and tag helpers.

rynowak: I think regardless of philosophy, it's a strong value-add to have tests that cover the exact HTML output by various html helper apis.

In my experience most changes that impact the HTML output are only caught in functional tests and it's hard to investigate.

dougbu: I'm agreeing on the value of testing to confirm the generated HTML. But I would rather that be done in tests that call `IHtmlHelper` methods directly. Here it just gets redundant.

Eilon: The tests will be too mechanical to provide any real value. The tests as they are now will catch real bugs. Let's stick with this approach.

ajaybhargavb:
Updated.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperCheckboxTest.cs
```diff
@@ -649,6 +649,54 @@ public void CheckBoxFor_WithComplexExpressions_DoesNotUseValuesFromViewDataDicti
             Assert.Equal(expected, HtmlContentUtilities.HtmlContentToString(html));
         }
 
+        [Fact]
+        public void CheckboxHelpers_UsesSpecifiedExpression()
```
dougbu: Careful of that 's' -- `CheckboxHelpers_Use...` or `CheckboxHelper_Uses...`.

dougbu: ... throughout the PR.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperHiddenTest.cs
```diff
@@ -835,6 +835,40 @@ public void HiddenFor_DoesNotUseAttributeValue()
...
 
+        [Fact]
+        public void HiddenHelpers_UsesSpecifiedExpression()
+        {
+            // Arrange
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper();
+
+            // Act
+            var hiddenResult = helper.Hidden("Property1");
+            var hiddenForResult = helper.HiddenFor(m => m.Property1);
```
dougbu: Not a fan of doing two independent tests in the same `[Fact]` though we have some ugly examples in these HTML helper tests.

Split into separate methods. Or (perhaps) pass `Func<IHtmlHelper<ForgetTheType>, IHtmlContent>` into a `[Theory]` (since the content matches here).

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperLabelExtensionsTest.cs
```diff
@@ -252,6 +252,38 @@ public void LabelFor_DisplaysVariableName()
...
+        [Fact]
+        public void LabelHelpers_UsesSpecifiedLabelText()
+        {
+            // Arrange
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper();
+
+            // Act
+            var labelResult = helper.Label("Property1", labelText: "Hello");
+            var labelForResult = helper.LabelFor(m => m.Property1, labelText: "Hello");
+            var labelForModelResult = helper.LabelForModel(labelText: "Hello");
```
dougbu: Split up.

dougbu: Also this last one is testing something different. Other calls get `<label>` for the `Property1` method but this gets the `<label>` for the entire model. The second is a rather special case since, in many cases, the top-level model cannot be used in HTML helpers.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperLabelExtensionsTest.cs
```diff
@@ -252,6 +252,38 @@ public void LabelFor_DisplaysVariableName()
...
+
+        [Fact]
+        public void LabelHelpers_DisplaysSpecifiedHtmlAttributes()
+        {
+            // Arrange
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper();
+
+            // Act
+            var labelForResult = helper.LabelFor(m => m.Property1, htmlAttributes: new { attr="value" });
+            var labelForModelResult = helper.LabelForModel(labelText: "Hello", htmlAttributes: new { attr="value" });
```
dougbu: Split these two different scenarios. Again, `LabelFor(m => m)` is the call that's analogous to `LabelForModel()` but both are special cases unless you're in a partial, `EditorFor()`, or similar.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperPasswordTest.cs
```diff
@@ -345,6 +345,40 @@ public static TheoryData PasswordFor_WithComplexExpressionsData
...
 
+        [Fact]
+        public void PasswordHelpers_UsesSpecifiedExpression()
+        {
+            // Arrange
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper();
+
+            // Act
+            var passwordResult = helper.Password("Property1");
+            var passwordForResult = helper.PasswordFor(m => m.Property1);
```
dougbu: `[Theory]` or two `[Fact]`s. (`[Theory]` would be :ok: here because expressions work out to the same thing and output is identical.)

dougbu: ... same for remaining tests. I don't see more that test the `...ForModel()` extensions.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperRadioButtonExtensionsTest.cs
```diff
@@ -0,0 +1,68 @@
...
+
+using Microsoft.AspNet.Mvc.Rendering;
+using Microsoft.AspNet.Mvc.TestCommon;
+using Xunit;
+
+namespace Microsoft.AspNet.Mvc.Core
+{
+    /// <summary>
+    /// Test the RadioButton extensions in <see cref="HtmlHelperInputExtensions" /> class.
+    /// </summary>
```
dougbu: Fine but don't think doc comments are at all required in a regular test class.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperRadioButtonExtensionsTest.cs
```diff
@@ -0,0 +1,68 @@
...
+    public class HtmlHelperRadioButtonExtensionsTest
+    {
+        [Fact]
+        public void RadioButtonHelpers_UsesSpecifiedExpression()
```
dougbu: This test doesn't demonstrate what you've got on the tin. Here, you're showing that `checked` isn't generated when `Property1` is something (anything) other than `"myvalue"`. Also need a test showing that `checked` _is_ generated when `Property1 == "myvalue"`. And it's also worth confirming whether the conditions are case-sensitive i.e. test what happens when `Property1 == "MyValue"`.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperHiddenTest.cs
```diff
@@ -835,6 +835,40 @@ public void HiddenFor_DoesNotUseAttributeValue()
             Assert.Equal(expected, HtmlContentUtilities.HtmlContentToString(result));
         }
 
+        [Fact]
+        public void HiddenHelpers_UsesSpecifiedExpression()
```
dougbu: `Property1` needs to be non-`null` to show anything interesting here.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperCheckboxTest.cs
```diff
@@ -649,6 +649,54 @@ public void CheckBoxFor_WithComplexExpressions_DoesNotUseValuesFromViewDataDicti
...
+        {
+            // Arrange
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper();
+
+            // Act
+            var checkboxResult = helper.CheckBox("Property1");
```
dougbu: Similar to `RadioButtonHelpers_UsesSpecifiedExpression()`, must demonstrate what happens when `Property1 == "false"`, `Property1 == "FALSE"`, `Property1 = "true", and `Property1 == "Not a bool"`. Last case may already be handled (see a test using `Assert.Throws<FormatException>` below).

Suspect tests exist already covering `bool` properties. Double-check please.

I believe `bool` property

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperRadioButtonExtensionsTest.cs
```diff
@@ -0,0 +1,68 @@
...
+            Assert.Equal(
+                "<input id=\"HtmlEncode[[Property1]]\" name=\"HtmlEncode[[Property1]]\" type=\"HtmlEncode[[radio]]\" value=\"HtmlEncode[[myvalue]]\" />",
+                HtmlContentUtilities.HtmlContentToString(radioButtonResult));
+            Assert.Equal(
+                "<input id=\"HtmlEncode[[Property1]]\" name=\"HtmlEncode[[Property1]]\" type=\"HtmlEncode[[radio]]\" value=\"HtmlEncode[[myvalue]]\" />",
+                HtmlContentUtilities.HtmlContentToString(radioButtonForResult));
+        }
+
+        [Fact]
+        public void RadioButtonHelpers_UsesSpecifiedIsChecked()
```
dougbu: Test the opposite case as well: `Property1 == "true"` with 'isChecked: false'.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperCheckboxTest.cs
```diff
@@ -649,6 +649,54 @@ public void CheckBoxFor_WithComplexExpressions_DoesNotUseValuesFromViewDataDicti
...
+        }
+
+        [Fact]
+        public void CheckboxHelpers_UsesSpecifiedIsChecked()
+        {
+            // Arrange
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper();
+
+            // Act
+            var checkboxResult = helper.CheckBox("Property1", isChecked: true);
```
dougbu: Test the opposite case as well: `Property1 == "true"` with 'isChecked: false'.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperPasswordTest.cs
```diff
@@ -345,6 +345,40 @@ public static TheoryData PasswordFor_WithComplexExpressionsData
             Assert.Equal(expected, HtmlContentUtilities.HtmlContentToString(result));
         }
 
+        [Fact]
+        public void PasswordHelpers_UsesSpecifiedExpression()
```
dougbu: No, `Password()` and `PasswordFor()` ignore the expression value, `ViewData`, `ModelState`, and so on. _Only_ the passed `htmlAttributes` or `value` arguments can results in a generated `value` attribute. (And in the `PasswordFor()` case, there's no `value` argument.)

Need to change this test name and test a non-`null` `Property1`

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperTextAreaExtensionsTest.cs
```diff
@@ -0,0 +1,87 @@
...
+
+namespace Microsoft.AspNet.Mvc.Core
+{
+    /// <summary>
+    /// Test the TextArea extensions in <see cref="HtmlHelperInputExtensions" /> class.
+    /// </summary>
+    public class HtmlHelperTextAreaExtensionsTest
+    {
+        [Fact]
+        public void TextAreaHelpers_UsesSpecifiedExpression()
```
dougbu: Need a non-`null` value to show that it's used. Check everywhere I didn't already make this comment too.

dougbu:
:watch:

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperCheckboxTest.cs
ajaybhargavb: There are existing checkbox tests that test all scenarios that you mentioned above (not just the `Assert.Throws` case). I'll add these cases for `RadioButton`.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperRadioButtonExtensionsTest.cs
ajaybhargavb: Added.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperCheckboxTest.cs
ajaybhargavb: Added.

ajaybhargavb:
Updated.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperPasswordTest.cs
```diff
@@ -345,6 +345,60 @@ public static TheoryData PasswordFor_WithComplexExpressionsData
             Assert.Equal(expected, HtmlContentUtilities.HtmlContentToString(result));
         }
 
+        [Fact]
+        public void Password_UsesSpecifiedExpressionButIgnoresValue()
```
dougbu: `Password_UsesSpecifiedExpressionForNames_IgnoresExpressionValue()`. Same for similar test below.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperRadioButtonExtensionsTest.cs
```diff
@@ -0,0 +1,157 @@
...
+            // Assert
+            Assert.Equal(
+                "<input id=\"HtmlEncode[[Property1]]\" name=\"HtmlEncode[[Property1]]\" type=\"HtmlEncode[[radio]]\" value=\"HtmlEncode[[myvalue]]\" />",
+                HtmlContentUtilities.HtmlContentToString(radioButtonForResult));
+        }
+
+        [Theory]
+        [InlineData("MyValue")]
+        [InlineData("myvalue")]
+        public void RadioButton_UsesSpecifiedExpressionCheckedWhenValueMatches(string value)
```
dougbu: `RadioButtion_CheckedWhenValueMatchesSpecifiedExpression` and same for similar test below.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperTextBoxExtensionsTest.cs
```diff
@@ -0,0 +1,130 @@
...
+        public void TextBox_UsesSpecifiedFormat()
+        {
+            // Arrange
+            var metadataProvider = new EmptyModelMetadataProvider();
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper(new ViewDataDictionary<TestModel>(metadataProvider));
+            helper.ViewContext.ClientValidationEnabled = false;
+            helper.ViewData.Model = new TestModel { Property1 = "propValue" };
+
+            // Act
+            var textBoxResult = helper.TextBox("Property1", value: "myvalue", format: "prefix: {0}");
```
dougbu: Add a similar test with same extension method but `value: null`.

dougbu:
Is this PR intended to be the last word on #453? If yes, must add more tests covering open items mentioned in that issue. I'm fine doing that separately (hmm, prefer it) but don't want to see the issue close immediately.

- Indexers and deeply nested lambdas for the various `*For` e.g `InputFor`, `DisplayFor` etc
- Scenarios gap compared to 5.2
- a few other methods are exposed publicly but used in other helpers (and therefore tested indirectly, or will be):
 - `@Html.Encode()`
 - `@Html.FormatValue()`
 - `@Html.GenerateIdFromName()`
 - `@Html.GetClientValidationRules()`
 - `@Html.GetFormMethodString()`
- Some details on remaining `DropDownList` and `ListBox` test gaps.  Test do not currently:
 - use non-`null` `optionLabel` and `htmlAttributes` arguments
 - confirm value sources and their priorities
 - mock an `IHtmlGenerator` and confirm how it is called
 - mock an `IHtmlHelper` and confirm how extension methods call that

My point in the second-last bullet ([here](https://github.com/aspnet/Mvc/issues/453#issuecomment-70395759)) was to write a few (not too many) tests to cover usage that isn't already handled indirectly.

"confirm value sources and their priorities" is important in general and a big part of "Scenarios gap compared to 5.2". Other than expression vs. explicit values, that isn't covered in any of the new test classes. Need to show `ModelStateDictionary` values, `ViewDataDictionary` values, and their precedence. (Or, in the case of, `Password[For]()` those values being ignored.) Have a look at tests like `HiddenUsesValuesFromModelState_OverExplicitSpecifiedValueAndPropertyValue()` in `HtmlHelperHiddenTest`.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperCheckboxTest.cs
```diff
@@ -649,6 +649,73 @@ public void CheckBoxFor_WithComplexExpressions_DoesNotUseValuesFromViewDataDicti
...
+        }
+
+        [Fact]
+        public void Checkbox_UsesSpecifiedHtmlAttributes()
+        {
+            // Arrange
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper();
+
+            // Act
+            var checkboxResult = helper.CheckBox("Property1", htmlAttributes: new { attr = "value" });
```
dougbu: Something more to track for another PR: `htmlAttributes` is another "value source". Not sure even the existing test classes cover including `checked`, `id`, `name`, `type` or `value` properties in the `htmlAttributes` object.

FYI the generated HTML might override a `name` and `value` (when the `value` argument is non-`null`) but `htmlAttributes` should "win" in general.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperValidationMessageExtensionsTest.cs
```diff
@@ -0,0 +1,135 @@
...
+using Microsoft.AspNet.Mvc.Rendering;
+using Microsoft.AspNet.Mvc.TestCommon;
+using Xunit;
+
+namespace Microsoft.AspNet.Mvc.Core
+{
+    /// <summary>
+    /// Test the ValidationMessage extensions in <see cref="HtmlHelperValidationExtensions" /> class.
+    /// </summary>
+    public class HtmlHelperValidationMessageExtensionsTest
```
dougbu: For another PR: This test class covers all extension methods but only when the property is valid and client-side validation is enabled. Need to confirm what happens for all combinations of those two characteristics e.g. invalid properties cause a `<span>` to be generated even if client-side validation is disabled.

dougbu:
:watch: just for one more test and a couple of renames.

ajaybhargavb:
@dougbu, no. My intention was to not close the issue.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperTextBoxExtensionsTest.cs
ajaybhargavb: Added.

ajaybhargavb:
Updated.

dougbu:
:shipit:

Please also update #453 to make it clear what work remains. Cross out the done bits in existing comments and add another comment with a "still to come" list (likely similar to my 3 comments above though I may have missed something). I don't recommend just linking to my comments above because that buries the information.

PR #3919, opened 2016-01-13
 * StringOutput set proper ContentType

### LINKED ISSUE #3692: ASP.NET MVC returns Content-Type text/plain even if application/xml was requested in case of return type is string
This is my controller:

```c#
[Route("")]
public class ValuesController : Controller
{
    //This always returns text/plain
    [HttpPost("/text")]
    [Produces("application/xml", "application/json")]
    public string GetText() => "Some text";

    [HttpPost("/textarray")]
    [Produces("application/xml", "application/json")]
    public IEnumerable<string> GetTextarray() => new[] { "Some text", "Some text2" };
}
```

GetText() always returns text/plain even if I specify "Accept: application/xml" in request header. This happens only in case of return type "string".  I expect get xml instead of plain text. See sample requests below:

Request 1:
GET /text HTTP/1.1
Host: localhost:5004
Accept: application/xml

Response 1:
HTTP/1.1 200 OK
Date: Fri, 04 Dec 2015 20:13:00 GMT
Content-Type: text/plain; charset=utf-8
Server: Kestrel

Some text

Request 2:
GET /textarray HTTP/1.1
Host: localhost:5004
Accept: application/xml

Response 2:
HTTP/1.1 200 OK
Date: Fri, 04 Dec 2015 20:13:04 GMT
Content-Type: application/json; charset=utf-8
Server: Kestrel

```xml
<ArrayOfString xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema"><string>Some text</string><string>Some text2</string></ArrayOfString>
```

ryanbrandenburg:
Fixes #3692

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/Formatters/StringOutputFormatter.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Formatters/StringOutputFormatter.cs
@@ -7,7 +7,6 @@ using System.Threading.Tasks;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Mvc.Internal;
 using Microsoft.Extensions.Primitives;
-using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNet.Mvc.Formatters
 {
@@ -34,7 +33,13 @@ namespace Microsoft.AspNet.Mvc.Formatters
             // always return it as a text/plain format.
             if (context.ObjectType == typeof(string) || context.Object is string)
             {
-                context.ContentType = new StringSegment(SupportedMediaTypes[0]);
+                if (!context.ContentType.HasValue)
+                {
+                    var mediaType = SupportedMediaTypes[0];
+                    var encoding = SupportedEncodings[0];
+                    context.ContentType = new StringSegment(MediaType.ReplaceEncoding(mediaType, encoding));
+                }
+
                 return true;
             }
 
--- a/test/Microsoft.AspNet.Mvc.Core.Test/Formatters/StringOutputFormatterTests.cs
+++ b/test/Microsoft.AspNet.Mvc.Core.Test/Formatters/StringOutputFormatterTests.cs
@@ -27,6 +27,48 @@ namespace Microsoft.AspNet.Mvc.Formatters
             }
         }
 
+        [Fact]
+        public void CanWriteResult_SetsAcceptContentType()
+        {
+            // Arrange
+            var formatter = new StringOutputFormatter();
+            var expectedContentType = new StringSegment("application/json");
+
+            var context = new OutputFormatterWriteContext(
+                new DefaultHttpContext(),
+                new TestHttpResponseStreamWriterFactory().CreateWriter,
+                typeof(string),
+                "Thisisastring");
+            context.ContentType = expectedContentType;
+
+            // Act
+            var result = formatter.CanWriteResult(context);
+
+            // Assert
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNet.Mvc.Core/Formatters/StringOutputFormatter.cs
```diff
@@ -34,8 +34,12 @@ public override bool CanWriteResult(OutputFormatterCanWriteContext context)
             // always return it as a text/plain format.
             if (context.ObjectType == typeof(string) || context.Object is string)
             {
-                context.ContentType = new StringSegment(SupportedMediaTypes[0]);
-                return true;
+                if (context.ContentType == null)
```
rynowak: `StringSegment`s can't be null.

rynowak:
Tests?

rynowak:
:watch:

ryanbrandenburg:
:up: :date:

src/Microsoft.AspNet.Mvc.Core/Formatters/StringOutputFormatter.cs
```diff
@@ -34,7 +33,11 @@ public override bool CanWriteResult(OutputFormatterCanWriteContext context)
             // always return it as a text/plain format.
             if (context.ObjectType == typeof(string) || context.Object is string)
             {
-                context.ContentType = new StringSegment(SupportedMediaTypes[0]);
+                if (!context.ContentType.HasValue)
+                {
+                    context.ContentType = new StringSegment(SupportedMediaTypes[0]);
```
rynowak: Ideally we'd also include an charset here. Look at what the other formatters do

test/Microsoft.AspNet.Mvc.Core.Test/Formatters/StringOutputFormatterTests.cs
```diff
@@ -35,9 +35,7 @@ public static IEnumerable<object[]> OutputFormatterContextValues
             bool expectedCanWriteResult)
         {
             // Arrange
-            var expectedContentType = expectedCanWriteResult ?
-                new StringSegment("text/plain") :
-                new StringSegment("application/json");
+            var expectedContentType = new StringSegment("application/json");
 
             var formatter = new StringOutputFormatter();
             var type = useDeclaredTypeAsString ? typeof(string) : typeof(object);
```
rynowak: Where's the unit test for the new behavior?

rynowak:
:watch:

ryanbrandenburg:
:up: :date: Added unit tests and set charset on default ContentType.

src/Microsoft.AspNet.Mvc.Core/Formatters/StringOutputFormatter.cs
```diff
@@ -20,7 +19,7 @@ public StringOutputFormatter()
         {
             SupportedEncodings.Add(Encoding.UTF8);
             SupportedEncodings.Add(Encoding.Unicode);
-            SupportedMediaTypes.Add("text/plain");
+            SupportedMediaTypes.Add("text/plain;charset=utf-8");
```
rynowak: This isn't right. Look at what the other formatters do.

ryanbrandenburg:
:up: :date:

rynowak:
:shipit:

PR #3920, opened 2016-01-13
Buffer IHtmlContent instead of writing synchronously

rynowak:


test/Microsoft.AspNet.Mvc.ViewFeatures.Test/TestViewBufferScope.cs
```diff
@@ -14,5 +17,11 @@ public TestViewBufferScope(int bufferSize = DefaultBufferSize)
         }
 
         public ViewBufferValue[] GetSegment() => new ViewBufferValue[_bufferSize];
+
+
```
dougbu: needs less newline-iness

src/Microsoft.AspNet.Mvc.Razor/RazorTextWriter.cs
```diff
@@ -99,15 +99,7 @@ public override void Write(string value)
...
-            var htmlTextWriter = TargetWriter as HtmlTextWriter;
-            if (htmlTextWriter == null)
-            {
-                value.WriteTo(TargetWriter, HtmlEncoder);
-            }
-            else
-            {
-                htmlTextWriter.Write(value);
-            }
+            value.WriteTo(TargetWriter, HtmlEncoder);
```
rynowak: We need to remove this optimization and put it back into the places where it's valuable. 

The reason is that often see a value show up here like: `TagHelperOutput`->`TagHelperContent`->`ViewBuffer`. We want 'flatten' this hierarchy, but without any hardcoding to types like `TagHelperOutput`. 

The solution is to remove this check, and let `TagHelperOutput` control its own behavior.

This ultimately plays much nicer if data gets copied into the `ViewBuffer`. That lets intermediate lists get GCed before they can graduate from gen0. `ViewBuffer` is pooled so there's no such issue.

src/Microsoft.AspNet.Mvc.Razor/RazorView.cs
```diff
@@ -233,7 +234,10 @@ private Task RenderPageCoreAsync(IRazorPage page, ViewContext context)
             if (bodyWriter.IsBuffering)
             {
                 // Only copy buffered content to the Output if we're currently buffering.
-                await bodyWriter.Buffer.WriteToAsync(context.Writer, _htmlEncoder);
+                using (var writer = _bufferScope.CreateWriter(context.Writer))
+                {
+                    await bodyWriter.Buffer.WriteToAsync(writer, _htmlEncoder);
+                }
```
rynowak: This is the code path where we encounter sync writes which hurt kestrel.

src/Microsoft.AspNet.Mvc.ViewFeatures/Buffer/ViewBuffer.cs
```diff
@@ -35,17 +37,14 @@ public ViewBuffer(IViewBufferScope bufferScope, string name)
 
             _bufferScope = bufferScope;
             _name = name;
+
+            Pages = new List<ViewBufferPage>();
```
rynowak: I changed this to track the count per-page. It's not a lot to allocate, and it makes this code so much simpler.

src/Microsoft.AspNet.Mvc.ViewFeatures/Buffer/ViewBufferTextWriter.cs
```diff
@@ -0,0 +1,172 @@
...
+using System.Buffers;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNet.Mvc.ViewFeatures.Buffer
+{
+    public class ViewBufferTextWriter : TextWriter
```
rynowak: This is the writer that buffers sync writes.

src/Microsoft.AspNet.Mvc.ViewFeatures/Buffer/ViewBufferValue.cs
```diff
@@ -11,6 +11,7 @@ namespace Microsoft.AspNet.Mvc.ViewFeatures.Buffer
     /// <summary>
     /// Encapsulates a string or <see cref="IHtmlContent"/> value.
     /// </summary>
+    [DebuggerDisplay("{DebuggerToString()}")]
```
rynowak: Required for sanity.

src/Microsoft.AspNet.Mvc.Razor/RazorTextWriter.cs
pranavkm: Could you add this as a note so it doesn't get optimized by accident in the future?

src/Microsoft.AspNet.Mvc.ViewFeatures/Buffer/ViewBuffer.cs
```diff
@@ -67,7 +66,49 @@ public IHtmlContentBuilder AppendHtml(IHtmlContent content)
...
+                // If we haven't started writing a page yet, then let's copy from the other buffer.
+                if (Pages.Count == 0 || Pages[Pages.Count - 1].IsFull)
+                {
+                    Pages.Add(otherPage);
+                    continue;
+                }
+
+                // If the other page is less than half full, let's blit it's contents if we
+                // have room.
+                var isLessThanHalfFull = (double)(otherPage.Capacity - otherPage.Count) / otherPage.Capacity > .5;
```
pranavkm: Is this the same as `2 * otherPage.Count < otherPage.Capacity`?

rynowak: yeah, that would be simpler.

src/Microsoft.AspNet.Mvc.ViewFeatures/Buffer/ViewBuffer.cs
```diff
@@ -67,7 +66,49 @@ public IHtmlContentBuilder AppendHtml(IHtmlContent content)
...
+                var otherPage = otherBuffer.Pages[i];
+
+                // If we haven't started writing a page yet, then let's copy from the other buffer.
+                if (Pages.Count == 0 || Pages[Pages.Count - 1].IsFull)
+                {
+                    Pages.Add(otherPage);
+                    continue;
+                }
+
+                // If the other page is less than half full, let's blit it's contents if we
```
pranavkm: > let's copy

I don't know if blit will make policheck unhappy.

src/Microsoft.AspNet.Mvc.ViewFeatures/Buffer/ViewBuffer.cs
```diff
@@ -67,7 +66,49 @@ public IHtmlContentBuilder AppendHtml(IHtmlContent content)
...
+                    }
+
+                    currentPage.Count += otherPage.Count;
+
+                    // Now we can return this page, and it can be reused in the scope of this request.
+                    _bufferScope.ReturnSegment(otherPage.Buffer);
+                }
+
+                // Otherwise, let's just take the the page from the other buffer.
+                Pages.Add(otherPage);
```
pranavkm: Should this be in an `else`?

rynowak: yeah bro!

src/Microsoft.AspNet.Mvc.Razor/RazorTextWriter.cs
rynowak: Sure

src/Microsoft.AspNet.Mvc.ViewFeatures/Buffer/ViewBufferPage.cs
```diff
@@ -0,0 +1,33 @@
...
+        public ViewBufferPage(ViewBufferValue[] buffer)
+        {
+            Buffer = buffer;
+        }
+
+        public void Append(ViewBufferValue value)
+        {
+            if (IsFull)
+            {
+                throw new InvalidOperationException();
```
pranavkm: Exception message.

src/Microsoft.AspNet.Mvc.ViewFeatures/Buffer/ViewBufferPage.cs
```diff
@@ -0,0 +1,33 @@
...
+            }
+
+            Buffer[Count++] = value;
+        }
+
+        public ViewBufferValue[] Buffer { get; }
```
pranavkm: Properties before method.

src/Microsoft.AspNet.Mvc.ViewFeatures/Buffer/ViewBufferTextWriter.cs
```diff
@@ -0,0 +1,172 @@
...
+        public override Encoding Encoding => _inner.Encoding;
+
+        public override void Flush()
+        {
+            // Don't do anything. We'll call FlushAsync.
+        }
+
+        public override async Task FlushAsync()
+        {
+            if (_pages.Count == 0 || (_currentPage == 0 && _currentIndex == 0))
```
pranavkm: You could generalize the second condition into the loop if it's a common case `if (i == _currentPage && _currentIndex == 0)`

src/Microsoft.AspNet.Mvc.ViewFeatures/Buffer/ViewBufferTextWriter.cs
```diff
@@ -0,0 +1,172 @@
...
+        private char[] GetCurrentPage()
+        {
+            char[] page;
+            if (_pages.Count == 0)
+            {
+                Debug.Assert(_currentPage == 0);
+                Debug.Assert(_currentIndex == 0);
+
+                page = _pool.Rent(PageSize);
+                _pages.Add(page);
```
pranavkm: Wouldn't this (and the other `Rent`) require a try-catch to prevent leaks?

pranavkm:
:watch:

src/Microsoft.AspNet.Mvc.ViewFeatures/Buffer/ViewBufferPage.cs
rynowak: Removed this check, we'll just let the array access throw. This is going to be `.Internal` soon.

src/Microsoft.AspNet.Mvc.ViewFeatures/Buffer/ViewBufferTextWriter.cs
rynowak: I don't see how that's an improvement.

src/Microsoft.AspNet.Mvc.ViewFeatures/Buffer/ViewBufferTextWriter.cs
rynowak: sure

rynowak:
Will update after adding tests for `ViewBufferTextWriter`

src/Microsoft.AspNet.Mvc.ViewFeatures/Buffer/ViewBufferTextWriter.cs
pranavkm: It would avoid the 0-char `WriteAsync` although I imagine that no-ops to start with.

rynowak: ok, I'll just make this into a `count == 0`.

rynowak:
@pranavkm updated

rynowak:
Removed design tag

src/Microsoft.AspNet.Mvc.ViewFeatures/Buffer/ViewBufferPage.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        public int Capacity => Buffer.Length;
+
+        public int Count { get; set; }
+
+        public bool IsFull => Count == Capacity;
+
+        public void Append(ViewBufferValue value) => Buffer[Count++] = value;
```
pranavkm: :100: points

pranavkm:
:shipit:

PR #3923, opened 2016-01-13
Reduce allocations in GlobbingUrlBuilder

pranavkm:
The earlier PR https://github.com/aspnet/Mvc/pull/3896 got closed when I deleted the base branch.


pranavkm:
### Before

**string**
string

**Enumerable + string[]**
enumerable

**int[]**
int

### After
after


This accounts for ~2.8 MB of allocations (4%) for a the tag helper workload.

pranavkm:
cc @rynowak \ @dougbu

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -307,45 +310,66 @@ private void BuildGlobbedLinkTags(TagHelperAttributeList attributes, TagHelperCo
...
-                    .AppendHtml("<script>")
-                    .AppendHtml(JavaScriptResources.GetEmbeddedJavaScript(FallbackJavaScriptResourceName))
-                    .AppendHtml("\"")
-                    .AppendHtml(JavaScriptEncoder.Encode(FallbackTestProperty))
-                    .AppendHtml("\",\"")
-                    .AppendHtml(JavaScriptEncoder.Encode(FallbackTestValue))
-                    .AppendHtml("\",")
-                    .AppendHtml(JavaScriptStringArrayEncoder.Encode(JavaScriptEncoder, fallbackHrefs))
-                    .AppendHtml(");</script>");
+                builder.AppendHtml(JavaScriptEncoder.Encode(valueToWrite));
```
pranavkm: I tried using the `HtmlContentTextWriter` here - however the encoder writes values as characters and we end up allocating strings for each of these.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
```diff
@@ -73,93 +80,105 @@ public GlobbingUrlBuilder(IFileProvider fileProvider, IMemoryCache cache, PathSt
...
-            var includePatterns = include.Split(PatternSeparator, StringSplitOptions.RemoveEmptyEntries);
-            var excludePatterns = exclude?.Split(PatternSeparator, StringSplitOptions.RemoveEmptyEntries);
-
-            if (includePatterns.Length == 0)
-            {
-                return Enumerable.Empty<string>();
-            }
-
-            if (Cache != null)
+            var cacheKey = new GlobbingUrlKey(include, exclude);
```
pranavkm: @rynowak the cache identifier does not capture the values of `include` \ `exclude` which is what we want to unique-ify on.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
```diff
@@ -73,93 +80,105 @@ public GlobbingUrlBuilder(IFileProvider fileProvider, IMemoryCache cache, PathSt
...
+
+            string lastAddedEntry = null;
+            for (var i = 0; i < globbedUrls.Count; i++)
+            {
+                if (!string.Equals(staticUrl, globbedUrls[i], StringComparison.Ordinal))
+                {
+                    lastAddedEntry = globbedUrls[i];
+                    urls.Add(lastAddedEntry);
+                }
+            }
```
rynowak: This block needs a comment about what it means :+1:

rynowak:
Perf things look overall good to me :shipit: - get a signoff from someone who knows this code better :laughing:

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
pranavkm: The variable was left over from the earlier iteration. I'll clean it up.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
```diff
@@ -73,93 +80,104 @@ public GlobbingUrlBuilder(IFileProvider fileProvider, IMemoryCache cache, PathSt
...
             }
 
-            // Add urls that match the globbing patterns specified
-            var matchedUrls = ExpandGlobbedUrl(includePattern, excludePattern);
-            urls.UnionWith(matchedUrls);
+            // The staticUrl always appears first in the sequence.
+            var urls = new List<string>(1 + globbedUrls.Count)
+            {
+                staticUrl
+            };
```
dougbu: Suggest changing `ExpandGlobbedUrl()` to return a concrete `List<string>` and `Insert(0, staticUrl)` by default. Allocate a new `List<string>` only if `ReferenceEquals(EmptyList, globbedUrls)`. Of course `EmptyList` would need to be a `List<string>` then.
Note I'm not suggesting you change the `BuildUrlList()` return type.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
```diff
@@ -73,93 +80,104 @@ public GlobbingUrlBuilder(IFileProvider fileProvider, IMemoryCache cache, PathSt
...
+                var excludeTokenizer = new StringTokenizer(exclude, ',');
                 var trimmedExcludePatterns = new List<string>();
-                for (var i = 0; i < excludePatterns.Length; i++)
+                foreach (var excludePattern in excludeTokenizer)
                 {
-                    trimmedExcludePatterns.Add(TrimLeadingTildeSlash(excludePatterns[i]));
+                    trimmedExcludePatterns.Add(TrimLeadingTildeSlash(excludePattern));
                 }
                 matcher.AddExcludePatterns(trimmedExcludePatterns);
             }
```
dougbu: Probably more readable if you
- factor out this block into a separate method for readability
- do the `exclude` parsing in `ExpandGlobbedUrl()` for consistency

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
```diff
@@ -73,93 +80,104 @@ public GlobbingUrlBuilder(IFileProvider fileProvider, IMemoryCache cache, PathSt
...
 
             var matches = matcher.Execute(_baseGlobbingDirectory);
+            var matchedUrls = new List<string>();
+            foreach (var matchedPath in matches.Files)
+            {
+                // Resolve the path to site root
+                var relativePath = new PathString("/" + matchedPath.Path);
+                var matchedUrl = RequestPathBase.Add(relativePath).ToString();
```
dougbu: Does this end up with `/base//relativePath` if `RequestPathBase.EndsWith('/')`?

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
```diff
@@ -189,60 +207,172 @@ public int Compare(string x, string y)
...
+                        return string.Compare(x, xExtIndex, y, yExtIndex, length, StringComparison.Ordinal);
+                    }
+
+                    return xExtIndex - yExtIndex;
                 }
 
-                var xSegments = xNoExt.Split(new[] { '/' }, StringSplitOptions.RemoveEmptyEntries);
-                var ySegments = yNoExt.Split(new[] { '/' }, StringSplitOptions.RemoveEmptyEntries);
+                var xNoExt = xExtIndex >= 0 ? x.Substring(0, xExtIndex) : x;
+                var yNoExt = yExtIndex >= 0 ? y.Substring(0, yExtIndex) : y;
```
dougbu: Not sure if this extra `string` allocation shows up. But could add a `StringTokenizer` constructor overload that takes a `StringSegment` to avoid this `Substring()`.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
```diff
@@ -189,60 +207,172 @@ public int Compare(string x, string y)
...
+                var result = 0;
+                var xEnumerator = new StringTokenizer(xNoExt, '/').GetEnumerator();
+                var yEnumerator = new StringTokenizer(yNoExt, '/').GetEnumerator();
+                StringSegment xSegment;
+                StringSegment ySegment;
+                while (TryGetNextSegment(ref xEnumerator, out xSegment))
                 {
-                    // Different path depths so shallower path wins
-                    return xSegments.Length.CompareTo(ySegments.Length);
+                    if (!TryGetNextSegment(ref yEnumerator, out ySegment))
```
dougbu: You're probably fixing a bug here e.g. wouldn't previous code place `/ZZ/a.txt` earlier than `/aa/bb/c.txt`? Please ensure you have tests covering such comparisons.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
```diff
@@ -189,60 +207,172 @@ public int Compare(string x, string y)
...
+                    {
+                        // Different path depths (right is shorter), so shallower path wins.
+                        return 1;
+                    }
+
+                    if (result != 0)
+                    {
+                        // Once we've determined that a segment differs, we need to ensure that the two paths
+                        // are of equal depth.
```
dougbu: Please add an example to this comment. That would make the intended order clearer. (For example, I suspect my above "fixing a bug here" comment is incorrect and it'll still place `string`s in that order.)

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
```diff
@@ -189,60 +207,172 @@ public int Compare(string x, string y)
...
+                }
+            }
 
-                    var xToY = string.Compare(xSegment, ySegment, StringComparison.Ordinal);
-                    if (xToY != 0)
+            private static bool TryGetNextSegment(ref StringTokenizer.Enumerator enumerator, out StringSegment segment)
+            {
+                while (enumerator.MoveNext())
+                {
+                    if (enumerator.Current.HasValue && enumerator.Current.Length > 0)
```
dougbu: When would a `Matcher` return paths containing `//`? Probably worth a code comment.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
```diff
@@ -189,60 +207,172 @@ public int Compare(string x, string y)
...
 
-                // Should't get here, but if we do, hey, they're the same :)
-                return 0;
+                segment = default(StringSegment);
+                return false;
             }
         }
 
-        private static string TrimLeadingTildeSlash(string value)
+        private static string TrimLeadingTildeSlash(StringSegment value)
```
dougbu: Actually `TrimLeadingTildeSlashOr(Back)Slash` but I'll leave this up to you.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
```diff
@@ -189,60 +207,172 @@ public int Compare(string x, string y)
...
             {
-                result = result.Substring(2);
+                value = new StringSegment(value.Buffer, value.Offset + 2, value.Length - 2);
             }
-            else if (result.StartsWith("/", StringComparison.Ordinal) ||
-                result.StartsWith("\\", StringComparison.Ordinal))
+            else if (value.StartsWith("/", StringComparison.Ordinal) ||
+                value.StartsWith("\\", StringComparison.Ordinal))
             {
                 // Trim the leading slash as the matcher runs from the provided root only anyway
```
dougbu: If `Matcher` works fine with leading `/` or `\\`, why bother?

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
pranavkm: That would changed the cachedValue - we could add the staticUrl to the cache key which wasn't previously done.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
pranavkm: No - `PathString` accounts for trailing slashes.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
pranavkm: Sure

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
pranavkm: No idea - but this sort of emulates the `RemoveEmtpyEntries` behavior.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
pranavkm: No idea?

pranavkm: Not sure - tests fail if I remove this - so I'm leaving this in.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
pranavkm: I could do this as a follow up. The substring doesn't show up significantly in profiles as yet.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
pranavkm: It's covered in the `OrdersGlobbedMatchResultsCorrectly` test.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/LinkTagHelperTest.cs
```diff
@@ -279,7 +281,10 @@ public static TheoryData RunsWhenRequiredAttributesArePresent_Data
             var output = MakeTagHelperOutput("link");
             var hostingEnvironment = MakeHostingEnvironment();
             var viewContext = MakeViewContext();
-            var globbingUrlBuilder = new Mock<GlobbingUrlBuilder>();
+            var globbingUrlBuilder = new Mock<GlobbingUrlBuilder>(
```
pranavkm: I removed the test-only ctor but forgot to update these tests.

pranavkm:
Updated

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
dougbu: Makes sense.

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
dougbu: :ok:

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/GlobbingUrlBuilder.cs
dougbu: :+1:

src/Microsoft.AspNet.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -307,45 +310,66 @@ private void BuildGlobbedLinkTags(TagHelperAttributeList attributes, TagHelperCo
...
+                .AppendHtml(JavaScriptResources.GetEmbeddedJavaScript(FallbackJavaScriptResourceName))
+                .AppendHtml("\"")
+                .AppendHtml(JavaScriptEncoder.Encode(FallbackTestProperty))
+                .AppendHtml("\",\"")
+                .AppendHtml(JavaScriptEncoder.Encode(FallbackTestValue))
+                .AppendHtml("\",");
+
+            builder.AppendHtml("[");
+
+                // Perf: Avoid allocating enumerator
```
dougbu: De-indent

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/JavaScriptStringArrayEncoder.cs
```diff
@@ -1,43 +0,0 @@
...
-    /// <summary>
-    /// Methods for encoding <c>string[]</c>s for use as a JavaScript array literal.
-    /// </summary>
-    public static class JavaScriptStringArrayEncoder
-    {
-        /// <summary>
-        /// Encodes a <c>string[]</c> for safe use as a JavaScript array literal in many contexts, including
-        /// inline in an HTML file.
-        /// </summary>
-        public static string Encode(JavaScriptEncoder encoder, string[] values)
```
dougbu: Don't think I noticed this change before. Why not pass the `TagHelperContent` into this method? Nice to keep the `BuildFallbackBlock()` length down somewhat.

dougbu:
:watch: for one new question

src/Microsoft.AspNet.Mvc.TagHelpers/Internal/JavaScriptStringArrayEncoder.cs
pranavkm: The earlier code modified the list in place (https://github.com/aspnet/Mvc/pull/3923/files#diff-5574912c61dd39d9a17a1c579ae2a7c0L322) based on the value of `AppendVersion`. Since we return the cached `IReadOnlyList`, I changed the code to calculate the fallBackUrl in place and encode it. Passing `AppendVersion` + `IFileVersionProvider` seems weird. I could refactor this bit out in to a separate method in `LinkTagHelper` if it's easier to read.

dougbu: > I could refactor this bit out in to a separate method in `LinkTagHelper` if it's easier to read.

Please

pranavkm:
Updated

dougbu:
:shipit:

pranavkm:
Merged in https://github.com/aspnet/Mvc/commit/200fb23ba5b99e3b88defb02e0d66830d8e4ebdc

PR #3927, opened 2016-01-13
Removing redundant null logger after hosting fix

JunTaoLuo:
I have resolved https://github.com/aspnet/Hosting/issues/532. I think this is the only place where you added a null logger in https://github.com/aspnet/Mvc/pull/3796 right? @pranavkm


pranavkm:
Yup, that's the only location

:shipit:

PR #3929, opened 2016-01-14
[Fixes #3928] Ensure we throw at UseMvc during application startup if AddMvc wasn't called

### LINKED ISSUE #3928: Ensure we throw an error during startup when MVC services haven't been added to DI
This is the counterpart to https://github.com/aspnet/Routing/issues/276.

javiercn:
We were already doing this check, so I just added a specific unit test for it and did some cleanup (The error message was obsolete) to return a similar message to the one Routing returns in the same scenario.


moozzyk:
@davidfowl https://github.com/aspnet/SignalR-Server/commit/da96c1d9abae49e6aadeefd30942ad0269a65996 

I don't think we run any tests for SignalR Server ATM...

davidfowl:
@moozzyk good we do it in the right place.

src/Microsoft.AspNet.Mvc.Core/Builder/MvcApplicationBuilderExtensions.cs
```diff
@@ -77,7 +79,13 @@ public static IApplicationBuilder UseMvcWithDefaultRoute(this IApplicationBuilde
 
             // Verify if AddMvc was done before calling UseMvc
             // We use the MvcMarkerService to make sure if all the services were added.
-            MvcServicesHelper.ThrowIfMvcNotRegistered(app.ApplicationServices);
+            if (app.ApplicationServices.GetService(typeof(MvcMarkerService)) == null)
+            {
+                throw new InvalidOperationException(Resources.FormatUnableToFindServices(
+                    nameof(IServiceCollection),
+                    "AddMvc",
```
dougbu: Use `nameof(MvcServiceCollectionExtensions.AddMvc)`. Since this is in the Core repo, should also mention `nameof(MvcCoreServiceCollectionExtensions.AddMvcCore)`.

test/Microsoft.AspNet.Mvc.Core.Test/Builder/MvcApplicationBuilderExtensionsTest.cs
```diff
@@ -0,0 +1,34 @@
...
+        [Fact]
+        public void UseMvc_ThrowsInvalidOperationException_IfRoutingMarkerServiceIsNotRegistered()
+        {
+            // Arrange
+            var applicationBuilderMock = new Mock<IApplicationBuilder>();
+            applicationBuilderMock
+                .Setup(s => s.ApplicationServices)
+                .Returns(Mock.Of<IServiceProvider>());
+
+            // Act
```
dougbu: `// Act & Assert` and remove division just below.

test/Microsoft.AspNet.Mvc.Core.Test/Builder/MvcApplicationBuilderExtensionsTest.cs
```diff
@@ -0,0 +1,34 @@
...
+            var error = Assert.Throws<InvalidOperationException>(
```
dougbu: May be worthwhile being consistent with most similar tests: Use `exception` or `ex`. `error` doesn't mean anything here.

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.Core/Builder/MvcApplicationBuilderExtensions.cs
javiercn: Can't do that due to the current factoring of the code. MvcServiceCollectionExtensions lives in Microsoft.AspNet.Mvc, this piece of code lives in Microsoft.AspNet.Mvc.Core

javiercn:
@dougbu updated

dougbu:
:shipit:

src/Microsoft.AspNet.Mvc.Core/Builder/MvcApplicationBuilderExtensions.cs
rynowak: Don't mention `AddMvcCore`. That's for people who know what they are doing.

src/Microsoft.AspNet.Mvc.Core/Resources.resx
```diff
@@ -208,7 +208,7 @@
     <value>An unescaped '[' token is not allowed inside of a replacement token. Use '[[' to escape.</value>
   </data>
   <data name="UnableToFindServices" xml:space="preserve">
-    <value>Unable to find the required services. Please add all the required services by calling '{0}' inside the call to '{1}' or '{2}' in the application startup code.</value>
+    <value>Unable to find the required services. Please add all the required services by calling '{0}.{1}' or '{2}.{3}' inside the call to '{4}' in the application startup code.</value>
   </data>
```
rynowak: I don't want to mention mvc core here, that's an advanced scenario that we shouldn't steer beginners toward.

rynowak:
:watch:

src/Microsoft.AspNet.Mvc.Core/Builder/MvcApplicationBuilderExtensions.cs
javiercn: @dougbu Are you ok with what @rynowak said?

dougbu: Yup.

javiercn:
@rynowak updated

rynowak:
:shipit:

PR #3931, opened 2016-01-14
Cleaning up VirtualFileResult tests

pranavkm:


test/Microsoft.AspNet.Mvc.Core.Test/VirtualFileResultTest.cs
```diff
@@ -241,73 +267,29 @@ public void Constructor_SetsContentTypeAndParameters()
             Assert.Equal(expectedData, contents);
         }
 
-        [Theory]
```
pranavkm: These were written when VIrtualFileResult did work with the specified paths. We don't do anything interesting with it now.

test/Microsoft.AspNet.Mvc.Core.Test/VirtualFileResultTest.cs
```diff
@@ -241,73 +267,29 @@ public void Constructor_SetsContentTypeAndParameters()
...
-        [InlineData("\\SubFolder\\SubFolderTestFile.txt")]
-        [InlineData("/SubFolder\\SubFolderTestFile.txt")]
-        [InlineData("\\SubFolder/SubFolderTestFile.txt")]
-        [InlineData("./SubFolder/SubFolderTestFile.txt")]
-        [InlineData(".\\SubFolder\\SubFolderTestFile.txt")]
-        [InlineData("./SubFolder\\SubFolderTestFile.txt")]
-        [InlineData(".\\SubFolder/SubFolderTestFile.txt")]
-        [InlineData("~/SubFolder/SubFolderTestFile.txt")]
-        [InlineData("~/SubFolder\\SubFolderTestFile.txt")]
-        public void ExecuteResultAsync_ThrowsDirectoryNotFound_IfFileProviderCanNotFindTheDirectory(string path)
```
pranavkm: This tested a `Assert.ThrowsAsync` without awaiting it. Awaiting it fails the test since we no longer throw `DirectoryNotFoundException`.

pranavkm:
cc @javiercn

pranavkm:
Ping

test/Microsoft.AspNet.Mvc.Core.Test/VirtualFileResultTest.cs
dougbu: Not quiet parsing your comment. Shouldn't we still at least test a few longer paths, including those that start with dots and slashes?

pranavkm: The tests were meant for behavior that `FileResult` used to support - https://github.com/aspnet/Mvc/commit/ddc74e5396c5a09e950c5d0029bc23d452bc7783#diff-79ab8ced19b9f57a37b11093d8533ed4L112 to detect if a path was rooted . That code's been removed and we pass the path as is (with the exception of removing the tilde).

dougbu: I agree that you've got tests which verify the `path` is passed unchanged (or with tilde removed) to the underlying `IFileProvider` in a few simple cases. What I'm asking is a few more cases added to the `ExecuteResultAsync_ReturnsFiles_ForDifferentPaths()` and perhaps `ExecuteResultAsync_TrimsTilde_BeforeInvokingFileProvider()` test data. That is, the new pass-through behaviour isn't tested enough.

dougbu:
:watch:

test/Microsoft.AspNet.Mvc.Core.Test/VirtualFileResultTest.cs
pranavkm: Ok.

pranavkm:
Updated

test/Microsoft.AspNet.Mvc.Core.Test/VirtualFileResultTest.cs
```diff
@@ -184,10 +184,10 @@ public void Constructor_SetsContentTypeAndParameters()
         [Theory]
         [InlineData("FilePathResultTestFile.txt")]
         [InlineData("TestFiles/FilePathResultTestFile.txt")]
+        [InlineData("TestFiles/../FilePathResultTestFile.txt")]
         [InlineData("TestFiles\\FilePathResultTestFile.txt")]
-        [InlineData("~/FilePathResultTestFile.txt")]
-        [InlineData("~/TestFiles/FilePathResultTestFile.txt")]
-        [InlineData("~/TestFiles\\FilePathResultTestFile.txt")]
+        [InlineData("TestFiles\\..\\FilePathResultTestFile.txt")]
+        [InlineData(@"\\..//?><|""&@#\c:\..\? /..txt")]
```
dougbu: :smiley_cat:

dougbu:
:shipit:

Issue #3868, opened 2016-01-06
Should BeginForm and FormTagHelper include an antiforgery token for GET


rynowak:
We could be smart and skip the antiforgery token in the form when it's for a `GET`.

We'd probably want to have the same behavior for `BeginForm` and `<form>` taghelper

NTaylorMullen:
Ooo, That'd be fancy. :+1: I like.

Eilon:
Yeah we don't want it on GETs by default. But the user should be able to force it on or off always. So let's make sure the anti-forgery option in each of these mechanisms is a nullable `bool`, with `null` meaning "auto".

PR #3933, opened 2016-01-15
Exclude Antiforgery token in form with method Get

ajaybhargavb:
Issue - #3868 

@NTaylorMullen @rynowak

### DIFF
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/FormTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/FormTagHelper.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using Microsoft.AspNet.Mvc.Rendering;
 using Microsoft.AspNet.Mvc.ViewFeatures;
 using Microsoft.AspNet.Razor.TagHelpers;
@@ -68,7 +69,8 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
         /// <summary>
         /// Whether the antiforgery token should be generated.
         /// </summary>
-        /// <value>Defaults to <c>false</c> if user provides an <c>action</c> attribute; <c>true</c> otherwise.</value>
+        /// <value>Defaults to <c>false</c> if user provides an <c>action</c> attribute
+        /// or if the <c>method</c> is <see cref="FormMethod.Get"/>; <c>true</c> otherwise.</value>
         [HtmlAttributeName(AntiforgeryAttributeName)]
         public bool? Antiforgery { get; set; }
 
@@ -80,6 +82,13 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
         /// </remarks>
         [HtmlAttributeName(RouteAttributeName)]
         public string Route { get; set; }
+
+        /// <summary>
+        /// The HTTP method to use.
+        /// </summary>
+        /// <remarks>Passed through to the generated HTML in all cases.</remarks>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public string Method { get; set; }
 
         /// <summary>
         /// Additional parameters for the route.
@@ -121,6 +130,10 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
             if (output == null)
             {
                 throw new ArgumentNullException(nameof(output));
+            }
+            if (Method != null)
+            {
+                output.CopyHtmlAttribute(nameof(Method), context);
             }
 
             var antiforgeryDefault = true;
@@ -194,6 +207,11 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
                 {
                     output.MergeAttributes(tagBuilder);
                     output.PostContent.AppendHtml(tagBuilder.InnerHtml);
+                }
+
+                if (string.Equals(Method, FormMethod.Get.ToString(), StringComparison.OrdinalIgnoreCase))
+                {
+                    antiforgeryDefault = false;
                 }
             }
 
--- a/src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/HtmlHelperFormExtensions.cs
+++ b/src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/HtmlHelperFormExtensions.cs
@@ -34,7 +34,7 @@ namespace Microsoft.AspNet.Mvc.Rendering
                 controllerName: null,
                 routeValues: null,
                 method: FormMethod.Post,
-                suppressAntiforgery: false,
+                antiforgery: null,
                 htmlAttributes: null);
         }
 
@@ -43,9 +43,11 @@ namespace Microsoft.AspNet.Mvc.Rendering
         /// match the current request.
         /// </summary>
         /// <param name="htmlHelper">The <see cref="IHtmlHelper"/> instance this method extends.</param>
-        /// <param name="suppressAntiforgery">
-        /// If <c>true</c>, suppresses the generation an &lt;input&gt; of type "hidden" with an antiforgery token. By
-        /// default &lt;form&gt; elements will automatically include an antiforgery token.
+        /// <param name="antiforgery">
+        /// If <c>true</c>, &lt;form&gt; elements will include an antiforgery token.
+        /// If <c>false</c>, suppresses the generation an &lt;input&gt; of type "hidden" with an antiforgery token.
+        /// If <c>null</c>, &lt;form&gt; elements will include an antiforgery token only if
+        /// <paramref name="method"/> is not <see cref="FormMethod.Get"/>.
         /// </param>
         /// <returns>
         /// An <see cref="MvcForm"/> instance which renders the &lt;/form&gt; end tag when disposed.
@@ -53,7 +55,7 @@ namespace Microsoft.AspNet.Mvc.Rendering
         /// <remarks>
         /// In this context, "renders" means the method writes its output using <see cref="ViewContext.Writer"/>.
         /// </remarks>
-        public static MvcForm BeginForm(this IHtmlHelper htmlHelper, bool suppressAntiforgery)
+        public static MvcForm BeginForm(this IHtmlHelper htmlHelper, bool? antiforgery)
         {
             if (htmlHelper == null)
             {
@@ -66,7 +68,7 @@ namespace Microsoft.AspNet.Mvc.Rendering
                 controllerName: null,
                 routeValues: null,
                 method: FormMethod.Post,
-                suppressAntiforgery: suppressAntiforgery,
+                antiforgery: antiforgery,
                 htmlAttributes: null);
         }
 
@@ -94,7 +96,7 @@ namespace Microsoft.AspNet.Mvc.Rendering
                 controllerName: null,
                 routeValues: null,
                 method: method,
-                suppressAntiforgery: false,
+                antiforgery: null,
                 htmlAttributes: null);
         }
 
@@ -130,7 +132,7 @@ namespace Microsoft.AspNet.Mvc.Rendering
                 controllerName: null,
                 routeValues: null,
                 method: method,
-                suppressAntiforgery: false,
+                antiforgery: null,
                 htmlAttributes: htmlAttributes);
         }
 
@@ -140,9 +142,11 @@ namespace Microsoft.AspNet.Mvc.Rendering
         /// </summary>
         /// <param name="htmlHelper">The <see cref="IHtmlHelper"/> instance this method extends.</param>
         /// <param name="method">The HTTP method for processing the form, either GET or POST.</param>
-        /// <param name="suppressAntiforgery">
-        /// If <c>true</c>, suppresses the generation an &lt;input&gt; of type "hidden" with an antiforgery token. By
-        /// default &lt;form&gt; elements will automatically include an antiforgery token.
+        /// <param name="antiforgery">
+        /// If <c>true</c>, &lt;form&gt; elements will include an antiforgery token.
+        /// If <c>false</c>, suppresses the generation an &lt;input&gt; of type "hidden" with an antiforgery token.
+        /// If <c>null</c>, &lt;form&gt; elements will include an antiforgery token only if
+        /// <paramref name="method"/> is not <see cref="FormMethod.Get"/>.
         /// </param>
         /// <param name="htmlAttributes">
         /// An <see cref="object"/> that contains the HTML attributes for the element. Alternatively, an
@@ -158,7 +162,7 @@ namespace Microsoft.AspNet.Mvc.Rendering
         public static MvcForm BeginForm(
             this IHtmlHelper htmlHelper,
             FormMethod method,
-            bool suppressAntiforgery,
+            bool? antiforgery,
             object htmlAttributes)
         {
             if (htmlHelper == null)
@@ -171,7 +175,7 @@ namespace Microsoft.AspNet.Mvc.Rendering
                 controllerName: null,
                 routeValues: null,
                 method: method,
-                suppressAntiforgery: suppressAntiforgery,
+                antiforgery: antiforgery,
                 htmlAttributes: htmlAttributes);
         }
 
@@ -205,7 +209,7 @@ namespace Microsoft.AspNet.Mvc.Rendering
                 controllerName: null,
                 routeValues: routeValues,
                 method: FormMethod.Post,
-                suppressAntiforgery: false,
+                antiforgery: null,
                 htmlAttributes: null);
         }
 
@@ -237,7 +241,7 @@ namespace Microsoft.AspNet.Mvc.Rendering
                 controllerName,
                 routeValues: null,
                 method: FormMethod.Post,
-                suppressAntiforgery: false,
+                antiforgery: null,
                 htmlAttributes: null);
         }
 
@@ -277,7 +281,7 @@ namespace Microsoft.AspNet.Mvc.Rendering
                 controllerName,
                 routeValues,
                 FormMethod.Post,
-                suppressAntiforgery: false,
+                antiforgery: null,
                 htmlAttributes: null);
         }
 
@@ -311,7 +315,7 @@ namespace Microsoft.AspNet.Mvc.Rendering
                 controllerName,
                 routeValues: null,
                 method: method,
-                suppressAntiforgery: false,
+                antiforgery: null,
                 htmlAttributes: null);
         }
 
@@ -353,7 +357,7 @@ namespace Microsoft.AspNet.Mvc.Rendering
                 controllerName,
                 routeValues,
                 method,
-                suppressAntiforgery: false,
+                antiforgery: null,
                 htmlAttributes: null);
         }
 
@@ -393,7 +397,7 @@ namespace Microsoft.AspNet.Mvc.Rendering
                 controllerName,
                 routeValues: null,
                 method: method,
-                suppressAntiforgery: false,
+                antiforgery: null,
                 htmlAttributes: htmlAttributes);
         }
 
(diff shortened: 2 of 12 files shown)


src/Microsoft.AspNet.Mvc.TagHelpers/FormTagHelper.cs
```diff
@@ -124,6 +125,14 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
             }
 
             var antiforgeryDefault = true;
+            if (output.Attributes.ContainsName("method"))
```
NTaylorMullen: Hmm, we could bind `method`. That way we can avoid all of this tostringing madness below. That way we can just pass it through to the HtmlGenerator

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/IHtmlHelper.cs
```diff
@@ -133,7 +135,7 @@ public interface IHtmlHelper
             string controllerName,
             object routeValues,
             FormMethod method,
-            bool suppressAntiforgery,
+            bool? suppressAntiforgery,
```
NTaylorMullen: Hmmm, are we able to not expose the nullable bool to public API and have extension methods that simply don't take supressAntiforgery?

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/IHtmlHelper.cs
ajaybhargavb: @dougbu, What is your opinion on adding another overload for `BeginForm` and `BeginRouteForm` on `IHtmlHelper`?

src/Microsoft.AspNet.Mvc.TagHelpers/FormTagHelper.cs
dougbu: Yeah, I didn't bind the `method` attribute because `DefaultHtmlGenerator` didn't need it. If we need it now, just bind it.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/IHtmlHelper.cs
dougbu: Why would such an overload be useful? Why is it useful to make `suppressAntiforgery` nullable? (Hint, it should _suppress_ antiforgery when `true` and leave the system to its own whims when `false`.)

NTaylorMullen: I just don't like the nullable aspect of it :smile:

dougbu: My point is `null` adds nothing to the mix because `false` should mean "do what you wanna do".

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/HtmlHelperFormExtensions.cs
```diff
@@ -44,8 +44,10 @@ public static MvcForm BeginForm(this IHtmlHelper htmlHelper)
         /// </summary>
         /// <param name="htmlHelper">The <see cref="IHtmlHelper"/> instance this method extends.</param>
         /// <param name="suppressAntiforgery">
-        /// If <c>true</c>, suppresses the generation an &lt;input&gt; of type "hidden" with an antiforgery token. By
-        /// default &lt;form&gt; elements will automatically include an antiforgery token.
+        /// If <c>true</c>, suppresses the generation an &lt;input&gt; of type "hidden" with an antiforgery token.
+        /// If <c>false</c> &lt;form&gt; elements will include an antiforgery token.
+        /// By default, &lt;form&gt; elements will include an antiforgery token only if
+        /// <paramref name="method"/> is not <see cref="FormMethod.Get"/>.
```
dougbu: This PR should have no affect on this API except to change the words after `By default`.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/IHtmlHelper.cs
dougbu: Put another way: What is the use case for forcing an antiforgery token?

Name doesn't make much sense if we think there's a real scenario here for `null` as distinct from `false` either. Even with IntelliSense help, hard to read `suppressAntiforgery: false` as "force addition of an antiforgery token": User just asked for token not to be _suppressed_. So would have to reverse the polarity and name it `includeAntiforgery` with `null` meaning "default", `false` meaning "no", and `true` meaning "yes".

src/Microsoft.AspNet.Mvc.TagHelpers/FormTagHelper.cs
dougbu: > That way we can just pass it through to the `HtmlGenerator`

Note passing it into the `IHtmlGenerator` isn't useful per se and `FormMethod.Get.ToString()` will still be necessary. `GenerateForm()` doesn't deal with the antiforgery token but would need the `method` value to add the bound attribute back. And we need the `string` comparison to determine the `antiforgeryDefault` value.

src/Microsoft.AspNet.Mvc.TagHelpers/FormTagHelper.cs
```diff
@@ -124,6 +125,14 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
...
+            {
+                var methodValue = output.Attributes["method"].Value.ToString();
+                if (string.Equals(methodValue, FormMethod.Get.ToString(), StringComparison.OrdinalIgnoreCase))
+                {
+                    antiforgeryDefault = false;
+                }
+            }
```
dougbu: Whole block is miss-placed because `antiforgeryDefault` will be `false` anyhow if `output.Attributes.ContainsName(HtmlActionAttributeName)`. Move down below the `else` at [line 157](https://github.com/aspnet/Mvc/pull/3933/files#diff-8bd6f4dfc2f4a07e5b08261979376682R157).

src/Microsoft.AspNet.Mvc.TagHelpers/FormTagHelper.cs
dougbu: I take that back. Don't change the call to `GenerateForm()`. Instead unconditionally use `output.CopyHtmlAttribute()` for the (HTML) `method` attribute.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/IHtmlHelper.cs
dougbu: BTW everyone using HTML helpers can always call `Html.AntiForgery()` to make the token addition unconditional.

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/IHtmlHelper.cs
rynowak: Ugh

Eilon: Chatted in person with @ajaybhargavb and @dougbu . We want the new behavior, but we'll rename the parameter to match what the tag helper uses: `BeginForm(...., bool? antiForgery)`. And the behavior and docs will also be the same as the tag helper: true = on, false = off, null = smart.

We'll also post updates to the Announcements repo and related discussion item to let people know we renamed the property and flipped the behavior.

ajaybhargavb:
:up: :date:

src/Microsoft.AspNet.Mvc.TagHelpers/FormTagHelper.cs
```diff
@@ -82,6 +83,12 @@ public override int Order
         public string Route { get; set; }
 
         /// <summary>
+        /// The http method to use.
+        /// </summary>
+        /// <remarks>Passed through to the generated HTML in all cases.</remarks>
+        public string Method { get; set; }
```
NTaylorMullen: Lets add an `EditorBrowsable.Never` decorator to this. No value in showing consumers we're binding their `Method` attribute.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/HtmlHelperFormExtensions.cs
```diff
@@ -43,9 +43,11 @@ public static MvcForm BeginForm(this IHtmlHelper htmlHelper)
...
         /// </summary>
         /// <param name="htmlHelper">The <see cref="IHtmlHelper"/> instance this method extends.</param>
-        /// <param name="suppressAntiforgery">
-        /// If <c>true</c>, suppresses the generation an &lt;input&gt; of type "hidden" with an antiforgery token. By
-        /// default &lt;form&gt; elements will automatically include an antiforgery token.
+        /// <param name="antiforgery">
+        /// If <c>true</c>, &lt;form&gt; elements will include an antiforgery token.
+        /// If <c>false</c> suppresses the generation an &lt;input&gt; of type "hidden" with an antiforgery token.
+        /// By default, &lt;form&gt; elements will include an antiforgery token only if
+        /// <paramref name="method"/> is not <see cref="FormMethod.Get"/>.
```
NTaylorMullen: Shouldnt we document the `null` behavior?

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/HtmlHelperFormExtensions.cs
ajaybhargavb: The `By default` part actually talks about the `null` case.

src/Microsoft.AspNet.Mvc.TagHelpers/FormTagHelper.cs
```diff
@@ -82,6 +83,12 @@ public override int Order
         public string Route { get; set; }
 
         /// <summary>
+        /// The http method to use.
```
Eilon: `http` --> `HTTP`

src/Microsoft.AspNet.Mvc.TagHelpers/FormTagHelper.cs
```diff
@@ -124,6 +131,14 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
             }
 
             var antiforgeryDefault = true;
+            if (Method != null)
+            {
+                output.CopyHtmlAttribute(nameof(Method), context);
+                if (string.Equals(Method, FormMethod.Get.ToString(), StringComparison.OrdinalIgnoreCase))
+                {
+                    antiforgeryDefault = false;
+                }
```
dougbu: Move this inner block down into the `else` at [line 163)(https://github.com/aspnet/Mvc/pull/3933/files#diff-8bd6f4dfc2f4a07e5b08261979376682R163). Note you don't need to double-check `Method != null` there.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/HtmlHelperFormExtensions.cs
dougbu: Rephrase to explicitly say `If <c>null</c>,`.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/HtmlHelperFormExtensions.cs
```diff
@@ -43,9 +43,11 @@ public static MvcForm BeginForm(this IHtmlHelper htmlHelper)
         /// match the current request.
         /// </summary>
         /// <param name="htmlHelper">The <see cref="IHtmlHelper"/> instance this method extends.</param>
-        /// <param name="suppressAntiforgery">
-        /// If <c>true</c>, suppresses the generation an &lt;input&gt; of type "hidden" with an antiforgery token. By
-        /// default &lt;form&gt; elements will automatically include an antiforgery token.
+        /// <param name="antiforgery">
+        /// If <c>true</c>, &lt;form&gt; elements will include an antiforgery token.
+        /// If <c>false</c> suppresses the generation an &lt;input&gt; of type "hidden" with an antiforgery token.
```
dougbu: Be consistent: Comma after `<c>false</c>`.

dougbu: Also better if you use the longer description of what gets generated in the first (`If <c>true</c>`) sentence.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/HtmlHelperFormExtensions.cs
```diff
@@ -140,9 +142,11 @@ public static MvcForm BeginForm(this IHtmlHelper htmlHelper, FormMethod method)
         /// </summary>
         /// <param name="htmlHelper">The <see cref="IHtmlHelper"/> instance this method extends.</param>
         /// <param name="method">The HTTP method for processing the form, either GET or POST.</param>
-        /// <param name="suppressAntiforgery">
-        /// If <c>true</c>, suppresses the generation an &lt;input&gt; of type "hidden" with an antiforgery token. By
-        /// default &lt;form&gt; elements will automatically include an antiforgery token.
+        /// <param name="antiforgery">
```
dougbu: Ditto all doc comment comments for the other `<param name="antiforgery">`s. They should all be identical when you're done.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperFormExtensionsTest.cs
```diff
@@ -397,10 +397,14 @@ public void BeginFormWithMethodParameter_CallsHtmlGeneratorWithExpectedValues(Fo
...
-                .Returns(HtmlString.Empty)
-                .Verifiable();
+
+            if (method != FormMethod.Get)
+            {
+                htmlGenerator
+                    .Setup(g => g.GenerateAntiforgery(htmlHelper.ViewContext))
+                    .Returns(HtmlString.Empty)
+                    .Verifiable();
+            }
```
dougbu: Better to split tests in two than to make the setup conditional.

ajaybhargavb: Yes but there are too many tests that input `FormMethod` in theory and I felt that it was really unnecessary to split every single one.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperFormTest.cs
```diff
@@ -477,7 +516,46 @@ public void BeginForm_EndForm_SuppressAntiforgeryToken()
             Assert.NotNull(writer);
 
             // Act & Assert
-            using (var form = htmlHelper.BeginForm(FormMethod.Post, suppressAntiforgery: true, htmlAttributes: null))
+            using (var form = htmlHelper.BeginForm(FormMethod.Post, antiforgery: false, htmlAttributes: null))
```
dougbu: Add a test demonstrating that `antiforgery: true` "wins" even if `method: FormMethod.Get`. That is the `<form> element will contain an `<antiforgery>` element in that case.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperFormExtensionsTest.cs
dougbu: Why not split the data sets in two as well?

dougbu:
:watch:

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperFormExtensionsTest.cs
ajaybhargavb: There would be too many tests testing almost the same thing. We could split one of these tests in two just to demonstrate this but I would suggest not to split every test in two.

dougbu: Please split. And if we're doing the same test multiple times, feed free to nuke some bits (but include PR comments so we don't have to ask why something died an ignoble death).

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperFormTest.cs
ajaybhargavb: Added.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/HtmlHelperFormExtensions.cs
ajaybhargavb: Done.

ajaybhargavb:
:up: :date:

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperFormExtensionsTest.cs
dougbu: Discussed offline. Since you're not creating new tests and it wouldn't be a quick change, fine as-is.

src/Microsoft.AspNet.Mvc.ViewFeatures/Rendering/HtmlHelperFormExtensions.cs
```diff
@@ -43,9 +43,11 @@ public static MvcForm BeginForm(this IHtmlHelper htmlHelper)
...
         /// </summary>
         /// <param name="htmlHelper">The <see cref="IHtmlHelper"/> instance this method extends.</param>
-        /// <param name="suppressAntiforgery">
-        /// If <c>true</c>, suppresses the generation an &lt;input&gt; of type "hidden" with an antiforgery token. By
-        /// default &lt;form&gt; elements will automatically include an antiforgery token.
+        /// <param name="antiforgery">
+        /// If <c>true</c>, &lt;form&gt; elements will include an antiforgery token.
+        /// If <c>false</c>, suppresses the generation an &lt;input&gt; of type "hidden" with an antiforgery token.
+        /// If <c>null</c>, &lt;form&gt; elements will include an antiforgery token only if
+        /// <paramref name="method"/> is not <see cref="FormMethod.Get"/>.
```
dougbu: Up to you but `... unless <paramref name="method"/> is <see cref="FormMethod.Get"/>.` reads slightly better.

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/Rendering/HtmlHelperFormTest.cs
```diff
@@ -487,6 +565,47 @@ public void BeginForm_EndForm_SuppressAntiforgeryToken()
                 writer.GetStringBuilder().ToString());
         }
 
+        [Theory]
+        [InlineData(FormMethod.Get)]
+        [InlineData(FormMethod.Post)]
+        public void BeginForm_EndForm_DoesNotSuppressAntiforgeryTokenWhenAntiforgeryIsTrue(FormMethod method)
```
dougbu: Up to you but `Begin[Route]Form_EndForm_MethodDoesNotSuppressAntiforgeryTokenWhenAntiforgeryIsTrue()` would be slightly more complete. Or perhaps `Begin[Route]Form_EndForm_AntiforgeryTrueOverridesMethod()`.

dougbu:
:shipit:

Issue #3386, opened 2015-10-22
Select tag helper with static options


tuespetre:
Given a select tag bound to a model property using `asp-for` that has only a few static options, the option matching the value of the model property is not automatically selected. I understand that the select options could be moved out into a getter on the model but it's also kind of nice just having them in the view. Not sure if this is a `wontfix` or not so I'm posting it here. Thanks for looking!

Using beta8

dougbu:
Please provide more information on what you're seeing.

The `<select>` and `<option>` tag helpers work together to ensure the correct `<option>` element(s) is / are selected. If something is going wrong here, it's a scenario variant we aren't currently testing.

tuespetre:
I was having this problem in two cases:

1. Error on my part, was not binding correctly in the controller.
2. When the static options were text (like, full comments or sentences) and did not have the 'value' attribute set

So the first case was my bad; the second case is IMO a discrepancy between how HTML works (per spec) and how the tag helper resolves selected values. In HTML, the value of an `<option>` is equivalent to its `textContent` IDL attribute when there is no value attribute defined. That's kind of what I thought the first case was going to be too.

For now I will just move the static options for my second case into a group of constant values and use an `IEnumerable<SelectListItem>` getter on my view model to spit them out, assigning the constants to both the `Text` and `Value` properties of `SelectListItem` so the tag helper picks up their values.

Assuming this will be a `wontfix`, go ahead and close if you feel the same :)

dougbu:
> Assuming this will be a `wontfix`

Not a good assumption.

The `<option>` tag helper checks its body if the `value` attribute is not provided. So this should work. Could you upload a repro project to a GitHub repo?

tuespetre:
> Not a good assumption

Sorry! Nothing on your (team's) part, just that I didn't know how much ripple effect something like that would cause. I will look closer at what I'm doing and report back. Thanks for your help!

tuespetre:
Alright. Cloned `Mvc` and `Common` and added them to my solution's `global.json` to debug.

Here is a sample of the HTML:

```
<select class="form-control" asp-for="@Model.CpniQuestion">
  <option></option>
  <option>The last 4 digits of the primary account holder’s Social Security number</option>
</select>
```

Here is the value of `childContent.GetContent()`:
`"The last 4 digits of the primary account holder’s Social Security number"`

Here is the value of `encodedValues.First()`:
`"The last 4 digits of the primary account holder&#x2019;s Social Security number"`

[...]

PR #3937, opened 2016-01-15
Check encoded and unencoded values against element body in `OptionTagHelper`

dougbu:
- #3386
- initialize comparison `HashSet` with unencoded values to ensure both are checked
- address correctness and perf issues in this code
  - `context.Items[typeof(SelectTagHelper)]` entry read as `ICollection` but written as `IReadOnlyCollection`
    - `IReadOnlyCollection` worse because it does not include `Contains()`, causing Linq use
  - every `<option>` element recalculated the encoded values and created a `HashSet` to contain them
    - add `CurrentValues` type to cache this `HashSet` in `context.Items`
  - each `OptionTagHelper` created the additional `HashSet` even if `Value` was bound

### DIFF
--- a/src/Microsoft.AspNet.Mvc.TagHelpers/OptionTagHelper.cs
+++ b/src/Microsoft.AspNet.Mvc.TagHelpers/OptionTagHelper.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Mvc.Rendering;
+using Microsoft.AspNet.Mvc.TagHelpers.Internal;
 using Microsoft.AspNet.Mvc.ViewFeatures;
 using Microsoft.AspNet.Razor.TagHelpers;
 
@@ -55,8 +56,8 @@ namespace Microsoft.AspNet.Mvc.TagHelpers
         /// <inheritdoc />
         /// <remarks>
         /// Does nothing unless <see cref="TagHelperContext.Items"/> contains a
-        /// <see cref="SelectTagHelper"/> <see cref="Type"/> entry and that entry is a non-empty
-        /// <see cref="ICollection{string}"/> instance. Also does nothing if the associated &lt;option&gt; is already
+        /// <see cref="SelectTagHelper"/> <see cref="Type"/> entry and that entry is a non-<c>null</c>
+        /// <see cref="CurrentValues"/> instance. Also does nothing if the associated &lt;option&gt; is already
         /// selected.
         /// </remarks>
         public override async Task ProcessAsync(TagHelperContext context, TagHelperOutput output)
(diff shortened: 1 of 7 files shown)


dougbu:
Took me a while to convince myself but just the `HashSet<string>` (including entries and enumerators) reduction saves about 2% or 20 MB.

Overall scenario is down by much more (~170 MB in total). Mainly the site runs smoother with the change. So `ViewBufferValue[]` drops significantly.

##### Before
static options 0
static options 0

##### After
static options 1
static options 1

src/Microsoft.AspNet.Mvc.TagHelpers/SelectTagHelper.cs
```diff
@@ -94,7 +95,8 @@ public override void Init(TagHelperContext context)
 
             // Whether or not (not being highly unlikely) we generate anything, could update contained <option/>
             // elements. Provide selected values for <option/> tag helpers.
-            context.Items[typeof(SelectTagHelper)] = _currentValues;
+            var currentValues = _currentValues == null ? null : new CurrentValues(_currentValues);
+            context.Items[typeof(SelectTagHelper)] = currentValues;
```
dougbu: Current `OptionTagHelper` doesn't need a `CurrentValues` instance when `Values` is empty. But allocating one in that case keeps the semantics consistent with `IHtmlGenerator.GetCurrentValues()` and the expression value.

dougbu:
FYI CI builds are failing due to inconsistent `Newtonsoft.Json` major versions.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/IHtmlGenerator.cs
```diff
@@ -303,7 +303,7 @@ public interface IHtmlGenerator
             string optionLabel,
             string expression,
             IEnumerable<SelectListItem> selectList,
-            IReadOnlyCollection<string> currentValues,
+            ICollection<string> currentValues,
```
dougbu: Already spoke to @Eilon about this minor correction in the API. Not a big deal since we expect few users to interact with an `IHtmlGenerator`.

test/Microsoft.AspNet.Mvc.TagHelpers.Test/OptionTagHelperTest.cs
```diff
@@ -233,7 +236,7 @@ public class OptionTagHelperTest
                             "text")
                     },
                     {
-                        "HtmlEncode[[text]]", null, null, new [] { "text", },
+                        "HtmlEncode[[text]]", null, null, new CurrentValues(new HashSet<string>(new [] { "text", })),
```
dougbu: Forgot to add a similar entry using `"text"` as the element's content e.g. the following will now also generate `selected="selected"`:
``` c#
  "text", null, null, new CurrentValues(new HashSet<string>(new [] { "text", })),
  ...
```

src/Microsoft.AspNet.Mvc.TagHelpers/OptionTagHelper.cs
```diff
@@ -55,8 +56,8 @@ public override int Order
         /// <inheritdoc />
         /// <remarks>
         /// Does nothing unless <see cref="TagHelperContext.Items"/> contains a
-        /// <see cref="SelectTagHelper"/> <see cref="Type"/> entry and that entry is a non-empty
-        /// <see cref="ICollection{string}"/> instance. Also does nothing if the associated &lt;option&gt; is already
+        /// <see cref="SelectTagHelper"/> <see cref="Type"/> entry and that entry is a non-<c>null</c>
+        /// <see cref="CurrentValues"/> instance. Also does nothing if the associated &lt;option&gt; is already
```
NTaylorMullen: Do we typically mention `Internal` types in doc comments?

NTaylorMullen:
:shipit:

src/Microsoft.AspNet.Mvc.TagHelpers/OptionTagHelper.cs
dougbu: Pretty sure I've seen a few instances of it.

dougbu:
fa8c2ea

PR #3941, opened 2016-01-16
[Fixes #3752] Cleanup Controller invocation pipeline

### LINKED ISSUE #3752: Make IViewComponentActivator abstraction responsible for creation of view components
While the current design allows intercepting the creation of controllers (using the `IControllerActivator` abstraction), this is currently almost impossible with view components. Currently, the `DefaultViewComponentInvoker` is responsible of creating view components by calling directly into the `IServiceProvider` inside its private `CreateComponent` method. Although this method actually calls the existing `Activate` method on the `IViewComponentActivator` abstraction, this abstraction actually expects a created instance; it is not responsible of creating instances, while it should be.

Because of this, intercepting the creation of view components basically means either creating a complete copy of the `DefaultViewComponentInvoker` class, or replacing the `IServiceProvider` with your own. Creating your own copy of `DefaultViewComponentInvoker` is obviously awful, and replacing the default `IServiceProvder` forces a user to replace the complete DI pipeline with its own container, but this isn't always possible, either because the user applies Pure DI (DI without container) or uses a DI library that is incompatible with the DI abstractions given by the new ASP.NET framework.

Intercepting the creation of view component classes however is very important, because -as [the documentation states](https://docs.asp.net/projects/mvc/en/latest/views/view-components.html)- view components are "mini-controller" that will almost always require dependencies of their own.

The addition of this extra 'view components' concept doesn't seem very useful, since we've always been able to do the same with controllers in MVC 5. Since the docs describe them as "mini-controllers", it would be more useful to merge those concepts back together again; this also removes the problem, because controllers already have abstractions that allow intercepting their creation.

Either an `IViewComponentActivator` abstraction should be changed so that it is responsible for creation of components as well (or it should delegate to another abstraction that is responsible of creating it), or the concepts of controllers and view components should be merged together.

javiercn:
* Added a Release method to IControllerActivator
* Changed Create in IControllerActivator to take in a ControllerActionContext
* Move the check to determine if a controller can be instantiated into the controller activator.
* Move logic for disposing controllers into the controller activator and make release on the
  controller factory delegate into the activator.
* Changed release methods to take in a controller context.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/Controllers/DefaultControllerActivator.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Controllers/DefaultControllerActivator.cs
@@ -2,6 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Reflection;
+using Microsoft.AspNetCore.Mvc.Core;
+using Microsoft.AspNetCore.Mvc.Infrastructure;
 using Microsoft.AspNetCore.Mvc.Internal;
 
 namespace Microsoft.AspNetCore.Mvc.Controllers
@@ -19,23 +22,64 @@ namespace Microsoft.AspNetCore.Mvc.Controllers
         /// <param name="typeActivatorCache">The <see cref="ITypeActivatorCache"/>.</param>
         public DefaultControllerActivator(ITypeActivatorCache typeActivatorCache)
         {
+            if (typeActivatorCache == null)
+            {
+                throw new ArgumentNullException(nameof(typeActivatorCache));
+            }
+
             _typeActivatorCache = typeActivatorCache;
         }
+
         /// <inheritdoc />
-        public virtual object Create(ActionContext actionContext, Type controllerType)
+        public virtual object Create(ControllerContext controllerContext)
         {
-            if (actionContext == null)
+            if (controllerContext == null)
             {
-                throw new ArgumentNullException(nameof(actionContext));
+                throw new ArgumentNullException(nameof(controllerContext));
             }
 
-            if (controllerType == null)
+            if (controllerContext.ActionDescriptor == null)
             {
-                throw new ArgumentNullException(nameof(controllerType));
+                throw new ArgumentException(Resources.FormatPropertyOfTypeCannotBeNull(
+                    nameof(ControllerContext.ActionDescriptor),
+                    nameof(ControllerContext)));
             }
 
-            var serviceProvider = actionContext.HttpContext.RequestServices;
-            return _typeActivatorCache.CreateInstance<object>(serviceProvider, controllerType);
+            var controllerTypeInfo = controllerContext.ActionDescriptor.ControllerTypeInfo;
+            if (controllerTypeInfo.IsValueType ||
+                controllerTypeInfo.IsInterface ||
+                controllerTypeInfo.IsAbstract ||
+                (controllerTypeInfo.IsGenericType && controllerTypeInfo.IsGenericTypeDefinition))
+            {
+                var message = Resources.FormatValueInterfaceAbstractOrOpenGenericTypesCannotBeActivated(
+                    controllerTypeInfo.FullName,
+                    GetType().FullName);
+
+                throw new InvalidOperationException(message);
+            }
+
+            var serviceProvider = controllerContext.HttpContext.RequestServices;
+            return _typeActivatorCache.CreateInstance<object>(serviceProvider, controllerTypeInfo.AsType());
+        }
+
+        /// <inheritdoc />
+        public virtual void Release(ControllerContext context, object controller)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            if (controller == null)
+            {
+                throw new ArgumentNullException(nameof(controller));
+            }
+
+            var disposable = controller as IDisposable;
+            if (disposable != null)
+            {
+                disposable.Dispose();
+            }
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Mvc.Core/Controllers/DefaultControllerFactory.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Controllers/DefaultControllerFactory.cs
@@ -31,6 +31,16 @@ namespace Microsoft.AspNetCore.Mvc.Controllers
             IControllerActivator controllerActivator,
             IEnumerable<IControllerPropertyActivator> propertyActivators)
         {
+            if (controllerActivator == null)
+            {
+                throw new ArgumentNullException(nameof(controllerActivator));
+            }
+
+            if (propertyActivators == null)
+            {
+                throw new ArgumentNullException(nameof(propertyActivators));
+            }
+
             _controllerActivator = controllerActivator;
             _propertyActivators = propertyActivators.ToArray();
         }
@@ -61,20 +71,7 @@ namespace Microsoft.AspNetCore.Mvc.Controllers
                     nameof(ControllerContext)));
             }
 
-            var controllerType = context.ActionDescriptor.ControllerTypeInfo.AsType();
-            var controllerTypeInfo = controllerType.GetTypeInfo();
-            if (controllerTypeInfo.IsValueType ||
-                controllerTypeInfo.IsInterface ||
-                controllerTypeInfo.IsAbstract ||
-                (controllerTypeInfo.IsGenericType && controllerTypeInfo.IsGenericTypeDefinition))
-            {
-                var message = Resources.FormatValueInterfaceAbstractOrOpenGenericTypesCannotBeActivated(
-                    controllerType.FullName, 
-                    GetType().FullName);
-                throw new InvalidOperationException(message);
-            }
-
-            var controller = _controllerActivator.Create(context, controllerType);
+            var controller = _controllerActivator.Create(context);
             foreach (var propertyActivator in _propertyActivators)
             {
                 propertyActivator.Activate(context, controller);
@@ -84,14 +81,19 @@ namespace Microsoft.AspNetCore.Mvc.Controllers
         }
 
         /// <inheritdoc />
-        public virtual void ReleaseController(object controller)
+        public virtual void ReleaseController(ControllerContext context, object controller)
         {
-            var disposableController = controller as IDisposable;
-
-            if (disposableController != null)
+            if (context == null)
             {
-                disposableController.Dispose();
+                throw new ArgumentNullException(nameof(context));
             }
+
+            if (controller == null)
+            {
+                throw new ArgumentNullException(nameof(controller));
+            }
+
+            _controllerActivator.Release(context, controller);
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Mvc.Core/Controllers/IControllerActivator.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Controllers/IControllerActivator.cs
@@ -13,7 +13,13 @@ namespace Microsoft.AspNetCore.Mvc.Controllers
         /// <summary>
         /// Creates a controller.
         /// </summary>
-        /// <param name="context">The <see cref="ActionContext"/> for the executing action.</param>
-        object Create(ActionContext context, Type controllerType);
+        /// <param name="context">The <see cref="ControllerContext"/> for the executing action.</param>
+        object Create(ControllerContext context);
+
+        /// <summary>
+        /// Releases a controller.
+        /// </summary>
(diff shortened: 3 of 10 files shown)


javiercn:
@Eilon can you take a look at the public api changes?

javiercn:
PR is not final, cleanup, missing test and docs to come

src/Microsoft.AspNet.Mvc.Core/Controllers/ServiceBasedControllerActivator.cs
```diff
@@ -13,19 +13,26 @@ namespace Microsoft.AspNet.Mvc.Controllers
...
         }
+
+        /// <inheritdoc />
+        public virtual void Release(ControllerContext context, object controller)
+        {
+            var disposable = controller as IDisposable;
+            if (disposable != null)
+            {
+                disposable.Dispose();
+            }
```
rynowak: This method should be a no-op. The DI is responsible for doing disposal if necessary.

rynowak:
:shipit: after removing Dispose from the services-based-thingy

src/Microsoft.AspNet.Mvc.Core/Controllers/ServiceBasedControllerActivator.cs
davidfowl: Yar

src/Microsoft.AspNet.Mvc.Core/Controllers/DefaultControllerActivator.cs
```diff
@@ -19,23 +21,63 @@ public class DefaultControllerActivator : IControllerActivator
...
+            {
+                throw new ArgumentNullException(nameof(typeActivatorCache));
+            }
+
             _typeActivatorCache = typeActivatorCache;
         }
+
         /// <inheritdoc />
-        public virtual object Create(ActionContext actionContext, Type controllerType)
+        public virtual object Create(ControllerContext actionContext)
```
pranavkm: `ControllerContext controllerContext`

src/Microsoft.AspNet.Mvc.Core/Controllers/DefaultControllerActivator.cs
```diff
@@ -19,23 +21,63 @@ public class DefaultControllerActivator : IControllerActivator
...
+
+                throw new InvalidOperationException(message);
             }
 
             var serviceProvider = actionContext.HttpContext.RequestServices;
-            return _typeActivatorCache.CreateInstance<object>(serviceProvider, controllerType);
+            return _typeActivatorCache.CreateInstance<object>(serviceProvider, controllerTypeInfo.AsType());
+        }
+
+        public virtual void Release(ControllerContext context, object controller)
```
pranavkm: `<inheritdoc />`

src/Microsoft.AspNet.Mvc.Core/Controllers/DefaultControllerActivator.cs
```diff
@@ -19,23 +21,63 @@ public class DefaultControllerActivator : IControllerActivator
...
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            if (controller == null)
+            {
+                throw new ArgumentNullException(nameof(controller));
+            }
+
+            var disposable = controller as IDisposable;
```
pranavkm: `(controller as IDisposable)?.Dispose();`

src/Microsoft.AspNet.Mvc.Core/Controllers/ServiceBasedControllerActivator.cs
```diff
@@ -13,19 +13,26 @@ namespace Microsoft.AspNet.Mvc.Controllers
...
             if (actionContext == null)
             {
                 throw new ArgumentNullException(nameof(actionContext));
             }
 
-            if (controllerType == null)
-            {
-                throw new ArgumentNullException(nameof(controllerType));
-            }
+            var controllerType = actionContext.ActionDescriptor.ControllerTypeInfo?.AsType();
```
pranavkm: Should this throw if the `ControllerTypeInfo` is null? `GetRequiredService` might throw, but it might not be a particularly useful message.

test/Microsoft.AspNet.Mvc.Core.Test/Controllers/DefaultControllerActivatorTest.cs
```diff
@@ -81,6 +140,19 @@ public TypeDerivingFromControllerWithServices(TestService service)
             public TestService TestService { get; }
         }
 
+        private IServiceProvider GetServices()
+        {
+            var metadataProvider = new EmptyModelMetadataProvider();
+            var services = new Mock<IServiceProvider>();
```
pranavkm: Using `ServiceCollection` might be a bit cleaner here.

test/Microsoft.AspNet.Mvc.Core.Test/Controllers/DefaultControllerActivatorTest.cs
```diff
@@ -46,16 +98,23 @@ public void Create_TypeActivatesTypesWithServices()
             serviceProvider.Setup(s => s.GetService(typeof(TestService)))
```
pranavkm: Could you modify the tab indentation for this so that it uses VS defaults?

test/Microsoft.AspNet.Mvc.Core.Test/Controllers/DefaultControllerActivatorTest.cs
```diff
@@ -81,6 +140,19 @@ public TypeDerivingFromControllerWithServices(TestService service)
             public TestService TestService { get; }
         }
 
+        private IServiceProvider GetServices()
+        {
+            var metadataProvider = new EmptyModelMetadataProvider();
+            var services = new Mock<IServiceProvider>();
+            services.Setup(s => s.GetService(typeof(IUrlHelper)))
+                    .Returns(Mock.Of<IUrlHelper>());
```
pranavkm: Tab to VS defaults.

test/Microsoft.AspNet.Mvc.Core.Test/Controllers/DefaultControllerFactoryTest.cs
```diff
@@ -226,22 +195,11 @@ public void DefaultControllerFactory_DisposesIDisposableController()
             // Act + Assert
             Assert.False(controller.Disposed);
 
-            factory.ReleaseController(controller);
+            factory.ReleaseController(Mock.Of<ControllerContext>(), controller);
```
pranavkm: `new ControllerContext()`

test/Microsoft.AspNet.Mvc.Core.Test/Controllers/DefaultControllerFactoryTest.cs
```diff
@@ -257,7 +215,12 @@ private IServiceProvider GetServices()
 
         private static DefaultControllerFactory CreateControllerFactory(IControllerActivator controllerActivator = null)
         {
-            controllerActivator = controllerActivator ?? Mock.Of<IControllerActivator>();
+            var controllerActivatorMock = new Mock<IControllerActivator>();
+            controllerActivatorMock
```
pranavkm: Why do you have to set this up to Dispose?

test/Microsoft.AspNet.Mvc.Core.Test/Controllers/ServiceBasedControllerActivatorTest.cs
```diff
@@ -50,13 +55,17 @@ public void Create_ThrowsIfControllerIsNotRegisteredInServiceProvider()
...
+                new RouteData(),
+                new ControllerActionDescriptor
+                {
+                    ControllerTypeInfo = typeof(DIController).GetTypeInfo()
+                }));
 
             // Act and Assert
             var ex = Assert.Throws<InvalidOperationException>(
-                        () => activator.Create(actionContext, typeof(DIController)));
+                        () => activator.Create(actionContext));
```
pranavkm: Indentation.

pranavkm:
:shipit:

src/Microsoft.AspNet.Mvc.Core/Controllers/DefaultControllerActivator.cs
javiercn: I think it's clearer this way.

src/Microsoft.AspNet.Mvc.Core/Controllers/ServiceBasedControllerActivator.cs
javiercn: This was a mistake on my part, It should be .AsType() as in DefaultControllerActivator

test/Microsoft.AspNet.Mvc.Core.Test/Controllers/DefaultControllerActivatorTest.cs
javiercn: This is code moved from DefaultControllerFactoryTest

test/Microsoft.AspNet.Mvc.Core.Test/Controllers/DefaultControllerActivatorTest.cs
javiercn: will do

test/Microsoft.AspNet.Mvc.Core.Test/Controllers/DefaultControllerFactoryTest.cs
javiercn: Because the factory delegates on the activator and the activator is expected to call dispose on the object

pranavkm: We should move the `DefaultControllerFactory_DisposesIDisposableController` test out in that case. We're testing the behavior of the activator, not the controller factory.

javiercn: We are testing that it delegates into the activator

pranavkm: Then the verification needs to be if `controllerActivatorMock.Release` is invoked. You shouldn't need to dispose it.

test/Microsoft.AspNet.Mvc.Core.Test/Controllers/DefaultControllerActivatorTest.cs
javiercn: I think i've done this, I don't know why github still shows it, still is not part of my changes :)

test/Microsoft.AspNet.Mvc.Core.Test/Controllers/DefaultControllerActivatorTest.cs
```diff
@@ -26,16 +29,65 @@ public void Create_CreatesInstancesOfTypes(Type type)
             {
                 RequestServices = serviceProvider.Object
             };
-            var actionContext = new ActionContext(httpContext,
-                                                  new RouteData(),
-                                                  new ActionDescriptor());
+
+            var actionContext = new ControllerContext(
```
rynowak: `context` :laughing:

test/Microsoft.AspNet.Mvc.Core.Test/Controllers/DefaultControllerActivatorTest.cs
```diff
@@ -44,18 +96,25 @@ public void Create_TypeActivatesTypesWithServices()
...
+
             var httpContext = new DefaultHttpContext
             {
                 RequestServices = serviceProvider.Object
             };
-            var actionContext = new ActionContext(httpContext,
-                                                  new RouteData(),
-                                                  new ActionDescriptor());
+
+            var actionContext = new ControllerContext(
```
rynowak: `context`

test/Microsoft.AspNet.Mvc.Core.Test/Controllers/ServiceBasedControllerActivatorTest.cs
```diff
@@ -26,12 +27,16 @@ public void Create_GetsServicesFromServiceProvider()
                 RequestServices = serviceProvider.Object
             };
             var activator = new ServiceBasedControllerActivator();
-            var actionContext = new ActionContext(httpContext,
-                                                  new RouteData(),
-                                                  new ActionDescriptor());
+            var actionContext = new ControllerContext(new ActionContext(
```
rynowak: `context`

rynowak:
:shipit: x2

PR #3942, opened 2016-01-16
[Fixes #3752] Refactor and cleanup view components

### LINKED ISSUE #3752: Make IViewComponentActivator abstraction responsible for creation of view components

javiercn:
* Moved instantiation of view components into DefaultViewComponentActivator
* Introduced IViewComponentFactory to handling setup of new view component instances.
* Added a release method on IViewComponentActivator for handling tear down of view
  component instances.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/DependencyInjection/MvcViewFeaturesMvcCoreBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/DependencyInjection/MvcViewFeaturesMvcCoreBuilderExtensions.cs
@@ -118,8 +118,10 @@ namespace Microsoft.Extensions.DependencyInjection
             //
             // View Components
             //
+            
             // These do caching so they should stay singleton
             services.TryAddSingleton<IViewComponentSelector, DefaultViewComponentSelector>();
+            services.TryAddSingleton<IViewComponentFactory, DefaultViewComponentFactory>();
             services.TryAddSingleton<IViewComponentActivator, DefaultViewComponentActivator>();
             services.TryAddSingleton<
                 IViewComponentDescriptorCollectionProvider,
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Properties/Resources.Designer.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Properties/Resources.Designer.cs
@@ -858,6 +858,22 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
             return string.Format(CultureInfo.CurrentCulture, GetString("ViewComponent_AmbiguousMethods"), p0, p1, p2);
         }
 
+        /// <summary>
+        /// The type '{0}' cannot be activated by '{1}' because it is either a value type, an interface, an abstract class or an open generic type.
+        /// </summary>
+        internal static string ValueInterfaceAbstractOrOpenGenericTypesCannotBeActivated
+        {
+            get { return GetString("ValueInterfaceAbstractOrOpenGenericTypesCannotBeActivated"); }
+        }
+
+        /// <summary>
+        /// The type '{0}' cannot be activated by '{1}' because it is either a value type, an interface, an abstract class or an open generic type.
+        /// </summary>
+        internal static string FormatValueInterfaceAbstractOrOpenGenericTypesCannotBeActivated(object p0, object p1)
+        {
+            return string.Format(CultureInfo.CurrentCulture, GetString("ValueInterfaceAbstractOrOpenGenericTypesCannotBeActivated"), p0, p1);
+        }
+
         private static string GetString(string name, params string[] formatterNames)
         {
             var value = _resourceManager.GetString(name);
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Resources.resx
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Resources.resx
@@ -277,4 +277,7 @@
   <data name="ViewComponent_AmbiguousMethods" xml:space="preserve">
     <value>View component '{0}' must have exactly one public method named '{1}' or '{2}'.</value>
   </data>
+  <data name="ValueInterfaceAbstractOrOpenGenericTypesCannotBeActivated" xml:space="preserve">
+    <value>The type '{0}' cannot be activated by '{1}' because it is either a value type, an interface, an abstract class or an open generic type.</value>
+  </data>
 </root>
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentActivator.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentActivator.cs
@@ -2,9 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Concurrent;
 using System.Reflection;
-using Microsoft.Extensions.Internal;
+using Microsoft.AspNetCore.Mvc.Internal;
+using Microsoft.AspNetCore.Mvc.ViewFeatures;
 
 namespace Microsoft.AspNetCore.Mvc.ViewComponents
 {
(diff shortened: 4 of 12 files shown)


javiercn:
@Eilon can you take a look at the public api changes?

src/Microsoft.AspNet.Mvc.ViewFeatures/DependencyInjection/MvcViewFeaturesMvcCoreBuilderExtensions.cs
```diff
@@ -119,9 +119,11 @@ internal static void AddViewServices(IServiceCollection services)
             //
             // View Components
             //
+            services.TryAddSingleton<IViewComponentActivator, DefaultViewComponentActivator>();
+            
             // These do caching so they should stay singleton
             services.TryAddSingleton<IViewComponentSelector, DefaultViewComponentSelector>();
-            services.TryAddSingleton<IViewComponentActivator, DefaultViewComponentActivator>();
+            services.TryAddSingleton<IViewComponentFactory, DefaultViewComponentFactory>();
```
rynowak: The activator is the one that belongs with this comment. Personally I'm game for just deleting the comments. Everything for the most part in the framework is singleton by default.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentActivator.cs
```diff
@@ -18,49 +18,68 @@ namespace Microsoft.AspNet.Mvc.ViewComponents
...
+                var message = Resources.FormatValueInterfaceAbstractOrOpenGenericTypesCannotBeActivated(
+                    componentType.FullName,
+                    GetType().FullName);
+
+                throw new InvalidOperationException(message);
             }
+
+            var viewComponent = _typeActivatorCache.CreateInstance<object>(
+            context.ViewContext.HttpContext.RequestServices,
+            context.ViewComponentDescriptor.Type);
```
rynowak: indent

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentActivator.cs
```diff
@@ -18,49 +18,68 @@ namespace Microsoft.AspNet.Mvc.ViewComponents
...
-
             if (context == null)
             {
                 throw new ArgumentNullException(nameof(context));
             }
 
-            var propertiesToActivate = _injectActions.GetOrAdd(
-                viewComponent.GetType(),
-                _getPropertiesToActivate);
+            var componentType = context.ViewComponentDescriptor.Type.GetTypeInfo();
```
rynowak: weren't you going to change this to be a typeinfo?

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentFactory.cs
```diff
@@ -0,0 +1,85 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Concurrent;
+using System.Reflection;
+using Microsoft.Extensions.Internal;
+
+namespace Microsoft.AspNet.Mvc.ViewComponents
+{
+    public class DefaultViewComponentFactory : IViewComponentFactory
```
rynowak: Docs

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/IViewComponentFactory.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Microsoft.AspNet.Mvc.ViewComponents
+{
+    public interface IViewComponentFactory
+    {
+        object CreateViewComponent(ViewComponentContext context);
+
+        void ReleaseViewComponent(ViewComponentContext context, object component);
```
rynowak: docs

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewComponentResultTest.cs
```diff
@@ -572,30 +573,30 @@ public IEnumerable<ViewComponentDescriptor> GetViewComponents()
...
+            using (var reader = new StreamReader(response.Body))
             {
-                return new HtmlString("Hello, " + name);
+                return reader.ReadToEnd();
             }
         }
+    }
 
-        private class AsyncTextViewComponent : ViewComponent
+    public class TextViewComponent : ViewComponent
```
rynowak: private

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewComponentResultTest.cs
```diff
@@ -572,30 +573,30 @@ public IEnumerable<ViewComponentDescriptor> GetViewComponents()
...
-            public Task<HtmlString> InvokeAsync(string name)
-            {
-                return Task.FromResult(new HtmlString("Hello-Async, " + name));
-            }
+            return new HtmlString("Hello, " + name);
         }
+    }
 
-        private static string ReadBody(HttpResponse response)
+    public class AsyncTextViewComponent : ViewComponent
```
rynowak: private

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentActivator.cs
```diff
@@ -18,49 +18,68 @@ namespace Microsoft.AspNet.Mvc.ViewComponents
...
 
-            for (var i = 0; i < propertiesToActivate.Length; i++)
+            if (!componentType.IsClass ||
+                !componentType.IsPublic ||
```
rynowak: we shouldn't be checking `IsPublic` here. We don't discover public types by default. But there's nothing stopping it from working.

rynowak: This check is different from the code in the controller factory. Can these be the same? why/why not? (I have more faith in the code in controller factory).

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentFactory.cs
rynowak: Needs tests too

rynowak:
:watch:

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentActivator.cs
javiercn: Decided not to do it as part of this change

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentActivator.cs
javiercn: I got this from looking at

https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/ViewComponentConventions.cs#L75

If we change it here, we might want to change it there. This would mean that ViewComponents could be non public classes (like internal classes and nested classes) which I'm not sure if we want to allow

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewComponentResultTest.cs
javiercn: needed to change this as part of the check that I introduced on DefaultViewComponentActivator

test/Microsoft.AspNet.Mvc.ViewFeatures.Test/ViewComponentResultTest.cs
javiercn: needed to change this as part of the check that I introduced on DefaultViewComponentActivator, I think this makes the test more compliant with the expectations we have for view components

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentActivator.cs
rynowak: > We don't discover public types by default. 

This should stay. 

> But there's nothing stopping it from working in the factory.

If you use extensibility to tell us to use some random private class as a VC, then it should be supported by the factory.


> If we change it here, we might want to change it there.

We should not change `ViewComponentConventions`. What I'm saying is that the factory **could** instantiate any instantiable type. We shouldn't put a limitation in the factory.

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentActivator.cs
rynowak: :ok:

javiercn:
@rynowak updated

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentFactory.cs
```diff
@@ -0,0 +1,90 @@
...
+    /// <summary>
+    /// Default implementation for <see cref="IViewComponentFactory"/>.
+    /// </summary>
+    public class DefaultViewComponentFactory : IViewComponentFactory
+    {
+        private readonly IViewComponentActivator _activator;
+        private readonly Func<Type, PropertyActivator<ViewComponentContext>[]> _getPropertiesToActivate;
+        private readonly ConcurrentDictionary<Type, PropertyActivator<ViewComponentContext>[]> _injectActions;
+
+        public DefaultViewComponentFactory(IViewComponentActivator activator)
```
rynowak: `/// <inheritdoc />`

rynowak: Docs

rynowak:
:shipit:

PR #3955, opened 2016-01-20
[Fixes #3874][Design] Null passed as arguments to controller method parameters when no InputFormatter matches

### LINKED ISSUE #3874: Null passed as arguments to controller method parameters when no InputFormatter matches
if the request's content-type is missing or not matching any InputFormatter then controller methods are passed null as arguments where data need to be deserialized. In the example below 'employee' is then null when the method is called. Instead the code 415 should be returned and the controller method should NOT be called.

```public IActionResult UpdateEmployee(uint employeerId, [FromBody] Employee employee)```

P.S:
A workaround is to create a middleware the code below to force the error it is not very pretty and breaks if other content types than json is added because I am not aware how to check properly any of the registered InputFormatter's match rather then hardcoding it.

```
       async Task Invoke(HttpContext context)
        {
          string method = context.Request.Method;
          if (method == "PUT" || method == "POST" || method == "PATCH")
          {
                RequestHeaders typedHeaders = context.Request.GetTypedHeaders();
                MediaTypeHeaderValue contentType = typedHeaders.ContentType;
                if (contentType==null || contentType.SubType!=@"json")
                {
                    context.Response.StatusCode = Status415UnsupportedMediaType;
                    return;
                }
          }

          await next.Invoke(context);
        }
```

### PR DESCRIPTION
**Pull request by javiercn**
* Add an UnsupportedContentType to the ModelState dictionary when no formatter can read the body.
* Add a filter to the pipeline that searches for that specific exception and transforms the response into 415.

### DIFF
--- a/test/Microsoft.AspNetCore.Mvc.FunctionalTests/WebApiCompatShimActionSelectionTest.cs
+++ b/test/Microsoft.AspNetCore.Mvc.FunctionalTests/WebApiCompatShimActionSelectionTest.cs
@@ -3,6 +3,7 @@
 
 using System.Net;
 using System.Net.Http;
+using System.Text;
 using System.Threading.Tasks;
 using Newtonsoft.Json;
 using Xunit;
@@ -290,9 +291,6 @@ namespace Microsoft.AspNetCore.Mvc.FunctionalTests
         [InlineData("GET", "api/Admin/Test?name=mario&ssn=123456", "GetUserByNameAndSsn")]
         [InlineData("GET", "api/Admin/Test?name=mario&ssn=123456&age=3", "GetUserByNameAgeAndSsn")]
         [InlineData("GET", "api/Admin/Test/5?random=9", "GetUser")]
-        [InlineData("POST", "api/Admin/Test", "PostUser")]
-        [InlineData("POST", "api/Admin/Test?name=mario&age=10", "PostUserByNameAndAge")]
-
         // Note: Normally the following would not match DeleteUserByIdAndOptName because it has 'id' and 'age' as parameters while the DeleteUserByIdAndOptName action has 'id' and 'name'.
         // However, because the default value is provided on action parameter 'name', having the 'id' in the request was enough to match the action.
         [InlineData("DELETE", "api/Admin/Test/6?age=10", "DeleteUserByIdAndOptName")]
@@ -317,6 +315,25 @@ namespace Microsoft.AspNetCore.Mvc.FunctionalTests
         {
             // Arrange
             var request = new HttpRequestMessage(new HttpMethod(httpMethod), "http://localhost/" + requestUrl);
+
+            // Act
+            var response = await Client.SendAsync(request);
+
+            // Assert
+            var data = Assert.Single(response.Headers.GetValues("ActionSelection"));
+            var result = JsonConvert.DeserializeObject<ActionSelectionResult>(data);
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            Assert.Equal(expectedActionName, result.ActionName);
+        }
+
+        [Theory]
+        [InlineData("api/Admin/Test", "PostUser")]
+        [InlineData("api/Admin/Test?name=mario&age=10", "PostUserByNameAndAge")]
+        public async Task LegacyActionSelection_OverloadedAction_WithUnnamedAction(string requestUrl, string expectedActionName)
+        {
+            // Arrange
+            var request = new HttpRequestMessage(HttpMethod.Post, "http://localhost/" + requestUrl);
+            request.Content = new StringContent("{}", Encoding.UTF8, "application/json");
 
             // Act
             var response = await Client.SendAsync(request);
@@ -378,7 +395,6 @@ namespace Microsoft.AspNetCore.Mvc.FunctionalTests
         [InlineData("GET", "api/Blog/Test/GetUser/4?id=3", "GetUser")]
         [InlineData("GET", "api/Blog/Test/GetUserByNameAgeAndSsn?name=user&age=90&ssn=123456789", "GetUserByNameAgeAndSsn")]
         [InlineData("GET", "api/Blog/Test/GetUserByNameAndSsn?name=user&ssn=123456789", "GetUserByNameAndSsn")]
-        [InlineData("POST", "api/Blog/Test/PostUserByNameAndAddress?name=user", "PostUserByNameAndAddress")]
         public async Task LegacyActionSelection_RouteWithActionName(string httpMethod, string requestUrl, string expectedActionName)
         {
             // Arrange
@@ -394,6 +410,25 @@ namespace Microsoft.AspNetCore.Mvc.FunctionalTests
             Assert.Equal(expectedActionName, result.ActionName);
         }
 
+        [Fact]
+        public async Task LegacyActionSelection_RouteWithActionName()
+        {
+            // Arrange
+            var request = new HttpRequestMessage(
+                HttpMethod.Post,
+                "http://localhost/api/Blog/Test/PostUserByNameAndAddress?name=user");
+            request.Content = new StringContent("{}", Encoding.UTF8, "application/json");
+
+            // Act
+            var response = await Client.SendAsync(request);
+
+            // Assert
+            var data = Assert.Single(response.Headers.GetValues("ActionSelection"));
+            var result = JsonConvert.DeserializeObject<ActionSelectionResult>(data);
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            Assert.Equal("PostUserByNameAndAddress", result.ActionName);
+        }
+
         [Theory]
         [InlineData("GET", "api/Blog/Test/getusers", "GetUsers")]
         [InlineData("GET", "api/Blog/Test/getuseR/1", "GetUser")]
@@ -401,7 +436,6 @@ namespace Microsoft.AspNetCore.Mvc.FunctionalTests
         [InlineData("GET", "api/Blog/Test/GetUser/4?Id=3", "GetUser")]
         [InlineData("GET", "api/Blog/Test/GetUserByNameAgeandSsn?name=user&age=90&ssn=123456789", "GetUserByNameAgeAndSsn")]
         [InlineData("GET", "api/Blog/Test/getUserByNameAndSsn?name=user&ssn=123456789", "GetUserByNameAndSsn")]
-        [InlineData("POST", "api/Blog/Test/PostUserByNameAndAddress?name=user", "PostUserByNameAndAddress")]
         public async Task LegacyActionSelection_RouteWithActionName_Casing(string httpMethod, string requestUrl, string expectedActionName)
         {
             // Arrange
@@ -415,6 +449,25 @@ namespace Microsoft.AspNetCore.Mvc.FunctionalTests
             var result = JsonConvert.DeserializeObject<ActionSelectionResult>(data);
             Assert.Equal(HttpStatusCode.OK, response.StatusCode);
             Assert.Equal(expectedActionName, result.ActionName);
+        }
+
+        [Fact]
+        public async Task LegacyActionSelection_RouteWithActionName_Casing()
+        {
+            // Arrange
+            var request = new HttpRequestMessage(
+                HttpMethod.Post,
+                "http://localhost/api/Blog/Test/PostUserByNameAndAddress?name=user");
+            request.Content = new StringContent("{}", Encoding.UTF8, "application/json");
+
+            // Act
+            var response = await Client.SendAsync(request);
+
+            // Assert
+            var data = Assert.Single(response.Headers.GetValues("ActionSelection"));
+            var result = JsonConvert.DeserializeObject<ActionSelectionResult>(data);
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            Assert.Equal("PostUserByNameAndAddress", result.ActionName);
         }
 
         [Theory]
@@ -438,19 +491,36 @@ namespace Microsoft.AspNetCore.Mvc.FunctionalTests
             Assert.Equal(expectedActionName, result.ActionName);
         }
 
-
         [Theory]
         [InlineData("GET", "api/Admin/ParameterAttribute/2", "GetUser")]
         [InlineData("GET", "api/Admin/ParameterAttribute?id=2", "GetUser")]
         [InlineData("GET", "api/Admin/ParameterAttribute?myId=2", "GetUserByMyId")]
-        [InlineData("POST", "api/Admin/ParameterAttribute/3?name=user", "PostUserNameFromUri")]
-        [InlineData("POST", "api/Admin/ParameterAttribute/3", "PostUserNameFromBody")]
         [InlineData("DELETE", "api/Admin/ParameterAttribute/3?name=user", "DeleteUserWithNullableIdAndName")]
         [InlineData("DELETE", "api/Admin/ParameterAttribute?address=userStreet", "DeleteUser")]
         public async Task LegacyActionSelection_ModelBindingParameterAttribute_AreAppliedWhenSelectingActions(string httpMethod, string requestUrl, string expectedActionName)
         {
             // Arrange
             var request = new HttpRequestMessage(new HttpMethod(httpMethod), "http://localhost/" + requestUrl);
+            request.Content = new StringContent("{}", Encoding.UTF8, "application/json");
+
+            // Act
+            var response = await Client.SendAsync(request);
+
+            // Assert
+            var data = Assert.Single(response.Headers.GetValues("ActionSelection"));
+            var result = JsonConvert.DeserializeObject<ActionSelectionResult>(data);
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            Assert.Equal(expectedActionName, result.ActionName);
+        }
+
+        [Theory]
+        [InlineData("api/Admin/ParameterAttribute/3?name=user", "PostUserNameFromUri")]
+        [InlineData("api/Admin/ParameterAttribute/3", "PostUserNameFromBody")]
+        public async Task LegacyActionSelection_Post_ModelBindingParameterAttribute_AreAppliedWhenSelectingActions(string requestUrl, string expectedActionName)
+        {
+            // Arrange
+            var request = new HttpRequestMessage(HttpMethod.Post, "http://localhost/" + requestUrl);
+            request.Content = new StringContent("{}", Encoding.UTF8, "application/json");
 
             // Act
             var response = await Client.SendAsync(request);
--- a/test/WebSites/BasicWebSite/Controllers/ActionConstraints/ConsumesAttribute_PassThroughController.cs
+++ b/test/WebSites/BasicWebSite/Controllers/ActionConstraints/ConsumesAttribute_PassThroughController.cs
@@ -10,9 +10,9 @@ namespace BasicWebSite.Controllers.ActionConstraints
     public class ConsumesAttribute_PassThroughController : Controller
     {
         [Consumes("application/json")]
-        public Product CreateProduct([FromBody] Product_Json jsonInput)
+        public IActionResult CreateProduct(Product_Json jsonInput)
         {
-            return jsonInput;
+            return Content("ConsumesAttribute_PassThrough_Product_Json");
         }
     }
 }
--- a/test/WebSites/BasicWebSite/Controllers/ActionConstraints/ConsumesAttribute_WithFallbackActionController.cs
+++ b/test/WebSites/BasicWebSite/Controllers/ActionConstraints/ConsumesAttribute_WithFallbackActionController.cs
@@ -10,20 +10,22 @@ namespace BasicWebSite.Controllers.ActionConstraints
     public class ConsumesAttribute_WithFallbackActionController : Controller
     {
         [Consumes("application/json")]
-        public Product CreateProduct([FromBody] Product_Json jsonInput)
+        [ActionName("CreateProduct")]
+        public IActionResult CreateProductJson()
         {
-            return jsonInput;
+            return Content("CreateProduct_Product_Json");
         }
 
         [Consumes("application/xml")]
-        public Product CreateProduct([FromBody] Product_Xml xmlInput)
+        [ActionName("CreateProduct")]
+        public IActionResult CreateProductXml()
         {
-            return xmlInput;
+            return Content("CreateProduct_Product_Xml");
         }
 
-        public Product CreateProduct([FromBody] Product_Text defaultInput)
+        public IActionResult CreateProduct()
         {
-            return defaultInput;
+            return Content("CreateProduct_Product_Text");
         }
     }
 }
--- a/test/WebSites/FiltersWebSite/Controllers/JsonOnlyController.cs
+++ b/test/WebSites/FiltersWebSite/Controllers/JsonOnlyController.cs
@@ -44,12 +44,14 @@ namespace FiltersWebSite.Controllers
 
             public void OnResultExecuting(ResultExecutingContext context)
             {
-                var options = context.HttpContext.RequestServices.GetRequiredService<IOptions<MvcOptions>>();
-                var jsonFormatter = options.Value.OutputFormatters.OfType<JsonOutputFormatter>().Single();
-
                 // Update the output formatter collection to only return JSON. 
-                var result = (ObjectResult)context.Result;
-                result.Formatters.Add(jsonFormatter);
+                if (context.Result is ObjectResult)
+                {
+                    var options = context.HttpContext.RequestServices.GetRequiredService<IOptions<MvcOptions>>();
+                    var jsonFormatter = options.Value.OutputFormatters.OfType<JsonOutputFormatter>().Single();
+                    var result = (ObjectResult)context.Result;
+                    result.Formatters.Add(jsonFormatter);
+                }
             }
         }
     }
--- a/test/WebSites/FormatterWebSite/Controllers/InputFormatterController.cs
+++ b/test/WebSites/FormatterWebSite/Controllers/InputFormatterController.cs
@@ -9,35 +9,6 @@ namespace FormatterWebSite.Controllers
 {
     public class InputFormatterController : Controller
     {
-        [HttpPost]
-        public object ActionHandlesError([FromBody] DummyClass dummy)
-        {
-            if (!ModelState.IsValid)
-            {
-                // Body model binder normally reports errors for parameters using the empty name.
-                var parameterBindingErrors = ModelState["dummy"]?.Errors ?? ModelState[string.Empty]?.Errors;
-                if (parameterBindingErrors != null && parameterBindingErrors.Count != 0)
-                {
-                    return new ErrorInfo
-                    {
-                        ActionName = "ActionHandlesError",
-                        ParameterName = "dummy",
-                        Errors = parameterBindingErrors.Select(x => x.ErrorMessage).ToList(),
-                        Source = "action"
-                    };
-                }
-            }
-
-            return dummy;
-        }
-
-        [HttpPost]
-        [ValidateBodyParameter]
-        public object ActionFilterHandlesError([FromBody] DummyClass dummy)
-        {
-            return dummy;
-        }
-
         public IActionResult ReturnInput([FromBody] string test)
         {
             if (!ModelState.IsValid)
--- a/src/Microsoft.AspNetCore.Mvc.Core/Internal/MvcCoreMvcOptionsSetup.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Internal/MvcCoreMvcOptionsSetup.cs
@@ -48,6 +48,9 @@ namespace Microsoft.AspNetCore.Mvc.Internal
             options.ModelBinders.Add(new GenericModelBinder());
             options.ModelBinders.Add(new MutableObjectModelBinder());
 
+            // Set up filters
+            options.Filters.Add(new UnsupportedContentTypeFilter());
+
             // Set up default output formatters.
             options.OutputFormatters.Add(new HttpNoContentOutputFormatter());
             options.OutputFormatters.Add(new StringOutputFormatter());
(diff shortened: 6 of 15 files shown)

### CODE REVIEW

**Comment by javiercn** [comment]
@rynowak I think this is what we talked about, It needs design refinements and cleanup, but I think this is a good start point

**Comment by dougbu** [comment]
@javiercn how does this fix relate to #3874? This PR hides a framework-provided error message from most users but doesn't change MVC's visible behaviour otherwise. Seems like you're fixing another problem in the original poster's scenario. Might need a new bug or to clarify what "fix" means here.

**Comment by dougbu** [comment]
... Note I'm not objecting to this cleanup. Just thinking it's more similar to #3215.

**Comment by dougbu** [comment]
Separately: Are you also working to clean up the other error case like this one? MVC adds an error message to the `ModelStateDictionary` when the chosen `InputFormatter` has no supported encodings. See https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs#L156-L159. In that case, may make sense to simply `throw` if the `SupportedEncodings` list is empty.

(Hmm, odd that `InputFormatter.SelectCharacterEncoding()` silently uses its default character encoding even if the request specifically identified its content differently.)

File: samples/MvcSandbox/Controllers/HomeController.cs [hunk 1]
```diff
@@ -11,5 +11,11 @@ public IActionResult Index()
         {
             return View();
         }
+
+        [HttpPost]
+        public IActionResult Custom([FromBody] int i)
+        {
+            return Ok(i);
+        }
```
rynowak: Revert before checking in

src/Microsoft.AspNet.Mvc.Core/DependencyInjection/MvcCoreMvcCoreBuilderExtensions.cs
```diff
@@ -94,6 +95,22 @@ internal static void AddAuthorizationServices(IServiceCollection services)
...
+            }
+
+            AddFilterServices(builder.Services);
+            return builder;
+        }
+
+        internal static void AddFilterServices(IServiceCollection services)
+        {
+            services.TryAddSingleton<UnsupportedContentTypeFilter>();
+        }
```
rynowak: What is this stuff for?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/BodyModelBinder.cs
```diff
@@ -84,9 +84,15 @@ public Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindingContex
 
             if (formatter == null)
             {
-                var unsupportedContentType = Resources.FormatUnsupportedContentType(
+                var message = Resources.FormatUnsupportedContentType(
                     bindingContext.OperationBindingContext.HttpContext.Request.ContentType);
-                bindingContext.ModelState.AddModelError(modelBindingKey, unsupportedContentType);
+
+                var unsupportedContentType = new UnsupportedContentTypeException(message);
```
rynowak: `exception`?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/UnsupportedContentTypeFilter.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            var modelState = context.ModelState;
+            if (modelState.IsValid)
+            {
+                return false;
+            }
+
+            var unsupportedContentTypeExceptions = modelState.SelectMany(
+                kvp => kvp.Value.Errors.Select(e => new KeyValuePair<string, Exception>(kvp.Key, e.Exception))
+                    .Where(e => e.Value != null && e.Value is UnsupportedContentTypeException));
```
rynowak: don't use linq here #allocations

src/Microsoft.AspNet.Mvc.Core/MvcOptions.cs
```diff
@@ -23,7 +23,10 @@ public MvcOptions()
         {
             CacheProfiles = new Dictionary<string, CacheProfile>(StringComparer.OrdinalIgnoreCase);
             Conventions = new List<IApplicationModelConvention>();
-            Filters = new FilterCollection();
+            Filters = new FilterCollection()
+            {
+                new UnsupportedContentTypeFilter()
+            };
```
rynowak: Don't add this by default, make it part of the options-setup

src/Microsoft.AspNet.Mvc.Core/ModelBinding/UnsupportedContentTypeFilter.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Mvc.Filters;
+
+namespace Microsoft.AspNet.Mvc.ModelBinding
+{
+    public class UnsupportedContentTypeFilter : IActionFilter
```
rynowak: Docs

src/Microsoft.AspNet.Mvc.Core/ModelBinding/UnsupportedContentTypeFilter.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        public void OnActionExecuting(ActionExecutingContext context)
+        {
+            if (HasUnsupportedContentTypeError(context))
+            {
+                context.Result = new HttpStatusCodeResult(415);
```
rynowak: `StatusCodes.<something>`

rynowak:
@dougbu - I think the issues around supported encodings need some more discussion. I think we can use the same mechanism, but that it's worth a discussion.

-------------------

A few points:

- Some formatters don't use encodings, they are for binary formats (more common with output). There needs to be a way to express that.

- Defaulting to `SupportedEncodings[0]` is really just our way of saying "utf-8 by default" in practice.

- The whole `SupportedEncodings` thing is pretty low value, but @blowdart wants to keep it since the default works for 99% of users.

-------------

Once you do these things, it's pretty easy to see how this leads you down a path of being able to use the same mechanism for reporting an issue.

`SelectCharacterEncoding` can just throw if the `charset` was specified but not a match. If `SupportedEncodings` is empty, we ignore `charset`. If `charset` was not specified we assume utf-8.

rynowak:
:watch: - design looks like it's on track

dougbu:
@rynowak I'm fine w/ the approach for supported encodings that you outlined [here](https://github.com/aspnet/Mvc/pull/3955#issuecomment-173310276) with one proviso:

> Right now that works when `SupportedEncodings` is empty.

Things don't work in that case today. `ModelBinding.IsValid` will be `false`. If we remove [the addition](https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/Formatters/InputFormatter.cs#L157) to the `ModelStateDictionary`, we'd also be removing the last framework-provided error message user's can't override (hurrah) and not covered in other bugs. Even better the "don't use encodings" scenario would be supported. (Perhaps not the most discoverable approach but it would work.)

Do we need a separate bug to track supporting this scenario?

rynowak:
Discussed with @javiercn - going to open a separate issue for this.

> Things don't work in that case today. ModelBinding.IsValid will be false. If we remove the addition to the ModelStateDictionary, we'd also be removing the last framework-provided error message user's can't override (hurrah) and not covered in other bugs. 

You might still remove the filter, in which case this would be just another modelstate error.

javiercn:
@rynowak Updated

javiercn:
@rynowak Updated

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
```diff
@@ -869,6 +880,55 @@ public void Reset()
...
+            /// <inheritdoc />
+            public bool MoveNext()
+            {
+                return _enumerator.MoveNext();
+            }
+
+            /// <inheritdoc />
+            public void Reset()
+            {
+            }
```
rynowak: `_enumerator.Reset()`

src/Microsoft.AspNet.Mvc.WebApiCompatShim/WebApiCompatShimOptionsSetup.cs
```diff
@@ -25,6 +27,8 @@ public void Configure(MvcOptions options)
             options.Conventions.Add(new WebApiRoutesApplicationModelConvention(area: DefaultAreaName));
 
             // Add an action filter for handling the HttpResponseException.
+            var filter = options.Filters.FirstOrDefault(f => f is UnsupportedContentTypeFilter);
+            options.Filters.Remove(filter);
```
rynowak: Undo this.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/BodyModelBinderTests.cs
```diff
@@ -203,7 +203,7 @@ public class BodyModelBinderTests
             // Key is empty because this was a top-level binding.
             var entry = Assert.Single(bindingContext.ModelState);
             Assert.Equal(string.Empty, entry.Key);
-            var errorMessage = Assert.Single(entry.Value.Errors).ErrorMessage;
+            var errorMessage = Assert.Single(entry.Value.Errors).Exception.Message;
```
rynowak: Isn't errormessage also set to this value?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/BodyModelBinder.cs
```diff
@@ -84,9 +84,11 @@ public Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindingContex
 
             if (formatter == null)
             {
-                var unsupportedContentType = Resources.FormatUnsupportedContentType(
+                var message = Resources.FormatUnsupportedContentType(
                     bindingContext.OperationBindingContext.HttpContext.Request.ContentType);
-                bindingContext.ModelState.AddModelError(modelBindingKey, unsupportedContentType);
+
+                var exception = new UnsupportedContentTypeException(message);
+                bindingContext.ModelState.AddModelError(modelBindingKey, exception, bindingContext.ModelMetadata);
```
rynowak: Is there an overload that also lets you specify an error message?

test/WebSites/BasicWebSite/Controllers/ActionConstraints/ConsumesAttribute_WithFallbackActionController.cs
```diff
@@ -10,20 +10,20 @@ namespace BasicWebSite.Controllers.ActionConstraints
...
         [Consumes("application/xml")]
-        public Product CreateProduct([FromBody] Product_Xml xmlInput)
+        public IActionResult CreateProduct(Product_Xml xmlInput)
         {
-            return xmlInput;
+            return Content("CreateProduct_Product_Xml");
         }
 
-        public Product CreateProduct([FromBody] Product_Text defaultInput)
+        public IActionResult CreateProduct(Product_Text defaultInput)
```
rynowak: You don't really even need a paraemter here.

test/WebSites/FiltersWebSite/Controllers/JsonOnlyController.cs
```diff
@@ -48,8 +48,11 @@ public void OnResultExecuting(ResultExecutingContext context)
                 var jsonFormatter = options.Value.OutputFormatters.OfType<JsonOutputFormatter>().Single();
```
rynowak: Move into the `if`

rynowak:
:watch: for small comments

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
javiercn: There is no Reset() on the underlying enumerator. I can change it to throw an exception()

dougbu: How is that possible? `Reset()` is part of the `IEnumerator` interface (and therefore `IEnumerator<T>`.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/BodyModelBinder.cs
javiercn: No, the only overload for exceptions is that one

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
rynowak: No worries then. I was just being pedantic. The existence of `Reset()` is for COM interop.

@dougbu it's probably an explicit implementation of `Reset()` on the underlying type.

test/Microsoft.AspNet.Mvc.Core.Test/ModelBinding/BodyModelBinderTests.cs
javiercn: image

src/Microsoft.AspNet.Mvc.Core/ModelBinding/BodyModelBinder.cs
dougbu: 1. We should not place `string`s from our resources directly into a `ModelError`. That leads to more framework-provided error messages that users can't override, something we've recently _fixed_. Extend `IModelBindingMessageProvider` and so on to allow overrides, get the message from somewhere else (that users can override e.g. a `DataAnnotations` attribute), or don't call the `string` overloads of `[Try]AddModelError()`.
2. If it's necessary to provide both a message and an `Exception`, just call `[Try]AddModelError()` twice with the same `key`.

dougbu: Put another way, excellent that you're not calling `AddModelError()` with a `string` message any more.

javiercn: @dougbu This is not an error destined to end users. End users/developers will see a 415 (Unsupported media type) response. @rynowak thoughts?

rynowak: Yeah, I'm fine with just the exception and no message since we're changing the default behavior.

javiercn: For 2 I don't think it's necessary to provide both. The exception will suffice as it's not an error users will make. (It's a client error (browser, API client).

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
```diff
@@ -657,8 +657,19 @@ public bool TryGetValue(string key, out ModelStateEntry value)
...
 
+        /// <summary>
+        /// Returns an enumerator that iterates through the <see cref="ModelStateDictionary"/>.
+        /// </summary>
+        /// <returns>
+        /// A <see cref="ModelStateDictionary.Enumerator"/> structure for the <see cref="ModelStateDictionary"/>.
+        /// </returns>
+        public Enumerator GetEnumerator()
+        {
+            return new Enumerator(_innerDictionary.GetEnumerator());
```
dougbu: Think I'm missing something. How does wrapping an allocated `IEnumerator` in a `struct` help?

src/Microsoft.AspNet.Mvc.Core/ModelBinding/BodyModelBinder.cs
dougbu: Sounds good. One clarification:
> it's not an error users will make

That's not relevant because it is an error users could _see_ in their browser whenever `UnsupportedContentTypeFilter` is disabled. `ModelError.ErrorMessage` is for display to users regardless of who or what caused the error.

src/Microsoft.AspNet.Mvc/MvcServiceCollectionExtensions.cs
```diff
@@ -44,7 +44,6 @@ public static IMvcBuilder AddMvc(this IServiceCollection services, Action<MvcOpt
 
             builder.AddApiExplorer();
             builder.AddAuthorization();
-
             // Order added affects options setup order
```
dougbu: Curious: Why is this file changing?

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
dougbu: ... Not seeing a reason for any change to this class.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/UnsupportedContentTypeException.cs
```diff
@@ -0,0 +1,29 @@
...
+using System;
+using Microsoft.AspNet.Mvc.Formatters;
+
+namespace Microsoft.AspNet.Mvc.ModelBinding
+{
+    /// <summary>
+    /// The <see cref="Exception"/> that is added to model state when a model binder for the body of the request is
+    /// unable to understand the request content type header.
+    /// </summary>
+    public class UnsupportedContentTypeException : Exception
```
dougbu: Suggest moving this to the `Microsoft.AspNet.Mvc.Core` package and the `Microsoft.AspNet.Mvc.Internal` namespace. Doesn't seem to be an abstraction users should care about.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/UnsupportedContentTypeFilter.cs
```diff
@@ -0,0 +1,55 @@
...
+    /// with an Unsupported Media Type (415) response.
+    /// </summary>
+    public class UnsupportedContentTypeFilter : IActionFilter
+    {
+        /// <inheritdoc />
+        public void OnActionExecuting(ActionExecutingContext context)
+        {
+            if (HasUnsupportedContentTypeError(context))
+            {
+                context.Result = new UnsupportedMediaTypeResult();
```
dougbu: `UnsupportedMediaTypeResult` isn't in this PR.

test/Microsoft.AspNet.Mvc.FunctionalTests/ConsumesAttributeTests.cs
```diff
@@ -49,11 +49,11 @@ public ConsumesAttributeTests(MvcTestFixture<BasicWebSite.Startup> fixture)
...
             // Act
             var response = await Client.SendAsync(request);
-            var product = JsonConvert.DeserializeObject<Product>(await response.Content.ReadAsStringAsync());
+            var body = await response.Content.ReadAsStringAsync();
 
             // Assert
-            Assert.Equal(HttpStatusCode.NoContent, response.StatusCode);
-            Assert.Null(product);
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            Assert.Equal("ConsumesAttribute_PassThrough_Product_Json", body);
```
dougbu: How are these test changes related to the new filter?

test/Microsoft.AspNet.Mvc.FunctionalTests/InputFormatterTests.cs
```diff
@@ -60,47 +60,6 @@ public InputFormatterTests(MvcTestFixture<FormatterWebSite.Startup> fixture)
...
-            request.Content = content;
-            var response = await Client.SendAsync(request);
-
-            var responseBody = await response.Content.ReadAsStringAsync();
-            var result = JsonConvert.DeserializeObject<FormatterWebSite.ErrorInfo>(responseBody);
-
-            // Assert
-            Assert.Equal(1, result.Errors.Count);
-            Assert.Equal("Unsupported content type '" + requestContentType + "'.",
-                         result.Errors[0]);
```
dougbu: Why remove this test? All that's changing here should be `ErrorMessage != null` -> `Exception != null`.

dougbu: Never mind, thought this was an integration test.

test/WebSites/BasicWebSite/Controllers/ActionConstraints/ConsumesAttribute_PassThroughController.cs
```diff
@@ -10,9 +10,9 @@ namespace BasicWebSite.Controllers.ActionConstraints
     public class ConsumesAttribute_PassThroughController : Controller
     {
         [Consumes("application/json")]
-        public Product CreateProduct([FromBody] Product_Json jsonInput)
+        public IActionResult CreateProduct(Product_Json jsonInput)
         {
-            return jsonInput;
+            return Content("ConsumesAttribute_PassThrough_Product_Json");
```
dougbu: This is no longer testing `[Consumes]` because that attribute only interacts w/ `[FromBody]` parameters.

test/WebSites/BasicWebSite/Controllers/ActionConstraints/ConsumesAttribute_WithFallbackActionController.cs
```diff
@@ -10,20 +10,20 @@ namespace BasicWebSite.Controllers.ActionConstraints
     public class ConsumesAttribute_WithFallbackActionController : Controller
```
dougbu: These tests were previously about `[Consumes]` (which interacts w/ `[FromBody]`) and how formatted content round trips. What are they about now?

test/WebSites/FormatterWebSite/Controllers/InputFormatterController.cs
```diff
@@ -9,35 +9,6 @@ namespace FormatterWebSite.Controllers
...
-            return dummy;
-        }
-
-        [HttpPost]
-        [ValidateBodyParameter]
-        public object ActionFilterHandlesError([FromBody] DummyClass dummy)
-        {
-            return dummy;
-        }
-
```
dougbu: Huh? These actions look like they're part of valid and unrelated tests.

src/Microsoft.AspNet.Mvc.Core/ModelBinding/UnsupportedContentTypeFilter.cs
dougbu: Found it.

dougbu:
:watch:

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
javiercn: That enumerator is a struct, we just don't want to expose Dictionary<T,K>.Enumerator directly as it's leaking an implementation detail

src/Microsoft.AspNet.Mvc/MvcServiceCollectionExtensions.cs
javiercn: Might be due to the rebase? I don't recall changing that, still the change looks right and is harmless

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/UnsupportedContentTypeException.cs
javiercn: I think it's fine here, TooManyModelErrorsException is on the same place and I can't argue why one is there and the other is not. I see this exception more as a contract than as an implementation detail. @rynowak can chip on this.

test/Microsoft.AspNet.Mvc.FunctionalTests/ConsumesAttributeTests.cs
javiercn: The tests used to receive the product in the body and just return null, which produced 204. With this change, the tests returned 415, so we removed the [FromBody] from the controllers (so that action selection happens normally) and changed the response so that we can verify the right action was hit.

test/WebSites/BasicWebSite/Controllers/ActionConstraints/ConsumesAttribute_PassThroughController.cs
javiercn: It also interacts during action selection

test/WebSites/BasicWebSite/Controllers/ActionConstraints/ConsumesAttribute_WithFallbackActionController.cs
javiercn: The action selection component of it, you can have [Consumes] globally to express that your API only supports json for example, or you can use consumes to disambiguate between two actions (action selection again).

test/WebSites/FormatterWebSite/Controllers/InputFormatterController.cs
javiercn: I don't think so, the behavior for these two actions is covered by other tests already, these tests only check that there's a value in the model state when content-type is invalid and that you can handle it either on the action or using a filter. Given that we have a filter that does this by default now and that we have test to cover this behavior, I don't see any reason to keep these around

javiercn:
@rynowak updated

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/UnsupportedContentTypeException.cs
rynowak: No. Move these out. I just missed it in my review. `TooManyModelErrorsException` is only here because the code that throws it is in this assembly. It's a bit of a wart.

:trollface: Please ask for permission before putting any new code in `.Abstractions` :laughing:

javiercn: I will move it out then

javiercn:
@rynowak updated, moved the exception to core

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
dougbu: >  don't want to expose Dictionary.Enumerator directly as it's leaking an implementation detail

In what way is that enumerator exposed? The wrapper enumerator you're creating has identical semantics and the layer of indirection hides nothing. Regardless, this isn't a worthwhile change and it has no relation to the rest of this PR. Revert changes to this class.

src/Microsoft.AspNet.Mvc/MvcServiceCollectionExtensions.cs
dougbu: The change makes it less clear what the comment refers to. If it doesn't go away when you next rebase, revert changes to this file.

test/WebSites/FormatterWebSite/Controllers/InputFormatterController.cs
dougbu: :ok:

test/Microsoft.AspNet.Mvc.FunctionalTests/ConsumesAttributeTests.cs
dougbu: :ok:

dougbu:
:watch:

Please rebase to clear artifacts from previous rebase(s) and get the paths updated.

javiercn:
@dougbu updated

test/Microsoft.AspNetCore.Mvc.FunctionalTests/WebApiCompatShimActionSelectionTest.cs
```diff
@@ -329,6 +327,25 @@ public WebApiCompatShimActionSelectionTest(MvcTestFixture<WebApiCompatShimWebSit
         }
 
         [Theory]
+        [InlineData("api/Admin/Test", "PostUser")]
+        [InlineData("api/Admin/Test?name=mario&age=10", "PostUserByNameAndAge")]
+        public async Task LegacyActionSelection_OverloadedAction_WithUnnamedAction(string requestUrl, string expectedActionName)
```
dougbu: Are these `LegacyActionSelection_...` tests something you added? A rebase artifact? ...

dougbu: Since the `WebApiCompatShimWebSite` hasn't changed, I suspect these changes don't belong in this PR.

test/WebSites/BasicWebSite/Controllers/ActionConstraints/ConsumesAttribute_PassThroughController.cs
dougbu: :ok:

test/WebSites/BasicWebSite/Controllers/ActionConstraints/ConsumesAttribute_WithFallbackActionController.cs
dougbu: :ok:

dougbu:
:shipit: but revert anything that shouldn't be included when you squash the commits and rebase the last time. Also get @rynowak's squirrel.

test/Microsoft.AspNetCore.Mvc.FunctionalTests/WebApiCompatShimActionSelectionTest.cs
javiercn: These actually belong to the PR, as discussed with @rynowak I kept the new behavior of returning 415 when no input formatter can be found even in the web api compat shim. As part of that we try to read the body of the request and in order to make the Web API compat shim test pass, I separated the ones that used POST into their own test method and added a dummy body (it's not used on the action) to keep the test happy.

What we care about is the action selection piece of it, so adding a dummy body is just ok to prevent the BodyModelBinder and UnsupportedContentTypeFilter from triggering a 415

javiercn:
@rynowak ping

rynowak:
:shipit:

PR #3965, opened 2016-01-21
Added caching for ActionConstraints

ajaybhargavb:
@rynowak / @Eilon

### DIFF
--- a/src/Microsoft.AspNet.Mvc.Core/Internal/ActionConstraintCache.cs
+++ b/src/Microsoft.AspNet.Mvc.Core/Internal/ActionConstraintCache.cs
@@ -1,0 +1,200 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Mvc.Abstractions;
+using Microsoft.AspNet.Mvc.ActionConstraints;
+using Microsoft.AspNet.Mvc.Infrastructure;
+
+namespace Microsoft.AspNet.Mvc.Internal
+{
+    public class ActionConstraintCache
+    {
+        private readonly IActionDescriptorCollectionProvider _collectionProvider;
+        private readonly IActionConstraintProvider[] _actionConstraintProviders;
+
+        private volatile InnerCache _currentCache;
+
+        public ActionConstraintCache(
+            IActionDescriptorCollectionProvider collectionProvider,
+            IEnumerable<IActionConstraintProvider> actionConstraintProviders)
+        {
+            _collectionProvider = collectionProvider;
+            _actionConstraintProviders = actionConstraintProviders.OrderBy(item => item.Order).ToArray();
+        }
+
+        private InnerCache CurrentCache
+        {
+            get
+            {
+                var current = _currentCache;
+                var actionDescriptors = _collectionProvider.ActionDescriptors;
+
+                if (current == null || current.Version != actionDescriptors.Version)
+                {
+                    current = new InnerCache(actionDescriptors.Version);
+                    _currentCache = current;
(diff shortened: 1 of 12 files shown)


src/Microsoft.AspNet.Mvc.Core/Infrastructure/DefaultActionSelector.cs
```diff
@@ -33,10 +35,12 @@ public class DefaultActionSelector : IActionSelector
         public DefaultActionSelector(
             IActionSelectorDecisionTreeProvider decisionTreeProvider,
             IEnumerable<IActionConstraintProvider> actionConstraintProviders,
```
rynowak: We shouldn't need the providers here anymore

src/Microsoft.AspNet.Mvc.Core/Internal/ActionConstraintCache.cs
```diff
@@ -0,0 +1,213 @@
...
+                    var item = items[i];
+                    if (item.IsReusable)
+                    {
+                        Items[i] = item;
+                    }
+                    else
+                    {
+                        Items[i] = new ActionConstraintItem(item.Metadata);
+                    }
+                }
```
rynowak: It looks like this is an extra defensive copy of `items` - the filter cache code may have had this too. Can you take a look and tell me if you agree?

rynowak:
:watch: for small comments

src/Microsoft.AspNet.Mvc.Core/Internal/ActionConstraintCache.cs
ajaybhargavb: I agree. We do the same thing here https://github.com/aspnet/Mvc/pull/3965/files#diff-450e0a08195e9893206a2926f8b38335R106

ajaybhargavb:
:up: :date:

src/Microsoft.AspNet.Mvc.Core/Internal/ActionConstraintCache.cs
```diff
@@ -0,0 +1,200 @@
...
+using System.Diagnostics;
+using System.Linq;
+using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Mvc.Abstractions;
+using Microsoft.AspNet.Mvc.ActionConstraints;
+using Microsoft.AspNet.Mvc.Infrastructure;
+
+namespace Microsoft.AspNet.Mvc.Internal
+{
+    public class ActionConstraintCache
```
pranavkm: Would we have a flag that would ignore missing doc comments when we set `xmlDocs: true`? If not, consider adding some docs here.

ajaybhargavb: Yes. We have suppress error support.

rynowak:
:shipit:

src/Microsoft.AspNet.Mvc.Core/Internal/ActionConstraintCache.cs
```diff
@@ -0,0 +1,200 @@
...
+        {
+            var cache = CurrentCache;
+
+            CacheEntry entry;
+            if (cache.Entries.TryGetValue(action, out entry))
+            {
+                return GetActionConstraintsFromEntry(entry, httpContext, action);
+            }
+
+            if (action.ActionConstraints == null || action.ActionConstraints.Count == 0)
```
javiercn: ```action.ActionConstraints?..Count == 0```?

src/Microsoft.AspNet.Mvc.Core/Internal/ActionConstraintCache.cs
```diff
@@ -0,0 +1,200 @@
...
+                {
+                    current = new InnerCache(actionDescriptors.Version);
+                    _currentCache = current;
+                }
+
+                return current;
+            }
+        }
+
+        public IReadOnlyList<IActionConstraint> GetActionConstraints(HttpContext httpContext, ActionDescriptor action)
```
javiercn: null check arguments

PR #3975, opened 2016-01-23
[Fixes #3952] ViewComponentDescriptor.Type should be a TypeInfo instance instead of a Type instance

### LINKED ISSUE #3952: ViewComponentDescriptor.Type should be a TypeInfo instance instead of a Type instance
ViewComponentDescriptor.Type is of type System.Type while ControllerActionDescriptor.ControllerTypeInfo is of type TypeInfo. 

We should change ViewComponentDescriptor.Type to ViewComponentDescriptor.TypeInfo and change the type of the property to be TypeInfo.

In general, TypeInfo represents the information of a type, while Type respresents a type at runtime. We should use TypeInfo through the stack and do AsType where we need an instance of Type.

I'm not sure if there is a scenario where AsType might return null and if we should guard against that.

javiercn:


javiercn:
@rynowak

src/Microsoft.AspNet.Mvc.ViewFeatures/ViewComponents/DefaultViewComponentDescriptorProvider.cs
```diff
@@ -74,7 +74,7 @@ private static ViewComponentDescriptor CreateDescriptor(TypeInfo typeInfo)
             {
                 FullName = ViewComponentConventions.GetComponentFullName(typeInfo),
                 ShortName = ViewComponentConventions.GetComponentName(typeInfo),
-                Type = type,
+                TypeInfo = type.GetTypeInfo(),
```
rynowak: We already have the `TypeInfo` in a local here

rynowak:
:shipit:

PR #3987, opened 2016-01-25
Update the documentation for IViewComponentHelper.InvokeAsync

### LINKED ISSUE #3974: Update the documentation for IViewComponentHelper.InvokeAsync to be more precise about what the behavior of `argument` values is.
From: https://github.com/aspnet/Mvc/issues/3970#issuecomment-174086656:

It might be helpful to update the documentation from
> /// Arguments to be passed to the invoked view component method.

We should additional docs that specify that we allow anonymous types or dictionaries where the keys and values map to the argument names and values in the `Invoke` \ `InvokeAsync` method on the ViewComponent.

pranavkm:
Fixes #3974

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/IViewComponentHelper.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/IViewComponentHelper.cs
@@ -16,7 +16,11 @@ namespace Microsoft.AspNetCore.Mvc
         /// Invokes a view component with the specified <paramref name="name"/>.
         /// </summary>
         /// <param name="name">The name of the view component.</param>
-        /// <param name="arguments">Arguments to be passed to the invoked view component method.</param>
+        /// <param name="arguments">
+        /// Arguments to be passed to the invoked view component method.
+        /// An <see cref="object"/> containing arguments to be passed to the invoked view component method.
+        /// Alternatively, an <see cref="IDictionary{string, object}"/> instance containing the invocation arguments.
+        /// </param>
         /// <returns>A <see cref="Task"/> that on completion returns the rendered <see cref="IHtmlContent" />.
         /// </returns>
         Task<IHtmlContent> InvokeAsync(string name, object arguments);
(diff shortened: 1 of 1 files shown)


pranavkm:
cc @dougbu

src/Microsoft.AspNetCore.Mvc.ViewFeatures/IViewComponentHelper.cs
```diff
@@ -16,7 +16,14 @@ public interface IViewComponentHelper
         /// Invokes a view component with the specified <paramref name="name"/>.
         /// </summary>
         /// <param name="name">The name of the view component.</param>
-        /// <param name="arguments">Arguments to be passed to the invoked view component method.</param>
+        /// <param name="arguments">
+        /// Arguments to be passed to the invoked view component method.
```
dougbu: Wrap first paragraph with `<para>...</para>`.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/IViewComponentHelper.cs
```diff
@@ -16,7 +16,14 @@ public interface IViewComponentHelper
...
+        /// <para>
+        /// The argument can either be an <see cref="System.Collections.Generic.IDictionary{string, object}"/>
+        /// instance, or an anonymous object containing names and values of the arguments for the invoked view
+        /// component method.
```
dougbu: Why invent new words? That is, use something more like the wording for `htmlAttributes` parameters in `IHtmlHelper` and `IHtmlGenerator`:
``` c#
/// An <see cref="object"/> containing arguments to be passed to the invoked view component method.
/// Alternatively, an <see cref="IDictionary{string, object}"/> instance containing the invocation arguments.
```

Note the `IDictionary<TK, TV>` case should come second since we expect it to be used less often i.e. only in extra-dynamic cases.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/IViewComponentHelper.cs
dougbu: ... if a separate paragraph remains. I removed it in my suggestion below.

dougbu:
:watch:

src/Microsoft.AspNetCore.Mvc.ViewFeatures/IViewComponentHelper.cs
pranavkm: > /// An <see cref="object"/> containing arguments to be passed to the invoked view component method.

sounds sounds just like 
> Arguments to be passed to the invoked view component method.

which the user complained wasn't very useful.

dougbu: > which the user complained wasn't very useful.

Perhaps w/o the `object` part I could agree. But the framework doesn't care about anonymous `object`s versus any other thing in the system with `public` properties. And the second sentence makes things even more clear.

pranavkm:
Updated

src/Microsoft.AspNetCore.Mvc.ViewFeatures/IViewComponentHelper.cs
dougbu: This sentence now just repeats the "Arguments to be passed to the invoked view component method." words. Remove it.

dougbu:
:shipit: with 2 one-line removals.

Issue #3972, opened 2016-01-22
Add caching for Validators


ajaybhargavb:

ajaybhargavb:
8c4bcf14c75f14f8b6e135d7822f009e67964e11

ajaybhargavb:
47351aac7a124148f84a484e187abbe0f24b4efc

PR #3990, opened 2016-01-25
Added caching for validators

ajaybhargavb:
#3972 

Will push ClientValidator change as a separate commit/PR.
@rynowak

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/Validation/IModelValidatorProvider.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/Validation/IModelValidatorProvider.cs
@@ -13,8 +13,9 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding.Validation
         /// </summary>
         /// <param name="context">The <see cref="ModelValidatorProviderContext"/>.</param>
         /// <remarks>
-        /// Implementations should add <see cref="IModelValidator"/> instances to
-        /// <see cref="ModelValidatorProviderContext.Validators"/>.
+        /// Implementations should add the <see cref="IModelValidator"/> instances to the appropriate
+        /// <see cref="ValidatorItem"/> instance which should be added to
+        /// <see cref="ModelValidatorProviderContext.Results"/>.
         /// </remarks>
         void GetValidators(ModelValidatorProviderContext context);
     }
--- a/src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/Validation/ModelValiatorProviderContext.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/Validation/ModelValiatorProviderContext.cs
@@ -14,9 +14,11 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding.Validation
         /// Creates a new <see cref="ModelValidatorProviderContext"/>.
         /// </summary>
         /// <param name="modelMetadata">The <see cref="ModelBinding.ModelMetadata"/>.</param>
-        public ModelValidatorProviderContext(ModelMetadata modelMetadata)
+        /// <param name="items">The list of <see cref="ValidatorItem"/>s.</param>
+        public ModelValidatorProviderContext(ModelMetadata modelMetadata, IList<ValidatorItem> items)
         {
             ModelMetadata = modelMetadata;
+            Results = items;
         }
 
         /// <summary>
@@ -39,11 +41,11 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding.Validation
         }
 
         /// <summary>
-        /// Gets the list of <see cref="IModelValidator"/> instances. <see cref="IModelValidatorProvider"/> instances
-        /// should add validators to this list when
+        /// Gets the list of <see cref="ValidatorItem"/> instances. <see cref="IModelValidatorProvider"/> instances
+        /// should add the appropriate <see cref="ValidatorItem.Validator"/> properties when
         /// <see cref="IModelValidatorProvider.GetValidators(ModelValidatorProviderContext)"/>
         /// is called.
         /// </summary>
-        public IList<IModelValidator> Validators { get; } = new List<IModelValidator>();
+        public IList<ValidatorItem> Results { get; }
     }
 }
--- a/src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/Validation/ValidatorItem.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/Validation/ValidatorItem.cs
@@ -1,0 +1,45 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Mvc.ModelBinding.Validation
+{
+    /// <summary>
+    /// Used to associate validators with <see cref="ValidatorMetadata"/> instances
+    /// as part of <see cref="ModelValidatorProviderContext"/>. An <see cref="IModelValidator"/> should
+    /// inspect <see cref="ModelValidatorProviderContext.Results"/> and set <see cref="Validator"/> and
+    /// <see cref="IsReusable"/> as appropriate.
+    /// </summary>
+    public class ValidatorItem
+    {
+        /// <summary>
+        /// Creates a new <see cref="ValidatorItem"/>.
+        /// </summary>
+        public ValidatorItem()
+        {
+        }
+
+        /// <summary>
+        /// Creates a new <see cref="ValidatorItem"/>.
+        /// </summary>
+        /// <param name="validatorMetadata">The <see cref="ValidatorMetadata"/>.</param>
+        public ValidatorItem(object validatorMetadata)
+        {
+            ValidatorMetadata = validatorMetadata;
+        }
+
+        /// <summary>
+        /// Gets the metadata associated with the <see cref="Validator"/>.
+        /// </summary>
+        public object ValidatorMetadata { get; }
+
+        /// <summary>
+        /// Gets or sets the <see cref="IModelValidator"/>.
+        /// </summary>
+        public IModelValidator Validator { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether or not <see cref="Validator"/> can be reused across requests.
+        /// </summary>
+        public bool IsReusable { get; set; }
+    }
+}
--- a/src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs
@@ -143,6 +143,7 @@ namespace Microsoft.Extensions.DependencyInjection
                 return new DefaultCompositeMetadataDetailsProvider(options.ModelMetadataDetailsProviders);
             }));
             services.TryAddSingleton<IObjectModelValidator, DefaultObjectValidator>();
+            services.TryAddSingleton<ValidatorCache>();
 
             //
             // Random Infrastructure
--- a/src/Microsoft.AspNetCore.Mvc.Core/Internal/ActionConstraintCache.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Internal/ActionConstraintCache.cs
@@ -153,12 +153,13 @@ namespace Microsoft.AspNetCore.Mvc.Internal
             }
             
             var actionConstraints = new IActionConstraint[count];
-            for (int i = 0, j = 0; i < items.Count; i++)
+            var actionConstraintIndex = 0;
+            for (int i = 0; i < items.Count; i++)
             {
                 var actionConstraint = items[i].Constraint;
                 if (actionConstraint != null)
                 {
-                    actionConstraints[j++] = actionConstraint;
+                    actionConstraints[actionConstraintIndex++] = actionConstraint;
                 }
             }
 
--- a/src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultModelValidatorProvider.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultModelValidatorProvider.cs
@@ -18,12 +18,21 @@ namespace Microsoft.AspNetCore.Mvc.Internal
         public void GetValidators(ModelValidatorProviderContext context)
         {
             //Perf: Avoid allocations here
-            for (var i = 0; i < context.ValidatorMetadata.Count; i++)
+            for (var i = 0; i < context.Results.Count; i++)
             {
-                var validator = context.ValidatorMetadata[i] as IModelValidator;
+                var validatorItem = context.Results[i];
+
+                // Don't overwrite anything that was done by a previous provider.
+                if (validatorItem.Validator != null)
+                {
+                    continue;
+                }
+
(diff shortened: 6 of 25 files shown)


src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Validation/ModelValidatorProviderContext.cs
```diff
@@ -39,11 +40,11 @@ public IReadOnlyList<object> ValidatorMetadata
...
 
         /// <summary>
-        /// Gets the list of <see cref="IModelValidator"/> instances. <see cref="IModelValidatorProvider"/> instances
+        /// Gets the list of <see cref="ValidatorItem"/> instances. <see cref="IModelValidatorProvider"/> instances
         /// should add validators to this list when
         /// <see cref="IModelValidatorProvider.GetValidators(ModelValidatorProviderContext)"/>
         /// is called.
         /// </summary>
-        public IList<IModelValidator> Validators { get; } = new List<IModelValidator>();
+        public IList<ValidatorItem> Results { get; private set; }
```
rynowak: No need for the private setter.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Validation/ModelValidatorProviderContext.cs
```diff
@@ -39,11 +40,11 @@ public IReadOnlyList<object> ValidatorMetadata
         }
 
         /// <summary>
-        /// Gets the list of <see cref="IModelValidator"/> instances. <see cref="IModelValidatorProvider"/> instances
+        /// Gets the list of <see cref="ValidatorItem"/> instances. <see cref="IModelValidatorProvider"/> instances
         /// should add validators to this list when
```
rynowak: I'd change this comment to say `should set the appropriate ValidatorItem.Validator properties when called`.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Validation/ValidatorItem.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
+{
+    public class ValidatorItem
```
rynowak: Docs yo.

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Validation/ValidatorItem.cs
```diff
@@ -0,0 +1,23 @@
...
+        public ValidatorItem()
+        {
+        }
+
+        public ValidatorItem(object validatorMetadata)
+        {
+            ValidatorMetadata = validatorMetadata;
+        }
+
+        public object ValidatorMetadata { get; private set; }
```
rynowak: No need for private setter

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Validation/ValidatorItem.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Mvc.ModelBinding.Validation
+{
+    public class ValidatorItem
+    {
+        public ValidatorItem()
```
rynowak: What's this constructor for?

src/Microsoft.AspNet.Mvc.Core/Internal/DefaultModelValidatorProvider.cs
```diff
@@ -18,12 +18,14 @@ public class DefaultModelValidatorProvider : IModelValidatorProvider
         public void GetValidators(ModelValidatorProviderContext context)
         {
             //Perf: Avoid allocations here
-            for (var i = 0; i < context.ValidatorMetadata.Count; i++)
+            for (var i = 0; i < context.Results.Count; i++)
             {
-                var validator = context.ValidatorMetadata[i] as IModelValidator;
+                var validatorItem = context.Results[i];
```
rynowak: Add a 'skip' here if `validatorItem.Validator != null`

src/Microsoft.AspNet.Mvc.Core/Internal/ValidatorCache.cs
```diff
@@ -0,0 +1,144 @@
...
+
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Diagnostics;
+using Microsoft.AspNet.Mvc.ModelBinding;
+using Microsoft.AspNet.Mvc.ModelBinding.Validation;
+
+namespace Microsoft.AspNet.Mvc.Internal
+{
+    public class ValidatorCache
```
rynowak: Trusting you that this code is a copy paste with renames and whatnot.

ajaybhargavb: Plus I removed the InnerCache. I suggest you look at this too.

rynowak: That seems fine

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Validation/ValidatorItem.cs
ajaybhargavb: This is for mocking in some tests and stuff like these https://github.com/aspnet/Mvc/pull/3990/files#diff-4538fda6cd1ba9234665211044c79245R95

rynowak: Gotcha, seems fine

src/Microsoft.AspNet.Mvc.Core/Internal/DefaultModelValidatorProvider.cs
ajaybhargavb: Added.

ajaybhargavb:
:up: :date:

rynowak:
:shipit:

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Validation/ValidatorItem.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+        }
+
+        /// <summary>
+        /// Creates a new <see cref="ValidatorItem"/>.
+        /// </summary>
+        /// <param name="validatorMetadata">The <see cref="ValidatorMetadata"/>.</param>
+        public ValidatorItem(object validatorMetadata)
+        {
+            ValidatorMetadata = validatorMetadata;
```
javiercn: Is it ok for the validatorMetadata to be null in case there is no metadata?

src/Microsoft.AspNet.Mvc.Core/Internal/ValidatorCache.cs
```diff
@@ -0,0 +1,144 @@
...
+
+            ExecuteProvider(validationProvider, metadata, items);
+
+            return ExtractValidators(items);
+        }
+
+        private void ExecuteProvider(IModelValidatorProvider validatorProvider, ModelMetadata metadata, List<ValidatorItem> items)
+        {
+            var context = new ModelValidatorProviderContext(metadata, items);
+
```
javiercn: nit: no need for this blank line

src/Microsoft.AspNet.Mvc.Core/Internal/ValidatorCache.cs
```diff
@@ -0,0 +1,144 @@
...
+                }
+            }
+
+            if (count == 0)
+            {
+                return EmptyArray;
+            }
+
+            var validators = new IModelValidator[count];
+            for (int i = 0, j = 0; i < items.Count; i++)
```
javiercn: This is deceiving, j is not used as part of the condition of the for nor it's incremented on the step, which is the expectation in for loops, also it's not a good practice to modify the loop variable outside the step part of the for statement.

So just keep things simple, declare j outside of the for and change ```int i``` to ```var i```.

test/Microsoft.AspNet.Mvc.DataAnnotations.Test/Internal/DataAnnotationsModelValidatorProviderTest.cs
```diff
@@ -99,13 +100,24 @@ public void IValidatableObjectGetsAValidator()
             var mockValidatable = new Mock<IValidatableObject>();
             var metadata = _metadataProvider.GetMetadataForType(mockValidatable.Object.GetType());
 
-            var providerContext = new ModelValidatorProviderContext(metadata);
+            var providerContext = new ModelValidatorProviderContext(metadata, GetValidatorItems(metadata));
```
javiercn: Is there an scenario in which a user needs to create a ModelValidatorProviderContext?
GetValidatorItems seems like a product of metadata, which could be computed internally on the constructor. I know we want to keep context classes as simple as possible, but doing it in the constructor reduces the complexity of the context class and helps prevents incorrect usages, we can keep the two parameter constructor as an override of the default logic if it makes sense or helps in other areas like unit testing, bear in mind that I don't have a lot of context on this area. @rynowak thoughts?

I also have seen the lines in GetValidatorItems inside the ValidatorCache (I think code) so seems to me that this is logic we care about/might be common in this area

src/Microsoft.AspNet.Mvc.Abstractions/ModelBinding/Validation/ValidatorItem.cs
ajaybhargavb: Yes, for the cases when we directly add a validator.

src/Microsoft.AspNet.Mvc.Core/Internal/ValidatorCache.cs
ajaybhargavb: I kind of agree with what you are saying but having a variable declaration like `var j = 0` outside a for loop looks odd to me.

javiercn:
:shipit:

test/Microsoft.AspNet.Mvc.DataAnnotations.Test/Internal/DataAnnotationsModelValidatorProviderTest.cs
ajaybhargavb: Spoke offline. We are not doing this because `context.Results` will contain items added by a previous provider and creating a new context with a fresh set of items is not very common.

src/Microsoft.AspNet.Mvc.Core/Internal/ValidatorCache.cs
ajaybhargavb: Spoke offline. Will make this change and rename `j` to something else.

src/Microsoft.AspNet.Mvc.Core/Internal/DefaultObjectValidator.cs
```diff
@@ -13,20 +13,27 @@ namespace Microsoft.AspNet.Mvc.Internal
...
         public DefaultObjectValidator(
-            IModelMetadataProvider modelMetadataProvider)
+            IModelMetadataProvider modelMetadataProvider,
+            ValidatorCache validatorCache)
         {
             if (modelMetadataProvider == null)
             {
                 throw new ArgumentNullException(nameof(modelMetadataProvider));
             }
+            if (validatorCache == null)
```
pranavkm: New line prior to this.

src/Microsoft.AspNet.Mvc.Core/Internal/ValidatorCache.cs
```diff
@@ -0,0 +1,144 @@
...
+using Microsoft.AspNet.Mvc.ModelBinding;
+using Microsoft.AspNet.Mvc.ModelBinding.Validation;
+
+namespace Microsoft.AspNet.Mvc.Internal
+{
+    public class ValidatorCache
+    {
+        private readonly IReadOnlyList<IModelValidator> EmptyArray = new IModelValidator[0];
+
+        private volatile ConcurrentDictionary<ModelMetadata, CacheEntry> _cacheEntries = new ConcurrentDictionary<ModelMetadata, CacheEntry>();
```
pranavkm: `readonly volatile`

test/Microsoft.AspNet.Mvc.Core.Test/Internal/ValidatorCacheTest.cs
```diff
@@ -0,0 +1,110 @@
...
+            // Assert - 1
+            Assert.Collection(
+                validators1,
+                v => Assert.Same(metadata.ValidatorMetadata[0], Assert.IsType<DataAnnotationsModelValidator>(v).Attribute), // Copied by provider
+                v => Assert.Same(metadata.ValidatorMetadata[1], Assert.IsType<DataAnnotationsModelValidator>(v).Attribute)); // Copied by provider
+
+            // Act - 2
+            var validators2 = cache.GetValidators(metadata, validatorProvider);
+
+            Assert.NotSame(validators1, validators2);
```
pranavkm: ` // Assert - 2`

test/Microsoft.AspNet.Mvc.Core.Test/Internal/ValidatorCacheTest.cs
```diff
@@ -0,0 +1,110 @@
...
+            // Assert - 1
+            Assert.Collection(
+                validators1,
+                v => Assert.Same(metadata.ValidatorMetadata[0], Assert.IsType<DataAnnotationsModelValidator>(v).Attribute), // Copied by provider
+                v => Assert.Same(metadata.ValidatorMetadata[1], Assert.IsType<DataAnnotationsModelValidator>(v).Attribute)); // Copied by provider
+
+            // Act - 2
+            var validators2 = cache.GetValidators(metadata, validatorProvider);
+
+            Assert.Same(validators1, validators2);
```
pranavkm: ` // Assert - 2`

test/Microsoft.AspNet.Mvc.Core.Test/Internal/DefaultModelValidatorProviderTest.cs
```diff
@@ -131,17 +131,28 @@ public void GetValidators_FromModelMetadataType_MergedValidators()
...
-            Assert.Single(validators, v => ((DataAnnotationsModelValidator)v).Attribute is RegularExpressionAttribute);
-            Assert.Single(validators, v => ((DataAnnotationsModelValidator)v).Attribute is StringLengthAttribute);
+            Assert.Equal(2, validatorItems.Count);
+            Assert.Single(validatorItems, v => ((DataAnnotationsModelValidator)v.Validator).Attribute is RegularExpressionAttribute);
+            Assert.Single(validatorItems, v => ((DataAnnotationsModelValidator)v.Validator).Attribute is StringLengthAttribute);
+        }
+
+        private static IList<ValidatorItem> GetValidatorItems(ModelMetadata metadata)
+        {
+            var items = new List<ValidatorItem>(metadata.ValidatorMetadata.Count);
```
pranavkm: return `metadata.ValidatorMetadata.Select(m => new ValidatorItem(m)).ToList()`;

src/Microsoft.AspNet.Mvc.Core/Internal/ValidatorCache.cs
rynowak: There's no need for volatile if it's readonly

PR #3995, opened 2016-01-26
Adds support for binding to flags enums

tuespetre:
See #3977

Issue #3977, opened 2016-01-23
Support flagged enums for various scenarios


tuespetre:
There are several times I would have like to have this functionality; sometimes when you have a user-facing system that involves lots of 'flag-setting', flagged enums can be much easier to manage than an explosion of boolean fields.

I have noticed that GetEnumSelectList throws an exception for flagged enums, and that model binding does not seem to properly account for flagged enums, nor do checkboxes and tag helpers behave as desired (at least as desired by myself...)

I would like to propose the following behaviors:
- During model binding, if there is an array of values for a given key and the binding target is a flagged enum, interpret each value of the array as a member of that enum and 'or' them all together instead of just taking the first value present in the array
- When determining the 'current values' for a flagged enum in the model, only consider those combined values that are explicitly defined for the enum (i.e. 3, 7, -1, etc.)
- During generation of checkbox controls and select list options, in order to determine whether they are selected or checked or not, check to see if the model value contains the flags represented by the value specified for the checkbox/option. 
- During generation of select list items for an enum, don't throw an exception if the enum is flagged. Just proceed as normal

Please let me know any thoughts/concerns.

dougbu:
The general concern is that `[Flags]` `enum`s have wildly-varying semantics that cannot consistently be turned into a `IEnumerable<SelectListItem>`. Just for example, how should mask values (those w/ more than one bit set) in the `enum` be handled -- ignored, used to group the constituent bits, ...?

The system should work fine if you pass your own select list into the helpers. Please let us know the details if this is not the case.

dougbu:
Hit "Comment" and then remembered my "should work fine" is probably not correct. A `<select mutliple="multiple">` element for a `[Flags]` `enum` won't be bound correctly if the user selects more than one value. A set of checkboxes that an action or other user code or's together would work better.

tuespetre:
I'm sorry, it was 1:30AM here when I typed that; it could have been written better. I will get back to you later today and make a better write-up.

tuespetre:
Alright, so there are three areas of behavior surrounding `[Flags]` `enum`s that could change:

1. Binding values to the model
    - Currently, if multiple values come in for a `[Flags]` `enum`, only the first value will be applied. I believe that the default binding behavior should instead be to 'or' the values together.
2. Generating an `IEnumerable<SelectListItem>` for a `[Flags]` `enum`
[...]

Issue #3953, opened 2016-01-19
`SelectListItem` manipulation is quite expensive


dougbu:
`SelectListItem`, `List<SelectListItem>` and related allocations quickly overwhelm any scenario using `Html.ListBox()`, `<select asp-for="...">` or similar helpers with a non-empty `IEnumerable<SelectListItem>` argument or `asp-items` attribute value. For example `SelectListItem`s rise to be the fourth-most allocated type in a scenario involving ten `Html.ListBoxFor()` calls per request despite 50 `Html.LabelFor()` and 40 `Html.IdFor()` calls per request (see #3918).

Very little of the relevant code has changed since MVC 5. It definitely hasn't been optimized. Example perf issues:
- Extensive Linq use (including Linq syntax though a `foreach` would have been clearer) when enumerating items in a `MultiSelectList.
- Extensive Linq use (e.g. `GroupBy()`) in `DefaultHtmlGenerator.GenerateGroupsAndOptions()`.
- Defensive deep copies of `IEnumerable<SelectListItem>` values passed to `DefaultHtmlGenerator.UpdateSelectListItemsWithDefaultValue()`. Could avoid updating user's collection other ways e.g. tracking selection in a bit vector or using `currentValues.Contains(value)` directly in `GenerateOption()`.

dougbu:
@Eilon another one for the backlog?

dougbu:
Overall:
![overall](https://cloud.githubusercontent.com/assets/6431421/12433909/83336896-beb8-11e5-9f90-a44b82226e4a.png)

Directly `SelectListItem`-related:
![selectlistitem](https://cloud.githubusercontent.com/assets/6431421/12433912/859c854a-beb8-11e5-8153-4aff33edf548.png)

dougbu:
Should have mentioned: The directly `SelectListItem`-related allocations are worth more than 12% of the memory used in this scenario.

rynowak:
> Should have mentioned: The directly SelectListItem-related allocations are worth more than 12% of the memory used in this scenario.

To add to this, the benchmark is creating and caching the items up-front, it's already the most optimized usage a user can get with this helper.

tuespetre:
I noticed major improvements somewhere between beta7 and rc1 in a page I have with a very large amount of `<select>` elements. I would be interested in trying this out; can you point me in the right direction as far as running the performance tests and getting their results, or is this something that stays internal?

dougbu:
[...]

PR #3999, opened 2016-01-26
Flags enum support for some select-related classes (see #3977)

tuespetre:
Intended to build upon #3995, although that commit is not included in this PR.
- Allow use of `GetEnumSelectList` helpers with Flags enums
  - The test data for the related tests was very strange, as all of the test methods were accepting three parameters but only using one. I discerned which tests were using which parameters and split the test data out to make things a little more straightforward.
- Make `DefaultHtmlGenerator` detect multiple values for Flags enums
  - I did reorder a small handful of test data parameters related to this just because they were coded to expect a trimmed value first before an untrimmed value. I am not sure if this is anything significant or not.
  - This is using the old `private static readonly IDictionary<Type,Delegate>` trick to cache the delegates used to squeeze the current values out of an enum value. I'm not sure how friendly everyone is towards that pattern.
- Remove `EnumNamesAndValues` from metadata classes
  - Though this was technically a `public` item, the only place it was used was during `GetCurrentValues` in the `DefaultHtmlGenerator`. It was kind of an awkward property and the one place it was used did not seem performance-friendly. At first I thought maybe it held the `[Display]` text for the enum values but that did not seem to be the case.
- Make `<select>` tag helper behave as `<select multiple>` with Flags enums
  - Perhaps the tag helper should only behave this way with a Flags enum if the 'multiple' attribute is actually present? Or maybe the tag helper should strictly require the 'multiple' attribute to function as a multi-select and throw an exception if the model type is invalid?


Eilon:
@tuespetre at this time we are not planning to take this feature because we are working on stabilizing the RC2 release. We will revisit this in the next release.

tuespetre:
Thanks @Eilon; question for you and probably @rynowak / @dougbu: what's up with the `EnumNamesAndValues` property on the metadata? As it stands it seems to be part of the 'public API' but as I mentioned in this PR it doesn't seem to be used anywhere besides one place, and in that one place it seems like it would have an unnecessary performance impact (uses LINQ when it doesn't need to, etc.)

In my mind, that is closely related to #3953; I was just curious about this and wondering if you had some other reason for it to be part of the 'public API'.

rynowak:
> As it stands it seems to be part of the 'public API' but as I mentioned in this PR it doesn't seem to be used anywhere besides one place

I seem to recall this was added to support that one place it's used :+1: The `ModelMetadata` is a cache of stuff that could be derived from reflection + attributes. The list of display-names for enum values is really independent of how we render HTML for an enum, hence the layering.


> it seems like it would have an unnecessary performance impact

What code are you talking about specifically? Our approach with perf is to measure first. Lot's of code has an 'unnecessary' cost - but generally we're looking for the right tradeoffs.

Is this code much more readable as is? [This method](https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/Internal/FilterCache.cs#L136) used to be about 5 lines before we ran into the need to optimize it :+1: 

Does this code have an impact in the big picture (in a real application)?

If it's something trivial like turning a linq into the equivalent looping construct and it runs on every request, and it's not a big decrease in readability, then it seems like a good idea.

dougbu:
> ... it runs on every request ...

The Linq in question runs once per type, not on every request. See [this block](https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.DataAnnotations/Internal/DataAnnotationsMetadataProvider.cs#L107-L141) in the `DataAnnotationsMetadataProvider`.

tuespetre:
@rynowak:
> What code are you talking about specifically?
@dougbu:
> The Linq in question runs once per type, not on every request.

This is the code I am talking about; it will run every time the `DefaultHtmlGenerator` gets the current values for a given model expression:

https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs#L1023

> If it's something trivial like turning a linq into the equivalent looping construct and it runs on every request, and it's not a big decrease in readability, then it seems like a good idea.

That's about what it would be; if `EnumNamesAndValues` was also a `IReadOnlyDictionary<string, Enum>` instead, it would eliminate the need to compare string values.

> I seem to recall this was added to support that one place it's used :+1: The ModelMetadata is a cache of stuff that could be derived from reflection + attributes. The list of display-names for enum values is really independent of how we render HTML for an enum, hence the layering.

Thanks for the explanation. There's also this guy here, maybe it could be a cached delegate stored in the `ModelMetadata`:

https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs#L1007

dougbu:
@tuespetre are you seeing any of this Linq in performance measurements? Separately we'd have to be very careful about caching at either site you linked to: Values change per request.

tuespetre:
> @tuespetre are you seeing any of this Linq in performance measurements? Separately we'd have to be very careful about caching at either site you linked to: Values change per request.

To be honest I haven't; I have had issues with trying to get the performance benchmarks going like we discussed on #3953. I may try that again sometime today or tonight depending on how things go. As for the caching bit, it would just be caching a delegate instead of generating one with `MakeGenericMethod()` each time.

I think I will try to get those benchmarks going again later and pick up this discussion back over on #3953. Thanks!

Eilon:
Re-opening because we are still considering this.

mkArtakMSFT:
Hi @tuespetre. Sorry for our late response on this.
As a lot of time has passed since this PR has been opened, there are a lot of conflicts already, so we're going to close this PR.

Issue #3622, opened 2015-11-23
Reduce allocations of ModelBindingContext


rynowak:
We allocate too many `ModelBindingContext` objects:
![image](https://cloud.githubusercontent.com/assets/1430011/11345865/959790be-91ce-11e5-9271-9854190c252c.png)

This is the allocations for 3000 requests to bind 3 fields. It's about 1% of allocated bytes.

We should be able to (at a minimum) use one context per-level of the model binding tree, or use one shared context and mutate/restore state as necessary. Currently we're allocating O(N) contexts, where N is the number of nodes (8 per request for this example).

PR #4007, opened 2016-01-28
Reduce allocations for binding context

### LINKED ISSUE #3621: Change IModelBinder to mutate context instead of returning Task<ModelBindingResult>
The pattern of returning `Task<T>` is allocating too much for our usage in model binding. 

image

These are the allocations for 3000 requests and model binding 3 fields. In context, this is about 1% of allocations for the app.

The root cause is that we can't use a cached task, because `Task<ModelBindingResult>` has state. We should put the 'result' on the `ModelBindingContext` and then change `BindAsync` to return `Task` (non-generic).

lodejard:
Reduces allocation for binding context by having result be present in the context rather than returned as a values. Also reduces allocations by having the binding context push/pop state onto value-type stack rather than create child binding context per object property.

Fixes #3621 #3622


src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
```diff
@@ -289,12 +282,57 @@ public IValueProvider ValueProvider
...
+            public string BinderModelName;
+            public BindingSource BindingSource;
+            public Type BinderType;
+            public bool FallbackToEmptyPrefix;
+            public bool IsTopLevelObject;
+
+            public ModelBindingResult? Result;
+        };
+
+        State _state;
```
davidfowl: stylecop 13040: Move these to the top of the file.

lodejard: lol

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CollectionModelBinder.cs
```diff
@@ -259,7 +265,7 @@ private Task<CollectionResult> BindComplexCollection(ModelBindingContext binding
         }
 
         // Internal for testing.
-        internal class CollectionResult
+        public class CollectionResult
```
davidfowl: ?

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
```diff
@@ -57,18 +60,18 @@ public class DictionaryModelBinder<TKey, TValue> : CollectionModelBinder<KeyValu
...
             // Update the existing successful but empty ModelBindingResult.
             var metadataProvider = bindingContext.OperationBindingContext.MetadataProvider;
             var valueMetadata = metadataProvider.GetMetadataForType(typeof(TValue));
-            var valueBindingContext = ModelBindingContext.CreateChildBindingContext(
-                bindingContext,
-                valueMetadata,
-                fieldName: bindingContext.FieldName,
-                modelName: bindingContext.ModelName,
-                model: null);
+            //var valueBindingContext = ModelBindingContext.CreateChildBindingContext(
```
davidfowl: Dead code?

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/FormCollectionModelBinder.cs
```diff
@@ -9,6 +9,7 @@
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc.ModelBinding.Validation;
 using Microsoft.Extensions.Primitives;
+using System.Diagnostics;
```
davidfowl: Sort usings

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/FormCollectionModelBinder.cs
```diff
@@ -17,13 +18,16 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
     /// </summary>
     public class FormCollectionModelBinder : IModelBinder
     {
+        private static Task _complete = Task.FromResult(0);
```
davidfowl: Why not use the Task helper used elsewhere?

dougbu:
@lodejard do you have before / after numbers for a model binding scenario e.g. @rynowak's `BasicViews`? I suspect this looks good but would be great to confirm expectations.

davidfowl:
This change looks really good. This directly affects authoring a model binder right?

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/IModelBindingContext.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Mvc.ModelBinding.Validation;
+
+namespace Microsoft.AspNetCore.Mvc.ModelBinding
+{
+    public interface IModelBindingContext
```
dougbu: Why add this interface? It seems to cause a lot of additional churn, there's one implementation, and no extensibility point here.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
```diff
@@ -260,28 +257,27 @@ public ModelStateDictionary ModelState
         /// Passed into the model binding system. Should not be <c>true</c> when <see cref="IsTopLevelObject"/> is
         /// <c>false</c>.
         /// </remarks>
-        public bool FallbackToEmptyPrefix { get; set; }
+        public bool FallbackToEmptyPrefix { get { return _state.FallbackToEmptyPrefix; } set { _state.FallbackToEmptyPrefix = value; } }
```
dougbu: Long enough to start wrapping this. Same for a few other properties in this class.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
```diff
@@ -289,12 +285,54 @@ public IValueProvider ValueProvider
...
+
+        public ModelBindingResult? Result
+        {
+            get
+            {
+                return _state.Result;
+            }
+            set
+            {
+#if DEBUG
```
dougbu: Why is this check debug-only?

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContextDisposable.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Mvc.ModelBinding
+{
+    public struct ModelBindingContextDisposable : IDisposable
```
dougbu: `/// docs` and some blank lines between members

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ArrayModelBinder.cs
```diff
@@ -16,16 +16,17 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
...
             if (bindingContext == null)
             {
                 throw new ArgumentNullException(nameof(bindingContext));
             }
+            Debug.Assert(bindingContext.Result == null);
 
             if (bindingContext.ModelMetadata.IsReadOnly)
             {
-                return ModelBindingResult.NoResultAsync;
+                return Internal.TaskCache.CompletedTask;
```
dougbu: Add a `using` for the `Internal` namespace.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
```diff
@@ -289,12 +285,54 @@ public IValueProvider ValueProvider
...
+                    throw new ArgumentException(nameof(ModelBindingResult));
+                }
+#endif
+
+                _state.Result = value;
+            }
+        }
+
+
+        private struct State
```
pranavkm: Would it be better to have a class here instead of struct with these many fields? (Is there a cost to the copy semantics at this point?)

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/BinderTypeBasedModelBinder.cs
```diff
@@ -31,13 +38,14 @@ public Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindingContex
             {
                 // Return NoResult so that we are able to continue with the default set of model binders,
```
dougbu: Comment needs some love since `NoResult` is no more.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/BinderTypeBasedModelBinder.cs
```diff
@@ -31,13 +38,14 @@ public Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindingContex
...
                 // if there is no specific model binder provided.
-                return ModelBindingResult.NoResultAsync;
+                bindingContext.Result = null;
```
dougbu: No need for this line; just asserted that `Result` is already `null`.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/BodyModelBinder.cs
```diff
@@ -58,13 +65,8 @@ public Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindingContex
         /// <returns>
         /// A <see cref="Task{ModelBindingResult}"/> which when completed returns a <see cref="ModelBindingResult"/>.
         /// </returns>
-        private async Task<ModelBindingResult> BindModelCoreAsync(ModelBindingContext bindingContext)
+        private async Task BindModelCoreAsync(IModelBindingContext bindingContext)
         {
-            if (bindingContext == null)
-            {
-                throw new ArgumentNullException(nameof(bindingContext));
-            }
```
dougbu: This shouldn't be removed.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ArrayModelBinder.cs
dougbu: Same throughout this PR.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CollectionModelBinder.cs
```diff
@@ -138,8 +141,8 @@ protected object CreateInstance(Type targetType)
         // Used when the ValueProvider contains the collection to be bound as a single element, e.g. the raw value
         // is [ "1", "2" ] and needs to be converted to an int[].
         // Internal for testing.
-        internal async Task<CollectionResult> BindSimpleCollection(
-            ModelBindingContext bindingContext,
+        public async Task<CollectionResult> BindSimpleCollection(
```
dougbu: Was previously `internal` only for testing. Why `public` now?

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CollectionModelBinder.cs
```diff
@@ -164,12 +161,20 @@ protected object CreateInstance(Type targetType)
...
                 object boundValue = null;
-                var result =
-                    await bindingContext.OperationBindingContext.ModelBinder.BindModelAsync(innerBindingContext);
-                if (result != null && result.IsModelSet)
+
+                using (bindingContext.PushContext(
+                    elementMetadata,
+                    fieldName: bindingContext.FieldName,
+                    modelName: bindingContext.ModelName,
+                    model: null))
```
dougbu: Why isn't the loop over `values` inside this `using`? Should only need to reset `Result` to `null` inside the loop.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerArgumentBinder.cs
```diff
@@ -118,15 +118,16 @@ public class ControllerArgumentBinder : IControllerActionArgumentBinder
                 parameter.BindingInfo,
                 parameter.Name);
 
-            var modelBindingResult = await operationContext.ModelBinder.BindModelAsync(modelBindingContext);
-            if (modelBindingResult.IsModelSet)
+            await operationContext.ModelBinder.BindModelAsync(modelBindingContext);
+            var modelBindingResult = modelBindingContext.Result;
+            if (modelBindingResult != null && modelBindingResult.Value.IsModelSet)
```
dougbu: Think `HasValue` is much more consistent w/ the rest of this repo than `!= null` (for `Nullable<T>`s).

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
```diff
@@ -85,18 +82,12 @@ public ModelBindingContext()
             };
         }
 
-        public static ModelBindingContext CreateChildBindingContext(
-            ModelBindingContext parent,
+        public ModelBindingContextDisposable PushContext(
             ModelMetadata modelMetadata,
             string fieldName,
             string modelName,
             object model)
```
dougbu: `PushContext()` is the moral equivalent of `CreateChildBindingContext()` but I don't see any calls needing anything but `model: null`. Suggest removing this parameter.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CompositeModelBinder.cs
```diff
@@ -36,24 +37,27 @@ public CompositeModelBinder(IList<IModelBinder> modelBinders)
         /// <inheritdoc />
         public IList<IModelBinder> ModelBinders { get; }
 
-        public virtual Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindingContext)
+        public async Task BindModelAsync(IModelBindingContext bindingContext)
```
dougbu: Don't recall why this was `virtual` but why remove that?

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CompositeModelBinder.cs
```diff
@@ -129,43 +136,30 @@ private static ModelBindingContext CreateNewBindingContext(ModelBindingContext o
...
             }
             else
             {
-                newBindingContext.ModelName = string.Empty;
+                bindingContext.ModelName = string.Empty;
             }
 
-            return newBindingContext;
+            bindingContext.ValueProvider = valueProvider;
+            bindingContext.FallbackToEmptyPrefix = false;
```
dougbu: Suggest resetting `FallbackToEmptyPrefix` for current context as part of `PushContext()`.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
```diff
@@ -79,21 +76,26 @@ public class DictionaryModelBinder<TKey, TValue> : CollectionModelBinder<KeyValu
                 // that culture when rendering a form.
                 var convertedKey = ModelBindingHelper.ConvertTo<TKey>(kvp.Key, culture: null);
 
-                valueBindingContext.ModelName = kvp.Value;
+                using (bindingContext.PushContext(
+                    modelMetadata: valueMetadata,
+                    fieldName: bindingContext.FieldName,
+                    modelName: kvp.Value,
+                    model: null))
```
dougbu: As in the `CollectionModelBinder`, can do this outside the loop rather than per iteration.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/IModelBindingContext.cs
lodejard: An additional step would be to move the `ModelBindingContext` implementation out of abstractions and into core. The class has a lot of implementation that puts some gritty code into an abstractions assembly...

lodejard:
@dougbu no numbers yet, but absolutely yes numbers will be needed before merging this is justified

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/GenericModelBinder.cs
```diff
@@ -34,27 +40,26 @@ public Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindingContex
...
-                ModelBindingResult.Failed(bindingContext.ModelName);
-
-            // Were able to resolve a binder type.
-            // Always tell the model binding system to skip other model binders.
-            return modelBindingResult;
+            await binder.BindModelAsync(bindingContext);
+            if (bindingContext.Result == null)
+            {
+                // Were able to resolve a binder type.
+                // Always tell the model binding system to skip other model binders.
```
dougbu: Not your change but something has been lost here. The code at one point matched the comment -- always returning something _other_ than `NoResult`. Likely the comment needs to change.

lodejard:
@davidfowl it does affect it, the imodelbinder bindmodelasync method changes. however the properties/logistics/context/etc aren't changed by the PR, just how the result is returned and how a child context created

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/KeyValuePairModelBinder.cs
```diff
@@ -23,7 +31,8 @@ public sealed class KeyValuePairModelBinder<TKey, TValue> : IModelBinder
                     ModelBindingHelper.CastOrDefault<TKey>(keyResult.Model),
                     ModelBindingHelper.CastOrDefault<TValue>(valueResult.Model));
 
-                return ModelBindingResult.Success(bindingContext.ModelName, model);
+                bindingContext.Result = ModelBindingResult.Success(bindingContext.ModelName, model);
+                return;
             }
             else if (!keyResult.IsModelSet && valueResult.IsModelSet)
```
dougbu: Could you please remove the `else` bits after `return`s?

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/KeyValuePairModelBinder.cs
```diff
@@ -52,37 +63,40 @@ public sealed class KeyValuePairModelBinder<TKey, TValue> : IModelBinder
...
-                    return ModelBindingResult.Success(bindingContext.ModelName, model);
+                    bindingContext.Result = ModelBindingResult.Success(bindingContext.ModelName, model);
                 }
 
-                return ModelBindingResult.NoResult;
+                return;
             }
         }
 
         internal async Task<ModelBindingResult> TryBindStrongModel<TModel>(
```
dougbu: Couldn't this return just `Task` now?

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ArrayModelBinder.cs
lodejard: It seemed to make more tactical to have a qualified reference to the single static than to expose all of the types in the internal namespace to a file that doesn't use anything else in there, no?

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ModelBinderExtensions.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Mvc.ModelBinding
+{
+    public static class ModelBinderExtensions
+    {
+        public static async Task<ModelBindingResult> BindModelResultAsync(
```
dougbu: Looks like this has the potential to undo some gains main in this PR. Could we instead push the new pattern all the way down?

dougbu: Never mind. Thought I was seeing this used in product code.

test/Microsoft.AspNet.Mvc.Core.Test/out.txt
```diff
@@ -0,0 +1,320 @@
+xUnit.net DNX Runner (32-bit DNX 4.5.1)
```
dougbu: ?? huh

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
lodejard: Well, that is the entire point of the PR - currently each object (the entire model graph including dictionary and list entries) being model bound allocates an additional model binding context. So the assignment of all of the fields from one class to the next is already happening. The reason it's a struct in a Stack is that each of those child contexts can now be fields in a contiguous array w/out allocations.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ArrayModelBinder.cs
dougbu: No. We do namespace qualification only when there's conflicts. Even you'll see a named `using` almost as often.

src/Microsoft.AspNetCore.Mvc.WebApiCompatShim/HttpRequestMessage/HttpRequestMessageModelBinder.cs
```diff
@@ -14,16 +14,16 @@ namespace Microsoft.AspNetCore.Mvc.WebApiCompatShim
...
             if (bindingContext.ModelType == typeof(HttpRequestMessage))
             {
                 var model = bindingContext.OperationBindingContext.HttpContext.GetHttpRequestMessage();
                 bindingContext.ValidationState.Add(model, new ValidationStateEntry() { SuppressValidation = true });
-                return ModelBindingResult.SuccessAsync(bindingContext.ModelName, model);
+                bindingContext.Result = ModelBindingResult.Success(bindingContext.ModelName, model);
             }
 
-            return ModelBindingResult.NoResultAsync;
+            return Internal.TaskCache.CompletedTask;
```
dougbu: @davidfowl this is a good example of the changes someone who wrote an RC1 model binder would need to make.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/IModelBindingContext.cs
dougbu: Could have been done in a separate PR and made this easier to grok.

If we need to do the whole thing as one, needs lots of `/// doc` love and blank lines.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBindingHelper.cs
```diff
@@ -553,7 +554,7 @@ public static class ModelBindingHelper
         }
 
         // Internal for tests
-        internal static string GetPropertyName(Expression expression)
+        public static string GetPropertyName(Expression expression)
```
dougbu: Why now `public`?

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CollectionModelBinder.cs
lodejard: I believe we have standardized on avoiding internal/internalsvisibleto - i'll get clarification on this one.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CollectionModelBinder.cs
lodejard: Pushing/popping guarantees an entirely reset internal state per item, setting Result to null isn't as strong. It was probably outside of the loop before to reduce allocations, which isn't necessary with push/pop structs onto a stack.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContextDisposable.cs
```diff
@@ -0,0 +1,34 @@
...
+using System;
+
+namespace Microsoft.AspNetCore.Mvc.ModelBinding
+{
+    /// <summary>
+    /// Return value of <see cref="IModelBindingContext.PushContext"/>. Should be disposed
+    /// by caller when child binding context state should be popped off of 
+    /// the <see cref="IModelBindingContext"/>.
+    /// </summary>
+    public struct ModelBindingContextDisposable : IDisposable
```
dougbu: We shouldn't need this class at all. Model binding system intentionally catches almost no `Exception`s except those which occur inside an `InputFormatter` or while performing a conversion. So we can unwind stack frames without worrying about the `ModelBindingContext` state. Just have `PushContext()` return `void` and change call sites to use `PopContext()` explicitly.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerArgumentBinder.cs
lodejard: Why is it called Nullable if you can't use =, ==, and != `null`?

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CollectionModelBinder.cs
dougbu: Well, if we need such strong guarantees, something is wrong elsewhere. But you're right it shouldn't be as big a deal inside the loop now.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
lodejard: :ok: - makes me wonder why CreateChildBindingContext had an object model parameter in the first place?

dougbu:
:watch:

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
dougbu: Dunno but was fairly obvious looking at this PR that the parameter isn't needed _now_ :smiley_cat:

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CompositeModelBinder.cs
lodejard: Resetting FallbackToEmptyPrefix? It is assigned to false - I'm not sure what you're suggesting.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
lodejard: Same remark as the other instance.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CompositeModelBinder.cs
dougbu: I'm suggesting the assignment should happen in `PushContext()` and not here. Should _never_ be `true` in an inner context.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/KeyValuePairModelBinder.cs
lodejard: This method pushes a new child context because the property name and type are different, so the result is returned and that context is popped when it returns.

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ModelBinderExtensions.cs
lodejard: Yeah, was only to reduce the delta on the unit tests by an order of magnitude

test/Microsoft.AspNet.Mvc.Core.Test/out.txt
lodejard: D'oh!

src/Microsoft.AspNetCore.Mvc.WebApiCompatShim/HttpRequestMessage/HttpRequestMessageModelBinder.cs
lodejard: :ok:

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBindingHelper.cs
lodejard: Because it was used by unit tests - same as before, need to follow up on confirming policy

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContextDisposable.cs
lodejard: There are several places where methods are much more maintainable with a using statement. Keeping the context at the right state in the face of an uncaught exception is one case, another is when a method has non-trivial if/else or return flow control. The using statement is always implicitly correct.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
lodejard: Found one (might be the only one) in `MutableObjectModelBinder` - looks like it's to *update* a child object on a property in the case that it's already assigned by the default constructor

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/IModelBindingContext.cs
```diff
@@ -0,0 +1,149 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Mvc.ModelBinding.Validation;
+
+namespace Microsoft.AspNetCore.Mvc.ModelBinding
+{
+    /// <summary>
+    /// A context that contains operating information for model binding and validation.
```
dougbu: (Realize you copied many of these comments from `ModelBindingContext` but ...) Bit confusing because while this context includes some information gathered for later validation, it is not used during that process.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/IModelBindingContext.cs
```diff
@@ -0,0 +1,149 @@
...
+        string BinderModelName { get; set; }
+
+        /// <summary>
+        /// Gets the <see cref="Type"/> of an <see cref="IModelBinder"/> associated with the
+        /// <see cref="Model"/>.
+        /// </summary>
+        Type BinderType { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value which represents the <see cref="BindingSource"/> associated with the
```
dougbu: `ModelBinding.BindingSource` since properties bind more tightly than types. (The `ModelBindingContext` class happens to contain a lot of incorrect comment links.)

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/IModelBindingContext.cs
```diff
@@ -0,0 +1,149 @@
...
+        ModelMetadata ModelMetadata { get; set; }
+
+        /// <summary>
+        /// Gets or sets the name of the model. This property is used as a key for looking up values in
+        /// <see cref="IValueProvider"/> during model binding.
+        /// </summary>
+        string ModelName { get; set; }
+
+        /// <summary>
+        /// Gets or sets the <see cref="ModelStateDictionary"/> used to capture <see cref="ModelState"/> values
```
dougbu: `ModelBinding.ModelState`

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/IModelBindingContext.cs
```diff
@@ -0,0 +1,149 @@
...
+        /// <summary>
+        /// Gets the type of the model.
+        /// </summary>
+        /// <remarks>
+        /// The <see cref="ModelMetadata"/> property must be set to access this property.
+        /// </remarks>
+        Type ModelType { get; }
+
+        /// <summary>
+        /// Represents the <see cref="OperationBindingContext"/> associated with this context.
```
dougbu: `ModelBinding.OperationBindingContext`

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/IModelBindingContext.cs
```diff
@@ -0,0 +1,149 @@
...
+        /// <para>
+        /// On completion returns a <see cref="ModelBindingResult"/> which
+        /// represents the result of the model binding process.
+        /// </para>
+        /// <para>
+        /// If model binding was successful, the <see cref="ModelBindingResult"/> should be a value created
+        /// with <see cref="ModelBindingResult.Success"/>. If model binding failed, the
+        /// <see cref="ModelBindingResult"/> should be a value created with <see cref="ModelBindingResult.Failed"/>.
+        /// If there was no data, or this model binder cannot handle the operation, the
+        /// <see cref="ModelBindingResult"/> should be null.
```
dougbu: `null` -> `<c>null</c>`

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/IModelBindingContext.cs
```diff
@@ -0,0 +1,149 @@
...
+        /// </para>
+        /// </summary>
+        ModelBindingResult? Result { get; set; }
+
+        /// <summary>
+        /// Pushes a layer of state onto this context. Model binders will call this as part of recursion when binding properties
+        /// or collection items.
+        /// </summary>
+        /// <param name="modelMetadata"><see cref="ModelMetadata"/> to assign to the <see cref="IModelBindingContext.ModelMetadata"/> property</param>
```
dougbu: `IModelBindingContext.` shouldn't be needed in any of these comments. But first `ModelMetadata` should be either `Value` (consistent with other `<param>`s) or `ModelBinding.ModelMetadata`.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/IModelBindingContext.cs
```diff
@@ -0,0 +1,149 @@
...
+        /// <param name="fieldName">Value to assign to the <see cref="IModelBindingContext.FieldName"/> property</param>
+        /// <param name="modelName"><see cref="IModelBindingContext.ModelName"/> property</param>
+        /// <param name="model"><see cref="IModelBindingContext.Model"/> property</param>
```
dougbu: `Value to assign to the ...` for `modelName` and `model` as well.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
```diff
@@ -289,12 +310,60 @@ public IValueProvider ValueProvider
...
         /// customize validation behavior for a model object.
         /// </summary>
-        public ValidationStateDictionary ValidationState { get; set; }
+        public ValidationStateDictionary ValidationState
+        {
+            get { return _state.ValidationState; }
+            set { _state.ValidationState = value; }
+        }
+
+        public ModelBindingResult? Result
```
dougbu: `/// <inheritdoc />` and, if you like, can slim most of the other documentation down to this.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
dougbu: Pity. Could add an overload w/o the parameter but that would just churn your PR.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContextDisposable.cs
dougbu: Seems like a lot of ceremony for the less-common "non-trivial .. return flow" cases. That is, I can see the benefits where it's needed but am not a fan of the increased nesting where it wraps the trivial cases. Let's chock this up to a style thing.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContextDisposable.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        private readonly IModelBindingContext _context;
+
+        /// <summary>
+        /// Initializes the disposable for a pushed context.
```
dougbu: `disposable` -> `<see cref="ModelBindingContextDisposable"/>`

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContextDisposable.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// </summary>
+        /// <param name="context"></param>
```
dougbu: Fill in the blanks.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/IModelBindingContext.cs
```diff
@@ -0,0 +1,149 @@
...
+
+using System;
+using Microsoft.AspNetCore.Mvc.ModelBinding.Validation;
+
+namespace Microsoft.AspNetCore.Mvc.ModelBinding
+{
+    /// <summary>
+    /// A context that contains operating information for model binding and validation.
+    /// </summary>
+    public interface IModelBindingContext
```
rynowak: Should this just be an abstract class rather than an interface? Most of these properties are just get/set defaults, and the disposable could be a nested struct. 

We don't have any context interfaces that I can think of and there seem to be some good reasons for this to be a class.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/IModelBindingContext.cs
```diff
@@ -0,0 +1,149 @@
...
+        /// <remarks>
+        /// Passed into the model binding system. Should not be <c>true</c> when <see cref="IsTopLevelObject"/> is
+        /// <c>false</c>.
+        /// </remarks>
+        bool FallbackToEmptyPrefix { get; set; }
+
+        /// <summary>
+        /// Gets or sets the name of the current field being bound.
+        /// </summary>
+        string FieldName { get; set; }
```
rynowak: Why are these settable? I expected that you need to call `PushState` to set this (and a few others)

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
```diff
@@ -9,14 +10,10 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
...
+    public class ModelBindingContext : IModelBindingContext
     {
-        private string _fieldName;
-        private ModelMetadata _modelMetadata;
-        private string _modelName;
-        private ModelStateDictionary _modelState;
-        private OperationBindingContext _operationBindingContext;
-        private IValueProvider _valueProvider;
+        State _state;
+        Stack<State> _stack = new Stack<State>();
```
rynowak: coding standards...

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
```diff
@@ -9,14 +10,10 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
     /// <summary>
     /// A context that contains operating information for model binding and validation.
     /// </summary>
-    public class ModelBindingContext
+    public class ModelBindingContext : IModelBindingContext
```
rynowak: I thought the reason for the separation was to put the implementation somewhere else... if we're not going to do that that or can't ... might as well have one class.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingResult.cs
```diff
@@ -13,18 +13,6 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
...
-        /// </summary>
-        public static readonly ModelBindingResult NoResult = new ModelBindingResult();
-
-        /// <summary>
-        /// Returns a completed <see cref="Task{ModelBindingResult}"/> representing the lack of a result. The model
-        /// binding system will continue to execute other model binders.
-        /// </summary>
-        public static readonly Task<ModelBindingResult> NoResultAsync = Task.FromResult(NoResult);
-
-        /// <summary>
```
rynowak: You should also remove the other static constructors for `Task<ModelBindingResult>` - they don't have a purpose anymore

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ByteArrayModelBinder.cs
```diff
@@ -12,12 +14,13 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
...
     {
         /// <inheritdoc />
-        public Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindingContext)
+        public Task BindModelAsync(IModelBindingContext bindingContext)
         {
             if (bindingContext == null)
             {
                 throw new ArgumentNullException(nameof(bindingContext));
             }
+            Debug.Assert(bindingContext.Result == null);
```
rynowak: This assert is about public state on a public class that's a parameter. In practice it will never get tripped, but in principle this is either supported (shouldn't assert) or not supported (should throw). I don't feel like these asserts add much value, and one could make the argument that they aren't correct.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
pranavkm: :+1:

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
lodejard: Which one in particular? private keyword?

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ControllerActionInvokerTest.cs
```diff
@@ -2091,7 +2091,7 @@ public void CreateActionResult_ReturnsObjectContentResult(Type type, object inpu
 
             var binder = new Mock<IModelBinder>();
```
dougbu: `new StubModelBinder()` ?

dougbu: Or did you get rid of `new Mock<IModelBinder>()` primarily where the setup was a bit more complex?

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/ArrayModelBinderTest.cs
```diff
@@ -6,6 +6,7 @@
 using Microsoft.AspNetCore.Http.Internal;
 using Moq;
 using Xunit;
+#pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
```
dougbu: This wasn't needed before and the `async Task` methods all contain `await` operators. Is it still needed and, if so, where does the problem lie?

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/ArrayModelBinderTest.cs
```diff
@@ -25,10 +26,10 @@ public class ArrayModelBinderTest
             var binder = new ArrayModelBinder<int>();
 
             // Act
-            var result = await binder.BindModelAsync(bindingContext);
+            var result = await binder.BindModelResultAsync(bindingContext);
 
             // Assert
-            Assert.NotEqual(ModelBindingResult.NoResult, result);
+            Assert.NotEqual(default(ModelBindingResult), result);
```
dougbu: Might be worth reintroducing `ModelBindingResult.NoResult` for tests only. Of course you've done the work already...

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
davidfowl: yea

pranavkm: `private readonly`

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/CollectionModelBinderTest.cs
```diff
@@ -390,28 +389,26 @@ public void CanCreateInstance_ReturnsExpectedValue(Type modelType, bool expected
 
         private static IModelBinder CreateIntBinder()
         {
-            Mock<IModelBinder> mockIntBinder = new Mock<IModelBinder>();
-            mockIntBinder
-                .Setup(o => o.BindModelAsync(It.IsAny<ModelBindingContext>()))
-                .Returns((ModelBindingContext mbc) =>
+#pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
+            return new StubModelBinder(async mbc =>
+#pragma warning restore CS1998 // Async method lacks 'await' operators and will run synchronously
```
dougbu: Could we convert most of these setups to avoid the "`async` but not `await`" mess? E.g. set `mbc.Result` in this lambda.

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/CompositeModelBinderTest.cs
```diff
@@ -321,56 +296,28 @@ public class CompositeModelBinderTest
...
             // Assert
-            Assert.NotEqual(ModelBindingResult.NoResult, result);
+            Assert.NotEqual(default(ModelBindingResult), result);
             Assert.True(result.IsModelSet);
             Assert.Equal("someName", result.Key);
             Assert.Null(result.Model);
         }
 
         [Fact]
-        public async Task BindModel_UnsuccessfulBind_BinderFails_ReturnsNoResult()
```
dougbu: Good to see `BindModel_UnsuccessfulBind_BinderFails_ReturnsNoResult()` go away. Was an artifact from time when top-level simple types were implicitly `[BindRequired]` (or maybe `[Required]`).

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/ModelBindingContextTest.cs
```diff
@@ -32,28 +33,36 @@ public void CreateChildBindingContext_CopiesProperties()
                 });
 
             var newModelMetadata = metadataProvider.GetMetadataForType(typeof(object));
-            
+
             // Act
```
dougbu: Isn't only the `PushContext()` in `// Act`?

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/ModelBindingContextTest.cs
```diff
@@ -13,7 +14,7 @@ public class ModelBindingContextTest
         public void CreateChildBindingContext_CopiesProperties()
```
dougbu: Rename tests that are now of `PushContext()`.

dougbu: And need a test demonstrating that `PopContext()` restores all values.

dougbu:
:watch:

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/IModelBindingContext.cs
lodejard: Well, as this PR is really more about perf than authoring comments, we should probably file a different bug to take a pass over public surface area and check the text for freshness.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/IModelBindingContext.cs
lodejard: Because they were settable on the context object originally? In other words... Reasons?

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/IModelBindingContext.cs
dougbu: If this interface lives, you're creating something new here. Besides that we tend to correct existing comments when adjacent code is touched in PRs.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/IModelBindingContext.cs
dougbu: Bet many properties were only settable for tests and due to the odd new context handling in `CompositeModelBinder`. Could probably make lots of setters `internal` to clarify the intended API. But interfaces hate `internal`.

rynowak: So if they're not supposed to be settable in the design, can we address it?

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ControllerActionInvokerTest.cs
lodejard: Yes

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
```diff
@@ -232,69 +82,93 @@ public ModelStateDictionary ModelState
...
 
         /// <summary>
-        /// Gets or sets the <see cref="ValidationStateDictionary"/>. Used for tracking validation state to
-        /// customize validation behavior for a model object.
+        /// Return value of <see cref="ModelBindingContext.EnterNestedScope"/>. Should be disposed
+        /// by caller when child binding context state should be popped off of 
+        /// the <see cref="ModelBindingContext"/>.
         /// </summary>
-        public ValidationStateDictionary ValidationState { get; set; }
+        public struct NestedScope : IDisposable
```
dougbu: @rynowak why did you suggest nesting this class? It has to be `public` to avoid exposure as just another `IDisposable`.

Or does the compiler check the runtime type when it gens for a `using`, allowing this to be `private`?

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContextDisposable.cs
```diff
@@ -0,0 +1,8 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Mvc.ModelBinding
+{
```
dougbu: Either remove the file or restore separate location for `NestedScope`.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingResult.cs
lodejard: :ok:

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/IModelBinder.cs
```diff
@@ -13,7 +13,7 @@ public interface IModelBinder
         /// <summary>
         /// Attempts to bind a model.
         /// </summary>
-        /// <param name="bindingContext">The <see cref="ModelBindingContext"/>.</param>
+        /// <param name="bindingContext">The <see cref="DefaultModelBindingContext"/>.</param>
```
dougbu: S/b `ModelBindingContext`. See a fair few other incorrect mentions of `DefaultModelBindingContext` in doc comments.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
```diff
@@ -9,160 +9,46 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
...
-        private ModelMetadata _modelMetadata;
-        private string _modelName;
-        private ModelStateDictionary _modelState;
-        private OperationBindingContext _operationBindingContext;
-        private IValueProvider _valueProvider;
-
         /// <summary>
-        /// Initializes a new instance of the <see cref="ModelBindingContext"/> class.
+        /// Gets or sets a model name which is explicitly set using an <see cref="IModelNameProvider"/>.
+        /// <see cref="Model"/>.
```
dougbu: Remove this line.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
```diff
@@ -171,60 +57,24 @@ public string FieldName
...
-                    throw new ArgumentNullException(nameof(value));
-                }
-
-                _modelName = value;
-            }
-        }
+        public abstract string ModelName { get; set; }
 
         /// <summary>
         /// Gets or sets the <see cref="ModelStateDictionary"/> used to capture <see cref="ModelState"/> values
```
dougbu: `ModelBinding.ModelState`

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
```diff
@@ -232,69 +82,93 @@ public ModelStateDictionary ModelState
         /// <remarks>
         /// The <see cref="ModelMetadata"/> property must be set to access this property.
         /// </remarks>
-        public Type ModelType => ModelMetadata?.ModelType;
+        public abstract Type ModelType { get; }
 
         /// <summary>
-        /// Gets or sets a model name which is explicitly set using an <see cref="IModelNameProvider"/>.
-        /// <see cref="Model"/>.
+        /// Represents the <see cref="ModelBinding.OperationBindingContext"/> associated with this context.
```
dougbu: `Represents` -> `Gets or sets`

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CompositeModelBinder.cs
dougbu: thx

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/DefaultModelBindingContext.cs
```diff
@@ -0,0 +1,324 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Mvc.ModelBinding.Validation;
+
+namespace Microsoft.AspNetCore.Mvc.ModelBinding
```
dougbu: Suggest `Microsoft.AspNetCore.Mvc.Internal`.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/BinderTypeBasedModelBinder.cs
```diff
@@ -3,8 +3,10 @@
 
 using System;
 using System.Collections.Concurrent;
+using System.Diagnostics;
```
dougbu: No longer need most of the `System.Diagnostics` additions.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CompositeModelBinder.cs
lodejard: Tried it - but changing where it was assigned broke tests - reverted.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/BodyModelBinder.cs
```diff
@@ -54,11 +60,11 @@ public Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindingContex
         /// <summary>
         /// Attempts to bind the model using formatters.
         /// </summary>
-        /// <param name="bindingContext">The <see cref="ModelBindingContext"/>.</param>
+        /// <param name="bindingContext">The <see cref="DefaultModelBindingContext"/>.</param>
```
rynowak: Not sure if the comment was made elsewhere but this is just the standard boilerplate so it would say `<see cref="ModelBindingContext"/>.`

However, this is a private method, so the docs can be deleted entirely, it's not really saying anything.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CompositeModelBinder.cs
```diff
@@ -14,7 +15,7 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
     /// instances.
     /// </summary>
     /// <remarks>
-    /// If no binder is available and the <see cref="ModelBindingContext"/> allows it,
+    /// If no binder is available and the <see cref="DefaultModelBindingContext"/> allows it,
```
rynowak: Should be `ModelBindingContext`

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CompositeModelBinder.cs
```diff
@@ -36,24 +37,26 @@ public CompositeModelBinder(IList<IModelBinder> modelBinders)
         /// <inheritdoc />
         public IList<IModelBinder> ModelBinders { get; }
 
-        public virtual Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindingContext)
+        public virtual async Task BindModelAsync(ModelBindingContext bindingContext)
```
rynowak: By making this async, we've undone an optimization here. Notice `BindModelAync` now has a state machine. If there's not an easy way to address this, let's not block on it.

*Before*:
image

*After:*
image

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/DefaultModelBindingContext.cs
rynowak: I think this is fine where it is. We don't have a good reason to hide it, and users might want to unit test their code.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/DefaultModelBindingContext.cs
```diff
@@ -0,0 +1,324 @@
...
+                if (value.HasValue && value.Value == default(ModelBindingResult))
+                {
+                    throw new ArgumentException(nameof(ModelBindingResult));
+                }
+
+                _state.Result = value;
+            }
+        }
+
+
```
rynowak: Don't need extra line

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/DefaultModelBindingContext.cs
```diff
@@ -0,0 +1,324 @@
...
+            public ModelStateDictionary ModelState;
+
+            public string BinderModelName;
+            public BindingSource BindingSource;
+            public Type BinderType;
+            public bool FallbackToEmptyPrefix;
+            public bool IsTopLevelObject;
+
+            public ModelBindingResult? Result;
+        };
```
rynowak: Don't need extra semicolon

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/DefaultModelBindingContext.cs
```diff
@@ -0,0 +1,324 @@
...
+            public OperationBindingContext OperationBindingContext;
+            public string FieldName;
+            public object Model;
+            public ModelMetadata ModelMetadata;
+            public string ModelName;
+
+            public IValueProvider ValueProvider;
+            public Func<ModelBindingContext, string, bool> PropertyFilter;
+            public ValidationStateDictionary ValidationState;
+            public ModelStateDictionary ModelState;
```
rynowak: OBC, ModelState, and ValidationState are really global things, you can't scope em. Removing these from `State` would trim a few bytes.

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/DefaultModelBindingContextTest.cs
```diff
@@ -0,0 +1,95 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Mvc.ModelBinding.Test
```
rynowak: Leave out `.Test`

rynowak:
:watch: for minor comments

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CompositeModelBinder.cs
rynowak: There's a bunch of these. I'd do a find and replace for `<see cref="DefaultModelBindingContext"/>` - there really shouldn't be any of these except on the class itself.

rynowak:
Some numbers for this change:

*Before:*
image

*After:*
image

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CompositeModelBinder.cs
dougbu: Fine as it is now w/ `FallbackToEmptyPrefix = false` in the other `EnterNestedScope()` overload.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
lodejard: I didn't write the text, just moved it from one class to another, would prefer to defer the copy edits for an API update sweep.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/DefaultModelBindingContext.cs
lodejard: :tired_face: :gun:

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerArgumentBinder.cs
lodejard: :see_no_evil:

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CompositeModelBinder.cs
lodejard: :see_no_evil:

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ModelBinderExtensions.cs
lodejard: :see_no_evil:

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ControllerActionInvokerTest.cs
lodejard: :see_no_evil:

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/ArrayModelBinderTest.cs
lodejard: :see_no_evil:

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/CompositeModelBinderTest.cs
lodejard: :see_no_evil:

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
lodejard: Is this comment wrong?

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
lodejard: :see_no_evil:

lodejard:
:unlock: :watch: for comments

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
dougbu: Yes, `<see cref="ModelState"/>` references the property it's describing and not the intended type.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
dougbu: FYI if you fixed this, would be no need for a follow-up pass on this file.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CompositeModelBinder.cs
```diff
@@ -36,69 +36,86 @@ public CompositeModelBinder(IList<IModelBinder> modelBinders)
...
+            //ModelBindingResult? result = null;
+            //using (bindingContext.EnterNestedScope())
+            //{
+            //    if (PrepareBindingContext(bindingContext))
+            //    {
+            //        await RunModelBinders(bindingContext);
+            //        result = bindingContext.Result;
+            //    }
+            //}
+            //bindingContext.Result = result;
```
dougbu: Remove commented-out code.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
rynowak: > prefer to defer the copy edits for an API update sweep.
> FYI if you fixed this, would be no need for a follow-up pass on this file.

To speak to this, there is no planned API update sweep I've heard of.

At the same time, this is an extensibility point used by users at a medium frequency. If there are issues caused by the changes in this PR we should fix them.

This isn't the place to do a broad pass on all the existing docs for formatting. Open a separate issue if that's needed so it doesn't get lost.

rynowak:
:shipit: from me after fixing minor comments, all of my major feedback has been addressed

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
lodejard: Understood, but this is specifically a tactical perf PR. These lines of text didn't even change - they just moved.

rynowak: Let's be practical. If we're arguing about one comment, just fixing will be simpler than the alternatives. If this is about the docs of this whole class, or whole feature area, then file and issue and move on.

lodejard: Ah - I hadn't seen your note to open a separate issue - I wouldn't have replied if I had read that. See #4026

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
lodejard: I believe in this comment it was intended to cref the property - there is no ModelState class, only the ModelStateDictionary

:see_no_evil: TBD in #4026 if needed

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBindingContext.cs
```diff
@@ -232,69 +81,93 @@ public ModelStateDictionary ModelState
...
-            set
-            {
-                if (value == null)
-                {
-                    throw new ArgumentNullException(nameof(value));
-                }
+        /// <param name="modelMetadata"><see cref="ModelBinding.ModelMetadata"/> to assign to the <see cref="ModelMetadata"/> property</param>
+        /// <param name="fieldName">Name to assign to the <see cref="FieldName"/> property</param>
+        /// <param name="modelName">Name to assign to the <see cref="ModelName"/> property</param>
+        /// <param name="model">Instance to assign to the <see cref="Model"/> property</param>
```
dougbu: End these new guys with a period.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CompositeModelBinder.cs
```diff
@@ -36,70 +36,77 @@ public CompositeModelBinder(IList<IModelBinder> modelBinders)
...
 
-            var newBindingContext = CreateNewBindingContext(bindingContext);
-            if (newBindingContext == null)
-            {
-                // Unable to find a value provider for this binding source. Binding will fail.
-                return ModelBindingResult.NoResultAsync;
-            }
-
-            return RunModelBinders(newBindingContext);
+            return RunModelBinders(bindingContext);
```
dougbu: No longer any value in having `RunModelBinders()` separate. Might not be a big deal since the old `CreateNewBindingContext()` rarely failed. But we should measure a couple of scenarios to be sure we're not losing perf w/o the previous optimization here.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CompositeModelBinder.cs
```diff
@@ -36,70 +36,77 @@ public CompositeModelBinder(IList<IModelBinder> modelBinders)
...
+                                    }
+
+                                    overallResult = bindingContext.Result;
+                                    return;
+                                }
+
+                                // Current binder should have been able to bind value but found nothing. Exit loop in a way that
+                                // tells caller to fall back to the empty prefix, if appropriate. Do not return result because it
+                                // means only "other binders are not applicable".
+                                overallResult = null;
```
dougbu: Confusing to read: How can `overallResult` be non-`null` here?

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/FormCollectionModelBinder.cs
```diff
@@ -4,9 +4,11 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Diagnostics;
```
dougbu: Please clean out the now-extra added `using`s for `System.Diagnostics`.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/MutableObjectModelBinder.cs
```diff
@@ -322,7 +324,7 @@ private static bool CanUpdateReadOnlyProperty(Type propertyType)
         /// Creates suitable <see cref="object"/> for given <paramref name="bindingContext"/>.
         /// </summary>
         /// <param name="bindingContext">The <see cref="ModelBindingContext"/>.</param>
-        /// <returns>An <see cref="object"/> compatible with <see cref="ModelBindingContext.ModelType"/>.</returns>
+        /// <returns>An <see cref="object"/> compatible with <see cref="DefaultModelBindingContext.ModelType"/>.</returns>
```
dougbu: `ModelBindingContext.ModelType`. Need similar changes a couple of other places in this `class`.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CompositeModelBinder.cs
lodejard: @rynowak do you believe this was covered with your measurements, or do you think this happens often enough to be measured and changed specifically?

test/Microsoft.AspNetCore.Mvc.IntegrationTests/BinderTypeBasedModelBinderIntegrationTest.cs
```diff
@@ -280,11 +282,17 @@ private class Address
...
         private class AddressModelBinder : IModelBinder
         {
-            public Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindingContext)
+            public Task BindModelAsync(ModelBindingContext bindingContext)
             {
+                if (bindingContext == null)
+                {
+                    throw new ArgumentNullException(nameof(bindingContext));
+                }
+                Debug.Assert(bindingContext.Result == null);
```
dougbu: Minor but you don't need any of this in test code.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CompositeModelBinder.cs
rynowak: The value in having these separate was in reducing the number of state machines that are allocated. I don't think we need to measure more, separate is better.

dougbu:
:shipit:

PR #4010, opened 2016-01-28
Validation nation

rynowak:


src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/BodyModelBinder.cs
```diff
@@ -65,9 +65,6 @@ public Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindingContex
                 throw new ArgumentNullException(nameof(bindingContext));
             }
 
-            // For compatibility with MVC 5.0 for top level object we want to consider an empty key instead of
-            // the parameter name/a custom name. In all other cases (like when binding body to a property) we
-            // consider the entire ModelName as a prefix.
```
dougbu: This comment still applies because errors the input formatter adds still use this prefix.

Further, conversion errors e.g. during XML input formatting can still lead to a `ModelState` entry w/ `Key == string.Empty`. That probably isn't what we want.

How about removing the `string.Empty` special case entirely? That is, removing the value-only `ModelState` entry doesn't clean up all `ModelState` entries this binder may add.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CancellationTokenModelBinder.cs
```diff
@@ -16,7 +17,9 @@ public Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindingContex
         {
             if (bindingContext.ModelType == typeof(CancellationToken))
             {
-                var model = bindingContext.OperationBindingContext.HttpContext.RequestAborted;
+                // Force boxing so we can maintain a reference to this token in validation state.
```
dougbu: Could you clarify this comment? See the `ModelBindingResult` and `ValidationStateDictionary` will store the same reference for the token. But the comment doesn't really make the implication of that consistency clear.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CompositeModelBinder.cs
```diff
@@ -76,12 +76,13 @@ public virtual Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindi
...
+                                entry = new ValidationStateEntry()
+                                {
+                                    Key = result.Key,
+                                    Metadata = bindingContext.ModelMetadata,
+                                };
                                 bindingContext.ValidationState.Add(result.Model, entry);
                             }
-
-                            entry.Key = entry.Key ?? result.Key;
-                            entry.Metadata = entry.Metadata ?? bindingContext.ModelMetadata;
```
dougbu: This previously cleaned up incomplete `ValidationStateDictionary` entries, not just the one added just above. Not sure how important that was but it certainly compensated for odd 3rd party binders.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/BodyModelBinder.cs
rynowak: Sure, I'll put this comment back. It wasn't intentional, just lost in the shuffle.

> How about removing the string.Empty special case entirely? That is, removing the value-only ModelState entry doesn't clean up all ModelState entries this binder may add.

What would you suggest instead? If we're adding binding errors we still need to provide a key.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CancellationTokenModelBinder.cs
rynowak: Not in the existing code they don't. The comment basically says not to remove the cast. I can make it more explicit if necessary.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CompositeModelBinder.cs
rynowak: I don't think this was good to have. This was fighting with placed in code where we want to have a `null` key on purpose.

The `ValidationVisitor` already knows how to compensate for incomplete entries, so this wasn't helping.

rynowak:
@dougbu - added fixes for failing tests

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/BodyModelBinder.cs
dougbu: @rynowak I suggest removing the comment and changing the following line to
``` c#
    var modelBindingKey = bindingContext.ModelName;
```

That would ensure the empty prefix isn't used by the JSON input formatter, when the XML input formatter `throw`s (adding an error), or in the returned `ModelBindingResult`. That would avoid all conflicting keys in the `ModelStateDictionary`, not just one used on the happy path.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CancellationTokenModelBinder.cs
dougbu: I agree you're fixing things here. I'm asking for a bit more in the comment because it didn't help me understand the issue i.e. what shouldn't be reverted later.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CompositeModelBinder.cs
dougbu: :ok:

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Validation/ValidationVisitor.cs
```diff
@@ -168,11 +169,12 @@ private bool Visit(ModelMetadata metadata, string key, object model)
             }
             else if ((entry != null && entry.SuppressValidation))
             {
-                SuppressValidation(key);
+                // Use the key on the entry, because we might not have entries in model state.
+                SuppressValidation(entry?.Key);
```
dougbu: Why `?.`? You've checked that `entry != null`.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Validation/ValidationVisitor.cs
```diff
@@ -168,11 +169,12 @@ private bool Visit(ModelMetadata metadata, string key, object model)
             }
             else if ((entry != null && entry.SuppressValidation))
```
dougbu: While here, could you remove the extra parentheses?

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/BodyModelBinderTests.cs
```diff
@@ -73,7 +72,7 @@ public class BodyModelBinderTests
             Assert.False(binderResult.IsModelSet);
             Assert.Null(binderResult.Model);
 
-            // Key is empty because this was a top-level binding.
+            // Key is the fieldName because this was a top-level binding.
             var entry = Assert.Single(bindingContext.ModelState);
             Assert.Equal(string.Empty, entry.Key);
```
dougbu: Doesn't the fact this `Key` is still empty indicate the problem remains?

test/Microsoft.AspNetCore.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
```diff
@@ -1355,13 +1344,10 @@ private class Person9
             Assert.NotNull(model.Customer.Address);
             Assert.Equal(AddressStreetContent, model.Customer.Address.Street);
 
-            Assert.Equal(1, modelState.Count);
+            Assert.Equal(0, modelState.Count);
```
dougbu: Like the later `Assert.Empty()` better.

dougbu:
:watch:

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/BodyModelBinder.cs
rynowak: But `bindingContext.ModelName` will be the empty string most of the time. You can change the code, but it doesn't solve a problem.

Do you have another suggestion?

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/BodyModelBinderTests.cs
rynowak: Not really in the same way. You've got a definite model state error because the body failed to deserialize. 

The bug we're fixing is about missing validation errors. 

Let's talk more and see if we can solve both issues.

test/Microsoft.AspNetCore.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
rynowak: These are all over the test code. It's not wrong. I'd suggest not cleaning it up.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/BodyModelBinder.cs
dougbu: The following code in `CompositeModelBinder` and the fact `ModelName` is initially equal to `FieldName` implies the opposite.
``` c#
    if (bindingSource != null && bindingSource.IsGreedy)
    {
        newBindingContext.ModelName = oldBindingContext.ModelName;
    }
```

But if I'm reading that wrong, how 'bout
``` c#
    var modelBindingKey = bindingContext.ModelName;
    if (string.IsNullOrEmpty(modelBindingKey)
    {
        modelBindingKey = bindingContext.FieldName;
    }
```

rynowak:
@dougbu - updated, squashed and rebased

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CancellationTokenModelBinder.cs
```diff
@@ -16,7 +17,12 @@ public Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindingContex
         {
             if (bindingContext.ModelType == typeof(CancellationToken))
             {
-                var model = bindingContext.OperationBindingContext.HttpContext.RequestAborted;
+                // We need to force boxing now, so we can insert the same reference to the boxed CancellationToken
+                // in both the ValidationState and ModelBindingResult.
+                //
+                // DO NOT simplify this code by removing the cast.
```
dougbu: :clap:

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/FormFileModelBinder.cs
```diff
@@ -43,15 +43,22 @@ public Task<ModelBindingResult> BindModelAsync(ModelBindingContext bindingContex
 
         private async Task<ModelBindingResult> BindModelCoreAsync(ModelBindingContext bindingContext)
         {
+            // If we're at the top level, then use the FieldName (paramter or property name).
+            // This handles the fact that there will be nothing in the ValueProviders for this parameter
+            // and so we'll do the right thing even though we 'fell-back' to the empty prefix.
+            var modelName = bindingContext.IsTopLevelObject
+                ? bindingContext.FieldName
```
dougbu: Thinking of what you did in `BodyModelBinder`, should this use `BinderModelName` if set?

test/Microsoft.AspNetCore.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
dougbu: I was referring to the assertion you _added_ @ [line 1350](https://github.com/aspnet/Mvc/pull/4010/files#diff-a17a31e64e5fd7e2f15d089bb238bed0R1350).

test/Microsoft.AspNetCore.Mvc.IntegrationTests/ValidationIntegrationTests.cs
```diff
@@ -1209,13 +1202,153 @@ private class Country
...
+                {
+                    BindingSource = BindingSource.Body
+                },
+                ParameterType = typeof(Greeting)
+            };
+
+            var operationContext = ModelBindingTestHelper.GetOperationBindingContext(
+                request =>
+                {
+                    request.Body = new MemoryStream(Encoding.UTF8.GetBytes("{ message: \"Hello There\" }"));
```
dougbu: Comment about string being too long would avoid future line-by-line comparison w/ `FromBody_WithEmptyPrefix_DoesNotSuppressUnrelatedErrors_Valid()`.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/BodyModelBinder.cs
dougbu: Like the final compromise.

dougbu:
:shipit: don't want to block you on https://github.com/aspnet/Mvc/pull/4010#discussion_r51300495 question. That would be a change in behaviour while you're mostly making `FormFileModelBinder` more consistent here.

test/Microsoft.AspNetCore.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
rynowak: I know. Both styles are all over the code, and not wrong.

test/Microsoft.AspNetCore.Mvc.IntegrationTests/ValidationIntegrationTests.cs
rynowak: Sure

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/FormFileModelBinder.cs
rynowak: Sure

rynowak: Sure

PR #4011, opened 2016-01-29
Fixing ArgumentOutOfRangeException in ExpressionRewriter

ajaybhargavb:
Issue - https://github.com/aspnet/Razor/issues/671

@rynowak

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor/Internal/ExpressionRewriter.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor/Internal/ExpressionRewriter.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq.Expressions;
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Mvc.Razor/Internal/ExpressionRewriter.cs
```diff
@@ -137,49 +137,38 @@ public override SyntaxNode VisitSimpleLambdaExpression(SimpleLambdaExpressionSyn
...
+                }
 
-            memberAccess = memberAccess.WithTrailingTrivia(innerTrivia);
+                var padding = token.Text.Length;
+                // Need to exclude the length of the member name from the padding.
+                while (charactersToExclude > 0 && padding > 0)
+                {
+                    padding--;
+                    charactersToExclude--;
+                }
```
rynowak: Isn't this just subtraction with `Math.Min(padding, 0)`? It would be easier to understand written like that

rynowak:
:shipit:

src/Microsoft.AspNetCore.Mvc.Razor/Internal/ExpressionRewriter.cs
ajaybhargavb: It is not straightforward. `Math.Min(padding, 0)` will always return 0. I can change this while to 
```C#
if (padding > charactersToExclude)
{
    padding -= charactersToExclude;
    charactersToExclude = 0;
}
else
{
    charactersToExclude -= padding;
    padding = 0;
}
```
But I think they are both equally readable.

dougbu: Like your update @ajaybhargavb and think it's more readable. The `while` looks strange.

PR #4012, opened 2016-01-29
Support overrides for client-side validation messages

dougbu:
- #2969
- `RemoteAttribute` did not support `IStringLocalizer` overrides
 - use same `MvcDataAnnotationsLocalizationOptions` property as for other `ValidationAttribute`s
- error message `NumericClientModelValidator` added could not be overridden
 - not related to `IStringLocalizer` because users have no way to set the resource lookup key
 - extend `IModelBindingMessageProvider` to add the necessary `Func<,>`
- also correct problem using resources with `RemoteAttribute` and add lots of tests

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/Metadata/IModelBindingMessageProvider.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/Metadata/IModelBindingMessageProvider.cs
@@ -51,5 +51,12 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding.Metadata
         /// </summary>
         /// <value>Default <see cref="string"/> is "The value '{0}' is invalid.".</value>
         Func<string, string> ValueIsInvalidAccessor { get; }
+
+        /// <summary>
+        /// Error message HTML and tag helpers add for client-side validation of numeric formats. Visible in the
+        /// browser if the field for a <c>float</c> property (for example) does not have a correctly-formatted value.
+        /// </summary>
+        /// <value>Default <see cref="string"/> is "The field {0} must be a number.".</value>
+        Func<string, string> ValueMustBeANumberAccessor { get; }
     }
 }
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/RemoteAttribute.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/RemoteAttribute.cs
@@ -6,12 +6,14 @@ using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
 using System.Globalization;
 using System.Linq;
-using Microsoft.AspNetCore.Mvc.ViewFeatures;
-using Microsoft.AspNetCore.Mvc.Internal;
+using Microsoft.AspNetCore.Mvc.DataAnnotations;
 using Microsoft.AspNetCore.Mvc.ModelBinding.Validation;
+using Microsoft.AspNetCore.Mvc.Routing;
+using Microsoft.AspNetCore.Mvc.ViewFeatures;
 using Microsoft.AspNetCore.Routing;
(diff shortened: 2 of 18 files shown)


src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/Metadata/IModelBindingMessageProvider.cs
```diff
@@ -51,5 +51,12 @@ public interface IModelBindingMessageProvider
         /// </summary>
         /// <value>Default <see cref="string"/> is "The value '{0}' is invalid.".</value>
         Func<string, string> ValueIsInvalidAccessor { get; }
+
+        /// <summary>
+        /// Error message HTML and tag helpers add for client-side validation of numeric formats. Visible in the
+        /// browser if the field for a <c>float</c> property (for example) does not have a correctly-formatted value.
+        /// </summary>
+        /// <value>Default <see cref="string"/> is "The field {0} must be a number.".</value>
+        Func<string, string> ValueMustBeANumberAccessor { get; }
```
dougbu: Already discussed this small addition offline w/ @Eilon.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/RemoteAttribute.cs
```diff
@@ -33,7 +37,7 @@ public class RemoteAttribute : ValidationAttribute, IClientModelValidator
         /// Intended for subclasses that support URL generation with no route, action, or controller names.
         /// </remarks>
         protected RemoteAttribute()
-            : base(Resources.RemoteAttribute_RemoteValidationFailed)
+            : base(errorMessageAccessor: () => Resources.RemoteAttribute_RemoteValidationFailed)
```
dougbu: Previous approach left `ErrorMessage` non-`null`, disallowing use of `ErrorMessageResourceName` and `ErrorMessageResourceType`.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/RemoteAttributeTest.cs
```diff
@@ -155,6 +159,70 @@ public void Constructor_WithActionControllerAreaName_UpdatesAreaRouteData(string
...
+            // Act
+            var message = attribute.FormatErrorMessage("Property1");
+
+            // Assert
+            Assert.Equal(expected, message);
+        }
+
+        [Fact]
+        [ReplaceCulture]
+        public void FormatErrorMessage_UsesErrorMessageFromResource()
```
dougbu: This test and any other that accessed the `protected string ErrorMessageString` from ValidationAttribute would have failed w/o `RemoteAttribute` constructor change.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/RemoteAttributeTest.cs
```diff
@@ -518,35 +884,59 @@ public void AddValidation_DoesNotTrounceExistingAttributes()
                 kvp => { Assert.Equal("data-val-remote-url", kvp.Key); Assert.Equal("original", kvp.Value); });
         }
 
-        private static ClientModelValidationContext GetValidationContext(IUrlHelper urlHelper)
```
dougbu: Remaining changes to this class are about rearranging things enough to allow `IModelMetadataProvider` and `IStringLocalizerFactory` overrides. Also made the various `Mock`s strict to ensure only the expected calls occur.

rynowak:
:shipit: from me - seems simple enough. Get @ryanbrandenburg to look as well.

dougbu:
Ah, guess I need that __@ Full Ryan__ approval :smiley_cat:

ryanbrandenburg:
Looks like a :shipit: to me.

dougbu:
26a14a2

PR #4017, opened 2016-01-29
Be smarter about StringBuilders

rynowak:
/cc @Eilon 

This saves about 40mb of allocations by moving a declaration out of the loop.

*Before*:
image

*After*:
image


pranavkm:
:shipit:

NTaylorMullen:
:shipit:

SajayAntony:
This will probably lower your allocation rate even further - https://github.com/dotnet/roslyn/blob/master/src/Compilers/Core/Portable/Collections/PooledStringBuilder.cs 
Maybe as a part of another change :)

rynowak:
> This will probably lower your allocation rate even further 

We already have this actually :+1: : https://github.com/aspnet/Common/blob/dev/src/Microsoft.Extensions.ObjectPool/ObjectPoolProviderExtensions.cs#L15

Going to hold off on integrating it for now since this is only allocating one per request.

PR #4019, opened 2016-01-30
Fix line endings in HTML helper tests for x-plat

dougbu:


dougbu:
FYI only, will push immediately.
Fixing Travis failures. No Windows impact or product code changes.

/cc @ajaybhargavb

Issue #3293, opened 2015-10-09
Allow to define section or/and body without the need to render it


danilodeven:
It would be nice if we could just prevent the render of the `body` or/and the `section` like this without get the exception:
```c#
@if (someLogicIsFalseHere) {
    RenderSection("adminMenu");
}
```
###### `InvalidOperationException: The following sections have been defined but have not been rendered by the page at '/Views/Shared/_Layout.cshtml': 'adminMenu'.`

I use the "section" to make various versions of the views using the same "layout.cshtml" for different browsers, users, permissions and requests types with the same controller and model.

The way I current make this in asp.net 4.5 is like the code bellow, but this workaround doesn't work properly in asp.net 5 beta 7. This was discussed in the issue https://github.com/aspnet/Mvc/issues/3283
```c#
@if (someLogicIsFalseHere) {
    WriteTo(TextWriter.Null, RenderBody());
    WriteTo(TextWriter.Null, RenderSection("adminMenu"));
    RenderSection("alternativeMenu");
}
```

Eilon:
This issue requires just a bit of design. We're happy to collaborate with folks on how this could be implemented.

pranavkm:
Does it execute the body at all? Or is the rendering the only thing that's supressed?

danilodeven:
This is giving the error just when the logic doesn't call the `@RenderSection("adminSection")` and works fine when it doesn't call the `@RenderBody()`. (the `@RenderBody` need to appears somewhere in the `Layout.cshtml` file at-list once.)
Appears that if suppress the obligation to render the defined sections would do the trick. The `@RenderSection()` should work like the `@RenderBody()`.

Bellow all the code I used to test:

`Home/Index.cshtml`
```html
@{
	Layout = "_Layout";
	ViewData["Title"] = "Home Page";
}
@section adminSection {
	<div>
		<p>You are administrator. Take some action:</p>
		<p>
			<button id="action1">See the new registers</button>
			<button id="action2">Do something else</button>
		</p>
	</div>
}
<div>
	<p>Welcome new user. Register your self</p>
	<form method="POST">
		<input name="fullName"/>
		<input name="email"/>
		<input type="submit"/>
	</form>
</div>
```
`Shared/_Layout.cshtml`
```html
<!DOCTYPE html>
<html>
<head>
	<meta charset="utf-8" />
	<meta name="viewport" content="width=device-width, initial-scale=1.0" />
[...]

PR #4024, opened 2016-02-01
Allow to define section or/and body without the need to render it

ctyar:
Issue - #3293


Eilon:
@ctyar thanks for sending the PR. The code overall looks good and I have some questions about the behavior:

1. Should it be an error to ignore a section that isn't defined? E.g. should `IgnoreSection("ASDFASDF");` throw an exception?
2. Did you want to change the error message for when a section is defined but isn't rendered or ignored? In the bug you mentioned a change, though I suggest slightly different wording to add: `To ignore an unrendered section call {n}("sectionName").` and pass in `nameof(IgnoreSection)` for the `{n}` substitution parameter (for whatever `n` is).

src/Microsoft.AspNetCore.Mvc.Razor/RazorPage.cs
```diff
@@ -40,6 +40,8 @@ public abstract class RazorPage : IRazorPage
         private TagHelperAttributeInfo _tagHelperAttributeInfo;
         private HtmlContentWrapperTextWriter _valueBuffer;
         private IViewBufferScope _bufferScope;
+        private bool _ignoreBody;
+        private readonly HashSet<string> _ignoredSections = new HashSet<string>();
```
pranavkm: Could we lazily initialize this? We could avoid allocating the `HashSet` if this feature isn't used. 
```C#
public void IgnoreSection(string name)
{
   ...
   if (_ignoredSections == null)
   {
      _ignoredSections = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
    }
    ...
}
```

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorPageTest.cs
```diff
@@ -472,6 +472,23 @@ public void IsSectionDefined_ThrowsIfPreviousSectionWritersIsNotRegistered()
...
+            var page = CreatePage(v =>
+            {
+            });
+            page.Path = path;
+            page.BodyContent = new HtmlString("some content");
+            page.IgnoreBody();
+
+            // Act & Assert (does not throw)
+            await page.ExecuteAsync();
+            page.EnsureRenderedBodyOrSections();
```
pranavkm: Could you verify that content written by the page - inside the delegate for `CreatePage` - gets written out (and not ignored?)

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorPageTest.cs
```diff
@@ -497,6 +514,28 @@ public void IsSectionDefined_ThrowsIfPreviousSectionWritersIsNotRegistered()
         }
 
         [Fact]
+        public async Task EnsureRenderedBodyOrSections_SucceedsIfDefinedSectionsAreNotRendered_AndIgnored()
```
pranavkm: One more test that verifies that `IgnoreSection` does not affect `EnsureRenderedBodyOrSections` for sections that aren't ignored?

pranavkm:
:watch:

Eilon:
@pranavkm can you follow up on this again?

pranavkm:
Sure. I was waiting to hear from @ctyar

ctyar:
@pranavkm Sorry for late response, I fixed some of feedbacks and I think I need more info for rest of them.
One more thing can't build the repo right now, should I push anyway?

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorPageTest.cs
ctyar: I'm not sure I understand what you mean here, this test wont write anything to the page
because body is ignored and no sections defined

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorPageTest.cs
ctyar: Isn't `EnsureRenderedBodyOrSections_SucceedsIfRenderBodyIsNotCalled_ButAllDefinedSectionsAreRendered` already checks this?

pranavkm: In this test, you'd do something along the lines of 

```C#
page.PreviousSectionWriters = new Dictionary<string, RenderAsyncDelegate>
{
           { "section1", _nullRenderAsyncDelegate },
           { "section2", _nullRenderAsyncDelegate },
};

page.IgnoreSection("section1");

// Assert section2 is rendered.
```
Just to make sure we don't do something bad and ignore all sections in general after a `IgnoreSection` is called.

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorPageTest.cs
pranavkm: The suggestion was to add a `v.WriteLiteral(expected);` inside `CreatePage` and do a `Assert.Equal(expected, page.RenderedContent);` at the end. The intent is to verify that it's suppresses `RenderBody` content and not any of the content written by the layout.

pranavkm:
Yup. If you can get your code to compile that should suffice for now. Our CI's been in a state of flux for a while now so it's hard to build about anything.

pranavkm:
Merged in https://github.com/aspnet/Mvc/commit/1e5b0b9becab1e6ba2a4978df880d74fba234439. Thanks for the PR!

PR #4025, opened 2016-02-01
 * Give message correct path.

### LINKED ISSUE #3930: Suggestion: error message indicates layout instead of current view.
When having a layout with a `RenderSection` required, and then you go to a View without that section defined, the error message is:

```
Section 'sectionName' is not defined in path '/Views/Shared/_Layout.cshtml'.
```

but I think it should be:

```
Section 'sectionName' is not defined in path '/Views/Home/Index.cshtml'.
```

ryanbrandenburg:
Fixes #3930

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor/RazorPage.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor/RazorPage.cs
@@ -443,7 +443,7 @@ namespace Microsoft.AspNetCore.Mvc.Razor
             // null  or false. Consequently defer the prefix generation until we encounter the attribute value.
             if (attributeValuesCount != 1)
             {
-               WritePositionTaggedLiteral(writer, prefix, prefixOffset);
+                WritePositionTaggedLiteral(writer, prefix, prefixOffset);
             }
         }
 
@@ -800,7 +800,11 @@ namespace Microsoft.AspNetCore.Mvc.Razor
             else if (required)
             {
                 // If the section is not found, and it is not optional, throw an error.
-                throw new InvalidOperationException(Resources.FormatSectionNotDefined(sectionName, Path));
+                var message = Resources.FormatSectionNotDefined(
+                    ViewContext.ExecutingFilePath,
+                    sectionName,
+                    ViewContext.View.Path);
+                throw new InvalidOperationException(message);
             }
             else
             {
--- a/test/Microsoft.AspNetCore.Mvc.FunctionalTests/RazorViewLocationSpecificationTest.cs
+++ b/test/Microsoft.AspNetCore.Mvc.FunctionalTests/RazorViewLocationSpecificationTest.cs
@@ -4,6 +4,7 @@
 using System.Net.Http;
 using System.Threading.Tasks;
 using Xunit;
+using System.Net;
 
 namespace Microsoft.AspNetCore.Mvc.FunctionalTests
 {
@@ -92,6 +93,22 @@ Non Shared Partial
 
             // Assert
             Assert.Equal(expected, body.Trim(), ignoreLineEndingDifferences: true);
+        }
+
+        [Fact]
+        public async Task PartialLayout_ThrowsIfRequiredSectionMissing()
+        {
+            // Arrange
+            var path = "http://localhost/PartialViewEngine/ViewPartialMissingSection";
+
+            // Act
+            var content = await (await Client.GetAsync(path)).Content.ReadAsStringAsync();
+
+            // Assert
+            Assert.Contains(
+                "The layout page '/Views/Shared/_PartialLayout.cshtml' cannot find the section " +
+                    "'section' in the content page '/Views/PartialViewEngine/PartialMissingSection.cshtml'.",
+                WebUtility.HtmlDecode(content));
         }
     }
 }
--- a/test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorPageTest.cs
+++ b/test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorPageTest.cs
@@ -284,11 +284,12 @@ namespace Microsoft.AspNetCore.Mvc.Razor
         public async Task RenderSection_ThrowsIfRequiredSectionIsNotFound()
         {
             // Arrange
+            var context = CreateViewContext(viewPath: "/Views/TestPath/Test.cshtml");
+            context.ExecutingFilePath = "/Views/Shared/_Layout.cshtml";
             var page = CreatePage(v =>
             {
-                v.Path = "/Views/TestPath/Test.cshtml";
                 v.RenderSection("bar");
-            });
+            }, context: context);
             page.PreviousSectionWriters = new Dictionary<string, RenderAsyncDelegate>
             {
                 { "baz", _nullRenderAsyncDelegate }
@@ -296,7 +297,9 @@ namespace Microsoft.AspNetCore.Mvc.Razor
 
             // Act & Assert
             var ex = await Assert.ThrowsAsync<InvalidOperationException>(() => page.ExecuteAsync());
-            Assert.Equal("Section 'bar' is not defined in path '/Views/TestPath/Test.cshtml'.", ex.Message);
+            var message = $"The layout page '/Views/Shared/_Layout.cshtml' cannot find the section 'bar'" +
+                " in the content page '/Views/TestPath/Test.cshtml'.";
+            Assert.Equal(message, ex.Message);
         }
 
         [Fact]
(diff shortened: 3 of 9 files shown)


src/Microsoft.AspNetCore.Mvc.Razor/RazorPage.cs
```diff
@@ -800,7 +800,8 @@ public Task<HtmlString> RenderSectionAsync(string name, bool required)
             else if (required)
             {
                 // If the section is not found, and it is not optional, throw an error.
-                throw new InvalidOperationException(Resources.FormatSectionNotDefined(sectionName, Path));
+                var message = Resources.FormatSectionNotDefined(sectionName, ViewContext.View.Path);
```
ryanbrandenburg: Someone more familiar with the internals of view/viewcontext might want to double check my logic that ViewContext.View.Path is always what we're looking for here.

dougbu: Use `ExecutingFilePath` which is updated as different pages execute. Will in this case mean the offending `_Layout.cshtml` file is identified, not the loaded view.

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorPageTest.cs
```diff
@@ -286,9 +286,9 @@ public static TheoryData<HtmlEncoder> HtmlEncoderData
             // Arrange
             var page = CreatePage(v =>
             {
-                v.Path = "/Views/TestPath/Test.cshtml";
+                v.Path = "/Views/Shared/_Layout.cshtml";
                 v.RenderSection("bar");
-            });
+            }, context: CreateViewContext(viewPath: "/Views/TestPath/Test.cshtml"));
```
dougbu: Add a new test rather than updating the existing one.

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorPageTest.cs
```diff
@@ -1188,7 +1188,7 @@ public static TheoryData WriteAttributeData
             return view.Object;
         }
 
-        private static ViewContext CreateViewContext(TextWriter writer = null)
+        private static ViewContext CreateViewContext(TextWriter writer = null, string viewPath = null)
```
dougbu: Shouldn't need to make changes here. `RazorView.Path` is almost-never referenced in product code (your incorrect update aside).

src/Microsoft.AspNetCore.Mvc.Razor/RazorPage.cs
dougbu: Oops, this one is strange. Looked at the bug and see we _want_ to use `View.Path` here. That's unusual but I can see the reasoning.

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorPageTest.cs
dougbu: Ignore previous comment here since product code _will_ be using `IView.Path` in some tests.

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorPageTest.cs
ryanbrandenburg: Do we still want a new test now that we're going with the end view? Because it seems like this is the exact scenario I'm working on, I don't really know what other test I would make.

Eilon: Yeah this seems like exactly the test that needs to be updated... the name of this test method couldn't be any better...

dougbu: :+1:

ryanbrandenburg:
:up: :date:

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorPageTest.cs
```diff
@@ -296,7 +297,8 @@ public static TheoryData<HtmlEncoder> HtmlEncoderData
 
             // Act & Assert
             var ex = await Assert.ThrowsAsync<InvalidOperationException>(() => page.ExecuteAsync());
-            Assert.Equal("Section 'bar' is not defined in path '/Views/TestPath/Test.cshtml'.", ex.Message);
+            var message = $"The layout page '{context.ExecutingFilePath}' cannot find the section 'bar' in the content page '{context.View.Path}'.";
```
Eilon: Hmm why is this using string interpolation? This will essentially check that "the code does what the code does." By using a constant string you can verify that "the code produces this exact result."

test/WebSites/RazorWebSite/Views/PartialViewEngine/PartialMissingSection.cshtml
```diff
@@ -0,0 +1,7 @@
+﻿@*
+    For more information on enabling MVC for empty projects, visit http://go.microsoft.com/fwlink/?LinkID=397860
```
Eilon: Feel free to remove these comments :smile:

src/Microsoft.AspNetCore.Mvc.Razor/RazorPage.cs
```diff
@@ -800,7 +800,11 @@ public Task<HtmlString> RenderSectionAsync(string name, bool required)
             else if (required)
             {
                 // If the section is not found, and it is not optional, throw an error.
-                throw new InvalidOperationException(Resources.FormatSectionNotDefined(sectionName, Path));
+                var message = Resources.FormatSectionNotDefined(
+                    ViewContext.ExecutingFilePath,
```
dougbu: Not that important but at least FYI: `ExecutingFilePath` and `Path` are identical here. Executing the layout page.

test/Microsoft.AspNetCore.Mvc.FunctionalTests/RazorViewLocationSpecificationTest.cs
```diff
@@ -93,5 +94,20 @@ public RazorViewLocationSpecificationTest(MvcTestFixture<RazorWebSite.Startup> f
...
         }
+
+        [Fact]
+        public async Task PartialLayout_ThrowsIfRequiredSectionMissing()
+        {
+            // Arrange
+            var path = "http://localhost/PartialViewEngine/ViewPartialMissingSection";
+
+            //Act
+            var ex = await Assert.ThrowsAsync<InvalidOperationException>(() => Client.GetAsync(path));
```
dougbu: This isn't a good functional test: Should never rely on `Exception`s across the client / server boundary. Could instead catch the expected `InvalidOperationException` in an `ExceptionFilterAttribute` and return the expected content. Bit uglier and definitely not recommended for production use (because the view could in general be left in a weird state) but could also use `try` / `catch` in the layout file.

test/WebSites/RazorWebSite/Views/PartialViewEngine/PartialMissingSection.cshtml
dougbu: Can also remove almost all content from the `.cshtml` files because they aren't included in an `HttpResponseMessage`. That is, unless you go with the "`try` / `catch` in the layout file" approach I suggested [above](https://github.com/aspnet/Mvc/pull/4025/files#r51515654).

test/Microsoft.AspNetCore.Mvc.FunctionalTests/RazorViewLocationSpecificationTest.cs
```diff
@@ -93,5 +94,20 @@ public RazorViewLocationSpecificationTest(MvcTestFixture<RazorWebSite.Startup> f
...
+
+            Assert.Equal(
+                "The layout page '/Views/Shared/_PartialLayout.cshtml' cannot find the section 'section'"
+                + "in the content page '/Views/PartialViewEngine/PartialMissingSection.cshtml'.",
```
dougbu: Looks like there's a space missing at the end of the first line. Does this test somehow succeed?

test/WebSites/RazorWebSite/Views/PartialViewEngine/ViewPartialMissingSection.cshtml
```diff
@@ -0,0 +1,9 @@
+﻿@*
+    For more information on enabling MVC for empty projects, visit http://go.microsoft.com/fwlink/?LinkID=397860
+*@
+@{
+    Layout = "/Views/Shared/_Layout.cshtml";
```
dougbu: Slightly prefer `Layout = null` here because the outer layout will never execute (again, unless you add a `try` block in the inner one).

dougbu:
:watch:

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorPageTest.cs
ryanbrandenburg: :see_no_evil:

ryanbrandenburg: :see_no_evil:

ryanbrandenburg:
:up: :date:

Eilon:
:shipit:

dougbu:
:shipit:

PR #4027, opened 2016-02-01
Reduce allocations in FormContext

rynowak:
This change reduces the allocations of `FormContext` instances and also of the resulting dictionary.

**Before** `(dictionary<string, bool> allocations)`
image

**After** `(dictionary<string, bool> allocations)`
image

The reduction in `FormContext` instances is worth about 400kb also.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Rendering/ViewContext.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Rendering/ViewContext.cs
@@ -15,9 +15,6 @@ namespace Microsoft.AspNetCore.Mvc.Rendering
     /// </summary>
     public class ViewContext : ActionContext
     {
-        // We need a default FormContext if the user uses HTML <form> instead of an MvcForm
-        private readonly FormContext _defaultFormContext = new FormContext();
-
         private FormContext _formContext;
         private DynamicViewData _viewBag;
 
@@ -84,7 +81,8 @@ namespace Microsoft.AspNetCore.Mvc.Rendering
             TempData = tempData;
             Writer = writer;
 
-            _formContext = _defaultFormContext;
+            FormContext = new FormContext();
+            
             ClientValidationEnabled = htmlHelperOptions.ClientValidationEnabled;
             Html5DateRenderingMode = htmlHelperOptions.Html5DateRenderingMode;
             ValidationSummaryMessageElement = htmlHelperOptions.ValidationSummaryMessageElement;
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.Mvc.ViewFeatures/Rendering/ViewContext.cs
```diff
@@ -15,9 +15,6 @@ namespace Microsoft.AspNetCore.Mvc.Rendering
     /// </summary>
     public class ViewContext : ActionContext
     {
-        // We need a default FormContext if the user uses HTML <form> instead of an MvcForm
-        private readonly FormContext _defaultFormContext = new FormContext();
```
rynowak: This plus the dictionary being non-lazy was what was so expensive.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Rendering/ViewContext.cs
```diff
@@ -144,14 +143,16 @@ public ViewContext()
...
             set
             {
-                // Never return a null form context, this is important for validation purposes.
-                _formContext = value ?? _defaultFormContext;
+                if (value == null)
+                {
+                    throw new ArgumentNullException(nameof(value));
+                }
+
+                _formContext = value;
```
rynowak: Opted for simpler semantics here rather than doing something lazy.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
```diff
@@ -801,8 +801,7 @@ public virtual IHtmlContent GenerateAntiforgery(ViewContext viewContext)
                 throw new ArgumentNullException(nameof(viewContext));
             }
 
-            var formContext = viewContext.ClientValidationEnabled ? viewContext.FormContext : null;
```
rynowak: This code wasn't really making use of the `FormContext`, discovered this while playing with lazy implementations. Refactoring for clarity.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/FormContext.cs
```diff
@@ -52,6 +51,19 @@ public IList<IHtmlContent> EndOfFormContent
...
+            get
+            {
+                if (_renderedFields == null)
+                {
+                    _renderedFields = new Dictionary<string, bool>(StringComparer.Ordinal);
+                }
+
+                return _renderedFields;
+            }
+        }
```
dougbu: This makes the code _look_ more performant but I doubt it matters beyond the dictionary for the outer ("non-form") `FormContext`. Most forms contain generated fields :smiling_imp: Unless we're optimizing at the level of one `Dictionary<,>` per request, I suggest leaving `_renderedFields` alone since it was simpler

On the other hand, the removal of a `new FormContext()` when calling the `ViewContext` copy constructor should be a significant win.

dougbu:
:shipit:

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/FormContext.cs
rynowak: This does matter. The alternative is to write complicated code in `ViewContext` to try and make the formcontext lazy there. I explained this earlier.

Issue #4004, opened 2016-01-27
Controller not passing arguments to ViewComponents


Dries-SF:
Hi,

I've run into a small problem which I wish to confirm.

With the recent changes to ViewComponents it is now called as such:

```C#
@await Component.InvokeAsync("Page", new { Model.MenuPageId })
```
Inside a Razorview the argument is passed correctly.
However the argument is not passed when calling it from a Controller as such:

```C#
    public class PageController : Controller
    {
        public IActionResult PagePreview(long id)
        {
            if (id == 0)
                return RedirectToAction("Index");
            var p = new Dictionary<string, object>();
            p.Add("menuPageId", id);
            return ViewComponent("Page",p);
        }
    }
```

I've also tried:

```C#
            return ViewComponent("Page",new { id } );
            return ViewComponent("Page",new { id } );
            return ViewComponent("Page",new {menuPageId = id});
```

ViewComponent code:
```C#
    public async Task<IViewComponentResult> InvokeAsync(long menuPageId)
        {
            IEnumerable<VMMenuPageEdit.VMArticle> q = from articlePI in _context.ArticlePageItems
                    join articleI in _context.Articles on articlePI.ArticleId equals articleI.Id
                    where articlePI.MenuPageId.Equals(menuPageId)
                    && articlePI.GetType().Equals(typeof(ArticlePageItem))
                    select new VMMenuPageEdit.VMArticle()
                    {
                        Id = articlePI.Id,
                        X = articlePI.XCoordinate,
                        Y = articlePI.YCoordinate,
                        Width = articlePI.Width,
                        Height = articlePI.Height,
                        ArticleId = articleI.Id,
                        Description = articleI.Description,
                        SellingPrice = articleI.SellingPrice
                    };

            return View(q);
        }
```

The menuPageId value in the ViewComponent remains 0 when passing through the Controller. It is the correct value when passing through a razor view.

Dependencies: 
```
    "EntityFramework.Commands": "7.0.0-*",
    "EntityFramework.MicrosoftSqlServer": "7.0.0-*",
    "Microsoft.AspNet.Diagnostics": "1.0.0-*",
    "Microsoft.AspNet.IISPlatformHandler": "1.0.0-*",
    "Microsoft.AspNet.Mvc": "6.0.0-*",
    "Microsoft.AspNet.Mvc.TagHelpers": "6.0.0-*",
    "Microsoft.AspNet.Mvc.ViewFeatures": "6.0.0-*",
[...]

PR #4028, opened 2016-02-01
Updated `Controller`s `ViewComponent` method to handle arguments.

NTaylorMullen:
- This bit was missed when changing the `ViewComponent` invocation pattern resulting in the inability to invoke `ViewComponent`s with arguments in a `Controller`.

#4004

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Controller.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Controller.cs
@@ -216,12 +216,16 @@ namespace Microsoft.AspNetCore.Mvc
         /// The view component name. Can be a view component
         /// <see cref="ViewComponents.ViewComponentDescriptor.ShortName"/> or
         /// <see cref="ViewComponents.ViewComponentDescriptor.FullName"/>.</param>
-        /// <param name="arguments">The arguments to pass to the view component.</param>
+        /// <param name="arguments">
+        /// An anonymous <see cref="object"/> containing arguments to be passed to the invoked view component method.
+        /// Alternatively, an <see cref="System.Collections.Generic.IDictionary{string, object}"/> instance containing
+        /// the invocation arguments.
+        /// </param>
         /// <returns>The created <see cref="ViewComponentResult"/> object for the response.</returns>
         [NonAction]
-        public virtual ViewComponentResult ViewComponent(string componentName, params object[] arguments)
+        public virtual ViewComponentResult ViewComponent(string componentName, object arguments)
         {
-            return new ViewComponentResult()
+            return new ViewComponentResult
             {
                 ViewComponentName = componentName,
                 Arguments = arguments,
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNetCore.Mvc.ViewFeatures/Controller.cs
```diff
@@ -216,12 +216,16 @@ public virtual PartialViewResult PartialView(string viewName, object model)
         /// The view component name. Can be a view component
         /// <see cref="ViewComponents.ViewComponentDescriptor.ShortName"/> or
         /// <see cref="ViewComponents.ViewComponentDescriptor.FullName"/>.</param>
-        /// <param name="arguments">The arguments to pass to the view component.</param>
+        /// <param name="arguments">
+        /// An <see cref="object"/> containing arguments to be passed to the invoked view component method.
```
rynowak: Should we specify that this should be an anonymous object? It's not clear from reading these docs what a user should **do**

test/Microsoft.AspNetCore.Mvc.FunctionalTests/BasicTests.cs
```diff
@@ -30,6 +30,31 @@ public BasicTests(MvcTestFixture<BasicWebSite.Startup> fixture)
...
+        public async Task CanRender_ViewComponentWithArgumentsFromController()
+        {
+            // Arrange
+            var expectedMediaType = MediaTypeHeaderValue.Parse("text/html; charset=utf-8");
+            var outputFile = "compiler/resources/BasicWebSite.PassThrough.Index.html";
+            var expectedContent =
+                await ResourceFile.ReadResourceAsync(_resourcesAssembly, outputFile, sourceFile: false);
+
+            // Act
+            // The host is not important as everything runs in memory and tests are isolated from each other.
```
rynowak: ???

test/Microsoft.AspNetCore.Mvc.FunctionalTests/compiler/resources/BasicWebSite.PassThrough.Index.html
```diff
@@ -0,0 +1 @@
+<p>The passed through value was: 123</p>
```
rynowak: was it?

rynowak:
:shipit: even though baselining is the **worst**

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Controller.cs
NTaylorMullen: Just copied and pasted from the `IViewComponentHelper`'s `InvokeAsync` method. Want me to update it in all locations?

test/Microsoft.AspNetCore.Mvc.FunctionalTests/BasicTests.cs
NTaylorMullen: Copy n pasted existing test. Figured this was consistent :smile:

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Controller.cs
rynowak: Yeah :-0

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Controller.cs
```diff
@@ -216,12 +216,16 @@ public virtual PartialViewResult PartialView(string viewName, object model)
         /// The view component name. Can be a view component
         /// <see cref="ViewComponents.ViewComponentDescriptor.ShortName"/> or
         /// <see cref="ViewComponents.ViewComponentDescriptor.FullName"/>.</param>
-        /// <param name="arguments">The arguments to pass to the view component.</param>
+        /// <param name="arguments">
+        /// An anonymous <see cref="object"/> containing arguments to be passed to the invoked view component method.
```
Eilon: Ack, this is **false**! There's no such thing as an **anonymous object**!!! There is only an anonymous **type**.

The parameters here are **objects** with properties representing the arguments of the invocation. Whether the caller uses an object of an anonymous type or not is irrelevant. It is 100% perfectly reasonable to use a named-type (or *onymous*, if you will), e.g. `return ViewComponent("MovieDetails", new MovieDetailInfo { id=123, Title="Great Movie", ... });`.

Please fix.

NTaylorMullen: Will fix that in [this](https://github.com/aspnet/Mvc/pull/4036) PR.

PR #4031, opened 2016-02-02
[Fixes #3915] Update FormFileModelBinder to avoid re-parsing the ContentDisposition header

### LINKED ISSUE #3915: Update FormFileModelBinder
https://github.com/aspnet/HttpAbstractions/issues/499

`IFormFile` now has a filename property, so we don't need to re-parse the content-disposition.

This is just code cleanup.

javiercn:


src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/FormFileModelBinder.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.IO;
```
javiercn: will remove this

rynowak:
:shipit: nice!

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/FormFileModelBinderTest.cs
```diff
@@ -82,17 +82,15 @@ public class FormFileModelBinderTest
...
             Assert.NotEqual(ModelBindingResult.NoResult, result);
-            var files = Assert.IsAssignableFrom<IList<IFormFile>>(result.Model);
-            Assert.Equal(2, files.Count);
+            var file = Assert.IsAssignableFrom<IFormFile>(result.Model);
+            Assert.Equal("file1.txt", file.FileName);
         }
 
         [Fact]
-        public async Task FormFileModelBinder_ExpectSingleFile_BindFirstFile()
+        public async Task FormFileModelBinder_ReturnsNoResult_WhenNoFilePosted()
```
dougbu: Correct test name. Code explicitly confirms that `NoResult` is _not_ returned.

dougbu: Seems like a general problem w/ these test names.

dougbu:
To confirm: The underlying code has an invariant that `IFormFile` implementations will never include double quotes in the `FileName` value?

If yes, :shipit: w/ corrected test names.

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/FormFileModelBinderTest.cs
javiercn: It's a problem with the diff actually, if you run diff --patience it will show the right diff. I don't know how to make GitHub do that though. I will show you in person

PR #4035, opened 2016-02-02
Reference Microsoft.NETCore.Platforms where necessary.

cesarbs:
cc @victorhurdugaci


JunTaoLuo:
I see, this repo has not been converted to building using dotnet yet so all the imports are required. Ignore my previous comment.

PR #4036, opened 2016-02-03
Enable invoking a `ViewComponent` without arguments inside of a controller.

NTaylorMullen:
- This was missed in the `Controller.ViewComponent` params arguments => anonymous object arguments PR [here](https://github.com/aspnet/Mvc/pull/4028).

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Controller.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Controller.cs
@@ -216,10 +216,36 @@ namespace Microsoft.AspNetCore.Mvc
         /// The view component name. Can be a view component
         /// <see cref="ViewComponents.ViewComponentDescriptor.ShortName"/> or
         /// <see cref="ViewComponents.ViewComponentDescriptor.FullName"/>.</param>
+        /// <returns>The created <see cref="ViewComponentResult"/> object for the response.</returns>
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Mvc.ViewFeatures/Controller.cs
```diff
@@ -216,6 +216,32 @@ public virtual PartialViewResult PartialView(string viewName, object model)
         /// The view component name. Can be a view component
         /// <see cref="ViewComponents.ViewComponentDescriptor.ShortName"/> or
         /// <see cref="ViewComponents.ViewComponentDescriptor.FullName"/>.</param>
+        /// <returns>The created <see cref="ViewComponentResult"/> object for the response.</returns>
+        [NonAction]
+        public virtual ViewComponentResult ViewComponent(string componentName)
```
NTaylorMullen: Instead of these extra methods we could also have default params in the other methods but we generally don't do that. Let me know if you feel differently.

pranavkm: Think we have method overloads for a slew of other Action methods too. Let's stick with this pattern.

pranavkm:
:shipit:

Eilon:
:shipit: for the updated doc comments too.

Issue #4016, opened 2016-01-29
Too much string allocation in Model Binding


rynowak:
With a large model binding operation (~100 form fields) - MVC is allocating way too many strings for binding.

Across 3000 requests, there's about 267mb of string being allocated. 41mb of that is the parsed representation of the form data. The other 226mb is MVC's overhead.

![image](https://cloud.githubusercontent.com/assets/1430011/12690175/81dc3776-c696-11e5-9148-6d7b1bf5964e.png)

SajayAntony:
What test are you using to get measure this? Do you have this in the [performance repo](https://github.com/aspnet/Performance/tree/dev/testapp)?

rynowak:
These are here: https://github.com/rynowak/mvc-sandbox/tree/master/workloads/BigModelBinding

SajayAntony:
Nice. @sivagms we need to work with @rynowak @troydai to get this into the perf repo :)

rynowak:
In order of importance, underlying issues here:

1). `PrefixContainer` allocates a comparer and a copy of the key for each lookup. This can happen 4x N / request, where N is the number of keys. Since we're using binary search, we likely don't need the allocation, we could iterate from the point of the first binary search. See comment here: https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/Internal/PrefixContainer.cs

Do we even need `PrefixContainer`? We could also likely do much better than binary search by using a 'real' prefix tree here if we care about lookup speed.

2). We generate lots and lots of strings for model names `ModelNames.CreatePropertyModelName` - should we cache these? Should we choose a different representation other than string?

3). Using jQuery Form Value Provider duplicates every key when using form data (can we remove this as a default?)

benaadams:
Hopefully halved strings created by form parsing with last commit of https://github.com/aspnet/HttpAbstractions/pull/550

rynowak:
That will definitely help, but unfortunately MVC is an order of magnitude worse than that code :scream:

benaadams:
Update, with captured data from recent changes plus post form reader+valuetask analysis 30k requests big model

String.Concat (2) allocs

![string-concat](http://aoa.blob.core.windows.net/aspnet/string-concat.png)

StringBuilder.ToString

![stringbuilder](http://aoa.blob.core.windows.net/aspnet/stringbuilder.png)

String.Concat (3) allocs

[...]

PR #4039, opened 2016-02-04
Replace comparer with a bespoke BinarySearch

rynowak:
> I felt a great disturbance in the Force, as if millions of unnecessary string allocations suddenly cried out in terror, and were suddenly silenced. I fear something performanc-ey has happened.

See #4016 - this should eliminate about 114mb of strings

image

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/Internal/PrefixContainer.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Internal/PrefixContainer.cs
@@ -265,9 +265,10 @@ namespace Microsoft.AspNetCore.Mvc.Internal
                     StringComparison.OrdinalIgnoreCase);
                 if (compare == 0)
                 {
-                    // Ok, so now we have a candiate that for which candidate.StartsWith(prefix) is
-                    // at least true. If the candidate is longer than the prefix, we need to look 
-                    // at the next character and see if it's a delimiter.
+                    Debug.Assert(candidate.StartsWith(prefix, StringComparison.OrdinalIgnoreCase));
+
+                    // Ok, so now we have a candiate that starts with the prefix. If the candidate is longer than
+                    // the prefix, we need to look at the next character and see if it's a delimiter.
                     if (candidate.Length == prefix.Length)
                     {
                         // Exact match
--- a/src/Microsoft.AspNetCore.Mvc.Core/Internal/PrefixContainer.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Internal/PrefixContainer.cs
@@ -41,19 +41,7 @@ namespace Microsoft.AspNetCore.Mvc.Internal
                 return _sortedValues.Length > 0; // only match empty string when we have some value
             }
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Mvc.Core/Internal/PrefixContainer.cs
```diff
@@ -259,25 +247,62 @@ private static int IndexOfDelimiter(string entry, int startIndex)
...
-            }
+            var start = 0;
+            var end = _sortedValues.Length - 1;
 
-            public int Compare(string x, string y)
+            while (start <= end)
             {
-                var testString = object.ReferenceEquals(x, _prefix) ? y : x;
-                if (IsPrefixMatch(_prefix, testString))
+                var pivot = start + (end - start >> 1);
```
rynowak: FYI - I did this for the #EilonPoints

src/Microsoft.AspNetCore.Mvc.Core/Internal/PrefixContainer.cs
```diff
@@ -259,25 +247,62 @@ private static int IndexOfDelimiter(string entry, int startIndex)
...
+                var candidate = _sortedValues[pivot];
+                var compare = string.Compare(
```
ryanbrandenburg: Why not "string.StartsWith"? Is it a perf thing or am I misunderstanding what this line does?

src/Microsoft.AspNetCore.Mvc.Core/Internal/PrefixContainer.cs
```diff
@@ -259,25 +247,62 @@ private static int IndexOfDelimiter(string entry, int startIndex)
...
+                    prefix,
+                    0,
+                    candidate,
+                    0,
+                    prefix.Length,
+                    StringComparison.OrdinalIgnoreCase);
+                if (compare == 0)
                 {
-                    return 0;
+                    // Ok, so now we have a candiate that for which candidate.StartsWith(prefix) is
```
ryanbrandenburg: If we stick with string.compare above we should change this to reference the function it actually uses.

rynowak: This isn't a reference to the other code, it's explaining that if you wrote that code it would be true. I'll make it a debug.assert instead.

src/Microsoft.AspNetCore.Mvc.Core/Internal/PrefixContainer.cs
rynowak: No. We don't just want to know if they are equal or not, we need to also know the ordering.

ryanbrandenburg: Ah, understood.

ryanbrandenburg:
Are there existing unit tests for this? If not we should probably add some.

rynowak:
@updated

ryanbrandenburg:
:shipit: once the tests reflect reality.

PR #4041, opened 2016-02-04
Re-enable few tests on Mono

ajaybhargavb:
@dougbu

### DIFF
--- a/test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ClientValidatorCacheTest.cs
+++ b/test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ClientValidatorCacheTest.cs
@@ -2,18 +2,17 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.ComponentModel.DataAnnotations;
+using System.Linq;
 using Microsoft.AspNetCore.Mvc.DataAnnotations.Internal;
 using Microsoft.AspNetCore.Mvc.ModelBinding;
 using Microsoft.AspNetCore.Mvc.ModelBinding.Validation;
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ClientValidatorCacheTest.cs
```diff
@@ -66,10 +62,7 @@ public void GetValidators_DoesNotCacheValidatorsWithIsReusableFalse()
             // Assert - 2
             Assert.NotSame(validators1, validators2);
 
-            Assert.Collection(
-                validators2,
-                v => Assert.Same(validators1[0], v), // Cached
-                v => Assert.NotSame(validators1[1], v)); // Not cached
+            Assert.Single(validators2.TakeWhile(v => validators1.Contains(v))); // only one validator is cached
```
dougbu: Might be easier to understand as
``` c#
    Assert.Contains(attribute1, validators2); // Cached
    Assert.NotContains(attribute2, validators2); // Not cached
```

Same for `GetValidators_DoesNotCacheValidatorsWithIsReusableFalse()` in the next file.

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ClientValidatorCacheTest.cs
```diff
@@ -2,18 +2,17 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.ComponentModel.DataAnnotations;
+using System.Linq;
```
dougbu: Is this just for the `TakeWhile()`?

dougbu:
:shipit:

PR #4044, opened 2016-02-04
[Design]Split OutputFormatter into OutputFormatter and TextOutputFormatter

javiercn:
WIP


src/Microsoft.AspNetCore.Mvc.Core/Formatters/TextOutputFormatter.cs
```diff
@@ -0,0 +1,246 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc.Core;
+using Microsoft.Extensions.Primitives;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNetCore.Mvc.Formatters
+{
+    /// <summary>
+    /// Writes an object to the output stream.
+    /// </summary>
```
rynowak: Polish this doc comment to say something about encodings

src/Microsoft.AspNetCore.Mvc.Core/Formatters/TextOutputFormatter.cs
```diff
@@ -0,0 +1,246 @@
...
+    public abstract class TextOutputFormatter : OutputFormatter
+    {
+        private IDictionary<string, string> _outputMediaTypeCache;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="OutputFormatter"/> class.
```
rynowak: merp

src/Microsoft.AspNetCore.Mvc.Core/Internal/MvcCoreLoggerExtensions.cs
```diff
@@ -154,7 +154,7 @@ static MvcCoreLoggerExtensions()
             _formatterSelected = LoggerMessage.Define<IOutputFormatter, string>(
                 LogLevel.Debug,
                 2,
-                "Selected output formatter '{OutputFormatter}' and content type '{ContentType}' to write the response.");
+                "Selected output formatter '{TextOutputFormatter}' and content type '{ContentType}' to write the response.");
```
rynowak: merp

src/Microsoft.AspNetCore.Mvc.Formatters.Json/JsonOutputFormatter.cs
```diff
@@ -132,16 +132,9 @@ protected virtual JsonSerializer CreateJsonSerializer()
             return _serializer;
         }
 
-        public override async Task WriteResponseBodyAsync(OutputFormatterWriteContext context)
+        public override async Task WriteResponseBodyAsync(OutputFormatterWriteContext context, Encoding selectedEncoding)
         {
-            if (context == null)
-            {
-                throw new ArgumentNullException(nameof(context));
-            }
```
rynowak: keep the null checks yo

rynowak:
:watch: - design looks right. Bring on the polish

src/Microsoft.AspNetCore.Mvc.Core/Formatters/TextOutputFormatter.cs
```diff
@@ -0,0 +1,246 @@
...
+            }
+
+            var selectedEncoding = SelectCharacterEncoding(context);
+            if (selectedEncoding != null)
+            {
+                // Override the content type value even if one already existed.
+                var mediaTypeWithCharset = GetMediaTypeWithCharset(selectedMediaType.Value, selectedEncoding);
+                selectedMediaType = new StringSegment(mediaTypeWithCharset);
+            }
+
```
javiercn: we should throw here if we can't find an encoding, it's clearly an application configuration error from the user

javiercn: I've implemented the check for SupportedEncodings > 0 inside SelectCharacterEncoding. By default we will never return null from that method as we will return the first character encoding if we can't find an encoding to use from the accept-charset header or from the content type header of the response.

If a user wants to change this behavior, they can override SelectCharacterEncoding, we will continue to work properly and they should take into account null encoding values inside their WriteBodyAsync(OutputFormatterContext context, Encoding encoding) implementation

javiercn:
@rynowak updated

src/Microsoft.AspNetCore.Mvc.Core/Formatters/TextOutputFormatter.cs
```diff
@@ -0,0 +1,258 @@
...
+        /// <inheritdoc />
+        public override async Task WriteResponseBodyAsync(OutputFormatterWriteContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            var encoding = MediaType.GetEncoding(context.ContentType);
+            await WriteResponseBodyAsync(context, encoding);
```
rynowak: so basically this won't ever get called right? Why not seal it and throw an exception

src/Microsoft.AspNetCore.Mvc.Core/Formatters/TextOutputFormatter.cs
rynowak: If the selected encoding is null it needs to return an error code.

rynowak:
:watch:

javiercn:
@rynowak Updated

rynowak:
Please see commits for minor comments, and then :shipit:

PR #4053, opened 2016-02-06
[Fixes #4013][Design] Added support for areas in 'a' and 'form' tag helpers

### LINKED ISSUE #4013: Add a TagHelper for areas in links
When i'm inside a razor view generated from an action outside of an area and I want to add a link to an action within an area I need to do the following:
```<a asp-action="Other" asp-route-area="Bug">taghelper</a>```
Given that we have asp-action and asp-controller tag helpers, we should also have an asp-area taghelpers as they are a first class concept in the framework

### PR DESCRIPTION
**Pull request by javiercn**
@NTaylorMullen 

Will delete everything that I've modified under samples/ and will add tests where appropiate.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.TagHelpers/AnchorTagHelper.cs
+++ b/src/Microsoft.AspNetCore.Mvc.TagHelpers/AnchorTagHelper.cs
@@ -14,6 +14,7 @@ namespace Microsoft.AspNetCore.Mvc.TagHelpers
     /// </summary>
     [HtmlTargetElement("a", Attributes = ActionAttributeName)]
     [HtmlTargetElement("a", Attributes = ControllerAttributeName)]
+    [HtmlTargetElement("a", Attributes = AreaAttributeName)]
     [HtmlTargetElement("a", Attributes = FragmentAttributeName)]
     [HtmlTargetElement("a", Attributes = HostAttributeName)]
     [HtmlTargetElement("a", Attributes = ProtocolAttributeName)]
@@ -24,6 +25,7 @@ namespace Microsoft.AspNetCore.Mvc.TagHelpers
     {
         private const string ActionAttributeName = "asp-action";
         private const string ControllerAttributeName = "asp-controller";
+        private const string AreaAttributeName = "asp-area";
         private const string FragmentAttributeName = "asp-fragment";
         private const string HostAttributeName = "asp-host";
         private const string ProtocolAttributeName = "asp-protocol";
@@ -66,6 +68,13 @@ namespace Microsoft.AspNetCore.Mvc.TagHelpers
         /// <remarks>Must be <c>null</c> if <see cref="Route"/> is non-<c>null</c>.</remarks>
         [HtmlAttributeName(ControllerAttributeName)]
         public string Controller { get; set; }
+
+        /// <summary>
+        /// The name of the area.
+        /// </summary>
+        /// <remarks>Must be <c>null</c> if <see cref="Route"/> is non-<c>null</c>.</remarks>
+        [HtmlAttributeName(AreaAttributeName)]
+        public string Area { get; set; }
 
         /// <summary>
         /// The protocol for the URL, such as &quot;http&quot; or &quot;https&quot;.
@@ -147,6 +156,7 @@ namespace Microsoft.AspNetCore.Mvc.TagHelpers
             {
                 if (Action != null ||
                     Controller != null ||
+                    Area != null ||
                     Route != null ||
                     Protocol != null ||
                     Host != null ||
@@ -159,6 +169,7 @@ namespace Microsoft.AspNetCore.Mvc.TagHelpers
                             "<a>",
                             ActionAttributeName,
                             ControllerAttributeName,
+                            AreaAttributeName,
                             RouteAttributeName,
                             ProtocolAttributeName,
                             HostAttributeName,
@@ -178,6 +189,17 @@ namespace Microsoft.AspNetCore.Mvc.TagHelpers
                     {
                         routeValues.Add(routeValue.Key, routeValue.Value);
                     }
+                }
+
+                if (Area != null)
+                {
+                    if (routeValues == null)
+                    {
+                        routeValues = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
+                    }
+
+                    // Unconditionally replace any value from asp-route-area. 
+                    routeValues["area"] = Area;
                 }
 
                 TagBuilder tagBuilder;
--- a/test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/AnchorTagHelperTest.cs
+++ b/test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/AnchorTagHelperTest.cs
@@ -187,6 +187,178 @@ namespace Microsoft.AspNetCore.Mvc.TagHelpers
             Assert.True(output.Content.IsEmpty);
         }
 
+        [Fact]
+        public async Task ProcessAsync_AddsAreaToRouteValuesAndCallsIntoActionLinkWithExpectedParameters()
+        {
+            // Arrange
+            var context = new TagHelperContext(
+                allAttributes: new TagHelperAttributeList(
+                    Enumerable.Empty<TagHelperAttribute>()),
+                items: new Dictionary<object, object>(),
+                uniqueId: "test");
+            var output = new TagHelperOutput(
+                "a",
+                attributes: new TagHelperAttributeList(),
+                getChildContentAsync: (useCachedResult, encoder) =>
+                {
+                    var tagHelperContent = new DefaultTagHelperContent();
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: samples/TagHelperSample.Web/Areas/Admin/Views/Admin/Index.cshtml [hunk 1]
```diff
@@ -0,0 +1,7 @@
+﻿@*
+    For more information on enabling MVC for empty projects, visit http://go.microsoft.com/fwlink/?LinkID=397860
+*@
+@{
+}
+
+<a asp-action="Index" asp-controller="Home" asp-area="">Go home!</a>
```
javiercn: Navigate outside an area from within an area

samples/TagHelperSample.Web/Views/Home/Index.cshtml
```diff
@@ -24,6 +24,7 @@
 <script>bar();</script>
 
 <h2>Index</h2>
+<a asp-action="Index" asp-Controller="Admin" asp-area="Admin" asp-route-area="Yolo">Go to admin!</a>
```
javiercn: asp-area always win over asp-route-area. This behavior is similar to asp-action over asp-route-action

src/Microsoft.AspNetCore.Mvc.TagHelpers/AnchorTagHelper.cs
```diff
@@ -180,6 +191,18 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
...
 
+                if (Area != null)
+                {
+                    if (routeValues == null)
+                    {
+                        routeValues = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
+                    }
+
+                    // If a user specifies asp-area and asp-route-area we want the first one to always win
+                    // so we unconditionally replace the value in the dictionary with that one.
```
dougbu: So many ones. Easy to get lost halfway through. Could just say `// Unconditionally replace any value from asp-route-area.` and be done.

samples/TagHelperSample.Web/Areas/Admin/Views/Admin/Index.cshtml
dougbu: To confirm: This generates the link we'd expect?

dougbu:
Design looks fine. Looking forward to tests.

samples/TagHelperSample.Web/Areas/Admin/Views/Admin/Index.cshtml
NTaylorMullen: @javiercn and I talked about this API and both agree it's a littttle wonky (but works). @Eilon / @dougbu can either of you think of something better for navigating outside of an area from inside one?

NTaylorMullen:
Design looks good pending answer to https://github.com/aspnet/Mvc/pull/4053/files#r52108212

samples/TagHelperSample.Web/Areas/Admin/Views/Admin/Index.cshtml
dougbu: Nope. The way Razor handles bound properties of type `string` prevents the more correct use of `null` to reference the root. That is, `asp-area=""` is unfortunately the same as `asp-area="@null"` -- except the second performs worse. And using an `object` property would be unhelpful for all the other cases.

Eilon: Thinking outside the box, there could be a different property to explicitly indicate "no area", e.g. `bool asp-use-main-area { get; set; }`. If that is `true` then we add `area = ""` to the route values. If it's `false` (or unset), use the newly added behavior here. If it's `true` and `asp-area` or asp-route-area` is set then we could consider throwing (or I'm a fan of silently ignoring because who cares).

Thoughts on this?

dougbu: > Thoughts on this?

I believe the pseudo-name of the "no area" area is `null`. If `string.Empty` is also correct, we should be fine w/ what @javiercn has done. Otherwise something like `asp-use-no-area` (name alternative) should add `["area"] = null` to the route values.

dougbu: /cc @rynowak

javiercn:
@dougbu @NTaylorMullen updated

test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/AnchorTagHelperTest.cs
```diff
@@ -187,6 +187,169 @@ public class AnchorTagHelperTest
...
+                Action = "Index",
+                Controller = "Home",
+                Area = "Admin",
+                Fragment = "hello=world",
+                Host = "contoso.com",
+                Protocol = "http",
+            };
+
+            // Act & Assert
+            await anchorTagHelper.ProcessAsync(context, output);
```
NTaylorMullen: Ehh, the ProcessAsync bit is really just an Act, same elsewhere.

NTaylorMullen:
We typically have a functional test to validate the end-to-end functionality but I don't feel that strongly about this one. Up to you.

:shipit:

test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/AnchorTagHelperTest.cs
```diff
@@ -187,6 +187,169 @@ public class AnchorTagHelperTest
...
+                });
+            output.Content.SetContent(string.Empty);
+
+            var generator = new Mock<IHtmlGenerator>();
+            var routeValues = new Dictionary<string, object> { { "area", "" } };
+
+            generator
+                .Setup(mock => mock.GenerateActionLink(
+                    It.IsAny<ViewContext>(),
+                    string.Empty,
```
dougbu: Use `string.Empty` everywhere to make it more obvious how this ties to [line 339](https://github.com/aspnet/Mvc/pull/4053/files#diff-c73d6268e22f9e83aec2f15c72329db6R339).

test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/AnchorTagHelperTest.cs
```diff
@@ -187,6 +187,169 @@ public class AnchorTagHelperTest
...
+                getChildContentAsync: (useCachedResult, encoder) =>
+                {
+                    var tagHelperContent = new DefaultTagHelperContent();
+                    tagHelperContent.SetContent("Something");
+                    return Task.FromResult<TagHelperContent>(tagHelperContent);
+                });
+            output.Content.SetContent(string.Empty);
+
+            var generator = new Mock<IHtmlGenerator>();
+            var routeValues = new Dictionary<string, object> { { "area", "" } };
```
dougbu: Don't include `area` in the dictionary. Testing that case already.

dougbu: Never mind. Just got confused as to how `routeValues` was used. Please rename this to `expectedRouteValues` in all the new tests to make it clearer for future readers.

test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/AnchorTagHelperTest.cs
dougbu: Yeah, silly me. This is the `linkText` argument.

test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/AnchorTagHelperTest.cs
dougbu: Also: Use `string.Empty` everywhere to make it more obvious how this ties to [line 339](https://github.com/aspnet/Mvc/pull/4053/files#diff-c73d6268e22f9e83aec2f15c72329db6R339).

test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/AnchorTagHelperTest.cs
```diff
@@ -187,6 +187,178 @@ public class AnchorTagHelperTest
...
+            // Assert
+            generator.Verify();
+
+            Assert.Equal("a", output.TagName);
+            Assert.Empty(output.Attributes);
+            Assert.True(output.Content.IsEmpty);
+        }
+
+        [Fact]
+        public async Task ProcessAsync_EmptyStringOnAspAreaClearsContextArea()
```
dougbu: Rename this test because there's nothing getting `Clear`ed in this test. All you're demonstrating is that `string.Empty` is passed through to the `IHtmlGenerator` correctly.

dougbu:
:watch: just for test fixes (both classes of course)

javiercn:
@dougbu updated

dougbu:
:shipit:

PR #4054, opened 2016-02-06
Make GetSupportedContentTypes on InputFormatter virtual

javiercn:
Spoke with @rynowak this was just an oversight during the previous implementation


rynowak:
:shipit:

PR #4055, opened 2016-02-06
[Fixes #4050] Throw an exception on CanRead/Write and GetSupportedContentTypes when the list of media types is empty

### LINKED ISSUE #4050: Should we throw in our formatters if the list of SupportedMediaTypes is empty?
In our input and output formatters we have a few pieces of code like 
```
SupportedMediaTypes.Count > 0 ? SupportedMediaTypes : null.
```

```
if(SupportedMediaTypes.Count > 0)...
```

Having no media types in the list after the formatter has been configured during startup is most likely a configuration error. This can happen in two scenarios:
* A user is configuring an existing formatter and somehow clears the supported media types list and forgets to add at least a media type.
* A user is implementing it's own formatter by extending one of our base classes and forgets do add media types to the list of supported media types inside the constructor or during the initialization of their app.

Should we just do this check in the places where we use it and throw an exception in that situation? The benefit is that the user gets an early error instead of a behavior that depends on the configuration of other pieces of the system. (Maybe the next formatter writes/reads the request/response) or we return something else like 406 or 415.

It also simplifies our implementations as we don't need to be checking if there is one element in the list of supported media types through the code

javiercn:
This provides an early error when the user missconfigures a formatter and simplifies our code a bit.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/Formatters/InputFormatter.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Formatters/InputFormatter.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Reflection;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Mvc.ApiExplorer;
+using Microsoft.AspNetCore.Mvc.Core;
 
 namespace Microsoft.AspNetCore.Mvc.Formatters
 {
@@ -43,6 +44,15 @@ namespace Microsoft.AspNetCore.Mvc.Formatters
         /// <inheritdoc />
         public virtual bool CanRead(InputFormatterContext context)
         {
+            if (SupportedMediaTypes.Count == 0)
+            {
+                var message = Resources.FormatFormatter_NoMediaTypes(
+                    GetType().FullName,
+                    nameof(SupportedMediaTypes));
+
+                throw new InvalidOperationException(message);
+            }
+
             if (!CanReadType(context.ModelType))
             {
                 return false;
@@ -111,6 +121,15 @@ namespace Microsoft.AspNetCore.Mvc.Formatters
         /// <inheritdoc />
         public virtual IReadOnlyList<string> GetSupportedContentTypes(string contentType, Type objectType)
         {
+            if (SupportedMediaTypes.Count == 0)
+            {
+                var message = Resources.FormatFormatter_NoMediaTypes(
+                    GetType().FullName,
+                    nameof(SupportedMediaTypes));
+
+                throw new InvalidOperationException(message);
+            }
+
             if (!CanReadType(objectType))
             {
                 return null;
@@ -119,7 +138,7 @@ namespace Microsoft.AspNetCore.Mvc.Formatters
             if (contentType == null)
             {
                 // If contentType is null, then any type we support is valid.
-                return SupportedMediaTypes.Count > 0 ? SupportedMediaTypes : null;
+                return SupportedMediaTypes;
             }
             else
             {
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNetCore.Mvc.Core/Formatters/InputFormatter.cs
```diff
@@ -169,6 +178,15 @@ public IReadOnlyList<string> GetSupportedContentTypes(string contentType, Type o
...
             }
 
+            if (SupportedMediaTypes?.Count == 0)
+            {
+                var message = Resources.FormatFormatter_NoMediaTypes(
+                    nameof(OutputFormatter),
+                    nameof(SupportedMediaTypes));
+
+                throw new InvalidOperationException(message);
+            }
```
rynowak: This should probably be above `CanReadType` - right?

src/Microsoft.AspNetCore.Mvc.Core/Formatters/InputFormatter.cs
```diff
@@ -62,6 +62,15 @@ public virtual bool CanRead(InputFormatterContext context)
...
```
rynowak: This should probably be above `CanReadType` - right?

src/Microsoft.AspNetCore.Mvc.Core/Formatters/InputFormatter.cs
```diff
@@ -169,6 +178,15 @@ public IReadOnlyList<string> GetSupportedContentTypes(string contentType, Type o
                 return null;
             }
 
+            if (SupportedMediaTypes?.Count == 0)
+            {
+                var message = Resources.FormatFormatter_NoMediaTypes(
+                    nameof(OutputFormatter),
```
rynowak: This isn't an output formatter

src/Microsoft.AspNetCore.Mvc.Core/Formatters/InputFormatter.cs
```diff
@@ -62,6 +62,15 @@ public virtual bool CanRead(InputFormatterContext context)
                 return false;
             }
 
+            if (SupportedMediaTypes?.Count == 0)
+            {
+                var message = Resources.FormatFormatter_NoMediaTypes(
+                    nameof(OutputFormatter),
```
rynowak: This isn't an output formatter

src/Microsoft.AspNetCore.Mvc.Core/Formatters/OutputFormatter.cs
```diff
@@ -166,6 +175,15 @@ public virtual bool CanWriteResult(OutputFormatterCanWriteContext context)
                 throw new ArgumentNullException(nameof(context));
             }
 
+            if (SupportedMediaTypes?.Count == 0)
+            {
+                var message = Resources.FormatFormatter_NoMediaTypes(
+                    nameof(OutputFormatter),
```
rynowak: Should we use `GetType()` here instead of hardcoding the base class name?

rynowak:
:watch: - needs tests also

javiercn:
@rynowak Updated

src/Microsoft.AspNetCore.Mvc.Core/Formatters/OutputFormatter.cs
```diff
@@ -15,18 +16,10 @@ namespace Microsoft.AspNetCore.Mvc.Formatters
...
-        protected OutputFormatter()
-        {
-            SupportedMediaTypes = new MediaTypeCollection();
-        }
-
-        /// <summary>
         /// Gets the mutable collection of media type elements supported by
         /// this <see cref="OutputFormatter"/>.
         /// </summary>
-        public MediaTypeCollection SupportedMediaTypes { get; }
```
rynowak: Solid. 5 PranavPoints awarded

src/Microsoft.AspNetCore.Mvc.Core/Formatters/OutputFormatter.cs
```diff
@@ -77,7 +79,7 @@ protected virtual bool CanWriteType(Type type)
 
                 return mediaTypes;
             }
-        }
+            }
```
rynowak: Fix indent

rynowak:
:shipit:

PR #4056, opened 2016-02-06
Less work on non-relative strings in UrlResolutionTagHelper

benaadams:
Earlier bail before scanning end to trim creating new string

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
@@ -219,23 +219,18 @@ namespace Microsoft.AspNetCore.Mvc.Razor.TagHelpers
         protected bool TryResolveUrl(string url, out string resolvedUrl)
         {
             resolvedUrl = null;
-            if (url == null)
+            var start = FindRelativeStart(url);
+            if (start == -1)
             {
                 return false;
             }
 
-            var trimmedUrl = url.Trim(ValidAttributeWhitespaceChars);
+            var trimmedUrl = CreateTrimmedString(url, start);
 
-            // Before doing more work, ensure that the URL we're looking at is app-relative.
-            if (trimmedUrl.Length >= 2 && trimmedUrl[0] == '~' && trimmedUrl[1] == '/')
-            {
-                var urlHelper = UrlHelperFactory.GetUrlHelper(ViewContext);
-                resolvedUrl = urlHelper.Content(trimmedUrl);
+            var urlHelper = UrlHelperFactory.GetUrlHelper(ViewContext);
+            resolvedUrl = urlHelper.Content(trimmedUrl);
 
-                return true;
-            }
-
-            return false;
+            return true;
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


benaadams:
Warning can't test as `Unable to locate Dependency System.Runtime >= 4.1.0-*`

src/Microsoft.AspNetCore.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
```diff
@@ -219,23 +219,48 @@ protected void ProcessUrlAttribute(string attributeName, TagHelperOutput output)
...
-                resolvedUrl = urlHelper.Content(trimmedUrl);
+                if (start > maxTestLength)
+                {
+                    return false;
+                }
 
-                return true;
+                var i = 0;
+                var ch = url[start];
+                for (; i < ValidAttributeWhitespaceChars.Length; i++)
```
pranavkm: Could we refactor this into a separate method - this piece of code is repeated 3 times over.

src/Microsoft.AspNetCore.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
```diff
@@ -250,41 +275,92 @@ protected bool TryResolveUrl(string url, out string resolvedUrl)
...
+                appRelativeUrl,
+                appRelativeUrl.Length - postTildeSlashUrlValue.Length,
+                postTildeSlashUrlValue);
+
+            return true;
+        }
+
+        private static string CreateTrimmedString(string input, int start)
+        {
+            int end = input.Length - 1;
```
pranavkm: `var`

src/Microsoft.AspNetCore.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
pranavkm: Maybe I'm not reading this right - but doesn't this always break (either in the loop or in the `if`?)

benaadams: two loops; two different breaks, snaffled it from [String.TrimHelper in coreclr](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/String.cs#L2793-L2827) added more var and changed to lazy + two phase

benaadams: Refactored

src/Microsoft.AspNetCore.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
```diff
@@ -250,41 +245,104 @@ protected bool TryResolveUrl(string url, out string resolvedUrl)
...
+        }
+
+        private static bool IsCharWhitespace(char ch)
+        {
+            var i = 0;
+            for (; i < ValidAttributeWhitespaceChars.Length; i++)
+            {
+                if (ValidAttributeWhitespaceChars[i] == ch)
+                {
+                    break;
```
pranavkm: `return true;` and you should be able to remove the outer `if` with a `return false`

src/Microsoft.AspNetCore.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
```diff
@@ -250,41 +245,104 @@ protected bool TryResolveUrl(string url, out string resolvedUrl)
...
             {
                 return false;
             }
 
-            var trimmedUrl = url.Trim(ValidAttributeWhitespaceChars);
+            var trimmedUrl = CreateTrimmedString(url, start);
+
+            var urlHelper = UrlHelperFactory.GetUrlHelper(ViewContext);
+            var appRelativeUrl = urlHelper.Content(trimmedUrl);
+            var postTildeSlashUrlValue = trimmedUrl.Substring(2);
```
pranavkm: We could avoid this substring although it makes the `EndsWith` check a lot messier.

pranavkm:
> Warning can't test as Unable to locate Dependency System.Runtime >= 4.1.0-*

It's on volatiledev. Hopefully we can have cidev sorted out today. 

:watch: for `IsCharWhitespace` changes.

src/Microsoft.AspNetCore.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
benaadams: this is very true :)

src/Microsoft.AspNetCore.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
benaadams: Yes thought about this - but wasn't as comfortable with it as it would potential knock-ons to  `IUrlHelper.GetUrlHelper`

benaadams:
`IsCharWhitespace` changed

src/Microsoft.AspNetCore.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs
benaadams: Actually it wouldn't; that was the reason I added back the trailing space trim. Must have been late :)

pranavkm:
Looks :shipit: to me. @NTaylorMullen \ @dougbu could you have a look?

NTaylorMullen:
Looks good. :shipit:

NTaylorMullen:
Thanks @benaadams! I've merged these in.

PR #4064, opened 2016-02-08
React to https://github.com/aspnet/Razor/pull/683

NTaylorMullen:
/cc @dougbu @rynowak

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor.Host/MvcRazorHost.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor.Host/MvcRazorHost.cs
@@ -112,7 +112,7 @@ namespace Microsoft.AspNetCore.Mvc.Razor
                     TagHelperContentGetContentMethodName = nameof(TagHelperContent.GetContent),
                     TagHelperOutputIsContentModifiedPropertyName = nameof(TagHelperOutput.IsContentModified),
                     TagHelperOutputContentPropertyName = nameof(TagHelperOutput.Content),
-                    TagHelperOutputGetChildContentAsyncMethodName = nameof(TagHelperExecutionContext.GetChildContentAsync)
+                    TagHelperOutputGetChildContentAsyncMethodName = nameof(TagHelperOutput.GetChildContentAsync)
                 })
             {
                 BeginContextMethodName = "BeginContext",
--- a/src/Microsoft.AspNetCore.Mvc.TagHelpers/LinkTagHelper.cs
+++ b/src/Microsoft.AspNetCore.Mvc.TagHelpers/LinkTagHelper.cs
@@ -245,9 +245,6 @@ namespace Microsoft.AspNetCore.Mvc.TagHelpers
                 return;
             }
 
-            // NOTE: Values in TagHelperOutput.Attributes may already be HTML-encoded.
-            var attributes = new TagHelperAttributeList(output.Attributes);
-
             if (AppendVersion == true)
             {
                 EnsureFileVersionProvider();
@@ -264,7 +261,7 @@ namespace Microsoft.AspNetCore.Mvc.TagHelpers
             var builder = new DefaultTagHelperContent();
             if (mode == Mode.GlobbedHref || mode == Mode.Fallback && !string.IsNullOrEmpty(HrefInclude))
             {
-                BuildGlobbedLinkTags(attributes, builder);
+                BuildGlobbedLinkTags(output.Attributes, builder);
                 if (string.IsNullOrEmpty(Href))
                 {
                     // Only HrefInclude is specified. Don't render the original tag.
@@ -306,8 +303,7 @@ namespace Microsoft.AspNetCore.Mvc.TagHelpers
                     continue;
                 }
 
-                attributes.SetAttribute(HrefAttributeName, url);
-                BuildLinkTag(attributes, builder);
+                BuildLinkTag(url, attributes, builder);
             }
         }
 
@@ -402,36 +398,54 @@ namespace Microsoft.AspNetCore.Mvc.TagHelpers
             }
         }
 
-        private void BuildLinkTag(TagHelperAttributeList attributes, TagHelperContent builder)
+        private void BuildLinkTag(string href, TagHelperAttributeList attributes, TagHelperContent builder)
         {
             builder.AppendHtml("<link ");
+
+            var addHref = true;
 
             // Perf: Avoid allocating enumerator
             for (var i = 0; i < attributes.Count; i++)
             {
                 var attribute = attributes[i];
-                var attributeValue = attribute.Value;
-                if (AppendVersion == true &&
-                    string.Equals(attribute.Name, HrefAttributeName, StringComparison.OrdinalIgnoreCase))
-                {
-                    // "href" values come from bound attributes and globbing. So anything but a non-null string is
-                    // unexpected but could happen if another helper targeting the same element does something odd.
-                    // Pass through existing value in that case.
-                    var attributeStringValue = attributeValue as string;
-                    if (attributeStringValue != null)
-                    {
-                        attributeValue = _fileVersionProvider.AddFileVersionToPath(attributeStringValue);
-                    }
-                }
 
-                builder
-                    .AppendHtml(attribute.Name)
-                    .AppendHtml("=\"")
-                    .Append(HtmlEncoder, attributeValue)
-                    .AppendHtml("\" ");
+                if (string.Equals(attribute.Name, HrefAttributeName, StringComparison.OrdinalIgnoreCase))
+                {
+                    addHref = false;
+
+                    AppendVersionedHref(attribute.Name, href, builder);
+                }
+                else
+                {
+                    AppendAttribute(attribute.Name, attribute.Value, builder);
+                }
+            }
+
+            if (addHref)
+            {
+                AppendVersionedHref(HrefAttributeName, href, builder);
             }
 
             builder.AppendHtml("/>");
+        }
+
+        private void AppendVersionedHref(string hrefName, string hrefValue, TagHelperContent builder)
+        {
+            if (AppendVersion == true)
+            {
+                hrefValue = _fileVersionProvider.AddFileVersionToPath(hrefValue);
+            }
+
+            AppendAttribute(hrefName, hrefValue, builder);
+        }
+
+        private void AppendAttribute(string key, object value, TagHelperContent builder)
+        {
+            builder
+                .AppendHtml(key)
+                .AppendHtml("=\"")
+                .Append(HtmlEncoder, value)
+                .AppendHtml("\" ");
         }
 
         private enum Mode
--- a/src/Microsoft.AspNetCore.Mvc.TagHelpers/ScriptTagHelper.cs
+++ b/src/Microsoft.AspNetCore.Mvc.TagHelpers/ScriptTagHelper.cs
@@ -208,9 +208,6 @@ namespace Microsoft.AspNetCore.Mvc.TagHelpers
                 return;
             }
 
-            // NOTE: Values in TagHelperOutput.Attributes may already be HTML-encoded.
-            var attributes = new TagHelperAttributeList(output.Attributes);
-
             if (AppendVersion == true)
             {
                 EnsureFileVersionProvider();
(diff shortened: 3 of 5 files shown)


src/Microsoft.AspNetCore.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -289,11 +286,7 @@ private void BuildFallbackBlock(TagHelperAttributeList attributes, DefaultTagHel
                        .AppendHtml("||document.write(\"");
 
                 // May have no "src" attribute in the dictionary e.g. if Src and SrcInclude were not bound.
-                if (!attributes.ContainsName(SrcAttributeName))
-                {
-                    // Need this entry to place each fallback source.
-                    attributes.Add(new TagHelperAttribute(SrcAttributeName, value: null));
-                }
+                var addSrc = !attributes.ContainsName(SrcAttributeName);
```
ryanbrandenburg: Is there a reason not to inline this at the if where it's used?

ryanbrandenburg:
Seems alright to me other than a nitpick, but I'll leave the squirreling to those who know the area better.

src/Microsoft.AspNetCore.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -402,7 +398,7 @@ private void EnsureFileVersionProvider()
             }
         }
 
-        private void BuildLinkTag(TagHelperAttributeList attributes, TagHelperContent builder)
+        private void BuildLinkTag(string newHref, TagHelperAttributeList attributes, TagHelperContent builder)
```
dougbu: Just `href`

src/Microsoft.AspNetCore.Mvc.TagHelpers/ScriptTagHelper.cs
dougbu: +1 Just need a `bool` variable above the `for (var i = 0; i < attributes.Count; i++)` and set to `true` if a `src` attribute is seen.

src/Microsoft.AspNetCore.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -410,30 +406,35 @@ private void BuildLinkTag(TagHelperAttributeList attributes, TagHelperContent bu
...
                 {
-                    // "href" values come from bound attributes and globbing. So anything but a non-null string is
-                    // unexpected but could happen if another helper targeting the same element does something odd.
-                    // Pass through existing value in that case.
-                    var attributeStringValue = attributeValue as string;
-                    if (attributeStringValue != null)
-                    {
-                        attributeValue = _fileVersionProvider.AddFileVersionToPath(attributeStringValue);
-                    }
+                    continue;
```
dougbu: Why is it :ok: to reorder the attributes in the `LinkTagHelper` but not in the `ScriptTagHelper`? Prefer choice made in the `ScriptTagHelper`.

src/Microsoft.AspNetCore.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -337,6 +325,20 @@ private void BuildFallbackBlock(TagHelperAttributeList attributes, DefaultTagHel
             }
         }
 
+        private void AppendEncodedVersionedSrc(string name, string srcValue, DefaultTagHelperContent builder)
```
dougbu: Just `src`

src/Microsoft.AspNetCore.Mvc.Razor.Host/MvcRazorHost.cs
```diff
@@ -112,7 +112,7 @@ internal MvcRazorHost(IChunkTreeCache chunkTreeCache, RazorPathNormalizer pathNo
                     TagHelperContentGetContentMethodName = nameof(TagHelperContent.GetContent),
                     TagHelperOutputIsContentModifiedPropertyName = nameof(TagHelperOutput.IsContentModified),
                     TagHelperOutputContentPropertyName = nameof(TagHelperOutput.Content),
-                    TagHelperOutputGetChildContentAsyncMethodName = nameof(TagHelperExecutionContext.GetChildContentAsync)
+                    TagHelperOutputGetChildContentAsyncMethodName = nameof(TagHelperOutput.GetChildContentAsync)
```
dougbu: Nice catch

dougbu:
:watch:

src/Microsoft.AspNetCore.Mvc.TagHelpers/LinkTagHelper.cs
NTaylorMullen: Fixed

NTaylorMullen:
:up: :date:

src/Microsoft.AspNetCore.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -289,12 +286,6 @@ private void BuildFallbackBlock(TagHelperAttributeList attributes, DefaultTagHel
                        .AppendHtml("||document.write(\"");
 
                 // May have no "src" attribute in the dictionary e.g. if Src and SrcInclude were not bound.
```
dougbu: Move this comment down to right above `if  (addSrc)` line.

src/Microsoft.AspNetCore.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -306,8 +303,7 @@ private void BuildGlobbedLinkTags(TagHelperAttributeList attributes, TagHelperCo
                     continue;
                 }
 
-                attributes.SetAttribute(HrefAttributeName, url);
```
dougbu: Does the `<link>` tag helper also have a case where the `href` attribute may not be bound? If so, the previous `SetAttribute(HrefAttributeName, url)` would have done the Right Thing:tm: and add the attribute if it wasn't already present. But new code in `BuildLinkTag()` expects `attributes` will _always_ contain `href`.

Bottom line: May need an `addHref` and handling similar to that in the `<script>` tag helper. Unless of course the case is impossible here.

dougbu:
:watch: for https://github.com/aspnet/Mvc/pull/4064#discussion-diff-52395144

src/Microsoft.AspNetCore.Mvc.TagHelpers/LinkTagHelper.cs
NTaylorMullen: :+1:

NTaylorMullen:
:up: :date:

src/Microsoft.AspNetCore.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -402,38 +398,60 @@ private void EnsureFileVersionProvider()
...
 
+        private void AppendEncodedVersionedHref(string hrefName, string hrefValue, TagHelperContent builder)
+        {
+            if (AppendVersion == true)
+            {
+                hrefValue = _fileVersionProvider.AddFileVersionToPath(hrefValue);
+            }
+
+            // attribute.Key ("src") does not need to be JavaScript-encoded.
+            var encodedValue = JavaScriptEncoder.Encode(hrefValue);
```
dougbu: Remove this line. The `<link>` element's `href` attribute value never needs to be JavaScript encoded.

dougbu:
:shipit: but _definitely_ with the fix mentioned [here](https://github.com/aspnet/Mvc/pull/4064#discussion-diff-52408090). `<link>` tag helper broken in this PR without that. (Surprisingly the `<link>` and `<script>` tag helpers do have a few differences :smiling_imp:)

src/Microsoft.AspNetCore.Mvc.TagHelpers/LinkTagHelper.cs
NTaylorMullen: :+1:.

src/Microsoft.AspNetCore.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -402,38 +398,57 @@ private void EnsureFileVersionProvider()
...
+            // May have no "href" attribute in the dictionary e.g. if Href and HrefInclude were not bound.
+            if (addHref)
+            {
+                AppendEncodedVersionedHref(HrefAttributeName, href, builder);
             }
 
             builder.AppendHtml("/>");
         }
 
+        private void AppendEncodedVersionedHref(string hrefName, string hrefValue, TagHelperContent builder)
```
dougbu: No longer specially-`Encoded`. Suggest removing that from the name.

src/Microsoft.AspNetCore.Mvc.TagHelpers/LinkTagHelper.cs
```diff
@@ -402,38 +398,57 @@ private void EnsureFileVersionProvider()
...
+                    AppendAttribute(attribute.Name, attribute.Value, builder);
                 }
+            }
 
-                builder
-                    .AppendHtml(attribute.Name)
-                    .AppendHtml("=\"")
-                    .Append(HtmlEncoder, attributeValue)
-                    .AppendHtml("\" ");
+            // May have no "href" attribute in the dictionary e.g. if Href and HrefInclude were not bound.
```
dougbu: What does the state of the dictionary have to do with `asp-href-include` and the `HrefInclude` property? That attribute contributes possible values for the `href`, not something in the dictionary. Unless I'm missing something, this should just be `if Href was not bound`.

src/Microsoft.AspNetCore.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -337,6 +325,25 @@ private void BuildFallbackBlock(TagHelperAttributeList attributes, DefaultTagHel
...
+            string srcValue,
+            TagHelperContent builder,
+            bool javaScriptSrc = true)
+        {
+            if (AppendVersion == true)
+            {
+                srcValue = _fileVersionProvider.AddFileVersionToPath(srcValue);
+            }
+
+            if (javaScriptSrc)
```
dougbu: This should not be conditional. Appending the version could add characters that must be encoded.

src/Microsoft.AspNetCore.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -316,20 +307,17 @@ private void BuildFallbackBlock(TagHelperAttributeList attributes, DefaultTagHel
...
-                            // attribute.Key ("src") does not need to be JavaScript-encoded.
-                            var encodedValue = JavaScriptEncoder.Encode(attributeValue);
-
-                            AppendAttribute(builder, attribute.Name, encodedValue, escapeQuotes: true);
+                            addSrc = false;
+                            AppendEncodedVersionedSrc(attribute.Name, src, builder);
                         }
                     }
 
+                    // May have no "src" attribute in the dictionary e.g. if Src and SrcInclude were not bound.
```
dougbu: Again, end with `if Src was not bound.`

src/Microsoft.AspNetCore.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -337,6 +325,25 @@ private void BuildFallbackBlock(TagHelperAttributeList attributes, DefaultTagHel
             }
         }
 
+        private void AppendEncodedVersionedSrc(
+            string srcName,
+            string srcValue,
+            TagHelperContent builder,
+            bool javaScriptSrc = true)
```
dougbu: Give this a better name e.g. `generateForDocumentWrite` and remove default value.

src/Microsoft.AspNetCore.Mvc.TagHelpers/LinkTagHelper.cs
NTaylorMullen: Legacyy comment. Removed.

src/Microsoft.AspNetCore.Mvc.TagHelpers/ScriptTagHelper.cs
NTaylorMullen: Talked off line.

dougbu: Never mind

dougbu:
:shipit: (still) though naming is hard.

PR #4066, opened 2016-02-09
Optimize PrefixContainer

rynowak:
There's really no need for us to sanitize null. This code handles null
correctly. Additionally, CopyTo should be able to do better than our
hand-written foreach in the worst case, and avoids an enumerator
allocation.

Removed a custom implementation of IndexOfAny just as cleanup.

This is sampling data from the BigModelBinding benchmark (100 form fields). Creating each `PrefixContainer` happens once per value-provider per request. Shown here are the "hot" methods in user code.

**Before**
image

Note that there are two code paths for `get_PrefixContainer` for 6.23% of the request's execution time.

**After**
image

Down to 2.64%. Notice that the call showing up is `new PrefixContainer(...)` instead of `get_PrefixContainer` - so the ctor is no longer inlined. This seems fine since the ctor call isn't too hot.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/Internal/PrefixContainer.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Internal/PrefixContainer.cs
@@ -14,6 +14,9 @@ namespace Microsoft.AspNetCore.Mvc.Internal
     /// </summary>
     public class PrefixContainer
     {
+        private static readonly string[] EmptyArray = new string[0];
+        private static readonly char[] Delimiters = new char[] { '[', '.' };
+
         private readonly ICollection<string> _originalValues;
         private readonly string[] _sortedValues;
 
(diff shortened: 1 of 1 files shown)


pranavkm:
For posterity, we dug up the history for `PrefixContainer` and the null check was essentially a guard that eventually got converted to `ToArrayWithoutNulls` (https://github.com/ASP-NET-MVC/aspnetwebstack/commit/b3638e50c6369c0ccd49ed74d84342b26c95d782?diff=split#diff-2935db0aebdda3691e350cf59e3c5162L14). Like @rynowak mentioned, it doesn't look like it's actually needed.

pranavkm:
:shipit:

src/Microsoft.AspNetCore.Mvc.Core/Internal/PrefixContainer.cs
```diff
@@ -14,6 +14,9 @@ namespace Microsoft.AspNetCore.Mvc.Internal
     /// </summary>
     public class PrefixContainer
     {
+        private static readonly string[] EmptyArray = new string[0];
```
dougbu: Yup, we definitely need at least one `EmptyArray` field per class :wink:

src/Microsoft.AspNetCore.Mvc.Core/Internal/PrefixContainer.cs
```diff
@@ -25,8 +28,17 @@ public PrefixContainer(ICollection<string> values)
             }
 
             _originalValues = values;
-            _sortedValues = ToArrayWithoutNulls(_originalValues);
```
dougbu: To confirm: None of `IFormCollection`, `IQueryCollection` or `RouteValueDictionary` allow `null` keys? I had a quick look at the relevant code and can't find an implementation not backed by a real `Dictionary<K,V>`. But this method must have been added for reasons -- right...

rynowak: It doesn't matter. `string.Compare` does the right thing with null.

PR #4069, opened 2016-02-10
ModelStateDictionary StartsWithPrefix efficiencies

benaadams:
Looking at your hot functions in #4066 top one is StartsWithPrefix

Currently it first does a string equals OrdinalIgnoreCase; then if not found does a StartsWith test.

String equals will only match one entry; so doing it for every entry then doing StartsWith is unnecessary.

/cc @rynowak

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
@@ -681,6 +681,12 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
                 throw new ArgumentNullException(nameof(key));
             }
 
+            if (prefix.Length == 0)
+            {
+                // Everything is prefixed by the empty string
+                return true;
+            }
+
             if (key.Length < prefix.Length)
             {
                 return false;
@@ -688,7 +694,7 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
 
             if (!key.StartsWith(prefix, StringComparison.OrdinalIgnoreCase))
             {
-                if (key.StartsWith("[", StringComparison.OrdinalIgnoreCase))
+                if (key[0] == '[')
                 {
                     var subKey = key.Substring(key.IndexOf('.') + 1);
 
@@ -696,9 +702,9 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
                     {
                         return false;
                     }
-
-                    if (string.Equals(prefix, subKey, StringComparison.OrdinalIgnoreCase))
+                    else if (prefix.Length == subKey.Length)
                     {
+                        // prefix == subKey
                         return true;
                     }
 
@@ -711,23 +717,16 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
             }
             else if (key.Length == prefix.Length)
             {
+                // key == prefix
                 return true;
             }
 
-            // Everything is prefixed by the empty string
-            if (prefix.Length == 0)
+            var charAfterPrefix = key[prefix.Length];
+            switch (charAfterPrefix)
             {
-                return true;
-            }
-            else
-            {
-                var charAfterPrefix = key[prefix.Length];
-                switch (charAfterPrefix)
-                {
-                    case '[':
-                    case '.':
-                        return true;
-                }
+                case '[':
+                case '.':
+                    return true;
             }
 
             return false;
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
```diff
@@ -714,6 +709,10 @@ public static bool StartsWithPrefix(string prefix, string key)
                     return false;
                 }
             }
+            else if (key.Length == prefix.Length)
+            {
+                return true;
```
dougbu: This should be `return (StringComparer.OrdinalIgnoreCase.Equals(key, prefix))` (or just `string.Equals(...Ordinal.IgnoreCase)`). No?

benaadams: `key.StartsWith(prefix, StringComparison.OrdinalIgnoreCase)` has returned `true` so if their lengths are the same they are equal?

dougbu: GitHub buried that bit, sorry.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
```diff
@@ -681,12 +681,7 @@ public static bool StartsWithPrefix(string prefix, string key)
                 throw new ArgumentNullException(nameof(key));
             }
 
-            if (StringComparer.OrdinalIgnoreCase.Equals(key, prefix))
```
dougbu: Does this really show up in perf measurements?

benaadams: Looking at your hot functions in #4066 top one is StartsWithPrefix

Currently it first does a string equals OrdinalIgnoreCase; then if not found does a StartsWith test.

String equals will only match one entry; so doing it for every entry then doing StartsWith is unnecessary.

benaadams: Updated desc

bbowyersmyth:
There is another one in the middle there for `subKey`. If `subKey` starts with `prefix` then the following `if` just needs to check lengths.

`if (key.StartsWith("[", StringComparison.OrdinalIgnoreCase))` can use `StringComparison.Ordinal`.

benaadams:
@bbowyersmyth changed and tweaked

bbowyersmyth:
If getting rid of the `subKey` allocation is desirable you could use `string.Compare` with a start index instead of the `StartsWith`. It is a bit more of a heavyweight with adding pre-checks and execution though.

LGTM.

benaadams:
@bbowyersmyth something like that?

bbowyersmyth:
Looking good. Proposed a couple of changes to `string.Compare` https://github.com/dotnet/coreclr/pull/3113

rynowak:
This looks solid overall, this had been something on my todo list, thanks @benaadams + @bbowyersmyth 

I'm going to take a crack later this week at rewriting this as a sorted list or a prefix tree. The reason that `StartsWithPrefix` is so hot is that we have to do a linear scan inside `PrefixEnumerable`.  We need to make this faster **and** choose a better data structure.

rynowak:
:shipit: from me, merging.

rynowak:
6c2c777bdcf9b286c90f4f819c31e005a427678a

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
```diff
@@ -681,54 +681,51 @@ public static bool StartsWithPrefix(string prefix, string key)
...
             }
 
+            var subKeyIndex = 0;
             if (!key.StartsWith(prefix, StringComparison.OrdinalIgnoreCase))
             {
-                if (key.StartsWith("[", StringComparison.OrdinalIgnoreCase))
+                if (key[0] == '[')
                 {
-                    var subKey = key.Substring(key.IndexOf('.') + 1);
+                    subKeyIndex = key.IndexOf('.') + 1;
```
rynowak: Looking into this more with @dougbu all of this code is actually incorrect. This implies that `[0].productId` is a match for `productId`. We're going to fix the behavior of this method and change the code that relies on this behavior (`ClearValidationState`) to do the right thing in isolation

PR #4070, opened 2016-02-10
Fast-path PrefixEnumerator

benaadams:
Use reference test, also only test for equality if exact prefix match found

/cc @rynowak

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs
@@ -782,9 +782,9 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
         public struct PrefixEnumerator : IEnumerator<KeyValuePair<string, ModelStateEntry>>
         {
             private readonly ModelStateDictionary _dictionary;
-            private readonly string _prefix;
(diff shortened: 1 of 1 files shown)


benaadams:
Looking at your hot functions in #4066 second one is MoveNext

rynowak:
Did you measure this change? What's the impact?

This one seems a little less clear-cut since it makes the code more complicated, the struct bigger, and all of this is going to be rewritten anyway.

rynowak:
Self assigning to track

benaadams:
@rynowak I'm having build issues with grumpy feeds, was hoping aspnetcidev would be happy soon. Should I move to volatile?

rynowak:
volatile should be fine for now - I checked in a bunch of **massive** improvements this morning, so things should be much better.

benaadams:
Might need a couple runs with the profiler, but pre and post calls

pre-calls

post-calls

benaadams:
Think it can be better and simpler

benaadams:
post-post

@rynowak Is better now?

rynowak:
Yeah it does look better. Seems worth taking for now... just FYI I am going to rewrite this whole class if I have time this week.

benaadams:
Yar, is still a big player on that binding test

rynowak:
a33369acd8b072fa41ac8f6255414e70ddf70418

PR #4076, opened 2016-02-10
Fix behavior of StartsWithPrefix

rynowak:
This undoes a behavior change introduced in
7b18d1d3f1f04ba8952b13fd2b0f903e684bf78d.

The intent was to have ClearValidationState do the right thing for a case
where a collection was bound to the empty prefix, and then used again with
TryUpdateModel.

This change was implemented by saying that a key like "[0].Foo" is a match
for the prefix of "Foo". This isn't really right, and it's only
interesting for the ClearValidationState case.

The problem is that we don't know what the keys look like for a
collection. We can assume that they start with [0] but that's not really a
guarantee, it's a guess.

This change fixes the behavior of StartsWithModel, and move the
responsibility for this case back into ClearValidationState.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/Internal/PrefixContainer.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Internal/PrefixContainer.cs
@@ -173,45 +173,7 @@ namespace Microsoft.AspNetCore.Mvc.Internal
             }
         }
 
-        public static bool IsPrefixMatch(string prefix, string testString)
-        {
-            if (testString == null)
-            {
-                return false;
-            }
-
-            if (prefix.Length == 0)
-            {
-                return true; // shortcut - non-null testString matches empty prefix
-            }
-
-            if (prefix.Length > testString.Length)
-            {
-                return false; // not long enough
-            }
-
-            if (!testString.StartsWith(prefix, StringComparison.OrdinalIgnoreCase))
-            {
-                return false; // prefix doesn't match
-            }
-
-            if (testString.Length == prefix.Length)
-            {
-                return true; // exact match
-            }
-
-            // invariant: testString.Length > prefix.Length
-            switch (testString[prefix.Length])
-            {
-                case '.':
-                case '[':
-                    return true; // known delimiters
-
-                default:
-                    return false; // not known delimiter
-            }
-        }
-
+        // This is tightly coupled to the definition at ModelStateDictionary.StartsWithPrefix
         private int BinarySearch(string prefix)
         {
             var start = 0;
--- a/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBindingHelper.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBindingHelper.cs
@@ -513,10 +513,7 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
             }
 
             var modelMetadata = metadataProvider.GetMetadataForType(modelType);
-
-            // Clear ModelStateDictionary entries for the model so that it will be re-validated.
             var modelState = actionContext.ModelState;
-            ClearValidationStateForModel(modelType, modelState, metadataProvider, prefix);
 
             var operationBindingContext = new OperationBindingContext
             {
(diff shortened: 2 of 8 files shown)


src/Microsoft.AspNetCore.Mvc.Core/Internal/PrefixContainer.cs
```diff
@@ -173,45 +173,7 @@ private static void GetKeyFromNonEmptyPrefix(string prefix, string entry, IDicti
             }
         }
 
-        public static bool IsPrefixMatch(string prefix, string testString)
```
rynowak: Deleting an almost-duplicate of this code

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBindingHelper.cs
```diff
@@ -654,28 +653,31 @@ internal static string GetPropertyName(Expression expression)
...
+                        if (kvp.Key.Length > 0 && kvp.Key[0] == '[')
+                        {
+                            // Starts with an indexer
+                            kvp.Value.Errors.Clear();
+                            kvp.Value.ValidationState = ModelValidationState.Unvalidated;
+                        }
+                    }
                 }
-
-                foreach (var property in modelMetadata.Properties)
```
rynowak: Notice that the existing code isn't really correct either. It only handles collections of POCOs correctly, not collections of simple types.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBindingHelper.cs
```diff
@@ -636,17 +636,16 @@ internal static string GetPropertyName(Expression expression)
         /// Clears <see cref="ModelStateDictionary"/> entries for <see cref="ModelMetadata"/>.
         /// </summary>
         /// <param name="modelMetadata">The <see cref="ModelMetadata"/>.</param>
+        /// <param name="modelstate">The <see cref="ModelStateDictionary"/>.</param>
         /// <param name="modelKey">The entry to clear. </param>
-        /// <param name="modelMetadataProvider">The <see cref="IModelMetadataProvider"/>.</param>
         public static void ClearValidationStateForModel(
```
rynowak: I wonder if we'd be better off doing something really simple here - like clearing the whole dictionary if you give us the empty prefix. All of the stuff we can try is imprecise unless you used a prefix.

test/Microsoft.AspNetCore.Mvc.Abstractions.Test/ModelBinding/ModelStateDictionaryTest.cs
```diff
@@ -505,9 +503,9 @@ public void GetFieldValidity_ReturnsUnvalidated_IfAnyItemInSubtreeIsInvalid()
         }
 
         [Theory]
+        [InlineData("")]
         [InlineData("user")]
         [InlineData("user.Age")]
-        [InlineData("product")]
         public void GetFieldValidity_ReturnsInvalid_IfAllKeysAreValidatedAndAnyEntryIsInvalid(string key)
```
rynowak: This was really 2 tests.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBindingHelper.cs
dougbu: There are really only three cases w/ an empty prefix:
- A simple type: Just remove the `""` entry, if any.
- A collection: Remove anything w/ a key that starts w/ `'['`.
- A POCO: Remove anything that's a prefix match for a property in the type.

That's fairly precise.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBindingHelper.cs
```diff
@@ -636,17 +636,16 @@ internal static string GetPropertyName(Expression expression)
         /// Clears <see cref="ModelStateDictionary"/> entries for <see cref="ModelMetadata"/>.
         /// </summary>
         /// <param name="modelMetadata">The <see cref="ModelMetadata"/>.</param>
+        /// <param name="modelstate">The <see cref="ModelStateDictionary"/>.</param>
         /// <param name="modelKey">The entry to clear. </param>
-        /// <param name="modelMetadataProvider">The <see cref="IModelMetadataProvider"/>.</param>
```
dougbu: Why change the `public` surface? Suggest the old method could exist as another overload, calling this one.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBindingHelper.cs
```diff
@@ -654,28 +653,31 @@ internal static string GetPropertyName(Expression expression)
...
             // If modelkey is empty, we need to iterate through properties (obtained from ModelMetadata) and
             // clear validation state for all entries in ModelStateDictionary that start with each property name.
             // If modelkey is non-empty, clear validation state for all entries in ModelStateDictionary
             // that start with modelKey
             if (string.IsNullOrEmpty(modelKey))
             {
-                var modelMetadata = metadataProvider.GetMetadataForType(modelType);
-                var elementMetadata = modelMetadata.ElementMetadata;
-                if (elementMetadata != null)
+                if (modelMetadata.IsCollectionType)
```
dougbu: Use `IsEnumerableType`. `ElementMetadata` is non-`null` for all `IEnumerable` types.

src/Microsoft.AspNetCore.Mvc.Core/Internal/PrefixContainer.cs
dougbu: Minor but this version looks slightly cleaner (except for the `switch` statement). Suggest copying comments like `// not long enough` and `// prefix doesn't match`. Also good to make the `StartsWith()` part `return` early.

dougbu:
:watch:

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBindingHelper.cs
rynowak: Sure

src/Microsoft.AspNetCore.Mvc.Core/Internal/PrefixContainer.cs
rynowak: A lot of these comments just document the obvious.

rynowak:
@dougbu updated

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBindingHelper.cs
```diff
@@ -659,27 +659,76 @@ internal static string GetPropertyName(Expression expression)
...
+                        entry.Errors.Clear();
+                        entry.ValidationState = ModelValidationState.Unvalidated;
+                    }
                 }
             }
             else
             {
+                // If model key is non-empty, we just want to clear all keys with that prefix. We expect
+                // model binding to have only used this key (and suffixes) for all entries related to
+                // this model.
```
dougbu: Liking these comments. Now we might remember what we decided a few months from now -- and why :clap:

src/Microsoft.AspNetCore.Mvc.Core/Internal/PrefixContainer.cs
dougbu: Thx for the copying you did.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBindingHelper.cs
dougbu: > Notice that the existing code isn't really correct ~~either~~

New code looks predictable and certainly correct enough. Problems should only occur when multiple collections or multiple complex objects bind using the empty prefix.

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/ModelBindingHelperTest.cs
```diff
@@ -764,10 +800,11 @@ public void ClearValidationStateForCollectionsModel_EmtpyModelKey(string modelKe
         [InlineData("product.Order[0].Address.Street")]
         [InlineData("product.Category.Name")]
         [InlineData("product.Order")]
-        public void ClearValidationStateForModel_NonEmtpyModelKey(string prefix)
+        public void ClearValidationState_ForModel_NonEmtpyModelKey(string prefix)
```
dougbu: `ForModel` -> `ForComplexTypeModel` ?

dougbu:
Thought this PR was going to remove the `ModelBindingHelper.ClearValidationStateForModel()` calls from `ControllerBase.TryValidateModel()` and `ModelBindingHelper.TryUpdateModelAsync()`?

dougbu:
:shipit: if another PR is coming for removing our use of `ClearValidationStateForModel()`;
:watch: otherwise.

rynowak:
Yeah, you're right I forgot about that. Will get back to this on Monday.

Issue #4078, opened 2016-02-11
TagBuilder can't build valueless attributes


geertdoornbos:
The Tagbuilder can't build a tag that had attributes with no value, like:

`<script src="..." async></script>`

It is currently possible to build alternatives:
`<script src="..." async=""></script>`
or
`<script src="..." async="async"></script>`

All three ways should be possible, where the first is the most clean way.

dougbu:
Suspect you'd be better served using `TagHelperOutput`. On one hand `TagBuilder` is intended for HTML generation from code and we long-ago chose to emit only XHTML syntax. On the other, `TagHelperOutput` is intended to maintain the original syntax (from a `.cshtml` file) when filtering that HTML through a tag helper.

In any case `TagBuilder` does not support minimized attributes, sorts the generated attributes alphabetically by name, and more.

PR #4079, opened 2016-02-11
Update TagBuilder to support valueless attributes.

geertdoornbos:
Ref. #4078 

This change makes it possible to add valueless attributes. For instance, the async attribute: <script src="..." async>

Attributes can now be rendered in the three ways, specified by w3c: http://www.w3.org/TR/html5/infrastructure.html#boolean-attribute

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Rendering/TagBuilder.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Rendering/TagBuilder.cs
@@ -204,9 +204,12 @@ namespace Microsoft.AspNetCore.Mvc.Rendering
 
                     writer.Write(" ");
                     writer.Write(key);
-                    writer.Write("=\"");
-                    encoder.Encode(writer, attribute.Value);
-                    writer.Write("\"");
+                    if (attribute.Value != null)
+                    {
+                        writer.Write("=\"");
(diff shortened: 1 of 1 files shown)


rynowak:
This change looks good, needs a unit test though. Similar to https://github.com/aspnet/Mvc/blob/dev/test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/TagBuilderTest.cs#L117

dougbu:
If we did this (and something was proposed offline ages ago and shot down), we can't do it this way. `null` is a valid attribute value and it results in `name=""` which is semantically different from `name` or `name="name"`. So this PR would result in a breaking change.

geertdoornbos:
Don't want a breaking change of course. Thanks for the quick feedback.
I didn't know TagHelper is bound to XHTML syntax. 
Ah semantics, if someone passes a null value for a string type attribute and generates code; the parser could see it as a boolean type attribute.

But it's 2016 now :smile: time to leave TagHelper behind I guess, if it can't support html5 style.
Thanks!

benaadams:
Switch on if `<!DOCTYPE html>` is at top of page?

geertdoornbos:
@benaadams I don't think there is a good way to do that inside TagBuilder. Altering TagBuilder to support html5 style is also not a good plan. Thanks anyway.

PR #4090, opened 2016-02-12
[Fixes #3774] DataAnnotation validation ignored

### LINKED ISSUE #3774: DataAnnotation validation ignored
DataAnnotation validation is ignored if route parameter has the same name as object variable passed in the body of the HTTP request

Model:

```c#
public class TransferInfo
{
        [Range(1, int.MaxValue)] //or CustomValidationAttribute
        public int AccountId { get; set; }
        public double Amount { get; set; }
}
```

Controller:

```c#
[HttpPost("{accountId:int}")]
public IActionResult Transfer(int accountId, [FromBody] TransferInfo transferInfo)
{
    if (!ModelState.IsValid)
    {
        return HttpBadRequest();
    }
    ...
}

```

Posted data (JSON):

```json
{
     "accountId": 0,
     "amount": 250.0
}
```

ModelState is valid even if I check that accountId must be greater than 0, using Range or Custom validation attribute.

If I update my TransferInfo class definition and rename "AccountId" by "ToAccountId" and try to send:

```json
{
     "toAccountId: 0,
     "amount": 250.0
}
```

ModelState Validation will failed

kichalla:
@rynowak

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Validation/ValidationVisitor.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Validation/ValidationVisitor.cs
@@ -98,7 +98,7 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding.Validation
         protected virtual bool ValidateNode()
         {
             var state = _modelState.GetValidationState(_key);
-            if (state == ModelValidationState.Unvalidated)
+            if (state != ModelValidationState.Invalid)
             {
                 var validators = _validatorCache.GetValidators(_metadata, _validatorProvider);
 
--- a/test/Microsoft.AspNetCore.Mvc.IntegrationTests/ValidationIntegrationTests.cs
+++ b/test/Microsoft.AspNetCore.Mvc.IntegrationTests/ValidationIntegrationTests.cs
@@ -117,6 +117,9 @@ namespace Microsoft.AspNetCore.Mvc.IntegrationTests
         [MemberData(nameof(MultipleActionParametersAndValidationData))]
         public async Task MultipleActionParameter_ValidModelState(List<ParameterDescriptor> parameters)
         {
+            // Since validation attribute is only present on the FromBody model's property(TransferInfo's AccountId),
+            // validation should not trigger for the parameter which is bound from Uri.
+
             // Arrange
             var actionDescriptor = new ControllerActionDescriptor()
             {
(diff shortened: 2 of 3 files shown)


kichalla:
@dougbu

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Validation/ValidationVisitor.cs
```diff
@@ -98,7 +98,7 @@ public bool Validate(ModelMetadata metadata, string key, object model)
         protected virtual bool ValidateNode()
         {
             var state = _modelState.GetValidationState(_key);
-            if (state == ModelValidationState.Unvalidated)
+            if (state != ModelValidationState.Invalid)
```
rynowak: Let's add a comment here. Something like

```
// Rationale: we might see the same model state key used for two different objects.
// We want to run validation unless it's already known that this key is invalid.
```

rynowak:
:shipit: with a comment from me. Get signoff from @dougbu as well

test/Microsoft.AspNetCore.Mvc.IntegrationTests/ValidationIntegrationTests.cs
```diff
@@ -18,6 +20,133 @@ namespace Microsoft.AspNetCore.Mvc.IntegrationTests
...
+        private class TestController { }
+
+        public static TheoryData<List<ParameterDescriptor>> MultipleActionParametersAndValidationData
+        {
+            get
+            {
+                return new TheoryData<List<ParameterDescriptor>>
+                {
+                    // Irrespective of the order in which the parameters are defined on the action,
+                    // the validation on the TransferInfo's AccountId should occur.
```
dougbu: May be worth mentioning the `accountId` is bound using the prefix `"accountId"` while `transferInfo` is bound using the empty prefix (and the `TransferInfo` property names).

test/Microsoft.AspNetCore.Mvc.IntegrationTests/ValidationIntegrationTests.cs
```diff
@@ -18,6 +20,133 @@ namespace Microsoft.AspNetCore.Mvc.IntegrationTests
...
+                actionDescriptor: actionDescriptor);
+
+            var modelState = operationContext.ActionContext.ModelState;
+
+            // Act
+            var arguments = await argumentBinder.BindActionArgumentsAsync(
+                (ControllerContext)operationContext.ActionContext, new TestController());
+
+            // Assert
+            Assert.True(modelState.IsValid);
```
dougbu: Confirm the values in the two arguments.

dougbu:
:shipit: w/ slight test updates.

kichalla:
7337f501120cfcd9fb46166b38c360a86282b879

PR #4095, opened 2016-02-15
[Fixes #4085] removing redundant 'http' prefixes from helper methods.

cottsak:
Thanks for the discussion on https://github.com/aspnet/Mvc/issues/4085

Can this be merged @danroth27 with the build in the state it is?

PR #4107, opened 2016-02-16
[Fixes #4085] Controller helper inconsistency: HttpNotFound(), Ok()

### LINKED ISSUE #4085: Controller helper inconsistency: HttpNotFound(), Ok()
I can `return HttpNotFound()` or I can `return Ok()` from a Controller. Why the inconsistent naming?

It should be one of these two alternatives IMO:

*  `return HttpNotFound()` and `return HttpOk()`, or
*  `return NotFound()` and `return Ok()`

I personally prefer the latter like the previous MVC incarnations. Thoughts folks?
cc @JakeGinnivan

kichalla:
@Eilon

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/StatusCodeResult.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/StatusCodeResult.cs
@@ -12,14 +12,14 @@ namespace Microsoft.AspNetCore.Mvc
     /// Represents an <see cref="ActionResult"/> that when executed will
     /// produce an HTTP response with the given response status code.
     /// </summary>
-    public class HttpStatusCodeResult : ActionResult
+    public class StatusCodeResult : ActionResult
     {
         /// <summary>
-        /// Initializes a new instance of the <see cref="HttpStatusCodeResult"/> class
+        /// Initializes a new instance of the <see cref="StatusCodeResult"/> class
         /// with the given <paramref name="statusCode"/>.
         /// </summary>
         /// <param name="statusCode">The HTTP status code of the response.</param>
-        public HttpStatusCodeResult(int statusCode)
+        public StatusCodeResult(int statusCode)
         {
             StatusCode = statusCode;
         }
@@ -38,7 +38,7 @@ namespace Microsoft.AspNetCore.Mvc
             }
 
             var factory = context.HttpContext.RequestServices.GetRequiredService<ILoggerFactory>();
-            var logger = factory.CreateLogger<HttpStatusCodeResult>();
+            var logger = factory.CreateLogger<StatusCodeResult>();
 
             logger.HttpStatusCodeResultExecuting(StatusCode);
 
--- a/test/Microsoft.AspNetCore.Mvc.Core.Test/ControllerBaseTest.cs
+++ b/test/Microsoft.AspNetCore.Mvc.Core.Test/ControllerBaseTest.cs
@@ -765,7 +765,7 @@ namespace Microsoft.AspNetCore.Mvc.Core.Test
             var controller = new TestableController();
 
             // Act
-            var result = controller.HttpUnauthorized();
+            var result = controller.Unauthorized();
 
             // Assert
             Assert.IsType<HttpUnauthorizedResult>(result);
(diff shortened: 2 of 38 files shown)


javiercn:
@kichalla A lot of these ones were inconsistent to maintain compatibility with MVC, at the time I remember that the decision was to keep compatibility vs consistency. I personally like this change, but just wanted to make sure this was taken in consideration

rynowak:
It was.

kichalla:
@javiercn More details here: https://github.com/aspnet/Mvc/issues/4085

test/Microsoft.AspNetCore.Mvc.Core.Test/ControllerBaseTest.cs
```diff
@@ -765,7 +765,7 @@ public void HttpUnauthorized_SetsStatusCode()
             var controller = new TestableController();
 
             // Act
-            var result = controller.HttpUnauthorized();
+            var result = controller.Unauthorized();
```
dougbu: Should change all test names that mention `HttpBlah` as well.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ControllerUnitTestabilityTests.cs
```diff
@@ -688,22 +688,22 @@ public IActionResult CreatedAtRoute_Action(string routeName, object routeValues,
 
             public IActionResult HttpBadRequest_Action()
```
dougbu: Suggest correcting these action names (and the URIs in the test class) too.

dougbu:
:watch:

test/Microsoft.AspNetCore.Mvc.Core.Test/ControllerBaseTest.cs
Eilon: Not too concerned about these. I think the test names / URLs are pretty clear either way, no?

dougbu: In other renaming PRs, we've updated test names et cetera to match the new method names. But :ok: if you're fine w/ the inconsistencies.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ControllerUnitTestabilityTests.cs
dougbu: Suspect @Eilon wants you to ignore me here (see https://github.com/aspnet/Mvc/pull/4107#discussion-diff-53069339).

test/Microsoft.AspNetCore.Mvc.Core.Test/ControllerBaseTest.cs
Eilon: The goal is clarity and maintainability.

dougbu:
Was about to say: :shipit: but use #4095 for the `ControllerBase` changes.

> The goal is clarity and maintainability.

Sounds good.

But I'm confused by renaming the result classes. Didn't @Eilon say _not_ to do that?

src/Microsoft.AspNetCore.Mvc.Core/StatusCodeResult.cs
```diff
@@ -12,14 +12,14 @@ namespace Microsoft.AspNetCore.Mvc
     /// Represents an <see cref="ActionResult"/> that when executed will
     /// produce an HTTP response with the given response status code.
     /// </summary>
-    public class HttpStatusCodeResult : ActionResult
+    public class StatusCodeResult : ActionResult
```
dougbu: This one isn't like the others. @Eilon does it need to change?

kichalla: @Eilon approved on this change offline

Eilon: Yes this was discussed and spec'ed and approved.

dougbu: :+1: thx

dougbu:
:shipit: but use #4095 for the initial `ControllerBase` changes. Add the rest in a second commit.

kichalla:
fd3ee499872465ccc2c5534b58f8ad113a8c0190

PR #4109, opened 2016-02-17
Small Refactorings

calebnelton:
Was not paying attention in WebEx meeting ... this was more fun.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/BindingSource.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/BindingSource.cs
@@ -183,7 +183,7 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
         /// <inheritdoc />
         public bool Equals(BindingSource other)
         {
-            return other == null ? false : string.Equals(other.Id, Id, StringComparison.Ordinal);
+            return string.Equals(other?.Id, Id, StringComparison.Ordinal);
         }
 
         /// <inheritdoc />
--- a/src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultApplicationModelProvider.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultApplicationModelProvider.cs
@@ -592,8 +592,7 @@ namespace Microsoft.AspNetCore.Mvc.Internal
             var routeTemplateProvider =
                 attributes
                 .OfType<IRouteTemplateProvider>()
-                .Where(a => !IsSilentRouteAttribute(a))
-                .SingleOrDefault();
+                .SingleOrDefault(a => !IsSilentRouteAttribute(a));
 
             if (routeTemplateProvider != null)
             {
--- a/src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ValueProviderResult.cs
@@ -96,7 +96,7 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
         public override bool Equals(object obj)
         {
             var other = obj as ValueProviderResult?;
-            return other.HasValue ? Equals(other.Value) : false;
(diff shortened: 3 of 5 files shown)


src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultApplicationModelProvider.cs
```diff
@@ -395,10 +395,7 @@ protected virtual PropertyModel CreatePropertyModel(PropertyInfo propertyInfo)
                 // public void Foo() { }
                 //
                 // Is one action.
-                if (!(routeProvider is IActionHttpMethodProvider))
-                {
-                    createActionForSilentRouteProviders = false;
-                }
+                createActionForSilentRouteProviders &= routeProvider is IActionHttpMethodProvider;
```
Eilon: Not personally a fan of the readability or debuggability of this (it was easy to set a breakpoint in the `if()` before).

Eilon:
@calebnelton thanks for the PR. Can you squash everything into one commit?

I should note that we generally prefer avoiding such changes because they don't offer a significant improvement of the code, and it takes a fair bit of time from many people to work on reviewing, merging, and verifying the PR. Feel free to discuss with us through issues any additional changes you might have in mind.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/BindingSource.cs
```diff
@@ -183,7 +183,7 @@ public virtual bool CanAcceptDataFrom(BindingSource bindingSource)
         /// <inheritdoc />
         public bool Equals(BindingSource other)
         {
-            return other == null ? false : string.Equals(other.Id, Id, StringComparison.Ordinal);
+            return other != null && string.Equals(other.Id, Id, StringComparison.Ordinal);
```
pranavkm: `public bool Equals(BindingSource other) => string.Equals(other?.Id, Id, StringComparison.Ordinal);`

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultApplicationModelProvider.cs
```diff
@@ -592,8 +589,7 @@ protected virtual bool IsAction(TypeInfo typeInfo, MethodInfo methodInfo)
             var routeTemplateProvider =
                 attributes
                 .OfType<IRouteTemplateProvider>()
-                .Where(a => !IsSilentRouteAttribute(a))
-                .SingleOrDefault();
+                .SingleOrDefault(a => !IsSilentRouteAttribute(a));
```
pranavkm: This should probably be `FirstOrDefault`.cc  @rynowak

rynowak: That would hide a bug. It's intended that this method only be called when one or zero attributes is a non-silent route provided. TLDR all of this code is complicated on purpose.

pranavkm: Should we throw a nicer exception in that case? `SingleOrDefault`'s message isn't particularly intuitive.

pranavkm: Regardless, this change is fine as is.

pranavkm:
:watch:

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultApplicationModelProvider.cs
calebnelton: I'll remove it when I rebase and squash into a single commit.

calebnelton:
@Eilon Squashed commits

Code comment from @pranavkm incorporated also

pranavkm:
Great. I'll merge this in soon

pranavkm:
Merged in https://github.com/aspnet/Mvc/commit/5b0b0b95a7d6d40c4506abb105186042934a9b4b

pranavkm:
Thanks for the PR!

Issue #4093, opened 2016-02-14
RequireHttpsAttribute needs some devmode love


TheWyzard44:
If you could amend the Constructor

`public RequireHttpsAttribute(int devPort = 443, int stagingPort = 443, int productionPort = 443)`

 and the main body to something more like:

    #if DEBUG
                var thisPort = DevPort;
    #else
    #if STAGING
                var thisPort = StagingPort;
    #else
                var thisPort = ProductionPort;
    #endif
    #endif
                var oldport = ":";
                var newport = "";
                var request = filterContext.HttpContext.Request;
                if (request.Host.Value.Contains(':'))
                    oldport = ":" + request.Host.Value.Split(':')[1];
                if (thisPort != 443)
                    newport = ":" + thisPort.ToString();
                var newUrl = string.Concat(
                    "https://",
                    request.Host.ToUriComponent().Replace(oldport, "") + newport,
                    request.PathBase.ToUriComponent(),
                    request.Path.ToUriComponent(),
                    request.QueryString.ToUriComponent());
                // redirect to HTTPS version of page
                filterContext.Result = new RedirectResult(newUrl, permanent: true);

or something Similar, then we could:
`[RequireHttps(devPort: 44391)]`
`public class SomeController {...}`

And that would make our dev lives a little easier! ;)  I'm not crazy about this in regards to separation of concerns...  but the alternative of putting that #if block on every single controller isn't much prettier.

Thank you for your consideration

EDIT:  For now I've implemented this as a CustomRequireHttpsAttribute, and perhaps that's the best answer for those of us who woudn't mind mixing in environmental details with our code.  Maintainability is about the same with both approaches but this would lead to easier reading.  Or perhaps there's a better idea out there?  This scenario is not for redirecting an entire site to https (where we could set this up in the middleware), but instead for a large selection of controllers that actually deal in secure data, but also where the site does generate enough public traffic that saving some encryption overhead is beneficial when there's no sensitive data being passed.

At minimum the current implementation does not address custom ports.  If the request is http://something:8080 then it will redirect to https://something:8080 which won't work (2 protocols on the same port).

Staging is superfluous, in this example, but having the ability to address custom ports in some way, without the environmental details may be another solution to the dilemma.  99% of the time, I'm sure production will be 443, and staging could be manipulated.  And that leaves dev, which doesn't always have the option of binding 443, let alone a simple port 80, so local testing won't be possible otherwise.

Eilon:
[...]

PR #4113, opened 2016-02-18
[Fixes #4112 #4093] Adding support for custom SSL port

### LINKED ISSUE #4112: RequireHttpsAttribute doesn't change the port
RequireHttpsAttribute changes the scheme but not the port, so `http://localhost:44380/` gets redirected to `https://localhost:44380/`, which is invalid because you can't serve both http and https on the same port.

The only way this ever worked was if the original request's host header did not include the port (e.g. `localhost` rather than `localhost:80`)

There's even a (wrong) test:
https://github.com/aspnet/Mvc/commit/25838cee55216bd02432390996b95c5480b23e94#diff-ba666c0ff94ddf047b6f11caa0ac94c9R42

@sebastienros @blowdart @javiercn

sebastienros:
New optional MvcOptions.SslPort. If not defined the redirection uses an empty port (default),
otherwise the custom port is used.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/MvcOptions.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/MvcOptions.cs
@@ -127,5 +127,11 @@ namespace Microsoft.AspNetCore.Mvc
         /// Gets a list of <see cref="IValueProviderFactory"/> used by this application.
         /// </summary>
         public IList<IValueProviderFactory> ValueProviderFactories { get; }
+
+        /// <summary>
+        /// Gets or sets the SSL port that is used by this application when <see cref="RequireHttpsAttribute"/>
+        /// is used. If not set the port won't be specified in the secured URL e.g. https://localhost/path.
+        /// </summary>
+        public int? SslPort { get; set; }
     }
 }
--- a/src/Microsoft.AspNetCore.Mvc.Core/RequireHttpsAttribute.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/RequireHttpsAttribute.cs
@@ -4,6 +4,8 @@
 using System;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc.Filters;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.AspNetCore.Mvc
 {
@@ -35,10 +37,25 @@ namespace Microsoft.AspNetCore.Mvc
             }
             else
             {
+                var optionsAccessor = filterContext.HttpContext.RequestServices.GetRequiredService<IOptions<MvcOptions>>();
+
                 var request = filterContext.HttpContext.Request;
+
+                var host = request.Host;
+                if (optionsAccessor.Value.SslPort.HasValue && optionsAccessor.Value.SslPort > 0)
+                {
+                    // a specific SSL port is specified
+                    host = new HostString(host.Host, optionsAccessor.Value.SslPort.Value);
+                }
+                else
+                {
+                    // clear the port
+                    host = new HostString(host.Host);
+                }
+                
                 var newUrl = string.Concat(
                     "https://",
-                    request.Host.ToUriComponent(),
+                    host.ToUriComponent(),
                     request.PathBase.ToUriComponent(),
                     request.Path.ToUriComponent(),
                     request.QueryString.ToUriComponent());
--- a/test/Microsoft.AspNetCore.Mvc.Core.Test/RequireHttpsAttributeTests.cs
+++ b/test/Microsoft.AspNetCore.Mvc.Core.Test/RequireHttpsAttributeTests.cs
@@ -1,11 +1,14 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Internal;
 using Microsoft.AspNetCore.Mvc.Abstractions;
 using Microsoft.AspNetCore.Mvc.Filters;
 using Microsoft.AspNetCore.Routing;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
 using Xunit;
 
 namespace Microsoft.AspNetCore.Mvc
@@ -37,7 +40,7 @@ namespace Microsoft.AspNetCore.Mvc
                 return new TheoryData<string, string, string, string, string>
                 {
                     { "localhost", null, null, null, "https://localhost" },
-                    { "localhost:5000", null, null, null, "https://localhost:5000" },
+                    { "localhost:5000", null, null, null, "https://localhost" },
                     { "localhost", "/pathbase", null, null, "https://localhost/pathbase" },
                     { "localhost", "/pathbase", "/path", null, "https://localhost/pathbase/path" },
                     { "localhost", "/pathbase", "/path", "?foo=bar", "https://localhost/pathbase/path?foo=bar" },
@@ -67,6 +70,7 @@ namespace Microsoft.AspNetCore.Mvc
         {
             // Arrange
             var requestContext = new DefaultHttpContext();
+            requestContext.RequestServices = CreateServices();
             requestContext.Request.Scheme = "http";
             requestContext.Request.Method = "GET";
             requestContext.Request.Host = HostString.FromUriComponent(host);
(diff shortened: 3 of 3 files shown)


test/Microsoft.AspNetCore.Mvc.Core.Test/RequireHttpsAttributeTests.cs
```diff
@@ -6,6 +6,9 @@
 using Microsoft.AspNetCore.Mvc.Abstractions;
 using Microsoft.AspNetCore.Mvc.Filters;
 using Microsoft.AspNetCore.Routing;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+using System;
```
davidfowl: Nit: sort using a properly

src/Microsoft.AspNetCore.Mvc.Core/RequireHttpsAttribute.cs
```diff
@@ -43,6 +47,13 @@ protected virtual void HandleNonHttpsRequest(AuthorizationFilterContext filterCo
                     request.Path.ToUriComponent(),
                     request.QueryString.ToUriComponent());
 
+                // use a specific SSL port if it is provided in the options,
+                // clear it otherwise to use an empty default.
+                var uri = new UriBuilder(newUrl);
```
davidfowl: Only do this if ssl port is non null

src/Microsoft.AspNetCore.Mvc.Core/MvcOptions.cs
```diff
@@ -127,5 +127,11 @@ public int MaxModelValidationErrors
         /// Gets a list of <see cref="IValueProviderFactory"/> used by this application.
         /// </summary>
         public IList<IValueProviderFactory> ValueProviderFactories { get; }
+
+        /// <summary>
+        /// Gets or sets the SSL port that is used by this application. If not set
+        /// the port won't be specified in secured url, e.g., https://localhost/path.
```
dougbu: - "URL" in all caps
- Suggest removing the commas.
- Suggest mentioning `[RequireHttps]` because nothing else uses this value. Otherwise customers may go searching for "secured URL".

src/Microsoft.AspNetCore.Mvc.Core/RequireHttpsAttribute.cs
```diff
@@ -4,6 +4,8 @@
 using System;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc.Filters;
+using Microsoft.Extensions.Options;
+using Microsoft.Extensions.DependencyInjection;
```
dougbu: nit: Sort `using`s here too.

src/Microsoft.AspNetCore.Mvc.Core/RequireHttpsAttribute.cs
dougbu: Suspect the only case where the original `newUrl` can be used is if both `SslPort` is `null` and there's no port in the `request.Host` string. Do we have a way to inspect `request.Host` that's less icky than instantiating a `UriBuilder`?

src/Microsoft.AspNetCore.Mvc.Core/RequireHttpsAttribute.cs
```diff
@@ -43,6 +47,13 @@ protected virtual void HandleNonHttpsRequest(AuthorizationFilterContext filterCo
                     request.Path.ToUriComponent(),
                     request.QueryString.ToUriComponent());
```
dougbu: Does `UriBuilder` work better with or without the puny-coding that `ToUriComponent()` adds?

test/Microsoft.AspNetCore.Mvc.Core.Test/RequireHttpsAttributeTests.cs
```diff
@@ -36,8 +41,8 @@ public void OnAuthorization_AllowsTheRequestIfItIsHttps()
                 // host, pathbase, path, query, expectedRedirectUrl
                 return new TheoryData<string, string, string, string, string>
                 {
-                    { "localhost", null, null, null, "https://localhost" },
-                    { "localhost:5000", null, null, null, "https://localhost:5000" },
+                    { "localhost", null, null, null, "https://localhost/" },
+                    { "localhost:5000", null, null, null, "https://localhost/" },
                     { "localhost", "/pathbase", null, null, "https://localhost/pathbase" },
                     { "localhost", "/pathbase", "/path", null, "https://localhost/pathbase/path" },
                     { "localhost", "/pathbase", "/path", "?foo=bar", "https://localhost/pathbase/path?foo=bar" },
```
dougbu: - Suggest adding an IPv6 example to this theory since the colons might confuse `UriBuilder` if `ToUriComponent()` isn't what it wants.
- Similar suggestion for adding a "本地主機:80" test.

Eilon: Do we need to be testing `UriBuilder` here?

dougbu: My concern is the `UriBuilder` / `ToUriComponent()` interaction, not either on its own.

src/Microsoft.AspNetCore.Mvc.Core/RequireHttpsAttribute.cs
sebastienros: I favored simplicity and efficiency over trying to re-invent port parsing and cause more issues. Was expecting this comment though and maybe someone knows a solution. If nobody chimes in I will keep it. And yes I confirm @davidfowl's comment is missing the fact that we need to clear the port when redirecting to https and `SslPort` is `null`, which fixes #4112.

sebastienros: After talking to @Tratcher we need to expose the port and domain separately in `HostString`. And be able to construct a new one from domain and port. I will then remove `UriBuilder` usage. https://github.com/aspnet/HttpAbstractions/issues/567

test/Microsoft.AspNetCore.Mvc.Core.Test/RequireHttpsAttributeTests.cs
```diff
@@ -141,6 +149,47 @@ public void HandleNonHttpsRequestExtensibility()
             Assert.Equal(StatusCodes.Status404NotFound, result.StatusCode);
         }
 
+        [Theory]
+        [InlineData("http://localhost", null, "https://localhost/")]
+        [InlineData("http://localhost:5000", null, "https://localhost/")]
+        [InlineData("http://[2001:db8:a0b:12f0::1]", null, "https://[2001:db8:a0b:12f0::1]/")]
+        [InlineData("http://[2001:db8:a0b:12f0::1]:5000", null, "https://[2001:db8:a0b:12f0::1]/")]
+        [InlineData("http://localhost:5000/path", null, "https://localhost/path")]
```
kichalla: Can you add variation(s) having query string too?

src/Microsoft.AspNetCore.Mvc.Core/RequireHttpsAttribute.cs
sebastienros: `UriBuilder` is removed in favor of the new features in `HostString`

test/Microsoft.AspNetCore.Mvc.Core.Test/RequireHttpsAttributeTests.cs
sebastienros: Done

src/Microsoft.AspNetCore.Mvc.Core/RequireHttpsAttribute.cs
sebastienros: I integrated the new feature from `HostString`, can you confirm it's ok from your side?

test/Microsoft.AspNetCore.Mvc.Core.Test/RequireHttpsAttributeTests.cs
kichalla: Thanks @sebastienros Just FYI...you seem to have squashed your commits. This makes it difficult to find the differences(to see how the feedback was addressed) from earlier commit. You could squash the commits once you get a sign off.

kichalla:
:shipit:

javiercn:
Looks good to me, :shipit:

test/Microsoft.AspNetCore.Mvc.Core.Test/RequireHttpsAttributeTests.cs
```diff
@@ -16,6 +19,8 @@ public class RequireHttpsAttributeTests
         public void OnAuthorization_AllowsTheRequestIfItIsHttps()
         {
             // Arrange
+            var options = new TestOptionsManager<MvcOptions>();
```
dougbu: Remove. This isn't used.

src/Microsoft.AspNetCore.Mvc.Core/RequireHttpsAttribute.cs
dougbu: LGTM

dougbu:
:shipit:

PR #4120, opened 2016-02-20
[Fixes #3867] Abstracting Cache tag helper storage

### LINKED ISSUE #3867: CacheTagHelper should not depend on IMemoryCache
`IMemoryCache` is a perfect match for single node scenarios in CacheTagHelper but it breaks if the website runs on more nodes, as each node would return different contents. The issue here is that there is no way to change it.

I think it would be better to either:
- Depend on `IDitributedCache` as the default implementation uses `IMemoryCache` which means it would work the same as today on a single instance, and not be broken on multiple instances.
- Or even better, define a custom interface for extensibility reasons so that each customer could decide how to store the cached tags: explicitly `IMemoryCache`, `IDistributedCache` or another implementation. Each site is different, and locking down the cache tag storage might be an issue. And having the to share the same implementation for controllers and tags might not make sense either.

A different approach would be to use a message bus implementation to invalidate `IMemoryCache` entries across nodes, but I don't think it's been planned either.

/cc @NTaylorMullen

sebastienros:


src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/MemoryHtmlFragmentCache.cs
```diff
@@ -0,0 +1,103 @@
...
+
+                var stringBuilder = new StringBuilder();
+                using (var writer = new StringWriter(stringBuilder))
+                {
+                    content.WriteTo(writer, context.HtmlEncoder);
+                }
+
+                var result = new StringBuilderHtmlContent(stringBuilder);
+
+                _memoryCache.Set(key, result, options);
```
kichalla: Curious to see how the get and set would be with a distribute cache implementation...could we have a test(not sure if a sample is better) related to it..

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/IHtmlFragmentCache.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Html;
+
+namespace Microsoft.AspNetCore.Mvc.TagHelpers.Cache
+{
+    public interface IHtmlFragmentCache
+    {
+        Task<IHtmlContent> SetAsync(object key, Func<Task<IHtmlContent>> renderContent, HtmlFragmentCacheContext context);
```
kichalla: Should the key be a `string`? The Key property and the auto generated key on the cache tag helper are strings. Also the key on IDistributedCache is a string.

sebastienros: Could be right. Also if no key is provided the tag helper generates one automatically, which can't be distributed, but is also a string. Will change.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/MemoryHtmlFragmentCache.cs
sebastienros: It would be simpler actually, no scope to use/support. And yes, I want to add a default implementation in the module for `IDistributedCache`. I created the PR to allow comments on this phase. 
@Eilon should we have instead of two different implementations by default, a single one that we can switch from options on each tag helper, like Damian wanted initially? We could register the memory cache like now, and throw a meaningful exception if the `IDistributedCache` is not registered when the option is triggered.

Eilon: I think that's an avenue worth investigating, yeah.

src/Microsoft.AspNetCore.Mvc.Razor/DependencyInjection/MvcRazorMvcCoreBuilderExtensions.cs
```diff
@@ -142,7 +140,7 @@ internal static void AddRazorViewEngineServices(IServiceCollection services)
             // Only want one ITagHelperActivator so it can cache Type activation information. Types won't conflict.
             services.TryAddSingleton<ITagHelperActivator, DefaultTagHelperActivator>();
 
-            // Consumed by the Cache tag helper to cache results across the lifetime of the application.
+            // Consumed by the default IChunkTreeCache
```
pranavkm: ?

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/MemoryHtmlFragmentCache.cs
sebastienros: Done. Please take a look when you can, like now.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
```diff
@@ -0,0 +1,218 @@
...
+
+            var stringBuilder = new StringBuilder();
+            using (var writer = new StringWriter(stringBuilder))
+            {
+                content.WriteTo(writer, context.HtmlEncoder);
+            }
+
+            var serialized = Encoding.UTF8.GetBytes(stringBuilder.ToString());
+
+            cache.Set(key, serialized, options);
```
kichalla: SetAsync

sebastienros: Done

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HtmlFragmentCacheContext.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// <summary>
+        /// Gets or sets the duration from last access that the cache entry should be evicted.
+        /// </summary>
+        public TimeSpan? ExpiresSliding { get; set; }
+
+        /// <summary>
+        /// Gets wether this context has options or not. This should be called before accessing
+        /// the <see cref="Options"/> values.
+        /// </summary>
+        public bool HasOptions => _options != null && _options.Count > 0;
```
pranavkm: `_options?.Count > 0`

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
```diff
@@ -0,0 +1,218 @@
...
+                CacheItemPriority priority;
+                if (Enum.TryParse(context.Options[PriorityOptionName], out priority))
+                {
+                    options.SetPriority(priority);
+                }
+            }
+
+            // Create an entry link scope and flow it so that any tokens related to the cache entries
+            // created within this scope get copied to this scope.
+            using (var link = _memoryCache.CreateLinkingScope())
```
pranavkm: Would the user have an expectation that triggers created in inner contexts cause the outer cache to expire? Would documentation suffice to explain that this wouldn't work for IDistributedCache?

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HtmlFragmentCacheContext.cs
pranavkm: This will get you :100: points

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
```diff
@@ -0,0 +1,218 @@
+﻿using System;
```
pranavkm: License header.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
```diff
@@ -0,0 +1,218 @@
...
+
+                var stringBuilder = new StringBuilder();
+                using (var writer = new StringWriter(stringBuilder))
+                {
+                    content.WriteTo(writer, context.HtmlEncoder);
+                }
+
+                var result = new StringBuilderHtmlContent(stringBuilder);
+
+                _memoryCache.Set(key, result, options);
```
pranavkm: `return _memoryCache.Set(key, result, options);`

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HtmlFragmentCacheContext.cs
sebastienros: :scream_cat:

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
sebastienros: Where document explains the magic behind `IMemoryCache` tokens, a note that using the distributed cache removes all of that.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
```diff
@@ -0,0 +1,219 @@
...
+
+                var stringBuilder = new StringBuilder();
+                using (var writer = new StringWriter(stringBuilder))
+                {
+                    content.WriteTo(writer, context.HtmlEncoder);
+                }
+
+                var result = new StringBuilderHtmlContent(stringBuilder);
+
+                return _memoryCache.Set(key, result, options) as IHtmlContent;
```
pranavkm: `_memoryCache.Set<IHtmlContent>(key, result, options);`. This API is kinda awful to use.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
```diff
@@ -0,0 +1,219 @@
...
+
+            if (context.ExpiresSliding != null)
+            {
+                options.SetSlidingExpiration(context.ExpiresSliding.Value);
+            }
+
+            var content = await renderContent();
+
+            var stringBuilder = new StringBuilder();
+            using (var writer = new StringWriter(stringBuilder))
```
pranavkm: Looks like this could you'd need something like `HttpResponseStreamWriter` to avoid creating large strings here. @rynowak, would you recommend using that type here?

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
```diff
@@ -0,0 +1,219 @@
...
+        public bool GetMemoryValue(string key, out IHtmlContent value)
+        {
+            return _memoryCache.TryGetValue(key, out value);
+        }
+
+        public async Task<IHtmlContent> GetDistributedValueAsync(string key, IDistributedCache cache)
+        {
+            var encoded = await cache.GetAsync(key);
+
+            if(encoded == null)
```
pranavkm: formatting.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
```diff
@@ -0,0 +1,219 @@
...
+            {
+                return null;
+            }
+
+            var content = Encoding.UTF8.GetString(encoded);
```
pranavkm: You could create a `ByteArrayHtmlString` and simply write the encoded bytes to the output as part of the `IHtmlContent.WriteTo` call.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
```diff
@@ -0,0 +1,219 @@
...
+            return new HtmlEncodedString(content);
+        }
+
+        private bool EnsureDistributedService(HtmlFragmentCacheContext context, out IDistributedCache distributedCache)
+        {
+            bool distributed;
+            if (context.Options.ContainsKey(DistributedOptionName)
+                && bool.TryParse(context.Options[DistributedOptionName], out distributed))
```
pranavkm: `&&` on previous line.

pranavkm: Does it need to do `&& distributed`?

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
```diff
@@ -0,0 +1,219 @@
...
+            if(encoded == null)
+            {
+                return null;
+            }
+
+            var content = Encoding.UTF8.GetString(encoded);
+            return new HtmlEncodedString(content);
+        }
+
+        private bool EnsureDistributedService(HtmlFragmentCacheContext context, out IDistributedCache distributedCache)
```
pranavkm: `TryGetDistributedService`?

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelper.cs
```diff
@@ -132,17 +132,38 @@ public override int Order
         public TimeSpan? ExpiresSliding { get; set; }
 
         /// <summary>
-        /// Gets or sets the <see cref="CacheItemPriority"/> policy for the cache entry.
-        /// </summary>
-        [HtmlAttributeName(CachePriorityAttributeName)]
-        public CacheItemPriority? Priority { get; set; }
```
pranavkm: Could we light some of these properties up conditionally (i.e. have them throw at runtime if you're using the wrong cache?)

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelper.cs
```diff
@@ -160,23 +181,15 @@ public override int Order
...
-                    }
+                var cacheContext = CreateCacheContext();
+
+                result = await Cache.GetValueAsync(key, cacheContext);
+
+                if (result == null)
+                {                    
+                    // The cache implementation might rewrite the result using the encoder or intercept 
+                    // its cache tokens
+                    result = await Cache.SetAsync(key, async () => await output.GetChildContentAsync(), cacheContext);
```
pranavkm: `await Cache.SetAsync(key, output.GetChildContentAsync, cacheContext);`

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/IHtmlFragmentCache.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Html;
+
+namespace Microsoft.AspNetCore.Mvc.TagHelpers.Cache
+{
+    public interface IHtmlFragmentCache
+    {
+        Task<IHtmlContent> SetAsync(string key, Func<Task<IHtmlContent>> renderContent, HtmlFragmentCacheContext context);
```
pranavkm: If the idea is to only ever use this for `CacheTagHelper`, could we pass in the `TagHelperOutput` here (instead of the Func?

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelper.cs
```diff
@@ -132,17 +132,38 @@ public override int Order
...
-        public CacheItemPriority? Priority { get; set; }
-
-        /// <summary>
         /// Gets or sets the value which determines if the tag helper is enabled or not.
         /// </summary>
         [HtmlAttributeName(EnabledAttributeName)]
         public bool Enabled { get; set; } = true;
 
+        /// <summary>
+        /// Gets or sets a key to dicriminate cached entries.
```
pranavkm: > discriminate

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelper.cs
```diff
@@ -193,12 +206,35 @@ public override int Order
...
         // Internal for unit testing
         internal string GenerateKey(TagHelperContext context)
         {
             var builder = new StringBuilder(CacheKeyPrefix);
+
+            // Use the distributed key as the dicriminator if specified to 
+            // ensure all nodes get the same key
             builder.Append(CacheKeyTokenSeparator)
-                   .Append(context.UniqueId);
+                   .Append(Key ?? context.UniqueId);
```
pranavkm: Unrelated to your PR, but could you indent it to VS defaults (4 spaces?)

src/Microsoft.AspNetCore.Mvc.TagHelpers/DependencyInjection/TagHelperExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+using System;
+using Microsoft.AspNetCore.Mvc.TagHelpers.Cache;
+using Microsoft.Extensions.Caching.Memory;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    public static class TagHelperServicesExtensions
+    {
+        public static IMvcCoreBuilder AddCacheTagHelpers(this IMvcCoreBuilder builder)
```
pranavkm: `AddCacheTagHelper`? The plural feels weird.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
```diff
@@ -0,0 +1,219 @@
...
+    /// </summary>
+    public class HybridHtmlFragmentCache : IHtmlFragmentCache
+    {
+        private const string PriorityOptionName = "priority";
+        private const string DistributedOptionName = "distributed";
+
+        private readonly IMemoryCache _memoryCache;
+        private readonly IServiceProvider _serviceProvider;
+
+        public HybridHtmlFragmentCache(IMemoryCache memoryCache, IServiceProvider serviceProvider)
```
pranavkm: You could add a second ctor which accepts the `IDistributedCache`. The DI is expected to use the longest ctor that works:

```C#
public HybridHtmlFragmentCache(IMemoryCache memoryCache)
{ }

public HybridHtmlFragmentCache(IMemoryCache memoryCache, IDistributedCache distributedCache)
{}
```

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
```diff
@@ -0,0 +1,219 @@
...
+            bool distributed;
+            if (context.Options.ContainsKey(DistributedOptionName)
+                && bool.TryParse(context.Options[DistributedOptionName], out distributed))
+            {
+                // resolve IDistributedCache dynamically. Because it's not registered by MVC (as opposed
+                // to IMemoryCache) we need to warn the user that the service is required for this option
+                distributedCache = _serviceProvider.GetService<IDistributedCache>();
+                if (distributedCache == null)
+                {
+                    throw new NotImplementedException(Resources.FormatCacheTagHelper_NoDistributedCacheService(nameof(IDistributedCache), DistributedOptionName));
```
Eilon: Not the best choice of exception type. NotSupported sounds more like it.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
Eilon: But who would call that ctor then? The way the code works right now it gets the IDistributedCache itself, on demand - from *within* an instance of this type.

src/Microsoft.AspNetCore.Mvc.TagHelpers/DependencyInjection/TagHelperExtensions.cs
Eilon: :+1:

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
pranavkm: Since this instance is instantiated by the DI it would call the right constructor. It would avoid the `IServiceProvider.GetService` call that @davidfowl loves to hate.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
sebastienros: Thanks for the suggestion.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
sebastienros: No, because by default there are not IDistributedCache implementation registered, even if someone doesn't use the `distributed` option it would fail at construction with a DI exception stating the service can't be resolved.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
sebastienros: Thanks.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
sebastienros: This is some code that was in `CacheTagHelper` before this PR, I wouldn't change anything that is working unless there is an issue open for that. This is out of the scope of this PR.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
kichalla: I think `Try` pattern is when you do not throw an error but here we do.

sebastienros: I named it like this before calling it this way, because it can throw an exception, and AFAIK the `TryXYZ` pattern is not supposed to throw anything.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
sebastienros: Correct, thank you.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/IHtmlFragmentCache.cs
sebastienros: I intentionally didn't make it dependent on any `CacheTagHelper` concept. But if removing the lambda justifies it, I will. Waiting for others to chime in.

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelper.cs
sebastienros: No, `CacheTagHelper` doesn't know about `IMemoryCache` and `IDistributedCache`.

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelper.cs
sebastienros: You really read everything. Get back on reddit!

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelper.cs
sebastienros: Ooooh !

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
Eilon: Precisely.

pranavkm: Couldn't you simply do a null check instead - `if (IsDistributedEnabled() && _distributedCache == null) { throw .. }`?

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelper.cs
pranavkm: It doesn't have to. It would set the property on the context which would throw if you're using a feature the backing cache doesn't support.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
Eilon: But you never get `null` from a DI ctor. DI will have already thrown.

sebastienros: @Eilon @pranavkm just showed me a feature in the DI, I will try it, he might be right.

Eilon: Ah, using `IEnumerable<T>`? This just seems to be getting more complicated :smile: But perhaps it's the right way to go...

pranavkm: No, using the longest matching feature (https://github.com/aspnet/DependencyInjection/blob/dev/src/Microsoft.Extensions.DependencyInjection.Specification.Tests/DependencyInjectionSpecificationTests.cs#L542). It'll construct it with `IDistributedCache` if the service is registered, otherwise pick the constructor with the single `IMemoryCache` argument.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
sebastienros: Doesn't work on the `TextWriter`.

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelper.cs
sebastienros: Nope, only the properties globally supported by the cache storage implementations should be there. Any other implementation can have custom options. If we want to through an exception is a parameter is set but not supported, any implementation can do that, not the context object.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HtmlFragmentCacheContext.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// </summary>
+        public TimeSpan? ExpiresAfter { get; set; }
+
+        /// <summary>
+        /// Gets or sets the duration from last access that the cache entry should be evicted.
+        /// </summary>
+        public TimeSpan? ExpiresSliding { get; set; }
+
+        /// <summary>
+        /// Gets wether this context has options or not. This should be called before accessing
```
NTaylorMullen: https://www.vocabulary.com/dictionary/wether

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
```diff
@@ -0,0 +1,217 @@
...
+    /// </summary>
+    public class HybridHtmlFragmentCache : IHtmlFragmentCache
+    {
+        private const string PriorityOptionName = "priority";
+        private const string DistributedOptionName = "distributed";
+
+        private readonly IMemoryCache _memoryCache;
+        private readonly IDistributedCache _distributedCache;
+
+        public HybridHtmlFragmentCache(IMemoryCache memoryCache)
```
NTaylorMullen: Will need docs everywhere.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
```diff
@@ -0,0 +1,217 @@
...
+            if(encoded == null)
+            {
+                return null;
+            }
+
+            var content = Encoding.UTF8.GetString(encoded);
+            return new HtmlEncodedString(content);
+        }
+
+        private bool EnsureDistributedService(HtmlFragmentCacheContext context)
```
NTaylorMullen: May also want to validate `_distributedCache` is not `null` and if so to provide a nice error

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
```diff
@@ -0,0 +1,217 @@
...
+                    content.WriteTo(writer, context.HtmlEncoder);
+                }
+
+                var result = new StringBuilderHtmlContent(stringBuilder);
+
+                return _memoryCache.Set<IHtmlContent>(key, result, options);
+            }
+        }
+
+        public async Task<IHtmlContent> SetDistributedAsync(string key, Func<Task<IHtmlContent>> renderContent, HtmlFragmentCacheContext context, IDistributedCache cache)
```
NTaylorMullen: private

NTaylorMullen: Should do the same elsewhere, looks like there's too many `public` methods in this class

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/IHtmlFragmentCache.cs
NTaylorMullen: If we have this take in a `TagHelperOutput` we'd probably want to change the names of all these interfaces to `IHtmlTagHelperFragmentCache` etc. Might be too specific.

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelper.cs
```diff
@@ -35,18 +34,19 @@ public class CacheTagHelper : TagHelper
...
         private const string EnabledAttributeName = "enabled";
+        private IDictionary<string, string> _optionValues;
         private static readonly char[] AttributeSeparator = new[] { ',' };
-
         /// <summary>
         /// Creates a new <see cref="CacheTagHelper"/>.
         /// </summary>
-        /// <param name="memoryCache">The <see cref="IMemoryCache"/>.</param>
-        public CacheTagHelper(IMemoryCache memoryCache, HtmlEncoder htmlEncoder)
+        /// <param name="cache">The <see cref="IMemoryCache"/>.</param>
```
NTaylorMullen: `<see cref="IMemoryCache"/>` > `IHtmlFragmentCache `

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelper.cs
```diff
@@ -132,17 +132,38 @@ public override int Order
...
+        /// Additional options for the cache.
+        /// </summary>
+        [HtmlAttributeName(DictionaryAttributePrefix = OptionValuesPrefix)]
+        public IDictionary<string, string> OptionValues
+        {
+            get
+            {
+                if (_optionValues == null)
+                {
+                    _optionValues = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
```
NTaylorMullen: Why ignorecase for options?

sebastienros: Because the ones in `CacheTagHelper` are. So why are they like this in `CacheTagHelper`, just a personal preference as they are taken from html attributes.

src/Microsoft.AspNetCore.Mvc/project.json
```diff
@@ -16,6 +16,7 @@
     "Microsoft.AspNetCore.Mvc.Formatters.Json": "1.0.0-*",
     "Microsoft.AspNetCore.Mvc.Localization": "1.0.0-*",
     "Microsoft.AspNetCore.Mvc.Razor": "1.0.0-*",
+    "Microsoft.AspNetCore.Mvc.TagHelpers": "1.0.0-*",
```
NTaylorMullen: Did you run this bit by @DamianEdwards ?

NTaylorMullen:
:watch:

src/Microsoft.AspNetCore.Mvc/project.json
dougbu: +1 we have left the MVC tag helpers as optional to date. There's only two included by default and they aren't in the TagHelpers package.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
sebastienros: :dancers:

sebastienros:
:up::date:

src/Microsoft.AspNetCore.Mvc/project.json
DamianEdwards: Before now though, Tag Helpers didn't need their own services in DI. This is going to be a pain to use if it's not part of the MVC package, as you'll have to add the Tag Helpers package, and explicitly add the Tag Helper services to DI, and register them in the views. So many steps makes me :cry: so if there's no other objection to this, I say we keep it, or find a way to make the experience of adding and use the Tag Helpers not suck.

DamianEdwards:
I'm having a little trouble understanding how this will be used. Can you provide some usage examples please?

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HtmlFragmentCacheContext.cs
```diff
@@ -0,0 +1,69 @@
...
+    {
+        private IDictionary<string, string> _options;
+
+        /// <summary>
+        /// Creates a new <see cref="HtmlFragmentCacheContext"/>.
+        /// </summary>
+        /// <param name="htmlEncoder">The <see cref="HtmlEncoder"/>.</param>
+        public HtmlFragmentCacheContext(HtmlEncoder htmlEncoder)
+        {
+            HtmlEncoder = htmlEncoder;
```
NTaylorMullen: Should probably `null` check the `HtmlEncoder`, otherwise it'll explode when used in the `HybridHtmlFragmentCache`

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
```diff
@@ -0,0 +1,255 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Html;
+using Microsoft.Extensions.Caching.Distributed;
+using Microsoft.Extensions.Caching.Memory;
+
+namespace Microsoft.AspNetCore.Mvc.TagHelpers.Cache
+{
+    /// <summary>
+    /// This implementation of <see cref="IHtmlFragmentCache"/> is able to switch from 
+    /// <see cref="IMemoryCache"/> to <see cref="IDistributedCache"/>
```
NTaylorMullen: `.`

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
```diff
@@ -0,0 +1,255 @@
...
+            // The content is stored using UTF8 encoding on the distributed cache service.
+            var serialized = Encoding.UTF8.GetBytes(stringBuilder.ToString());
+
+            await cache.SetAsync(key, serialized, options);
+
+            return content;
+        }
+
+        /// <inheritdoc />
+        public Task<IHtmlContent> GetValueAsync(string key, HtmlFragmentCacheContext context)
```
NTaylorMullen: nit: Group public methods with each other.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HybridHtmlFragmentCache.cs
```diff
@@ -0,0 +1,255 @@
...
+                else
+                {
+                    // no existing cache entry
+                    return Task.FromResult<IHtmlContent>(null);
+                }
+            }
+        }
+
+        /// <summary>
+        /// Retrieves a cache entry from the <see cref="IMemoryCache"/> service.
```
NTaylorMullen: lmao, note: you don't need to doc private methods

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/IHtmlFragmentCache.cs
```diff
@@ -0,0 +1,33 @@
...
+namespace Microsoft.AspNetCore.Mvc.TagHelpers.Cache
+{
+    /// <summary>
+    /// An implementation of this interface provides a service to cache html fragments.
+    /// One client of this service is <see cref="CacheTagHelper"/>.
+    /// </summary>
+    public interface IHtmlFragmentCache
+    {
+        /// <summary>
+        /// Evaluates the <see cref="IHtmlContent"/> to render and caches it.
```
NTaylorMullen: `cache it`

src/Microsoft.AspNetCore.Mvc.TagHelpers/DependencyInjection/TagHelperExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Mvc.TagHelpers.Cache;
+using Microsoft.Extensions.Caching.Memory;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    public static class TagHelperServicesExtensions
```
NTaylorMullen: Docs on these public guys

NTaylorMullen:
:watch: just for doc comment fixes.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/IHtmlFragmentCache.cs
sebastienros: Don't think so. `Evaluates` and `caches` have the same subject in this case. Changing it would give it another meaning, which you might find better, but let's ask @dougbu.

dougbu: Agree w/ @sebastienros. We write summaries about the method and that thingy is consistently singular. So `Evaluates` and `caches` are consistent, correct, read nicely, are generally better together (than another alternative), and [some other words].

src/Microsoft.AspNetCore.Mvc/project.json
sebastienros: > Before now though, Tag Helpers didn't need their own services in DI

They did, `IMemoryCache`, which was registered by MVC _by chance_. Now it's `IMemoryCache` and `IHtmlFragmentCache` which are also registered by MVC, on purpose for this feature. So it doesn't change, nothing more to register to use `CacheTagHelper` from MVC.

Before you had to:
- Reference TagHelper package
- Import the tag helpers in the view

And now it's
- Import the tag helpers in the view

And _Optionally_ register `IDistributedCache` if you want to use it

sebastienros:
:up::date:

src/Microsoft.AspNetCore.Mvc/project.json
sebastienros: Usage:

```
<cahe key="myentry" options-distributed="true">
My distributed cached content 
</cache>
```

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/HtmlFragmentCacheContext.cs
```diff
@@ -0,0 +1,75 @@
...
+    {
+        private IDictionary<string, string> _options;
+
+        /// <summary>
+        /// Creates a new <see cref="HtmlFragmentCacheContext"/>.
+        /// </summary>
+        /// <param name="htmlEncoder">The <see cref="HtmlEncoder"/>.</param>
+        public HtmlFragmentCacheContext(HtmlEncoder htmlEncoder)
+        {
+
```
NTaylorMullen: Nuke

NTaylorMullen:
:shipit:

sebastienros:
Deprecated

PR #4139, opened 2016-02-23
Moving DNX dependencies to test only

pranavkm:


test/Microsoft.AspNetCore.Mvc.FunctionalTests/DnxRoslynCompilationService.cs
```diff
@@ -0,0 +1,268 @@
...
+using Microsoft.Extensions.CompilationAbstractions;
+using Microsoft.Extensions.Options;
+using Microsoft.Extensions.PlatformAbstractions;
+
+namespace Microsoft.AspNetCore.Mvc.Razor.Internal
+{
+    /// <summary>
+    /// A type that uses Roslyn to compile C# content and <see cref="ILibraryExporter"/> to find out references.
+    /// </summary>
+    public class DnxRoslynCompilationService : ICompilationService
```
pranavkm: This is used for running tests from VS \ `dnx test` in command line for dnxcore50 tfm. I'll follow up with another PR that switches to the CLI test runner which removes dnx dependencies in command line builds.

test/Microsoft.AspNetCore.Mvc.Razor.Test/Internal/DefaultRoslynCompilationServiceTest.cs
```diff
@@ -1,393 +0,0 @@
-// Copyright (c) .NET Foundation. All rights reserved.
```
pranavkm: Will revive this in a follow up - https://github.com/aspnet/Mvc/issues/4140

test/WebSites/RazorWebSite/Startup.cs
```diff
@@ -22,6 +22,8 @@ public void ConfigureServices(IServiceCollection services)
                     options.ViewLocationExpanders.Add(new NonMainPageViewLocationExpander());
 #if DNXCORE50
                     options.ParseOptions = options.ParseOptions.WithPreprocessorSymbols("DNXCORE50", "DNXCORE50_CUSTOM_DEFINE");
+#else
```
pranavkm: This + the change in FrameworkSpecificHelper is to let the test work in VS.

pranavkm:
cc @pakrym

PR #4151, opened 2016-02-25
[Perf] One UrlHelper per context and pool StringBuilders used by it

mnltejaswini:
I modified https://github.com/aspnet/Performance/blob/dev/testapp/BasicViews/Views/TagHelpers/Index.cshtml 
 to contain 6 action Links with href generated by TagHelpers. 
 And the loadtest is executed with the following settings
 loadtest -k -n 3000 --rps 100 

The total objects(StringBuilder, char[] and UrlHelper)  saved is 2% and bytes saved is 1.5%

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
@@ -15,6 +15,10 @@ namespace Microsoft.AspNetCore.Mvc.Routing
     /// </summary>
     public class UrlHelper : IUrlHelper
     {
+
+        // Perf: Share the StringBuilder object across multiple calls of GenerateURL for this UrlHelper
+        private StringBuilder _stringBuilder;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="UrlHelper"/> class using the specified action context and
         /// action selector.
@@ -201,6 +205,16 @@ namespace Microsoft.AspNetCore.Mvc.Routing
             });
         }
 
+        private StringBuilder GetStringBuilder()
+        {
+            if(_stringBuilder == null)
+            {
+                _stringBuilder = new StringBuilder();
+            }
+
+            return _stringBuilder;
+        }
+
         /// <summary>
         /// Generates the URL using the specified components.
         /// </summary>
@@ -219,29 +233,38 @@ namespace Microsoft.AspNetCore.Mvc.Routing
             // VirtualPathData.VirtualPath returns string.Empty instead of null.
             Debug.Assert(pathData.VirtualPath != null);
 
-            var builder = new StringBuilder();
-            if (string.IsNullOrEmpty(protocol) && string.IsNullOrEmpty(host))
+            var builder = GetStringBuilder();
+            try
             {
-                AppendPathAndFragment(builder, pathData, fragment);
-                // We're returning a partial URL (just path + query + fragment), but we still want it to be rooted.
-                if (builder.Length == 0 || builder[0] != '/')
+                if (string.IsNullOrEmpty(protocol) && string.IsNullOrEmpty(host))
                 {
-                    builder.Insert(0, '/');
+                    AppendPathAndFragment(builder, pathData, fragment);
+                    // We're returning a partial URL (just path + query + fragment), but we still want it to be rooted.
+                    if (builder.Length == 0 || builder[0] != '/')
+                    {
+                        builder.Insert(0, '/');
+                    }
                 }
+                else
+                {
+                    protocol = string.IsNullOrEmpty(protocol) ? "http" : protocol;
+                    builder.Append(protocol);
+
+                    builder.Append("://");
+
+                    host = string.IsNullOrEmpty(host) ? HttpContext.Request.Host.Value : host;
+                    builder.Append(host);
+                    AppendPathAndFragment(builder, pathData, fragment);
+                }
+
+                var path = builder.ToString();
+                return path;
             }
-            else
+            finally
             {
-                protocol = string.IsNullOrEmpty(protocol) ? "http" : protocol;
-                builder.Append(protocol);
-
-                builder.Append("://");
-
-                host = string.IsNullOrEmpty(host) ? HttpContext.Request.Host.Value : host;
-                builder.Append(host);
-                AppendPathAndFragment(builder, pathData, fragment);
+                // Clear the StringBuilder so that it can reused for the next call.
+                builder.Clear();
             }
-
-            return builder.ToString();
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelperFactory.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelperFactory.cs
@@ -1,6 +1,10 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc.Core;
+
 namespace Microsoft.AspNetCore.Mvc.Routing
 {
     /// <summary>
@@ -11,7 +15,31 @@ namespace Microsoft.AspNetCore.Mvc.Routing
         /// <inheritdoc />
         public IUrlHelper GetUrlHelper(ActionContext context)
         {
-            return new UrlHelper(context);
+            var httpContext = context.HttpContext;
+
+            if (httpContext == null)
+            {
+                throw new ArgumentException(Resources.FormatPropertyOfTypeCannotBeNull(
+                    nameof(ActionContext.HttpContext),
+                    nameof(ActionContext)));
+            }
+
+            if (httpContext.Items == null)
+            {
+                throw new ArgumentException(Resources.FormatPropertyOfTypeCannotBeNull(
+                    nameof(HttpContext.Items),
+                    nameof(HttpContext)));
+            }
+
+            // Perf: Create only one UrlHelper per context
+            var urlHelper = httpContext.Items[typeof(IUrlHelper)] as IUrlHelper;
+            if (urlHelper == null)
+            {
+                urlHelper = new UrlHelper(context);
+                httpContext.Items[typeof(IUrlHelper)] = urlHelper;
+            }
+
+            return urlHelper;
         }
     }
 }
--- a/test/Microsoft.AspNetCore.Mvc.Core.Test/LocalRedirectResultTest.cs
+++ b/test/Microsoft.AspNetCore.Mvc.Core.Test/LocalRedirectResultTest.cs
@@ -3,6 +3,7 @@
 
 using System;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Internal;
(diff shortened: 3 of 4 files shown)


src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelperFactory.cs
```diff
@@ -9,9 +12,40 @@ namespace Microsoft.AspNetCore.Mvc.Routing
...
+            }
+        }
         public IUrlHelper GetUrlHelper(ActionContext context)
         {
-            return new UrlHelper(context);
+            var httpContext = context.HttpContext;
+            IUrlHelper urlHelper = null;
+            if (httpContext != null && httpContext.Features != null)
+            {
+                urlHelper = httpContext.Features[typeof(IUrlHelper)] as IUrlHelper;
```
rynowak: This should go in `HttpContext.Items` - features is something different entirely

rynowak:
What's the impact of using `HttpContext.Items` with the `UrlHelper` and **not** pooling the stringbuilders? I'm really worried about contention on the pool with this change.

mnltejaswini:
0.75% savings in total objects and 0.5% in total bytes allocated

rynowak:
> 0.75% savings in total objects and 0.5% in total bytes allocated

Does that include keeping the stringbuilder as a field of the UrlHelper? The idea is that if you generate 8 links on a page, it's still just one builder per request. I'd be much more comfortable with that than pooling the builder per-operation.

mnltejaswini:
Discussed with Ryan offline. I will change UrlHelper to use a single StringBuilder object and remove pooling.

src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelperFactory.cs
```diff
@@ -1,17 +1,37 @@
...
-    {
-        /// <inheritdoc />
+    {        
         public IUrlHelper GetUrlHelper(ActionContext context)
         {
-            return new UrlHelper(context);
+            var httpContext = context.HttpContext;
+            IUrlHelper urlHelper = null;
+
+            /// Perf: Create only one UrlHelper per context
```
mnltejaswini: This comment will be corrected

samples/UrlHelperSample.Web/CustomUrlHelper.cs
```diff
@@ -2,9 +2,11 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Text;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.Routing;
+using Microsoft.Extensions.ObjectPool;
```
rynowak: Undo

src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
```diff
@@ -6,6 +6,7 @@
 using System.Text;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Routing;
+using Microsoft.Extensions.ObjectPool;
```
rynowak: Undo

src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
```diff
@@ -28,6 +33,7 @@ public UrlHelper(ActionContext actionContext)
             }
 
             ActionContext = actionContext;
+            _stringBuilder = new StringBuilder();
```
rynowak: It would be valuable in some scenarios to make this field lazy. There are uses of the `UrlHelper` that don't call `GenerateUrl`

src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelperFactory.cs
```diff
@@ -1,17 +1,37 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Text;
+using Microsoft.Extensions.ObjectPool;
```
rynowak: Remove using for the pool

src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelperFactory.cs
```diff
@@ -1,17 +1,37 @@
...
 namespace Microsoft.AspNetCore.Mvc.Routing
 {
     /// <summary>
     /// A default implementation of <see cref="IUrlHelperFactory"/>.
     /// </summary>
     public class UrlHelperFactory : IUrlHelperFactory
-    {
-        /// <inheritdoc />
+    {        
         public IUrlHelper GetUrlHelper(ActionContext context)
```
rynowak: Don't erase the docs

src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelperFactory.cs
```diff
@@ -1,17 +1,37 @@
...
         {
-            return new UrlHelper(context);
+            var httpContext = context.HttpContext;
+            IUrlHelper urlHelper = null;
+
+            /// Perf: Create only one UrlHelper per context
+            if (httpContext != null && httpContext.Items != null)
```
rynowak: This isn't something that should ever happen.

rynowak: Throw an exception if this is the case. We have a resource already set up for this case: "property X or Y cannot be null".

test/Microsoft.AspNetCore.Mvc.Core.Test/Routing/UrlHelperTest.cs
```diff
@@ -12,6 +12,7 @@
 using Microsoft.AspNetCore.Mvc.Abstractions;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.ObjectPool;
```
rynowak: Undo

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/RemoteAttributeTest.cs
```diff
@@ -19,6 +20,7 @@
 using Microsoft.Extensions.Logging.Testing;
 using Microsoft.Extensions.Options;
 using Microsoft.Extensions.WebEncoders.Testing;
+using Microsoft.Extensions.ObjectPool;
```
rynowak: Undo

rynowak:
General idea of the changes looks right, needs some cleanup :watch:

src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
```diff
@@ -15,6 +16,10 @@ namespace Microsoft.AspNetCore.Mvc.Routing
     /// </summary>
     public class UrlHelper : IUrlHelper
     {
+
+        /// Perf: Share the StringBuilder object across multiple calls of GenerateURL for this UrlHelper
+        private StringBuilder _stringBuilder;
```
pranavkm: `readonly`

src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
```diff
@@ -199,7 +205,7 @@ public virtual string Link(string routeName, object values)
                 Protocol = HttpContext.Request.Scheme,
                 Host = HttpContext.Request.Host.ToUriComponent()
             });
-        }
+        }
```
pranavkm: Ctrl K - D (remove extra whitespaces)

src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelperFactory.cs
mnltejaswini: One of the unit tests LocalRedirectResult mocks HttpContext, in which case the Items and Features are null.

rynowak:
Fix the test then instead of adding product behavior for something that won't ever happen.

src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
```diff
@@ -201,6 +205,16 @@ public virtual string Link(string routeName, object values)
             });
         }
 
+        private StringBuilder GetStringBuilder()
+        {
+            if(_stringBuilder == null)
```
rynowak: Fix spacing here

src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
```diff
@@ -219,29 +233,39 @@ protected virtual string GenerateUrl(string protocol, string host, VirtualPathDa
...
+                return path;
             }
-
-            return builder.ToString();
+            finally
+            {
+                // Clear the StringBuilder so that it can reused for the next call.
+                builder.Clear();
+            }
+
```
rynowak: Remove extra blank line

src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelperFactory.cs
```diff
@@ -11,7 +13,27 @@ public class UrlHelperFactory : IUrlHelperFactory
...
+            var httpContext = context.HttpContext;
+
+            if (httpContext == null)
+            {
+                throw new ArgumentNullException(nameof(httpContext));
+            }
+
+            if (httpContext.Items == null)
+            {
+                throw new ArgumentNullException(nameof(httpContext.Items));
```
rynowak: This is not correct. Use `ArgumentException` with FormatPropertyOfTypeCannotBeNull

samples/UrlHelperSample.Web/CustomUrlHelper.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Text;
```
rynowak: Revert this file

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/RemoteAttributeTest.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Text;
```
rynowak: Revert

rynowak:
:watch:

rynowak:
:shipit:

tmds:
Hi @rynowak @mnltejaswini can you explain to me why this code is thread safe? If there is a single UrlHelper per HttpContext, does that mean it is only accessed by one thread at a time?

davidfowl:
It's not thread safe and doesn't need to be. Accessing the http context from multiple threads at the same time is unsupported, there are no guarantees if you do that, nothing on it is thread safe.

tmds:
ok, thanks!

Issue #4140, opened 2016-02-23
Revive DefaultRoslynCompilationServicesTest


pranavkm:
We had tests for the DNX based RoslynCompilationService, but not the DependencyModel based compilation service. We need to revive these.

pranavkm:
Also revive DefaultAssemblyProviderTests.

PR #4154, opened 2016-02-25
Use the entry assembly when DependencyContext is null

### LINKED ISSUE #4136: DependencyContextAssemblyProvider assumes the entry assembly has a .deps.json file
Tests in the aspnet/Performance repo are failing because [`DependencyContextAssemblyProvider.GetCandidateLibraries()`](https://github.com/aspnet/Mvc/blob/25eb50120eceb62fd24ab5404210428fcdf0c400/src/Microsoft.AspNetCore.Mvc.Core/Infrastructure/DependencyContextAssemblyProvider.cs#L64) assumes that `DependencyContext.Default` is not null. Our tests are throwing a `System.NullReferenceException` when accessing `DependencyContext.Default` as it's null.

Note that the current implementation of [`DependencyContext.get_Default()`](https://github.com/dotnet/cli/blob/master/src/Microsoft.Extensions.DependencyModel/DependencyContext.cs#L50) allows for it to be null when the entry assembly (in our case xunit.console.exe) doesn't provide a dependencies description file (in our case xunit.console.deps.json).

Let me know if I'm interpreting something wrong, this has the Performance repository blocked aspnet/Performance#62.

@NTaylorMullen @muratg @SajayAntony

pranavkm:
Fixes #4136

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/Infrastructure/DefaultAssemblyProvider.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Infrastructure/DefaultAssemblyProvider.cs
@@ -17,6 +17,7 @@ namespace Microsoft.AspNetCore.Mvc.Infrastructure
     /// </summary>
     public class DefaultAssemblyProvider : IAssemblyProvider
     {
+        private readonly Assembly _entryAssembly;
         private readonly DependencyContext _dependencyContext;
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


pranavkm:
cc @davidfowl

pranavkm:
Will add tests as part of https://github.com/aspnet/Mvc/issues/4140

davidfowl:
:shipit: We also need to look at what happens to compilation in this world.

PR #4159, opened 2016-02-25
Make ServiceCollectionExtensions consistent

javiercn:


src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs
```diff
@@ -43,16 +60,15 @@ public static IMvcCoreBuilder AddMvcCore(this IServiceCollection services)
                 throw new ArgumentNullException(nameof(services));
             }
 
-            ConfigureDefaultServices(services);
-
-            AddMvcCoreServices(services);
-
-            if (setupAction != null)
+            if (setupAction == null)
```
dougbu: Why change this to `throw` when passed a `null` delegate? The previous code seemed more usable, allowing users to call the full overload all the time if they prefer.

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs
```diff
@@ -22,8 +22,16 @@
...
+    /// Extension methods for setting up essential MVC services in an <see cref="IServiceCollection" />.
+    /// </summary>
     public static class MvcCoreServiceCollectionExtensions
     {
+        /// <summary>
+        /// Adds essential MVC services to the specified <see cref="IServiceCollection" />.
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection" /> to add services to.</param>
+        /// <returns>An <see cref="IMvcCoreBuilder"/> that can be used to further configure the MVC services.</returns>
         public static IMvcCoreBuilder AddMvcCore(this IServiceCollection services)
```
dougbu: Are `IMvcCoreBuilder` and `MvcCoreBuilder` already fully-documented?

src/Microsoft.AspNetCore.Mvc/MvcServiceCollectionExtensions.cs
```diff
@@ -60,11 +44,31 @@ public static IMvcBuilder AddMvc(this IServiceCollection services, Action<MvcOpt
...
+
+        /// <summary>
+        /// Adds MVC services to the specified <see cref="IServiceCollection" />.
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection" /> to add services to.</param>
+        /// <param name="setupAction">An action delegate to configure the provided <see cref="MvcOptions"/>.</param>
+        /// <returns>An <see cref="IMvcBuilder"/> that can be used to further configure the MVC services.</returns>
+        public static IMvcBuilder AddMvc(this IServiceCollection services, Action<MvcOptions> setupAction)
+        {
+            if (services == null)
```
dougbu: Same question about adding this new error case.

dougbu:
:watch:

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs
javiercn: Some repositories did this, some other repositories did what you mention. I spoke to @Eilon about it and we set on just throwing in this case. 
Why would you call this with null instead of just calling the other overload? We don't consider this to be an scenario.
If you still want to do this, you only need to pass in ```_=>{}```. instead of null.

src/Microsoft.AspNetCore.Mvc/MvcServiceCollectionExtensions.cs
javiercn: Same answer as above

javiercn:
Things to pay attention to:
* Comments (There's been a lot of copy paste, so mistakes can happen).
* Method names
* Null checks for arguments
* Overload with configuration action calls parameterless overload.
* Parameterless overload does the work and brings in direct depenendencies
  (Like routing on MVC, for most repositories it means making sure AddOptions is called).

javiercn:
@dougbu Updated

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/IMvcBuilder.cs
```diff
@@ -3,8 +3,14 @@
 
 namespace Microsoft.Extensions.DependencyInjection
 {
+    /// <summary>
+    /// Marker interface for configuring MVC services.
```
dougbu: What is "marker" about this interface? Usually reserve that word for empty interfaces that just say "treat me special".

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs
dougbu: :ok:

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs
```diff
@@ -31,9 +39,18 @@ public static IMvcCoreBuilder AddMvcCore(this IServiceCollection services)
...
+            AddMvcCoreServices(services);
+
+            return new MvcCoreBuilder(services);
         }
 
+        /// <summary>
+        /// Adds essential MVC services to the specified <see cref="IServiceCollection" />.
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection" /> to add services to.</param>
+        /// <param name="setupAction">An action delegate to configure the provided <see cref="MvcOptions"/>.</param>
```
dougbu: Pick one: `An <see cref="Action"/> ...` or `A delegate ...`.

src/Microsoft.AspNetCore.Mvc.Core/Internal/MvcBuilder.cs
```diff
@@ -6,8 +6,17 @@
...
+    /// <summary>
+    /// Allows fine grained configuration of MVC services.
+    /// </summary>
     public class MvcBuilder : IMvcBuilder
     {
+        /// <summary>
+        /// Initializes a new <see cref="MvcBuilder"/> instance.
+        /// </summary>
+        /// <param name="services">
+        /// The <see cref="IServiceCollection"/> where MVC services are configured.
```
dougbu: Suggest sticking w/ `The <see cref="IServiceCollection" /> to add services to.`

src/Microsoft.AspNetCore.Mvc.Core/Internal/MvcCoreBuilder.cs
```diff
@@ -6,8 +6,15 @@
...
 {
+    /// <summary>
+    /// Allows fine grained configuration of essential MVC services.
+    /// </summary>
     public class MvcCoreBuilder : IMvcCoreBuilder
     {
+        /// <summary>
+        /// Initializes a new instance of <see cref="MvcCoreBuilder"/>.
+        /// </summary>
+        /// <param name="services"></param>
```
dougbu: Fill in the blank.

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs
```diff
@@ -22,8 +22,16 @@
 
 namespace Microsoft.Extensions.DependencyInjection
 {
+    /// <summary>
+    /// Extension methods for setting up essential MVC services in an <see cref="IServiceCollection" />.
+    /// </summary>
     public static class MvcCoreServiceCollectionExtensions
     {
+        /// <summary>
+        /// Adds essential MVC services to the specified <see cref="IServiceCollection" />.
```
dougbu: Suggest `<see cref="IServiceCollection" />` -> `<paramref name="services"/>`.

src/Microsoft.AspNetCore.Mvc/MvcServiceCollectionExtensions.cs
```diff
@@ -16,7 +16,7 @@ public static class MvcServiceCollectionExtensions
         /// Adds MVC services to the specified <see cref="IServiceCollection" />.
```
dougbu: Old comments but use `<paramref name="services"/>` here too.

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/AttributeRoutingTest.cs
```diff
@@ -193,13 +193,16 @@ private static IServiceProvider CreateServices(params ActionDescriptor[] actions
                 .SetupGet(o => o.Value)
                 .Returns(new RouteOptions());
 
-            return new ServiceCollection()
-                .AddSingleton<IInlineConstraintResolver>(new DefaultInlineConstraintResolver(routeOptions.Object))
+            var services = new ServiceCollection();
+            services.AddSingleton<IInlineConstraintResolver>(new DefaultInlineConstraintResolver(routeOptions.Object))
```
dougbu: Leave first `.AddSingleton()` on its own line.

dougbu:
Nice to see these classes documented.

dougbu:
:watch:

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs
javiercn: Yes

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/IMvcBuilder.cs
javiercn: Any suggestion?

dougbu: "An interface ..."

src/Microsoft.AspNetCore.Mvc/MvcServiceCollectionExtensions.cs
javiercn: Id rather keep <see cref="IServiceCollection" />. Its way more useful to have the type directly available on the documentation where you can see the type definition on 1 click than having to navigate to the parameter and then to the parameter. Also, this being an extension method, its weird to make reference to the 'this' parameter. Overall, I don't think it makes a lot of difference to use one or the other, it's clear in both cases.

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/AttributeRoutingTest.cs
javiercn: Is this what you mean?
```
            var services = new ServiceCollection();
            services.AddSingleton<IInlineConstraintResolver>(new DefaultInlineConstraintResolver(routeOptions.Object))
                .AddSingleton<UrlEncoder>(new UrlTestEncoder());

            services.AddRouting();

            services
                .AddSingleton(actionDescriptorProvider.Object)
                .AddSingleton<ILoggerFactory>(NullLoggerFactory.Instance);
```

dougbu: That's unchanged from what you had. I meant
```c#
services
    .AddSingleton<IInlineConstraintResolver>(new DefaultInlineConstraintResolver(routeOptions.Object))
    .Add...
```

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs
dougbu: Actually it was "No" but you've fixed that in later iterations :smiley_cat:

src/Microsoft.AspNetCore.Mvc/MvcServiceCollectionExtensions.cs
dougbu: Yeah :ok: that makes sense.

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs
dougbu: As discussed [here](https://github.com/aspnet/Mvc/pull/4159#discussion-diff-54208453), ignore this comment too. `services` is a `this` parameter.

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/IMvcBuilder.cs
Eilon: Agree with @dougbu, drop "marker" from this.

dougbu:
This change looks :ok: but I'm not comfortable about introducing the new error cases discussed [here](https://github.com/aspnet/Mvc/pull/4159#discussion-diff-54161395). The new `Exception` case doesn't match how numerous overloads and extension methods in MVC and other repos work. In fact it's often impossible for the longest overload to check parameter values for `null` e.g. when they're called from extension methods.

@javiercn please either change the extension methods back to doing the work in the longest overload or have someone else sign off on this.

javiercn:
@Eilon Am I good to go here?

Eilon:
Yes, :shipit: . There's nothing wrong with this pattern.

PR #4161, opened 2016-02-26
 * Linux FS is case sensitive

ryanbrandenburg:


dougbu:
:shipit:

muratg:
How did this work in our CI?

ryanbrandenburg:
My guess is that the casing was the real problem here, so since aspnetCI is windows (case insensitive) vs the failing builds on Travis/linux (case sensitive) it didn't have any trouble with the path.

ryanbrandenburg:
Looks like this fixes the problems we were having but there's something else that was hiding under it.

ryanbrandenburg:
I'm just going to check this in and make the other fix as it's own PR since it improves things at least.

PR #4170, opened 2016-02-26
Transition to Netstandard.

NTaylorMullen:
- dotnet5.X => netstandard1.y (where y = x-1).
- DNXCore50 => netstandardapp1.5.
- Applied the same changes to ifdefs.

### DIFF
--- a/samples/ActionConstraintSample.Web/project.json
+++ b/samples/ActionConstraintSample.Web/project.json
@@ -12,9 +12,12 @@
     "Microsoft.AspNetCore.Server.Kestrel": "1.0.0-*"
   },
   "frameworks": {
-    "net451": { },
-    "dnxcore50": {
-      "imports": "portable-net451+win8",
+    "net451": {},
+    "netstandardapp1.5": {
+      "imports": [
+        "dnxcore50",
+        "portable-net451+win8"
+      ],
       "dependencies": {
         "NETStandard.Library": "1.0.0-*"
       }
--- a/src/Microsoft.AspNetCore.Mvc.ApiExplorer/project.json
+++ b/src/Microsoft.AspNetCore.Mvc.ApiExplorer/project.json
@@ -8,7 +8,9 @@
   "compilationOptions": {
     "warningsAsErrors": true,
     "keyFile": "../../tools/Key.snk",
-    "nowarn": [ "CS1591" ],
+    "nowarn": [
+      "CS1591"
+    ],
     "xmlDoc": true
   },
   "dependencies": {
@@ -24,8 +26,11 @@
   },
   "frameworks": {
     "net451": {},
-    "dotnet5.4": {
-      "imports": "portable-net451+win8"
+    "netstandard1.3": {
+      "imports": [
+        "dotnet5.4",
+        "portable-net451+win8"
+      ]
     }
   }
 }
--- a/test/Microsoft.AspNetCore.Mvc.TestDiagnosticListener.Sources/project.json
+++ b/test/Microsoft.AspNetCore.Mvc.TestDiagnosticListener.Sources/project.json
@@ -3,10 +3,13 @@
   "shared": "**/*.cs",
   "frameworks": {
     "net451": {},
-    "dotnet5.4": {
+    "netstandard1.3": {
       "dependencies": {
         "System.Runtime": "4.1.0-*"
-      }
+      },
+      "imports": [
+        "dotnet5.4"
+      ]
     }
   }
 }
--- a/test/WebSites/RazorWebSite/Services/FrameworkSpecificHelper.cs
+++ b/test/WebSites/RazorWebSite/Services/FrameworkSpecificHelper.cs
@@ -9,8 +9,8 @@ namespace RazorWebSite
         {
 #if NET451 || DNX451
             return "This method is running from NET451";
-#elif DNXCORE50
-            return "This method is running from DNXCORE50";
+#elif NETSTANDARDAPP1_5
+            return "This method is running from NETSTANDARDAPP1_5";
 #endif
         }
 
@@ -21,10 +21,10 @@ namespace RazorWebSite
         }
 #endif
 
-#if DNXCORE50_CUSTOM_DEFINE
-        public string ExecuteDnxCore50Operation()
+#if NETSTANDARDAPP1_5_CUSTOM_DEFINE
+        public string ExecuteNetStandardApp1_5Operation()
         {
-            return "This method is only defined in DNXCORE50";
+            return "This method is only defined in NETSTANDARDAPP1_5";
         }
 #endif
     }
--- a/test/WebSites/RazorWebSite/project.json
+++ b/test/WebSites/RazorWebSite/project.json
@@ -23,13 +23,16 @@
         ]
       }
     },
-    "dnxcore50": {
+    "netstandardapp1.5": {
       "compilationOptions": {
         "define": [
-          "DNXCORE50_CUSTOM_DEFINE"
+          "NETSTANDARDAPP1_5_CUSTOM_DEFINE"
         ]
       },
-      "imports": "portable-net451+win8"
+      "imports": [
+        "dnxcore50",
+        "portable-net451+win8"
+      ]
     }
   }
 }
--- a/samples/CustomRouteSample.Web/project.json
+++ b/samples/CustomRouteSample.Web/project.json
@@ -13,8 +13,11 @@
   },
   "frameworks": {
     "net451": {},
-    "dnxcore50": {
-      "imports": "portable-net451+win8",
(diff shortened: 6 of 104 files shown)


NTaylorMullen:
Doing this as a verification of an interesting repo before doing this across the board.

/cc @Eilon @dougbu @ryanbrandenburg @rynowak @ajaybhargavb @pranavkm @kichalla @javiercn @sebastienros

samples/ActionConstraintSample.Web/project.json
```diff
@@ -12,9 +12,12 @@
...
   "frameworks": {
-    "net451": { },
-    "dnxcore50": {
-      "imports": "portable-net451+win8",
+    "net451": {},
+    "netstandardapp1.5": {
+      "imports": [
+        "dnxcore50",
+        "portable-net451+win8"
+      ],
```
dougbu: Could you explain what this means? Is it somewhat analogous to having this project support `net451`, `dnx451` and `dnxcore50`?

Also `dnxcore50` --- really?

src/Microsoft.AspNetCore.Mvc.ApiExplorer/project.json
```diff
@@ -8,7 +8,9 @@
   "compilationOptions": {
     "warningsAsErrors": true,
     "keyFile": "../../tools/Key.snk",
-    "nowarn": [ "CS1591" ],
+    "nowarn": [
+      "CS1591"
+    ],
```
dougbu: Curious: Is the VS editor still doing the Wrong Thing:tm: with one-element arrays?

samples/ActionConstraintSample.Web/project.json
NTaylorMullen: So the portable bits are here because there's an external dependency in the chain that doesn't directly target `dnxcore50` or `netstandardapp1.5` but does target something compatible with the `portable-net451+win8`.

As for why we have the `dnxcore50` in the imports now instead of at the top level: This is a temporary stop-gap to enable us to work with packages that haven't 100% transitioned to netstandard yet. Aka, if a package supports netstandard it'll use the netstandard target, if not it falls back to the imports.

test/Microsoft.AspNetCore.Mvc.TestDiagnosticListener.Sources/project.json
```diff
@@ -3,10 +3,13 @@
   "shared": "**/*.cs",
   "frameworks": {
     "net451": {},
-    "dotnet5.4": {
+    "netstandard1.3": {
```
dougbu: Why does this project need anything but `net451` in `frameworks`? We don't build a package for this project and ignore the compiled DLL is always ignored.

src/Microsoft.AspNetCore.Mvc.ApiExplorer/project.json
NTaylorMullen: Don't beleive so, this is newtonsoft reformatting when I spit the file back out.

samples/ActionConstraintSample.Web/project.json
dougbu: Thanks for the detail.

> there's an external dependency in the chain

Curious: What external dependency?

test/Microsoft.AspNetCore.Mvc.TestDiagnosticListener.Sources/project.json
NTaylorMullen: Maybe for a reason :smile:. Don't know. Just transitioning here.

samples/ActionConstraintSample.Web/project.json
NTaylorMullen: Probably newtonsoft

samples/ActionConstraintSample.Web/project.json
```diff
@@ -12,9 +12,12 @@
     "Microsoft.AspNetCore.Server.Kestrel": "1.0.0-*"
   },
   "frameworks": {
-    "net451": { },
-    "dnxcore50": {
-      "imports": "portable-net451+win8",
+    "net451": {},
+    "netstandardapp1.5": {
```
dougbu: `netstandardapp1.5` versus `netstandard1.3` seems a bit arbitrary. The samples are used as libraries when invoked from the functional test project. So why isn't `netstandard1.3` also listed?

Similarly all test projects are treated as libraries when invoked through `xunit.console.exe`, `dotnet test` or even `dnx test`.

Guess I'm wondering why the app versus not facet isn't implicit in `"emitEntryPoint": true` or not.

dougbu: Talked offline: Nobody knows.

test/WebSites/RazorWebSite/Services/FrameworkSpecificHelper.cs
```diff
@@ -21,10 +21,10 @@ public string ExecuteNet451Operation()
         }
 #endif
 
-#if DNXCORE50_CUSTOM_DEFINE
+#if NETSTANDARDAPP1_5_CUSTOM_DEFINE
         public string ExecuteDnxCore50Operation()
```
pranavkm: `ExecuteNetStandardAppOperation`

test/WebSites/RazorWebSite/project.json
```diff
@@ -23,13 +23,16 @@
         ]
       }
     },
-    "dnxcore50": {
+    "netstandardapp1.5": {
       "compilationOptions": {
         "define": [
           "DNXCORE50_CUSTOM_DEFINE"
```
pranavkm: `NETSTANDARD_CUSTOM_DEFINE` (doesn't really matter since it's just a token)

pranavkm:
Changes look :shipit: to me. Might want to have @davidfowl \ @anurse do a cursory glance over the changes to see if they spot some oddity.

samples/ActionConstraintSample.Web/project.json
anurse: It's because soon, `netstandardapp` will actually refer to a specific set of APIs that will be assumed to be available in the shared framework. `netstandard` refers to the standard library alone.

anurse:
Looks right to me.

test/WebSites/RazorWebSite/project.json
khellang: If `DNXCORE50_CUSTOM_DEFINE` became `NETSTANDARDAPP1_5_CUSTOM_DEFINE` (in code), shouldn't this match?

pranavkm: In which case, yes. I assumed he missed it in both places.

ciel:
This is so confusing. So now we have to update all of our framework references __again__? Is `netstandardapp1.5` the last time at least? Or is it going to change again?

anurse:
Unfortunately this is life in the preview train :(. This name is not yet final, so feel free to wait to update to avoid extra churn :).

ciel:
I honestly hope it turns out to be something less garish than "netstandardapp1.5". That just looks awful. Perhaps "netcore1.5"

ciel:
I doubt I can make a difference, especially this late into things, but
would you be kind enough to point me to the correct team/repository for
voicing my feelings on this naming convention?

anurse:
I think we're pretty aware of the concerns and we're taking them into account. As we make changes, as always, feel free to voice concerns in GitHub :). It's a very iterative process, which is what's causing all the churn right now. /cc @DamianEdwards

ciel:
I'm not trying to sound bratty or rude, I just think that we've got a chance to define some solid naming convention and ought to take it.

We're calling this new stuff "dot net core", so it seems to me that the best strategy is to go with that...

"netcore"
"netcore1.5"

etc.

anurse:
@ciel No worries, it wasn't taken as such! We're thinking along some similar lines. There are always a few extra constraints to worry about but I think we'll hit somewhere near to that.

ciel:
I don't know how difficult it would be, but have you considered adding another tier to the frameworks schema to make it easier to support newer tools and versions going forward? For example...

```
"frameworks": {
    "imports": [ ],
    "netstandard" : {
        "1.5" : {
            "dependencies": {}
        },
        "dependencies": {}
    },
    "net": {
        "451" : {
            "dependencies": {}
        },  
        "dependencies": {}
    }
}
```

this way, you could check for a specific __type__ of framework first, then if necessary narrow it down by version. I think this would make tooling in Visual Studio easier, too - since it could target specific types of framework instead of specific versions all the time. But perhaps I don't understand how the tooling in Visual Studio works.

ciel:
I'm still totally lost on this. I thought the final implementation was `netstandard1.5`? Why do the mvc samples still use `netstandardapp1.5`?

Issue #3482, opened 2015-11-05
FormFileModelBinder should support IFormFileCollection


rynowak:
Doing this:
```
        [HttpPost]
        public IActionResult Index(IFormFileCollection files)
        {
            return View();
        }
```

Gives you this:
```
System.ArgumentException: Object of type 'System.Collections.Generic.List`1[Microsoft.AspNet.Http.IFormFile]' cannot be converted to type 'Microsoft.AspNet.Http.IFormFileCollection'.
   at System.RuntimeType.TryChangeType(Object value, Binder binder, CultureInfo culture, Boolean needsSpecialCast)
   at System.RuntimeType.CheckValue(Object value, Binder binder, CultureInfo culture, BindingFlags invokeAttr)
   at System.Reflection.MethodBase.CheckArguments(Object[] parameters, Binder binder, BindingFlags invokeAttr, CultureInfo culture, Signature sig)
   at System.Reflection.RuntimeMethodInfo.InvokeArgumentsCheck(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at Microsoft.AspNet.Mvc.Controllers.ControllerActionExecutor.ExecuteAsync(MethodInfo actionMethodInfo, Object instance, Object[] orderedActionArguments)
   at Microsoft.AspNet.Mvc.Controllers.ControllerActionInvoker.<InvokeActionAsync>d__6.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd(Task task)
   at Microsoft.AspNet.Mvc.Controllers.FilterActionInvoker.<InvokeActionFilterAsync>d__48.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at Microsoft.AspNet.Mvc.Controllers.FilterActionInvoker.<InvokeAsync>d__39.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
[...]

Issue #4130, opened 2016-02-22
`CollectionModelBinder` and `GenericModelBinder` conspire to create two collection instances


dougbu:
In the normal case, `MutableObjectModelBinder` activates all properties and ensures `bindingContext.Model != null` before `GenericModelBinder` gets involved. However `bindingContext.Model == null` for top-level collection parameters and the `GenericModelBinder` ecosystem handles that case inefficiently. This is a call to measure performance for top-level collections and determine if activation is a significant drain. If activation is a problem for that scenario, we know where to look.

----

In the top-level collection case, first `GenericModelBinder` calls `collectionBinder.CanCreateInstance()` (where `collectionBinder` is always a `CollectionModelBinder` or a subclass). Unless `collectionBinder` is an `ArrayModelBinder`, this attempts to create a collection instance, compares it with `null`, and discards the instance.

Then `CollectionModelBinder` often needs to either create an empty instance (because model binding failed) or activate another instance because the intermediate `List<T>` cannot be used. This always happens when using `ArrayModelBinder` or `DictionaryModelBinder` but can easily occur with the base `CollectionModelBinder` as well.

I don't have the data available but recall @rynowak showing high costs associated with `System.Activator` use, though the parameterless constructor case isn't as bad as some.

Separately, the intermediate `List<T>` may be unavoidable for `ArrayModelBinder` but probably isn't necessary for most other collection binding scenarios. As long as `bindingContext.Model` or the activated collection implement `ICollection<TElement>`, should be fine with that instance.

Note the `GenericModelBinder` ecosystem appears correct. We should not change the visible behaviour of this part of the model binding system.

dougbu:
Partially addressed in d3c2463. `CollectionModelBinder` is now a bit more efficient. But it still creates an intermediate collection when property value could be used.

PR #4172, opened 2016-02-26
Correct `Type.IsAssignableFrom()` polarity

dougbu:
- #3482
- see new tests; many failed without fixes in the product code
 - `ModelBindingHelper.ConvertValuesToCollectionType<T>()` -> `CreateCompatibleCollection<T>()`
- make `FormFileModelBinder` / `HeaderModelBinder` collection handling consistent w/ `GenericModelBinder`++
 - add checks around creating collections and leaving non-top-level collections `null` (not empty)
 - add checks around binding collections to read-only properties
- add support for binding `IFormFileCollection` properties

nit: Correct a few existing test names since nothing is not the same as `ModelBindingResult.Failed()`

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CollectionModelBinder.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CollectionModelBinder.cs
@@ -7,9 +7,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
-#if NETSTANDARD1_3
 using System.Reflection;
-#endif
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Mvc.Internal;
 using Microsoft.AspNetCore.Mvc.ModelBinding.Validation;
@@ -45,7 +43,6 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
                     }
 
                     bindingContext.Result = ModelBindingResult.Success(bindingContext.ModelName, model);
-                    return;
                 }
 
                 return;
@@ -85,7 +82,7 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
 
             if (valueProviderResult != ValueProviderResult.None)
             {
-                // If we did simple binding, then modelstate should be updated to reflect what we bound for ModelName. 
+                // If we did simple binding, then modelstate should be updated to reflect what we bound for ModelName.
                 // If we did complex binding, there will already be an entry for each index.
                 bindingContext.ModelState.SetModelValue(
                     bindingContext.ModelName,
@@ -93,13 +90,20 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
             }
 
             bindingContext.Result = ModelBindingResult.Success(bindingContext.ModelName, model);
-            return;
         }
 
         /// <inheritdoc />
         public virtual bool CanCreateInstance(Type targetType)
         {
-            return CreateEmptyCollection(targetType) != null;
+            if (targetType.IsAssignableFrom(typeof(List<TElement>)))
+            {
+                // Simple case such as ICollection<TElement>, IEnumerable<TElement> and IList<TElement>.
+                return true;
+            }
+
+            return targetType.GetTypeInfo().IsClass &&
+                !targetType.GetTypeInfo().IsAbstract &&
+                typeof(ICollection<TElement>).IsAssignableFrom(targetType);
         }
 
         /// <summary>
@@ -126,15 +130,7 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
         /// <returns>An instance of <paramref name="targetType"/>.</returns>
         protected object CreateInstance(Type targetType)
         {
-            try
-            {
-                return Activator.CreateInstance(targetType);
-            }
-            catch (Exception)
-            {
-                // Details of exception are not important.
-                return null;
-            }
+            return Activator.CreateInstance(targetType);
         }
 
         // Used when the ValueProvider contains the collection to be bound as a single element, e.g. the raw value
--- a/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
@@ -113,10 +113,7 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
                 return collection.ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
             }
 
-            var newCollection = CreateInstance(targetType);
-            CopyToModel(newCollection, collection);
-
-            return newCollection;
+            return base.ConvertToCollectionType(targetType, collection);
         }
 
         /// <inheritdoc />
@@ -128,7 +125,18 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
                 return new Dictionary<TKey, TValue>();
             }
 
-            return CreateInstance(targetType);
+            return base.CreateEmptyCollection(targetType);
+        }
+
+        public override bool CanCreateInstance(Type targetType)
+        {
+            if (targetType.IsAssignableFrom(typeof(Dictionary<TKey, TValue>)))
+            {
+                // Simple case such as IDictionary<TKey, TValue>.
+                return true;
+            }
+
+            return base.CanCreateInstance(targetType);
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/FormFileModelBinder.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/FormFileModelBinder.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.Diagnostics;
 using System.Linq;
 #if NETSTANDARD1_3
@@ -12,7 +13,6 @@ using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc.Internal;
 using Microsoft.AspNetCore.Mvc.ModelBinding.Validation;
-using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNetCore.Mvc.ModelBinding
 {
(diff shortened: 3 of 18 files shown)


src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/FormFileModelBinder.cs
```diff
@@ -36,66 +36,151 @@ public Task BindModelAsync(ModelBindingContext bindingContext)
...
+                    // Note this cast may fail if the runtime model implements IEnumerable<IFormFile> but not
+                    // ICollection<IFormFile>. Give up in then: Assuming we're not in an odd corner case where the
+                    // property is settable and its declared type is actually assignable from List<IFormFile>.
+                    collection = bindingContext.Model as ICollection<IFormFile>;
+                    collection?.Clear();
+                }
             }
 
-            if (value == null)
+            return collection;
```
rynowak: Is all this stuff really necessary? This just seems so far from simple. At this point you're already using `IFormFile` so you're already coding against our apis. I understand why the code is complicated, just wondering if we need it.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/FormFileModelBinder.cs
```diff
@@ -36,66 +36,151 @@ public Task BindModelAsync(ModelBindingContext bindingContext)
...
+                        // Special-case creating a custom IFormFileCollection if the property is settable and
+                        // currently null. Use a List<IFormFile> for now. Will create a FileCollection later.
+                        collection = new List<IFormFile>();
+                    }
+                    else
+                    {
+                        // Note this call may return null if the model type cannot be activated.
+                        collection = ModelBindingHelper.CreateCompatibleCollection<IFormFile>(
+                            bindingContext.ModelType,
+                            capacity: null);
```
rynowak: Should we bake more of this knowledge into this utility method?

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBindingHelper.cs
```diff
@@ -778,46 +778,51 @@ internal static TModel CastOrDefault<TModel>(object model)
...
-                var result = (ICollection<T>)Activator.CreateInstance(modelType);
-                foreach (var value in values)
+                try
                 {
-                    result.Add(value);
+                    return (ICollection<T>)Activator.CreateInstance(modelType);
+                }
+                catch
+                {
+                    // Ignore any errors attempting to create the model.
```
rynowak: Why would we ignore these errors? This intentionally propagated the exception before

rynowak:
:watch: I don't really have a gripe with the code changes or tests here.... but this just seems like something that becomes much harder to explain to anyone how it works and why.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/FormFileModelBinder.cs
dougbu: Couple of thoughts
- The read-only `ICollection<T>` stuff is definitely something I'd like to see supported, consistent w/ other collection properties in the system.
- Less sure about the `try` block around activating the user's type. But that's fairly localized and not repeated.
- Could simplify things a bit if we were more consistent about read-only scalars. The code here and in `CollectionModelBinder` assumes `MutableObjectModelBinder` or `TryUpdateModelAsync()` has already excluded the read-only scalar case. But it complicates the code to avoid checking `IsReadOnly` for the scalar case. Bottom line we could centralize some of this in a `Helper<T>` that checks
```c#
bindingContext.ModelMetadata.IsReadOnly &&
  (bindingContext.Model == null && !typeof(ICollection<T>).IsAssignableFrom(bindingContext.ModelType) ||
   bindingContext.ModelType = typeof(T[]))
```
- `ModelBindingHelper.CreateCompatibleCollection<T>()` could take a `ModelBindingContext` parameter instead of passing `bindingCotnext.ModelType`. With the above out of the way, this helper could handle the bare `T` and `T[]` cases.
- That would simplify this block to something like
``` c#
if (CanReuseOrCreateModel(bindingContext))
{
    return null;
}

ICollection<IFormFile> collection;
if (bindingContext.Model == null && bindingContext.ModelType == typeof(IFormFileCollection))
{
    collection = new List<IFormFile>();
}
else
{
    collection = CreateCompatibleCollection<IFormFile>(ModelType, capacity: null);
}
```
- Can't combine further w/o pushing the `IFormFileCollection` stuff down where it doesn't belong.

Thoughts?

rynowak: That looks much better. I'd assume the header code is pretty similar without the `if` right?

dougbu: > I'd assume the header code is pretty similar without the `if` right?

Actually will have a similar layout except the `if` condition will be `bindingContext.ModelType == typeof(string)` because no collection, intermediate or otherwise, is needed in that case.

I'm going to try a few things out and see what feels best, keeping `GenericModelBinder` and friends in mind...

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBindingHelper.cs
dougbu: Intentional or not, `Exception`s while activating are handled inconsistently within the model binding system. Many are caught, especially in `CollectionModelBinder` and `SimpleTypeModelBinder` but some propagate e.g. when creating the target binder in `GenericModelBinder`. This change at least makes `FormFileModelBinder` and `HeaderModelBinder` consistent with `CollectionModelBinder`.

Separately we can certainly discuss whether e.g. `MutableObjectModelBinder` should allow activation `Exception`s to propagate given another binder could handle the POCO type. My current vote would be that the model binding system should _always_ let every applicable model binder run regardless of issues some binder hits.

dougbu:
:up::date:

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBindingHelper.cs
rynowak: > Intentional or not, Exceptions while activating are handled inconsistently within the model binding system

I'm not 100% convinced that's bad. There's a big spectrum here, and while consistency is nice, it's secondary to providing the best user experience -- I realise that's a bit of a canard as well - "best user experience" is 100% subject to debate and discussion.

> This change at least makes FormFileModelBinder and HeaderModelBinder consistent with CollectionModelBinder.

And I'm not sure that's a good thing. The binder for collections is a "default" binder - it handles it if nothing more specialized jumped in. The binders for header and form handle specific cases that you opt in to! I think this needs more discussion if the end result is a `try/catch`.

Additionally, neither of these binders even handled this case before, both of them were very constrained in what they do, so I don't buy any arguments about consistency. This PR fundamentally changes the policy of all of them.

dougbu: > end result is a `try/catch/eat`

The end result is the model binding system continues with the rest of the current model as well as other action parameters. The behaviour without this PR is a HTTP/50x response containing no information to help the end user (whoever wrote the model).

Note I wrote these helpers with use in `CollectionModelBinder` and #4130 in mind as well, avoiding the need to create an instance when `GenericModelBinder` chooses a binder. But even adding that use case to the mix, I'll grant we're talking about a strange corner case involving an `ICollection<T>` implementation that lacks a parameterless constructor or otherwise can't be activated.

Would logging be satisfactory?

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/FormFileModelBinder.cs
```diff
@@ -33,69 +33,115 @@ public Task BindModelAsync(ModelBindingContext bindingContext)
             // using Task.FromResult. If you need to make changes of this nature, profile
             // allocations afterwards and look for Task<ModelBindingResult>.
```
rynowak: This comment is stale and can be removed. Bonus #RyanPoints

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/FormFileModelBinder.cs
```diff
@@ -33,69 +33,115 @@ public Task BindModelAsync(ModelBindingContext bindingContext)
...
+            }
+
+            ICollection<IFormFile> postedFiles;
+            if (createFileCollection)
+            {
+                postedFiles = new List<IFormFile>();
+            }
+            else
+            {
+                postedFiles = ModelBindingHelper.GetCompatibleCollection<IFormFile>(bindingContext, capacity: null);
```
rynowak: Should we add another overload?

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/HeaderModelBinder.cs
```diff
@@ -41,34 +39,37 @@ public Task BindModelAsync(ModelBindingContext bindingContext)
...
-            if (bindingContext.ModelType == typeof(string))
+
+            object model;
+            if (ModelBindingHelper.CanGetCompatibleCollection<string>(bindingContext))
             {
-                string value = request.Headers[headerName];
-                if (value != null)
+                if (bindingContext.ModelType == typeof(string))
                 {
+                    string value = request.Headers[headerName];
```
rynowak: Can you do var and cast here? That might make the conversion more obvious.

rynowak:
:shipit: after removing the try/catch

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBindingHelper.cs
dougbu: Decided offline to remove the `try` / `catch` / `eat` here and to file a bug about the pattern in `CollectionModelBinder`.

dougbu: See #4193. Will fix that in next iteration as well.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/FormFileModelBinder.cs
dougbu: Didn't update all of these but did touch the couple in files already part of this PR. Please confirm new wording (there's still _an_ optimization) in next iteration.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/FormFileModelBinder.cs
dougbu: done

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/HeaderModelBinder.cs
dougbu: done

dougbu:
:up::date:

rynowak:
:shipit: again!

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/CollectionModelBinder.cs
```diff
@@ -93,13 +90,20 @@ public class CollectionModelBinder<TElement> : ICollectionModelBinder
...
 
             bindingContext.Result = ModelBindingResult.Success(bindingContext.ModelName, model);
-            return;
         }
 
         /// <inheritdoc />
         public virtual bool CanCreateInstance(Type targetType)
         {
-            return CreateEmptyCollection(targetType) != null;
+            if (targetType.IsAssignableFrom(typeof(List<TElement>)))
```
dougbu: Should result in slightly better performance because it's no longer creating an instance to check if it can create an instance. Touches on #4181 as well.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/DictionaryModelBinder.cs
```diff
@@ -113,10 +113,7 @@ public class DictionaryModelBinder<TKey, TValue> : CollectionModelBinder<KeyValu
                 return collection.ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
             }
 
-            var newCollection = CreateInstance(targetType);
-            CopyToModel(newCollection, collection);
-
-            return newCollection;
+            return base.ConvertToCollectionType(targetType, collection);
```
dougbu: Just avoiding some code duplication. Has the slight side effect of allowing this binder to handle `IList<KeyValuePair<K,V>>` but that's not a big deal either way.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBindingHelper.cs
```diff
@@ -748,46 +748,146 @@ internal static TModel CastOrDefault<TModel>(object model)
...
+        /// non-<c>null</c> value.
+        /// </summary>
+        /// <typeparam name="T">The element type of the <see cref="ICollection{T}"/> required.</typeparam>
+        /// <param name="bindingContext">The <see cref="ModelBindingContext"/>.</param>
+        /// <returns>
+        /// <c>true</c> if <see cref="M:GetCompatibleCollection{T}"/> is likely to return a usable non-<c>null</c>
+        /// value; <c>false</c> otherwise.
+        /// </returns>
+        /// <remarks>"Usable" in this context means the property can be set or its value reused.</remarks>
+        public static bool CanGetCompatibleCollection<T>(ModelBindingContext bindingContext)
```
dougbu: @rynowak you might want to look at this code again or at the most recent commit. Simplified slightly and reduced block nesting.

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/DictionaryModelBinderTest.cs
```diff
@@ -388,9 +388,8 @@ public class DictionaryModelBinderTest
                     { typeof(IDictionary<int, int>), true },
                     { typeof(Dictionary<int, int>), true },
                     { typeof(SortedDictionary<int, int>), true },
-                    { typeof(IList<KeyValuePair<int, int>>), false },
-                    { typeof(DictionaryWithInternalConstructor<int, int>), false },
-                    { typeof(DictionaryWithThrowingConstructor<int, int>), false },
+                    { typeof(IList<KeyValuePair<int, int>>), true },
```
dougbu: Like I said, now supports `IList<KeyValuePair<int, int>>` though that wasn't a goal (or a downside).

dougbu:
d3c2463

PR #4173, opened 2016-02-26
Simplify `IsAssignableFrom()` use

dougbu:
- standardize on the `Type` extension method; less verbiage
- `ModelMetadata` had a redundant `IsAssignableFrom()` call
- `ModelBindingHelper.ValidateBindingContext()` over-engineered and used just once
 - do useful bit inline in `KeyValuePairModelBinder`
- found while investigating #3482 but not part of a fix for that issue

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelMetadata.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelMetadata.cs
@@ -406,7 +406,7 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
                 var enumerableType = ClosedGenericMatcher.ExtractGenericInterface(ModelType, typeof(IEnumerable<>));
                 ElementType = enumerableType?.GenericTypeArguments[0];
 
-                if (ElementType == null && typeof(IEnumerable).IsAssignableFrom(ModelType))
+                if (ElementType == null)
                 {
                     // ModelType implements IEnumerable but not IEnumerable<T>.
                     ElementType = typeof(object);
--- a/src/Microsoft.AspNetCore.Mvc.ApiExplorer/DefaultApiDescriptionProvider.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ApiExplorer/DefaultApiDescriptionProvider.cs
@@ -4,10 +4,11 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+#if DOTNET5_4
 using System.Reflection;
+#endif
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Mvc.Abstractions;
-using Microsoft.AspNetCore.Mvc.ActionConstraints;
 using Microsoft.AspNetCore.Mvc.Controllers;
 using Microsoft.AspNetCore.Mvc.Formatters;
 using Microsoft.AspNetCore.Mvc.Internal;
@@ -135,9 +136,9 @@ namespace Microsoft.AspNetCore.Mvc.ApiExplorer
                     apiDescription.SupportedResponseFormats.Add(format);
                 }
             }
-            
+
             // It would be possible here to configure an action with multiple body parameters, in which case you
-            // could end up with duplicate data. 
+            // could end up with duplicate data.
             foreach (var parameter in apiDescription.ParameterDescriptions.Where(p => p.Source == BindingSource.Body))
             {
                 var formats = GetRequestFormats(action, requestMetadataAttributes, parameter.Type);
@@ -427,7 +428,7 @@ namespace Microsoft.AspNetCore.Mvc.ApiExplorer
 
             // If the method is declared to return IActionResult or a derived class, that information
             // isn't valuable to the formatter.
-            if (typeof(IActionResult).GetTypeInfo().IsAssignableFrom(unwrappedType.GetTypeInfo()))
+            if (typeof(IActionResult).IsAssignableFrom(unwrappedType))
             {
                 return null;
             }
@@ -577,7 +578,7 @@ namespace Microsoft.AspNetCore.Mvc.ApiExplorer
 
             public ParameterDescriptor Parameter { get; }
 
-            // Avoid infinite recursion by tracking properties. 
+            // Avoid infinite recursion by tracking properties.
             private HashSet<PropertyKey> Visited { get; }
 
             public void WalkParameter(ApiParameterDescriptionContext context)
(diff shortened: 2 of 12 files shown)


src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelMetadata.cs
```diff
@@ -406,7 +406,7 @@ private void InitializeTypeInformation()
                 var enumerableType = ClosedGenericMatcher.ExtractGenericInterface(ModelType, typeof(IEnumerable<>));
                 ElementType = enumerableType?.GenericTypeArguments[0];
 
-                if (ElementType == null && typeof(IEnumerable).IsAssignableFrom(ModelType))
+                if (ElementType == null)
```
dougbu: See [line 393](https://github.com/aspnet/Mvc/pull/4173/files#diff-438d1905c3cf0d136b093a030abe0d41R393). That guarantees the second part of the former condition is `true`.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBindingHelper.cs
```diff
@@ -744,35 +743,6 @@ internal static void ValidateBindingContext(ModelBindingContext bindingContext)
...
-        {
-            ValidateBindingContext(bindingContext);
-
-            if (bindingContext.ModelType != requiredType)
-            {
-                var message = Resources.FormatModelBinderUtil_ModelTypeIsWrong(bindingContext.ModelType, requiredType);
-                throw new ArgumentException(message, nameof(bindingContext));
-            }
-
-            if (!allowNullModel && bindingContext.Model == null)
```
dougbu: `KeyValuePairModelBinder` passed `allowNullModel: true`

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBindingHelper.cs
```diff
@@ -744,35 +743,6 @@ internal static void ValidateBindingContext(ModelBindingContext bindingContext)
...
-            {
-                var message = Resources.FormatModelBinderUtil_ModelCannotBeNull(requiredType);
-                throw new ArgumentException(message, nameof(bindingContext));
-            }
-
-            if (bindingContext.Model != null &&
-                !bindingContext.ModelType.GetTypeInfo().IsAssignableFrom(requiredType.GetTypeInfo()))
```
dougbu: `!bindingContext.ModelType.GetTypeInfo().IsAssignableFrom(requiredType.GetTypeInfo())` is always `false` because `bindingContext.ModelType == requiredType` to reach here. (Yeah, this method only dealt correctly with `sealed` types anyhow.)

dougbu:
@rynowak (though anyone could review this cleanup) you should be happy to see I'm killing `ModelBindingHelper.ValidateBindingContext()` to death.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/KeyValuePairModelBinder.cs
```diff
@@ -17,9 +17,13 @@ public sealed class KeyValuePairModelBinder<TKey, TValue> : IModelBinder
...
-            ModelBindingHelper.ValidateBindingContext(bindingContext,
-                                                      typeof(KeyValuePair<TKey, TValue>),
-                                                      allowNullModel: true);
+            if (bindingContext.ModelType != typeof(KeyValuePair<TKey, TValue>))
+            {
+                var message = Resources.FormatModelBinderUtil_ModelTypeIsWrong(
+                    bindingContext.ModelType,
+                    typeof(KeyValuePair<TKey, TValue>));
+                throw new ArgumentException(message, nameof(bindingContext));
+            }
```
rynowak: Would it be all that harmful to just have this return instead of throwing? That's more inline with how model binders actually work. Then we could kill the badly named resource.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBindingHelper.cs
```diff
@@ -744,35 +743,6 @@ internal static void ValidateBindingContext(ModelBindingContext bindingContext)
             }
         }
 
-        internal static void ValidateBindingContext(
```
rynowak: :clap:

src/Microsoft.AspNetCore.Mvc.ApiExplorer/DefaultApiDescriptionProvider.cs
```diff
@@ -4,10 +4,11 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+#if DOTNET5_4
```
pranavkm: But you won't get free remove and sort usings anymore :crying_cat_face:

rynowak:
:shipit:

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/KeyValuePairModelBinder.cs
dougbu: Hmm, depends on whether we think anything but `GenericModelBinder` will ever interact with this class. If this is only a `GenericModelBinder` toy, `Debug.Assert()` would be appropriate. Otherwise returning might make sense.

But I do agree `throw`ing seems un-model binder-ish.

dougbu: @rynowak I lean toward `Debug.Assert()`. Your thoughts?

src/Microsoft.AspNetCore.Mvc.ApiExplorer/DefaultApiDescriptionProvider.cs
dougbu: Actually you still do. What this does is avoid remove and sort thinking `System.Reflection` isn't needed when the file is open in full .NET mode.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/KeyValuePairModelBinder.cs
rynowak: Debug.Assert isn't really correct either if it's a public API.

dougbu: Will go with just returning. Also filed #4181 for other oddities related to behaving like these types are `public`.

dougbu:
f889965

src/Microsoft.AspNetCore.Mvc.TagHelpers/SelectTagHelper.cs
```diff
@@ -5,7 +5,9 @@
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+#if DOTNET5_6
 using System.Reflection;
```
davidfowl: what?

davidfowl:
Why would we add more ifdefs?

PR #4174, opened 2016-02-26
Test fix: Change `CollectionModelBinderTest` to update `ModelMetadata.IsReadOnly`

dougbu:
- unrelated to #3482 except that I discovered the issue while investigating that issue
- tests previously set `BindingDetails.IsReadOnly` for a `Type` and that was ignored
 - same for `DictionaryModelBinderTest`


rynowak:
:shipit: - seems like a good change

dougbu:
49ffeb1

PR #4180, opened 2016-02-27
Implement MoveTo semantics in Mvc

rynowak:
- Simplify things that used to rely on HtmlTextWriter. This behavior is
  now the default.

- Simplify a whole mess of Razor TextWriter code.

- Integration test for merging of TagHelper buffers back into the 'main'
  buffer.


Mvc.NoFun.sln
```diff
@@ -1,7 +1,6 @@
-
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.24720.0
+VisualStudioVersion = 14.0.24711.0
```
rynowak: Ignore. I'll undo later

global.json
```diff
@@ -1,3 +1,3 @@
 {
-    "projects": ["src", "test/WebSites", "samples"]
+    "projects": ["src", "test/WebSites", "samples", "d:\\k\\HtmlAbstractions\\src", "d:\\k\\Razor\\src"]
```
rynowak: Ignore. I'll undo later

src/Microsoft.AspNetCore.Mvc.Localization/LocalizedHtmlString.cs
```diff
@@ -100,15 +100,7 @@ public void WriteTo(TextWriter writer, HtmlEncoder encoder)
...
 
-            var htmlTextWriter = writer as HtmlTextWriter;
-            if (htmlTextWriter == null)
-            {
-                FormatValue(writer, encoder, Value, _arguments);
-            }
-            else
-            {
-                htmlTextWriter.Write(this);
-            }
```
rynowak: This kind of thing is no longer needed. You now get it for free.

src/Microsoft.AspNetCore.Mvc.Razor/RazorPage.cs
```diff
@@ -223,9 +223,7 @@ public void StartTagHelperWritingScope(HtmlEncoder encoder)
 
             // We need to replace the ViewContext's Writer to ensure that all content (including content written
             // from HTML helpers) is redirected.
-            var buffer = new ViewBuffer(BufferScope, Path);
-            var writer = new HtmlContentWrapperTextWriter(buffer, ViewContext.Writer.Encoding);
-            ViewContext.Writer = writer;
+            ViewContext.Writer = new ViewBufferTextWriter(buffer, ViewContext.Writer.Encoding);
```
rynowak: Just simplification here. There now one class for this writer instead of two.

src/Microsoft.AspNetCore.Mvc.Razor/RazorPage.cs
```diff
@@ -316,18 +314,7 @@ public static void WriteTo(TextWriter writer, HtmlEncoder encoder, object value)
...
-                {
-                    htmlContent.WriteTo(writer, encoder);
-                }
-                else
-                {
-                    // This special case allows us to keep buffering as IHtmlContent until we get to the 'final'
-                    // TextWriter.
-                    htmlTextWriter.Write(htmlContent);
-                }
-
```
rynowak: Considering bringing back `HtmlTextWriter` just to satisfy this special case. The other option is to always wrap the given writer in `ViewBufferTextWriter` inside Razor if it isn't one already.

src/Microsoft.AspNetCore.Mvc.Razor/RazorPage.cs
```diff
@@ -353,7 +340,10 @@ private static void WriteTo(TextWriter writer, HtmlEncoder encoder, string value
         {
             if (!string.IsNullOrEmpty(value))
             {
-                encoder.Encode(writer, value);
+                // Perf: Encode write away instead of writing it character-by-character.
+                // character-by-character isn't efficient when using a writer backed by a ViewBuffer.
+                var encoded = encoder.Encode(value);
+                writer.Write(encoded);
```
rynowak: This was a MAJOR win for the DJ site.

rynowak: We could potentially also tie this to ViewBuffer and do something tricky when it's the same HTML encoder.

src/Microsoft.AspNetCore.Mvc.Razor/RazorPage.cs
```diff
@@ -566,10 +555,11 @@ public virtual void EndWriteAttributeTo(TextWriter writer)
 
             if (value != null)
             {
+                // Perf: We'll use this buffer for all of the attribute values and then clear it to
+                // reduce allocations.
                 if (_valueBuffer == null)
                 {
-                    var buffer = new ViewBuffer(BufferScope, Path);
-                    _valueBuffer = new HtmlContentWrapperTextWriter(buffer, Output.Encoding);
```
rynowak: Creating a buffer for this small of content was not good. We just use one stringwriter per page now.

src/Microsoft.AspNetCore.Mvc.Razor/RazorView.cs
```diff
@@ -99,24 +99,32 @@ public string Path
...
+            var writer = context.Writer as ViewBufferTextWriter;
+            if (writer == null)
+            {
+                Debug.Assert(_bufferScope != null);
+
+                // If we get here, this is likely the top-level page (not a partial) - this means
+                // that context.Writer is wrapping the output stream. We need to buffer, so create a buffered writer.
+                var buffer = new ViewBuffer(_bufferScope, page.Path, pageSize: 256);
+                writer = new ViewBufferTextWriter(buffer, context.Writer.Encoding, _htmlEncoder, context.Writer);
+            }
```
rynowak: Another big bugfix here. If we were rendering a partial, we'd wrap a buffered writer in a buffer. This way we can reuse instead of being greedy.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/MemoryPoolViewBufferScope.cs
```diff
@@ -13,7 +13,7 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
     /// </summary>
     public class MemoryPoolViewBufferScope : IViewBufferScope, IDisposable
     {
-        public static readonly int SegmentSize = 512;
+        public static readonly int MinimumSize = 16;
```
rynowak: I significantly reduced the chunk size here and it had a large positive impact. We need to do more tuning, but the values in this PR are a good start.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/PagedBufferedTextWriter.cs
```diff
@@ -0,0 +1,211 @@
...
+using System.Buffers;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
+{
+    public class PagedBufferedTextWriter : TextWriter
```
rynowak: This is a rename. Just ignore, no new code here.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ViewBufferTextWriter.cs
```diff
@@ -1,211 +1,436 @@
...
+    /// <para>
+    /// A <see cref="TextWriter"/> that is backed by a unbuffered writer (over the Response stream) and/or a 
+    /// <see cref="ViewBuffer"/>
+    /// </para>
+    /// <para>
+    /// When <c>Flush</c> or <c>FlushAsync</c> is invoked, the writer copies all content from the buffer to
+    /// the writer and switches to writing to the unbuffered writer for all further write operations.
+    /// </para>
+    /// </summary>
     public class ViewBufferTextWriter : TextWriter
```
rynowak: Combines the best features of `RazorTextWriter` and `HtmlContentWrapperTextWriter`

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Rendering/MvcForm.cs
```diff
@@ -63,27 +70,24 @@ protected virtual void GenerateEndForm()
...
                 {
-                    if (htmlWriter == null)
-                    {
-                        content.WriteTo(writer, htmlEncoder);
-                    }
-                    else
-                    {
-                        htmlWriter.Write(content);
-                    }
+                    viewBufferWriter.Write(content);
```
rynowak: I don't love having this special case here. We could consider bringing back `HtmlTextWriter` for this purpose, or just saying Razor + ViewBufferTextWriter and joined at the hip and shrug.

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorPageTest.cs
```diff
@@ -113,7 +114,7 @@ public class RazorPageTest
             await page.ExecuteAsync();
 
             // Assert
-            var pageOutput = page.Output.ToString();
```
rynowak: This is just cleanup. This code assumed that `.Output` was always a `StringWriter`

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorPageTest.cs
```diff
@@ -224,6 +225,138 @@ public static TheoryData<HtmlEncoder> HtmlEncoderData
             await page.ExecuteAsync();
         }
 
+        // This is an integration test for ensuring that ViewBuffer segments used by
+        // TagHelpers can be merged back into the 'main' segment where possible.
+        [Fact]
+        public async Task TagHelperScopes_ViewBuffersCanCombine()
```
rynowak: These kinds of tests are ugly, but IMO worth having at least 1.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Internal/HtmlContentWrapperTextWriterTest.cs
```diff
@@ -1,186 +0,0 @@
...
-using System.Text;
-using System.Text.Encodings.Web;
-using System.Threading.Tasks;
-using Microsoft.AspNetCore.Html;
-using Microsoft.AspNetCore.Mvc.Rendering;
-using Xunit;
-
-namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
-{
-    public class HtmlContentWrapperTextWriterTest
```
rynowak: These tests were largely redundant with `RazorTestWriter`'s tests.

global.json
dougbu: How can we ignore the fact you're stuck at `K`? :trollface:

src/Microsoft.AspNetCore.Mvc.Razor/RazorPage.cs
dougbu: What's the exact invariant here? That is, in our usage, what types can `ViewContext.Writer` have e.g. just `StringWriter` and `ViewBufferTextWriter`?

src/Microsoft.AspNetCore.Mvc.Razor/RazorPage.cs
dougbu: Could also special case `value is IHtmlContentContainer && writer is IHtmlContentContainer`. Would that avoid the worst? If the performances are high in number, could revisit later -- as long as this PR leaves the correctnesses high.

src/Microsoft.AspNetCore.Mvc.Razor/RazorPage.cs
dougbu: Unfortunately both `ViewContext.Writer` and the `HtmlEncoder` are available in user code. So even if we avoid `Encode(TextWriter, string)` completely, users can shoot themselves in both feet simultaneously. I'm also not sure the `ViewBuffer` would need to detect matching encoders. How about instead overriding `Write(char)` in `ViewBufferTextWriter` to use a `bool _isUsingSetAsideBuffer` and a `StringBuilder _setAsideBuffer`, where the `_setAsideBuffer` is lazily initialized but reused.

src/Microsoft.AspNetCore.Mvc.Razor/RazorView.cs
```diff
@@ -99,24 +99,32 @@ public string Path
             }
 
             _bufferScope = context.HttpContext.RequestServices.GetRequiredService<IViewBufferScope>();
-            var bodyWriter = await RenderPageAsync(RazorPage, context, invokeViewStarts: true);
+            var bodyWriter = await RenderPageContentAsync(RazorPage, context, invokeViewStarts: true);
             await RenderLayoutAsync(context, bodyWriter);
         }
 
-        private async Task<RazorTextWriter> RenderPageAsync(
+        private async Task<ViewBufferTextWriter> RenderPageContentAsync(
```
dougbu: Why? Doesn't fit well w/ other method names in this class.

src/Microsoft.AspNetCore.Mvc.Razor/RazorPage.cs
rynowak: In our usage I think this is *always* `ViewBufferTextWriter`

src/Microsoft.AspNetCore.Mvc.Razor/RazorPage.cs
rynowak: Yeah, let me just special case `ViewBufferTextWriter` directly. I don't want to create more abstractions.

src/Microsoft.AspNetCore.Mvc.Razor/RazorPage.cs
rynowak: Yeah, that was what I had in mind, but it was just an idea. Not going to tack on more things here. :see_no_evil:

src/Microsoft.AspNetCore.Mvc.Razor/RazorView.cs
rynowak: Will undo.

src/Microsoft.AspNetCore.Mvc.Razor/RazorView.cs
```diff
@@ -99,24 +99,32 @@ public string Path
...
-            var buffer = new ViewBuffer(_bufferScope, page.Path);
-            var razorTextWriter = new RazorTextWriter(context.Writer, buffer, _htmlEncoder);
+            var writer = context.Writer as ViewBufferTextWriter;
+            if (writer == null)
+            {
+                Debug.Assert(_bufferScope != null);
+
+                // If we get here, this is likely the top-level page (not a partial) - this means
+                // that context.Writer is wrapping the output stream. We need to buffer, so create a buffered writer.
+                var buffer = new ViewBuffer(_bufferScope, page.Path, pageSize: 256);
```
dougbu: How 'bout `const`s on the `ViewBuffer` class so we can tweak these in one place?

src/Microsoft.AspNetCore.Mvc.Razor/RazorPage.cs
dougbu: Users can set `ViewContext.Writer` too. Suggest we `throw` here if the property has an unexpected type where it's important. Bill Murray would be [chanting](https://www.youtube.com/watch?v=e9mf3Bypyk8) that "It Just Doesn't Matter" in a lot of this code but there's a few places where we shouldn't make the wrong assumption.

dougbu: Oooops meant to remove "here"...

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/IViewBufferScope.cs
```diff
@@ -12,9 +12,9 @@ public interface IViewBufferScope
     {
         /// <summary>
         /// Gets a <see cref="T:ViewBufferValue[]"/>.
-        /// </summary>
-        /// <returns>The <see cref="T:ViewBufferValue[]"/>.</returns>
-        ViewBufferValue[] GetSegment();
+        /// <param name="size">The minimum size of the returned segment.</param>
+        /// <returns>The <see cref="ViewBufferValue[]"/>.</returns>
```
ajaybhargavb: `T:ViewBufferValue[]`. Did this not fail during compilation?

rynowak: I don't build things. Looks like my merge undid your change. I'll put it back.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/MemoryPoolViewBufferScope.cs
```diff
@@ -27,7 +27,7 @@ public class MemoryPoolViewBufferScope : IViewBufferScope, IDisposable
         /// The <see cref="ArrayPool{ViewBufferValue}"/> for creating <see cref="ViewBufferValue"/> instances.
         /// </param>
         /// <param name="charPool">
-        /// The <see cref="ArrayPool{Char}"/> for creating <see cref="ViewBufferTextWriter"/> instances.
+        /// The <see cref="ArrayPool{char}"/> for creating <see cref="PagedBufferedTextWriter"/> instances.
```
ajaybhargavb: `ArrayPool{Char}`. Undo this as well.

src/Microsoft.AspNetCore.Mvc.Razor/RazorView.cs
rynowak: Fixed

src/Microsoft.AspNetCore.Mvc.Razor/RazorPage.cs
```diff
@@ -354,7 +353,10 @@ private static void WriteTo(TextWriter writer, HtmlEncoder encoder, string value
         {
             if (!string.IsNullOrEmpty(value))
             {
-                encoder.Encode(writer, value);
+                // Perf: Encode write away instead of writing it character-by-character.
```
dougbu: write -> right

global.json
```diff
@@ -1,3 +1,3 @@
 {
```
dougbu: Don't check in your `UpgradeLog.htm` file.

rynowak: Removed

src/Microsoft.AspNetCore.Mvc.Razor/RazorPage.cs
dougbu: :+1:

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/IViewBufferScope.cs
```diff
@@ -12,9 +12,9 @@ public interface IViewBufferScope
     {
         /// <summary>
         /// Gets a <see cref="T:ViewBufferValue[]"/>.
-        /// </summary>
+        /// <param name="size">The minimum size of the returned segment.</param>
```
dougbu: - Restore `</summary>`.
- Why does the `size` description say "returned" for a "Get" method?
- Should this parameter be named `pageSize`?

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/MemoryPoolViewBufferScope.cs
```diff
@@ -36,8 +36,13 @@ public MemoryPoolViewBufferScope(ArrayPool<ViewBufferValue> viewBufferPool, Arra
         }
 
         /// <inheritdoc />
-        public ViewBufferValue[] GetSegment()
+        public ViewBufferValue[] GetSegment(int size)
         {
+            if (size <= 0)
```
dougbu: This class is `.Internal`. Shouldn't this be a `Debug.Assert()`?

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ViewBufferTextWriter.cs
```diff
@@ -1,211 +1,436 @@
...
         {
-            if (_pages.Count == 0)
+            if (buffer == null)
             {
-                return;
+                throw new ArgumentNullException(nameof(buffer));
             }
 
-            for (var i = 0; i <= _currentPage; i++)
+            if (index < 0)
```
ajaybhargavb: `if (index < 0 || index >= buffer.Length)`?

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/MemoryPoolViewBufferScope.cs
```diff
@@ -60,7 +65,7 @@ public ViewBufferValue[] GetSegment()
 
             try
             {
-                segment = _viewBufferPool.Rent(SegmentSize);
+                segment = _viewBufferPool.Rent(Math.Max(size, MinimumSize));
```
dougbu: Given all the callers are also `.Internal`, do we really need to worry about incorrect calls to this method?

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/MemoryPoolViewBufferScope.cs
rynowak: Public method in a public class. Debug.Assert is for validating assumptions.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ViewBuffer.cs
```diff
@@ -254,5 +210,139 @@ public void WriteTo(TextWriter writer, HtmlEncoder encoder)
         }
 
         private string DebuggerToString() => _name;
+
+        public void CopyTo(IHtmlContentBuilder destination)
```
dougbu: `null`-check `destination` here and in `MoveTo()`

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/MemoryPoolViewBufferScope.cs
rynowak: What's the suggestion? I'm not going to make the code less correct than it already is because of `.Internal`

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/IViewBufferScope.cs
rynowak: I fixed the docs. Will rename this method and use pageSize.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ViewBuffer.cs
```diff
@@ -254,5 +210,139 @@ public void WriteTo(TextWriter writer, HtmlEncoder encoder)
...
+                    {
+                        destination.AppendHtml((IHtmlContent)value.Value);
+                    }
+                }
+            }
+
+            for (var i = 0; i < Pages.Count; i++)
+            {
+                var page = Pages[i];
+                Array.Clear(page.Buffer, 0, page.Capacity);
```
dougbu: Why bother clearing the buffer?

rynowak: It can contain arbitrary objects. If this page is returned to the pool, they could hang out for an arbitrary amount of time. Basically I'm being extra careful

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ViewBuffer.cs
```diff
@@ -254,5 +210,139 @@ public void WriteTo(TextWriter writer, HtmlEncoder encoder)
...
+                        length: page.Count);
+
+                    destinationPage.Count += page.Count;
+
+                    // Now we can return this page, and it can be reused in the scope of this request.
+                    _bufferScope.ReturnSegment(page.Buffer);
+                }
+                else
+                {
+                    // Otherwise, let's just take the the page from the other buffer.
```
dougbu: Didn't know you liked the The. But regardless, what does this comment mean?

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ViewBuffer.cs
```diff
@@ -254,5 +210,139 @@ public void WriteTo(TextWriter writer, HtmlEncoder encoder)
...
+
+        private void MoveTo(ViewBuffer destination)
+        {
+            for (var i = 0; i < Pages.Count; i++)
+            {
+                var page = Pages[i];
+
+                var destinationPage = destination.Pages.Count == 0 ? null : destination.Pages[destination.Pages.Count - 1];
+
+                // If the other page is less or equal to than half full, let's copy it's to the current page if
```
dougbu: Huh? "to than" ... "it's" [what] ...

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ViewBuffer.cs
rynowak: Yeah, comment needs an update, this code was cut/pasted from `WriteTo`

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ViewBuffer.cs
```diff
@@ -254,5 +210,139 @@ public void WriteTo(TextWriter writer, HtmlEncoder encoder)
...
+                // possible.
+                var isLessThanHalfFull = 2 * page.Count <= page.Capacity;
+                if (isLessThanHalfFull &&
+                    destinationPage != null &&
+                    destinationPage.Capacity - destinationPage.Count >= page.Count)
```
dougbu: Could you clarify why `isLessThanHalfFull` matters as long as there's room in `destinationPage` for `page`'s content?

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ViewBuffer.cs
dougbu: Note `isLessThanHalfFull` is about the page from _this_ `ViewBuffer`, not "the other page".

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ViewBuffer.cs
rynowak: It's to avoid copying lots of items when it could just be a simple link-list append. Not new code.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ViewBuffer.cs
```diff
@@ -254,5 +210,139 @@ public void WriteTo(TextWriter writer, HtmlEncoder encoder)
...
+                        sourceArray: page.Buffer,
+                        sourceIndex: 0,
+                        destinationArray: destinationPage.Buffer,
+                        destinationIndex: destinationPage.Count,
+                        length: page.Count);
+
+                    destinationPage.Count += page.Count;
+
+                    // Now we can return this page, and it can be reused in the scope of this request.
+                    _bufferScope.ReturnSegment(page.Buffer);
```
dougbu: Need `Array.Clear()` here for same reason as you've got one in the other `MoveTo()` method.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ViewBuffer.cs
```diff
@@ -254,5 +210,139 @@ public void WriteTo(TextWriter writer, HtmlEncoder encoder)
...
+            private readonly string _unencoded;
+
+            public EncodingWrapper(string unencoded)
+            {
+                _unencoded = unencoded;
+            }
+
+            public void WriteTo(TextWriter writer, HtmlEncoder encoder)
+            {
+                encoder.Encode(writer, _unencoded);
```
dougbu: Won't this fall into the `char`-by-`char` pit?

rynowak: char-by-char is **good** on a real textwriter.

rynowak: And WriteTo is only ever called on a real textwriter

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ViewBufferTextWriter.cs
```diff
@@ -1,211 +1,436 @@
...
+        public bool IsBuffering { get; private set; } = true;
+
+        /// <summary>
+        /// Gets the <see cref="ViewBuffer"/>.
+        /// </summary>
+        public ViewBuffer Buffer { get; }
+
+        private TextWriter Inner { get; }
+
+        private HtmlEncoder HtmlEncoder { get; }
```
dougbu: Suggest `readonly` fields instead.

rynowak: why?

dougbu: It's properties that need justification. Why not use fields?

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/MemoryPoolViewBufferScope.cs
dougbu: My suggestion would be to remove the `Math.Max()` call as well as `MinimumSize`. This code isn't part of a public API.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/MemoryPoolViewBufferScope.cs
```diff
@@ -36,8 +36,13 @@ public MemoryPoolViewBufferScope(ArrayPool<ViewBufferValue> viewBufferPool, Arra
         }
 
         /// <inheritdoc />
-        public ViewBufferValue[] GetSegment()
+        public ViewBufferValue[] GetPage(int size)
```
dougbu: `size` -> `pageSize` ?

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/MemoryPoolViewBufferScope.cs
```diff
@@ -36,8 +36,13 @@ public MemoryPoolViewBufferScope(ArrayPool<ViewBufferValue> viewBufferPool, Arra
...
         {
+            if (size <= 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(size));
```
dougbu: This is a `public` class in an `Internal` namespace. I don't get why anything but `Debug` assertions are needed.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/MemoryPoolViewBufferScope.cs
rynowak: There's some non-zero likelihood that we will move this to a public place next week. I don't see any reason to take this out :see_no_evil:

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ViewBufferTextWriter.cs
rynowak: Yeah you're right, I just default to doing that. Changed these two

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/MemoryPoolViewBufferScope.cs
rynowak: Still don't see any reason to take this out. :see_no_evil:

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ViewBuffer.cs
rynowak: Fixed

src/Microsoft.AspNetCore.Mvc.Razor/RazorView.cs
```diff
@@ -128,13 +136,12 @@ public string Path
...
 
                 await RenderPageCoreAsync(page, context);
-                return razorTextWriter;
+                return writer;
             }
             finally
             {
                 context.Writer = oldWriter;
                 context.ExecutingFilePath = oldFilePath;
-                razorTextWriter.Dispose();
```
dougbu: Doesn't `ViewBufferTextWriter` still implement `IDisposable`?

src/Microsoft.AspNetCore.Mvc.Razor/RazorView.cs
```diff
@@ -233,10 +240,25 @@ private Task RenderPageCoreAsync(IRazorPage page, ViewContext context)
 
             if (bodyWriter.IsBuffering)
             {
-                // Only copy buffered content to the Output if we're currently buffering.
-                using (var writer = _bufferScope.CreateWriter(context.Writer))
+                // If IsBuffering - then we've got a bunch of content in the view buffer. How to best deal with it
```
dougbu: Think I'm missing something though it's probably an existing facet of the design. The layout page was passed `bodyWriter.Buffer` and may have rendered that content before we get here. What prevents writing that information out twice?

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ViewBuffer.cs
dougbu: :ok: though I suggest `page.Capacity` -> `page.Count` since there's no reason to `.Clear()` more than `page.Count`.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ViewBuffer.cs
```diff
@@ -254,5 +230,153 @@ public void WriteTo(TextWriter writer, HtmlEncoder encoder)
...
+                        sourceArray: page.Buffer,
+                        sourceIndex: 0,
+                        destinationArray: destinationPage.Buffer,
+                        destinationIndex: destinationPage.Count,
+                        length: page.Count);
+
+                    destinationPage.Count += page.Count;
+
+                    // Now we can return the source page, and it can be reused in the scope of this request.
+                    Array.Clear(page.Buffer, 0, page.Capacity);
```
dougbu: Suggest `page.Capacity` -> `page.Count`

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ViewBufferTextWriter.cs
```diff
@@ -1,211 +1,435 @@
...
-                var page = _pages[i];
+                throw new ArgumentOutOfRangeException(nameof(index));
+            }
 
-                var count = i == _currentPage ? _currentIndex : page.Length;
-                if (count > 0)
-                {
-                    await _inner.WriteAsync(page, 0, count);
-                }
+            if (count < 0 || (index + count > buffer.Length))
```
dougbu: Could `index + count` overflow? Suggest `buffer.Length - index` < count`.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Rendering/MvcForm.cs
dougbu: :shrug:

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ViewBuffer.cs
rynowak: Sure, sounds good.

src/Microsoft.AspNetCore.Mvc.Razor/RazorView.cs
rynowak: Fixed elsewhere in code :see_no_evil:

src/Microsoft.AspNetCore.Mvc.Razor/RazorView.cs
rynowak: Added some comments that really clear this up.

src/Microsoft.AspNetCore.Mvc.Razor/RazorPage.cs
dougbu: Never mind. Couldn't find a case where the [chant](https://www.youtube.com/watch?v=e9mf3Bypyk8) isn't applicable.

src/Microsoft.AspNetCore.Mvc.Razor/RazorView.cs
dougbu: Looks good

dougbu:
:shipit:

PR #4192, opened 2016-02-29
[Fixes #4099] Add a functional test for creating a form with a file upload field

### LINKED ISSUE #4099: create a form with upload a photo
How can I create a basic model (username, password etc...) and with a photo and submit it to the controller? Please, example...

javiercn:


test/WebSites/FilesWebSite/Models/User.cs
```diff
@@ -0,0 +1,28 @@
...
+        public int Age { get; set; }
+        public IFormFile Biography { get; set; }
+
+        public string ReadBiography()
+        {
+            if (Biography != null)
+            {
+                using (var reader = new StreamReader(Biography.OpenReadStream()))
+                {
+                    return reader.ReadToEnd();
```
kichalla: nit: ReadToEndAsync ?

kichalla:
:shipit:

test/WebSites/FilesWebSite/Models/User.cs
javiercn: No need to complicate things

kichalla: I agree that this is just a test, but if our plan is to show this as a sample to our users, I think this should be `async`

javiercn: fair enough :)

ajaybhargavb:
:shipit:

test/WebSites/FilesWebSite/Models/User.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace FilesWebSite.Models
+{
+    public class User
+    {
+        public string Name { get; set; }
+        public int Age { get; set; }
+        public IFormFile Biography { get; set; }
+
+        public async Task<string> ReadBiography()
```
kichalla: nit: `ReadBiographyAsync`

Issue #3903, opened 2016-01-11
Optimize invoking action method


rynowak:
Here's a quick look at an execution profile for an MVC app rendering a pretty significant Razor View.

![image](https://cloud.githubusercontent.com/assets/1430011/12243799/4283d3a8-b855-11e5-8c8c-f573820e7e28.png)

Notice that `RuntimeMethodInfo.Invoke` is eating about 3% of execution time. We should experiment with computing a cached delegate here.

rynowak:
@Eilon @danroth27 - recommending next milestone for this one. Low pri and no public API impact.

Eilon:
@rynowak what do you define "next milestone" as? We'd like to get things done in RC2, or post-RTM. We have only a small window in between those.

rynowak:
> what do you define "next milestone" as?

I'd start by putting this below wherever your cutline is for RC2. If there's time then we can pull it back. We can definitely ship RTM without doing this and it has no public API impact.

Eilon:
That'd be the backlog :smile: We're already in the process of trimming down RC2 items, so unfortunately that means this will have to wait as well.

SajayAntony:
@rynowak do you think you want to assign this to @mnltejaswini ?

rynowak:
Sounds good.

PR #4208, opened 2016-03-02
[Perf] [Design] Optimize controller action invoke

mnltejaswini:
Fixes https://github.com/aspnet/Mvc/issues/3903 
1. Added a general purpose method executor that provides an alternative to MethodInfo.Invoke()
2. This executor creates a delegate for the method which can be cached for performance benefit.
3. The delegate is created dynamically using the Linq expressions.
2. Controller action methods are invoked using the cached executor.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionExecutor.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionExecutor.cs
@@ -35,6 +35,15 @@ namespace Microsoft.AspNetCore.Mvc.Internal
         }
 
         public static Task<object> ExecuteAsync(
+            ObjectMethodExecutor actionMethodExecutor,
+            object instance,
+            IDictionary<string, object> actionArguments)
+        {
+            var orderedArguments = PrepareArguments(actionArguments, actionMethodExecutor.MethodInfo.GetParameters());
+            return ExecuteAsync(actionMethodExecutor, instance, orderedArguments);
+        }
+
+        public static Task<object> ExecuteAsync(
             MethodInfo actionMethodInfo,
             object instance,
             object[] orderedActionArguments)
@@ -56,6 +65,30 @@ namespace Microsoft.AspNetCore.Mvc.Internal
                 actionMethodInfo.ReturnType,
                 actionMethodInfo.Name,
                 actionMethodInfo.DeclaringType);
+        }
+
+        public static Task<object> ExecuteAsync(
+            ObjectMethodExecutor actionMethodExecutor,
+            object instance,
+            object[] orderedActionArguments)
+        {
+            object invocationResult = null;
+            try
+            {
+                invocationResult = actionMethodExecutor.Execute(instance, orderedActionArguments);
+            }
+            catch (Exception actionExecutionException)
+            {
+                // Capturing the exception and the original callstack and rethrow for external exception handlers.
+                var exceptionDispatchInfo = ExceptionDispatchInfo.Capture(actionExecutionException);
+                exceptionDispatchInfo.Throw();
+            }
+
+            return CoerceResultToTaskAsync(
+                invocationResult,
+                actionMethodExecutor.MethodInfo.ReturnType,
+                actionMethodExecutor.MethodInfo.Name,
+                actionMethodExecutor.MethodInfo.DeclaringType);
         }
 
         // We need to CoerceResult as the object value returned from methodInfo.Invoke has to be cast to a Task<T>.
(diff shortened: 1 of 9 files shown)


src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionExecutor.cs
```diff
@@ -58,6 +67,30 @@ private static Task<object> Convert<T>(object taskAsObject)
...
+            try
+            {
+                invocationResult = actionMethodExecutor.Execute(instance, orderedActionArguments);
+            }
+            catch (Exception actionExecutionException)
+            {
+                // Capturing the exception and the original callstack and rethrow for external exception handlers.
+                var exceptionDispatchInfo = ExceptionDispatchInfo.Capture(actionExecutionException);
+                exceptionDispatchInfo.Throw();
+            }
```
rynowak: This try/catch shouldn't be needed for the expressions. This was only needed for the reflection case because the exception was wrapped by `TargetInvocationException`

src/Microsoft.AspNetCore.Mvc.Core/Internal/FilterCache.cs
```diff
@@ -53,7 +55,10 @@ public IFilterMetadata[] GetFilters(ActionContext actionContext)
             CacheEntry entry;
             if (cache.Entries.TryGetValue(actionDescriptor, out entry))
             {
-                return GetFiltersFromEntry(entry, actionContext);
+                if (entry.Filters != null || entry.Items != null)
```
rynowak: Rename this class to `ControllerActionInvokerCache`

rynowak: rename `entry.Items` to `entry.FilterItems`

rynowak: Rename this method to `GetCacheEntry` and return a new type.

```
public ControllerActionInvokerCacheEntry GetCacheEntry(ActionContext actionContext) { ... }

public struct ControllerActionInvokerCacheEntry
{
    public IFilterMetadata[] Filters { get; }
    public ObjectMethodExecutor Executor { get; }
}
```

src/Microsoft.AspNetCore.Mvc.Core/Internal/FilterCache.cs
```diff
@@ -86,10 +91,54 @@ public IFilterMetadata[] GetFilters(ActionContext actionContext)
                 entry = new CacheEntry(items);
             }
 
-            cache.Entries.TryAdd(actionDescriptor, entry);
+            cache.Entries.AddOrUpdate(actionDescriptor,
+                                      entry,
+                                      (key, value) =>
+                                      {
+                                          return new CacheEntry(entry.Filters, entry.Items, value.ActionMethodExecutor);
+                                      });
```
rynowak: Don't use a capturing lambda here, find another way.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
```diff
@@ -0,0 +1,100 @@
...
+        public object Execute(object target, object[] parameters)
+        {
+            return _executor(target, parameters);
+        }
+
+        private static ActionExecutor GetExecutor(MethodInfo methodInfo, TypeInfo targetTypeInfo)
+        {
+            // Parameters to executor
+            ParameterExpression targetParameter = Expression.Parameter(typeof(object), "target");
+            ParameterExpression parametersParameter = Expression.Parameter(typeof(object[]), "parameters");
```
rynowak: I know this is a design PR, but make sure to do a pass here for style. (use var, etc).

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ControllerActionExecutorTests.cs
```diff
@@ -32,24 +32,22 @@ public class ControllerActionExecutorTests
 
         private delegate dynamic ReturnTaskAsDynamicValue(int i, string s);
 
-        [Fact]
-        public async Task AsyncAction_WithVoidReturnType()
+        [Theory,
+         InlineData(true),
+         InlineData(false)]
```
rynowak: just close each attribute instead of using a comma, there's nothing like this elsewhere in the codebase.

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ControllerActionExecutorTests.cs
```diff
@@ -99,17 +92,16 @@ public class ControllerActionExecutorTests
...
 
             // Assert
             Assert.Equal(inputParam1, result);
         }
 
-        [Fact]
-        public async Task AsyncAction_WithAsyncKeywordThrows()
+        [Theory,
+         InlineData(true),
+         InlineData(false)]
```
rynowak: Instead of doing this to the tests, can we just remove the `MethodInfo` support?

rynowak:
:watch:

mnltejaswini:
AS these comments involved code changes related to ViewComponentInvoker, I ended up creating a new branch and created a new pull request https://github.com/aspnet/Mvc/pull/4242 . Hence this pull request is closed.

PR #4212, opened 2016-03-03
Added tests for DropDownList and ListBox extensions

ajaybhargavb:
Issue #453 

This is only a part of the tests. There is more to come.

@dougbu

### DIFF
--- a/test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperDropDownListExtensionsTest.cs
+++ b/test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperDropDownListExtensionsTest.cs
@@ -1,0 +1,257 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
+using Microsoft.AspNetCore.Mvc.Rendering;
+using Microsoft.AspNetCore.Mvc.TestCommon;
+using Microsoft.AspNetCore.Mvc.ViewFeatures;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Mvc.Core
+{
+    public class HtmlHelperDropDownListExtensionsTest
+    {
+        private static readonly List<SelectListItem> BasicSelectList = new List<SelectListItem>
+        {
+            new SelectListItem { Text = "Zero",  Value = "0"},
+            new SelectListItem { Text = "One",   Value = "1"},
+            new SelectListItem { Text = "Two",   Value = "2"},
+            new SelectListItem { Text = "Three", Value = "3"},
+        };
+
+        [Fact]
+        public void DropDownList_FindsSelectList()
+        {
+            // Arrange
+            var expectedHtml = "<select id=\"HtmlEncode[[Property1]]\" name=\"HtmlEncode[[Property1]]\">" +
+                "<option value=\"HtmlEncode[[0]]\">HtmlEncode[[Zero]]</option>" + Environment.NewLine +
+                "<option value=\"HtmlEncode[[1]]\">HtmlEncode[[One]]</option>" + Environment.NewLine +
+                "<option value=\"HtmlEncode[[2]]\">HtmlEncode[[Two]]</option>" + Environment.NewLine +
+                "<option selected=\"HtmlEncode[[selected]]\" value=\"HtmlEncode[[3]]\">HtmlEncode[[Three]]</option>" + Environment.NewLine +
+                "</select>";
+            var metadataProvider = new EmptyModelMetadataProvider();
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper(new ViewDataDictionary<TestModel>(metadataProvider));
+            helper.ViewContext.ClientValidationEnabled = false;
+            helper.ViewData.ModelState.SetModelValue("Property1", 3, "3");
+            helper.ViewData["Property1"] = BasicSelectList;
+
+            // Act
+            var dropDownListResult = helper.DropDownList("Property1");
+
+            // Assert
+            Assert.Equal(expectedHtml, HtmlContentUtilities.HtmlContentToString(dropDownListResult));
+        }
+
+        [Fact]
+        public void DropDownList_FindsSelectList_UsesOptionLabel()
+        {
+            // Arrange
+            var expectedHtml = "<select id=\"HtmlEncode[[Property1]]\" name=\"HtmlEncode[[Property1]]\">" +
+                "<option value=\"\">HtmlEncode[[--select--]]</option>" + Environment.NewLine +
+                "<option value=\"HtmlEncode[[0]]\">HtmlEncode[[Zero]]</option>" + Environment.NewLine +
+                "<option selected=\"HtmlEncode[[selected]]\" value=\"HtmlEncode[[1]]\">HtmlEncode[[One]]</option>" + Environment.NewLine +
+                "<option value=\"HtmlEncode[[2]]\">HtmlEncode[[Two]]</option>" + Environment.NewLine +
+                "<option value=\"HtmlEncode[[3]]\">HtmlEncode[[Three]]</option>" + Environment.NewLine +
+                "</select>";
+            var metadataProvider = new EmptyModelMetadataProvider();
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper(new ViewDataDictionary<TestModel>(metadataProvider));
+            helper.ViewContext.ClientValidationEnabled = false;
+            helper.ViewData.ModelState.SetModelValue("Property1", 1, "1");
+            helper.ViewData["Property1"] = BasicSelectList;
+
+            // Act
+            var dropDownListResult = helper.DropDownList("Property1", "--select--");
+
+            // Assert
+            Assert.Equal(expectedHtml, HtmlContentUtilities.HtmlContentToString(dropDownListResult));
+        }
+
+        [Fact]
+        public void DropDownList_UsesSpecifiedExpressionAndSelectList()
+        {
+            // Arrange
+            var expectedHtml = "<select id=\"HtmlEncode[[Property2]]\" name=\"HtmlEncode[[Property2]]\">" +
+                "<option selected=\"HtmlEncode[[selected]]\" value=\"HtmlEncode[[4]]\">HtmlEncode[[Four]]</option>" + Environment.NewLine +
+                "<option value=\"HtmlEncode[[5]]\">HtmlEncode[[Five]]</option>" + Environment.NewLine +
+                "</select>";
+            var selectList = new List<SelectListItem>
+            {
+                new SelectListItem { Text = "Four", Value = "4" },
+                new SelectListItem { Text = "Five", Value = "5" },
+            };
+            var metadataProvider = new EmptyModelMetadataProvider();
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper(new ViewDataDictionary<TestModel>(metadataProvider));
+            helper.ViewContext.ClientValidationEnabled = false;
+            helper.ViewData.Model = new TestModel { Property2 = "4" };
+
+            // Act
+            var dropDownListResult = helper.DropDownList("Property2", selectList);
+
+            // Assert
+            Assert.Equal(expectedHtml, HtmlContentUtilities.HtmlContentToString(dropDownListResult));
+        }
+
+        [Fact]
+        public void DropDownList_UsesSpecifiedExpressionAndSelectListAndHtmlAttributes()
+        {
+            // Arrange
+            var expectedHtml = "<select id=\"HtmlEncode[[Property1]]\" Key=\"HtmlEncode[[Value]]\" name=\"HtmlEncode[[Property1]]\">" +
+                "<option selected=\"HtmlEncode[[selected]]\" value=\"HtmlEncode[[4]]\">HtmlEncode[[Four]]</option>" + Environment.NewLine +
+                "<option value=\"HtmlEncode[[5]]\">HtmlEncode[[Five]]</option>" + Environment.NewLine +
+                "</select>";
+            var selectList = new List<SelectListItem>
+            {
+                new SelectListItem { Text = "Four", Value = "4" },
+                new SelectListItem { Text = "Five", Value = "5" },
+            };
+            var metadataProvider = new EmptyModelMetadataProvider();
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper(new ViewDataDictionary<TestModel>(metadataProvider));
+            helper.ViewContext.ClientValidationEnabled = false;
+            helper.ViewData.ModelState.SetModelValue("Property1", 4, "4");
+            helper.ViewData["Property1"] = BasicSelectList;
+
+            // Act
+            var dropDownListResult = helper.DropDownList("Property1", selectList, new { Key = "Value", name = "CustomName" });
+
+            // Assert
+            Assert.Equal(expectedHtml, HtmlContentUtilities.HtmlContentToString(dropDownListResult));
+        }
+
+        [Fact]
+        public void DropDownList_UsesExpressionAndSpecifiedSelectListAndOptionLabel()
+        {
+            // Arrange
+            var expectedHtml = "<select id=\"HtmlEncode[[Property2]]\" name=\"HtmlEncode[[Property2]]\">" +
+                "<option value=\"\">HtmlEncode[[--select--]]</option>" + Environment.NewLine +
+                "<option value=\"HtmlEncode[[4]]\">HtmlEncode[[Four]]</option>" + Environment.NewLine +
+                "<option selected=\"HtmlEncode[[selected]]\" value=\"HtmlEncode[[5]]\">HtmlEncode[[Five]]</option>" + Environment.NewLine +
+                "</select>";
+            var selectList = new List<SelectListItem>
+            {
+                new SelectListItem { Text = "Four", Value = "4" },
+                new SelectListItem { Text = "Five", Value = "5" },
+            };
+            var metadataProvider = new EmptyModelMetadataProvider();
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper(new ViewDataDictionary<TestModel>(metadataProvider));
+            helper.ViewContext.ClientValidationEnabled = false;
+            helper.ViewData.Model = new TestModel { Property2 = "5" };
+
+            // Act
+            var dropDownListResult = helper.DropDownList("Property2", selectList, "--select--");
+
+            // Assert
+            Assert.Equal(expectedHtml, HtmlContentUtilities.HtmlContentToString(dropDownListResult));
+        }
+
+        [Fact]
+        public void DropDownListFor_NullSelectListFindsListFromViewData()
+        {
+            // Arrange
+            var expectedHtml = "<select id=\"HtmlEncode[[Property1]]\" name=\"HtmlEncode[[Property1]]\">" +
+                "<option selected=\"HtmlEncode[[selected]]\" value=\"HtmlEncode[[0]]\">HtmlEncode[[Zero]]</option>" + Environment.NewLine +
+                "<option value=\"HtmlEncode[[1]]\">HtmlEncode[[One]]</option>" + Environment.NewLine +
+                "<option value=\"HtmlEncode[[2]]\">HtmlEncode[[Two]]</option>" + Environment.NewLine +
+                "<option value=\"HtmlEncode[[3]]\">HtmlEncode[[Three]]</option>" + Environment.NewLine +
+                "</select>";
+            var metadataProvider = new EmptyModelMetadataProvider();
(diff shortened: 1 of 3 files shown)


test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperFormTest.cs
```diff
@@ -114,7 +114,6 @@ public class HtmlHelperFormTest
                     },
                 };
             }
-
```
dougbu: :clap:

dougbu:
I'm pretty sure we have a fair number of tests for the `<select>` tag helper. Do they cover any HTML helper-applicable scenarios that you haven't included here? (I'm thinking a fair number of the `<select>` tests will deal with static `<option>`s and other things the HTML helpers don't do.)

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperDropDownListExtensionsTest.cs
```diff
@@ -0,0 +1,232 @@
...
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
+using Microsoft.AspNetCore.Mvc.Rendering;
+using Microsoft.AspNetCore.Mvc.TestCommon;
+using Microsoft.AspNetCore.Mvc.ViewFeatures;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Mvc.Core
+{
+    public class HtmlHelperDropDownListExtensionsTest
```
dougbu: In both of these classes, need to add more complex `SelectListItem`s, especially those that have a null `Value`, are `Disabled` or `Selected`, or are grouped. Note that `Selected` is used only when the helper is given a `null` expression.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperDropDownListExtensionsTest.cs
```diff
@@ -0,0 +1,232 @@
...
+        private static readonly List<SelectListItem> BasicSelectList = new List<SelectListItem>
+        {
+            new SelectListItem { Text = "Zero",  Value = "0"},
+            new SelectListItem { Text = "One",   Value = "1"},
+            new SelectListItem { Text = "Two",   Value = "2"},
+            new SelectListItem { Text = "Three", Value = "3"},
+        };
+
+        [Fact]
+        public void DropDownList_UsesSpecifiedExpression()
```
dougbu: `DropDownList_FindsSelectList()`. This test isn't using the specified expression. It finds the select list at the named location in VDD and that's it. Actually a case we have a bug about -- somewhere in the backlog -- because users can't put both the expression and the select list in VDD.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperDropDownListExtensionsTest.cs
```diff
@@ -0,0 +1,232 @@
...
+                "<option value=\"HtmlEncode[[4]]\">HtmlEncode[[Four]]</option>" + Environment.NewLine +
+                "<option value=\"HtmlEncode[[5]]\">HtmlEncode[[Five]]</option>" + Environment.NewLine +
+                "</select>";
+            var metadataProvider = new EmptyModelMetadataProvider();
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper(new ViewDataDictionary<TestModel>(metadataProvider));
+            helper.ViewContext.ClientValidationEnabled = false;
+            helper.ViewData.Model = new TestModel { Property1 = BasicSelectList };
+
+            // Act
+            var dropDownListResult = helper.DropDownList("Property1", new List<SelectListItem>
```
dougbu: Never do this. Give `Property1` a real value when passing the select list in. Muck with the value and you'll see `selected="selected"` start to show up in the HTML.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperDropDownListExtensionsTest.cs
dougbu: Update all names where you aren't providing an expression.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperDropDownListExtensionsTest.cs
```diff
@@ -0,0 +1,232 @@
...
+                "<option value=\"HtmlEncode[[4]]\">HtmlEncode[[Four]]</option>" + Environment.NewLine +
+                "<option value=\"HtmlEncode[[5]]\">HtmlEncode[[Five]]</option>" + Environment.NewLine +
+                "</select>";
+            var metadataProvider = new EmptyModelMetadataProvider();
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper(new ViewDataDictionary<TestModel>(metadataProvider));
+            helper.ViewContext.ClientValidationEnabled = false;
+            helper.ViewData.Model = new TestModel { Property1 = BasicSelectList };
+
+            // Act
+            var dropDownListForResult = helper.DropDownListFor(m => m.Property1, new List<SelectListItem>
```
dougbu: In this case it's fine to place the select list in VDD directly: `helper.ViewData["Property1"] = BasicSelectList;` But the passed expression should _never_ refer to a select list. Give this property a real value in the model.

dougbu:
:watch:

ajaybhargavb:
@dougbu, Some of the complex cases we discussed are already extensively covered in `HtmlHelperSelectTest`. So I've added only tests relevant to testing the extension methods that were not tested before.

ajaybhargavb:
:up: :date:

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperDropDownListExtensionsTest.cs
ajaybhargavb: https://github.com/aspnet/Mvc/pull/4212#issuecomment-199420905

dougbu:
> already extensively covered in `HtmlHelperSelectTest`

Do `HtmlHelperDropDownListExtensionsTest` and `HtmlHelperListBoxExtensionsTest` test only extension methods? Does `HtmlHelperSelectTest` test just methods implemented directly in `HtmlHelper` and `HtmlHelper<T>`? If not, should split things up in that way or for example change `HtmlHelperSelectTest` to always call the longest overloads (the `HtmlHelper` and `HtmlHelper<T>` methods). (Definitely don't want test classes named `BlahExtensionsTest` testing non-extension methods. But adding lots of `blah: null` arguments in `HtmlHelperSelectTest` seems less important / urgent.)

ajaybhargavb:
> Do HtmlHelperDropDownListExtensionsTest and HtmlHelperListBoxExtensionsTest test only extension methods? Does HtmlHelperSelectTest test just methods implemented directly in HtmlHelper and HtmlHelper<T>?

Yes and Yes. They are already split in the way you suggested.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperDropDownListExtensionsTest.cs
```diff
@@ -0,0 +1,259 @@
...
+            var expectedHtml = "<select id=\"HtmlEncode[[Property1]]\" name=\"HtmlEncode[[Property1]]\">" +
+                "<option value=\"HtmlEncode[[0]]\">HtmlEncode[[Zero]]</option>" + Environment.NewLine +
+                "<option value=\"HtmlEncode[[1]]\">HtmlEncode[[One]]</option>" + Environment.NewLine +
+                "<option value=\"HtmlEncode[[2]]\">HtmlEncode[[Two]]</option>" + Environment.NewLine +
+                "<option value=\"HtmlEncode[[3]]\">HtmlEncode[[Three]]</option>" + Environment.NewLine +
+                "</select>";
+            var metadataProvider = new EmptyModelMetadataProvider();
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper(new ViewDataDictionary<TestModel>(metadataProvider));
+            helper.ViewContext.ClientValidationEnabled = false;
+            helper.ViewData.Model = new TestModel { Property1 = BasicSelectList };
```
dougbu: Why does `TestModel` have a property of type `IEnumerable<SelectListItem>`? If it were a view model, the property would _never_ be looked up in this manner; the view model property would be used explicitly.

The usual patterns are
``` c#
class ViewModel
{
    string Property1 { get; set; }
    IEnumerable<SelectListItem> Property1SelectList { get; }
}
...
@Html.DropDownListFor(m => m.Property1, Model.Property1SelectList)
```
or
``` c#
class RealModel
{
    string Property1 { get; set; }
}
...
public IActionResult Index()
{
    ViewData["Property1"] = new SelectList(...);
    return View();
}
...
@Html.DropDownListFor(m => m.Property1)
```

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperDropDownListExtensionsTest.cs
```diff
@@ -0,0 +1,259 @@
...
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper(new ViewDataDictionary<TestModel>(metadataProvider));
+            helper.ViewContext.ClientValidationEnabled = false;
+            helper.ViewData.Model = new TestModel { Property2 = "4" };
+
+            // Act
+            var dropDownListResult = helper.DropDownList("Property2", new List<SelectListItem>
+            {
+                new SelectListItem { Text = "Four", Value = "4" },
+                new SelectListItem { Text = "Five", Value = "5" },
+            });
```
dougbu: Set up the `List<SelectListItem>` under `// Arrange` rather than inline here.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperDropDownListExtensionsTest.cs
```diff
@@ -0,0 +1,259 @@
...
+
+            // Assert
+            Assert.Equal(expectedHtml, HtmlContentUtilities.HtmlContentToString(dropDownListForResult));
+        }
+
+        [Fact]
+        public void DropDownListFor_UsesSpecifiedSelectListAndHtmlAttributes()
```
dougbu: Mention the expression in this name.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperDropDownListExtensionsTest.cs
```diff
@@ -0,0 +1,259 @@
...
+                "<option value=\"HtmlEncode[[2]]\">HtmlEncode[[Two]]</option>" + Environment.NewLine +
+                "<option value=\"HtmlEncode[[3]]\">HtmlEncode[[Three]]</option>" + Environment.NewLine +
+                "</select>";
+            var metadataProvider = new EmptyModelMetadataProvider();
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper(new ViewDataDictionary<TestModel>(metadataProvider));
+            helper.ViewContext.ClientValidationEnabled = false;
+            helper.ViewData["Property1"] = BasicSelectList;
+
+            // Act
+            var dropDownListForResult = helper.DropDownListFor(m => m.Property1, null);
```
dougbu: Name the parameter. (Surprised this call isn't ambiguous.)

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperDropDownListExtensionsTest.cs
```diff
@@ -0,0 +1,259 @@
...
+                    new SelectListItem { Text = "Five", Value = "5" },
+                },
+                new { Key = "Value", name = "CustomName" });
+
+            // Assert
+            Assert.Equal(expectedHtml, HtmlContentUtilities.HtmlContentToString(dropDownListResult));
+        }
+
+        [Fact]
+        public void DropDownList_UsesSpecifiedSelectListAndOptionLabel()
```
dougbu: This test also uses the expression.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperDropDownListExtensionsTest.cs
```diff
@@ -0,0 +1,259 @@
...
+            // Arrange
+            var expectedHtml = "<select id=\"HtmlEncode[[Property2]]\" name=\"HtmlEncode[[Property2]]\">" +
+                "<option value=\"\">HtmlEncode[[--select--]]</option>" + Environment.NewLine +
+                "<option value=\"HtmlEncode[[4]]\">HtmlEncode[[Four]]</option>" + Environment.NewLine +
+                "<option value=\"HtmlEncode[[5]]\">HtmlEncode[[Five]]</option>" + Environment.NewLine +
+                "</select>";
+            var metadataProvider = new EmptyModelMetadataProvider();
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper(new ViewDataDictionary<TestModel>(metadataProvider));
+            helper.ViewContext.ClientValidationEnabled = false;
+            helper.ViewData["Property1"] = BasicSelectList;
```
dougbu: Leave out; adding an `IEnumerable<SelectListItem>` for `Property1` has no impact on the HTML generated for `Property2`.

dougbu:
:watch:

ajaybhargavb:
:up: :date:

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperDropDownListExtensionsTest.cs
ajaybhargavb: Removed. I wonder why this didn't get outdated.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperDropDownListExtensionsTest.cs
```diff
@@ -0,0 +1,256 @@
...
+            var selectList = new List<SelectListItem>
+            {
+                new SelectListItem { Text = "Four", Value = "4" },
+                new SelectListItem { Text = "Five", Value = "5" },
+            };
+            var metadataProvider = new EmptyModelMetadataProvider();
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper(new ViewDataDictionary<TestModel>(metadataProvider));
+            helper.ViewContext.ClientValidationEnabled = false;
+            helper.ViewData.ModelState.SetModelValue("Property1", 4, "4");
+            helper.ViewData["Property1"] = BasicSelectList;
```
dougbu: Weird case due to this extra select list. Just set `ViewData.Model.Property1`.

dougbu: And, yes, this does represent a showcase @NTaylorMullen

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperDropDownListExtensionsTest.cs
dougbu: Agree it's odd this comment wasn't hidden.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperDropDownListExtensionsTest.cs
```diff
@@ -0,0 +1,256 @@
...
+        [Fact]
+        public void DropDownListFor_NullSelectListFindsListFromViewData()
+        {
+            // Arrange
+            var expectedHtml = "<select id=\"HtmlEncode[[Property1]]\" name=\"HtmlEncode[[Property1]]\">" +
+                "<option value=\"HtmlEncode[[0]]\">HtmlEncode[[Zero]]</option>" + Environment.NewLine +
+                "<option value=\"HtmlEncode[[1]]\">HtmlEncode[[One]]</option>" + Environment.NewLine +
+                "<option value=\"HtmlEncode[[2]]\">HtmlEncode[[Two]]</option>" + Environment.NewLine +
+                "<option value=\"HtmlEncode[[3]]\">HtmlEncode[[Three]]</option>" + Environment.NewLine +
+                "</select>";
```
dougbu: Good that an `<option>` has `selected="selected"` in every `DropDownList()` test. Use the `SetModelValue()` trick to do the same here.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperDropDownListExtensionsTest.cs
```diff
@@ -0,0 +1,256 @@
...
+                new SelectListItem { Text = "Four", Value = "4" },
+                new SelectListItem { Text = "Five", Value = "5" },
+            };
+            var metadataProvider = new EmptyModelMetadataProvider();
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper(new ViewDataDictionary<TestModel>(metadataProvider));
+            helper.ViewContext.ClientValidationEnabled = false;
+
+            // Act
+            var dropDownListForResult = helper.DropDownListFor(
+                m => m.Property2, selectList, "--select--");
```
dougbu: All on one line.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperDropDownListExtensionsTest.cs
```diff
@@ -0,0 +1,256 @@
...
+                "<option value=\"HtmlEncode[[5]]\">HtmlEncode[[Five]]</option>" + Environment.NewLine +
+                "</select>";
+            var selectList = new List<SelectListItem>
+            {
+                new SelectListItem { Text = "Four", Value = "4" },
+                new SelectListItem { Text = "Five", Value = "5" },
+            };
+            var metadataProvider = new EmptyModelMetadataProvider();
+            var helper = DefaultTemplatesUtilities.GetHtmlHelper(new ViewDataDictionary<TestModel>(metadataProvider));
+            helper.ViewContext.ClientValidationEnabled = false;
```
dougbu: Set `Property2` to ensure something is selected.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperListBoxExtensionsTest.cs
```diff
@@ -0,0 +1,178 @@
...
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
+using Microsoft.AspNetCore.Mvc.Rendering;
+using Microsoft.AspNetCore.Mvc.TestCommon;
+using Microsoft.AspNetCore.Mvc.ViewFeatures;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Mvc.Core
+{
+    public class HtmlHelperListBoxExtensionsTest
```
dougbu: Double-check that this file is the same as `HtmlHelperDropDownListExtensionsTest`, modulo the expected details and less the overloads `ListBox()` and `ListBoxFor()` don't support.

dougbu:
:shipit: after fixing bits mentioned above.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperDropDownListExtensionsTest.cs
ajaybhargavb: This is to show that the passed in select list is preferred over the one in `ViewData` as opposed to the case in `DropDownListFor_NullSelectListFindsListFromViewData`.

dougbu: :shrug: probably not worth testing but alright.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/HtmlHelperListBoxExtensionsTest.cs
ajaybhargavb: Checked. ListBox test covers the same cases as DropDownList test minus the missing `OptionLabel` overloads.

PR #4227, opened 2016-03-04
Remove IMvcRazorHost.MainClassPrefix

pranavkm:


pranavkm:
cc @rynowak  - we spoke about refactoring this a while back. Saw this was left over when I was working on adding tests.

src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultRoslynCompilationService.cs
```diff
@@ -99,7 +95,7 @@ public CompilationResult Compile(RelativeFileInfo fileInfo, string compilationCo
 
             var startTimestamp = _logger.IsEnabled(LogLevel.Debug) ? Stopwatch.GetTimestamp() : 0;
 
-            var assemblyName = Path.GetRandomFileName();
+            var assemblyName = fileInfo.FileInfo.Name + "." + Path.GetRandomFileName();
```
NTaylorMullen: Are all file names valid as parts of an assembly name?

pranavkm: I'll have a look.

dougbu: Why is the assembly name changing / getting longer here?

dougbu:
@pranavkm what issue does this PR solve?

src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultRoslynCompilationService.cs
pranavkm: It made debugging in the Modules window a tad simpler. I'll get this tracked in a separate work item

pranavkm:
This was some left over code that we wanted to remove the last time this file was refactored. At that point precompilation was using the property which made it tricky.

src/Microsoft.AspNetCore.Mvc.Razor.Host/IMvcRazorHost.cs
```diff
@@ -23,11 +23,6 @@ public interface IMvcRazorHost : IDisposable
         GeneratorResults GenerateCode(string rootRelativePath, Stream inputStream);
 
         /// <summary>
-        /// Represent the prefix off the main entry class in the view.
-        /// </summary>
-        string MainClassNamePrefix { get; }
```
pranavkm: cc @Eilon

Eilon: :+1:

pranavkm:
Updated. Moved the assembly name mangling to a separate work item - https://github.com/aspnet/Mvc/issues/4239

src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultRoslynCompilationService.cs
```diff
@@ -158,9 +154,8 @@ public CompilationResult Compile(RelativeFileInfo fileInfo, string compilationCo
                         assembly = LoadStream(ms, assemblySymbols: null);
                     }
 
-                    var type = assembly
-                        .GetExportedTypes()
-                        .First(t => t.Name.StartsWith(_classPrefix, StringComparison.Ordinal));
+                    Debug.Assert(assembly.GetExportedTypes().Length == 1);
```
NTaylorMullen: This breaks if you have an `@functions` that defines a nested `public class`

pranavkm: Ok. I guess we can look for non-nested types.

pranavkm:
Updated

dougbu:
:shipit:

NTaylorMullen:
:shipit:

PR #4232, opened 2016-03-04
Refactoring CacheTagHelper

### LINKED ISSUE #3867: CacheTagHelper should not depend on IMemoryCache

### LINKED ISSUE #4147: Consider storing a Task<> in cache tag
We're wondering if the concurrency behavior of #CacheTag is quite right.

Since we store the computed content in a `StringBuilder` - on cache miss, every thread to reach that block needs to recompute the content. If the content takes a long time to compute, it could mean a lot of extra DB calls or just general processing.

It seems beneficial to have the cache store `Task<>` so that we can quickly address the 'miss' and have other threads aside from the one computing the data wait on the task.

sebastienros:
- Introducing a new distributed cache tag helper
- Sharing base implementation for both cache tag helper
- Preventing concurrent execution of cache tag helpers

Fixes #4147 , Fixes #3867

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperFormatter.cs
+++ b/src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperFormatter.cs
@@ -1,0 +1,47 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc.Rendering;
+
+namespace Microsoft.AspNetCore.Mvc.TagHelpers.Cache
+{
+    /// <summary>
+    /// Implements <see cref="IDistributedCacheTagHelperFormatter"/> by serializing the content
+    /// in UTF8.
+    /// </summary>
+    public class DistributedCacheTagHelperFormatter : IDistributedCacheTagHelperFormatter
+    {
+        
+        /// <inheritdoc />
+        public Task<byte[]> SerializeAsync(DistributedCacheTagHelperFormattingContext context)
+        {
+            if (context == null) 
+            {
+                throw new ArgumentNullException(nameof(context));    
+            }
+            
+            if (context.Html == null) 
+            {
+                throw new ArgumentNullException(nameof(context.Html));    
+            }
+
+            var serialized = Encoding.UTF8.GetBytes(context.Html.ToString());
+            return Task.FromResult(serialized);
+        }
+
+        /// <inheritdoc />
+        public Task<HtmlString> DeserializeAsync(byte[] value)
+        {
+            if (value == null) 
+            {
+                throw new ArgumentNullException(nameof(value));    
+            }
+            
+            var content = Encoding.UTF8.GetString(value);
+            return Task.FromResult(new HtmlString(content));
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperFormattingContext.cs
+++ b/src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperFormattingContext.cs
@@ -1,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Mvc.Rendering;
+
+namespace Microsoft.AspNetCore.Mvc.TagHelpers.Cache
+{
+    /// <summary>
+    /// Represents an object containing the information to serialize with <see cref="IDistributedCacheTagHelperFormatter" />.
+    /// </summary>
+    public class DistributedCacheTagHelperFormattingContext
+    {
+        /// <summary>
+        /// Gets the <see cref="HtmlString"/> instance.
+        /// </summary>
+        public HtmlString Html { get; set; }  
+    }
+}
--- a/src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperStorage.cs
+++ b/src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperStorage.cs
@@ -1,0 +1,54 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Caching.Distributed;
+
+namespace Microsoft.AspNetCore.Mvc.TagHelpers.Cache
+{
+    /// <summary>
+    /// Implements <see cref="IDistributedCacheTagHelperStorage"/> by storing the content
+    /// in using <see cref="IDistributedCache"/> as the store.
+    /// </summary>
+    public class DistributedCacheTagHelperStorage : IDistributedCacheTagHelperStorage
+    {
+        private readonly IDistributedCache _distributedCache;
+
+        /// <summary>
+        /// Creates a new <see cref="DistributedCacheTagHelperStorage"/>.
+        /// </summary>
+        /// <param name="distributedCache">The <see cref="IDistributedCache"/> to use.</param>
+        public DistributedCacheTagHelperStorage(IDistributedCache distributedCache)
+        {
+            _distributedCache = distributedCache;
+        }
+
+        /// <inheritdoc />
+        public Task<byte[]> GetAsync(string key)
+        {
+            if (key == null)
+            {
+                throw new ArgumentNullException(nameof(key));
+            }
+            
+            return _distributedCache.GetAsync(key);
+        }
+
+        /// <inheritdoc />
+        public Task SetAsync(string key, byte[] value, DistributedCacheEntryOptions options)
+        {
+            if (key == null)
+            {
+                throw new ArgumentNullException(nameof(key));
+            }
+
+            if (value == null)
+            {
+                throw new ArgumentNullException(nameof(value));
+            }
+
+            return _distributedCache.SetAsync(key, value, options);
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelper.cs
+++ b/src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelper.cs
@@ -2,15 +2,12 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
 using System.IO;
-using System.Security.Cryptography;
 using System.Text;
 using System.Text.Encodings.Web;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Html;
-using Microsoft.AspNetCore.Mvc.Rendering;
-using Microsoft.AspNetCore.Mvc.ViewFeatures;
 using Microsoft.AspNetCore.Razor.TagHelpers;
 using Microsoft.Extensions.Caching.Memory;
 using Microsoft.Extensions.Primitives;
@@ -20,44 +17,22 @@ namespace Microsoft.AspNetCore.Mvc.TagHelpers
     /// <summary>
     /// <see cref="TagHelper"/> implementation targeting &lt;cache&gt; elements.
     /// </summary>
-    public class CacheTagHelper : TagHelper
+    public class CacheTagHelper : CacheTagHelperBase
     {
         /// <summary>
         /// Prefix used by <see cref="CacheTagHelper"/> instances when creating entries in <see cref="MemoryCache"/>.
         /// </summary>
         public static readonly string CacheKeyPrefix = nameof(CacheTagHelper);
-        private const string VaryByAttributeName = "vary-by";
-        private const string VaryByHeaderAttributeName = "vary-by-header";
-        private const string VaryByQueryAttributeName = "vary-by-query";
-        private const string VaryByRouteAttributeName = "vary-by-route";
-        private const string VaryByCookieAttributeName = "vary-by-cookie";
-        private const string VaryByUserAttributeName = "vary-by-user";
-        private const string ExpiresOnAttributeName = "expires-on";
-        private const string ExpiresAfterAttributeName = "expires-after";
-        private const string ExpiresSlidingAttributeName = "expires-sliding";
         private const string CachePriorityAttributeName = "priority";
-        private const string CacheKeyTokenSeparator = "||";
-        private const string EnabledAttributeName = "enabled";
-        private static readonly char[] AttributeSeparator = new[] { ',' };
 
         /// <summary>
         /// Creates a new <see cref="CacheTagHelper"/>.
         /// </summary>
         /// <param name="memoryCache">The <see cref="IMemoryCache"/>.</param>
         /// <param name="htmlEncoder">The <see cref="HtmlEncoder"/> to use.</param>
-        public CacheTagHelper(IMemoryCache memoryCache, HtmlEncoder htmlEncoder)
+        public CacheTagHelper(IMemoryCache memoryCache, HtmlEncoder htmlEncoder) : base(htmlEncoder)
         {
             MemoryCache = memoryCache;
-            HtmlEncoder = htmlEncoder;
-        }
-
-        /// <inheritdoc />
-        public override int Order
-        {
-            get
-            {
-                return -1000;
-            }
         }
 
         /// <summary>
@@ -66,83 +41,10 @@ namespace Microsoft.AspNetCore.Mvc.TagHelpers
         protected IMemoryCache MemoryCache { get; }
 
         /// <summary>
-        /// Gets the <see cref="System.Text.Encodings.Web.HtmlEncoder"/> which encodes the content to be cached.
-        /// </summary>
-        protected HtmlEncoder HtmlEncoder { get; }
-
-        /// <summary>
-        /// Gets or sets the <see cref="ViewContext"/> for the current executing View.
-        /// </summary>
-        [HtmlAttributeNotBound]
-        [ViewContext]
-        public ViewContext ViewContext { get; set; }
-
-        /// <summary>
-        /// Gets or sets a <see cref="string" /> to vary the cached result by.
-        /// </summary>
-        [HtmlAttributeName(VaryByAttributeName)]
-        public string VaryBy { get; set; }
-
-        /// <summary>
-        /// Gets or sets the name of a HTTP request header to vary the cached result by.
-        /// </summary>
-        [HtmlAttributeName(VaryByHeaderAttributeName)]
-        public string VaryByHeader { get; set; }
-
-        /// <summary>
-        /// Gets or sets a comma-delimited set of query parameters to vary the cached result by.
-        /// </summary>
-        [HtmlAttributeName(VaryByQueryAttributeName)]
-        public string VaryByQuery { get; set; }
-
-        /// <summary>
-        /// Gets or sets a comma-delimited set of route data parameters to vary the cached result by.
-        /// </summary>
-        [HtmlAttributeName(VaryByRouteAttributeName)]
-        public string VaryByRoute { get; set; }
-
-        /// <summary>
-        /// Gets or sets a comma-delimited set of cookie names to vary the cached result by.
-        /// </summary>
-        [HtmlAttributeName(VaryByCookieAttributeName)]
-        public string VaryByCookie { get; set; }
-
-        /// <summary>
-        /// Gets or sets a value that determines if the cached result is to be varied by the Identity for the logged in
-        /// <see cref="Http.HttpContext.User"/>.
-        /// </summary>
-        [HtmlAttributeName(VaryByUserAttributeName)]
-        public bool VaryByUser { get; set; }
-
-        /// <summary>
-        /// Gets or sets the exact <see cref="DateTimeOffset"/> the cache entry should be evicted.
-        /// </summary>
-        [HtmlAttributeName(ExpiresOnAttributeName)]
-        public DateTimeOffset? ExpiresOn { get; set; }
-
-        /// <summary>
-        /// Gets or sets the duration, from the time the cache entry was added, when it should be evicted.
-        /// </summary>
-        [HtmlAttributeName(ExpiresAfterAttributeName)]
-        public TimeSpan? ExpiresAfter { get; set; }
-
-        /// <summary>
-        /// Gets or sets the duration from last access that the cache entry should be evicted.
-        /// </summary>
-        [HtmlAttributeName(ExpiresSlidingAttributeName)]
-        public TimeSpan? ExpiresSliding { get; set; }
-
-        /// <summary>
         /// Gets or sets the <see cref="CacheItemPriority"/> policy for the cache entry.
         /// </summary>
         [HtmlAttributeName(CachePriorityAttributeName)]
         public CacheItemPriority? Priority { get; set; }
-
-        /// <summary>
-        /// Gets or sets the value which determines if the tag helper is enabled or not.
-        /// </summary>
-        [HtmlAttributeName(EnabledAttributeName)]
-        public bool Enabled { get; set; } = true;
 
         /// <inheritdoc />
         public override async Task ProcessAsync(TagHelperContext context, TagHelperOutput output)
@@ -156,87 +58,84 @@ namespace Microsoft.AspNetCore.Mvc.TagHelpers
             {
                 throw new ArgumentNullException(nameof(output));
             }
+            
+            IHtmlContent content = null;
 
-            IHtmlContent result = null;
             if (Enabled)
             {
                 var key = GenerateKey(context);
-                if (!MemoryCache.TryGetValue(key, out result))
+                MemoryCacheEntryOptions options;
+    
+                while (content == null)
                 {
-                    // Create an entry link scope and flow it so that any tokens related to the cache entries
-                    // created within this scope get copied to this scope.
-                    using (var link = MemoryCache.CreateLinkingScope())
+                    Task<IHtmlContent> result = null;
+                    
+                    if (!MemoryCache.TryGetValue(key, out result))
                     {
-                        var content = await output.GetChildContentAsync();
+                        var tokenSource = new CancellationTokenSource();
 
-                        var stringBuilder = new StringBuilder();
-                        using (var writer = new StringWriter(stringBuilder))
+                        // Create an entry link scope and flow it so that any tokens related to the cache entries
+                        // created within this scope get copied to this scope.
+
+                        options = GetMemoryCacheEntryOptions();
+                        options.AddExpirationToken(new CancellationChangeToken(tokenSource.Token));
+
+                        var tcs = new TaskCompletionSource<IHtmlContent>();
+                         
+                        MemoryCache.Set(key, tcs.Task, options);
+                        
+                        try 
                         {
-                            content.WriteTo(writer, HtmlEncoder);
+                            using (var link = MemoryCache.CreateLinkingScope())
+                            {
+                                result = ProcessContentAsync(output);
+                                content = await result;
+                                options.AddEntryLink(link);
+                            }
+                            
+                            // The entry is set instead of assigning a value to the 
+                            // task so that the expiration options are are not impacted 
+                            // by the time it took to compute it.
+                            
+                            MemoryCache.Set(key, result, options);
                         }
+                        catch
+                        {
+                            // Remove the worker task from the cache in case it can't complete.
+                            tokenSource.Cancel();
+                            throw;
+                        }
+                        finally
+                        {
+                            // If an exception occurs, ensure the other awaiters 
+                            // render the output by themselves.
+                            tcs.SetResult(null);
+                        }
+                    }
+                    else
+                    {
+                        // There is either some value already cached (as a Task)
+                        // or a worker processing the output. In the case of a worker,
+                        // the result will be null, and the request will try to acquire
+                        // the result from memory another time.
 
-                        result = new StringBuilderHtmlContent(stringBuilder);
-                        MemoryCache.Set(key, result, GetMemoryCacheEntryOptions(link));
+                        content = await result;
                     }
                 }
+            }
+            else
+            {
+                content = await output.GetChildContentAsync();
             }
 
             // Clear the contents of the "cache" element since we don't want to render it.
             output.SuppressOutput();
-            if (Enabled)
-            {
-                output.Content.SetContent(result);
-            }
-            else
-            {
-                result = await output.GetChildContentAsync();
-                output.Content.SetContent(result);
-            }
+
+            output.Content.SetContent(content);
         }
 
         // Internal for unit testing
-        internal string GenerateKey(TagHelperContext context)
-        {
-            var builder = new StringBuilder(CacheKeyPrefix);
-            builder.Append(CacheKeyTokenSeparator)
-                   .Append(context.UniqueId);
-
-            var request = ViewContext.HttpContext.Request;
-
-            if (!string.IsNullOrEmpty(VaryBy))
-            {
-                builder.Append(CacheKeyTokenSeparator)
-                       .Append(nameof(VaryBy))
-                       .Append(CacheKeyTokenSeparator)
-                       .Append(VaryBy);
-            }
-
-            AddStringCollectionKey(builder, nameof(VaryByCookie), VaryByCookie, request.Cookies, (c, key) => c[key]);
-            AddStringCollectionKey(builder, nameof(VaryByHeader), VaryByHeader, request.Headers, (c, key) => c[key]);
-            AddStringCollectionKey(builder, nameof(VaryByQuery), VaryByQuery, request.Query, (c, key) => c[key]);
-            AddVaryByRouteKey(builder);
-
-            if (VaryByUser)
-            {
-                builder.Append(CacheKeyTokenSeparator)
-                       .Append(nameof(VaryByUser))
-                       .Append(CacheKeyTokenSeparator)
-                       .Append(ViewContext.HttpContext.User?.Identity?.Name);
-            }
-
-            // The key is typically too long to be useful, so we use a cryptographic hash
-            // as the actual key (better randomization and key distribution, so small vary
-            // values will generate dramatically different keys).
-            using (var sha = SHA256.Create())
-            {
-                var contentBytes = Encoding.UTF8.GetBytes(builder.ToString());
-                var hashedBytes = sha.ComputeHash(contentBytes);
-                return Convert.ToBase64String(hashedBytes);
-            }
-        }
-
-        // Internal for unit testing
-        internal MemoryCacheEntryOptions GetMemoryCacheEntryOptions(IEntryLink entryLink)
+        internal MemoryCacheEntryOptions GetMemoryCacheEntryOptions()
         {
             var options = new MemoryCacheEntryOptions();
             if (ExpiresOn != null)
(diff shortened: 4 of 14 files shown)


src/Microsoft.AspNetCore.Mvc.Razor/DependencyInjection/MvcRazorMvcCoreBuilderExtensions.cs
```diff
@@ -133,6 +135,7 @@ internal static void AddRazorViewEngineServices(IServiceCollection services)
 
             // Consumed by the Cache tag helper to cache results across the lifetime of the application.
             services.TryAddSingleton<IMemoryCache, MemoryCache>();
+            services.TryAddSingleton<IDistributedCache, LocalCache>();
```
sebastienros: This change will be removed once @javiercn 's PR is merged.

sebastienros:
@lodejard Please take a look. I implemented the the way you asked.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperFormattingContext.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        /// <summary>
+        /// Gets the <see cref="TagHelperContext"/> instance.
+        /// </summary>
+        public TagHelperContext Context {get; set; }
+        
+        /// <summary>
+        /// Gets the <see cref="HtmlString"/> instance.
+        /// </summary>
+        public HtmlString HtmlString { get; set; }
```
NTaylorMullen: If this is required ( I see you null checking it in methods that consume the context) should just make this { get; } with a ctor that takes this.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperFormatter.cs
```diff
@@ -0,0 +1,47 @@
...
+            {
+                throw new ArgumentNullException(nameof(context));    
+            }
+            
+            if (context.HtmlString == null) 
+            {
+                throw new ArgumentNullException(nameof(context.HtmlString));    
+            }
+            
+            var serialized = Encoding.UTF8.GetBytes(context.HtmlString.ToString());
```
NTaylorMullen: Hmm, should we be using `WriteTo` here instead of `ToString()`? `HtmlString` is an `IHtmlCOntent` after all

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperFormattingContext.cs
NTaylorMullen: `HtmlEncodedString`? I had a similar comment in one of my Razor PRs, sounded like we were moving away from `HtmlString`. Could also just do `IHtmlContent`

NTaylorMullen: Rename `Html`

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperFormatter.cs
```diff
@@ -0,0 +1,47 @@
...
+        /// <inheritdoc />
+        public Task<HtmlString> DeserializeAsync(byte[] value)
+        {
+            if (value == null) 
+            {
+                throw new ArgumentNullException(nameof(value));    
+            }
+            
+            var content = Encoding.UTF8.GetString(value);
+            return Task.FromResult(new HtmlString(content));
```
NTaylorMullen: Similar comments here as [this](https://github.com/aspnet/Mvc/pull/4232/files#r55100329).

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperFormatter.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc.Rendering;
+
+namespace Microsoft.AspNetCore.Mvc.TagHelpers.Cache
+{
+    /// <summary>
+    /// Implements <see cref="IDistributedCacheTagHelperFormatter"/> by serializing the content
+    /// in UTF8.
+    /// </summary>
+    public class DistributedCacheTagHelperFormatter : IDistributedCacheTagHelperFormatter
```
NTaylorMullen: Hmm, is it worth it for this to be a service (instead of inlined in the distributed cache tag helper)?

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelper.cs
```diff
@@ -156,87 +56,71 @@ public override int Order
...
+            IHtmlContent content = null;
 
-            IHtmlContent result = null;
             if (Enabled)
             {
                 var key = GenerateKey(context);
-                if (!MemoryCache.TryGetValue(key, out result))
+                MemoryCacheEntryOptions options;
+    
+                while(content == null)
```
NTaylorMullen: This is odd, so if `MemoryCache` correctly retrieves a value but it's not complete content will be `null`?

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelper.cs
```diff
@@ -156,87 +56,71 @@ public override int Order
...
                 {
-                    // Create an entry link scope and flow it so that any tokens related to the cache entries
-                    // created within this scope get copied to this scope.
-                    using (var link = MemoryCache.CreateLinkingScope())
+                    Task<IHtmlContent> result = null;
+                    
+                    if (!MemoryCache.TryGetValue(key, out result))
```
NTaylorMullen: Note: If you're trying to protect against multi-threading this will race. In the end all the caches will have a single value but they may have executed the body X number of times.

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelper.cs
```diff
@@ -156,87 +56,71 @@ public override int Order
...
 
             // Clear the contents of the "cache" element since we don't want to render it.
             output.SuppressOutput();
-            if (Enabled)
-            {
-                output.Content.SetContent(result);
-            }
-            else
+
+            if (!Enabled)
```
NTaylorMullen: Just move this up to be an else of the original `if (!Enabled)`

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelper.cs
```diff
@@ -259,134 +143,17 @@ internal MemoryCacheEntryOptions GetMemoryCacheEntryOptions(IEntryLink entryLink
...
-            options.AddEntryLink(entryLink);
             return options;
         }
 
-        private static void AddStringCollectionKey(
-            StringBuilder builder,
-            string keyName,
-            string value,
-            IDictionary<string, StringValues> sourceCollection)
+        protected override string GetUniqueId(TagHelperContext context)
```
NTaylorMullen: Re-order methods in this class to be public -> protected -> private etc.

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelperBase.cs
```diff
@@ -0,0 +1,305 @@
...
+
+                // Perf: Avoid allocating enumerator
+                for (var i = 0; i < values.Count; i++)
+                {
+                    var item = values[i];
+
+                    builder
+                        .Append(item)
+                        .Append(CacheKeyTokenSeparator)
+                        .Append(accessor(sourceCollection, item))
```
NTaylorMullen: Do you need the `accessor`? Unless i'm missing something it looks like it always evaluates to `sourceCollection[item]`

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelperBase.cs
```diff
@@ -0,0 +1,305 @@
...
+            }
+        }
+
+        protected static void AddStringCollectionKey(
+            StringBuilder builder,
+            string keyName,
+            string value,
+            IDictionary<string, StringValues> sourceCollection)
+        {
+            if (!string.IsNullOrEmpty(value))
```
NTaylorMullen: Instead of nesting just do:

```C#
{
    return;
}
```

Same with other methods where this applies

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelperBase.cs
NTaylorMullen: Actually, looks like you don't need this method at all.

src/Microsoft.AspNetCore.Mvc.TagHelpers/DependencyInjection/TagHelperExtensions.cs
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Mvc.TagHelpers.Cache;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    public static class TagHelperServicesExtensions
```
NTaylorMullen: Docs

src/Microsoft.AspNetCore.Mvc.TagHelpers/DependencyInjection/TagHelperExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        /// <summary>
+        ///  Adds MVC cache tag helper services to the application.
+        /// </summary>
+        public static IMvcCoreBuilder AddCacheTagHelper(this IMvcCoreBuilder builder)
```
NTaylorMullen: params docs

src/Microsoft.AspNetCore.Mvc.TagHelpers/DependencyInjection/TagHelperExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            AddTagHelperServices(builder.Services);
+            return builder;
+        }
+
+        private static void AddTagHelperServices(IServiceCollection services)
```
NTaylorMullen: Might as well just inline. Doesn't look like you use this elsewhere

src/Microsoft.AspNetCore.Mvc.TagHelpers/DistributedCacheTagHelper.cs
```diff
@@ -0,0 +1,214 @@
...
+    /// <see cref="TagHelper"/> implementation targeting &lt;distributed-cache&gt; elements.
+    /// </summary>
+    [HtmlTargetElement("distributed-cache", Attributes = "name")]
+    public class DistributedCacheTagHelper : CacheTagHelperBase
+    {
+        /// <summary>
+        /// Prefix used by <see cref="DistributedCacheTagHelper"/> instances when creating entries in <see cref="IDistributedCacheTagHelperStorage"/>.
+        /// </summary>
+        public static readonly string CacheKeyPrefix = nameof(DistributedCacheTagHelper);
+        private const string NameName = "name";
```
NTaylorMullen: lmao, rename to `NameAttributeName`

src/Microsoft.AspNetCore.Mvc.TagHelpers/DistributedCacheTagHelper.cs
```diff
@@ -0,0 +1,214 @@
...
+using Microsoft.Extensions.Caching.Distributed;
+using Microsoft.Extensions.Caching.Memory;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNetCore.Mvc.TagHelpers
+{
+    /// <summary>
+    /// <see cref="TagHelper"/> implementation targeting &lt;distributed-cache&gt; elements.
+    /// </summary>
+    [HtmlTargetElement("distributed-cache", Attributes = "name")]
```
NTaylorMullen: `Attributes = NameAttributeName`

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelperBase.cs
sebastienros: Let's ask @rynowak 
https://github.com/aspnet/Mvc/commit/32da2b8c466caa6d423b1502c6bca918397a5425

src/Microsoft.AspNetCore.Mvc.TagHelpers/DistributedCacheTagHelper.cs
```diff
@@ -0,0 +1,214 @@
...
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            if (output == null)
+            {
+                throw new ArgumentNullException(nameof(output));
+            }
+
+            Task<IHtmlContent> result = null;
+            IHtmlContent content = null;
```
NTaylorMullen: Move this down closer to where it's used.

src/Microsoft.AspNetCore.Mvc.TagHelpers/DistributedCacheTagHelper.cs
```diff
@@ -0,0 +1,214 @@
...
+            
+            // Create a cancellation token that will be used 
+            // to release the task from the memory cache.
+            var tokenSource = new CancellationTokenSource();
+
+            if (Enabled)
+            {
+                var key = GenerateKey(context);
+
+                byte[] value = await Storage.GetAsync(key);
```
NTaylorMullen: var

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperFormattingContext.cs
sebastienros: It's not required, it's `Formatter` implementation specific to define if it is.

sebastienros: Using `HtmlString` as the content needs to be already serialized.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperFormatter.cs
sebastienros: Similar answer.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperFormatter.cs
sebastienros: The goal of the extra service is to provide an extensibility point in the html serialization only (gzip, encryption, ...)

src/Microsoft.AspNetCore.Mvc.TagHelpers/DistributedCacheTagHelper.cs
```diff
@@ -0,0 +1,214 @@
...
+
+                if(value == null)
+                {
+                    // Is there any request already processing the value?
+                    if (!MemoryCache.TryGetValue(key, out result))
+                    {
+                        result = Task.Run(async () =>
```
NTaylorMullen: Why not just create a method and then do something like:

```C#
result = CacheEntry(...);
```

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelper.cs
sebastienros: `content` is null when then null TCS is assigned, meaning some worker is already computing the value, and this request needs to await it to get the actual value.

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelper.cs
sebastienros: No, and there is a unit test to ensure it doesn't. Only the _worker_ request processes the content, the other ones are awaiting a TCS.

src/Microsoft.AspNetCore.Mvc.TagHelpers/DistributedCacheTagHelper.cs
```diff
@@ -0,0 +1,214 @@
...
+
+                            var stringBuilder = new StringBuilder();
+                            using (var writer = new StringWriter(stringBuilder))
+                            {
+                                processedContent.WriteTo(writer, HtmlEncoder);
+                            }
+
+                            var formattingContext = new DistributedCacheTagHelperFormattingContext
+                            {
+                                HtmlString = new HtmlString(stringBuilder.ToString()),
```
NTaylorMullen: Hmm, you seem to always have the string/stringbuilder representation of this context. What does wrapping it in an `HtmlString` enable?

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperFormattingContext.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    /// <summary>
+    /// Represents an object containing the information to serialize with <see cref="IDistributedCacheTagHelperFormatter" />.
+    /// </summary>
+    public class DistributedCacheTagHelperFormattingContext
+    {
+        /// <summary>
+        /// Gets the <see cref="TagHelperContext"/> instance.
+        /// </summary>
+        public TagHelperContext Context {get; set; }
```
NTaylorMullen: What scenario does having this property enable?

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelper.cs
sebastienros: Are you sure the `SuppressOutput` call won't revert it? I kept the same flow as the previous version.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperFormatter.cs
NTaylorMullen: :+1:

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelper.cs
sebastienros: This can be the case if two requests enter the `TryGet` before the `Set` is called. Louis' argument is that it's not worth locking it as the probability is really small, and the impact not critical.

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelper.cs
NTaylorMullen: We're just calling `GetChildContentAsync`. That shouldn't revert anything.

NTaylorMullen:
:watch:

src/Microsoft.AspNetCore.Mvc.TagHelpers/DistributedCacheTagHelper.cs
sebastienros: Too many parameters

pranavkm: Should this use a `TaskCompletionSource`? Going by https://channel9.msdn.com/Events/TechEd/Europe/2013/DEV-B318 which recommends libraries against using `Task.Run`.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperStorage.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// <param name="distributedCache">The <see cref="IDistributedCache"/> to use.</param>
+        public DistributedCacheTagHelperStorage(IDistributedCache distributedCache)
+        {
+            _distributedCache = distributedCache;
+        }
+
+        /// <inheritdoc />
+        public Task<byte[]> GetAsync(string key)
+        {
+            return _distributedCache.GetAsync(key);
```
pranavkm: Null check for `key`

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelper.cs
```diff
@@ -156,87 +56,58 @@ public override int Order
...
-                            content.WriteTo(writer, HtmlEncoder);
+                            result = ProcessContentAsync(output);
+                            content = await result;
+                            options.AddEntryLink(link);
                         }
-
-                        result = new StringBuilderHtmlContent(stringBuilder);
-                        MemoryCache.Set(key, result, GetMemoryCacheEntryOptions(link));
+                        
+                        MemoryCache.Set(key, result, options);
```
pranavkm: Why not `TrySetResult` on the TCS instead? Something like this:

```C#
retry:
TaskCompletionSource<IHtmlContent> tcs = null;

if (!MemoryCache.TryGetValue(key, out task))
{
    tcs = new TaskCompletionSource<IHtmlContent>();
    MemoryCache.Set<Task<IHtmlContent>>(key, tcs.Task, options);
}

if (tcs != null)
{
    // Create cache entry
    try
    {
        content = await GetChildContentAsync();
        tcs.TrySetResult(content);
        
        return content;
    }
    catch (Exception ex)
    {
        tcs.TrySetException(ex);
        throw;
    }
}

goto retry;
```

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelperBase.cs
```diff
@@ -0,0 +1,307 @@
...
+        /// </summary>
+        [HtmlAttributeName(EnabledAttributeName)]
+        public bool Enabled { get; set; } = true;
+
+        protected abstract string GetUniqueId(TagHelperContext context);
+
+        protected abstract string GetKeyPrefix(TagHelperContext context);
+
+        // Internal for unit testing
+        internal string GenerateKey(TagHelperContext context)
```
pranavkm: `protected internal`

pranavkm: Could we make the storage support non-string keys? It'd be nice if we can change this code to use a type with `IEquatable<>` to avoid the string allocations.

src/Microsoft.AspNetCore.Mvc.TagHelpers/DistributedCacheTagHelper.cs
```diff
@@ -0,0 +1,194 @@
...
+                        MemoryCache.Set(key, result, options);
+                    }
+                }
+                else
+                {
+                    // If the deserialization fails, it can return null, for instance when the 
+                    // value is not in the expected format.
+                    var deserialized = await Formatter.DeserializeAsync(value);
+                    
+                    if(deserialized != null)
```
pranavkm: Seems like something we should log (if we don't plan on throwing here)

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelperBase.cs
```diff
@@ -0,0 +1,311 @@
...
+using Microsoft.AspNetCore.Mvc.ViewFeatures;
+using Microsoft.AspNetCore.Razor.TagHelpers;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNetCore.Mvc.TagHelpers
+{
+    /// <summary>
+    /// <see cref="TagHelper"/> base implementation for caching elements.
+    /// </summary>
+    public abstract class CacheTagHelperBase : TagHelper
```
lodejard: Should have a CacheKey equatable object. Most of this logic moves to that class, including ToString and hashing. I would strongly recommend not having an abstract base class to share code - favor composition over inheritance.

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelper.cs
sebastienros: I am adding a code comment for clarification. The idea is that the second setter will use the actual options, so that the expiration is not impacted by the time it takes to render the content.

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelperBase.cs
sebastienros: Created the issue for the `CacheKey` https://github.com/aspnet/Mvc/issues/4246

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelperBase.cs
sebastienros: Louis had the same remark, https://github.com/aspnet/Mvc/issues/4246

sebastienros:
:up::date:

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperFormattingContext.cs
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Mvc.Rendering;
+
+namespace Microsoft.AspNetCore.Mvc.TagHelpers.Cache
+{
+    /// <summary>
+    /// Represents an object containing the information to serialize with <see cref="IDistributedCacheTagHelperFormatter" />.
+    /// </summary>
+    public class DistributedCacheTagHelperFormattingContext
```
NTaylorMullen: @lodejard how do you feel about either one of these alternatives:

1. Remove this entirely and bake in the `Key` and `Html` into the `IDistributedCacheTagHelperFormatter.SerializeAsync` signature.
1. Add a constructor to this class and force the user to provide the `Html`/`Key`. Would mean we remove the setter on the properties here.

Both of these scenarios would make it overwhelmingly clear which parameters are required to properly cache something in a distributed cache without the need to do ugly bits like `if (context.Html == null) { throw ... }` in the `IDistributedCacheTagHelperFormatter.SerializeAsync` impl

src/Microsoft.AspNetCore.Mvc.TagHelpers/DependencyInjection/TagHelperExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// <param name="builder">The <see cref="IMvcCoreBuilder"/>.</param>
+        /// <returns>The <see cref="IMvcCoreBuilder"/>.</returns>
+        public static IMvcCoreBuilder AddCacheTagHelper(this IMvcCoreBuilder builder)
+        {
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            builder.Services.TryAddTransient<IDistributedCacheTagHelperStorage, DistributedCacheTagHelperStorage>();
```
NTaylorMullen: Just double checking: We'll be relying on someone else to call addcaching?

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelper.cs
```diff
@@ -156,87 +56,69 @@ public override int Order
...
+                            // The entry is set instead of assigning a value to the 
+                            // task so that the expiration options are are not impacted 
+                            // by the time it took to compute it.
+                            
+                            MemoryCache.Set(key, result, options);
+                            tcs.SetResult(null);
+                        }
+                        catch (Exception ex)
+                        {
+                            tcs.TrySetException(ex);
```
lodejard: Should the other awaiters fail, or should they be given a SetResult(null) so they can take a crack at rendering?

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperFormattingContext.cs
lodejard: I don't mind this class, though I would take it a step further and consider putting the Key, Html, and Data all on the context. SerializeAsync and DeserializeAsync could then ship Html->Data and Html<-Data respectively, and could stash and verify the Key inside the Data as well. Though that last bit is more a part of the next issue dealing with making a proper key class instead of doing string manipulation.

src/Microsoft.AspNetCore.Mvc.TagHelpers/DependencyInjection/TagHelperExtensions.cs
lodejard: oh, yeah - this should also call through to the TryAdd default IMemoryCache/IDistributedCache?

lodejard:
Just added a few last remarks, then :shipit: from me

lodejard:
Oh, also! We need to file a bug on the memory cache - it should link tokens slightly differently than it does - they currently link in the setter and don't roll up the parent property, so it won't work in all cases exactly like you'd want it to.

NTaylorMullen:
:shipit:

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperFormatter.cs
```diff
@@ -0,0 +1,47 @@
...
+        
+        /// <inheritdoc />
+        public Task<byte[]> SerializeAsync(DistributedCacheTagHelperFormattingContext context)
+        {
+            if (context == null) 
+            {
+                throw new ArgumentNullException(nameof(context));    
+            }
+            
+            if (context.Html == null)
```
pranavkm: Can you move this check to the setter?

pranavkm: Oh nevermind. You can instantiate it without ever invoking the setter

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelper.cs
pranavkm: Formatting (space after `while`)

src/Microsoft.AspNetCore.Mvc.TagHelpers/CacheTagHelperBase.cs
```diff
@@ -0,0 +1,311 @@
...
+            // as the actual key (better randomization and key distribution, so small vary
+            // values will generate dramatically different keys).
+            using (var sha = SHA256.Create())
+            {
+                var contentBytes = Encoding.UTF8.GetBytes(builder.ToString());
+                var hashedBytes = sha.ComputeHash(contentBytes);
+                return Convert.ToBase64String(hashedBytes);
+            }
+        }
+        protected abstract string GetUniqueId(TagHelperContext context);
```
pranavkm: New line preceding this.

test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/CacheTagHelperTest.cs
```diff
@@ -777,6 +782,71 @@ public void UpdateCacheEntryOptions_CopiesTriggersFromEntryLink()
...
+        {
+            // Arrange
+            var id = "unique-id";
+            var childContent = "some-content";
+            var resetEvent1 = new ManualResetEvent(false);
+            var resetEvent2 = new ManualResetEvent(false);
+            var calls = 0;
+            var cache = new MemoryCache(new MemoryCacheOptions());
+
+            TagHelperContext tagHelperContext1 = GetTagHelperContext(id + 1);
```
pranavkm: `var`

test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/CacheTagHelperTest.cs
```diff
@@ -777,6 +782,71 @@ public void UpdateCacheEntryOptions_CopiesTriggersFromEntryLink()
...
+            {
+                ViewContext = GetViewContext(),
+                Enabled = true
+            };
+
+            // Act
+
+            var task1 = Task.Run(async () =>
+            {
+                resetEvent1.WaitOne();
```
pranavkm: Can you add the wait duration here so the test doesn't hang forever in case something goes wrong? ~5-10 seconds should be good.

test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/DistributedCacheTagHelperTest.cs
```diff
@@ -0,0 +1,926 @@
...
+            var key = cacheTagHelper.GenerateKey(tagHelperContext);
+
+            // Assert
+            Assert.Equal(expected, key);
+        }
+
+        [Theory]
+        [InlineData("Cookie0", "DistributedCacheTagHelper||some-name||VaryByCookie(Cookie0||Cookie0Value)")]
+        [InlineData("Cookie0,Cookie1",
+                    "DistributedCacheTagHelper||some-name||VaryByCookie(Cookie0||Cookie0Value||Cookie1||Cookie1Value)")]
```
pranavkm: Use VS default indentation (4 spaces from `[` on previous line)

test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/DistributedCacheTagHelperTest.cs
```diff
@@ -0,0 +1,926 @@
...
+                Mock.Of<IDistributedCacheTagHelperStorage>(), 
+                Mock.Of<IDistributedCacheTagHelperFormatter>(), 
+                new HtmlTestEncoder())
+            {
+                ViewContext = GetViewContext(),
+                VaryByQuery = varyByQuery,
+                Name = name
+            };
+            cacheTagHelper.ViewContext.HttpContext.Request.QueryString =
+                new Http.QueryString("?sortoption=Adorability&Category=cats&sortOrder=");
```
pranavkm: Does `QueryString` conflict with other types? If not, move the `Http.` to a using.

test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/DistributedCacheTagHelperTest.cs
```diff
@@ -0,0 +1,926 @@
...
+            Assert.Equal(GetHashedBytes(expected), key);
+        }
+
+        [Fact]
+        public void GenerateKey_WithMultipleVaryByOptions_CreatesCombinedKey()
+        {
+            // Arrange
+            var name = "some-name";
+            var expected = GetHashedBytes("DistributedCacheTagHelper||some-name||VaryBy||custom-value||" +
+                                          "VaryByHeader(content-type||text/html)||VaryByUser||someuser");
```
pranavkm: Indentation.

test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/DistributedCacheTagHelperTest.cs
```diff
@@ -0,0 +1,926 @@
...
+        }
+
+        [Fact]
+        public async Task ProcessAsync_DoesNotCache_IfDisabled()
+        {
+            // Arrange
+            var childContent = "original-child-content";
+            var memoryCache = new MemoryCache(new MemoryCacheOptions());
+            var storage = new Mock<IDistributedCacheTagHelperStorage>();
+            storage.CallBase = true;
```
pranavkm: Shouldn't have to `Callbase` to an interface

test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/DistributedCacheTagHelperTest.cs
```diff
@@ -0,0 +1,926 @@
...
+                ViewContext = GetViewContext(),
+                Enabled = false
+            };
+
+            // Act
+            await cacheTagHelper.ProcessAsync(tagHelperContext, tagHelperOutput);
+
+            // Assert
+            Assert.Equal(childContent, tagHelperOutput.Content.GetContent());
+            storage.Verify(c => c.SetAsync(
```
pranavkm: `storage.Verify();`

pranavkm: Didn't see the `Times.Never`. In which case, remove the `Setup \ Verifiable` call from Line 318

test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/DistributedCacheTagHelperTest.cs
```diff
@@ -0,0 +1,926 @@
...
+        [Fact]
+        public async Task ProcessAsync_ReturnsCachedValue_IfEnabled()
+        {
+            // Arrange
+            var childContent = "original-child-content";
+            var memoryCache = new MemoryCache(new MemoryCacheOptions());
+            var storage = new Mock<IDistributedCacheTagHelperStorage>();
+            storage.CallBase = true;
+            var value = Encoding.UTF8.GetBytes(childContent);
+            storage.Setup(c => c.SetAsync(
```
pranavkm: Remove this since you're using `.Verify` later on.

test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/DistributedCacheTagHelperTest.cs
```diff
@@ -0,0 +1,926 @@
...
+        
+        [Fact]
+        public async Task ProcessAsync_UsesExpiresAfter_ToExpireCacheEntry()
+        {
+            // Arrange - 1
+            var currentTime = new DateTimeOffset(2010, 1, 1, 0, 0, 0, TimeSpan.Zero);
+            var childContent1 = "original-child-content";
+            var clock = new Mock<ISystemClock>();
+            clock.SetupGet(p => p.UtcNow)
+                 .Returns(() => currentTime);
```
pranavkm: Indentation.

pranavkm: Couple of other instances of this in tests that follow.

test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/DistributedCacheTagHelperTest.cs
```diff
@@ -0,0 +1,926 @@
...
+        }
+
+        [Fact]
+        public async Task ProcessAsync_ComputesValueOnce_WithConcurrentRequests()
+        {
+            // Arrange
+            var childContent = "some-content";
+            var resetEvent1 = new ManualResetEvent(false);
+            var resetEvent2 = new ManualResetEvent(false);
+            int calls = 0;
```
pranavkm: `var`

test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/DistributedCacheTagHelperTest.cs
```diff
@@ -0,0 +1,926 @@
...
+            var formatter = GetFormatter();
+            var memoryCache = new MemoryCache(new MemoryCacheOptions());
+            var storage = GetStorage();
+            TagHelperContext tagHelperContext1 = GetTagHelperContext();
```
pranavkm: `var` all day every day.

test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/DistributedCacheTagHelperTest.cs
```diff
@@ -0,0 +1,926 @@
...
+            var deserialized = await formatter.DeserializeAsync(serialized);
+            
+            // Assert
+            Assert.Equal(deserialized.ToString(), content);
+        }
+
+        private static ViewContext GetViewContext()
+        {
+            var actionContext = new ActionContext(new DefaultHttpContext(), new RouteData(), new ActionDescriptor());
+            return new ViewContext(actionContext,
```
pranavkm: Indentation.

pranavkm:
:shipit: once you've addressed formatting.

test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/DistributedCacheTagHelperTest.cs
sebastienros: Just for fun, can you do a git blame on the original file? Enjoy!

pranavkm: We switched the indentation rules half way thru :8ball:

sebastienros:
:up::date: I also added two unit tests to ensure that when an exception is thrown by the worker thread, the other awaiting requests are processing the output themselves.

sebastienros:
:up::date:

Introduced a new service to handle gating state, instead of using `IMemoryCache`.

sebastienros:
Announcements related to this change:
https://github.com/aspnet/Announcements/issues/160
https://github.com/aspnet/Announcements/issues/161

PR #4243, opened 2016-03-07
Revive DefaultAssemblyProviderTests

pranavkm:
Partial fix for https://github.com/aspnet/Mvc/issues/4140

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc/project.json
+++ b/src/Microsoft.AspNetCore.Mvc/project.json
@@ -11,7 +11,8 @@
     "nowarn": [
       "CS1591"
     ],
-    "xmlDoc": true
+    "xmlDoc": true,
+    "preserveCompilationContext": true
   },
   "dependencies": {
     "Microsoft.AspNetCore.Mvc.ApiExplorer": "1.0.0-*",
--- a/test/Microsoft.AspNetCore.Mvc.Core.Test/Infrastructure/DefaultAssemblyProviderTests.cs
+++ b/test/Microsoft.AspNetCore.Mvc.Core.Test/Infrastructure/DefaultAssemblyProviderTests.cs
@@ -1,0 +1,216 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyModel;
+using Xunit;
+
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNetCore.Mvc/project.json
```diff
@@ -11,7 +11,8 @@
     "nowarn": [
       "CS1591"
     ],
-    "xmlDoc": true
+    "xmlDoc": true,
+    "preserveCompilationContext": true
```
pranavkm: cc @pakrym \ @davidfowl. Is this a terrible thing to do?

pranavkm:
cc @NTaylorMullen

src/Microsoft.AspNetCore.Mvc/project.json
pakrym: Yes, why are you doing it anyway?

pranavkm: We have a test that verifies if a list of package names we have is near identical to this assembly's dependencies. In the past we used the `ILibraryManager` to discover these dependencies.

test/Microsoft.AspNetCore.Mvc.Core.Test/Infrastructure/DefaultAssemblyProviderTests.cs
```diff
@@ -0,0 +1,216 @@
...
+        }
+
+        private static RuntimeLibrary GetLibrary(string name, params string[] dependencyNames)
+        {
+            var dependencies = dependencyNames?.Select(d => new Dependency(d, "1.0.0")) ?? new Dependency[0];
+
+            return new RuntimeLibrary(
+                "package",
+                name,
+                "1.0.0",
```
NTaylorMullen: Anyway we can grab the version in a way that wont require code updates every time we ship?

pranavkm: We don't look at versions at all in our provider. I could change it to some random version if that makes it easier to grock.

src/Microsoft.AspNetCore.Mvc/project.json
pranavkm: I guess I can add this flag to the test and exclude any test assembly.

test/Microsoft.AspNetCore.Mvc.Core.Test/Infrastructure/DefaultAssemblyProviderTests.cs
NTaylorMullen: Ya make it random so it doesn't raise questions

NTaylorMullen:
:shipit:

PR #4249, opened 2016-03-08
[Fixes #4237] Include Assembly-Qualified type name in ActionDescriptor.DisplayName

### LINKED ISSUE #4237: Include Assembly-Qualified Type Name in ActionDescriptor.DisplayName
`DisplayName` is used when we have startup/configuration-related errors or a runtime action selection ambiguity. A frequent cause of these kinds of ambiguities is after an assembly rename - you can have old build outputs sitting around. You can't really diagnose that from the message with the fully-qualified name, you have to go assembly-qualified.

javiercn:


rynowak:
No test update?

Code change looks right, but I'd expect to see some test updates for this.

javiercn:
I was having trouble restoring yesterday. Will update tests today, I don't expect a lot of churn there.

rynowak:
Provisional :shipit: if the changes to tests are straightforward.

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ControllerActionDescriptorProviderTests.cs
```diff
@@ -38,6 +38,20 @@ public void GetDescriptors_GetsDescriptorsOnlyForValidActions()
...
+            // Arrange
+            var provider = GetProvider(typeof(PersonController).GetTypeInfo());
+
+            // Act
+            var descriptors = provider.GetDescriptors();
+            var descriptor = descriptors.Single(ad => ad.Name == nameof(PersonController.GetPerson));
+
+            // Assert
+            Assert.Equal($"Microsoft.AspNetCore.Mvc.Internal.ControllerActionDescriptorProviderTests+PersonController, Microsoft.AspNetCore.Mvc.Core.Test, Version=1.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60.GetPerson", descriptor.DisplayName);
+        }
```
javiercn: @rynowak This is how  the method name would look with the AssemblyQualifiedName, I find it a bit ugly for error messages, maybe we could do something like "GetPerson - (Microsoft.AspNetCore.Mvc.Internal.ControllerActionDescriptorProviderTests+PersonController, Microsoft.AspNetCore.Mvc.Core.Test, Version=1.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60)"

javiercn:
@rynowak Updated, I think this looks way more reasonable

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionDescriptorBuilder.cs
```diff
@@ -739,11 +733,13 @@ private static bool IsAttributeRoutedAction(ControllerActionDescriptor actionDes
...
+            var declaringType = actionDescriptor.MethodInfo.DeclaringType.GetTypeInfo();
             var methodFullName = string.Format(
                 CultureInfo.InvariantCulture,
-                "{0}.{1}",
-                actionDescriptor.MethodInfo.DeclaringType.FullName,
-                actionDescriptor.MethodInfo.Name);
+                "{0}.{1} ({2})",
+                declaringType.FullName,
+                actionDescriptor.MethodInfo.Name,
+                declaringType.Assembly.GetName().Name);
```
rynowak: Why can't we just `DisplayName` here?

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ControllerActionDescriptorProviderTests.cs
```diff
@@ -38,6 +38,20 @@ public void GetDescriptors_GetsDescriptorsOnlyForValidActions()
...
+        {
+            // Arrange
+            var provider = GetProvider(typeof(PersonController).GetTypeInfo());
+
+            // Act
+            var descriptors = provider.GetDescriptors();
+            var descriptor = descriptors.Single(ad => ad.Name == nameof(PersonController.GetPerson));
+
+            // Assert
+            Assert.Equal($"Microsoft.AspNetCore.Mvc.Internal.ControllerActionDescriptorProviderTests+PersonController.GetPerson (Microsoft.AspNetCore.Mvc.Core.Test)", descriptor.DisplayName);
```
pranavkm: Could we do `{typeof(PersonControllerType).FullName}.{nameof(PersonController.GetPerson)}` here?

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionDescriptorBuilder.cs
javiercn: Fair enough, I was actually going to comment that it was weird that we were using DeclaringType

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ControllerActionDescriptorProviderTests.cs
javiercn: Sure

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionDescriptorBuilder.cs
javiercn: What about the InvariantCulture part, do we care?

src/Microsoft.AspNetCore.Mvc.Core/Controllers/ControllerActionDescriptor.cs
```diff
@@ -24,9 +24,10 @@ public override string DisplayName
                 if (base.DisplayName == null && ControllerTypeInfo != null && MethodInfo != null)
                 {
                     base.DisplayName = string.Format(
```
rynowak: This should be using invariant culture

rynowak:
:shipit: after adding invariant culture to our string.format.

We should (as a rule) always be passing a culture/comparer into string methods.

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ControllerActionDescriptorProviderTests.cs
```diff
@@ -38,6 +38,20 @@ public void GetDescriptors_GetsDescriptorsOnlyForValidActions()
...
+        {
+            // Arrange
+            var provider = GetProvider(typeof(PersonController).GetTypeInfo());
+
+            // Act
+            var descriptors = provider.GetDescriptors();
+            var descriptor = descriptors.Single(ad => ad.Name == nameof(PersonController.GetPerson));
+
+            // Assert
+            Assert.Equal($"{typeof(PersonController).FullName}.{nameof(PersonController.GetPerson)} (Microsoft.AspNetCore.Mvc.Core.Test)", descriptor.DisplayName);
```
pranavkm: Maybe the assembly name too. `var assemblyName = GetType().GetTypeInfo().Assembly.GetName().Name;` (in case this assembly ever gets renamed)

pranavkm:
:shipit:

src/Microsoft.AspNetCore.Mvc.Core/Controllers/ControllerActionDescriptor.cs
```diff
@@ -23,10 +24,11 @@ public override string DisplayName
             {
                 if (base.DisplayName == null && ControllerTypeInfo != null && MethodInfo != null)
                 {
-                    base.DisplayName = string.Format(
-                        "{0}.{1}",
+                    base.DisplayName = string.Format(CultureInfo.InvariantCulture,
```
rynowak: #YOLOBREAK

PR #4250, opened 2016-03-08
[Fixes #4102] DefaultHtmlGenerator should attempt to properly format values for `<input type="date">` and `<input type="time">`

### LINKED ISSUE #4102: DefaultHtmlGenerator should attempt to properly format values for `<input type="date">` and `<input type="time">`
When binding a `DateTime` model value to an `<input type="date" />` element or an `<input type="time">` element, one would expect the output to be a [valid date string according to the HTML5 specification.](https://www.w3.org/TR/html5/single-page.html#valid-date-string)

Likewise, when binding a `DateTime` or `TimeSpan` model value to an `<input type="time" />` element, one would expect the output to be a [valid time string according to the HTML5 specification.](https://www.w3.org/TR/html5/single-page.html#valid-time-string)

javiercn:


dougbu:
:watch: Nice but I wanna see test delta.

javiercn:
@dougbu Test changes are included now

dougbu:
:shipit: (surprised how little we're using `DateTime[Offset]` properties with the default formatting mode)

Suggest building in the Entropy repo with this change pulled in (using `global.json`). Tests there may be affected.

PR #4253, opened 2016-03-08
React to aspnet/Razor#684.

NTaylorMullen:
- Modified `UrlResolutionTagHelper` to utilize new CSS selector attributes to restrict when it applies. It now only appies to tags that have their values starting with `~/`.
- `UrlResolutionTagHelper` no longer applies to dynamic content such as `href="@SomethingResultingInTildaSlash"`.
- Updated tests to reflect new behavior.


NTaylorMullen:
FYI mostly.

dougbu:
:shipit:

PR #4255, opened 2016-03-09
[Perf] Optimize creation of VDD

### LINKED ISSUE #3906: Optimize creation of VDD
Here's an execution profile from an MVC app with a fairly substantial Razor Page:

image

This use of the Activator for VDD is costing about 3% of total execution time, we can find a way to do better.

mnltejaswini:
Fixes #3906 
ViewDataDictionary of TModel type is now created using the cached delegate as opposed to Activator.CreateInstance for improved performance. 
Tagged as Design because build fails with unrelated errors.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor/RazorPageActivator.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor/RazorPageActivator.cs
@@ -3,7 +3,9 @@
 
 using System;
 using System.Collections.Concurrent;
+using System.Diagnostics;
 using System.Reflection;
+using System.Linq.Expressions;
 using Microsoft.AspNetCore.Mvc.ModelBinding;
 using Microsoft.AspNetCore.Mvc.Razor.Internal;
 using Microsoft.AspNetCore.Mvc.Rendering;
@@ -17,6 +19,11 @@ namespace Microsoft.AspNetCore.Mvc.Razor
     /// <inheritdoc />
     public class RazorPageActivator : IRazorPageActivator
     {
+        private delegate ViewDataDictionary CreateViewDataNested(ViewDataDictionary source);
+        private delegate ViewDataDictionary CreateViewDataRoot(
+            IModelMetadataProvider metadataProvider,
+            ModelStateDictionary modelState);
+
         // Name of the "public TModel Model" property on RazorPage<TModel>
         private const string ModelPropertyName = "Model";
         private readonly ConcurrentDictionary<Type, PageActivationInfo> _activationInfo;
@@ -63,17 +70,14 @@ namespace Microsoft.AspNetCore.Mvc.Razor
             if (context.ViewData == null)
             {
                 // Create ViewDataDictionary<TModel>(IModelMetadataProvider, ModelStateDictionary).
-                return (ViewDataDictionary)Activator.CreateInstance(
-                    activationInfo.ViewDataDictionaryType,
+                return activationInfo.CreateViewDataRoot(
                     _metadataProvider,
                     context.ModelState);
             }
             else if (context.ViewData.GetType() != activationInfo.ViewDataDictionaryType)
             {
                 // Create ViewDataDictionary<TModel>(ViewDataDictionary).
-                return (ViewDataDictionary)Activator.CreateInstance(
-                    activationInfo.ViewDataDictionaryType,
-                    context.ViewData);
+                return activationInfo.CreateViewDataNested(context.ViewData);
             }
 
             return context.ViewData;
@@ -96,6 +100,8 @@ namespace Microsoft.AspNetCore.Mvc.Razor
             return new PageActivationInfo
             {
                 ViewDataDictionaryType = viewDataType,
+                CreateViewDataNested = GetCreateViewDataNested(viewDataType),
+                CreateViewDataRoot = GetCreateViewDataRoot(viewDataType),
                 PropertyActivators = PropertyActivator<ViewContext>.GetPropertiesToActivate(
                     type,
                     typeof(RazorInjectAttribute),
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Mvc.ViewFeatures/Resources.resx
```diff
@@ -280,4 +280,7 @@
   <data name="ValueInterfaceAbstractOrOpenGenericTypesCannotBeActivated" xml:space="preserve">
     <value>The type '{0}' cannot be activated by '{1}' because it is either a value type, an interface, an abstract class or an open generic type.</value>
   </data>
+  <data name="ViewData_CannotCreateOfTModelType" xml:space="preserve">
+    <value>Cannot find matching constructor to create ViewDataDictionary of type '{0}'.</value>
+  </data>
```
rynowak: When will this error ever be surfaced? This seems like it should just be a debug.assert.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ViewDataDictionaryFactory.cs
```diff
@@ -0,0 +1,105 @@
...
+using Microsoft.AspNetCore.Mvc.ModelBinding;
+
+namespace Microsoft.AspNetCore.Mvc.ViewFeatures
+{
+    public static class ViewDataDictionaryFactory
+    {
+        internal delegate object ViewDataDictionaryCreator(object[] arguments);
+
+        private static readonly ConcurrentDictionary<ConstructorInfo, ViewDataDictionaryCreator> 
+            _viewDataDictionaryCreateDelegateCache = new ConcurrentDictionary<ConstructorInfo, ViewDataDictionaryCreator>();
```
rynowak: We don't want to add another (separate) cache if possible. RazorPageActivator already has a cache - just include this with the existing data.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ViewDataDictionaryFactory.cs
```diff
@@ -0,0 +1,105 @@
...
+using System.Collections.Generic;
+using System.Reflection;
+using System.Linq.Expressions;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
+
+namespace Microsoft.AspNetCore.Mvc.ViewFeatures
+{
+    public static class ViewDataDictionaryFactory
+    {
+        internal delegate object ViewDataDictionaryCreator(object[] arguments);
```
rynowak: No internal

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ViewDataDictionaryFactory.cs
```diff
@@ -0,0 +1,105 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Reflection;
+using System.Linq.Expressions;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
+
+namespace Microsoft.AspNetCore.Mvc.ViewFeatures
```
rynowak: `Microsoft.AspNetCore.Mvc.ViewFeatures.Internal`

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ViewDataDictionaryFactory.cs
```diff
@@ -0,0 +1,105 @@
...
+                var valueCast = Expression.Convert(valueObj, paramInfo.ParameterType);
+
+                // valueCast is "(Ti) parameters[i]"
+                parameters.Add(valueCast);
+            }
+
+            var newExp = Expression.New(constructor, parameters);
+            var castMethodCall = Expression.Convert(newExp, typeof(object));
+            var lambda = Expression.Lambda<ViewDataDictionaryCreator>(castMethodCall, parametersParameter);
+            return lambda.Compile();
```
rynowak: Instead of making these take an array, let's just make two delegate types

rynowak:
:watch: Direction looks good. Would like to see the cache on `RazorPageActivator` used instead, and a little cleanup avoiding the use of `object[]` that needs to be allocated per-request

src/Microsoft.AspNetCore.Mvc.Razor/RazorPageActivator.cs
```diff
@@ -63,7 +63,7 @@ private ViewDataDictionary CreateViewDataDictionary(ViewContext context, PageAct
             if (context.ViewData == null)
             {
                 // Create ViewDataDictionary<TModel>(IModelMetadataProvider, ModelStateDictionary).
-                return (ViewDataDictionary)Activator.CreateInstance(
+                return ViewDataDictionaryFactory.Create(
```
pranavkm: Would using `ActivatorUtilities` with a null `IServiceProvider` work? Alternatively could we add an overload to it that supports it? Seems like we might be able to reuse a bunch of code if we do that.

rynowak: I don't want to mix type activation with this. We're also **already** doing a cache lookup in this code, I don't want to add more.

mnltejaswini:
Perf comparisons
--------------------------------------
Testapp: https://github.com/aspnet/Performance/tree/dev/testapp/LargeStaticView
loadtest -k -n 10000 -c 250 http://localhost:5000/TagHelpers

Before fix
------------
RazorPageActivator.Activate takes ~448ms of CPU time and the hot spot is 
image


After fix
---------
RazorPageActivator takes ~178ms of CPU time and the hot spot is no longer Activator.CreateInstance

image

src/Microsoft.AspNetCore.Mvc.Razor/RazorPageActivator.cs
```diff
@@ -17,6 +18,10 @@ namespace Microsoft.AspNetCore.Mvc.Razor
     /// <inheritdoc />
     public class RazorPageActivator : IRazorPageActivator
     {
+        // Delegates to create ViewDataDictionary of TModel
```
rynowak: This comment doesn't explain enough to be useful. Either explain **why** this is special or just remove the comment.

src/Microsoft.AspNetCore.Mvc.Razor/RazorPageActivator.cs
```diff
@@ -17,6 +18,10 @@ namespace Microsoft.AspNetCore.Mvc.Razor
...
+        private delegate ViewDataDictionary ViewDataCreateFromSource(ViewDataDictionary source);
+        private delegate ViewDataDictionary ViewDataCreateFromModel(IModelMetadataProvider metadataProvider, ModelStateDictionary modelState);
```
rynowak: Suggestions for naming 

`ViewDataCreateFromSource` -> `CreateViewDataNested`
`ViewDataCreateFromModel` -> `CreateViewDataRoot`

src/Microsoft.AspNetCore.Mvc.Razor/RazorPageActivator.cs
```diff
@@ -104,6 +108,41 @@ private PageActivationInfo CreateViewActivationInfo(Type type)
...
         }
 
+        private ViewDataCreateFromSource GetViewDataCreatorFromSource(Type pageType, Type viewDataDictionaryType)
+        {
+            var parameterTypes = new Type[] { typeof(ViewDataDictionary) };
+            var matchingConstructor = viewDataDictionaryType.GetConstructor(parameterTypes);
+            if (matchingConstructor == null)
+            {
+                var message = Resources.FormatViewCannotBeActivated(pageType.FullName, GetType().FullName);
+                throw new InvalidOperationException();
```
rynowak: Why and when would a user ever see this exception? This should be a `Debug.Assert`

src/Microsoft.AspNetCore.Mvc.Razor/RazorPageActivator.cs
```diff
@@ -104,6 +108,41 @@ private PageActivationInfo CreateViewActivationInfo(Type type)
...
+            }
+
+            var parameters = new ParameterExpression[] { Expression.Parameter(parameterTypes[0]) };
+            var newExp = Expression.New(matchingConstructor, parameters);
```
rynowak: don't abbreviate in names - `newExpression`

rynowak:
:watch:

rynowak:
:shipit: - looks good!

Issue #4258, opened 2016-03-09
Consider adding SignInResult/SignOutResult and SignIn/SignOut


PinpointTownes:
MVC 6 has really cool authentication-specific `ActionResult`s and helpers for `Forbid` and `Challenge`, but nothing for `SignIn` and `SignOut`.

It would really make this kind of actions much clearer:

```csharp
[HttpPost("~/connect/logout"), ValidateAntiForgeryToken]
public async Task<IActionResult> Logout() {
    var response = HttpContext.GetOpenIdConnectResponse();
    if (response != null) {
        return View("Error", response);
    }

    // ...

    // Ask the cookies middleware to delete the local cookie created
    // when the user agent is redirected from the external identity provider
    // after a successful authentication flow (e.g Google or Facebook).
    await HttpContext.Authentication.SignOutAsync("ServerCookie");

    // Redirect the user agent to the post_logout_redirect_uri specified by the client application.
    await HttpContext.Authentication.SignOutAsync(OpenIdConnectServerDefaults.AuthenticationScheme);

    return new EmptyResult();
}
```

-> 

```csharp
[HttpPost("~/connect/logout"), ValidateAntiForgeryToken]
public async Task<IActionResult> Logout() {
    var response = HttpContext.GetOpenIdConnectResponse();
    if (response != null) {
        return View("Error", response);
    }

    // ...

    return SignOut("ServerCookie", OpenIdConnectServerDefaults.AuthenticationScheme);
}
```

I'm fine submitting a PR if you like the general idea :clap:

rynowak:
If it would make the templated controller code more understandable/declarative then it seems like a good idea. I'm not familiar enough with this part of the system on my own to understand how much of this could be refactored into a result type.

@blowdart thoughts?

louislewis2:
I like the general thoughts behind this idea. Seeing this included would certainly be a plus.

PinpointTownes:
For some reasons, I'm seeing weird issues with DependencyModel in the MVC tests.
Restoring doesn't seem to help. I'll investigate.

![image](https://cloud.githubusercontent.com/assets/6998306/13643380/c3a1e9a4-e620-11e5-9120-a05e6fab62c5.png)

PinpointTownes:
```js
"Microsoft.Extensions.DependencyModel/1.0.0-dev-1452673922": {
  "type": "package",
  "dependencies": {
    "Newtonsoft.Json": "7.0.1",
[...]

PR #4259, opened 2016-03-09
Introduce SignInResult/SignOutResult and SignIn/SignOut

PinpointTownes:
Fixes https://github.com/aspnet/Mvc/issues/4258.

(FWIW, that would be great if the authentication helpers had overloads accepting a `params string[]` argument for `authenticationSchemes`).

/cc @rynowak @HaoK @pranavkm

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/ChallengeResult.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/ChallengeResult.cs
@@ -101,8 +101,10 @@ namespace Microsoft.AspNetCore.Mvc
             var loggerFactory = context.HttpContext.RequestServices.GetRequiredService<ILoggerFactory>();
             var logger = loggerFactory.CreateLogger<ChallengeResult>();
 
+            logger.ChallengeResultExecuting(AuthenticationSchemes);
+
             var authentication = context.HttpContext.Authentication;
-            if (AuthenticationSchemes.Count > 0)
+            if (AuthenticationSchemes != null && AuthenticationSchemes.Count > 0)
             {
                 foreach (var scheme in AuthenticationSchemes)
                 {
@@ -113,8 +115,6 @@ namespace Microsoft.AspNetCore.Mvc
             {
                 await authentication.ChallengeAsync(Properties);
             }
-
-            logger.ChallengeResultExecuting(AuthenticationSchemes);
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Mvc.Core/ControllerBase.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/ControllerBase.cs
@@ -924,21 +924,12 @@ namespace Microsoft.AspNetCore.Mvc
             => new ChallengeResult();
 
         /// <summary>
-        /// Creates a <see cref="ChallengeResult"/> with the specified authentication scheme.
-        /// </summary>
-        /// <param name="authenticationScheme">The authentication scheme to challenge.</param>
-        /// <returns>The created <see cref="ChallengeResult"/> for the response.</returns>
-        [NonAction]
-        public virtual ChallengeResult Challenge(string authenticationScheme)
-            => new ChallengeResult(authenticationScheme);
-
-        /// <summary>
         /// Creates a <see cref="ChallengeResult"/> with the specified authentication schemes.
         /// </summary>
         /// <param name="authenticationSchemes">The authentication schemes to challenge.</param>
         /// <returns>The created <see cref="ChallengeResult"/> for the response.</returns>
         [NonAction]
-        public virtual ChallengeResult Challenge(IList<string> authenticationSchemes)
+        public virtual ChallengeResult Challenge(params string[] authenticationSchemes)
             => new ChallengeResult(authenticationSchemes);
 
         /// <summary>
@@ -952,29 +943,17 @@ namespace Microsoft.AspNetCore.Mvc
             => new ChallengeResult(properties);
 
         /// <summary>
-        /// Creates a <see cref="ChallengeResult"/> with the specified specified authentication scheme and
-        /// <paramref name="properties" />.
-        /// </summary>
-        /// <param name="authenticationScheme">The authentication scheme to challenge.</param>
-        /// <param name="properties"><see cref="AuthenticationProperties"/> used to perform the authentication
-        /// challenge.</param>
-        /// <returns>The created <see cref="ChallengeResult"/> for the response.</returns>
-        [NonAction]
-        public virtual ChallengeResult Challenge(string authenticationScheme, AuthenticationProperties properties)
-            => new ChallengeResult(authenticationScheme, properties);
-
-        /// <summary>
         /// Creates a <see cref="ChallengeResult"/> with the specified specified authentication schemes and
         /// <paramref name="properties" />.
         /// </summary>
-        /// <param name="authenticationSchemes">The authentication schemes to challenge.</param>
         /// <param name="properties"><see cref="AuthenticationProperties"/> used to perform the authentication
         /// challenge.</param>
+        /// <param name="authenticationSchemes">The authentication schemes to challenge.</param>
         /// <returns>The created <see cref="ChallengeResult"/> for the response.</returns>
         [NonAction]
         public virtual ChallengeResult Challenge(
-            IList<string> authenticationSchemes,
-            AuthenticationProperties properties)
+            AuthenticationProperties properties,
+            params string[] authenticationSchemes)
             => new ChallengeResult(authenticationSchemes, properties);
 
         /// <summary>
@@ -986,21 +965,12 @@ namespace Microsoft.AspNetCore.Mvc
             => new ForbidResult();
 
         /// <summary>
-        /// Creates a <see cref="ForbidResult"/> with the specified authentication scheme.
-        /// </summary>
-        /// <param name="authenticationScheme">The authentication scheme to challenge.</param>
-        /// <returns>The created <see cref="ForbidResult"/> for the response.</returns>
-        [NonAction]
-        public virtual ForbidResult Forbid(string authenticationScheme)
-            => new ForbidResult(authenticationScheme);
-
-        /// <summary>
         /// Creates a <see cref="ForbidResult"/> with the specified authentication schemes.
         /// </summary>
         /// <param name="authenticationSchemes">The authentication schemes to challenge.</param>
         /// <returns>The created <see cref="ForbidResult"/> for the response.</returns>
         [NonAction]
-        public virtual ForbidResult Forbid(IList<string> authenticationSchemes)
+        public virtual ForbidResult Forbid(params string[] authenticationSchemes)
             => new ForbidResult(authenticationSchemes);
 
         /// <summary>
@@ -1014,28 +984,61 @@ namespace Microsoft.AspNetCore.Mvc
             => new ForbidResult(properties);
 
         /// <summary>
-        /// Creates a <see cref="ForbidResult"/> with the specified specified authentication scheme and
-        /// <paramref name="properties" />.
-        /// </summary>
-        /// <param name="authenticationScheme">The authentication scheme to challenge.</param>
-        /// <param name="properties"><see cref="AuthenticationProperties"/> used to perform the authentication
-        /// challenge.</param>
-        /// <returns>The created <see cref="ForbidResult"/> for the response.</returns>
-        [NonAction]
-        public virtual ForbidResult Forbid(string authenticationScheme, AuthenticationProperties properties)
-            => new ForbidResult(authenticationScheme, properties);
-
-        /// <summary>
         /// Creates a <see cref="ForbidResult"/> with the specified specified authentication schemes and
         /// <paramref name="properties" />.
         /// </summary>
-        /// <param name="authenticationSchemes">The authentication schemes to challenge.</param>
         /// <param name="properties"><see cref="AuthenticationProperties"/> used to perform the authentication
         /// challenge.</param>
+        /// <param name="authenticationSchemes">The authentication schemes to challenge.</param>
         /// <returns>The created <see cref="ForbidResult"/> for the response.</returns>
         [NonAction]
-        public virtual ForbidResult Forbid(IList<string> authenticationSchemes, AuthenticationProperties properties)
+        public virtual ForbidResult Forbid(AuthenticationProperties properties, params string[] authenticationSchemes)
             => new ForbidResult(authenticationSchemes, properties);
+
+        /// <summary>
+        /// Creates a <see cref="SignInResult"/> with the specified authentication scheme.
+        /// </summary>
+        /// <param name="principal">The <see cref="ClaimsPrincipal"/> containing the user claims.</param>
+        /// <param name="authenticationScheme">The authentication scheme to use for the sign-in operation.</param>
+        /// <returns>The created <see cref="SignInResult"/> for the response.</returns>
+        [NonAction]
+        public virtual SignInResult SignIn(ClaimsPrincipal principal, string authenticationScheme)
+            => new SignInResult(authenticationScheme, principal);
+
+        /// <summary>
+        /// Creates a <see cref="SignInResult"/> with the specified specified authentication scheme and
+        /// <paramref name="properties" />.
+        /// </summary>
+        /// <param name="principal">The <see cref="ClaimsPrincipal"/> containing the user claims.</param>
+        /// <param name="properties"><see cref="AuthenticationProperties"/> used to perform the sign-in operation.</param>
+        /// <param name="authenticationScheme">The authentication scheme to use for the sign-in operation.</param>
+        /// <returns>The created <see cref="SignInResult"/> for the response.</returns>
+        [NonAction]
+        public virtual SignInResult SignIn(
+            ClaimsPrincipal principal,
+            AuthenticationProperties properties,
+            string authenticationScheme)
+            => new SignInResult(authenticationScheme, principal, properties);
+
+        /// <summary>
+        /// Creates a <see cref="SignOutResult"/> with the specified authentication schemes.
+        /// </summary>
+        /// <param name="authenticationSchemes">The authentication schemes to use for the sign-out operation.</param>
+        /// <returns>The created <see cref="SignOutResult"/> for the response.</returns>
+        [NonAction]
+        public virtual SignOutResult SignOut(params string[] authenticationSchemes)
+            => new SignOutResult(authenticationSchemes);
+
+        /// <summary>
+        /// Creates a <see cref="SignOutResult"/> with the specified specified authentication schemes and
+        /// <paramref name="properties" />.
+        /// </summary>
+        /// <param name="properties"><see cref="AuthenticationProperties"/> used to perform the sign-out operation.</param>
+        /// <param name="authenticationSchemes">The authentication scheme to use for the sign-out operation.</param>
+        /// <returns>The created <see cref="SignOutResult"/> for the response.</returns>
+        [NonAction]
+        public virtual SignOutResult SignOut(AuthenticationProperties properties, params string[] authenticationSchemes)
+            => new SignOutResult(authenticationSchemes, properties);
 
         /// <summary>
         /// Updates the specified <paramref name="model"/> instance using values from the controller's current
--- a/src/Microsoft.AspNetCore.Mvc.Core/SignInResult.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/SignInResult.cs
@@ -1,0 +1,95 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Security.Claims;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http.Authentication;
+using Microsoft.AspNetCore.Mvc.Core;
+using Microsoft.AspNetCore.Mvc.Internal;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Mvc
+{
+    /// <summary>
+    /// An <see cref="ActionResult"/> that on execution invokes <see cref="M:AuthenticationManager.SignInAsync"/>.
+    /// </summary>
+    public class SignInResult : ActionResult
+    {
+        /// <summary>
+        /// Initializes a new instance of <see cref="SignInResult"/> with the
+        /// specified authentication scheme.
+        /// </summary>
+        /// <param name="authenticationScheme">The authentication scheme to use when signing in the user.</param>
+        /// <param name="principal">The claims principal containing the user claims.</param>
+        public SignInResult(string authenticationScheme, ClaimsPrincipal principal)
+            : this(authenticationScheme, principal, properties: null)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of <see cref="SignInResult"/> with the
+        /// specified authentication scheme and <paramref name="properties"/>.
+        /// </summary>
+        /// <param name="authenticationScheme">The authentication schemes to use when signing in the user.</param>
+        /// <param name="principal">The claims principal containing the user claims.</param>
+        /// <param name="properties"><see cref="AuthenticationProperties"/> used to perform the sign-in operation.</param>
+        public SignInResult(string authenticationScheme, ClaimsPrincipal principal, AuthenticationProperties properties)
+        {
+            if (authenticationScheme == null)
+            {
+                throw new ArgumentNullException(nameof(authenticationScheme));
+            }
+
+            if (principal == null)
+            {
+                throw new ArgumentNullException(nameof(principal));
+            }
+
+            AuthenticationScheme = authenticationScheme;
+            Principal = principal;
+            Properties = properties;
+        }
+
+        /// <summary>
+        /// Gets or sets the authentication scheme that is used to perform the sign-in operation.
+        /// </summary>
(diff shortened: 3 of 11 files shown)


src/Microsoft.AspNetCore.Mvc.Core/SignInResult.cs
```diff
@@ -0,0 +1,120 @@
...
+            }
+
+            if (principal == null)
+            {
+                throw new ArgumentNullException(nameof(principal));
+            }
+
+            if (authenticationSchemes.Count == 0)
+            {
+                // TODO: move the exception message to the resources file.
```
PinpointTownes: Question: what's the recommended way to add new messages to the resources file? The `resx` KoreBuild target?

pranavkm: Yup - `build resx`

HaoK:
Meh, not sure its worth an extension method to do a foreach SignOut.

src/Microsoft.AspNetCore.Mvc.Core/ControllerBase.cs
```diff
@@ -1038,6 +1038,96 @@ public virtual ForbidResult Forbid(IList<string> authenticationSchemes, Authenti
...
+            => new SignInResult(authenticationScheme, principal);
+
+        /// <summary>
+        /// Creates a <see cref="SignInResult"/> with the specified authentication schemes.
+        /// </summary>
+        /// <param name="authenticationSchemes">The authentication schemes to use for the sign-in operation.</param>
+        /// <param name="principal">The <see cref="ClaimsPrincipal"/> containing the user claims.</param>
+        /// <returns>The created <see cref="SignInResult"/> for the response.</returns>
+        [NonAction]
+        public virtual SignInResult SignIn(IList<string> authenticationSchemes, ClaimsPrincipal principal)
```
HaoK: Not sure this really makes sense, would you really expect someone to sign the same principal into multiple schemes?  I would expect different principals would get signed into each scheme in general

src/Microsoft.AspNetCore.Mvc.Core/ControllerBase.cs
```diff
@@ -1038,6 +1038,96 @@ public virtual ForbidResult Forbid(IList<string> authenticationSchemes, Authenti
...
+        /// <summary>
+        /// Creates a <see cref="SignInResult"/> with the specified specified authentication schemes and
+        /// <paramref name="properties" />.
+        /// </summary>
+        /// <param name="authenticationSchemes">The authentication scheme to use for the sign-in operation.</param>
+        /// <param name="principal">The <see cref="ClaimsPrincipal"/> containing the user claims.</param>
+        /// <param name="properties"><see cref="AuthenticationProperties"/> used to perform the sign-in operation.</param>
+        /// <returns>The created <see cref="SignInResult"/> for the response.</returns>
+        [NonAction]
+        public virtual SignInResult SignIn(
```
HaoK: I'm not sure we should promote the idea of multiple sign ins at once, its fairly uncommon, and if you want to do it, you almost certainly want different principals which would make this method signature pretty gross looking

src/Microsoft.AspNetCore.Mvc.Core/SignInResult.cs
PinpointTownes: Sadly, it doesn't work on my machine:

```
& : The module '.build' could not be loaded. For more information, run
'Import-Module .build'.
At C:\Users\Chalet Kévin\Documents\GitHub\Forks\Mvc\.build\KoreBuild.ps1:64
char:2
+ &"$koreBuildFolder\Sake\0.2.2\tools\Sake.exe" -I $koreBuildFolder\sha ...
+  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (.build\Sake\0.2.2\tools\Sake.ex
   e:String) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : CouldNotAutoLoadModule
```
I guess it's related to https://github.com/aspnet/Security/issues/688#issuecomment-193683592, since I'm also using Windows 7.

/cc @Eilon

src/Microsoft.AspNetCore.Mvc.Core/ControllerBase.cs
```diff
@@ -1038,6 +1038,96 @@ public virtual ForbidResult Forbid(IList<string> authenticationSchemes, Authenti
...
+
+        /// <summary>
+        /// Creates a <see cref="SignOutResult"/> with the specified specified authentication schemes and
+        /// <paramref name="properties" />.
+        /// </summary>
+        /// <param name="authenticationSchemes">The authentication scheme to use for the sign-out operation.</param>
+        /// <param name="properties"><see cref="AuthenticationProperties"/> used to perform the sign-out operation.</param>
+        /// <returns>The created <see cref="SignOutResult"/> for the response.</returns>
+        [NonAction]
+        public virtual SignOutResult SignOut(IList<string> authenticationSchemes, AuthenticationProperties properties)
```
HaoK: I believe this has a similar small flaw in that potentially you might want to pass different AuthenticationProperties to each signout.  Maybe just limit things to a single sign in/sign out which is the most common case?

Any additional sign in/sign outs can be done outside of the result so there's not really any loss of flexibility without these overloads

src/Microsoft.AspNetCore.Mvc.Core/ControllerBase.cs
PinpointTownes: FYI, some ASOS users are doing that to create a cookie and a token with the same principal.

That said, you're probably right, it's surely not a common case.

HaoK: Yeah I think things look much cleaner with a single scheme for sign in / sign out.  You only need multiple schemes for challenge.

src/Microsoft.AspNetCore.Mvc.Core/ControllerBase.cs
PinpointTownes: > Maybe just limit things to a single sign in/sign out which is the most common case?

Specifying multiple schemes is actually more frequent that you might imagine (e.g with OIDC, you'll often want to remove the local cookie and redirect the user to the identity provider).

That said, if you really think the overloads accepting an `AuthenticationProperties` instance with multiple schemes is risky, we can probably remove them.

src/Microsoft.AspNetCore.Mvc.Core/ControllerBase.cs
PinpointTownes: Okay for `SignIn` but for `SignOut`, see https://github.com/aspnet/Mvc/pull/4259#discussion_r55586374.

src/Microsoft.AspNetCore.Mvc.Core/ControllerBase.cs
HaoK: Yep Identity does the same thing, we can leave signout taking multiple schemes just no properties, that seems good enough.

PinpointTownes: > I believe this has a similar small flaw in that potentially you might want to pass different AuthenticationProperties to each signout. Maybe just limit things to a single sign in/sign out which is the most common case?

Hum, FYI, both `ForbidResult` and `ChallengeResult` (and their corresponding helpers) accept multiple schemes with a single properties bag. Not sure why `SignInResult` and `SignOutResult` should get a different treatment.

HaoK: Those are needed because of how Authorize works, which builds a combined policy which can easily end up with multiple schemes needed compose the ClaimsPrincipal.

I guess if those have overloads that take multiple schemes and a single authentication property, we can follow the pattern for SignOut.

I maintain SignIn is different than the rest since that's the only one that's Saving a payload.

src/Microsoft.AspNetCore.Mvc.Core/SignInResult.cs
pranavkm: You might need to delete the .build directory. We made a bunch of changes to how we procure and layout the build scripts.

PinpointTownes: Nope, no luck :cry: 

I ended up manually adding the property/method in the designer file.

danroth27:
Thank you for the PR! However, this needs to wait until post 1.0.0. We aren't taking additional features into the current release.

PinpointTownes:
@danroth27 alright.

That said, it's not really a new feature stricto sensu, as it's mainly a copy/paste of existing code, which is itself a tiny wrapper around the authentication APIs.

davidfowl:
@danroth27 @Eilon this is such a small change. Do we like the API? If so can we merge it? It feels like polish since we have ChallengeResult, SignInResult and SIgnOut result can clean up some of the template code as well.

@Tratcher  @HaoK does it look good?

src/Microsoft.AspNetCore.Mvc.Core/SignInResult.cs
```diff
@@ -0,0 +1,86 @@
...
+        /// <summary>
+        /// Initializes a new instance of <see cref="SignInResult"/> with the
+        /// specified authentication scheme and <paramref name="properties"/>.
+        /// </summary>
+        /// <param name="authenticationScheme">The authentication schemes to use when signing in the user.</param>
+        /// <param name="principal">The claims principal containing the user claims.</param>
+        /// <param name="properties"><see cref="AuthenticationProperties"/> used to perform the sign-in operation.</param>
+        public SignInResult(string authenticationScheme, ClaimsPrincipal principal, AuthenticationProperties properties)
+        {
+            if (authenticationScheme == null)
```
PinpointTownes: Question: should we keep the null guards even if the properties are publicly settable?

Eilon: Either keep the null guards and make the properties read-only, or remove the null guards.

Eilon:
Fine, fine, we can take this :smile:

src/Microsoft.AspNetCore.Mvc.Core/SignInResult.cs
PinpointTownes: Actually, I wonder if we shouldn't simply move these guards to the `ExecuteAsync` method, prior to calling the authentication manager.

Anyway, I need to fix all the aspnet-contrib projects before going back to this PR. Everything stopped compiling/working and I can't find a work around...

src/Microsoft.AspNetCore.Mvc.Core/SignInResult.cs
```diff
@@ -0,0 +1,86 @@
...
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            var loggerFactory = context.HttpContext.RequestServices.GetRequiredService<ILoggerFactory>();
+            var logger = loggerFactory.CreateLogger<SignInResult>();
+
+            var authentication = context.HttpContext.Authentication;
+            await authentication.SignInAsync(AuthenticationScheme, Principal, Properties);
+
+            logger.SignInResultExecuting(AuthenticationScheme, Principal);
```
rynowak: Ideally this line would take place before the call to `SignInAsync`. This way if the 'actual' operation throws, we've already logged what we were trying to do.

src/Microsoft.AspNetCore.Mvc.Core/SignInResult.cs
rynowak: IMO better to validate on the parameters + properties. Generally `ExecuteAsync` won't be called in unit tests, so it be easy to use the result the wrong way.

src/Microsoft.AspNetCore.Mvc.Core/SignOutResult.cs
```diff
@@ -0,0 +1,104 @@
...
+            var logger = loggerFactory.CreateLogger<SignOutResult>();
+
+            var authentication = context.HttpContext.Authentication;
+            
+            for (var i = 0; i < AuthenticationSchemes.Count; i++)
+            {
+                await authentication.SignOutAsync(AuthenticationSchemes[i], Properties);
+            }
+
+            logger.SignOutResultExecuting(AuthenticationSchemes);
```
rynowak: Same comment about the logging ordering vs the actual operation

test/Microsoft.AspNetCore.Mvc.Core.Test/ForbidResultTest.cs
```diff
@@ -19,7 +19,7 @@ namespace Microsoft.AspNetCore.Mvc
     public class ForbidResultTest
```
rynowak: :heart_eyes:

test/Microsoft.AspNetCore.Mvc.Core.Test/SignInResultTest.cs
```diff
@@ -0,0 +1,86 @@
...
+
+            // Act
+            await result.ExecuteResultAsync(actionContext);
+
+            // Assert
+            authenticationManager.Verify();
+        }
+
+        [Fact]
+        public async Task ExecuteResultAsync_InvokesSignInAsyncOnAllConfiguredSchemes()
```
rynowak: Do we really need this test now that we don't have multiple scheme support?

src/Microsoft.AspNetCore.Mvc.Core/ControllerBase.cs
rynowak: > Any additional sign in/sign outs can be done outside of the result so there's not really any loss of flexibility without these overloads

Not adding anything in particular to the discussion, but that's gross :-p

rynowak:
:watch: from me for a few small things

@HaoK - any more feedback on this?

src/Microsoft.AspNetCore.Mvc.Core/SignInResult.cs
PinpointTownes: Agreed, I'll update that and fix the existing results.

test/Microsoft.AspNetCore.Mvc.Core.Test/SignInResultTest.cs
PinpointTownes: Well, it ensures the scheme used to invoke the authentication manager is really the one `SignInResult` takes as a parameter so it's still useful.

That said, I agree the name needs to be updated: `ExecuteResultAsync_InvokesSignInAsyncOnConfiguredScheme`?

src/Microsoft.AspNetCore.Mvc.Core/SignInResult.cs
PinpointTownes: I'm fine with dual properties + parameters validation, but we'll still have to move the "as at least one element" check of `SignOutResult.AuthenticationSchemes` elsewhere, since we could validate that for a constructor parameter, but not for a property.

Another approach might be to remove all the public setters and update `SignOutResult.AuthenticationSchemes` to expose a read-only list. TBH, I have no personal preference, as long as it's consistent everywhere.

rynowak: We generally want all the action results to have get/set properties. It's really handy to be able to modify things in filters. Speaking of, we should probably make a defensive copy of whatever list the user passes in.

PinpointTownes:
> (FWIW, that would be great if the authentication helpers had overloads accepting a params string[] argument for authenticationSchemes).

Any opinion?

test/Microsoft.AspNetCore.Mvc.Core.Test/SignInResultTest.cs
rynowak: This is really just the same code as the first test with different values passed to `new SignInResult` and into the mock verification. There's nothing different in the product code between these two cases right?

I wouldn't bother with both of these, but follow your :heart:

HaoK:
Looks fine to me :shipit:

rynowak:
@PinpointTownes - any updates? 

> (FWIW, that would be great if the authentication helpers had overloads accepting a params string[] argument for authenticationSchemes).

I wouldn't block this PR on that item, if you want to, go ahead and file that on the appropriate repo?

PinpointTownes:
Sadly no. Fixing all the aspnet-contrib/OpenIddict stuff to make DNX and CLI happy with the recent changes in corefx and NuGet took me much more time than I had expected (and that's an euphemism). 

The fact MVC itself no longer compiles on with DNX/VS is also extremely annoying, specially since CLI no longer works on Windows 7: https://github.com/dotnet/cli/issues/1966

TBH, I don't feel super comfortable not being able to run the unit tests before pushing new commits. But since you haven't updated MVC to include the frameworkAssemblies workaround that would make DNX happy and since CI builds always fail before running the tests, that's pretty much my only option. 

I'll try some yolodev tonight and fix all the reviewed points. 

---

By authentication helpers, I meant the SignOut/Challenge/Forbid methods exposed by ControllerBase, that all take IList<string>. We could update them to take array params without having to update AuthenticationManager.

rynowak:
> By authentication helpers, I meant the SignOut/Challenge/Forbid methods exposed by ControllerBase, that all take IList<string>. We could update them to take array params without having to update AuthenticationManager.

Yeah I think that would be a fine change

Sorry for all the difficulties :cry: if you don't manage to get to this I'm happy to fix up the remaining stuff and make sure sure it's included for RC2.

src/Microsoft.AspNetCore.Mvc.Core/ControllerBase.cs
PinpointTownes: Addressed in a previous commit.

src/Microsoft.AspNetCore.Mvc.Core/SignInResult.cs
PinpointTownes: I've added a tiny check in both `SignInResult.ExecuteAsync` and `SignOutResult.ExecuteAsync` to ensure the `AuthenticationScheme(s)` property is not left is an incoherent state (null or empty, basically).

src/Microsoft.AspNetCore.Mvc.Core/SignInResult.cs
PinpointTownes: Done.

src/Microsoft.AspNetCore.Mvc.Core/ChallengeResult.cs
```diff
@@ -101,8 +101,10 @@ public ChallengeResult(IList<string> authenticationSchemes, AuthenticationProper
             var loggerFactory = context.HttpContext.RequestServices.GetRequiredService<ILoggerFactory>();
             var logger = loggerFactory.CreateLogger<ChallengeResult>();
 
+            logger.ChallengeResultExecuting(AuthenticationSchemes);
+
             var authentication = context.HttpContext.Authentication;
-            if (AuthenticationSchemes.Count > 0)
+            if (AuthenticationSchemes != null && AuthenticationSchemes.Count > 0)
```
PinpointTownes: I've added a null check here, since `AuthenticationSchemes` can be null (there's no null guard in the constructor).

src/Microsoft.AspNetCore.Mvc.Core/SignInResult.cs
PinpointTownes: @rynowak do you strongly feel we should do some `ToArray`/`ToList` here? Authentication-related actions are probably not critical paths, so I guess the extra allocations are fine.

src/Microsoft.AspNetCore.Mvc.Core/SignInResult.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        /// <inheritdoc />
+        public override async Task ExecuteResultAsync(ActionContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            if (AuthenticationScheme == null)
```
PinpointTownes: Note: I'm not a huge fan of empty schemes (and empty strings in general), but @HaoK strongly feels `AuthenticationManager` shouldn't reject them.

PinpointTownes:
:date: PR updated :smile:

PinpointTownes:
> Yeah I think that would be a fine change

Alright, I've updated the `ControllerBase` methods but kept the results themselves as-is (they still take `IList<string>` parameters)

> Sorry for all the difficulties :cry: if you don't manage to get to this I'm happy to fix up the remaining stuff and make sure sure it's included for RC2.

No worries. Thanks for the offer: actually, it would be great if you could run the tests for me :smile:

src/Microsoft.AspNetCore.Mvc.Core/SignInResult.cs
```diff
@@ -0,0 +1,120 @@
...
+
+            if (principal == null)
+            {
+                throw new ArgumentNullException(nameof(principal));
+            }
+
+            if (authenticationSchemes.Count == 0)
+            {
+                // TODO: move the exception message to the resources file.
+                throw new ArgumentException("At least one scheme must be specified", nameof(authenticationSchemes));
```
rynowak: I'll take a look at fixing this

PinpointTownes: Hum, are you commenting on an old commit? I'm pretty sure I've fixed that (by manually adding the resource entry :trollface:)

src/Microsoft.AspNetCore.Mvc.Core/ControllerBase.cs
```diff
@@ -1029,15 +1029,80 @@ public virtual ForbidResult Forbid(string authenticationScheme, AuthenticationPr
...
+
+        /// <summary>
+        /// Creates a <see cref="SignOutResult"/> with the specified specified authentication scheme and
+        /// <paramref name="properties" />.
+        /// </summary>
+        /// <param name="authenticationScheme">The authentication scheme to use for the sign-out operation.</param>
+        /// <param name="properties"><see cref="AuthenticationProperties"/> used to perform the sign-out operation.</param>
+        /// <returns>The created <see cref="SignOutResult"/> for the response.</returns>
+        [NonAction]
+        public virtual SignOutResult SignOut(string authenticationScheme, AuthenticationProperties properties)
```
rynowak: These should use consistent ordering. IE scheme should be be last here and on  `SignIn`

PinpointTownes: The scheme parameter is always the first one, except when it's a params array. I'm fine changing the order of the parameters if you really want me to, but `properties` is an "optional" parameter that should ideally be the last.

src/Microsoft.AspNetCore.Mvc.Core/ControllerBase.cs
```diff
@@ -1029,15 +1029,80 @@ public virtual ForbidResult Forbid(string authenticationScheme, AuthenticationPr
...
+            AuthenticationProperties properties)
+            => new SignInResult(authenticationScheme, principal, properties);
+
+        /// <summary>
+        /// Creates a <see cref="SignOutResult"/> with the specified authentication scheme.
+        /// </summary>
+        /// <param name="authenticationScheme">The authentication scheme to use for the sign-out operation.</param>
+        /// <returns>The created <see cref="SignOutResult"/> for the response.</returns>
+        [NonAction]
+        public virtual SignOutResult SignOut(string authenticationScheme)
```
PinpointTownes: This overload is useless, I'll remove it.

src/Microsoft.AspNetCore.Mvc.Core/ControllerBase.cs
rynowak: Ideally that's a good principle. In practice putting `authenticationScheme` last is the right thing to do when you have `params` in play.

Otherwise `SignOut("foo", properties)` and `SignOut(properties, "foo")`

PinpointTownes: Okay, I'll remove the overloads taking a single scheme string, since there's no point having `SignOut(AuthenticationProperties properties, string authenticationScheme)` and `SignOut(AuthenticationProperties properties, params string[] authenticationSchemes)`.

rynowak: That's fine to remove but I'd still like to see `SignIn` and `SignOut` consistent

PinpointTownes: I've reordered the `SignIn` parameters, to make the scheme the last parameter.

For perfect consistency, `SignIn` would have to accept multiple schemes, but @HaoK doesn't want that.

src/Microsoft.AspNetCore.Mvc.Core/SignInResult.cs
rynowak: Yeah that was an oopsie on my part

rynowak:
:shipit: - looks good to me. I'm going to squash and run tests for you

PinpointTownes:
I can squash it if you want, it will be easier for you if you want to cherry-pick it.

rynowak:
go for it then :+1:

PinpointTownes:
Done :clap:

rynowak:
Will merge as soon as I'm able to.... (restore issues).

rynowak:
Still waiting on package/build issues, will update once I've been able to merge.

PinpointTownes:
Thanks!

Yeah, it's super messy ATM, the reversioning of the new `System.Runtime` package is literally killing me (https://github.com/openiddict/openiddict-core/issues/76#issuecomment-201058572)

rynowak:
Pushed as f9d24a8521fbdc82234544068d4a832ca8f3a53a Thanks :+1:

PinpointTownes:
Thanks @rynowak :clap: :clap:

PR #4261, opened 2016-03-09
RazorCompilationService should throw a meaningful error when a user p…

### LINKED ISSUE #4202: RazorCompilationService should throw a meaningful error when a user produces an application without `preserveCompilationContext`
When the compilation service encounters an application that has been compiled without the `preserveCompilationContext` flag - it attempts to compile the application without any reference assemblies. This invariably fails with messages such as `The type or namespace System could not be found.` See - https://cloud.githubusercontent.com/assets/4823412/13446609/1bd6c9dc-dfd2-11e5-94cc-fb7547f4c585.png. We should instead throw an exception that instructs users to add this flag to their project.json.

cc @davidfowl

pranavkm:
…roduces an application without `preserveCompilationContext`

Fixes #4202

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultRoslynCompilationService.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultRoslynCompilationService.cs
@@ -85,7 +85,6 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Internal
 #if NETSTANDARD1_5
             _razorLoadContext = new RazorLoadContext();
 #endif
-
         }
 
         /// <inheritdoc />
@@ -144,6 +143,18 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Internal
 
                     if (!result.Success)
                     {
+                        if (!compilation.References.Any() && !_applicationReferences.Value.Any())
+                        {
+                            // DependencyModel had no references specified and the user did not use the
+                            // CompilationCallback to add extra references. It is likely that the user did not specify
+                            // preserveCompilationContext in the app's project.json.
+                            throw new InvalidOperationException(
+                                Resources.FormatCompilation_DependencyContextIsNotSpecified(
+                                    fileInfo.RelativePath,
+                                    "project.json",
+                                    "preserveCompilationContext"));
+                        }
+
                         return GetCompilationFailedResult(
                             fileInfo.RelativePath,
                             compilationContent,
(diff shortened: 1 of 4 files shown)


pranavkm:
cc @pakrym \ @NTaylorMullen

rynowak:
Why not do this in `AddMvc` if your assembly provider returns nothing? This is a problem for APIs also.

pakrym:
LGTM

NTaylorMullen:
@rynowak how is it a problem for APIs? Just curious.

pranavkm:
@rynowak, at that point we use the current assembly as the single source - https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/Infrastructure/DefaultAssemblyProvider.cs#L71-L74. We had a discussion about it here - https://github.com/aspnet/Mvc/issues/4136#issuecomment-188615339 - and the plan was to not throw in that scenario.

rynowak:
Is this a new thing? I was doing this today and it was broken. @glennc ran into the same issue (Give it a try)

pranavkm:
The code's been around for a while. I'll investigate your issue.

pranavkm:
@rynowak your scenario is something we missed - if you compile without `preserveCompilationContext`, `DependencyContext.RuntimeLibraries` is populated, but `RuntimeLibrary.Dependencies` is always empty. Consequently Controller discovery fails. I'll run it by @davidfowl to see how we should account for this.

pakrym:
Wait for deps.json to get in(tomorrow) . It has dependencies for runtime libraries.

pranavkm:
Ok. I'll leave the assembly provider unchanged. But it does look like for an app running via `dotnet run` you're guaranteed to have a non-null `DependencyContext`. Checking for null `DependencyContext` almost never works. I'll change the conditional a bit.

rynowak:
Ok, I'm opening a bug then that track that. Because it's a thing we have.

pranavkm:
Updated the condition. I'll wait for new CLI bits to address https://github.com/aspnet/Mvc/issues/4266

src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultRoslynCompilationService.cs
```diff
@@ -144,6 +143,18 @@ public CompilationResult Compile(RelativeFileInfo fileInfo, string compilationCo
 
                     if (!result.Success)
                     {
+                        if (!compilation.References.Any() && !_applicationReferences.Value.Any())
```
pranavkm: @davidfowl \ @pakrym  Yeah, this is a guess at best. We could add an API to `DependencyContext` that says it doesn't have compilation context which might be nicer. Thoughts?

pakrym: context.CompilationLibraries.Any()?

pranavkm: Which is sort of what `~_applicationReferences.Value.Any()` represents. But it's a bit lossy compared to directly querying if the app is missing that flag in the project.json.

pranavkm:
@davidfowl is the heuristic we use here acceptable?

PR #4268, opened 2016-03-10
Don't reference facades in NuSpec

bricelam:
These can be removed entirely after dotnet/cli#164

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/project.json
+++ b/src/Microsoft.AspNetCore.Mvc.Core/project.json
@@ -51,12 +51,7 @@
     "System.Diagnostics.DiagnosticSource": "4.0.0-*"
   },
   "frameworks": {
-    "net451": {
-      "frameworkAssemblies": {
-        "System.Reflection": "",
-        "System.Runtime": ""
-      }
-    },
+    "net451": { },
     "netstandard1.3": {
       "imports": [
         "dotnet5.4",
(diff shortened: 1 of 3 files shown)


dougbu:
:shipit: as long as removed dependencies are still pulled in (transitively)

src/Microsoft.AspNetCore.Mvc.Core/project.json
```diff
@@ -51,12 +51,7 @@
     "System.Diagnostics.DiagnosticSource": "4.0.0-*"
   },
   "frameworks": {
-    "net451": {
-      "frameworkAssemblies": {
-        "System.Reflection": "",
-        "System.Runtime": ""
-      }
-    },
+    "net451": { },
```
kichalla: Just curious, shouldn't we have use the TFM `netstandard1.2` (instead which targets 451 & 452) here as its a library project?. My understanding is that we are supposed to use monikers like `net451` and `netstandardapp1.5` only in projects are actually run (ex: samples, tests).

pranavkm: We tried that change - https://github.com/aspnet/DependencyInjection/commit/99f4f17a88277c4edfabd988ae994db79c1b17a7 - but ended up reverting it because of odd CLI issues + issues with restore. I'll work on bringing it back soon.

Issue #4083, opened 2016-02-12
`<link>` and `<script>` tag helpers do not encode correctly


dougbu:
Some literal text, most obviously in `asp-fallback-href` and `asp-fallback-src` values, are not correctly encoded when written into the browser's document.

E.g. (yeah, an intentional obstacle course) in the `.cshtml` file
```html
    <script src="~/blank.js?a=b&c=d"
            asp-fallback-src='~/styles/site&#x1f4a9;.js?a=<"the" &#x1f4a9; title>'
            asp-fallback-test="false"
            data-foo="foo-data2"
            title='<"the" &#x1f4a9; title>'>
        // TagHelper script with comment in body, and extra properties.
    </script>
```

generates
```html
    <script src="/blank.js?a=b&amp;c=d" data-foo="foo-data2" title="<&quot;the&quot; &#x1f4a9; title>">
        // TagHelper script with comment in body, and extra properties.
    </script>
<script>(false||document.write("<script src=\"\/styles\/site\u0026#x1f4a9;.js?a=\u003C\u0022the\u0022 \u0026#x1f4a9; title\u003E\" data-foo=\"foo-data2\" title=\"\u003C\u0022the\u0022 \u0026#x1f4a9; title\u003E\"><\/script>"));</script>
```

The `src` attribute w/in the `document.write()` is correctly JavaScript encoded but the written HTML is
```html
<script src="/styles/site&#x1f4a9;.js?a=<"the" &#x1f4a9; title>" data-foo="foo-data2" title="<"the" &#x1f4a9; title>"></script>
```

Note the final `src` and `title` attributes contain double quotes and are delimited with double quotes.

===

Test encoders show the issue directly
- Main element HTML encodes the `src` attribute value (`src="HtmlEncode[[/blank.js]]"`). This is actually overzealous: The original text was `src="~/blank.js"` and so `src="HtmlEncode[[/]]blank.js"` was expected.
- Fallback element JavaScript encodes the `src` attribute value (`src=\"JavaScriptEncode[[/styles/site.js]]\"`) and pretty much every other attribute name and value. But the JavaScript string is written out as HTML and none of it is HTML encoded.

Problems are somewhat less extensive w/ the `<link>` tag helper because it does not include unknown attributes in the fallback elements. (Might be a separate bug there, not sure.)

dougbu:
ffe2d26

Issue #4084, opened 2016-02-12
Link tag helper should include original element's attributes in fallbacks


dougbu:
E.g. `.cshtml` contains:
```html
    <link rel="stylesheet" data-extra="test"
          asp-fallback-href="~/styles/site.css"
          asp-fallback-test-class="hidden"
          asp-fallback-test-property="visibility"
          asp-fallback-test-value="hidden">
```

Generated HTML with a few extra newlines:
```html
    <link rel="stylesheet" data-extra="test">
<meta name="x-stylesheet-fallback-test" content="" class="hidden" />
<script>
!function(a,b,c){
var d,e=document,
  f=e.getElementsByTagName("SCRIPT"),
  g=f[f.length-1].previousElementSibling,
  h=e.defaultView&&e.defaultView.getComputedStyle?e.defaultView.getComputedStyle(g):g.currentStyle;
if(h&&h[a]!==b)
  for(d=0;d<c.length;d++)
    e.write('<link rel="stylesheet" href="'+c[d]+'"/>')
}("visibility","hidden",["\/styles\/site.css"]);
</script>
```

Note the `data-extra` is not included in the `<link>` element in the JavaScript, ready to be written into the document. Should it?

DamianEdwards:
Given cases like http://codepen.io/tigt/post/async-css-without-javascript where you'd want extra attributes to get different behavior, then yes, it should probably preserve the extra attributes. I actually thought I remembered writing the code to do that... Does it work in the script TH?

dougbu:
> Does it work in the script TH?

For some definition of "work" i.e. modulo #4083, yes.

NTaylorMullen:
d60ed06c1998ced0af09ce1b86aa9f4b774495be

PR #4272, opened 2016-03-11
Correct HTML and JavaScript encoding of `<link>` and `<script>` attribute values

dougbu:
- #4083
- `<link>` tag helper did not HTML encode `href` values in fallback elements
- `<script>` tag helper did not correctly encode any attribute value in fallback elements
 - e.g. double quotes in literal strings would slip through
- only needed to change one existing unit test (!!); so added a bunch

nit: use `Process()`, not `ProcessAsync()` in `<script>` tag helper tests


src/Microsoft.AspNetCore.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -324,6 +342,37 @@ private void BuildFallbackBlock(TagHelperAttributeList attributes, DefaultTagHel
...
+                stringValue = htmlEncodedString.Value;
+                if (stringValue.Contains("\""))
+                {
+                    stringValue = stringValue.Replace("\"", "&quot;");
+                }
+            }
+            else
+            {
+                var writer = StringWriter;
+                RazorPage.WriteTo(writer, HtmlEncoder, value);
```
dougbu: This `static` method isn't fitting so well in `RazorPage` but I haven't got a better place to throw it. Will leave as-is unless someone has a good suggestion.

test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/LinkTagHelperTest.cs
```diff
@@ -634,24 +638,41 @@ public void RendersLinkTagsForGlobbedHrefResults()
             // Assert
             Assert.Equal("link", output.TagName);
             Assert.Equal("/css/site.css", output.Attributes["href"].Value);
-            Assert.Equal("<link rel=\"stylesheet\" href=\"HtmlEncode[[/base.css]]\" />", output.PostElement.GetContent());
+            var content = HtmlContentUtilities.HtmlContentToString(output, new HtmlTestEncoder());
+            Assert.Equal(expectedContent, content);
         }
 
         [Fact]
-        public void RendersLinkTagsForGlobbedHrefResults_UsingProvidedEncoder()
```
dougbu: This test was a dupe of `RendersLinkTagsForGlobbedHrefResults()`. Fairly different now. Same deal for `RendersScriptTagsForGlobbedSrcResults_UsesProvidedEncoder()` and `RendersScriptTagsForGlobbedSrcResults()`.

test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/LinkTagHelperTest.cs
```diff
@@ -761,7 +782,143 @@ public void RendersLinkTags_AddsFileVersion_WithRequestPathBase()
...
+        {
+            // Arrange
+            var expectedPostElement = Environment.NewLine +
+                "<meta name=\"x-stylesheet-fallback-test\" content=\"\" class=\"hidden\" />" +
+                "<script>!function(a,b,c){var d,e=document,f=e.getElementsByTagName(\"SCRIPT\")," +
+                "g=f[f.length-1].previousElementSibling," +
+                "h=e.defaultView&&e.defaultView.getComputedStyle?e.defaultView.getComputedStyle(g):g.currentStyle;" +
+                "if(h&&h[a]!==b)for(d=0;d<c.length;d++)e.write('<link rel=\"stylesheet\" href=\"'+c[d]+'\"/>')}(" +
+                "\"JavaScriptEncode[[visibility]]\",\"JavaScriptEncode[[hidden]]\"," +
+                "[\"JavaScriptEncode[[HtmlEncode[[/fallback.css?v=f4OxZX_x_FO5LcGBSKHWXfwtSx-j1ncoSt3SABJtkGk]]]]\"]);" +
```
dougbu: #4084 covers lack of unbound attributes in the fallback `<link>` element(s).

test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/LinkTagHelperTest.cs
```diff
@@ -761,7 +782,143 @@ public void RendersLinkTags_AddsFileVersion_WithRequestPathBase()
         }
 
         [Fact]
-        public void RendersLinkTags_GlobbedHref_AddsFileVersion()
+        public void RenderLinkTags_FallbackHref_WithFileVersion()
```
dougbu: Added because we had this test for the `<script>` tag helper but not here.

ryanbrandenburg:
:shipit:, seems straighforward.

src/Microsoft.AspNetCore.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -324,6 +342,37 @@ private void BuildFallbackBlock(TagHelperAttributeList attributes, DefaultTagHel
...
+        private string GetAttributeValue(object value)
+        {
+            string stringValue;
+            var htmlEncodedString = value as HtmlEncodedString;
+            if (htmlEncodedString != null)
+            {
+                // Value likely came from an HTML context in the .cshtml file but may still contain double quotes
+                // since attribute could have been enclosed in single quotes.
+                stringValue = htmlEncodedString.Value;
+                if (stringValue.Contains("\""))
```
pranavkm: This check shouldn't be required. Calling `stringValue.Replace` would no-op if it doesn't have a quote.

src/Microsoft.AspNetCore.Mvc.TagHelpers/ScriptTagHelper.cs
```diff
@@ -324,6 +342,37 @@ private void BuildFallbackBlock(TagHelperAttributeList attributes, DefaultTagHel
...
+                {
+                    stringValue = stringValue.Replace("\"", "&quot;");
+                }
+            }
+            else
+            {
+                var writer = StringWriter;
```
pranavkm: Why not use the property directly?

pranavkm:
:shipit:

src/Microsoft.AspNetCore.Mvc.TagHelpers/ScriptTagHelper.cs
dougbu: :ok_hand: double-checked the [native source](https://github.com/dotnet/coreclr/blob/f007485a78024b1953a7fe1ffa845be28e93ab6c/src/classlibnative/bcltype/stringnative.cpp#L836) and confirmed that.

src/Microsoft.AspNetCore.Mvc.TagHelpers/ScriptTagHelper.cs
dougbu: :see_no_evil:

dougbu:
ffe2d26

PR #4273, opened 2016-03-11
Target dnx451

bricelam:
Reverts 7b70da14bccb843c2f1daca5e7e2faa30e4b1a20 on DNX


dougbu:
:shipit:

bricelam:
Recalling. After chatting with @Eilon, we decided that adding the facades to user's `project.json` is an acceptable workaround.

PR #4274, opened 2016-03-11
Workaround for Json.Net bug

ajaybhargavb:
@rynowak @Eilon

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Formatters.Json/Internal/JsonArrayPool.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Formatters.Json/Internal/JsonArrayPool.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Buffers;
+using System.Collections.Generic;
 using Newtonsoft.Json;
 
 namespace Microsoft.AspNetCore.Mvc.Formatters.Json.Internal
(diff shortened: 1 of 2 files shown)


Eilon:
Reviewed in person so this looks :shipit: to me, but need @rynowak to confirm that this was the intention.

src/Microsoft.AspNetCore.Mvc.Formatters.Json/Internal/JsonArrayPool.cs
```diff
@@ -33,7 +39,11 @@ public void Return(T[] array)
                 throw new ArgumentNullException(nameof(array));
             }
 
-            _inner.Return(array);
+            if (_arrayTracker.Contains(array))
+            {
+                _arrayTracker.Remove(array);
```
dougbu: Can combine `Contains()` and `Remove()` calls: `if (_arrayTracker.Remove(array))`

ajaybhargavb: :+1:

rynowak: Look at me! I used the new reaction thing to agree with @dougbu

rynowak:
:shipit: - as workarounds go, this one is pretty great I gotta say :+1:

ajaybhargavb:
Closing this as it is no longer needed.

rynowak:
We should be able to use a new 8.0.3 package and so this won't be needed.

PR #4281, opened 2016-03-14
Ignore body/section now returns HtmlString.Empty

ctyar:
This is in fact a followup for the discussion here
https://github.com/aspnet/Mvc/pull/4024
I was late to fix some of @pranavkm points, and the PR merged before I address them

The missing parts

> Could you verify that content written by the page - inside the delegate for CreatePage - gets written out (and not ignored?)

based on this 
https://github.com/aspnet/Mvc/issues/3293#issuecomment-146991821
`IgnoreSection` only suppresses errors right now, and wont stop sections from rendering in other words:
If u call `RenderSection("section1")` and also `IgnoreSection("section1")` the section would be rendered
should I change this behavior?

> One more test that verifies that IgnoreSection does not affect EnsureRenderedBodyOrSections for sections that aren't ignored?

Done

Also I changed the return types so it would be easier to use

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor/RazorPage.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor/RazorPage.cs
@@ -672,9 +672,12 @@ namespace Microsoft.AspNetCore.Mvc.Razor
         /// <summary>
         /// In a Razor layout page, ignores rendering the portion of a content page that is not within a named section.
         /// </summary>
-        public void IgnoreBody()
+        /// <returns><see cref="HtmlString.Empty"/>.</returns>
+        public HtmlString IgnoreBody()
         {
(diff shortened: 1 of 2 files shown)


pranavkm:
@ctyar I added the additional test as part of the commit (https://github.com/aspnet/Mvc/blob/dev/test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorPageTest.cs#L698). We decided we wanted to avoid the pattern for `IgnoreSection`, `IgnoreBody` where it returns a `HtmlString`. They have to be invoked as part of a code section:

e.g. `@{ IgnoreSection("section-to-ignore"); }`

ctyar:
@pranavkm so I think this is not necessary anymore

pranavkm:
Yup.

PR #4285, opened 2016-03-14
Updating Json.Net version

ajaybhargavb:
@rynowak / @Eilon


rynowak:
You'll want to do a sweep for other places. JsonPatch repo for one.

JamesNK:
Up to you guys whether you think it is worth keeping everywhere in sync. The changes between 8.0.2 and 8.0.3 are minor and there will be a new version post RC2 with netstandard to update to anyway.

PR #4288, opened 2016-03-14
[Perf] Optimize CoerceResultToTaskAsync to remove MethodInfo.Invoke

### LINKED ISSUE #4267: Optimize CoerceResultToTaskAsync
image

This is about 1.2% of all execution time for a small app with a database, model binding and view. 

We should fold this into the existing cached delegate for action invocation if possible.

mnltejaswini:
Fixes #4267 
The controller action return values are the coerced to Task<object> type. This involved using MethodInfo.Invoke when the return type is a generic Task type. Folding this coercing process in the cached action delegate to reduce the perf penalty from reflection

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
@@ -3,14 +3,34 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc.Core;
+using Microsoft.Extensions.Internal;
 
 namespace Microsoft.AspNetCore.Mvc.Internal
 {
     public class ObjectMethodExecutor
     {
+        private ActionExecutorAsync _executorAsync;
         private ActionExecutor _executor;
+
+        private static readonly MethodInfo _convertOfTMethod =
+            typeof(ObjectMethodExecutor).GetRuntimeMethods().Single(methodInfo => methodInfo.Name == nameof(ObjectMethodExecutor.Convert));
+
+        private static readonly Expression<Func<object, Task<object>>> _createTaskFromResultExpression =
+            ((result) => Task.FromResult(result));
+
+        private static readonly MethodInfo _createTaskFromResultMethod = 
+            ((MethodCallExpression)_createTaskFromResultExpression.Body).Method;
+
+        private static readonly Expression<Func<object, string, Type, Task<object>>> _coerceTaskExpression =
+            ((result, methodName, declaringType) => ObjectMethodExecutor.CoerceTaskType(result, methodName, declaringType));
+
+        private static readonly MethodInfo _coerceMethod = ((MethodCallExpression)_coerceTaskExpression.Body).Method;
+
 
         private ObjectMethodExecutor(MethodInfo methodInfo)
         {            
@@ -20,6 +40,8 @@ namespace Microsoft.AspNetCore.Mvc.Internal
             }
             MethodInfo = methodInfo;
         }
+
+        private delegate Task<object> ActionExecutorAsync(object target, object[] parameters);
 
         private delegate object ActionExecutor(object target, object[] parameters);
 
@@ -31,7 +53,13 @@ namespace Microsoft.AspNetCore.Mvc.Internal
         {
             var executor = new ObjectMethodExecutor(methodInfo);
             executor._executor = GetExecutor(methodInfo, targetTypeInfo);
+            executor._executorAsync = GetExecutorAsync(methodInfo, targetTypeInfo);
             return executor;
+        }
+
+        public Task<object> ExecuteAsync(object target, object[] parameters)
+        {
+            return _executorAsync(target, parameters);
         }
 
         public object Execute(object target, object[] parameters)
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
```diff
@@ -3,15 +3,23 @@
...
 
 namespace Microsoft.AspNetCore.Mvc.Internal
 {
     public class ObjectMethodExecutor
     {
+        private ActionExecutorAsync _executorAsync;
         private ActionExecutor _executor;
 
+        private static readonly MethodInfo _convertOfTMethod =
+            typeof(ObjectMethodExecutor).GetRuntimeMethods().Single(methodInfo => methodInfo.Name == "Convert");
```
rynowak: use `nameof` here, it will prevent this from breaking if renamed

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
```diff
@@ -87,5 +103,151 @@ private static ActionExecutor WrapVoidAction(VoidActionExecutor executor)
...
+                    return convertedResult;
+                }
+
+                // This will be the case for types which have derived from Task and Task<T>
+                throw new InvalidOperationException(Resources.FormatActionExecutor_UnexpectedTaskInstance(
+                    methodInfo.Name,
+                    methodInfo.DeclaringType));
+            }
+
+            // Action methods which use dynamic keyword but return a Task or Task<T> will also be coerced to Task<object>
```
rynowak: I'd add more to this comment (or delete it). This gives the impression that `dynamic` is somehow special, it's not. It behaves just like `object`.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
```diff
@@ -87,5 +103,151 @@ private static ActionExecutor WrapVoidAction(VoidActionExecutor executor)
...
+            Expression<Func<object, Task<object>>> createTaskFromResultExpression = ((result) => Task.FromResult(result));
+            var createTaskFromResultMethodInfo = ((MethodCallExpression)createTaskFromResultExpression.Body).Method;
```
rynowak: These two lines can be hoisted into a constant if it helps.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
```diff
@@ -87,5 +103,151 @@ private static ActionExecutor WrapVoidAction(VoidActionExecutor executor)
...
+        {
+            var castMethodCall = Expression.Convert(methodCall, typeof(object));
+            Type returnType = methodCall.Type;
+
+            if (typeof(Task).IsAssignableFrom(returnType))
+            {
+                if (returnType == typeof(Task))
+                {
+                    Expression<Func<object, string, Type, Task<object>>> coerceTaskExpression = ((result, methodName, declaringType) => ObjectMethodExecutor.CoerceTaskType(result, methodName, declaringType));
+                    var coerceMethodInfo = ((MethodCallExpression)coerceTaskExpression.Body).Method;
```
rynowak: These two lines can be hoisted into a constant.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
```diff
@@ -87,5 +103,151 @@ private static ActionExecutor WrapVoidAction(VoidActionExecutor executor)
...
+        }
+
+        // We need to CoerceResult as the object value returned from methodInfo.Invoke has to be cast to a Task<T>.
+        // This is necessary to enable calling await on the returned task.
+        // i.e we need to write the following var result = await (Task<ActualType>)mInfo.Invoke.
+        // Returning Task<object> enables us to await on the result.
+        private static Expression GetCoerceMethodCallExpression(MethodCallExpression methodCall, MethodInfo methodInfo)
+        {
+            var castMethodCall = Expression.Convert(methodCall, typeof(object));
+            Type returnType = methodCall.Type;
```
rynowak: var

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
```diff
@@ -87,5 +103,151 @@ private static ActionExecutor WrapVoidAction(VoidActionExecutor executor)
...
+                            declaringType,
+                            actualTypeReturned.FullName));
+                }
+            }
+        }
+
+        private static Task<object> CoerceTaskType(object result, string methodName, Type declaringType)
+        {
+            var resultAsTask = result as Task;
+            ThrowIfWrappedTaskInstance(resultAsTask.GetType(), methodName, declaringType);
```
rynowak: If you're going to keep this around, let's remeasure. I'm concerned we might want to get rid of this too.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
```diff
@@ -87,5 +103,151 @@ private static ActionExecutor WrapVoidAction(VoidActionExecutor executor)
...
+        /// Cast Task of T to Task of object
+        /// </summary>
+        private static async Task<object> CastToObject<T>(Task<T> task)
+        {
+            return (object)await task;
+        }
+
+        private static Type GetTaskInnerTypeOrNull(Type type)
+        {
+            var genericType = ClosedGenericMatcher.ExtractGenericInterface(type, typeof(Task<>));
```
rynowak: This is a fairly expensive call. just a heads up.

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ControllerActionExecutorTests.cs
```diff
@@ -323,18 +323,15 @@ public class ControllerActionExecutorTests
...
 
-            // Act & Assert
-            var ex = await Assert.ThrowsAsync<InvalidOperationException>(
-               () => ExecuteAction(
-                   dynamicTaskMethod,
-                   _controller,
-                   actionParameters));
-            Assert.Equal(expectedException, ex.Message);
+            // Act
+            var result = await (Task<int>)(await ExecuteAction(
```
rynowak: The test name needs to change. Leave a comment here explaining that this isn't really a *supported* scenario, it's just something we don't check for specifically.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
mnltejaswini: typeof() expects a type where as nameof returns string. So this may not compile if I am not missing anything?

mnltejaswini:
Benchmark Test app : https://github.com/aspnet/Performance/tree/dev/testapp/BasicViews

loadtest -k -n 100000 -c 1000  http://localhost:5000


Before Optimization
--------------------------
image

After Optimization
-----------------------
image

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
rynowak: You can use `nameof(ObjectMethodExecutor.Convert)` right?

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
```diff
@@ -87,5 +115,146 @@ private static ActionExecutor WrapVoidAction(VoidActionExecutor executor)
...
+                            methodName,
+                            declaringType,
+                            actualTypeReturned.FullName));
+                }
+            }
+        }
+
+        private static Task<object> CoerceTaskType(object result, string methodName, Type declaringType)
+        {
+            var resultAsTask = result as Task;
```
rynowak: this should be `(Task)result`. You don't ever expect this to fail.

rynowak:
:shipit: after fixing minor issues

rynowak:
:shipit:

mnltejaswini:
will add one

PR #4289, opened 2016-03-15
Remove ServiceProvider

### LINKED ISSUE #4097: Don't expose ServiceProvider on ControllerBase
It's already available on HttpContext and we really don't want people to use it directly. Let's not expose it from multiple vectors.

ryanbrandenburg:
Remove the ServiceProvider from ControllerBase. People should use `HttpContext?.RequestServices` instead.

Fixes #4097


pranavkm:
:shipit: 

This might need an annoucement as a follow up.

davidfowl:
:shipit: !!!!!

Issue #3921, opened 2016-01-13
Remove uncached Reflection in `ExpressionHelper.GetExpressionText()`


dougbu:
The `IsSingleArgumentIndexer()` helper method performs uncached Reflection. `ModelMetadata` could be available in this code. Or cached information about the Linq expression could be reused.

danroth27:
Does this show up when profiling?

rynowak:
Yes, that's how we found it :+1: I don't have the data in front of me though, I was looking over @dougbu 's shoulder

dougbu:
`GetCustomAttribute<T>()` is responsible for a fair number of different allocations. Hard to see the total using these tools. But those calls cost more than 1% of the top two allocated types.

See `object[]`:
![object](https://cloud.githubusercontent.com/assets/6431421/12430575/2b6d130e-bea5-11e5-8058-23433c583e25.png)

And `string`:
![string](https://cloud.githubusercontent.com/assets/6431421/12430587/3aac54ce-bea5-11e5-920b-2c8c6a2a830a.png)

dougbu:
Hmm, since `string` allocations are part of this, we could consider this issue as a detail of #3918.

Eilon:
@dougbu @rynowak moving to backlog as we're a bit overbooked. This doesn't seem to involve API changes so we can pick this up another time.

mnltejaswini:
We optimized the current ExpressionHelper to the commonly occurring scenarios (which is also the targeted scenarios)  like member access. However, for the indexer expression, there is still cost from compiling the expression every time. This fix for this issues is not straightforward and needs more analysis. As indexer expressions are not the targeted scenarios, this performance issue will be closed as resolved. However we will track the issues with the indexer expressions separately based on the priority.

PR #4298, opened 2016-03-16
[Perf] Cache Member Access Expression Text to avoid ExpressionHelper.GetExpressionText()

mnltejaswini:
Part 1 Fix for  #3921

Benchmark Test app : https://github.com/aspnet/Performance/tree/dev/testapp/BigViews

loadtest -k -n 3000 --rps 250  http://localhost:5000

There are ~105MB of memory allocations coming from ExpressionHelper.GetExpressionText() which include types of string and stack. For simple lambda expression like member access, the expression text is similar to the expression and doesn't need any kind of evaluations. However, for expressions that include indexer of any kind (variable/constant indexer) we may need additional parsing.  The pull request has changes to cache the expression text for member access expressions. We do not cache the expression text if the expression has indexer.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor/RazorPageOfT.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor/RazorPageOfT.cs
@@ -38,6 +38,12 @@ namespace Microsoft.AspNetCore.Mvc.Razor
         public ViewDataDictionary<TModel> ViewData { get; set; }
 
         /// <summary>
+        /// Gets or sets the expression text cache for model expressions.
+        /// </summary>
+        [RazorInject]
+        private ExpressionTextCache ExpressionTextCache { get; set; }
+
+        /// <summary>
         /// Returns a <see cref="ModelExpression"/> instance describing the given <paramref name="expression"/>.
         /// </summary>
         /// <typeparam name="TValue">The type of the <paramref name="expression"/> result.</typeparam>
@@ -59,7 +65,7 @@ namespace Microsoft.AspNetCore.Mvc.Razor
                 _provider = Context.RequestServices.GetRequiredService<IModelMetadataProvider>();
             }
 
-            var name = ExpressionHelper.GetExpressionText(expression);
+            var name = ExpressionHelper.GetExpressionText(expression, ExpressionTextCache);
             var modelExplorer = ExpressionMetadataProvider.FromLambdaExpression(expression, ViewData, _provider);
             if (modelExplorer == null)
             {
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/DependencyInjection/MvcViewFeaturesMvcCoreBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/DependencyInjection/MvcViewFeaturesMvcCoreBuilderExtensions.cs
@@ -103,6 +103,7 @@ namespace Microsoft.Extensions.DependencyInjection
             services.TryAddTransient<IHtmlHelper, HtmlHelper>();
             services.TryAddTransient(typeof(IHtmlHelper<>), typeof(HtmlHelper<>));
             services.TryAddSingleton<IHtmlGenerator, DefaultHtmlGenerator>();
+            services.TryAddSingleton<ExpressionTextCache>();
 
             //
             // JSON Helper
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionHelper.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionHelper.cs
@@ -7,9 +7,8 @@ using System.Globalization;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
-using Microsoft.AspNetCore.Mvc.ViewFeatures.Internal;
 
-namespace Microsoft.AspNetCore.Mvc.ViewFeatures
+namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
 {
     public static class ExpressionHelper
     {
@@ -21,11 +20,24 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
 
         public static string GetExpressionText(LambdaExpression expression)
         {
+            return GetExpressionText(expression, expressionTextCache: null);
+        }
+
+        public static string GetExpressionText(LambdaExpression expression, ExpressionTextCache expressionTextCache)
+        {
             if (expression == null)
             {
                 throw new ArgumentNullException(nameof(expression));
             }
 
+            string expressionText;
+            if (expressionTextCache != null &&
+                expressionTextCache.Entries.TryGetValue(expression, out expressionText))
+            {
+                return expressionText;
+            }
+
+            var containsIndexers = false;
             // Split apart the expression string for property/field accessors to create its name
             var nameParts = new Stack<string>();
             var part = expression.Body;
@@ -34,6 +46,7 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
             {
                 if (part.NodeType == ExpressionType.Call)
                 {
+                    containsIndexers = true;
                     var methodExpression = (MethodCallExpression)part;
                     if (!IsSingleArgumentIndexer(methodExpression))
                     {
@@ -50,6 +63,7 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
                 }
                 else if (part.NodeType == ExpressionType.ArrayIndex)
                 {
+                    containsIndexers = true;
                     var binaryExpression = (BinaryExpression)part;
 
                     nameParts.Push(
@@ -89,7 +103,6 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
                 }
                 else
                 {
-                    // Unsupported.
                     break;
                 }
             }
@@ -100,12 +113,18 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
                 nameParts.Pop();
             }
 
+            expressionText = string.Empty;
             if (nameParts.Count > 0)
             {
-                return nameParts.Aggregate((left, right) => left + right).TrimStart('.');
+                expressionText = nameParts.Aggregate((left, right) => left + right).TrimStart('.');
             }
 
-            return string.Empty;
+            if (expressionTextCache != null && !containsIndexers)
+            {
+                expressionTextCache.Entries.TryAdd(expression, expressionText);
+            }
+
+            return expressionText;
         }
 
         private static string GetIndexerInvocation(
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionTextCache.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionTextCache.cs
@@ -1,0 +1,109 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq.Expressions;
+using Microsoft.Extensions.Internal;
+
+namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
+{
+    /// <summary>
+    /// This class holds the cache for the expression text that is computed by ExpressionHelper.
+    /// </summary>
+    public class ExpressionTextCache
+    {
+        /// <inheritdoc />
+        public ConcurrentDictionary<LambdaExpression, string> Entries { get; } = 
+            new ConcurrentDictionary<LambdaExpression, string>(LambdaExpressionComparer.Instance);
+
+        // This comparer is tightly coupled with the logic of ExpressionHelper.GetExpressionText.
+        // It is not designed to accurately compare any two arbitrary LambdaExpressions.
+        private class LambdaExpressionComparer : IEqualityComparer<LambdaExpression>
+        {
+            public static readonly LambdaExpressionComparer Instance = new LambdaExpressionComparer();
+
+            public bool Equals(LambdaExpression lambdaExpression1, LambdaExpression lambdaExpression2)
+            {
+                if (ReferenceEquals(lambdaExpression1,lambdaExpression2))
+                {
+                    return true;
+                }
+                // We will cache only pure member access expressions. Hence we compare two expressions 
+                // to be equal only if they are identical member access expressions.
+                var expression1 = lambdaExpression1.Body;
+                var expression2 = lambdaExpression2.Body;
+
+                while (true)
+                {
+                    if (expression1 == null || expression2 == null)
+                    {
+                        return false;
+                    }
+
+                    if (expression1.NodeType != expression2.NodeType)
+                    {
+                        return false;
+                    }
+
+                    if (expression1.NodeType == ExpressionType.MemberAccess)
+                    {
+                        var memberExpression1 = (MemberExpression)expression1;
+                        var memberName1 = memberExpression1.Member.Name;
+                        expression1 = memberExpression1.Expression;
+
+                        var memberExpression2 = (MemberExpression)expression2;
+                        var memberName2 = memberExpression2.Member.Name;
+                        expression2 = memberExpression2.Expression;
+
+                        // If identifier contains "__", it is "reserved for use by the implementation" and likely compiler-
+                        // or Razor-generated e.g. the name of a field in a delegate's generated class.
+                        if (memberName1.Contains("__") && memberName2.Contains("__"))
+                        {
+                            return true;
+                        }
+
+                        if (!string.Equals(memberName1, memberName2, StringComparison.OrdinalIgnoreCase))
+                        {
+                            return false;
+                        }
+                    }
+                    else 
+                    {
+                        return true;
+                    }
+                }
+            }
+
+            public int GetHashCode(LambdaExpression lambdaExpression)
+            {
+                var expression = lambdaExpression.Body;
+                var hashCodeCombiner = HashCodeCombiner.Start();
+
+                while (true)
+                {
+                    if (expression != null && expression.NodeType == ExpressionType.MemberAccess)
+                    {
+                        var memberExpression = (MemberExpression)expression;
+                        var memberName = memberExpression.Member.Name;
+
+                        if (memberName.Contains("__"))
+                        {
+                            break;
+                        }
+
+                        hashCodeCombiner.Add(memberName, StringComparer.OrdinalIgnoreCase);
+                        expression = memberExpression.Expression;
+                    }
+                    else
+                    {
+                        break;
+                    }
+                }
+
+                return hashCodeCombiner.CombinedHash;
+            }
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/HtmlHelperOfT.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/HtmlHelperOfT.cs
@@ -15,6 +15,8 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
 {
     public class HtmlHelper<TModel> : HtmlHelper, IHtmlHelper<TModel>
     {
+        private readonly ExpressionTextCache _expressionTextCache;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="HtmlHelper{TModel}"/> class.
         /// </summary>
@@ -24,7 +26,8 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
             IModelMetadataProvider metadataProvider,
             IViewBufferScope bufferScope,
             HtmlEncoder htmlEncoder,
-            UrlEncoder urlEncoder)
+            UrlEncoder urlEncoder,
+            ExpressionTextCache expressionTextCache)
             : base(
                 htmlGenerator,
                 viewEngine,
@@ -33,6 +36,12 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
                 htmlEncoder,
                 urlEncoder)
         {
+            if (expressionTextCache == null)
+            {
+                throw new ArgumentNullException(nameof(expressionTextCache));
+            }
+
+            _expressionTextCache = expressionTextCache;
         }
 
         /// <inheritdoc />
@@ -152,7 +161,7 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
                 new ViewDataDictionary<TModelItem>(ViewData, model: null),
                 MetadataProvider);
 
-            var expressionText = ExpressionHelper.GetExpressionText(expression);
+            var expressionText = ExpressionHelper.GetExpressionText(expression, _expressionTextCache);
             if (modelExplorer == null)
             {
                 throw new InvalidOperationException(Resources.FormatHtmlHelper_NullModelMetadata(expressionText));
@@ -352,7 +361,7 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
                 throw new ArgumentNullException(nameof(expression));
             }
 
-            return ExpressionHelper.GetExpressionText(expression);
+            return ExpressionHelper.GetExpressionText(expression, _expressionTextCache);
         }
 
         protected ModelExplorer GetModelExplorer<TResult>(Expression<Func<TModel, TResult>> expression)
--- a/test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorPageCreateModelExpressionTest.cs
+++ b/test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorPageCreateModelExpressionTest.cs
@@ -11,6 +11,7 @@ using Microsoft.AspNetCore.Mvc.ModelBinding;
 using Microsoft.AspNetCore.Mvc.Rendering;
 using Microsoft.AspNetCore.Mvc.ViewEngines;
 using Microsoft.AspNetCore.Mvc.ViewFeatures;
+using Microsoft.AspNetCore.Mvc.ViewFeatures.Internal;
 using Microsoft.AspNetCore.Routing;
 using Moq;
 using Xunit;
@@ -116,6 +117,9 @@ namespace Microsoft.AspNetCore.Mvc.Razor
             serviceProvider
                 .Setup(real => real.GetService(typeof(IModelMetadataProvider)))
                 .Returns(provider);
+            serviceProvider
+                .Setup(real => real.GetService(typeof(ExpressionTextCache)))
+                .Returns(new ExpressionTextCache());
 
             var httpContext = new Mock<HttpContext>();
             httpContext
--- a/test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Internal/ExpressionHelperTest.cs
+++ b/test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Internal/ExpressionHelperTest.cs
@@ -1,0 +1,242 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq.Expressions;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
+{
+    public class ExpressionHelperTest
+    {
+        private readonly ExpressionTextCache _expressionTextCache = new ExpressionTextCache();
+
+        public static IEnumerable<object[]> CachedExpressions
+        {
+            get
+            {
+                var key = "TestModel";
+                var myModel = new TestModel();
+
+                return new TheoryData<Expression>
+                {
+                    (Expression<Func<TestModel, Category>>)(model => model.SelectedCategory),
+                    (Expression<Func<TestModel, CategoryName>>)(model => model.SelectedCategory.CategoryName),
+                    (Expression<Func<TestModel, int>>)(testModel => testModel.SelectedCategory.CategoryId),
+                    (Expression<Func<TestModel, string>>)(model => model.SelectedCategory.CategoryName.MainCategory),
+                    (Expression<Func<TestModel, string>>)(testModel => key),
+                    (Expression<Func<TestModel, TestModel>>)(m => m),
+                    (Expression<Func<TestModel, Category>>)(m => myModel.SelectedCategory),
+                };
+            }
+        }
+
+        public static IEnumerable<object[]> IndexerExpressions
+        {
+            get
+            {
+                var i = 3;
+                var key = "TestModel";
+                var myModels = new List<TestModel>();
+
+                return new TheoryData<Expression>
+                {
+                    (Expression<Func<IList<TestModel>, Category>>)(model => model[2].SelectedCategory),
+                    (Expression<Func<IList<TestModel>, Category>>)(model => myModels[i].SelectedCategory),
+                    (Expression<Func<IList<TestModel>, CategoryName>>)(testModel => testModel[i].SelectedCategory.CategoryName),
+                    (Expression<Func<TestModel, int>>)(model => model.PreferredCategories[i].CategoryId),
+                    (Expression<Func<IDictionary<string, TestModel>, string>>)(model => model[key].SelectedCategory.CategoryName.MainCategory),
+                };
+            }
+        }
+
+        public static IEnumerable<object[]> EquivalentExpressions
+        {
+            get
+            {
+                var value = "Test";
+                var Model = "Test";
+
+                return new TheoryData<Expression, Expression>
+                {
+                    {
+                        (Expression<Func<TestModel, Category>>)(model => model.SelectedCategory),
+                        (Expression<Func<TestModel, Category>>)(model => model.SelectedCategory)
+                    },
+                    {
+                        (Expression<Func<TestModel, CategoryName>>)(model => model.SelectedCategory.CategoryName),
+                        (Expression<Func<TestModel, CategoryName>>)(model => model.SelectedCategory.CategoryName)
+                    },
+                    {
+                        (Expression<Func<TestModel, int>>)(testModel => testModel.SelectedCategory.CategoryId),
+                        (Expression<Func<TestModel, int>>)(testModel => testModel.SelectedCategory.CategoryId)
+                    },
+                    {
+                        (Expression<Func<TestModel, string>>)(model => model.SelectedCategory.CategoryName.MainCategory),
+                        (Expression<Func<TestModel, string>>)(model => model.SelectedCategory.CategoryName.MainCategory)
+                    },
+                    {
+                        (Expression<Func<TestModel, TestModel>>)(model => model),
+                        (Expression<Func<TestModel, TestModel>>)(m => m)
+                    },
+                    {
+                        (Expression<Func<TestModel, string>>)(model => value),
+                        (Expression<Func<TestModel, string>>)(m => value)
+                    },
+                    {
+                        // These two expressions are not actually equivalent. However ExpressionHelper returns 
+                        // string.Empty for these two expressions and hence they are considered as equivalent by the 
+                        // cache.
+                        (Expression<Func<TestModel, string>>)(m => Model),
+                        (Expression<Func<TestModel, TestModel>>)(m => m)
+                    },
+                };
+            }
+        }
+
+        public static IEnumerable<object[]> NonEquivalentExpressions
+        {
+            get
+            {
+                var value = "test";
+                var key = "TestModel";
+                var Model = "Test";
+                var myModel = new TestModel();
+
+                return new TheoryData<Expression, Expression>
+                {
+                    {
+                        (Expression<Func<TestModel, Category>>)(model => model.SelectedCategory),
+                        (Expression<Func<TestModel, CategoryName>>)(model => model.SelectedCategory.CategoryName)
+                    },
+                    {
+                        (Expression<Func<TestModel, string>>)(model => model.Model),
+                        (Expression<Func<TestModel, string>>)(model => model.Name)
+                    },
+                    {
+                        (Expression<Func<TestModel, CategoryName>>)(model => model.SelectedCategory.CategoryName),
+                        (Expression<Func<TestModel, string>>)(model => value)
+                    },
+                    {
+                        (Expression<Func<TestModel, string>>)(testModel => testModel.SelectedCategory.CategoryName.MainCategory),
+                        (Expression<Func<TestModel, string>>)(testModel => value)
+                    },
+                    {
+                        (Expression<Func<IList<TestModel>, Category>>)(model => model[2].SelectedCategory),
+                        (Expression<Func<TestModel, string>>)(model => model.SelectedCategory.CategoryName.MainCategory)
+                    },
+                    {
+                        (Expression<Func<TestModel, int>>)(testModel => testModel.SelectedCategory.CategoryId),
+                        (Expression<Func<TestModel, Category>>)(model => model.SelectedCategory)
+                    },
+                    {
+                        (Expression<Func<IDictionary<string, TestModel>, string>>)(model => model[key].SelectedCategory.CategoryName.MainCategory),
+                        (Expression<Func<TestModel, Category>>)(model => model.SelectedCategory)
+                    },
+                    {
+                        (Expression<Func<IList<TestModel>, Category>>)(model => model[2].SelectedCategory),
+                        (Expression<Func<IList<TestModel>, Category>>)(model => model[2].SelectedCategory)
+                    },
+                    {
+                        (Expression<Func<TestModel, string>>)(m => Model),
+                        (Expression<Func<TestModel, string>>)(m => m.Model)
+                    },
+                    {
+                        (Expression<Func<TestModel, TestModel>>)(m => m),
+                        (Expression<Func<TestModel, string>>)(m => m.Model)
+                    },
+                    {
+                        (Expression<Func<TestModel, string>>)(m => myModel.Name),
+                        (Expression<Func<TestModel, string>>)(m => m.Name)
+                    },
+                    {
+                        (Expression<Func<TestModel, string>>)(m => key),
+                        (Expression<Func<TestModel, string>>)(m => value)
+                    },
+                    {
+                        (Expression<Func<IDictionary<string, TestModel>, string>>)(model => model[key].SelectedCategory.CategoryName.MainCategory),
+                        (Expression<Func<IDictionary<string, TestModel>, string>>)(model => model[key].SelectedCategory.CategoryName.MainCategory)
+                    },
+                };
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(CachedExpressions))]
+        public void GetExpressionText_CachesExpression(LambdaExpression expression)
+        {
+            // Act - 1
+            var text1 = ExpressionHelper.GetExpressionText(expression, _expressionTextCache);
+
+            // Act - 2
+            var text2 = ExpressionHelper.GetExpressionText(expression, _expressionTextCache);
+
+            // Assert
+            Assert.Same(text1, text2); // Cached
+        }
+
+        [Theory]
+        [MemberData(nameof(IndexerExpressions))]
+        public void GetExpressionText_DoesNotCacheIndexerExpression(LambdaExpression expression)
+        {
+            // Act - 1
+            var text1 = ExpressionHelper.GetExpressionText(expression, _expressionTextCache);
+
+            // Act - 2
+            var text2 = ExpressionHelper.GetExpressionText(expression, _expressionTextCache);
+
+            // Assert
+            Assert.NotSame(text1, text2); // not cached
+        }
+
+        [Theory]
+        [MemberData(nameof(EquivalentExpressions))]
+        public void GetExpressionText_CacheEquivalentExpressions(LambdaExpression expression1, LambdaExpression expression2)
+        {
+            // Act - 1
+            var text1 = ExpressionHelper.GetExpressionText(expression1, _expressionTextCache);
+
+            // Act - 2
+            var text2 = ExpressionHelper.GetExpressionText(expression2, _expressionTextCache);
+
+            // Assert
+            Assert.Same(text1, text2);
+        }
+
+        [Theory]
+        [MemberData(nameof(NonEquivalentExpressions))]
+        public void GetExpressionText_CheckNonEquivalentExpressions(LambdaExpression expression1, LambdaExpression expression2)
+        {
+            // Act - 1
+            var text1 = ExpressionHelper.GetExpressionText(expression1, _expressionTextCache);
+
+            // Act - 2
+            var text2 = ExpressionHelper.GetExpressionText(expression2, _expressionTextCache);
+
+            // Assert
+            Assert.NotSame(text1, text2);
+        }
+
+        private class TestModel
+        {
+            public string Name { get; set; }
+            public string Model { get; set; }
+            public Category SelectedCategory { get; set; }
+
+            public IList<Category> PreferredCategories { get; set; }
+        }
+
+        private class Category
+        {
+            public int CategoryId { get; set; }
+            public CategoryName CategoryName { get; set; }
+        }
+
+        private class CategoryName
+        {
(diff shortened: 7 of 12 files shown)


mnltejaswini:
Benchmark Test app : https://github.com/aspnet/Performance/tree/dev/testapp/BigViews

loadtest -k -n 3000 --rps 250 http://localhost:5000


Before optimization
-------------------------

image

After optimization
----------------------
image

13% savings in total number of bytes and 15% savings in total number of objects allocated.

src/Microsoft.AspNetCore.Mvc.Razor/RazorPageOfT.cs
```diff
@@ -59,7 +61,12 @@ public ModelExpression CreateModelExpression<TValue>(Expression<Func<TModel, TVa
                 _provider = Context.RequestServices.GetRequiredService<IModelMetadataProvider>();
             }
 
-            var name = ExpressionHelper.GetExpressionText(expression);
+            if (_expressionTextCache == null)
+            {
+                _expressionTextCache = Context.RequestServices.GetRequiredService<ExpressionTextCache>();
+            }
```
dougbu: Why not use a new property with a `[RazorInject]` attribute?

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionTextCache.cs
```diff
@@ -0,0 +1,36 @@
...
+    {
+        private readonly ConcurrentDictionary<LambdaExpression, string> _cacheEntries  = 
+                new ConcurrentDictionary<LambdaExpression, string>();
+
+        public string Get(LambdaExpression expression)
+        {
+            if (expression == null)
+            {
+                throw new ArgumentNullException(nameof(expression));
+            }
```
dougbu: If this is `.Internal`, can assume calling code has already done the `null` checking. Feel free to `Debug.Assert()` that.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ExpressionHelper.cs
```diff
@@ -19,13 +19,16 @@ public static string GetExpressionText(string expression)
             return string.Equals(expression, "model", StringComparison.OrdinalIgnoreCase) ? string.Empty : expression;
         }
 
-        public static string GetExpressionText(LambdaExpression expression)
+        public static bool GetExpressionText(LambdaExpression expression, out string expressionText)
```
dougbu: API change makes it look like a `TryGetBlah()` method, which this isn't. Suggest instead passing the `ExpressionTextCache` _in_ and keeping the `isMemberAccessExpressionOnly` details private.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ExpressionHelper.cs
```diff
@@ -19,13 +19,16 @@ public static string GetExpressionText(string expression)
...
         {
             if (expression == null)
             {
                 throw new ArgumentNullException(nameof(expression));
             }
 
+            expressionText = string.Empty;
+            var isMemberAccessExpressionOnly = true;
```
dougbu: Suggest `var containsIndexers = false;` since this will remain `true` for variable access as well.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/HtmlHelper.cs
```diff
@@ -45,7 +45,8 @@ public class HtmlHelper : IHtmlHelper, IViewContextAware
             IModelMetadataProvider metadataProvider,
             IViewBufferScope bufferScope,
             HtmlEncoder htmlEncoder,
-            UrlEncoder urlEncoder)
+            UrlEncoder urlEncoder,
+            ExpressionTextCache expressionTextCache)
```
dougbu: No reason for `HtmlHelper` to have every constructor parameter that `HtmlHelper<T>` has. Leave this out because the class doesn't need or want the `ExpressionTextCache`.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Internal/ExpressionTextCacheTest.cs
```diff
@@ -0,0 +1,101 @@
...
+                data.Add((Expression<Func<TestModel,Category>>)(model => model.SelectedCategory));
+                data.Add((Expression<Func<TestModel, CategoryName>>)(model => model.SelectedCategory.CategoryName));
+                data.Add((Expression<Func<TestModel, int>>)(model => model.SelectedCategory.CategoryId));
+                data.Add((Expression<Func<TestModel, string>>)(model => model.SelectedCategory.CategoryName.MainCategory));
+
+                return data;
+            }
+        }
+
+        public static IEnumerable<object[]> IndexerExpressions
```
dougbu: Minor but it's easier to read if the data sets are immediately above the test methods where they're first used.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionTextCache.cs
```diff
@@ -0,0 +1,36 @@
...
+using System;
+using System.Collections.Concurrent;
+using System.Linq.Expressions;
+
+namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
+{
+    public class ExpressionTextCache
+    {
+        private readonly ConcurrentDictionary<LambdaExpression, string> _cacheEntries  = 
+                new ConcurrentDictionary<LambdaExpression, string>();
```
dougbu: I suspect this is fairly fragile and doesn't reuse the `string`s as much as it could. Performs reference equality which means it (a) won't work in any case hoisting in `ExpressionRewriter` doesn't happen and (b) won't detect that the expressions in `LabelFor(m => FirstName)` and `TextBoxFor(m => FirstName)` have the same name.

Add a test looking up two separate but identical lambda expressions. If I'm correct, will likely find multiple entries in the dictionary (and will need to implement a simple `IEqualityComparer<LambdaExpression>`). Otherwise I'm good :smiley_cat:

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Internal/ExpressionTextCacheTest.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        public static IEnumerable<object[]> MemberAccessExpressions
+        {
+            get
+            {
+                var data = new TheoryData<Expression>();
+                data.Add((Expression<Func<TestModel,Category>>)(model => model.SelectedCategory));
+                data.Add((Expression<Func<TestModel, CategoryName>>)(model => model.SelectedCategory.CategoryName));
+                data.Add((Expression<Func<TestModel, int>>)(model => model.SelectedCategory.CategoryId));
+                data.Add((Expression<Func<TestModel, string>>)(model => model.SelectedCategory.CategoryName.MainCategory));
```
dougbu: - Add a test of `var value = "valuable"; ... model => value ...` and use a different lambda parameter e.g. `q` at least once.
- Rather verbose. Why not use dictionary init syntax?

dougbu:
:watch:
Please also show the before and after dotMemory snapshots in a comment.

dougbu:
Oops, ignore my second comment about the dotMemory snapshots you've already included.

dougbu:
At `--rps 250` aren't many of the allocations un-collected, making it hard to compare runs? If not, you must have a rather hefty test rig!

src/Microsoft.AspNetCore.Mvc.ViewFeatures/DependencyInjection/MvcViewFeaturesMvcCoreBuilderExtensions.cs
```diff
@@ -105,6 +105,11 @@ internal static void AddViewServices(IServiceCollection services)
             services.TryAddSingleton<IHtmlGenerator, DefaultHtmlGenerator>();
 
             //
+            // Expression Helper
+            //
```
rynowak: This doesn't need it's own 'block' here it's not a top-level feature. Group it with `IHtmlGenerator` and the other misc infrastructure.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionTextCache.cs
rynowak: We **cannot** rely on reference equality here - that will result in unbounded memory growth.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionTextCache.cs
```diff
@@ -0,0 +1,86 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq.Expressions;
+
+namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
+{
+    public class ExpressionTextCache
+    {
+        public readonly ConcurrentDictionary<LambdaExpression, string> Entries  = 
+                new ConcurrentDictionary<LambdaExpression, string>(LambdaExpressionComparer.Instance);
```
dougbu: Odd to make a field `public`. Suggest instead
```c#
public ConcurrentDictionary<LambdaExpression, string> Entries { get; } = ...;
```

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionTextCache.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        private class LambdaExpressionComparer : IEqualityComparer<LambdaExpression>
```
dougbu: Move to the bottom of the file. Reasonable ordering w/in a `class` isn't mentioned in our [Engineering Guidelines](https://github.com/aspnet/Home/wiki/Engineering-guidelines) but is enforced for new code.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionTextCache.cs
```diff
@@ -0,0 +1,86 @@
...
+                        expression = null;
+                    }
+                    else
+                    {
+                        return lambdaExpression.GetHashCode();
+                    }
+                }
+
+                return hashCode;
+            }
```
dougbu: Rewrite this method to use `HashCodeCombiner`. You'll see a number of examples in this repo.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionTextCache.cs
```diff
@@ -0,0 +1,86 @@
...
+                        var memberExpression2 = (MemberExpression)expression2;
+                        expression2 = memberExpression2.Expression;
+                    }
+                    else if (expression1.NodeType == ExpressionType.Parameter)
+                    {
+                        expression1 = expression2 = null;
+                    }
+                    else
+                    {
+                        // should never come here.
```
dougbu: Will often reach here since code should check the cache before determining whether an expression includes an indexer or is supported at all. Remove the comment or change it to `// Unsupported.`

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionTextCache.cs
```diff
@@ -0,0 +1,86 @@
...
+                        return false;
+                    }
+                }
+
+                return true;
+            }
```
dougbu: This comparison does not match the hash code implementation and looks generally scary:
- Ignores `Member.Name` of a `MemberExpression`.
- Won't handle `m => m` correctly because it assumes the top-level node is a `MemberExpression`.
- Performs a reference equality comparison on an `Expression` and then checks parts of the expression.
- Does not check whether the `NodeType`s of the two expressions match.

The last two bullets could be fixed by changing [line 30](https://github.com/aspnet/Mvc/pull/4298/files#diff-cbc498e6135dcff3a95d99f007858dd4R30) to
``` c#
if (expression1.NodeType != expression2.NodeType)
```

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionTextCache.cs
```diff
@@ -0,0 +1,86 @@
...
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq.Expressions;
+
+namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
+{
+    public class ExpressionTextCache
```
dougbu: Add `/// doc comments`

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionTextCache.cs
```diff
@@ -0,0 +1,86 @@
...
+                    {
+                        var memberExpression1 = (MemberExpression)expression1;
+                        expression1 = memberExpression1.Expression;
+
+                        var memberExpression2 = (MemberExpression)expression2;
+                        expression2 = memberExpression2.Expression;
+                    }
+                    else if (expression1.NodeType == ExpressionType.Parameter)
+                    {
+                        expression1 = expression2 = null;
```
dougbu: Weird. Just `break`. In fact the loop should just be
``` c#
    while (true)
    {
    }
```

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionTextCache.cs
```diff
@@ -0,0 +1,86 @@
...
+                    {
+                        var memberExpression = (MemberExpression)expression;
+                        hashCode ^= memberExpression.Member.Name.GetHashCode();
+                        expression = memberExpression.Expression;
+                    }
+                    else if (expression.NodeType == ExpressionType.Parameter)
+                    {
+                        var parameterExpression = (ParameterExpression)expression;
+                        hashCode ^= parameterExpression.Name.GetHashCode();
+                        expression = null;
```
dougbu: - Just `break` here and make the loop `while (true)`.
- Name of the lambda parameter doesn't matter and isn't checked in `Equals()`. Don't include it in the hash code.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionTextCache.cs
dougbu: OIC, nothing else is in this `class`. Fine as is.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ExpressionHelper.cs
```diff
@@ -21,11 +21,26 @@ public static string GetExpressionText(string expression)
 
         public static string GetExpressionText(LambdaExpression expression)
         {
+            return GetExpressionText(expression, null);
```
dougbu: Name the `null` parameter i.e. `expressionTextCache: null`.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ExpressionHelper.cs
```diff
@@ -21,11 +21,26 @@ public static string GetExpressionText(string expression)
...
             }
 
+            string expressionText;
+            if (expressionTextCache != null &&
+                expressionTextCache.Entries.TryGetValue(expression, out expressionText))
+            {
+                return expressionText;
+            }
+
+            expressionText = string.Empty;
```
dougbu: Too far from where it's used. Do this in an `else` block for the `nameParts.Count > 0` condition.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ExpressionHelper.cs
```diff
@@ -163,5 +188,37 @@ public static bool IsSingleArgumentIndexer(Expression expression)
                 property => (string.Equals(defaultMember.MemberName, property.Name, StringComparison.Ordinal) &&
                     property.GetMethod == methodExpression.Method));
         }
+
+        public static bool EqualsTo(this Expression expression1, Expression expression2)
```
dougbu: Why does this exist? And why is it an extension method and in this class? Inline the useful parts into `LambdaExpressionComparer`.

Note the Engineering Guidelines again: We define extension methods in a few special cases, none where a `static` method would be sufficient.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/HtmlHelper.cs
```diff
@@ -155,6 +155,8 @@ public ITempDataDictionary TempData
         /// <inheritdoc />
         public UrlEncoder UrlEncoder { get; }
 
+        
+
```
dougbu: Revert changes to this file.

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorPageCreateModelExpressionTest.cs
```diff
@@ -95,6 +96,7 @@ private static TestRazorPage CreatePage(ViewContext viewContext)
             {
                 ViewContext = viewContext,
                 ViewData = (ViewDataDictionary<RazorPageCreateModelExpressionModel>)viewContext.ViewData,
+                ExpressionTextCache = new ExpressionTextCache()
```
dougbu: Include trailing commas for every property setting when using init syntax.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Internal/ExpressionTextCacheTest.cs
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        public static IEnumerable<object[]> IdenticalExpressions
+        {
+            get
+            {
+                return new TheoryData<Expression, Expression>
+                {
+                    { (Expression<Func<TestModel, Category>>)(model => model.SelectedCategory),
+                        (Expression<Func<TestModel, Category>>)(model => model.SelectedCategory) },
```
dougbu: Curly braces always go on their own lines (well, except when close parentheses or a semicolon belong on the same line). So
``` c#
{
    (Expression<Func<TestModel, Category>>)(model => model.SelectedCategory),
    (Expression<Func<TestModel, Category>>)(model => model.SelectedCategory)
},
```

Separately, are the casts actually necessary?

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Internal/ExpressionTextCacheTest.cs
```diff
@@ -0,0 +1,189 @@
...
+                var value = "test";
+                var key = "TestModel";
+
+                return new TheoryData<Expression, Expression>
+                {
+                    { (Expression<Func<TestModel, Category>>)(model => model.SelectedCategory),
+                        (Expression<Func<TestModel, CategoryName>>)(model => model.SelectedCategory.CategoryName) },
+
+                    { (Expression<Func<TestModel, CategoryName>>)(model => model.SelectedCategory.CategoryName),
+                        (Expression<Func<TestModel, string>>)(model => value) },
```
dougbu: Should test that `model => value` and `m => value` are identical.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Internal/ExpressionTextCacheTest.cs
```diff
@@ -0,0 +1,189 @@
...
+                return new TheoryData<Expression, Expression>
+                {
+                    { (Expression<Func<TestModel, Category>>)(model => model.SelectedCategory),
+                        (Expression<Func<TestModel, Category>>)(model => model.SelectedCategory) },
+
+                    { (Expression<Func<TestModel, CategoryName>>)(model => model.SelectedCategory.CategoryName),
+                        (Expression<Func<TestModel, CategoryName>>)(model => model.SelectedCategory.CategoryName) },
+
+                    { (Expression<Func<TestModel, int>>)(testmodel => testmodel.SelectedCategory.CategoryId),
+                        (Expression<Func<TestModel, int>>)(testmodel => testmodel.SelectedCategory.CategoryId) },
```
dougbu: Mix up the lambda parameter w/in these expressions e.g. `testModel => testModel` (note capital `M`) and `model => model` should be identical.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Internal/ExpressionTextCacheTest.cs
```diff
@@ -0,0 +1,189 @@
...
+                        (Expression<Func<TestModel, string>>)(testmodel => value) },
+
+                    { (Expression<Func<IList<TestModel>, Category>>)(model => model[2].SelectedCategory),
+                        (Expression<Func<TestModel, string>>)(model => model.SelectedCategory.CategoryName.MainCategory) },
+
+                    { (Expression<Func<TestModel, int>>)(testmodel => testmodel.SelectedCategory.CategoryId),
+                        (Expression<Func<TestModel, Category>>)(model => model.SelectedCategory) },
+
+                    { (Expression<Func<IDictionary<string, TestModel>, string>>)(model => model[key].SelectedCategory.CategoryName.MainCategory),
+                        (Expression<Func<TestModel, Category>>)(model => model.SelectedCategory) }
```
dougbu: Confirm a couple of indexer expressions, even if they look identical, result in `false`.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Internal/ExpressionTextCacheTest.cs
```diff
@@ -0,0 +1,189 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq.Expressions;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
+{
+    public class ExpressionTextCacheTest
```
dougbu: Actually `ExpressionHelperTest`

dougbu:
:watch:

dougbu:
:up::date:

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionTextCache.cs
```diff
@@ -0,0 +1,108 @@
...
+        {
+            public static readonly LambdaExpressionComparer Instance = new LambdaExpressionComparer();
+
+            public bool Equals(LambdaExpression lambdaExpression1, LambdaExpression lambdaExpression2)
+            {
+                // We will cache only pure member access expressions. Hence we compare two expressions 
+                // to be equal only if they are identical member access expressions.
+                var expression1 = lambdaExpression1.Body;
+                var expression2 = lambdaExpression2.Body;
+                var identical = false;
```
dougbu: `isEquivalent` because `identical` is a stretch.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionTextCache.cs
```diff
@@ -0,0 +1,108 @@
...
+
+                while (true)
```
dougbu: Return `true` if `ReferenceEquals(expression1, expression2)` before diving into checking the details.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionTextCache.cs
```diff
@@ -0,0 +1,108 @@
...
+                        var memberName2 = memberExpression2.Member.Name;
+                        expression2 = memberExpression2.Expression;
+
+                        if (memberName1.Contains("__") && memberName2.Contains("__"))
+                        {
+                            identical = true;
+                            break;
+                        }
+
+                        if (!memberName1.Equals(memberName2))
```
dougbu: `!string.Equals(memberName1, memberName2, StringComparison.OrdinalIgnoreCase)`

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionTextCache.cs
```diff
@@ -0,0 +1,108 @@
...
+using System.Diagnostics;
+using System.Linq.Expressions;
+using Microsoft.Extensions.Internal;
+
+namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
+{
+    /// <summary>
+    /// This class holds the cache for the expression text that is computed by ExpressionHelper.
+    /// Expressions of node type member, constant and parameter are cached.
+    /// However expressions that include indexers are not cached.
```
dougbu: Remove everything but the first sentence since they are about how this class is _used_, not about the class. Alternatively soften the wording to something about how the class is meant to be used.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionTextCache.cs
```diff
@@ -0,0 +1,108 @@
...
+                var hashCodeCombiner = HashCodeCombiner.Start();
+
+                while (true)
+                {
+                    if (expression != null && expression.NodeType == ExpressionType.MemberAccess)
+                    {
+                        var memberExpression = (MemberExpression)expression;
+                        var memberName = memberExpression.Member.Name;
+                        // If identifier contains "__", it is "reserved for use by the implementation" and likely compiler-
+                        // or Razor-generated e.g. the name of a field in a delegate's generated class.
```
dougbu: Move these comments above [line 57](https://github.com/aspnet/Mvc/pull/4298/files#diff-cbc498e6135dcff3a95d99f007858dd4R57). Good to have just one comment like this but it should be w/ the first related code.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionTextCache.cs
```diff
@@ -0,0 +1,108 @@
...
+                        if (memberName.Contains("__"))
+                        {
+                            break;
+                        }
+                        hashCodeCombiner.Add(memberName, StringComparer.Ordinal);
```
dougbu: `OrdinalIgnoreCase`

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionTextCache.cs
```diff
@@ -0,0 +1,108 @@
...
+    public class ExpressionTextCache
+    {
+        public ConcurrentDictionary<LambdaExpression, string> Entries { get; } = 
+            new ConcurrentDictionary<LambdaExpression, string>(LambdaExpressionComparer.Instance);
+
+        private class LambdaExpressionComparer : IEqualityComparer<LambdaExpression>
+        {
+            public static readonly LambdaExpressionComparer Instance = new LambdaExpressionComparer();
+
+            public bool Equals(LambdaExpression lambdaExpression1, LambdaExpression lambdaExpression2)
```
dougbu: `/// <inheritdoc />`

dougbu: Hmm, maybe not. Forgot this was a `private` class.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionTextCache.cs
```diff
@@ -0,0 +1,108 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq.Expressions;
+using Microsoft.Extensions.Internal;
```
dougbu: What requires `System`, `System.Collections.Generic`, or `System.Diagnostics`?

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ExpressionHelper.cs
```diff
@@ -21,11 +21,26 @@ public static string GetExpressionText(string expression)
...
             }
 
+            string expressionText;
+            if (expressionTextCache != null &&
+                expressionTextCache.Entries.TryGetValue(expression, out expressionText))
+            {
+                return expressionText;
+            }
+
+
```
dougbu: Just one blank line.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/HtmlHelperOfT.cs
```diff
@@ -33,11 +34,19 @@ public class HtmlHelper<TModel> : HtmlHelper, IHtmlHelper<TModel>
...
+                throw new ArgumentNullException(nameof(expressionTextCache));
+            }
+            ExpressionTextCache = expressionTextCache;
         }
 
         /// <inheritdoc />
         public new ViewDataDictionary<TModel> ViewData { get; private set; }
 
+        /// <inheritdoc />
+        public ExpressionTextCache ExpressionTextCache { get; }
```
dougbu: Just use a `private readonly` field. Can't use `/// <inheritdoc />` because it's not part of the interface but that's minor. No reason (so far) to expose this.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionTextCache.cs
```diff
@@ -0,0 +1,108 @@
...
+
+namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
```
dougbu: This should be in the same namespace as `ExpressionHelper`. I'm not positive why that one isn't `.Internal` but it isn't.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Internal/ExpressionHelperTest.cs
```diff
@@ -0,0 +1,227 @@
...
+    public class ExpressionHelperTest
+    {
+        private readonly ExpressionTextCache _expressionTextCache = new ExpressionTextCache();
+
+        public static IEnumerable<object[]> CachedExpressions
+        {
+            get
+            {
+                var key = "TestModel";
+
```
dougbu: Add a `var TestModel myModel` and confirm expressions using it (e.g. `model => myModel.SelectedCategory`) are cached.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Internal/ExpressionHelperTest.cs
```diff
@@ -0,0 +1,227 @@
...
+                };
+            }
+        }
+
+        public static IEnumerable<object[]> IndexerExpressions
+        {
+            get
+            {
+                var i = 3;
+                var key = "TestModel";
```
dougbu: Add a `var List<TestModel> myModel` and confirm expressions using it (e.g. `model => myModel[i].SelectedCategory`) are not cached.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Internal/ExpressionHelperTest.cs
```diff
@@ -0,0 +1,227 @@
...
+                {
+                    (Expression<Func<IList<TestModel>, Category>>)(model => model[2].SelectedCategory),
+                    (Expression<Func<IList<TestModel>, CategoryName>>)(testModel => testModel[i].SelectedCategory.CategoryName),
+                    (Expression<Func<TestModel, int>>)(model => model.PreferredCategories[i].CategoryId),
+                    (Expression<Func<IDictionary<string, TestModel>, string>>)(model => model[key].SelectedCategory.CategoryName.MainCategory),
+                };
+            }
+        }
+
+        public static IEnumerable<object[]> IdenticalExpressions
```
dougbu: `EquivalentExpressions`

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Internal/ExpressionHelperTest.cs
```diff
@@ -0,0 +1,227 @@
...
+                        (Expression<Func<TestModel, TestModel>>)(model => model),
+                        (Expression<Func<TestModel, TestModel>>)(m => m)
+                    },
+                    {
+                        (Expression<Func<TestModel, string>>)(model => value),
+                        (Expression<Func<TestModel, string>>)(m => value)
+                    },
+                    {
+                        (Expression<Func<TestModel, string>>)(m => Model),
+                        (Expression<Func<TestModel, TestModel>>)(m => m)
```
dougbu: Why do these two expressions use the same slot in the cache? Top-level nodes in the Linq expression should have different `NodeType`s.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Internal/ExpressionHelperTest.cs
```diff
@@ -0,0 +1,227 @@
...
+        [MemberData(nameof(IdenticalExpressions))]
+        public void GetExpressionText_CacheIdenticalExpressions(LambdaExpression expression1, LambdaExpression expression2)
+        {
+            // Act - 1
+            var text1 = ExpressionHelper.GetExpressionText(expression1, _expressionTextCache);
+
+            // Act - 2
+            var text2 = ExpressionHelper.GetExpressionText(expression2, _expressionTextCache);
+
+
```
dougbu: Just one blank line.

dougbu:
:watch: mostly for minor bits
/cc @rynowak please have a look too.

mnltejaswini:
:up::date

mnltejaswini:
:up::date:

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ExpressionTextCache.cs
```diff
@@ -0,0 +1,111 @@
...
+using Microsoft.Extensions.Internal;
+
+namespace Microsoft.AspNetCore.Mvc.ViewFeatures
+{
+    /// <summary>
+    /// This class holds the cache for the expression text that is computed by ExpressionHelper.
+    /// </summary>
+    public class ExpressionTextCache
+    {
+        public ConcurrentDictionary<LambdaExpression, string> Entries { get; } =
```
dougbu: `/// docs`

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ExpressionTextCache.cs
```diff
@@ -0,0 +1,111 @@
...
+                }
+                // We will cache only pure member access expressions. Hence we compare two expressions 
+                // to be equal only if they are identical member access expressions.
+                var expression1 = lambdaExpression1.Body;
+                var expression2 = lambdaExpression2.Body;
+                var isEquivalent = false;
+
+                while (true)
+                {
+                    if (expression1 == null || expression2 == null)
```
dougbu: Just wondering: Are either `((LambdaExpression)x).Body == null` or `((MemberExpression)y).Expression == null` actual cases? (Realize `ExpresisonHelper` uses `part != null` for its loop condition but am not sure that guard is necessary.)

If this can really happen, `expression1 == null && expression2 == null` should be considered equivalent.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/HtmlHelperOfT.cs
```diff
@@ -33,11 +34,19 @@ public class HtmlHelper<TModel> : HtmlHelper, IHtmlHelper<TModel>
...
+                throw new ArgumentNullException(nameof(expressionTextCache));
+            }
+
+            _expressionTextCache = expressionTextCache;
         }
 
         /// <inheritdoc />
         public new ViewDataDictionary<TModel> ViewData { get; private set; }
 
+        private readonly ExpressionTextCache _expressionTextCache;
```
dougbu: Move this to the top of the file, with the other fields.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ViewFeatures/ExpressionHelperTest.cs
```diff
@@ -0,0 +1,235 @@
...
+                        (Expression<Func<TestModel, TestModel>>)(model => model),
+                        (Expression<Func<TestModel, TestModel>>)(m => m)
+                    },
+                    {
+                        (Expression<Func<TestModel, string>>)(model => value),
+                        (Expression<Func<TestModel, string>>)(m => value)
+                    },
+                    {
+                        (Expression<Func<TestModel, string>>)(m => Model),
+                        (Expression<Func<TestModel, TestModel>>)(m => m)
```
dougbu: Add a comment about the detail that these two are _not_ actually equivalent. Mention the `string.Empty` special case in `ExpressionHelper`.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ViewFeatures/ExpressionHelperTest.cs
```diff
@@ -0,0 +1,235 @@
...
+        [MemberData(nameof(CachedExpressions))]
+        public void GetExpressionText_CachesExpression(LambdaExpression expression)
+        {
+            // Act - 1
+            var text1 = ExpressionHelper.GetExpressionText(expression, _expressionTextCache);
+
+            // Act - 2
+            var text2 = ExpressionHelper.GetExpressionText(expression, _expressionTextCache);
+
+            Assert.Same(text1, text2); // Cached
```
dougbu: Add `// Assert` in all of these test methods.

dougbu:
:shipit: for me, with some small touch-ups. Please have @rynowak read this through as well.

src/Microsoft.AspNetCore.Mvc.Razor/RazorPageOfT.cs
```diff
@@ -38,6 +38,12 @@ public TModel Model
         public ViewDataDictionary<TModel> ViewData { get; set; }
 
         /// <summary>
+        /// Gets or sets the expression text cache for model expressions.
+        /// </summary>
+        [RazorInject]
+        public ExpressionTextCache ExpressionTextCache { get; set; }
```
rynowak: This will show up in the user programming surface in Razor will it not?

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ExpressionHelper.cs
```diff
@@ -100,12 +114,18 @@ public static string GetExpressionText(LambdaExpression expression)
                 nameParts.Pop();
             }
 
+            expressionText = string.Empty;
             if (nameParts.Count > 0)
             {
-                return nameParts.Aggregate((left, right) => left + right).TrimStart('.');
+                expressionText = nameParts.Aggregate((left, right) => left + right).TrimStart('.');
```
rynowak: Given that this is going to still run for lots of expressions without a cache hit, can we just get rid of the silly Linq?

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ExpressionTextCache.cs
```diff
@@ -0,0 +1,111 @@
...
+        public ConcurrentDictionary<LambdaExpression, string> Entries { get; } = 
+            new ConcurrentDictionary<LambdaExpression, string>(LambdaExpressionComparer.Instance);
+
+        private class LambdaExpressionComparer : IEqualityComparer<LambdaExpression>
+        {
+            public static readonly LambdaExpressionComparer Instance = new LambdaExpressionComparer();
+
+            public bool Equals(LambdaExpression lambdaExpression1, LambdaExpression lambdaExpression2)
+            {
+                if (ReferenceEquals(lambdaExpression1,lambdaExpression2))
```
rynowak: spacing

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ExpressionTextCache.cs
```diff
@@ -0,0 +1,111 @@
...
+                        }
+
+                        if (!string.Equals(memberName1, memberName2, StringComparison.OrdinalIgnoreCase))
+                        {
+                            break;
+                        }
+                    }
+                    else 
+                    {
+                        isEquivalent = true;
```
rynowak: Wait so if the `NodeType` is **not** a `MemberAccess` it's always equivalent? I'd rather see whitelisting of what we want to handle and throw or fail for the others

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ExpressionTextCache.cs
```diff
@@ -0,0 +1,111 @@
...
+                        var memberName1 = memberExpression1.Member.Name;
+                        expression1 = memberExpression1.Expression;
+
+                        var memberExpression2 = (MemberExpression)expression2;
+                        var memberName2 = memberExpression2.Member.Name;
+                        expression2 = memberExpression2.Expression;
+
+                        // If identifier contains "__", it is "reserved for use by the implementation" and likely compiler-
+                        // or Razor-generated e.g. the name of a field in a delegate's generated class.
+                        if (memberName1.Contains("__") && memberName2.Contains("__"))
```
rynowak: Is this special casing for the hoisting case? Why is this beneficial? Why are two 'reserved' names always the same?

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ExpressionTextCache.cs
```diff
@@ -0,0 +1,111 @@
...
+                        {
+                            isEquivalent = true;
+                            break;
```
rynowak: This code would be much more understandable just returning a value instead of relying on `break` for control flow. `isEquivalent` is never set to true without a break, so you don't even need it.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ExpressionHelper.cs
mnltejaswini: All this will be optimized in part 2 where we replace the Stack with Stringbuilder

rynowak: :ok:

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ExpressionTextCache.cs
rynowak: Discussed in person, this is very tightly coupled to what `GetExpressionText` does.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ExpressionTextCache.cs
rynowak: Discussed in person, this doesn't make a big difference in the end. This code relies on `GetExpressionText` to be correct, and only add things to the cache when safe.

rynowak: And we can't throw here because we need this to run on arbitrary expressions without throwing.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ExpressionTextCache.cs
mnltejaswini: All the API s expect Expression<Func<TModel,TResult>> and hence ((LambdaExpression)x).Body == null   is not possible. 
((MemberExpression)y).Expression == null   is possible if we are accessing the static fields/properties. However, this is not currently a supported scenario.

mnltejaswini:
:up::date:

src/Microsoft.AspNetCore.Mvc.Razor/RazorPageOfT.cs
dougbu: Good point @rynowak. @mnltejaswini please make this property `private`. That'll work because `RazorPageActivator` includes non-`public` properties.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionTextCache.cs
```diff
@@ -0,0 +1,109 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
```
dougbu: Curious: What needs `System.Collections.Generic`?

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Internal/ExpressionHelperTest.cs
dougbu: thx

dougbu:
:shipit: w/ https://github.com/aspnet/Mvc/pull/4298#discussion-diff-57221115 change

rynowak:
:shipit: with https://github.com/aspnet/Mvc/pull/4298#discussion_r57221115 change

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionTextCache.cs
mnltejaswini: IEqualityComparer<T> is in Generic namespace

mnltejaswini:
:up::date:

src/Microsoft.AspNetCore.Mvc.Razor/RazorPageOfT.cs
```diff
@@ -20,6 +20,9 @@ public abstract class RazorPage<TModel> : RazorPage
     {
         private IModelMetadataProvider _provider;
 
+        [RazorInject]
+        private ExpressionTextCache _expressionTextCache { get; set; }
```
dougbu: - This is still a property and should be named `ExpressionTextCache`.
- Was placed correctly before. Add `private` members last.

dougbu:
:watch:

Please rerun the command-line build to confirm the lack of `ExpressionTextCache` in tests of `RazorPage<T>` doesn't break anything.

mnltejaswini:
I did make sure that build was success after making it private.

mnltejaswini:
:up::date:

dougbu:
:shipit:

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionHelper.cs
```diff
@@ -100,12 +113,18 @@ public static string GetExpressionText(LambdaExpression expression)
                 nameParts.Pop();
             }
 
+            expressionText = string.Empty;
             if (nameParts.Count > 0)
             {
-                return nameParts.Aggregate((left, right) => left + right).TrimStart('.');
+                expressionText = nameParts.Aggregate((left, right) => left + right).TrimStart('.');
```
davidfowl: string.join?

rynowak: Earlier comment on this got buried. Making this better is part of the next wave of changes

PR #4302, opened 2016-03-16
Remove DependencyModel dependency

pranavkm:
cc @Eilon \ @pakrym


pranavkm:
Verified Mvc against dotnet version 1.0.0-beta-001817.

pranavkm:
cc @NTaylorMullen

pakrym:
:shipit: if it works

src/Microsoft.AspNetCore.Mvc.Core/Infrastructure/DefaultAssemblyProvider.cs
```diff
@@ -88,7 +88,7 @@ public IEnumerable<Assembly> CandidateAssemblies
         /// </summary>
         /// <returns>A set of <see cref="Library"/>.</returns>
         // Internal for unit testing
-        protected internal virtual IEnumerable<RuntimeLibrary> GetCandidateLibraries()
```
NTaylorMullen: Why remove protected? It wasn't consumed in the tests that way previously. It was always called via the internal accessor method.

src/Microsoft.AspNetCore.Mvc.Razor/Internal/DependencyContextRazorViewEngineOptionsSetup.cs
```diff
@@ -80,7 +80,7 @@ private static void SetCompilationOptions(RazorViewEngineOptions options, Extens
         {
             var parseOptions = options.ParseOptions;
             parseOptions = parseOptions.WithPreprocessorSymbols(
-                parseOptions.PreprocessorSymbolNames.Concat(compilationOptions.Defines));
+                parseOptions.PreprocessorSymbolNames.Concat(compilationOptions.Defines ?? Enumerable.Empty<string>()));
```
NTaylorMullen: ?

NTaylorMullen:
Might be worth adding a loud header at the top of each of the copied files so no one attempts to modify them inbetween now and when we revert this change.

src/Microsoft.AspNetCore.Mvc.Razor/Internal/DependencyContextRazorViewEngineOptionsSetup.cs
pakrym: `compilationOptions.Defines` shouldn't be null, that's a bug I need to fix if it is

src/Microsoft.AspNetCore.Mvc.Core/Infrastructure/DefaultAssemblyProvider.cs
pranavkm: The type is `internal` now. So we can't expose it unless we make our private implementation public.

src/Microsoft.AspNetCore.Mvc.Razor/Internal/DependencyContextRazorViewEngineOptionsSetup.cs
pranavkm: Breaking change in DependencyContext. It returns null now if an app has no defines (instead of an Empty sequence).

src/Microsoft.AspNetCore.Mvc.Core/Infrastructure/DefaultAssemblyProvider.cs
NTaylorMullen: Ah i c

NTaylorMullen:
:shipit:

Eilon:
Not gonna look at the code, but :shipit: anyway :smile:

khellang:
What's the story here? This will still ship as a separate library, right? I'm asking cause we're using this in Nancy as well :smile:

davidfowl:
> What's the story here? This will still ship as a separate library, right? I'm asking cause we're using this in Nancy as well :smile:

The library is here https://github.com/dotnet/cli/tree/rel/1.0.0/src/Microsoft.Extensions.DependencyModel. this is a workaround to unblock ourselves in the short term.

PR #4303, opened 2016-03-16
Use grunt-newer to make TagHelpers incrementally build.

pranavkm:
The newer task prevents grunt from doing work in the event the file hasn't changed. This would allow us to run `build` locally without Microsoft.AspNetCore.Mvc.TagHelpers from recompiling every time


NTaylorMullen:
:shipit:

Eilon:
:shipit:, but we should really stop using this entirely to process 10 lines of JavaScript code.

pranavkm:
Spoke with @Eilon offline. We decided we'll remove the two grunt tasks in MvcSandbox and BasicWebsite. We'll rename the package.json \ gruntfile.js in Mvc.TagHelper so build doesn't run it and add instructions on how to regenerate the file in case it ever gets modified.

PR #4304, opened 2016-03-16
Remove ModelBinding features from ResourceFilters

rynowak:
Removes all of the ModelBinding configuration support from Resource
Filters. We haven't come up with concrete scenarios around these features,
and don't want to paint ourselves into a corner with extensibility.


javiercn:
:shipit:

rynowak:
:shipit:

PR #4313, opened 2016-03-18
Moved AuthorizeFilter and CorsAuthorizationFilter to a public namespace

### LINKED ISSUE #4310: Move AuthorizeFilter and CORS filters into a user friendly namespace
Looks like the some of our built-in filters that are tied to external attributes live in an internal namespace:

https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/Internal/AuthorizeFilter.cs
https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Cors/Internal/CorsAuthorizationFilter.cs

Most of the time you don't interact with these filters directly *unless* you are configuring them as a global filter. These filters should live in a user friendly namespace.

ajaybhargavb:
Issue #4310 

@rynowak

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Cors/CorsAuthorizationFilter.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Cors/CorsAuthorizationFilter.cs
@@ -7,10 +7,11 @@ using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Cors.Infrastructure;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc.Cors.Internal;
 using Microsoft.AspNetCore.Mvc.Filters;
 using Microsoft.Extensions.Primitives;
 
-namespace Microsoft.AspNetCore.Mvc.Cors.Internal
+namespace Microsoft.AspNetCore.Mvc.Cors
 {
     /// <summary>
     /// A filter which applies the given <see cref="CorsPolicy"/> and adds appropriate response headers.
--- a/src/Microsoft.AspNetCore.Mvc.Abstractions/Authorization/IAllowAnonymousFilter.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Abstractions/Authorization/IAllowAnonymousFilter.cs
@@ -3,12 +3,9 @@
 
 using Microsoft.AspNetCore.Mvc.Filters;
 
-namespace Microsoft.AspNetCore.Mvc.Internal
+namespace Microsoft.AspNetCore.Mvc.Authorization
 {
-    /// <summary>
-    /// An implementation of <see cref="IAllowAnonymousFilter"/>
-    /// </summary>
-    public class AllowAnonymousFilter : IAllowAnonymousFilter
+    public interface IAllowAnonymousFilter : IFilterMetadata
     {
     }
 }
--- a/src/Microsoft.AspNetCore.Mvc.Abstractions/Filters/IAllowAnonymousFilter.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Abstractions/Filters/IAllowAnonymousFilter.cs
@@ -1,9 +1,0 @@
-// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-namespace Microsoft.AspNetCore.Mvc.Filters
-{
-    public interface IAllowAnonymousFilter : IFilterMetadata
-    {
-    }
-}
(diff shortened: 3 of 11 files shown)


src/Microsoft.AspNetCore.Mvc.Core/AuthorizeFilter.cs
```diff
@@ -10,7 +10,7 @@
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Internal;
 
-namespace Microsoft.AspNetCore.Mvc.Internal
+namespace Microsoft.AspNetCore.Mvc
```
Eilon: Is there a Filters namespace this can go to? Not sure we want to pollute the top-level namespace with this. Is there a place where most other filters live?

src/Microsoft.AspNetCore.Mvc.Cors/CorsAuthorizationFilter.cs
```diff
@@ -7,10 +7,11 @@
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Cors.Infrastructure;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc.Cors.Internal;
 using Microsoft.AspNetCore.Mvc.Filters;
 using Microsoft.Extensions.Primitives;
 
-namespace Microsoft.AspNetCore.Mvc.Cors.Internal
+namespace Microsoft.AspNetCore.Mvc.Cors
```
Eilon: This one is fine.

Eilon:
:watch:

PinpointTownes:
I've always thought that `AuthorizeFilter` was a weird name. Have you considered renaming it? (e.g `AuthorizationFilter`)

src/Microsoft.AspNetCore.Mvc.Core/AuthorizeFilter.cs
rynowak: `.Filters` is for filter interfaces/contracts/etc. Putting implementations there is a case of grouping by component type where we do grouping by feature in 99.9% of other cases.

> Is there a place where most other filters live?

Most live in `.Internal` because the attribute class can be used as the global filter.

The cors filters can go to `.Cors`

The best solution for authorization really seems to be carving out an `.Authorization` namespace and putting a few more things from internal in there.

@Eilon thoughts?

rynowak: We could also revisit having a 'filter policy' namespace and using that, but that's more naming by component type

rynowak:
@PinpointTownes 

`AuthorizationFilter` makes it sound like a base class for `IAuthorizationFilter` - it's named just so as it's the implementation that's used when you use `[Authorize(...)]`

rynowak:
@ajaybhargavb - the test files should be moved in the file system as well

PinpointTownes:
@rynowak true, but it's actually not tied to the authorize attribute, since you can instantiate it manually with a custom policy and register it in the global filters bag without having to decorate your actions/controllers with the authorize attribute. 

Actually, it sounds like a DefaultAuthorizationFilter for me, since it's what the app model layer uses by default when seeing an action decorated with an authorize attribute.

src/Microsoft.AspNetCore.Mvc.Core/AuthorizeFilter.cs
Eilon: I'm fine with a .Authorization namespace, though we need to be careful about what we move there. I want to slow down significantly on moving stuff, but if there's a good set of stuff that is likely more obscure anyway, let's rip the Band-Aid off and do it.

rynowak: The two other things that would move here right away are `IAllowAnonymous` and the filter equivalent of the same, so that gives us 3 things

Eilon: Yeah sure let's do that.

ajaybhargavb:
:up: :date:

test/Microsoft.AspNetCore.Mvc.Cors.Test/Internal/CorsAuthorizationFilterTest.cs
```diff
@@ -17,7 +17,7 @@
 using Moq;
 using Xunit;
 
-namespace Microsoft.AspNetCore.Mvc.Cors.Internal
+namespace Microsoft.AspNetCore.Mvc.Cors
```
rynowak: This file should be moved as well

rynowak:
:shipit:

PR #4316, opened 2016-03-19
[Fixes #3732] Introduce ControllerAttribute and use it to mark base classes as controllers

### LINKED ISSUE #3732: Revisit Controller discovery conventions
We have a complicated set of rules for controller discovery that involve looking at names, names of base-classes, which assembly defines a class/base-class, and which assemblies are referenced from an assembly.

We should figure out if/how we can streamline these rules. Some questions to discuss:
----------

**Should we require you to register assemblies that we use for discovery?**

**Should we require the ...Controller suffix in all cases**?

**Do we need opt-out/opt-in attributes**?

javiercn:
* Types deriving from types with their name ending in Controller are not considered controllers.
* Types with ControllerAttribute applied to them or to any type in their inheritance hierarchy are considered controllers.
* Types with ControllerAttribute and NonControllerAttribute in their inheritance hierarchy are not considered controllers.
* Public not nested, concrete, closed generic types with their name ending in Controller are considered controllers.


javiercn:
@rynowak @Eilon Here are the changes to how we discover controllers that I mentioned the other day. I feel all changes are fine and make sense.

src/Microsoft.AspNetCore.Mvc.Core/Controllers/DefaultControllerTypeProvider.cs
```diff
@@ -44,79 +44,48 @@ public virtual IEnumerable<TypeInfo> ControllerTypes
...
             {
                 return false;
             }
-            if (!typeInfo.Name.EndsWith(ControllerTypeName, StringComparison.OrdinalIgnoreCase) &&
-                !DerivesFromController(typeInfo, candidateAssemblies))
-            {
-                return false;
-            }
+
             if (typeInfo.IsDefined(typeof(NonControllerAttribute)))
```
pranavkm: Couldn't we remove this?

davidfowl:
Does this mean all POCO controllers require this attribute?

src/Microsoft.AspNetCore.Mvc.Core/ControllerAttribute.cs
```diff
@@ -0,0 +1,17 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Mvc
+{
+    /// <summary>
+    /// Indicates that the type and any derived types that this attribute is applied to
+    /// are considered a controller by the default controller discovery mechanism, unless
+    /// <see cref="NonControllerAttribute"/> is applied to any type in the hierarchy.
```
dougbu: Does this mean any type with both `[Controller]` and `[NonController]` applied to it or its base classes will _not_ be a controller? I'm wondering in particular about the case where a base class has `[NonController]` and the type itself (or any closer base class) has `[Controller]`. Users may expect the nearest attribute overrides the other(s).

One alternative would be to remove `[NonController]` and then add a `bool IsController` property to this type. Get the `[Controller]` (inherited or not) and its `IsController` value wins. That would also avoid the compiler allowing odd cases like `[Controller]` and `[NonController]` on the _same_ type.

javiercn: @dougbu I've explictly been very simplistic about this behavior. NonController always wins over Controller. It is easier behavior to understand for our users. With other workitems that I'm readying for PR, adding or removing controllers to the list of discovered controller types will be really easy, that's why I chose to keep our convention simple and that would be the recommended way for our customers to tweak the results of discovery.

src/Microsoft.AspNetCore.Mvc.Core/Controllers/DefaultControllerTypeProvider.cs
javiercn: I don't think so, you still want to have the ability to explictly opt out of being a controller.

javiercn:
@davidfowl No, POCO controllers will still be considered controllers as long as they are suffixed with controllers. The change is that we don't use the Controller suffix on a base class to mark something as a controller. In that case, you need to use [Controller] on your base class. A clear example of this is that our current controller base classes work in the same way. ControllerBase is denoted as a controller with the [Controller] attribute and anything inheriting from it will be a controller unless marked with the [NonController] attribute. So to summarize the changes:

```
public class A : ControllerBase{}
```
Previously wouldn't be considered a controller, now it will be (it extends ControllerBase which has the ControllerAttribute on it)

```
public class PocoController {}
```
Worked before and will keep working

```
public class B : PocoController { }
```

Would have been considered a controller before as long as PocoController assembly referenced MVC. It will require the Controller attribute moving forward, either on B itself or on PocoController class.

```
public class BController : PocoController { }
```
Worked before and will continue working no matter whether BController or PocoController have a Controller attribute on them as it's suffixed with Controller

If you have any doubts I highly encourage you to look at the test changes, they are pretty descriptive. I also think that getting rid of the notion that something is considered a controller based on the fact that your base class references the MVC assembly or not is a good thing

src/Microsoft.AspNetCore.Mvc.Core/ControllerAttribute.cs
dougbu: @javiercn you're being more simplistic about the behaviour than users will expect. "Closest attribute wins" is fairly common and "this one always wins" is rather rare.

dougbu:
@rynowak mentions "This is identical to what we're already doing for View Components" in #3732 and his list of suggested rules. So why introduce `[NonController]` here when there's no `[NonViewComponent]`? Put another way, why would users inherit from a controller except to create another controller? (Plus "... except if ..." rules are an added complication.)

javiercn:
@dougbu Can you give a sample in where we use "closest attribute wins"? I also would say that I consider that the scenario in which [Controller] and [NonController] attributes are applied through a class hierarchy would not be a common case.

For getting rid of [NonController] attribute, I'd say that we could do it, but I think there are good reasons to keep it around. Unlike ViewComponent, the Controller suffix is a way more common word in software  design terminology and patterns and the risk of having a collision is way higher (SqlTransactionController) or Results on Github for Controller vs ViewComponent
image
image

Also it's way more likely for a user to write a controller than a view component (you can write a completely functional MVC application without writing a single view component out of the box while the opposite is not true). Point being, you need to already opt-in to use view components, you can't opt-in to write controllers.

That said, 99% of time, people follow conventions, so I don't consider that this will cause trouble to any user, even those that don't follow conventions. I think both conventions are pretty reasonable, that keeping the logic as simple as it is right now is good enough because it satisfies all common user scenarios and that if the convention that we have doesn't suit your needs it's easy to either write your own convention or to tweak the result of discovery to include or exclude specific types.

dougbu:
> Can you give a sample in where we use "closest attribute wins"?

1. Anywhere an attribute has properties. We don't e.g. merge properties from multiple `[Display]` attributes, we use the `Name`, `Order`, ... (if any) from the closest.
2. `[BindNever]` and `[BindRequired]` inherit from a common base and the code looks for the closest of that base type.

src/Microsoft.AspNetCore.Mvc.Core/ControllerAttribute.cs
Muchiachio: As an end user, I do agree with @dougbu, but MVC doesn't follow this rule already (e.g. `AlllowAnonymous` attribute). It would be better to not introduce `NonController` attribute at all.

rynowak: Personally I don't see a lot of value in `[NonController]`. I might have been the one to suggest it :laughing: - but the new rules don't really require an opt-out IMO.

src/Microsoft.AspNetCore.Mvc.Core/ControllerAttribute.cs
```diff
@@ -0,0 +1,17 @@
...
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false, Inherited = true)]
+    public class ControllerAttribute : Attribute
+    {
+    }
```
rynowak: This should let you set the controller name.

src/Microsoft.AspNetCore.Mvc.Core/Controllers/DefaultControllerTypeProvider.cs
```diff
@@ -44,79 +44,48 @@ public virtual IEnumerable<TypeInfo> ControllerTypes
...
-                {
-                    return true;
-                }
-
-                typeInfo = baseTypeInfo;
+                return false;
             }
 
-            return false;
+            return true;
```
rynowak: Sooooooo much simpler. The reason for this change is that it's now possible to **explain** the rules.

1. Inherit from a class with `[Controller]` (includes our base-classes)
2. Use the suffix "Controller"

Those are the only rules

test/WebSites/BasicWebSite/Controllers/ApplicationBaseController.cs
```diff
@@ -1,10 +0,0 @@
-// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-namespace BasicWebSite.Controllers
-{
-    public abstract class ApplicationBaseController
```
rynowak: Could keep this by using `[Controller]`. I agree though, not a super useful functional test

src/Microsoft.AspNetCore.Mvc.Core/ControllerBase.cs
```diff
@@ -23,6 +23,7 @@ namespace Microsoft.AspNetCore.Mvc
     /// <summary>
     /// A base class for an MVC controller without view support.
     /// </summary>
+    [Controller]
```
rynowak: Do `ApiController` as well

Eilon:
Re: `NonController` - I recommend keeping it, and making it be an "always wins" attribute. Think of it as "guarantee this type and its descendants will never be used as a controller`. Plus, it simplifies things.

Muchiachio:
@Eilon do you have any good scenario where `NonController` attribute could be used? I can't image someone, inheriting controller for it not to be a controller (unless they have service classes named `ServiceController` which could be considered controllers by the framework :smile:).

src/Microsoft.AspNetCore.Mvc.Core/ControllerAttribute.cs
javiercn: I think that is a separate feature from what we are doing here, can we do it as a later as a separate change? I'm not sure of all the implications that setting the controller name through an attribute the is inherited might have and it's definitely something that we can tackle as a separate issue.

test/WebSites/BasicWebSite/Controllers/ApplicationBaseController.cs
javiercn: I did that at first, but I don't think that test is proving anything that isn't being tested in unit tests, so I think it can be safely removed.

src/Microsoft.AspNetCore.Mvc.Core/ControllerAttribute.cs
rynowak: Sure, please log an issue. It would probably get backlogged at this point  :laughing:

javiercn: Filed https://github.com/aspnet/Mvc/issues/4323

Eilon:
@Muchiachio it's for exactly the POCO scenario where a class happens to have a name that ends with Controller and its assembly references MVC, but isn't actually a controller. Certainly not common, but an important security and functional feature to be available when it's needed.

javiercn:
@rynowak Updated

src/Microsoft.AspNetCore.Mvc.Core/Controllers/DefaultControllerTypeProvider.cs
dougbu: > Those are the only rules

Apparently not:
3. Unless class also inherits from a class w/ `[NonController]`

test/Microsoft.AspNetCore.Mvc.Core.Test/Controllers/DefaultControllerTypeProviderTest.cs
```diff
@@ -230,18 +230,31 @@ public void IsController_NoControllerSuffix()
...
+            // Act
+            var isController = provider.IsController(typeof(NoSuffixNoControllerAttribute).GetTypeInfo());
+
+            // Assert
+            Assert.False(isController);
+        }
+
         [Theory]
         [InlineData(typeof(BaseNonControllerController))]
         [InlineData(typeof(BaseNonControllerControllerChild))]
```
dougbu: Need a few more tests involving `[NonController]` e.g. a class w/ `[Controller]` and suffix "Controller" that inherits from something w/ `[NonController]`.

dougbu:
:watch: just for https://github.com/aspnet/Mvc/pull/4316#discussion-diff-57072894

test/WebSites/ControllersFromServicesClassLibrary/ResourcesController.cs
```diff
@@ -5,6 +5,7 @@
 
 namespace ControllersFromServicesClassLibrary
 {
+    [Controller]
```
rynowak: Why is this needed now? Is this because this class library doesn't reference MVC?

javiercn: There is a controller without suffix that inherits from it ```public class Inventory : ResourcesController```

test/Microsoft.AspNetCore.Mvc.Core.Test/Controllers/DefaultControllerTypeProviderTest.cs
javiercn: Added a test for that scenario

javiercn:
updated @rynowak @dougbu

test/Microsoft.AspNetCore.Mvc.WebApiCompatShimTest/ApiControllerActionDiscoveryTest.cs
```diff
@@ -57,7 +57,7 @@ public void GetActions_ApiControllerWithoutControllerSuffix_IsNotController()
             var controllerType = typeof(TestControllers.Blog).GetTypeInfo();
             var actions = results.Where(ad => ad.ControllerTypeInfo == controllerType).ToArray();
 
-            Assert.Empty(actions);
+            Assert.Equal(2,actions.Count());
```
dougbu: Fix spacing and don't use Linq for a simple array. Probably best to remove the unnecessary `ToArray()` just above.

test/WebSites/ControllersFromServicesClassLibrary/ResourcesController.cs
dougbu: In the real world, should users add `[Controller]` on the base `ResourcesController` or on their `Inventory` controller? Realize both work but what would we recommend? Whichever that is, do it here.

dougbu:
:shipit: with a commit description that includes the three simple rules discussed [here](https://github.com/aspnet/Mvc/pull/4316#discussion-diff-56843632). I have a couple of small comments above too. And please get @rynowak's squirrel.

test/WebSites/ControllersFromServicesClassLibrary/ResourcesController.cs
rynowak: We'd recommend putting the attribute on your base class.

rynowak:
:shipit: - this will need an announcement post

PR #4342, opened 2016-03-24
[Fixes #4089] Introducing application parts and features

### LINKED ISSUE #4089: MVC Assets from DI + Multiproject MVC Design Notes

javiercn:
Introduced the main abstractions and wired them up through the stack


src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/ApplicationPartManager.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc.ApplicationParts;
+
+namespace Microsoft.AspNetCore.Mvc.ApplicationParts
+{
+    public class ApplicationPartManager
+    {
+        public IList<IApplicationFeatureProvider> Providers { get; } =
```
rynowak: Suggestion `FeatureProviders`

src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/ApplicationPartManager.cs
```diff
@@ -0,0 +1,30 @@
...
+            new List<IApplicationFeatureProvider>();
+
+        public IList<ApplicationPart> ApplicationParts { get; } =
+            new List<ApplicationPart>();
+
+        public void PopulateFeature<T>(T feature)
```
rynowak: I think this is a good name

src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/AssemblyPart.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Mvc.ApplicationParts
+{
+    public class AssemblyPart : ApplicationPart, IExportTypes
+    {
+        private readonly Assembly _assembly;
```
rynowak: We should expose the assembly as a public property

src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/IApplicationFeatureProviderOfT.cs
```diff
@@ -0,0 +1,14 @@
...
+
+using System.Collections;
+using System.Collections.Generic;
+using System.Reflection;
+
+namespace Microsoft.AspNetCore.Mvc.ApplicationParts
+{
+    public interface IApplicationFeatureProvider<T> : IApplicationFeatureProvider
+    {
+        void GetFeature(IEnumerable<ApplicationPart> parts, T feature);
```
rynowak: This should probably also be `PopulateFeature`

src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/IExportTypes.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Mvc.ApplicationParts
+{
+    public interface IExportTypes
```
rynowak: This should probably be something like `IApplicationPartTypeProvider`

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/IMvcBuilder.cs
```diff
@@ -12,5 +14,10 @@ public interface IMvcBuilder
         /// Gets the <see cref="IServiceCollection"/> where MVC services are configured.
         /// </summary>
         IServiceCollection Services { get; }
+
+        /// <summary>
+        /// Gets a list of <see cref="ApplicationPart"/> instances that compose the application.
+        /// </summary>
+        ApplicationPartManager Manager { get; }
```
rynowak: `PartsManager`

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/IMvcCoreBuilder.cs
```diff
@@ -12,5 +14,10 @@ public interface IMvcCoreBuilder
         /// Gets the <see cref="IServiceCollection"/> where essential MVC services are configured.
         /// </summary>
         IServiceCollection Services { get; }
+
+        /// <summary>
+        /// Gets a list of <see cref="ApplicationPart"/> instances that compose the application.
+        /// </summary>
+        ApplicationPartManager Manager { get; }
```
rynowak: `PartsManager`

src/Microsoft.AspNetCore.Mvc.Core/Internal/MvcCoreMvcOptionsSetup.cs
```diff
@@ -17,14 +17,21 @@ namespace Microsoft.AspNetCore.Mvc.Internal
     /// <summary>
     /// Sets up default options for <see cref="MvcOptions"/>.
     /// </summary>
-    public class MvcCoreMvcOptionsSetup : ConfigureOptions<MvcOptions>
+    public class MvcCoreMvcOptionsSetup : IConfigureOptions<MvcOptions>
```
rynowak: Good cleanup here

rynowak:
:watch: - design seems right, and simple too. Left some suggestions about naming.

/cc @lodejard @Eilon

src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/ApplicationPartManager.cs
Eilon: Just make sure you never use `T` please :smile:

Eilon: (To be clear, always use `TSomething`.)

src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/IExportTypes.cs
Eilon: Yeah this current name says nothing about what it's for. Try to convey the semantics.

src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/AssemblyPart.cs
```diff
@@ -0,0 +1,27 @@
...
+            {
+                throw new ArgumentNullException(nameof(assembly));
+            }
+
+            _assembly = assembly;
+        }
+
+        public override string Name => _assembly.GetName().Name;
+
+        public IEnumerable<TypeInfo> Types => _assembly.DefinedTypes;
```
Eilon: Is DefinedTypes the right API here? Traditionally the API to use is `Assembly.GetExportedTypes()` https://msdn.microsoft.com/en-us/library/system.reflection.assembly.getexportedtypes(v=vs.110).aspx.

src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/IExportTypes.cs
Eilon: And, perhaps more interestingly, why is this even needed?

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs
```diff
@@ -39,10 +42,49 @@ public static IMvcCoreBuilder AddMvcCore(this IServiceCollection services)
...
+                foreach (var assembly in assemblies)
+                {
+                    manager.ApplicationParts.Add(new AssemblyPart(assembly));
+                }
+            }
+
+            return manager;
+        }
+
+        private static T GetServiceFromCollection<T>(IServiceCollection services)
```
Eilon: What is this? This looks evil, no?

Eilon: This method also looks like it can fail in so many different ways. What if there's no `ImplementationInstance`?

src/Microsoft.AspNetCore.Mvc.Core/MvcOptions.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using Microsoft.AspNetCore.Mvc.ApplicationModels;
+using Microsoft.AspNetCore.Mvc.ApplicationParts;
```
Eilon: ?

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/IMvcBuilder.cs
```diff
@@ -12,5 +14,10 @@ public interface IMvcBuilder
         /// Gets the <see cref="IServiceCollection"/> where MVC services are configured.
         /// </summary>
         IServiceCollection Services { get; }
+
+        /// <summary>
+        /// Gets a list of <see cref="ApplicationPart"/> instances that compose the application.
```
Eilon: This comment isn't quite accurate. The manager does more than just have a list of parts.

test/Microsoft.AspNetCore.Mvc.Core.Test/ApplicationParts/ApplicationPartManagerTest.cs
```diff
@@ -0,0 +1,65 @@
...
+            manager.ApplicationParts.Add(new TestPart());
+            manager.Providers.Add(new FeatureProvider(f => f.Value++));
+            manager.Providers.Add(new FeatureProvider(f => f.Value = f.Value * 3));
+
+            var feature = new Feature();
+            // Act
+            manager.PopulateFeature(feature);
+
+            // Assert
+            Assert.Equal(18, feature.Value);
```
Eilon: This seems like some fun math, but hard to follow. Can this use some plain old assertions or something? Or record all the calls and compare against to a known list of calls?

test/Microsoft.AspNetCore.Mvc.Core.Test/ApplicationParts/ApplicationPartManagerTest.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            public override string Name => "Test";
+        }
+
+        private class OtherPartType: ApplicationPart
+        {
+            public override string Name => "Other";
+        }
+
+        private class Feature
```
Eilon: `TestFeature`

test/Microsoft.AspNetCore.Mvc.Core.Test/ApplicationParts/ApplicationPartManagerTest.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            public int Value { get; set; }
+        }
+
+        private class FeatureProvider : IApplicationFeatureProvider<Feature>
```
Eilon: `TestFeatureProvider`

test/Microsoft.AspNetCore.Mvc.IntegrationTests/TestMvcOptions.cs
```diff
@@ -12,6 +13,7 @@
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.ObjectPool;
 using Microsoft.Extensions.Options;
+using Moq;
```
Eilon: Not needed?

test/Microsoft.AspNetCore.Mvc.Test/MvcServiceCollectionExtensionsTest.cs
```diff
@@ -22,6 +23,7 @@
 using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.FileProviders;
```
Eilon: ?

Eilon:
Overall looking pretty good! :watch:

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs
davidfowl: Yeah this is pretty bad

src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/AssemblyPart.cs
javiercn: I chose DefinedTypes for several reasons:
* We were using different APIs on different parts of the codebase.
* DefinedTypes gives you TypeInfo instances which is more aligned with what we use in all our descriptor classes. (An the only thing that contains actual useful stuff, so we can save people from typing a lot of GetTypeInfo())
* All our components (in the future features) already contain logic to ensure that the type meets their specific needs (public, non open genetic, non static, etc)
* Using GetExportedTypes make some customization scenarios a bit harder (like using non public types as controllers) as users would need to write their own part to expose the private types and their feature to select the ones they want.

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs
javiercn: This is the least evil way of getting to the application services. 
* We already followed this pattern in other parts of the code (AddControllersAsServices for example). 
* Building a container during startup is expensive, so we want to avoid that.
* A change in the list of services that gets passed to ConfigureServices is arguably a breaking change (for sure a change on behavior)
* I can't foresee any reason why we would change those services from instances to something else, most of them are just data.
* Even if something changes and they are not available, we have options. We can fallback to building a temporary container or we can just initialize MVC without doing discovery (which is not ideal, but it would work) and it would be reasonably discoverable by looking at the list of applicationparts.

src/Microsoft.AspNetCore.Mvc.Core/MvcOptions.cs
javiercn: leftover from when this was on options

javiercn:
@rynowak @Eilon Updated

rynowak:
Is this still `[Design]`?

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs
Eilon: Reaching into the unconstructed container is bad, but perhaps unavoidable.

But reading the `ImplementationInstance` property seems awful. It wouldn't surprise me at all if we change that at some point, and no one would know.

src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/AssemblyPart.cs
Eilon: `GetExportedTypes` would only be in our default implementation. I think that requiring a more advanced scenario that needs nested types or non-public types (which would be super weird) to do some extra work is reasonable, no?

Why are we interested in TypeInfo instead of Type? Does it offer any advantages? How do we reason about choosing one or the other?

Ultimately we can project Type to TypeInfo, so is that a problem here?

src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/IApplicationFeatureProviderOfT.cs
```diff
@@ -0,0 +1,23 @@
...
+
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.Mvc.ApplicationParts
+{
+    /// <summary>
+    /// A provider for a given <typeparamref name="T"/> feature.
+    /// </summary>
+    /// <typeparam name="T">The type of the feature.</typeparam>
+    public interface IApplicationFeatureProvider<T> : IApplicationFeatureProvider
```
Eilon: Please remember to avoid using just `T`. Use something like `TFeature` instead. The only type of place where just `T` is used is when it has no particular meaning, which is almost never.

src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/IApplicationPartTypeProvider.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Microsoft.AspNetCore.Mvc.ApplicationParts
+{
+    /// <summary>
+    /// Exports a set of types from an <see cref="ApplicationPart"/>.
+    /// </summary>
+    public interface IApplicationPartTypeProvider
+    {
+        /// <summary>
+        /// Gets the list of exported types.
```
Eilon: The irony of using the word `exported` in the comment amuses me :smile: But, ultimately, these docs don't seem to convey the purpose of this interface (both the type doc and this method doc).

test/Microsoft.AspNetCore.Mvc.Core.Test/ApplicationParts/ApplicationPartManagerTest.cs
```diff
@@ -0,0 +1,79 @@
...
+            manager.Providers.Add(new TestFeatureProvider((f, v) => f.Values.Add($"P1{v}")));
+            manager.Providers.Add(new TestFeatureProvider((f, v) => f.Values.Add($"P2{v}")));
+
+            var feature = new TestFeature();
+            
+            // Act
+            manager.PopulateFeature(feature);
+
+            // Assert
+            Assert.Equal(new[] { "P1A", "P1C", "P2A", "P2C" }, feature.Values.ToArray());
```
Eilon: Much nicer, easier to follow!

Eilon: BTW even though it's much easier to follow, I did have to read it a couple of times to understand why P1B/P2B don't show up. Can anything be done to the names of the test types to indicate that `TestPart` is of "interest" and that `OtherPartType` is "not of interest"?

Eilon:
:watch:

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs
rynowak: Is there a suggested alternative? Is the `IHostingEnvironment` or whatever we need to read the .deps file available as a static?

Eilon: We might need to come up with some alternative here - this code worries me because it's diving into implementation details of a completely unrelated component.

src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/ApplicationPartManager.cs
```diff
@@ -0,0 +1,47 @@
...
+            new List<IApplicationFeatureProvider>();
+
+        /// <summary>
+        /// Gets the list of <see cref="ApplicationPart"/>s.
+        /// </summary>
+        public IList<ApplicationPart> ApplicationParts { get; } =
+            new List<ApplicationPart>();
+
+        /// <summary>
+        /// Computes the value of the given <paramref name="feature"/> using the list of
```
rynowak: I was say "Populates the given ..."

src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/ApplicationPartManager.cs
```diff
@@ -0,0 +1,47 @@
...
+        /// <see cref="IApplicationFeatureProvider{TFeature}"/>s configured on the
+        /// <see cref="ApplicationPartManager"/>.
+        /// </summary>
+        /// <typeparam name="TFeature">The type of the feature.</typeparam>
+        /// <param name="feature">The feature instance to configure.</param>
```
rynowak: configure -> populate

src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/IApplicationFeatureProviderOfT.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        /// <summary>
+        /// Updates the <paramref name="feature"/> intance.
+        /// </summary>
+        /// <param name="parts">The list of <see cref="ApplicationPart"/>s of the
+        /// application.
+        /// </param>
+        /// <param name="feature">The feature instance to populate</param>
```
rynowak: .

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs
rynowak: Isn't the `IHostingEnvironment` available somewhere as a static?

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs
```diff
@@ -39,10 +42,49 @@ public static IMvcCoreBuilder AddMvcCore(this IServiceCollection services)
...
+                if (environment == null)
+                {
+                    // Consider just returning an empty application part collection. Makes
+                    // tests easier and its not a condition we expect in production scenarios
+                    // where IHostingEnvironment will always be part of the ServiceCollection
+                    // we get passed in.
+                    throw new InvalidOperationException();
+                }
+
+                var assemblies = new DefaultAssemblyProvider(environment).CandidateAssemblies;
```
rynowak: Where else is the assembly provider used? would it be possible to get rid of?

rynowak: @pranavkm - is this going to impact the DNX scenarios?

test/Microsoft.AspNetCore.Mvc.Core.Test/ApplicationParts/ApplicationPartManagerTest.cs
```diff
@@ -0,0 +1,79 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc.ApplicationParts;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Mvc.Core.Test.ApplicationParts
+{
+    public class ApplicationPartManagerTest
+    {
+
```
rynowak: Extra line

test/Microsoft.AspNetCore.Mvc.Core.Test/ApplicationParts/ApplicationPartManagerTest.cs
```diff
@@ -0,0 +1,79 @@
+﻿using System;
```
rynowak: Needs docs header and some sort + remove love

test/Microsoft.AspNetCore.Mvc.Core.Test/ApplicationParts/ApplicationPartManagerTest.cs
```diff
@@ -0,0 +1,79 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc.ApplicationParts;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Mvc.Core.Test.ApplicationParts
```
rynowak: Same namespace as the product code

test/Microsoft.AspNetCore.Mvc.Core.Test/ApplicationParts/ApplicationPartManagerTest.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+            // Arrange
+            var manager = new ApplicationPartManager();
+            manager.ApplicationParts.Add(new TestPart("A"));
+            manager.ApplicationParts.Add(new OtherPartType("B"));
+            manager.ApplicationParts.Add(new TestPart("C"));
+            manager.Providers.Add(new TestFeatureProvider((f, v) => f.Values.Add($"P1{v}")));
+            manager.Providers.Add(new TestFeatureProvider((f, v) => f.Values.Add($"P2{v}")));
+
+            var feature = new TestFeature();
```
rynowak: Would like to see another test with a different feature type.

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs
rynowak: If we can read the .deps file another way then the other issue to solve here is updating the functional tests.

I'd be fine if our functional test startup classes just *manually* filtered the parts collection. What we're doing now is magic because there was no parts collection to configure :laughing: 

That gets rid of all of the need for this.

/cc @davidfowl

javiercn: @Eilon @rynowak. I discussed at the time this with @davidfowl and @lodejard. If we have concerns I can setup sometime on Monday to discuss an alternative.

src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/IApplicationFeatureProviderOfT.cs
javiercn: This one slipped through the cracks :)

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs
rynowak: I'd suggest here falling back to `IApplicationEnvironment.ApplicationName` which is available in a static - or just using that directly if it's a fit. 

If you think this needs more discussion then go ahead and schedule it. I'm not averse to checking this in with somewhat spooky code here to unblock the other parts and fixing it before Wednesday. /cc @Eilon

Eilon: Yeah it's fine for right now. We can even revisit in RTM. So just log a bug for cleaning that up and we'll deal with it later.

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs
javiercn: We can't get rid of this completely yet, ViewComponents still depend on it, need to asses the impact on DNX

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs
javiercn: https://github.com/aspnet/Mvc/issues/4353

javiercn:
Updated, this is no design anymore

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreMvcBuilderExtensions.cs
```diff
@@ -8,6 +8,7 @@
 using Microsoft.AspNetCore.Mvc.Formatters;
 using Microsoft.AspNetCore.Mvc.Internal;
 using System.Linq;
+using Microsoft.AspNetCore.Mvc.ApplicationParts;
```
Eilon: Sort.

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreMvcBuilderExtensions.cs
```diff
@@ -59,6 +60,57 @@ public static class MvcCoreMvcBuilderExtensions
...
+        public static IMvcBuilder ConfigureApplicationPartManager(
+            this IMvcBuilder builder,
+            Action<ApplicationPartManager> setupAction)
+        {
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+
```
Eilon: ` `

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreMvcBuilderExtensions.cs
Eilon: +other files

Eilon: Sort

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreMvcBuilderExtensions.cs
```diff
@@ -59,6 +60,57 @@ public static class MvcCoreMvcBuilderExtensions
...
+        /// <returns>The <see cref="IMvcBuilder"/>.</returns>
+        public static IMvcBuilder ConfigureApplicationPartManager(
+            this IMvcBuilder builder,
+            Action<ApplicationPartManager> setupAction)
+        {
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
```
Eilon: `\r\n` (plus other files)

src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/IApplicationPartTypeProvider.cs
```diff
@@ -0,0 +1,19 @@
...
+
+using System.Collections.Generic;
+using System.Reflection;
+
+namespace Microsoft.AspNetCore.Mvc.ApplicationParts
+{
+    /// <summary>
+    /// Gives access to a set of types from an <see cref="ApplicationPart"/>.
+    /// </summary>
+    public interface IApplicationPartTypeProvider
```
Eilon: I still don't get why this interface is needed. It's implemented in exactly one place, but never consumed.

javiercn: Sorry, you are actually right, this will be needed as part of https://github.com/aspnet/Mvc/pull/4344. I'll move it to that PR.

Eilon: Will #4344 be updated to be based off this PR? Hard to tell what's different in that PR. BTW if they are different branches I think you can send a PR that is just the diff from this PR (if you make it a sub-branch).

Eilon: Oh actually it looks like that's what you did... but for some reason the other PR has all the Parts stuff too...

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreMvcCoreBuilderExtensions.cs
```diff
@@ -11,6 +11,7 @@
 using Microsoft.AspNetCore.Mvc.Internal;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 using System.Linq;
+using Microsoft.AspNetCore.Mvc.ApplicationParts;
```
rynowak: Sore

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreMvcCoreBuilderExtensions.cs
```diff
@@ -144,6 +145,57 @@ internal static void AddAuthorizationServices(IServiceCollection services)
...
+        public static IMvcCoreBuilder ConfigureApplicationPartManager(
+            this IMvcCoreBuilder builder,
+            Action<ApplicationPartManager> setupAction)
+        {
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+
```
rynowak: extra line

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreMvcBuilderExtensions.cs
javiercn: This is fixed

test/Microsoft.AspNetCore.Mvc.Core.Test/ApplicationParts/ApplicationPartManagerTest.cs
```diff
@@ -0,0 +1,148 @@
...
+            var feature = new InterestFeature();
+
+            // Act
+            manager.PopulateFeature(feature);
+
+            // Assert
+            Assert.Empty(feature.Values.ToArray());
+        }
+
+        private class InterestPart : ApplicationPart
```
rynowak: These names are just crazy town banana-pants.

Eilon: Yes, it seems my earlier comment with regard to names was taken a bit too literally... perhaps still some room for improvementizationalism.

javiercn: I took it literally, I'll name one Controllers and the other ViewComponents and that should be descriptive enough. Sorry about that :(

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreMvcBuilderExtensions.cs
```diff
@@ -59,6 +60,56 @@ public static class MvcCoreMvcBuilderExtensions
         }
 
         /// <summary>
+        /// Adds an <see cref="ApplicationPart"/> to the list of <see cref="ApplicationPartManager.ApplicationParts"/> on the
+        /// <see cref="IMvcBuilder.PartManager"/>.
+        /// </summary>
+        /// <param name="builder">The <see cref="IMvcBuilder"/>.</param>
+        /// <param name="part">The <see cref="ApplicationPart"/>.</param>
+        /// <returns>The <see cref="IMvcBuilder"/>.</returns>
+        public static IMvcBuilder AddApplicationPart(this IMvcBuilder builder, ApplicationPart part)
```
rynowak: I suggest instead of this `AddAssemblyPart` - and call the other one `ConfigureParts`

rynowak:
:shipit:

Eilon:
Much nicer, muchacho. :shipit:

test/Microsoft.AspNetCore.Mvc.Core.Test/ApplicationParts/ApplicationPartManagerTest.cs
```diff
@@ -0,0 +1,149 @@
...
+            public string Value { get; }
+        }
+
+        private class ViewComponentsPart : ApplicationPart
+        {
+            public ViewComponentsPart(string value)
+            {
+                Value = value;
+            }
+            public override string Name => "Other";
```
pranavkm: Need a new line here.

javiercn:
Updated, rebased and squashed and added a proper commit message @rynowak

Eilon:
No further comments from me, so I'll give a :shipit:

rynowak:
:shipit: x2 from me

PR #4344, opened 2016-03-24
[Fixes #4014] Moved controller discovery to the application part feature model

### LINKED ISSUE #4014: Add overload to AddControllerAsServices that uses the default controller discovery logic
It should use the default controller discovery logic baked into these https://github.com/aspnet/Mvc/blob/47351aac7a124148f84a484e187abbe0f24b4efc/src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs#L81-L88 services.

javiercn:
Moved controller discovery to the application part feature model and modified AddControllersAsServices implementation to rely on ControllersFeature.


rynowak:
Taking a look at this, I'll do my best to comment on the parts that have the most to do with parts that this PR adds specifically.

src/Microsoft.AspNetCore.Mvc.Core/Controllers/ControllersFeatureProvider.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc.ApplicationParts;
+
+namespace Microsoft.AspNetCore.Mvc.Controllers
+{
+    public class ControllersFeatureProvider : IApplicationFeatureProvider<ControllersFeature>
+    {
+        private const string ControllerTypeName = "Controller";
```
rynowak: This isn't really the controller type name, it's the suffix.

src/Microsoft.AspNetCore.Mvc.Core/Controllers/ControllersFeatureProvider.cs
```diff
@@ -0,0 +1,68 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc.ApplicationParts;
+
+namespace Microsoft.AspNetCore.Mvc.Controllers
+{
+    public class ControllersFeatureProvider : IApplicationFeatureProvider<ControllersFeature>
```
rynowak: Let's make the naming singular -> `ControllerFeature` and so on.

src/Microsoft.AspNetCore.Mvc.Core/Controllers/ControllersFeatureProvider.cs
```diff
@@ -0,0 +1,68 @@
...
+                {
+                    if (IsController(type) && !feature.Controllers.Contains(type))
+                    {
+                        feature.Controllers.Add(type);
+                    }
+                }
+            }
+        }
+
+        private bool IsController(TypeInfo typeInfo)
```
rynowak: There's really no harm in making this public if it's more testable this way.

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreMvcCoreBuilderExtensions.cs
```diff
@@ -11,6 +11,7 @@
 using Microsoft.AspNetCore.Mvc.Internal;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 using System.Linq;
+using Microsoft.AspNetCore.Mvc.Controllers;
```
rynowak: sort

test/Microsoft.AspNetCore.Mvc.Core.Test/Controllers/ControllersFeatureProviderTest.cs
```diff
@@ -0,0 +1,541 @@
...
+            var feature = new ControllersFeature();
+
+            // Act
+            manager.PopulateFeature(feature);
+
+            // Assert
+            Assert.Empty(feature.Controllers);
+        }
+
+
```
rynowak: Extra line

test/Microsoft.AspNetCore.Mvc.Core.Test/Controllers/ControllersFeatureProviderTest.cs
```diff
@@ -0,0 +1,541 @@
...
+            }
+
+            public override string Name => "Test";
+
+            public IEnumerable<TypeInfo> Types { get; }
+        }
+    }
+}
+
+// These controllers are used to test the DefaultControllerTypeProvider implementation
```
rynowak: comment a bit stale

test/Microsoft.AspNetCore.Mvc.Core.Test/Controllers/ControllersFeatureProviderTest.cs
```diff
@@ -0,0 +1,541 @@
...
+// which REQUIRES that they be public top-level classes. To avoid having to stub out the
+// implementation of this class to test it, they are just top level classes. Don't reuse
+// these outside this test - find a better way or use nested classes to keep the tests
+// independent.
+namespace Microsoft.AspNetCore.Mvc.DefaultControllerTypeProviderControllers
```
rynowak: namespace is the stale too

rynowak:
:watch: - changes look good overall, will wait for a rebase so that I don't have to review 3 PRs at once :laughing:

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreMvcCoreBuilderExtensions.cs
```diff
@@ -97,51 +97,23 @@ internal static void AddAuthorizationServices(IServiceCollection services)
...
+            foreach (var controller in feature.Controllers.Select(c => c.AsType()))
             {
-                throw new ArgumentNullException(nameof(builder));
+                builder.Services.TryAddTransient(controller, controller);
             }
 
-            ControllersAsServices.AddControllersAsServices(builder.Services, controllerTypes);
-            return builder;
-        }
+            builder.Services.Replace(ServiceDescriptor.Transient<IControllerActivator, ServiceBasedControllerActivator>());
```
javiercn: Is there a reason not to make this a singleton?

javiercn:
Updated, removed design label

src/Microsoft.AspNetCore.Mvc.Core/Controllers/ControllerFeatureProvider.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// <inheritdoc />
+        public void PopulateFeature(
+            IEnumerable<ApplicationPart> parts,
+            ControllerFeature feature)
+        {
+            foreach (var part in parts.OfType<IApplicationPartTypeProvider>())
+            {
+                foreach (var type in part.Types)
+                {
+                    if (IsController(type) && !feature.Controllers.Contains(type))
```
Eilon: Any perf concerns here? n<sup>2</sup> algorithm?

Eilon: (Especially considering the list of controllers will basically *never* already contain the controller, meaning the scan for `Contains` will always be its worst case and take `n` time to run.)

test/Microsoft.AspNetCore.Mvc.Core.Test/ApplicationParts/AssemblyPartTest.cs
```diff
@@ -22,6 +22,20 @@ public void AssemblyPart_Name_ReturnsAssemblyName()
...
+        {
+            // Arrange
+            var assembly = typeof(AssemblyPartTest).GetTypeInfo().Assembly;
+            var part = new AssemblyPart(assembly);
+
+            // Act
+            var types = part.Types;
+
+            // Assert
+            Assert.Equal(assembly.DefinedTypes, types);
```
Eilon: BTW I'd argue that this test is testing the specific implementation rather than the behavior. It would be safer to assert that the collections are equal, rather than checking that the collection *instance* is the same.

test/Microsoft.AspNetCore.Mvc.Core.Test/Controllers/ControllerFeatureProviderTest.cs
```diff
@@ -0,0 +1,546 @@
...
+
+    public class OpenGenericController<T> : Controller
+    {
+    }
+
+    public class DerivedGenericController : OpenGenericController<string>
+    {
+    }
+
+    public interface IController
```
Eilon: I see this is just cut & paste, but this confused me to no end at first. If we rename this to `ISomeController` would it still have the same value as a test? (From the looks of it yes.)

Eilon:
:watch:

src/Microsoft.AspNetCore.Mvc.Core/Controllers/ControllerFeatureProvider.cs
javiercn: The guard is there (an I believe it's necessary) in case you have multiple IApplicationFeatureProviders or you invoke Populate feature on the part manager with an already populated feature. I want to avoid duplicates as that will cause trouble at a later point in the pipeline.

I cared to the extent of putting IsController before in the clause to avoid unnecessary lookups, but this code is going to be called once during startup and we don't have data to suggest that it would be relevant in terms of time, so I'll take some measurements to make sure. What is the worst n we would consider in this scenario?

test/Microsoft.AspNetCore.Mvc.Core.Test/ApplicationParts/AssemblyPartTest.cs
javiercn: I believe that's what Equal in xunit does. Maybe I can just new up a collection with DefinedTypes as source to be sure? The purpose of this test is to make sure that something breaks if someone decides to change it later for GetExportedTypes or something like that, to ensure that such change doesn't go unnoticed

Eilon: Ah, yeah, not sure what xUnit does - you're probably right. But indeed perhaps comparing against a known good list of types is best here.

src/Microsoft.AspNetCore.Mvc.Core/Controllers/ControllerFeatureProvider.cs
Eilon: Completely agree that the check is necessary, in some manner. I am concerned about a not-unreasonable case where I have an app with 100 controllers in an assembly - this will do ~10,000 checks for `Contains`, right? Would be good to either measure, or come up with an algorithm that intrinsically has minimal cost (not sure exactly what that would be).

And, even though it's a one-time startup cost, we need to keep startup costs as low as possible.

src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/IApplicationPartTypeProvider.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Microsoft.AspNetCore.Mvc.ApplicationParts
+{
+    /// <summary>
+    /// Exposes a set of types from an <see cref="ApplicationPart"/>.
+    /// </summary>
+    public interface IApplicationPartTypeProvider
+    {
+        /// <summary>
+        /// Gets the list of available types.
```
rynowak: I was say `Gets the list of types defined by the part.` That's a little clearer

src/Microsoft.AspNetCore.Mvc.Core/Controllers/ControllerFeature.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Reflection;
+
+namespace Microsoft.AspNetCore.Mvc.Controllers
+{
+    /// <summary>
+    /// The list of controllers types in an MVC application.
```
rynowak: This should probably reference the part manager or application parts in some way

src/Microsoft.AspNetCore.Mvc.Core/Controllers/ControllerFeatureProvider.cs
rynowak: Isn't this all speculative perf-doing? (spoiler: speculative hand-waving about perf will *not* be part of my 10 top tips)





*There's also a simple solution here that resolves the perf concern and is probably more correct than what the code is doing.*

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreMvcCoreBuilderExtensions.cs
rynowak: There's a small tradeoff, the service provide has to hold on to it. We generally only make 'top-level' services into singletons. In this case neither the activator nor the factory are 'top-level'.

test/Microsoft.AspNetCore.Mvc.Core.Test/ApplicationParts/AssemblyPartTest.cs
rynowak: You have to look at the intellisense to be sure what xUnit does. It really **wants** to compare these as collections, but it's not always clear because there's `Assert.Equal(T t1, T t2)` and `Assert.Equal(IEnumerable<T> ts1, IEnumerable<T> ts2)`. You can always write `Assert.Equal<TypeInfo>(assembly.DefinedTypes, types);` to prove that it's doing what you want, but VS will want to 'simplify' this.

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ControllerActionDescriptorProviderTests.cs
```diff
@@ -1483,17 +1485,49 @@ public void GetDescriptors_SplitsConstraintsBasedOnControllerRoute_Override()
...
+            {
+                Types = types;
+            }
+
+            public override string Name => "Test";
+
+            public IEnumerable<TypeInfo> Types { get; }
+        }
+
+        private class TestProvider : IApplicationFeatureProvider<ControllerFeature>
```
rynowak: Should we just make this public?

test/Microsoft.AspNetCore.Mvc.WebApiCompatShimTest/ApiControllerActionDiscoveryTest.cs
```diff
@@ -406,20 +405,49 @@ private void Invoke(ControllerActionDescriptorProvider provider, ActionDescripto
...
+            {
+                foreach (var type in parts.OfType<IApplicationPartTypeProvider>().SelectMany(t => t.Types))
+                {
+                    feature.Controllers.Add(type);
+                }
             }
+        }
 
-            public override IEnumerable<TypeInfo> ControllerTypes
+        private class NamespaceFilteredControllersFeatureProvider : IApplicationFeatureProvider<ControllerFeature>
```
rynowak: Keep this in your brain as an example we can point people to if they want to write a custom feature.

rynowak:
:shipit: from me - get signoff from @Eilon on remaining feedback

src/Microsoft.AspNetCore.Mvc.Core/Controllers/ControllerFeatureProvider.cs
Eilon: n<sup>2</sup> algorithms generally aren't just a speculative concern - it is often a non-optimal *algorithm*, as opposed to just "choosing the wrong API". But, @javiercn and I chatted over email and it seems for even moderate sized sites (e.g. ~500 controllers) the cost is extremely low.

Eilon:
Aside from the few tiny things discussed, I'm :shipit: too.

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ControllerActionDescriptorProviderTests.cs
javiercn: TestPart and TestProvider i've been placing them adhoc where I needed them for tests, where would this go?

test/Microsoft.AspNetCore.Mvc.Core.Test/ApplicationParts/AssemblyPartTest.cs
javiercn: I added an Assert.NotSame(assembly.DefinedTypes, types) to validate that it's not doing reference equality, just to make the test a bit clearer (other options like comparing against a known list of types or passing a generic parameter seemed more cumbersome and more obscure respectively)

src/Microsoft.AspNetCore.Mvc.Core/Controllers/ControllerFeatureProvider.cs
javiercn: Not doing anything here

PR #4348, opened 2016-03-25
Added Tag helper factory to align tag helper creation with controllers and view components

### LINKED ISSUE #4087: No equivalent of AddControllersAsServices() for VC's and TagHelpers
If you have an external assembly, contianing a Controller, and a ViewComponent, you can register the controller with MVC by calling `AddControllersAsServices()`

However there is no equivalent for the ViewComponent.
The default behaviour seems to be if you have a project reference to the assembly then the VC is picked up, but if you don't have a project reference to it (i.e `ILibraryManager` is not aware of it) - because for example, it's a standalone "plugin" assembly, then VC's are not detected. 

Could you add a method similar to `AddControllersAsServices()` but for VC's?

Or perhaps consider adding some sort of "IncludeMvcServicesFromAssemblies()` method, that will make sure the assemblies are treated as candidates for all types of services that MVC looks for - i.e Controllers, VC's, and whatever else.

javiercn:
* Introduced a TagHelperFactory abstraction that does activation of the properties on a tag helper instance and that relies on an ITagHelperActivator to create the actual tag helper.
* Massaged the interfaces around to align them in naming and shape with the abstractions for View components and Controllers.
* Tag helper creation abstractions rely on generics as it makes getting and running tag helper initializers for a given tag helper much easier.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor/ITagHelperFactory.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor/ITagHelperFactory.cs
@@ -7,16 +7,15 @@ using Microsoft.AspNetCore.Razor.TagHelpers;
 namespace Microsoft.AspNetCore.Mvc.Razor
 {
     /// <summary>
-    /// Provides methods to create a tag helper.
+    /// Provides methods to create and initialize tag helpers.
     /// </summary>
-    public interface ITagHelperActivator
+    public interface ITagHelperFactory
     {
         /// <summary>
-        /// Creates an <see cref="ITagHelper"/>.
+        /// Creates a new tag helper for the specified <paramref name="context"/>.
         /// </summary>
-        /// <typeparam name="TTagHelper">The <see cref="ITagHelper"/> type.</typeparam>
-        /// <param name="context">The <see cref="ViewContext"/> for the executing view.</param>
+        /// <param name="context"><see cref="ViewContext"/> for the executing view.</param>
         /// <returns>The tag helper.</returns>
-        TTagHelper Create<TTagHelper>(ViewContext context) where TTagHelper : ITagHelper;
+        TTagHelper CreateTagHelper<TTagHelper>(ViewContext context) where TTagHelper : ITagHelper;
     }
 }
--- a/src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultTagHelperActivator.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultTagHelperActivator.cs
@@ -2,79 +2,45 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Concurrent;
-using System.Collections.Generic;
-using System.Reflection;
+using Microsoft.AspNetCore.Mvc.Internal;
 using Microsoft.AspNetCore.Mvc.Rendering;
-using Microsoft.AspNetCore.Mvc.ViewFeatures;
 using Microsoft.AspNetCore.Razor.TagHelpers;
-using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Internal;
 
 namespace Microsoft.AspNetCore.Mvc.Razor.Internal
 {
-    /// <inheritdoc />
+    /// <summary>
+    /// Default implementation of <see cref="ITagHelperActivator"/>.
+    /// </summary>
     public class DefaultTagHelperActivator : ITagHelperActivator
     {
-        private readonly ConcurrentDictionary<Type, PropertyActivator<ViewContext>[]> _injectActions;
-        private readonly Func<Type, PropertyActivator<ViewContext>[]> _getPropertiesToActivate;
+        private readonly ITypeActivatorCache _typeActivatorCache;
 
         /// <summary>
         /// Instantiates a new <see cref="DefaultTagHelperActivator"/> instance.
         /// </summary>
-        public DefaultTagHelperActivator()
+        /// <param name="typeActivatorCache">The <see cref="ITypeActivatorCache"/>.</param>
+        public DefaultTagHelperActivator(ITypeActivatorCache typeActivatorCache)
         {
-            _injectActions = new ConcurrentDictionary<Type, PropertyActivator<ViewContext>[]>();
-            _getPropertiesToActivate = type =>
-                PropertyActivator<ViewContext>.GetPropertiesToActivate(
-                    type,
-                    typeof(ViewContextAttribute),
-                    CreateActivateInfo);
+            if (typeActivatorCache == null)
+            {
+                throw new ArgumentNullException(nameof(typeActivatorCache));
+            }
+
+            _typeActivatorCache = typeActivatorCache;
         }
 
         /// <inheritdoc />
-        public void Activate<TTagHelper>(TTagHelper tagHelper, ViewContext context)
+        public TTagHelper Create<TTagHelper>(ViewContext context)
             where TTagHelper : ITagHelper
         {
-            if (tagHelper == null)
-            {
-                throw new ArgumentNullException(nameof(tagHelper));
-            }
-
             if (context == null)
             {
                 throw new ArgumentNullException(nameof(context));
             }
 
-            var propertiesToActivate = _injectActions.GetOrAdd(
-                tagHelper.GetType(),
-                _getPropertiesToActivate);
-
-            for (var i = 0; i < propertiesToActivate.Length; i++)
-            {
-                var activateInfo = propertiesToActivate[i];
-                activateInfo.Activate(tagHelper, context);
-            }
-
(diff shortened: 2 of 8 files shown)


src/Microsoft.AspNetCore.Mvc.Razor/ITagHelperActivator.cs
```diff
@@ -7,16 +7,22 @@
...
         /// <param name="context">The <see cref="ViewContext"/> for the executing view.</param>
-        void Activate<TTagHelper>(TTagHelper tagHelper, ViewContext context) where TTagHelper : ITagHelper;
+        TTagHelper Create<TTagHelper>(ViewContext context) where TTagHelper : ITagHelper;
+
+        /// <summary>
+        /// Releases a tag helper.
+        /// </summary>
+        /// <param name="context">The <see cref="ViewContext"/> for the executing view.</param>
+        /// <param name="tagHelper">The tag helper to release.</param>
+        void Release(ViewContext context, object tagHelper);
```
NTaylorMullen: Talked offline.

src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultTagHelperFactory.cs
```diff
@@ -0,0 +1,107 @@
...
+                throw new ArgumentNullException(nameof(activator));
+            }
+
+            _activator = activator;
+            _injectActions = new ConcurrentDictionary<Type, PropertyActivator<ViewContext>[]>();
+            _getPropertiesToActivate = type =>
+                PropertyActivator<ViewContext>.GetPropertiesToActivate(
+                    type,
+                    typeof(ViewContextAttribute),
+                    CreateActivateInfo);
```
NTaylorMullen: hoist

src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultTagHelperFactory.cs
```diff
@@ -0,0 +1,107 @@
...
+
+            return tagHelper;
+        }
+
+        private static void InitializeTagHelper<TTagHelper>(TTagHelper tagHelper, ViewContext context)
+            where TTagHelper : ITagHelper
+        {
+            // Run any tag helper initializers in the container
+            var serviceProvider = context.HttpContext.RequestServices;
+            var initializers = (IEnumerable<ITagHelperInitializer<TTagHelper>>)
```
NTaylorMullen: Why not `var initializers = serviceProvider.GetService<IEnumerable<ITagHelperInitializer<TTagHelper>>>();`?

NTaylorMullen:
:watch: for removing the release mechanics.

src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultTagHelperFactory.cs
javiercn: We don't have a dependency on DependencyInjection in this package and that extension method lives there.

dougbu: It's not brought in transitively? Surprising if so.

javiercn:
Updated

src/Microsoft.AspNetCore.Mvc.Razor/ITagHelperFactory.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc.Rendering;
+using Microsoft.AspNetCore.Razor.TagHelpers;
+
+namespace Microsoft.AspNetCore.Mvc.Razor
+{
+    /// <summary>
+    /// Provides methods for creation and disposal of tag helpers.
```
NTaylorMullen: no need for disposal anymore.

NTaylorMullen:
design looks good.

src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultTagHelperActivator.cs
```diff
@@ -2,79 +2,45 @@
...
             {
-                initializer.Initialize(tagHelper, context);
+                throw new ArgumentNullException(nameof(context));
             }
-        }
 
-        private static PropertyActivator<ViewContext> CreateActivateInfo(PropertyInfo property)
-        {
-            return new PropertyActivator<ViewContext>(property, viewContext => viewContext);
+            return _typeActivatorCache.CreateInstance<TTagHelper>(
```
javiercn: I'm wondering if we should put a guard similar to the one we have for controllers and view components, like in https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/Controllers/DefaultControllerActivator.cs#L49. This one already is constraining to TTagHelper extending ITagHelper, but I think that's not enought. @NTaylorMullen @rynowak

javiercn:
I don't think this needs more tests unless we add a guard on DefaultTagHelperActivator, we just spit responsibilities but the work is the same, so the tests for the factory should work for the activator too. I'll add some tests if you think otherwise

src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultTagHelperFactory.cs
javiercn: It was brought in transitively, intellisense was :trollface: me when I ctrl+. it

src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultTagHelperActivator.cs
NTaylorMullen: I think this is guarded enough. if it's `ITagHelper` then its possible to create it/run it.

src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultTagHelperFactory.cs
```diff
@@ -0,0 +1,92 @@
...
+{
+    /// <summary>
+    /// Default implementation for <see cref="ITagHelperFactory"/>.
+    /// </summary>
+    public class DefaultTagHelperFactory : ITagHelperFactory
+    {
+        private readonly ITagHelperActivator _activator;
+        private readonly ConcurrentDictionary<Type, PropertyActivator<ViewContext>[]> _injectActions;
+        private readonly Func<Type, PropertyActivator<ViewContext>[]> _getPropertiesToActivate;
+        private readonly Func<PropertyInfo, PropertyActivator<ViewContext>> _createActivateInfo = CreateActivateInfo;
```
NTaylorMullen: static

NTaylorMullen:
Add a unit test validating the `DefaultTagHelperActivator` creates `TagHelper`s by providing their ctor arguments via DI.

:shipit:

src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultTagHelperActivator.cs
rynowak: Referring to the code in the controller activator - that's currently on my list to review as a perf hotspot. 

I think it's fine to ask for forgiveness rather than permission here, and I wish we had done the same in the controller activator.

src/Microsoft.AspNetCore.Mvc.Razor/RazorPage.cs
```diff
@@ -192,13 +177,7 @@ private IViewBufferScope BufferScope
...
         public TTagHelper CreateTagHelper<TTagHelper>() where TTagHelper : ITagHelper
         {
-            var tagHelper = TypeActivatorCache.CreateInstance<TTagHelper>(
-                ViewContext.HttpContext.RequestServices,
-                typeof(TTagHelper));
-
-            TagHelperActivator.Activate(tagHelper, ViewContext);
-
-            return tagHelper;
+            return TagHelperFactory.CreateTagHelper<TTagHelper>(ViewContext);
```
rynowak: This is such an improvement over what was here before.

test/Microsoft.AspNetCore.Mvc.Razor.Test/Internal/DefaultTagHelperFactoryTest.cs
```diff
@@ -19,12 +19,12 @@
 
 namespace Microsoft.AspNetCore.Mvc.Razor.Internal
 {
-    public class DefaultTagHelperActivatorTest
+    public class DefaultTagHelperFactoryTest
```
rynowak: There should be 1-2 simple unit tests for the activator

rynowak:
:watch:

lukyanov84:
👍👍👍

javiercn:
updated

rynowak:
:shipit:

PR #4352, opened 2016-03-25
[Perf] Rewrite ExpressionHelper.GetExpressionText using StringBuilder

mnltejaswini:
Part 2 Fix for #3921 

ExpressionHelper.GetExpressionText() computes the string representation for the expression using Stack and String Concat operations. This pull request has changes to rewrite it using StringBuilder to optimize string and stack allocations. Also includes removing the Linq extension method "Any" in IsSingleArgumentIndexer to optimize delegate and closure allocations.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionHelper.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionHelper.cs
@@ -2,11 +2,11 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
+using System.Text;
 
 namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
 {
@@ -38,9 +38,10 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
             }
 
             var containsIndexers = false;
-            // Split apart the expression string for property/field accessors to create its name
-            var nameParts = new Stack<string>();
             var part = expression.Body;
+
+            // Builder to concatenate the names for property/field accessors within an expression to create a string.
+            var builder = new StringBuilder();
 
             while (part != null)
             {
@@ -54,10 +55,10 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
                         break;
                     }
 
-                    nameParts.Push(
-                        GetIndexerInvocation(
-                            methodExpression.Arguments.Single(),
-                            expression.Parameters.ToArray()));
+                    InsertIndexerInvocationText(
+                        builder,
+                        methodExpression.Arguments.Single(),
+                        expression);
 
                     part = methodExpression.Object;
                 }
@@ -66,10 +67,10 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
                     containsIndexers = true;
                     var binaryExpression = (BinaryExpression)part;
 
-                    nameParts.Push(
-                        GetIndexerInvocation(
-                            binaryExpression.Right,
-                            expression.Parameters.ToArray()));
+                    InsertIndexerInvocationText(
+                        builder,
+                        binaryExpression.Right,
+                        expression);
 
                     part = binaryExpression.Left;
                 }
@@ -87,19 +88,9 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
                         break;
                     }
 
-                    nameParts.Push("." + name);
+                    builder.Insert(0, name);
+                    builder.Insert(0, '.');
                     part = memberExpressionPart.Expression;
-                }
-                else if (part.NodeType == ExpressionType.Parameter)
-                {
-                    // When the expression is parameter based (m => m.Something...), we'll push an empty
-                    // string onto the stack and stop evaluating. The extra empty string makes sure that
-                    // we don't accidentally cut off too much of m => m.Model.
-                    nameParts.Push(string.Empty);
-
-                    // Exit loop. Have the entire name because the parameter cannot be used as an indexer; always the
-                    // leftmost expression node.
-                    break;
                 }
                 else
                 {
@@ -108,16 +99,23 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
             }
 
             // If parts start with "model", then strip that part away.
-            if (nameParts.Count > 0 && string.Equals(nameParts.Peek(), ".model", StringComparison.OrdinalIgnoreCase))
+            if (part == null || part.NodeType != ExpressionType.Parameter)
             {
-                nameParts.Pop();
+                var text = builder.ToString();
+                if (text.StartsWith(".model", StringComparison.OrdinalIgnoreCase))
+                {
+                    // 6 is the length of the string ".model".
+                    builder.Remove(0, 6);
+                }
             }
 
-            expressionText = string.Empty;
-            if (nameParts.Count > 0)
+            if (builder.Length > 0)
             {
-                expressionText = nameParts.Aggregate((left, right) => left + right).TrimStart('.');
+                // Trim the leading "." if present.
+                builder.Replace(".", string.Empty, 0, 1);
             }
+
+            expressionText = builder.ToString();
 
             if (expressionTextCache != null && !containsIndexers)
             {
@@ -127,21 +125,34 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
             return expressionText;
         }
 
-        private static string GetIndexerInvocation(
-            Expression expression,
-            ParameterExpression[] parameters)
+        private static void InsertIndexerInvocationText(
+            StringBuilder builder,
+            Expression indexExpression,
+            LambdaExpression parentExpression)
         {
-            if (expression == null)
+            if (builder == null)
             {
-                throw new ArgumentNullException(nameof(expression));
+                throw new ArgumentNullException(nameof(builder));
             }
 
-            if (parameters == null)
+            if (indexExpression == null)
             {
-                throw new ArgumentNullException(nameof(parameters));
+                throw new ArgumentNullException(nameof(indexExpression));
             }
 
-            var converted = Expression.Convert(expression, typeof(object));
+            if (parentExpression == null)
+            {
+                throw new ArgumentNullException(nameof(parentExpression));
+            }
+
+            if (parentExpression.Parameters == null)
+            {
+                throw new ArgumentException(Resources.FormatPropertyOfTypeCannotBeNull(
+                    nameof(parentExpression.Parameters),
+                    nameof(parentExpression)));
+            }
+
+            var converted = Expression.Convert(indexExpression, typeof(object));
             var fakeParameter = Expression.Parameter(typeof(object), null);
             var lambda = Expression.Lambda<Func<object, object>>(converted, fakeParameter);
             Func<object, object> func;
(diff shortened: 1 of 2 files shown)


mnltejaswini:
Benchmark Test app : https://github.com/aspnet/Performance/tree/dev/testapp/BigViews
loadtest -k --rps 10 -n 100 http://localhost:5000/Home/IndexUsingIndexers
loadtest -k --rps 10 -n 100 http://localhost:5000/Home/Index

Before Optimization
--------------------------
image

image


After Optimization
-----------------------
image

~0.5% savings in total bytes and 1% savings in number of objects allocated

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionHelper.cs
```diff
@@ -38,85 +38,97 @@ public static string GetExpressionText(LambdaExpression expression, ExpressionTe
...
+                    var text = builder.ToString();
+                    if (text.StartsWith(".model", StringComparison.OrdinalIgnoreCase))
+                    {
+                        // 6 is the length of the string ".model".
+                        builder.Remove(0, 6);
+                    }
                 }
-                else
+
+                if (builder.Length > 0)
```
dougbu: Remove this check. It's not useful.

dougbu: Sorry, reread the docs. That's a shame.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionHelper.cs
```diff
@@ -38,85 +38,97 @@ public static string GetExpressionText(LambdaExpression expression, ExpressionTe
...
+                        }
+
+                        builder.Insert(0, name);
+                        builder.Insert(0, '.');
+                        part = memberExpressionPart.Expression;
+                    }
+                    else if (part.NodeType == ExpressionType.Parameter)
+                    {
+                        // Exit loop. Have the entire name because the parameter cannot be used as an indexer; always the
+                        // leftmost expression node.
```
dougbu: Can't parse this comment. Also why does this block exists -- does the same thing as the regular `else` block?

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionHelper.cs
```diff
@@ -178,9 +190,18 @@ public static bool IsSingleArgumentIndexer(Expression expression)
             }
 
             // Find default property (the indexer) and confirm its getter is the method in this expression.
-            return declaringType.GetRuntimeProperties().Any(
-                property => (string.Equals(defaultMember.MemberName, property.Name, StringComparison.Ordinal) &&
-                    property.GetMethod == methodExpression.Method));
+            var runtimeProperties = declaringType.GetRuntimeProperties();
+            for (int i=0; i< runtimeProperties.Count(); i++ )
```
dougbu: Space expression as VS would and get rid of the repeated Linq eval:
``` c#
var runtimeProperties = declaringType.GetRuntimeProperties().ToArray();
for (int i = 0; i < runtimeProperties.Length; i++)
{
    var property = runtimeProperties[i];
    ...
}
```

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionHelper.cs
mnltejaswini: Yes :)

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionHelper.cs
```diff
@@ -38,85 +38,97 @@ public static string GetExpressionText(LambdaExpression expression, ExpressionTe
...
-            // Split apart the expression string for property/field accessors to create its name
-            var nameParts = new Stack<string>();
             var part = expression.Body;
+            expressionText = string.Empty;
 
-            while (part != null)
+            if (part != null &&
+                (part.NodeType == ExpressionType.Call ||
+                part.NodeType == ExpressionType.ArrayIndex ||
+                part.NodeType == ExpressionType.MemberAccess))
```
dougbu: What is the value of this condition?

dougbu: As discussed offline, even lazy-initializing the `StringBuilder` isn't likely that useful because this code executes only for a cache miss.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionHelper.cs
```diff
@@ -38,85 +38,97 @@ public static string GetExpressionText(LambdaExpression expression, ExpressionTe
...
             {
-                if (part.NodeType == ExpressionType.Call)
+                // Split apart the expression string for property/field accessors to create its name
+                var builder = new StringBuilder();
+                var parameters = expression.Parameters.ToArray();
```
dougbu: Only create this array when necessary. Probably best to do this in `GetIndexerInvocation()`.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionHelper.cs
```diff
@@ -38,85 +38,97 @@ public static string GetExpressionText(LambdaExpression expression, ExpressionTe
...
+                        containsIndexers = true;
+                        var methodExpression = (MethodCallExpression)part;
+                        if (!IsSingleArgumentIndexer(methodExpression))
+                        {
+                            // Unsupported.
+                            break;
+                        }
+                        var indexerInvokationResult = GetIndexerInvocation(
+                                methodExpression.Arguments.Single(),
+                                parameters);
```
dougbu: Remove extra indentation here and in `GetIndexerInvocation()` call just below.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionHelper.cs
```diff
@@ -38,85 +38,97 @@ public static string GetExpressionText(LambdaExpression expression, ExpressionTe
...
-
-                    // Exit loop. Have the entire name because the parameter cannot be used as an indexer; always the
-                    // leftmost expression node.
-                    break;
+                    var text = builder.ToString();
+                    if (text.StartsWith(".model", StringComparison.OrdinalIgnoreCase))
+                    {
+                        // 6 is the length of the string ".model".
+                        builder.Remove(0, 6);
+                    }
```
dougbu: `builder.Replace(".model", string,Empty, 0, 6);` to replace above block, avoiding the `ToString()`.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionHelper.cs
```diff
@@ -38,85 +38,97 @@ public static string GetExpressionText(LambdaExpression expression, ExpressionTe
...
+                        if (!IsSingleArgumentIndexer(methodExpression))
+                        {
+                            // Unsupported.
+                            break;
+                        }
+                        var indexerInvokationResult = GetIndexerInvocation(
+                                methodExpression.Arguments.Single(),
+                                parameters);
+
+                        builder.Insert(0, indexerInvokationResult);
```
dougbu: Pass the `builder` into `GetIndexerInvocation()`, avoiding the extra intermediate string.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Internal/ExpressionHelperTest.cs
```diff
@@ -12,6 +12,79 @@ public class ExpressionHelperTest
...
+        public static IEnumerable<object[]> ExpressionAndTexts
+        {
+            get
+            {
+                var i = 3;
+                var value = "Test";
+                var Model = new TestModel();
+                var key = "TestModel";
+                var myModels = new List<TestModel>();
+                var __DisplayClass = new List<TestModel>();
```
dougbu: Avoid writing code that violates the c# rules. Expressions such as `model => myModels[i].SelectedCategory` should be complex enough to hit compiler-generated field names. Set a breakpoint on relevant line and confirm instead of including this test.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Internal/ExpressionHelperTest.cs
```diff
@@ -163,6 +236,17 @@ public static IEnumerable<object[]> NonEquivalentExpressions
         }
 
         [Theory]
+        [MemberData(nameof(ExpressionAndTexts))]
+        public void GetExpressionText_ReturnsValidExpressionText(LambdaExpression expression, string expressionText)
```
dougbu: Minor but what is "valid"? Suggest `...ReturnsExpectedText`.

dougbu:
:watch:

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionHelper.cs
mnltejaswini: I thought of it but, StringBuilder Replace does case sensitive comparison of the original string. The original code does StringComparison.OrdinalIgnoreCase comparison.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionHelper.cs
mnltejaswini: Here the trade off is allocation versus computational cost. The ToArray() adds an allocation whereas the original one has more instructions. In my opinion, I would prefer the later one because in most of the cases, we break early as we find the matching property,

mnltejaswini:
:up::date:

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionHelper.cs
```diff
@@ -38,9 +38,9 @@ public static string GetExpressionText(LambdaExpression expression, ExpressionTe
             }
 
             var containsIndexers = false;
-            // Split apart the expression string for property/field accessors to create its name
-            var nameParts = new Stack<string>();
             var part = expression.Body;
+            // Split apart the expression string for property/field accessors to create its name
```
dougbu: Blank line b/4 comment.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionHelper.cs
dougbu: - What is rewriting the `Any()` to use _different_ Linq buying us? This doesn't look any better than the previous code.
- Still need to space the expressions correctly: `(int i = 0; i < runtimeProperties.Length; i++)`.

dougbu: Do a `foreach` here.

dougbu:
:watch:

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionHelper.cs
mnltejaswini: foreach allocates enumerator. However this is better than 2 allocations that was made by previous Any()

mnltejaswini:
:up::date:

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionHelper.cs
dougbu: Again, just use a `foreach`. The `Count()` and `ElementAt()` calls iterate multiple times when this loop only needs to go through the elements once.

dougbu: Put another way: You're talking about micro-optimizations that make the code harder to read and much harder to reason about. If something significant shows up later, we can revisit. For now write the obvious code.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionHelper.cs
dougbu: Comment was not addressed.

dougbu:
:shipit:

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionHelper.cs
```diff
@@ -108,17 +99,24 @@ public static string GetExpressionText(LambdaExpression expression, ExpressionTe
             }
 
             // If parts start with "model", then strip that part away.
-            if (nameParts.Count > 0 && string.Equals(nameParts.Peek(), ".model", StringComparison.OrdinalIgnoreCase))
+            if (part == null || part.NodeType != ExpressionType.Parameter)
             {
-                nameParts.Pop();
+                var text = builder.ToString();
```
dougbu: @mnltejaswini I think you said you had an idea about avoiding this extra `string` allocation. Whether I'm remembering things correctly or not, let's get this change in. Deal with this later (if it can be dealt with).

PR #4355, opened 2016-03-25
Refactor ModelBinding

rynowak:
Splits the responsibilities of an `IModelBinder` into `IModelBinderProvider` and `IModelBinder`.

Introduces `IModelBinderFactory`.

Some Cleanup of `ModelBinderContext`.


rynowak:
Some allocation data using **BigModelBinding**


**Before**
image

The highlighted `Type[]` is totally gone accounting for most of the difference in allocations.


**After**
image

rynowak:
**Some sampling data using BigModelBinding**

Before: my machine can sustain about ~1500rps
After: ~1750 rps

**Before**
image


**After**
image

Model binding has gone from 13.95% to 9.95% with this refactor and without applying any more specific optimizations.

Note that looking at the sampling data, GC is ranges from 60-64% of the total. The task and string allocations here are far and away the biggest issue.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
```diff
@@ -0,0 +1,229 @@
...
+
+            // OK this isn't a recursive case (yet) so "push" an entry on the stack and then ask the providers
+            // to create the binder.
+            stack.Add(new KeyValuePair<Key, DelegatingBinder>(key, null));
+
+            IModelBinder result = null;
+
+            for (var i = 0; i < _providers.Length; i++)
+            {
+                var factory = _providers[i];
```
davidfowl: nit: variable name provider

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/Metadata/ModelMetadataIdentity.cs
```diff
@@ -98,5 +99,31 @@ public ModelMetadataKind MetadataKind
...
+            return other.HasValue && Equals(other.Value);
+        }
+
+        /// <inheritdoc />
+        public override int GetHashCode()
+        {
+            var hash = new HashCodeCombiner();
+            hash.Add(ContainerType);
+            hash.Add(ModelType);
+            hash.Add(Name);
```
pranavkm: `, StringComparer.Ordinal);`

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/IModelBinderProvider.cs
```diff
@@ -0,0 +1,19 @@
...
+    /// instances in <c>MvcOptions</c>.
+    /// </summary>
+    public interface IModelBinderProvider
+    {
+        /// <summary>
+        /// Creates a <see cref="IModelBinder"/> based on <see cref="ModelBinderProviderContext"/>.
+        /// </summary>
+        /// <param name="context">The <see cref="ModelBinderProviderContext"/>.</param>
+        /// <returns>An <see cref="IModelBinder"/>.</returns>
+        IModelBinder Create(ModelBinderProviderContext context);
```
pranavkm: The API is already broken due to the change in parameter, but `GetBinder` in case you want to use the same name as Mvc5's binder provider.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/ComplexTypeModelBinder.cs
```diff
@@ -2,17 +2,36 @@
...
+        /// The <see cref="IDictionary{TKey, TValue}"/> of binders to use for binding properties.
+        /// </param>
+        public ComplexTypeModelBinder(IDictionary<ModelMetadata, IModelBinder> propertyBinders)
+        {
+            if (propertyBinders == null)
+            {
+                throw new ArgumentNullException(nameof(propertyBinders));
+            }
+
+            _propertyBinders = new Dictionary<ModelMetadata, IModelBinder>(propertyBinders);
```
pranavkm: Do we need to clone it? Seems unlikely that this would be mutated in the common case where it's created by the `ComplexTypeModelBinderProvider`

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBinderProviderContext.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Mvc.ModelBinding
+{
+    /// <summary>
+    /// A context object for <see cref="IModelBinderProvider.Create(ModelBinderProviderContext)"/>.
```
dougbu: FYI there's no need to specify the parameter types unless you're dealing w/ a method group. Leaving off the parenthetical part reduces churn if we add parameters later.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBinderProviderContext.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// </summary>
+    public abstract class ModelBinderProviderContext
+    {
+        /// <summary>
+        /// Creates an <see cref="IModelBinder"/> for the given <paramref name="metadata"/>.
+        /// </summary>
+        /// <param name="metadata">The <see cref="ModelMetadata"/> for the model.</param>
+        /// <returns>An <see cref="IModelBinder"/>.</returns>
+        public abstract IModelBinder CreateBinder(ModelMetadata metadata);
```
dougbu: Strange to surface a method on a context object. Could the `IModelBinderFactory` and `ModelBinderFactoryContext` be in Abstractions and included in this context?

src/Microsoft.AspNetCore.Mvc.Core/ControllerBase.cs
```diff
@@ -142,6 +143,31 @@ public IModelMetadataProvider MetadataProvider
         }
 
         /// <summary>
+        /// Gets or sets the <see cref="IModelBinderFactory"/>.
+        /// </summary>
+        public IModelBinderFactory ModelBinderFactory
```
dougbu: Why is this property `public` and why does it have a setter? Seems to enable some new scenario but I'm missing it. Note this does _not_ support the same scenarios as changing the `IModelBinders` in `Controller.ModelBinders` would do w/ existing code -- at least not in a straightforward fashion.

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultModelBindingContext.cs
```diff
@@ -104,18 +108,25 @@ public DefaultModelBindingContext()
                 throw new ArgumentNullException(nameof(modelName));
             }
 
+            var oldBindingSource = BindingSource;
```
dougbu: Remove unused variable. Or set it to `modelMetadata.BindingSource` and use it to shorten things in the next block.

src/Microsoft.AspNetCore.Mvc.Core/Internal/DelegatingBinder.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Mvc.ModelBinding.Internal
+{
+    public class DelegatingBinder : IModelBinder
```
dougbu: Doesn't need `///` comments but merits _some_ explanation w/in the class. Might mention that it initially serves as a placeholder but will be resolved prior to binder execution. And, when executing, delegates to the `Inner` binder.

Separately `DelegatingBinder` is a slightly odd name because, while delegation is what it does later on, "placeholdering" is why it exists and a more important role. Suggest `BinderPlaceholder`.

src/Microsoft.AspNetCore.Mvc.Core/Internal/DelegatingBinder.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        public bool IsInUse { get; set; }
```
dougbu: Remove unused property.

src/Microsoft.AspNetCore.Mvc.Core/Internal/MvcCoreMvcOptionsSetup.cs
```diff
@@ -37,17 +36,20 @@ public static void ConfigureMvc(MvcOptions options, IHttpRequestStreamReaderFact
...
+            options.ModelBinderProviders.Add(new SimpleTypeModelBinderProvider());
+            options.ModelBinderProviders.Add(new CancellationTokenModelBinderProvider());
+            options.ModelBinderProviders.Add(new ByteArrayModelBinderProvider());
+            options.ModelBinderProviders.Add(new FormFileModelBinderProvider());
+            options.ModelBinderProviders.Add(new FormCollectionModelBinderProvider());
+            options.ModelBinderProviders.Add(new KeyValuePairModelBinderProvider());
+            options.ModelBinderProviders.Add(new DictionaryModelBinderProvider());
+            options.ModelBinderProviders.Add(new ArrayModelBinderProvider());
+            options.ModelBinderProviders.Add(new CollectionModelBinderProvider());
+            options.ModelBinderProviders.Add(new ComplexTypeModelBinderProvider());
```
dougbu: Now I know the real reason for this PR. You just wanted to kill the name `MutableObjectModelBinder` to death. :trollface:

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/ArrayModelBinderProvider.cs
```diff
@@ -0,0 +1,35 @@
...
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            // We don't support binding readonly properties of arrays because we can't resize the
+            // existing value.
+            if (context.Metadata.ModelType.IsArray && !context.Metadata.IsReadOnly)
+            {
+                var elementType = context.Metadata.ModelType.GetElementType();
```
dougbu: Use `context.Metadata.ElementMetadata.ModelType` instead.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/BinderTypeModelBinder.cs
```diff
@@ -0,0 +1,63 @@
...
+        public BinderTypeModelBinder(Type binderType)
+        {
+            if (binderType == null)
+            {
+                throw new ArgumentNullException(nameof(binderType));
+            }
+
+            if (!typeof(IModelBinder).GetTypeInfo().IsAssignableFrom(binderType.GetTypeInfo()))
+            {
+                throw new InvalidOperationException(
```
dougbu: At this level, should be an `ArgumentException`. But why doesn't the `BinderTypeModelBinderProvider` do this check? (`InvalidOperationException` would make sense there.)

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/BinderTypeModelBinder.cs
```diff
@@ -0,0 +1,63 @@
...
+                throw new ArgumentNullException(nameof(bindingContext));
+            }
+
+            var requestServices = bindingContext.OperationBindingContext.HttpContext.RequestServices;
+            var binder = (IModelBinder)_factory(requestServices, arguments: null);
+
+            await binder.BindModelAsync(bindingContext);
+
+            // A model binder was specified by metadata and this binder handles all such cases.
+            // Always tell the model binding system to skip other model binders i.e. return non-null.
```
dougbu: This comment is no longer applicable.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/BinderTypeModelBinderProvider.cs
```diff
@@ -0,0 +1,30 @@
...
+        public IModelBinder Create(ModelBinderProviderContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            if (context.BindingInfo?.BinderType != null)
+            {
+                return new BinderTypeModelBinder(context.BindingInfo.BinderType);
```
dougbu: Why not create the real binder here? That is, why does `BinderTypeModelBinder` exist in the new world?

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/ByteArrayModelBinder.cs
```diff
@@ -5,10 +5,10 @@
...
 using Microsoft.AspNetCore.Mvc.Internal;
 
-namespace Microsoft.AspNetCore.Mvc.ModelBinding
+namespace Microsoft.AspNetCore.Mvc.ModelBinding.Binders
 {
     /// <summary>
-    /// ModelBinder to bind Byte Arrays.
+    /// ModelBinder to bind byte Arrays.
     /// </summary>
     public class ByteArrayModelBinder : IModelBinder
```
dougbu: Weird case here. Previously would fall through to using an `ArrayModelBinder<byte>` which supports additional submitted formats. Probably need to instantiate this with that binder as a fallback -- and use it.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/CollectionModelBinderProvider.cs
```diff
@@ -0,0 +1,62 @@
...
+                return (IModelBinder)Activator.CreateInstance(binderType, elementBinder);
+            }
+
+            // If the model type is IEnumerable<> then we need to know if we can assign a List<> to it, since
+            // that's what we would create.
+            var enumerableType = ClosedGenericMatcher.ExtractGenericInterface(modelType, typeof(IEnumerable<>));
+            if (enumerableType != null)
+            {
+                var listType = typeof(List<>).MakeGenericType(enumerableType.GenericTypeArguments);
+                if (modelType.GetTypeInfo().IsAssignableFrom(listType.GetTypeInfo()))
```
dougbu: Check whether the property is readonly here. Anything but `ICollection<>` is useful `iff` we can actually do the assignment.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/ComplexTypeModelBinder.cs
```diff
@@ -135,7 +148,13 @@ protected virtual bool CanBindProperty(ModelBindingContext bindingContext, Model
         /// </returns>
         protected virtual Task BindProperty(ModelBindingContext bindingContext)
         {
-            return bindingContext.OperationBindingContext.ModelBinder.BindModelAsync(bindingContext);
+            IModelBinder binder;
+            if (_propertyBinders.TryGetValue(bindingContext.ModelMetadata, out binder))
```
dougbu: The provider won't leave any property out. So why isn't it an error for `TryGetValue()` to fail?

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
```diff
@@ -0,0 +1,229 @@
...
+            if (binder == null)
+            {
+                var message = Resources.FormatCouldNotCreateIModelBinder(providerContext.Metadata.ModelType);
+                throw new InvalidOperationException(message);
+            }
+
+            if (context.CacheToken != null)
+            {
+                _cache.TryAdd(context.CacheToken, binder);
+            }
```
dougbu: - Needs a comment about only adding binders for top-level models to the cache.
- Shouldn't we cache the computed binder for `TryUpdateModel[Async]()` scenarios?

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
```diff
@@ -0,0 +1,229 @@
...
+            {
+                _factory = factory;
+                Metadata = factoryContext.Metadata;
+                BindingInfo = factoryContext.BindingInfo;
+
+                MetadataProvider = _factory._metadataProvider;
+                Stack = new List<KeyValuePair<Key, DelegatingBinder>>();
+            }
+
+            public DefaultModelBinderProviderContext(
```
dougbu: Perhaps `private` to make it more clear where this constructor overload is used.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
```diff
@@ -0,0 +1,229 @@
...
+
+            public Key(ModelMetadata metadata, object token)
+            {
+                _metadata = metadata;
+                _token = token;
+            }
+
+            public bool Equals(Key other)
+            {
+                return _metadata.Equals(other._metadata) && object.Equals(_token, other._token);
```
dougbu: Perhaps `ReferenceEquals()` for the tokens? Put another way, does `ParameterInfo` implement a useful `Equals()`?

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBindingHelper.cs
```diff
@@ -20,9 +20,9 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
     public static class ModelBindingHelper
     {
         /// <summary>
-        /// Updates the specified <paramref name="model"/> instance using the specified <paramref name="modelBinder"/>
-        /// and the specified <paramref name="valueProvider"/> and executes validation using the specified
-        /// <paramref name="validatorProvider"/>.
+        /// Updates the specified <paramref name="model"/> instance using the specified
+        /// <paramref name="modelBinderFactory"/> and the specified <paramref name="valueProvider"/> and executes 
+        /// validation using the specified <paramref name="validatorProvider"/>.
```
dougbu: If we meant for this class to be `public` and in a prominent namespace, this is a significant API change and does not support the previous scenarios in a straightforward fashion.

src/Microsoft.AspNetCore.Mvc.WebApiCompatShim/HttpRequestMessage/HttpRequestMessageModelBinderProvider.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// </summary>
+    public class HttpRequestMessageModelBinderProvider : IModelBinderProvider
+    {
+        /// <inheritdoc />
+        public IModelBinder Create(ModelBinderProviderContext context)
+        {
+            if (context.Metadata.ModelType == typeof(HttpRequestMessage))
+            {
+                return new HttpRequestMessageModelBinder();
+            }
```
dougbu: Why does similar code in `HttpRequestMessageModelBinder` remain?

test/Microsoft.AspNetCore.Mvc.Core.Test/ControllerBaseTest.cs
```diff
@@ -1635,10 +1635,14 @@ private static ControllerBase GetController(IModelBinder binder, IValueProvider
             var metadataProvider = TestModelMetadataProvider.CreateDefaultProvider();
             var httpContext = new DefaultHttpContext();
 
+            var binderFactory = new Mock<IModelBinderFactory>();
+            binderFactory
+                .Setup(f => f.CreateBinder(It.IsAny<ModelBinderFactoryContext>()))
+                .Returns(binder);
```
dougbu: Move down closer to `var controller` use.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBinderProviderContext.cs
rynowak: This isn't the same as calling `IModelBinderFactory.CreateBinder` - it has a different signature on purpose. Don't see any reason to change this.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBinderProviderContext.cs
rynowak: Sure, I'll do a pass and fix any other places I find like this in new code.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/IModelBinderProvider.cs
rynowak: Sure

src/Microsoft.AspNetCore.Mvc.Core/ControllerBase.cs
rynowak: Every service dependency on `Controller` works like this.

> Note this does not support the same scenarios as changing the IModelBinders in Controller.ModelBinders would do w/ existing code -- at least not in a straightforward fashion.

Not intended to be supported. 

This property was always 'unit testing only' in legacy code.

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultModelBindingContext.cs
rynowak: Removed

src/Microsoft.AspNetCore.Mvc.Core/Internal/DelegatingBinder.cs
rynowak: Sure

src/Microsoft.AspNetCore.Mvc.Core/Internal/DelegatingBinder.cs
rynowak: :+1:

src/Microsoft.AspNetCore.Mvc.Core/Internal/MvcCoreMvcOptionsSetup.cs
rynowak: I really did, it wasn't the best name.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/ArrayModelBinderProvider.cs
rynowak: :+1:

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/CancellationTokenModelBinderProvider.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        /// <inheritdoc />
+        public IModelBinder Create(ModelBinderProviderContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            if (context.Metadata.ModelType == typeof(CancellationToken))
```
pranavkm: Not going to suggest we do anything in this PR, but do we want to try and consolidate binder providers that match on ModelType (`if modelType == typeof(X) return y();`)? I imagine it could be mildy useful for user scenarios too where they want to provide binding for a specific type.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/BinderTypeModelBinder.cs
rynowak: They both do. I didn't see the harm in that. Thoughts?

Will change to argument exception.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/BinderTypeModelBinder.cs
rynowak: :dancer:

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/BinderTypeModelBinderProvider.cs
rynowak: `BinderTypeModelBinder` allows you to type-activate a binder pre-request. I'll open an issue and we can decide if we want to keep it.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
```diff
@@ -0,0 +1,229 @@
...
+                        return binder;
+                    }
+                    else
+                    {
+                        return entry.Value;
+                    }
+                }
+            }
+
+
```
pranavkm: Extra line

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/ByteArrayModelBinder.cs
rynowak: Where do you see that? I only see it returning `Failed` and `Success`

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
```diff
@@ -0,0 +1,229 @@
...
+            }
+
+            if (result == null && stack.Count > 1)
+            {
+                // Use a no-op binder if we're below the top level. At the top level, we throw.
+                result = NoOpBinder.Instance;
+            }
+            
+            // "pop"
+            var delegatingBinder = stack[stack.Count - 1].Value;
```
pranavkm: `Debug.Assert(stack.Count > 0);` since we're assuming it's non-empty.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/CollectionModelBinderProvider.cs
rynowak: OK

src/Microsoft.AspNetCore.Mvc.Core/Resources.resx
```diff
@@ -340,4 +340,7 @@
   <data name="Formatter_NoMediaTypes" xml:space="preserve">
     <value>No media types found in '{0}.{1}'. Add at least one media type to the list of supported media types.</value>
   </data>
+  <data name="CouldNotCreateIModelBinder" xml:space="preserve">
+    <value>Could not create a model binder a model object of type '{0}'.</value>
```
pranavkm: This doesn't sound right.

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/Binders/ArrayModelBinderProviderTest.cs
```diff
@@ -0,0 +1,105 @@
...
+            });
+
+            // Act
+            var result = provider.Create(context);
+
+            // Assert
+            Assert.IsType(typeof(ArrayModelBinder<>).MakeGenericType(modelType.GetElementType()), result);
+        }
+
+        public static TheoryData<int[]> ArrayModelData
```
pranavkm: Doesn't look like this is used

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/Binders/ArrayModelBinderProviderTest.cs
```diff
@@ -0,0 +1,105 @@
...
+                // Expect to be called with the element type to create a binder for elements.
+                Assert.Equal(typeof(int), m.ModelType);
+                return Mock.Of<IModelBinder>();
+            });
+
+            // Act
+            var result = provider.Create(context);
+
+            // Assert
+            Assert.IsType(typeof(ArrayModelBinder<>).MakeGenericType(typeof(int)), result);
```
pranavkm: Does `Assert.IsType<ArrayModelBinder<int>>(result);` not work?

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/Binders/ByteArrayModelBinderProviderTest.cs
```diff
@@ -0,0 +1,48 @@
...
+using Xunit;
+
+namespace Microsoft.AspNetCore.Mvc.ModelBinding.Binders
+{
+    public class ByteArrayModelBinderProviderTest
+    {
+        [Theory]
+        [InlineData(typeof(object))]
+        [InlineData(typeof(TestClass))]
+        [InlineData(typeof(IList<int>))]
```
pranavkm: Does it work when it's `IList<byte>`?

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/CollectionModelBinderProvider.cs
rynowak: Added check, comment, and test.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/ComplexTypeModelBinder.cs
rynowak: yeah, this should fail hard.

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/Binders/ComplexTypeModelBinderTest.cs
```diff
@@ -420,8 +418,9 @@ public void CanBindProperty_GetSetProperty(string property)
                 },
             };
 
+            var binder = CreateBinder(bindingContext.ModelMetadata);
```
pranavkm: Empty line after

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/ComplexTypeModelBinder.cs
rynowak: The only reason is to get the field typed as `Dictionary` and avoid enumerator alloc. Open to suggestions.

rynowak: But I guess we don't iterate it. hmm.

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/Binders/DictionaryModelBinderProviderTest.cs
```diff
@@ -0,0 +1,70 @@
...
+            {
+                if (m.ModelType == typeof(KeyValuePair<string, int>) ||
+                    m.ModelType == typeof(int) ||
+                    m.ModelType == typeof(string))
+                {
+                    return Mock.Of<IModelBinder>();
+                }
+                else
+                {
+                    Assert.False(false, "Not the right model type");
```
pranavkm: `Assert.True(false...`

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
rynowak: > Shouldn't we cache the computed binder for TryUpdateModel[Async]() scenarios?

It would be nice to do that - sure. Maybe using the metadata as the token?

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
rynowak: Sure

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
rynowak: Yeah you're right, we should be explicit about reference equals here.

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/Binders/KeyValuePairModelBinderProviderTest.cs
```diff
@@ -0,0 +1,66 @@
...
+            var context = new TestModelBinderProviderContext(typeof(KeyValuePair<string, int>));
+            context.OnCreatingBinder(m =>
+            {
+                if (m.ModelType == typeof(string) || m.ModelType == typeof(int))
+                {
+                    return Mock.Of<IModelBinder>();
+                }
+                else
+                {
+                    Assert.False(false, "Not the right model type");
```
pranavkm: Assert.True

pranavkm:
:shipit:

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBindingHelper.cs
rynowak: Spoke to @Eilon - he doesn't care.

src/Microsoft.AspNetCore.Mvc.WebApiCompatShim/HttpRequestMessage/HttpRequestMessageModelBinderProvider.cs
rynowak: Missed it, will take it out.

test/Microsoft.AspNetCore.Mvc.Core.Test/ControllerBaseTest.cs
rynowak: :ok:

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
dougbu: :+1:

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
rynowak: Fixed

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/ComplexTypeModelBinder.cs
dougbu: Could make the `propertyBinders` parameter a `Dictionary` if there's anything important in how it's used.

rynowak: removed the copy

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/Metadata/ModelMetadataIdentity.cs
rynowak: fixed

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelBinderProviderContext.cs
rynowak: :see_no_evil:

src/Microsoft.AspNetCore.Mvc.Core/ControllerBase.cs
rynowak: :see_no_evil:

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/BinderTypeModelBinderProvider.cs
rynowak: :see_no_evil:

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/ByteArrayModelBinder.cs
rynowak: :see_no_evil:

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBindingHelper.cs
rynowak: :see_no_evil:

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/ByteArrayModelBinder.cs
dougbu: Yeah, you're right. Every `byte[]` is handled here so e.g. `?byte[0]=1&byte[1]=2&...` wouldn't have worked before. Does make me wonder what the MVC 5 / Web API behaviours were...

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
rynowak: Fixed

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
rynowak: Sure

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/BinderTypeModelBinderProvider.cs
dougbu: thx

src/Microsoft.AspNetCore.Mvc.Core/Resources.resx
rynowak: Fixed

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/Binders/ArrayModelBinderProviderTest.cs
rynowak: Removed

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/Binders/ArrayModelBinderProviderTest.cs
rynowak: Yeah, bad copy-paste. Will fix

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/Binders/ByteArrayModelBinderProviderTest.cs
rynowak: No way bro

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/BinderTypeModelBinder.cs
dougbu: The only check in the provider is `context.BindingInfo?.BinderType != null`. Add this check there too and everything will be :peach:y.

> I didn't see the harm in that. Thoughts?

As long as we don't find hotspots in the checks themselves, :peach:y again.

rynowak:
Pushing an update to this @dougbu 

Please look at these two and make sure they are addressed to your satisfaction/opinion:
https://github.com/aspnet/Mvc/pull/4355#discussion_r57653080
https://github.com/aspnet/Mvc/pull/4355#discussion_r57653096

Here are the comments that are won't-fix/not-actionable
https://github.com/aspnet/Mvc/pull/4355#discussion_r57738137
https://github.com/aspnet/Mvc/pull/4355#discussion_r57739573
https://github.com/aspnet/Mvc/pull/4355#discussion_r57653093 (issue)
https://github.com/aspnet/Mvc/pull/4355#discussion_r57653128

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/Binders/ComplexTypeModelBinderTest.cs
rynowak: Fixed, but not showing in the diff

src/Microsoft.AspNetCore.Mvc.Core/Internal/MvcCoreMvcOptionsSetup.cs
rynowak: :see_no_evil:

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
```diff
@@ -0,0 +1,233 @@
...
+            {
+                _cache.TryAdd(context.CacheToken, binder);
+            }
+
+            return binder;
+        }
+
+        private IModelBinder CreateBinderCore(DefaultModelBinderProviderContext providerContext, object token)
+        {
+            // A non-null token will usually be passed in at the the top level (ParameterDescriptor likely).
```
dougbu: Now caching _all_ top level binders.

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/TestModelBinderProviderContext.cs
```diff
@@ -0,0 +1,69 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.Mvc.ModelBinding
```
dougbu: Should / could this be in `TestCommon` project?

dougbu:
:shipit:

Eilon:
Nice :smile:

PR #4359, opened 2016-03-28
[Fixes #4317] UrlResolutionTagHelper doesn't work with TagHelpers that run before and supress output

### LINKED ISSUE #4317: UrlResolutionTagHelper doesn't work with TagHelpers that run before and supress output.
If a user has a `TagHelper` and sets its order to `int.MinValue`, it'll run before the `UrlResolutionTagHelper`. Now, if that same `TagHelper` calls `output.SuppressOutput()` it will then force the `TagHelperOutput`s tag name to `null`. The problem this introduces is the `UrlResolutionTagHelper` relies on the `TagName` not being `null` when looking up corresponding attribute values for the tag so it ends up exploding [here](https://github.com/aspnet/Mvc/blob/fd984563c3f7e32c41a37c2d21040b6e4490d39c/src/Microsoft.AspNetCore.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs#L129).

We should check if `TagName == null` and if so, bail early.

javiercn:
Check if the TagHelperOutput.TagName is null and do nothing in that case.

### DIFF
--- a/test/Microsoft.AspNetCore.Mvc.Razor.Test/TagHelpers/UrlResolutionTagHelperTest.cs
+++ b/test/Microsoft.AspNetCore.Mvc.Razor.Test/TagHelpers/UrlResolutionTagHelperTest.cs
@@ -36,6 +36,35 @@ namespace Microsoft.AspNetCore.Mvc.Razor.TagHelpers
             }
         }
 
+        [Fact]
+        public void Process_DoesNothingIfTagNameIsNull()
+        {
+            // Arrange
+            var tagHelperOutput = new TagHelperOutput(
+                tagName: null,
+                attributes: new TagHelperAttributeList
+                {
+                    { "href", "~/home/index.html" }
+                },
+                getChildContentAsync: (useCachedResult, encoder) => Task.FromResult<TagHelperContent>(null));
+
+            var tagHelper = new UrlResolutionTagHelper(Mock.Of<IUrlHelperFactory>(), new HtmlTestEncoder());
+            var context = new TagHelperContext(
+                allAttributes: new TagHelperAttributeList(
+                    Enumerable.Empty<TagHelperAttribute>()),
+                items: new Dictionary<object, object>(),
+                uniqueId: "test");
+
+            // Act
+            tagHelper.Process(context, tagHelperOutput);
+
+            // Assert
+            var attribute = Assert.Single(tagHelperOutput.Attributes);
+            Assert.Equal("href", attribute.Name, StringComparer.Ordinal);
+            var attributeValue = Assert.IsType<string>(attribute.Value);
+            Assert.Equal("~/home/index.html", attributeValue, StringComparer.Ordinal);
+        }
+
         [Theory]
         [MemberData(nameof(ResolvableUrlData))]
         public void Process_ResolvesTildeSlashValues(string url, string expectedHref)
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNetCore.Mvc.Razor.Test/TagHelpers/UrlResolutionTagHelperTest.cs
```diff
@@ -36,6 +36,46 @@ public static TheoryData ResolvableUrlData
...
+            // Arrange
+            var tagHelperOutput = new TagHelperOutput(
+                tagName: null,
+                attributes: new TagHelperAttributeList
+                {
+                    { "href", "~/home/index.html" }
+                },
+                getChildContentAsync: (useCachedResult, encoder) => Task.FromResult<TagHelperContent>(null));
+
+            var urlHelperMock = new Mock<IUrlHelper>();
```
NTaylorMullen: No need for this guy.

test/Microsoft.AspNetCore.Mvc.Razor.Test/TagHelpers/UrlResolutionTagHelperTest.cs
```diff
@@ -36,6 +36,46 @@ public static TheoryData ResolvableUrlData
...
+            urlHelperMock
+                .Setup(urlHelper => urlHelper.Content(It.IsAny<string>()))
+                .Returns(new Func<string, string>(value => "/approot" + value.Substring(1)));
```
NTaylorMullen: `.Throws`

test/Microsoft.AspNetCore.Mvc.Razor.Test/TagHelpers/UrlResolutionTagHelperTest.cs
```diff
@@ -36,6 +36,46 @@ public static TheoryData ResolvableUrlData
...
+
+            var urlHelperFactory = new Mock<IUrlHelperFactory>();
+            urlHelperFactory
+                .Setup(f => f.GetUrlHelper(It.IsAny<ActionContext>()))
+                .Returns(urlHelperMock.Object);
```
NTaylorMullen: `.Throws`

test/Microsoft.AspNetCore.Mvc.Razor.Test/TagHelpers/UrlResolutionTagHelperTest.cs
```diff
@@ -36,6 +36,46 @@ public static TheoryData ResolvableUrlData
...
+                    { "href", "~/home/index.html" }
+                },
+                getChildContentAsync: (useCachedResult, encoder) => Task.FromResult<TagHelperContent>(null));
+
+            var urlHelperMock = new Mock<IUrlHelper>();
+            urlHelperMock
+                .Setup(urlHelper => urlHelper.Content(It.IsAny<string>()))
+                .Returns(new Func<string, string>(value => "/approot" + value.Substring(1)));
+
+            var urlHelperFactory = new Mock<IUrlHelperFactory>();
```
NTaylorMullen: `MockBehavior.Strict`. Can remove the rest.

NTaylorMullen:
:watch: for test changes.

javiercn:
Updated sorry for the accidnental rebase

NTaylorMullen:
:shipit:

PR #4369, opened 2016-03-29
[Fixes #4246] Introducing CacheTagKey

### LINKED ISSUE #4246: CacheKey equatable object in CacheTagHelper
- Prevent string allocations
- Fix hash collision by checking the serialized key with the actual content for `DistributedCacheTagHelper`

sebastienros:
Prevent string allocations in CacheTagHelper
Fix hash collision by checking the serialized key with the actual content for DistributedCacheTagHelper

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/CacheTagKey.cs
+++ b/src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/CacheTagKey.cs
@@ -1,0 +1,324 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Security.Cryptography;
+using System.Text;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Razor.TagHelpers;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.Extensions.Internal;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNetCore.Mvc.TagHelpers.Cache
+{
+    /// <summary>
+    /// An instance of <see cref="CacheTagKey"/> represents the state of <see cref="CacheTagHelper"/>
+    /// or <see cref="DistributedCacheTagHelper"/> keys.
+    /// </summary>
+    public class CacheTagKey : IEquatable<CacheTagKey>
+    {
+        private static readonly char[] AttributeSeparator = new[] { ',' };
+        private static readonly Func<IRequestCookieCollection, string, string> CookieAcccessor = (c, key) => c[key];
+        private static readonly Func<IHeaderDictionary, string, string> HeaderAccessor = (c, key) => c[key];
+        private static readonly Func<IQueryCollection, string, string> QueryAccessor = (c, key) => c[key];
+        private static readonly Func<RouteValueDictionary, string, string> RouteValueAccessor = (c, key) => c[key]?.ToString();
+
+        private const string CacheKeyTokenSeparator = "||";
+        private const string VaryByName = "VaryBy";
+        private const string VaryByHeaderName = "VaryByHeader";
+        private const string VaryByQueryName = "VaryByQuery";
+        private const string VaryByRouteName = "VaryByRoute";
+        private const string VaryByCookieName = "VaryByCookie";
+        private const string VaryByUserName = "VaryByUser";
+
+        private readonly string _key;
+        private readonly string _prefix;
+        private readonly string _varyBy;
+        private readonly DateTimeOffset? _expiresOn;
+        private readonly TimeSpan? _expiresAfter;
+        private readonly TimeSpan? _expiresSliding;
+        private readonly IList<KeyValuePair<string, string>> _headers;
+        private readonly IList<KeyValuePair<string, string>> _queries;
+        private readonly IList<KeyValuePair<string, string>> _routeValues;
+        private readonly IList<KeyValuePair<string, string>> _cookies;
+        private readonly bool _varyByUser;
+        private readonly string _username;
+
+        private string _generatedKey;
+        private int? _hashcode;
+
+        /// <summary>
+        /// Creates an instance of <see cref="CacheTagKey"/> for a specific <see cref="CacheTagHelper"/>.
+        /// </summary>
+        /// <param name="tagHelper">The <see cref="CacheTagHelper"/>.</param>
+        /// <param name="context">The <see cref="TagHelperContext"/>.</param>
+        /// <returns>A new <see cref="CacheTagKey"/>.</returns>
+        public CacheTagKey(CacheTagHelper tagHelper, TagHelperContext context)
+            : this(tagHelper)
+        {
+            _key = context.UniqueId;
+            _prefix = nameof(CacheTagHelper);
+        }
+
+        /// <summary>
+        /// Creates an instance of <see cref="CacheTagKey"/> for a specific <see cref="DistributedCacheTagHelper"/>.
+        /// </summary>
+        /// <param name="tagHelper">The <see cref="DistributedCacheTagHelper"/>.</param>
+        /// <returns>A new <see cref="CacheTagKey"/>.</returns>
+        public CacheTagKey(DistributedCacheTagHelper tagHelper)
+            : this((CacheTagHelperBase)tagHelper)
+        {
+            _key = tagHelper.Name;
+            _prefix = nameof(DistributedCacheTagHelper);
+        }
+
+        private CacheTagKey(CacheTagHelperBase tagHelper)
+        {
+            var httpContext = tagHelper.ViewContext.HttpContext;
+            var request = httpContext.Request;
+
+            _expiresAfter = tagHelper.ExpiresAfter;
+            _expiresOn = tagHelper.ExpiresOn;
+            _expiresSliding = tagHelper.ExpiresSliding;
+            _varyBy = tagHelper.VaryBy;
+            _cookies = ExtractCollection(tagHelper.VaryByCookie, request.Cookies, CookieAcccessor);
+            _headers = ExtractCollection(tagHelper.VaryByHeader, request.Headers, HeaderAccessor);
+            _queries = ExtractCollection(tagHelper.VaryByQuery, request.Query, QueryAccessor);
+            _routeValues = ExtractCollection(tagHelper.VaryByRoute, tagHelper.ViewContext.RouteData.Values, RouteValueAccessor);
+            _varyByUser = tagHelper.VaryByUser;
+
+            if (_varyByUser)
+            {
+                _username = httpContext.User?.Identity?.Name;
+            }
+        }
+
+        /// <summary>
+        /// Creates a <see cref="string"/> representation of the key.
+        /// </summary>
+        /// <returns>A <see cref="string"/> uniquely representing the key.</returns>
+        public string GenerateKey()
+        {
+            // Caching as the key is immutable and it can be called multiple times during a request.
+            if (_generatedKey != null)
+            {
+                return _generatedKey;
+            }
+
+            var builder = new StringBuilder(_prefix);
+            builder
+                .Append(CacheKeyTokenSeparator)
+                .Append(_key);
+
+            if (!string.IsNullOrEmpty(_varyBy))
+            {
+                builder
+                    .Append(CacheKeyTokenSeparator)
+                    .Append(VaryByName)
+                    .Append(CacheKeyTokenSeparator)
+                    .Append(_varyBy);
+            }
+
+            AddStringCollection(builder, VaryByCookieName, _cookies);
+            AddStringCollection(builder, VaryByHeaderName, _headers);
+            AddStringCollection(builder, VaryByQueryName, _queries);
+            AddStringCollection(builder, VaryByRouteName, _routeValues);
+
+            if (_varyByUser)
+            {
+                builder
+                    .Append(CacheKeyTokenSeparator)
+                    .Append(VaryByUserName)
+                    .Append(CacheKeyTokenSeparator)
+                    .Append(_username);
+            }
+
+            _generatedKey = builder.ToString();
+
+            return _generatedKey;
+        }
+
+        /// <summary>
+        /// Creates a hashed value of the key.
+        /// </summary>
+        /// <returns>A cryptographic hash of the key.</returns>
+        public string GenerateHashedKey()
+        {
+            var key = GenerateKey();
+
+            // The key is typically too long to be useful, so we use a cryptographic hash
+            // as the actual key (better randomization and key distribution, so small vary
+            // values will generate dramatically different keys).
+            using (var sha = SHA256.Create())
+            {
+                var contentBytes = Encoding.UTF8.GetBytes(key);
+                var hashedBytes = sha.ComputeHash(contentBytes);
+                return Convert.ToBase64String(hashedBytes);
+            }
+        }
+
+        /// <inheritdoc />
+        public override bool Equals(object obj)
+        {
+            var other = obj as CacheTagKey;
+            if (other == null)
+            {
+                return false;
+            }
+
+            return Equals(other);
+        }
+
+        /// <inheritdoc />
+        public bool Equals(CacheTagKey other)
+        {
+            return string.Equals(other._key, _key, StringComparison.Ordinal) &&
+                other._expiresAfter == _expiresAfter &&
+                other._expiresOn == _expiresOn &&
+                other._expiresSliding == _expiresSliding &&
+                string.Equals(other._varyBy, _varyBy, StringComparison.Ordinal) &&
+                AreSame(_cookies, other._cookies) &&
+                AreSame(_headers, other._headers) &&
+                AreSame(_queries, other._queries) &&
+                AreSame(_routeValues, other._routeValues) &&
+                _varyByUser == other._varyByUser && 
+                (!_varyByUser || string.Equals(other._username, _username, StringComparison.Ordinal));
+        }
+
+        /// <inheritdoc />
+        public override int GetHashCode()
+        {
+            // The hashcode is intentionally not using the computed 
+            // stringified key in order to prevent string allocations
+            // in the common case where it's not explicitly required.
+
+            // Caching as the key is immutable and it can be called
+            // multiple times during a request.
+            if (_hashcode.HasValue)
+            {
+                return _hashcode.Value;
+            }
+
+            var hashCodeCombiner = new HashCodeCombiner();
+
+            hashCodeCombiner.Add(_key, StringComparer.Ordinal);
+            hashCodeCombiner.Add(_expiresAfter);
+            hashCodeCombiner.Add(_expiresOn);
+            hashCodeCombiner.Add(_expiresSliding);
+            hashCodeCombiner.Add(_varyBy, StringComparer.Ordinal);
+            hashCodeCombiner.Add(_username, StringComparer.Ordinal);
+
+            CombineCollectionHashCode(hashCodeCombiner, VaryByCookieName, _cookies);
+            CombineCollectionHashCode(hashCodeCombiner, VaryByHeaderName, _headers);
+            CombineCollectionHashCode(hashCodeCombiner, VaryByQueryName, _queries);
+            CombineCollectionHashCode(hashCodeCombiner, VaryByRouteName, _routeValues);
+
+            _hashcode = hashCodeCombiner.CombinedHash;
+
+            return _hashcode.Value;
+        }
+        
+        private static IList<KeyValuePair<string, string>> ExtractCollection<TSourceCollection>(string keys, TSourceCollection collection, Func<TSourceCollection, string, string> accessor)
+        {
+            if (string.IsNullOrEmpty(keys))
+            {
+                return null;
+            }
+
+            var tokenizer = new StringTokenizer(keys, AttributeSeparator);
+
+            var result = new List<KeyValuePair<string, string>>();
+
+            foreach (var item in tokenizer)
+            {
+                var trimmedValue = item.Trim();
+
+                if (trimmedValue.Length != 0)
+                {
+                    var value = accessor(collection, trimmedValue.Value);
+                    result.Add(new KeyValuePair<string, string>(trimmedValue.Value, value ?? string.Empty));
+                }
+            }
+
+            return result;
+        }
+        
+        private static void AddStringCollection(
+            StringBuilder builder,
+            string collectionName,
+            IList<KeyValuePair<string, string>> values)
+        {
+            if (values == null || values.Count == 0)
+            {
+                return;
+            }
+
+            // keyName(param1=value1|param2=value2)
+            builder
+                .Append(CacheKeyTokenSeparator)
+                .Append(collectionName)
+                .Append("(");
+
+            for (var i = 0; i < values.Count; i++)
+            {
+                var item = values[i];
+
+                if (i > 0)
+                {
+                    builder.Append(CacheKeyTokenSeparator);
+                }
+
+                builder
+                    .Append(item.Key)
+                    .Append(CacheKeyTokenSeparator)
+                    .Append(item.Value);
+            }
+            
+            builder.Append(")");
+        }
+
+        private static void CombineCollectionHashCode(
+            HashCodeCombiner hashCodeCombiner,
+            string collectionName,
+            IList<KeyValuePair<string, string>> values)
+        {
+            if (values != null)
+            {
+                hashCodeCombiner.Add(collectionName, StringComparer.Ordinal);
+
+                for (var i = 0; i < values.Count; i++)
+                {
+                    var item = values[i];
+                    hashCodeCombiner.Add(item.Key);
+                    hashCodeCombiner.Add(item.Value);
+                }
+            }
+        }
+
+        private static bool AreSame(IList<KeyValuePair<string, string>> values1, IList<KeyValuePair<string, string>> values2)
+        {
+            if (values1 == values2)
+            {
+                return true;
+            }
+
+            if (values1 == null || values2 == null || values1.Count != values2.Count)
+            {
+                return false;
+            }
+
+            for (var i = 0; i < values1.Count; i++)
+            {
+                if (!string.Equals(values1[i].Key, values2[i].Key, StringComparison.Ordinal) ||
+                    !string.Equals(values1[i].Value, values2[i].Value, StringComparison.Ordinal))
+                {
+                    return false;
+                }
+            }
+
+            return true;
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperService.cs
+++ b/src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperService.cs
@@ -1,44 +1,81 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Concurrent;
 using System.IO;
+using System.Linq;
 using System.Text;
 using System.Text.Encodings.Web;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Html;
 using Microsoft.AspNetCore.Mvc.Rendering;
+using Microsoft.AspNetCore.Mvc.TagHelpers.Internal;
 using Microsoft.AspNetCore.Razor.TagHelpers;
 using Microsoft.Extensions.Caching.Distributed;
+using Microsoft.Extensions.Logging;
 
 namespace Microsoft.AspNetCore.Mvc.TagHelpers.Cache
 {
     /// <summary>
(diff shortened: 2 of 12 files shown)


src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperService.cs
```diff
@@ -16,13 +18,17 @@ namespace Microsoft.AspNetCore.Mvc.TagHelpers.Cache
     /// <summary>
     /// Implements <see cref="IDistributedCacheTagHelperService"/> and ensure
```
javiercn: nit: ensures

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperService.cs
```diff
@@ -16,13 +18,17 @@ namespace Microsoft.AspNetCore.Mvc.TagHelpers.Cache
...
     /// multiple concurrent requests are gated.
+    /// The entries are stored like this:
```
javiercn: This probably belongs on a <remarks></remarks> section and I'm not sure about using - to define "buillet points", I don't think we do it anywhere and it will most likely not be picked up by the documentation

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperService.cs
```diff
@@ -74,20 +82,57 @@ HtmlEncoder HtmlEncoder
...
-                            if (content == null)
+                            try
                             {
-                                content = await output.GetChildContentAsync();
+                                if (contentBuffer != null)
+                                {
+                                    content = await _formatter.DeserializeAsync(contentBuffer);
+                                }
                             }
+                            finally
```
javiercn: Shouldn't you have a catch somewhere around here?

sebastienros: This should actually be a catch, good _catch_

sebastienros:
:up: :date:

Added some logging when a value can't be deserialized
Moved `CacheTagKey` to the `Cache` namespace

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperService.cs
```diff
@@ -74,19 +93,59 @@ HtmlEncoder HtmlEncoder
...
                             content = formattingContext.Html;
                         }
                         else
                         {
-                            content = await _formatter.DeserializeAsync(value);
+                            // Extract the length of the serialized key
+                            byte[] contentBuffer = null;
+                            using (var buffer = new MemoryStream(value))
+                            {
+                                var keyLengthBuffer = new byte[sizeof(int)];
```
javiercn: I would make a constant for sizeof(int) and use that same constant on line 101 in place for keyLength.Length, if i'm not mistaken they represent the same concept and the constant makes it clearer

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperService.cs
```diff
@@ -74,19 +93,59 @@ HtmlEncoder HtmlEncoder
...
+                                buffer.Read(keyLengthBuffer, 0, keyLengthBuffer.Length);
+
+                                var keyLength = BitConverter.ToInt32(keyLengthBuffer, 0);
+                                var serializedKeyBuffer = new byte[keyLength];
+                                buffer.Read(serializedKeyBuffer, 0, serializedKeyBuffer.Length);
+
+                                // Ensure we are reading the expected key before continuing
+                                if (serializedKeyBuffer.SequenceEqual(serializedKey))
```
javiercn: Is this done for hash collisions?

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperService.cs
```diff
@@ -1,44 +1,61 @@
...
             _htmlEncoder = HtmlEncoder;
-
-            _workers = new ConcurrentDictionary<string, Task<IHtmlContent>>();
+            _logger = loggerFactory.CreateLogger<DistributedCacheTagHelperService>();
+            _workers = new ConcurrentDictionary<CacheTagKey, Task<IHtmlContent>>();
         }
 
         /// <inheritdoc />
-        public async Task<IHtmlContent> ProcessContentAsync(TagHelperOutput output, string key, DistributedCacheEntryOptions options)
+        public async Task<IHtmlContent> ProcessContentAsync(TagHelperOutput output, CacheTagKey key, DistributedCacheEntryOptions options)
```
javiercn: Isn't there a reasonable way to break this method into smaller ones? It's 120 lines long with several if-else, while and try cach loops nested and It makes really hard to understand what's going on.

test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/CacheTagKeyTest.cs
```diff
@@ -0,0 +1,393 @@
...
+            Assert.Equal(expected, key);
+        }
+
+        [Theory]
+        [InlineData("Accept-Language", "CacheTagHelper||testid||VaryByHeader(Accept-Language||en-us;charset=utf8)")]
+        [InlineData("X-CustomHeader,Accept-Encoding, NotAvailable",
+            "CacheTagHelper||testid||VaryByHeader(X-CustomHeader||Header-Value||Accept-Encoding||utf8||NotAvailable||)")]
+        [InlineData("X-CustomHeader,  , Accept-Encoding, NotAvailable",
+            "CacheTagHelper||testid||VaryByHeader(X-CustomHeader||Header-Value||Accept-Encoding||utf8||NotAvailable||)")]
+        public void GenerateKey_UsesVaryByHeader(string varyByHeader, string expected)
```
javiercn: Can you have vary by multiple headers simultaneously? What about headers with multiple values? Is that handled at the HttpRequest level?

javiercn:
Overall looks good to me, but I'll let others chip in. :watch:

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/CacheTagKey.cs
```diff
@@ -0,0 +1,358 @@
...
+                return false;
+            }
+
+            return Equals((CacheTagKey)obj);
+        }
+
+        /// <inheritdoc />
+        public bool Equals(CacheTagKey other)
+        {
+            if (!string.Equals(other._key, _key, StringComparison.Ordinal) ||
```
pranavkm: `return !string.Equals(...) ...`;

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/CacheTagKey.cs
```diff
@@ -0,0 +1,358 @@
...
+            return cacheKey;
+        }
+
+        /// <summary>
+        /// Creates an instance of <see cref="CacheTagKey"/> for a specific <see cref="DistributedCacheTagHelper"/>.
+        /// </summary>
+        /// <param name="tagHelper">The <see cref="DistributedCacheTagHelper"/>.</param>
+        /// <param name="context">The <see cref="TagHelperContext"/>.</param>
+        /// <returns>A new <see cref="CacheTagKey"/>.</returns>
+        public static CacheTagKey From(DistributedCacheTagHelper tagHelper, TagHelperContext context)
```
pranavkm: Seems like an odd API name. Couldn't we have regular ctors for this instead of factories?

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/CacheTagKey.cs
```diff
@@ -0,0 +1,358 @@
...
+            if (_varyByUser)
+            {
+                builder
+                    .Append(CacheKeyTokenSeparator)
+                    .Append(VaryByUserName)
+                    .Append(CacheKeyTokenSeparator)
+                    .Append(_username);
+            }
+
+            return _generatedKey = builder.ToString();
```
pranavkm: separate lines.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/CacheTagKey.cs
```diff
@@ -0,0 +1,358 @@
...
+            cacheKey._prefix = nameof(DistributedCacheTagHelper);
+
+            return cacheKey;
+        }
+
+        /// <summary>
+        /// Creates a <see cref="string"/> representation of the key.
+        /// </summary>
+        /// <returns>A <see cref="string"/> uniquely representing the key.</returns>
+        public string GenerateKey()
```
pranavkm: Kinda unfortunate we can't get rid of this

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/CacheTagKey.cs
```diff
@@ -0,0 +1,358 @@
...
+            // Caching as the key is immutable and it can be called
+            // multiple times during a request.
+            if (_hashcode.HasValue)
+            {
+                return _hashcode.Value;
+            }
+
+            var hashCodeCombiner = new HashCodeCombiner();
+
+            hashCodeCombiner.Add(_key);
```
pranavkm: `.Add(_key, StringComparer,OrdinalIgnoreCase);`. Need to use the same comparer as the `Equals` method

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/CacheTagKey.cs
```diff
@@ -0,0 +1,358 @@
...
+        }
+
+        private static IList<KeyValuePair<string, string>> ExtractCollection<TSourceCollection>(string keys, TSourceCollection collection, Func<TSourceCollection, string, string> accessor)
+        {
+            if (string.IsNullOrEmpty(keys))
+            {
+                return null;
+            }
+
+            var values = Tokenize(keys);
```
pranavkm: Could we use `StringTokenizer` here? Might avoid the list + array allocation.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperService.cs
sebastienros: Yes

test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/CacheTagKeyTest.cs
sebastienros: I don't know the exact answer. It's a more general question for the cache tag helper.

sebastienros:
:up: :date:

Some refactoring based on feedback, used the `StringTokenizer`, split some code in more fine grained methods for the service, use of constructors.

sebastienros:
:tada: 

image

I assume that deserves some :shipit:  ... plesae

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperService.cs
```diff
@@ -1,44 +1,61 @@
...
 
         public DistributedCacheTagHelperService(
             IDistributedCacheTagHelperStorage storage,
             IDistributedCacheTagHelperFormatter formatter,
-            HtmlEncoder HtmlEncoder 
-        )
+            HtmlEncoder HtmlEncoder,
+            ILoggerFactory loggerFactory)
         {
             _formatter = formatter;
```
javiercn: null check the parameters

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperService.cs
```diff
@@ -72,28 +92,48 @@ HtmlEncoder HtmlEncoder
...
                         else
                         {
-                            content = await _formatter.DeserializeAsync(value);
-
-                            // If the deserialization fails, it can return null, for instance when the 
-                            // value is not in the expected format.
-                            if (content == null)
+                            // The value was found in the storage, decode and ensure
+                            // there is no cache key hash collision
+                            byte[] decodedValue = Decode(value, serializedKey);
```
javiercn: :smile_cat:

test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/CacheTagKeyTest.cs
javiercn: I was asking for a test that combined multiple vary by different headers, although I think overall it's well tested

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperService.cs
javiercn: How is this method tested? I would expect at least one test in which the key is already on the distributed cache and another in which it is not (basically, test Encode, Decode)

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/CacheTagKey.cs
```diff
@@ -0,0 +1,319 @@
...
+using Microsoft.Extensions.Internal;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNetCore.Mvc.TagHelpers.Cache
+{
+    /// <summary>
+    /// An instance of <see cref="CacheTagKey"/> represents the state of <see cref="CacheTagHelper"/>
+    /// or <see cref="DistributedCacheTagHelper"/> keys.
+    /// </summary>
+    public class CacheTagKey : IEquatable<CacheTagKey>
```
javiercn: nit: this class is pretty fat, and a bunch of the code here is very cohesive (the stuff for generating the cryptographic key). I would put all those pieces of code inside a private nested class and make the CacheTagKey use it when it needs to generate a key.

That way you keep the data separate and identity of the key (all the parts of the key, like it's values, the definition of equality, etc) from the details of the implementation of the hashed key.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperService.cs
sebastienros: Lots of tests in `DistributedCacheTagHelperTest` maybe you don't see it in this PR changes.

javiercn:
Add a few more tests or tell me how the things I pointed out are being tested and :shipit: 
Come talk to me if you disagree and we'll solve it offline

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/CacheTagKey.cs
```diff
@@ -0,0 +1,319 @@
...
+
+        private const string CacheKeyTokenSeparator = "||";
+        private const string VaryByName = "VaryBy";
+        private const string VaryByHeaderName = "VaryByHeader";
+        private const string VaryByQueryName = "VaryByQuery";
+        private const string VaryByRouteName = "VaryByRoute";
+        private const string VaryByCookieName = "VaryByCookie";
+        private const string VaryByUserName = "VaryByUser";
+
+        private string _key;
```
NTaylorMullen: Can't all these except `_generatedKey` and `_hashcode` be `readonly`?

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/CacheTagKey.cs
```diff
@@ -0,0 +1,319 @@
...
+        private IList<KeyValuePair<string, string>> _headers;
+        private IList<KeyValuePair<string, string>> _queries;
+        private IList<KeyValuePair<string, string>> _routeValues;
+        private IList<KeyValuePair<string, string>> _cookies;
+        private bool _varyByUser;
+        private string _username;
+
+        private string _generatedKey;
+        private int? _hashcode;
+
```
NTaylorMullen: nit: whitespace

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/CacheTagKey.cs
```diff
@@ -0,0 +1,319 @@
...
+        /// <param name="context">The <see cref="TagHelperContext"/>.</param>
+        /// <returns>A new <see cref="CacheTagKey"/>.</returns>
+        public CacheTagKey(DistributedCacheTagHelper tagHelper, TagHelperContext context)
+            : this((CacheTagHelperBase)tagHelper, context)
+        {
+            _key = tagHelper.Name;
+            _prefix = nameof(DistributedCacheTagHelper);
+        }
+
+        private CacheTagKey(CacheTagHelperBase tagHelper, TagHelperContext context)
```
NTaylorMullen: Doesn't look like this needs the `TagHelperContext`

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/CacheTagKey.cs
```diff
@@ -0,0 +1,319 @@
...
+            _prefix = nameof(CacheTagHelper);
+        }
+
+        /// <summary>
+        /// Creates an instance of <see cref="CacheTagKey"/> for a specific <see cref="DistributedCacheTagHelper"/>.
+        /// </summary>
+        /// <param name="tagHelper">The <see cref="DistributedCacheTagHelper"/>.</param>
+        /// <param name="context">The <see cref="TagHelperContext"/>.</param>
+        /// <returns>A new <see cref="CacheTagKey"/>.</returns>
+        public CacheTagKey(DistributedCacheTagHelper tagHelper, TagHelperContext context)
```
NTaylorMullen: This overload doesn't need the `TagHelperContext`

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/CacheTagKey.cs
```diff
@@ -0,0 +1,319 @@
...
+            : this((CacheTagHelperBase)tagHelper, context)
+        {
+            _key = tagHelper.Name;
+            _prefix = nameof(DistributedCacheTagHelper);
+        }
+
+        private CacheTagKey(CacheTagHelperBase tagHelper, TagHelperContext context)
+        {
+            var httpContext = tagHelper.ViewContext.HttpContext;
```
NTaylorMullen: Don't need these guys unless you're varying by user. Move them down to that block.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/CacheTagKey.cs
```diff
@@ -0,0 +1,319 @@
...
+            var request = httpContext.Request;
+
+            _expiresAfter = tagHelper.ExpiresAfter;
+            _expiresOn = tagHelper.ExpiresOn;
+            _expiresSliding = tagHelper.ExpiresSliding;
+            _varyBy = tagHelper.VaryBy;
+            _cookies = ExtractCollection(tagHelper.VaryByCookie, request.Cookies, (c, key) => c[key]);
```
NTaylorMullen: hoist lambdas as `private static readonly` for all of these.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/CacheTagKey.cs
```diff
@@ -0,0 +1,319 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Creates a <see cref="string"/> representation of the key.
+        /// </summary>
+        /// <returns>A <see cref="string"/> uniquely representing the key.</returns>
+        public string GenerateKey()
+        {
+            // Caching as the key is immutable and it can be called
```
NTaylorMullen: nit: can all be one line.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/CacheTagKey.cs
```diff
@@ -0,0 +1,319 @@
...
+                other._expiresOn == _expiresOn &&
+                other._expiresSliding == _expiresSliding &&
+                string.Equals(other._varyBy, _varyBy, StringComparison.Ordinal) &&
+                AreSame(_cookies, other._cookies) &&
+                AreSame(_headers, other._headers) &&
+                AreSame(_queries, other._queries) &&
+                AreSame(_routeValues, other._routeValues) &&
+                _varyByUser == other._varyByUser &&
+                (!_varyByUser || string.Equals(other._username, _username, StringComparison.Ordinal)
+                );
```
pranavkm: Could this go on the previous line?

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/CacheTagKey.cs
```diff
@@ -0,0 +1,319 @@
...
+                    .Append(CacheKeyTokenSeparator)
+                    .Append(_varyBy);
+            }
+
+            AddStringCollection(builder, VaryByCookieName, _cookies);
+            AddStringCollection(builder, VaryByHeaderName, _headers);
+            AddStringCollection(builder, VaryByQueryName, _queries);
+            AddStringCollection(builder, VaryByRouteName, _routeValues);
+
+            if (_varyByUser)
```
NTaylorMullen: nit: If order doesn't matter lets group the conditional blocks together. Aka:

```C#
if (!string.IsNullOrEMpty(_varyBy))
{
...
}

if (_varyByUser)
{
...
}

AddStringCollection....
...
```

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/CacheTagKey.cs
```diff
@@ -0,0 +1,319 @@
...
+                var contentBytes = Encoding.UTF8.GetBytes(key);
+                var hashedBytes = sha.ComputeHash(contentBytes);
+                return Convert.ToBase64String(hashedBytes);
+            }
+        }
+
+        /// <inheritdoc />
+        public override bool Equals(object obj)
+        {
+            if (obj == null || !(obj is CacheTagKey))
```
NTaylorMullen:
```C#
var other = obj as CacheTagKey;
if (other == null)
{
    return false;
}

return Equals(other);
```

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/CacheTagKey.cs
NTaylorMullen: move up to previous line. Odd by itself.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/CacheTagKey.cs
```diff
@@ -0,0 +1,319 @@
...
+            hashCodeCombiner.Add(_expiresSliding);
+            hashCodeCombiner.Add(_varyBy, StringComparer.Ordinal);
+            hashCodeCombiner.Add(_username, StringComparer.Ordinal);
+
+            CombineCollectionHashCode(hashCodeCombiner, VaryByCookieName, _cookies);
+            CombineCollectionHashCode(hashCodeCombiner, VaryByHeaderName, _headers);
+            CombineCollectionHashCode(hashCodeCombiner, VaryByQueryName, _queries);
+            CombineCollectionHashCode(hashCodeCombiner, VaryByRouteName, _routeValues);
+
+            _hashcode = hashCodeCombiner;
```
NTaylorMullen: Be explicit instead of relying on converters. `_hashcode = hashCodeCombiner.CombinedHash;`

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/CacheTagKey.cs
```diff
@@ -0,0 +1,319 @@
...
+                    result.Add(new KeyValuePair<string, string>(trimmedValue.Value, value ?? ""));
+                }
+            }
+
+            return result;
+        }
+        
+        private static void AddStringCollection(
+            StringBuilder builder,
+            string collectionName,
```
NTaylorMullen: not used

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/CacheTagKey.cs
```diff
@@ -0,0 +1,319 @@
...
+        private static IList<KeyValuePair<string, string>> ExtractCollection<TSourceCollection>(string keys, TSourceCollection collection, Func<TSourceCollection, string, string> accessor)
+        {
+            if (string.IsNullOrEmpty(keys))
+            {
+                return null;
+            }
+
+            var tokenizer = new StringTokenizer(keys, AttributeSeparator);
+
+            IList<KeyValuePair<string, string>> result = new List<KeyValuePair<string, string>>();
```
NTaylorMullen: var

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/CacheTagKey.cs
```diff
@@ -0,0 +1,319 @@
...
+                    .Append(CacheKeyTokenSeparator)
+                    .Append(item.Value);
+            }
+            
+            builder.Append(")");
+        }
+
+        private static void CombineCollectionHashCode(
+            HashCodeCombiner hashCodeCombiner,
+            string collectionName,
```
NTaylorMullen: not used

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/CacheTagKey.cs
```diff
@@ -0,0 +1,319 @@
...
+            IList<KeyValuePair<string, string>> result = new List<KeyValuePair<string, string>>();
+
+            foreach (var item in tokenizer)
+            {
+                var trimmedValue = item.Trim();
+
+                if (trimmedValue.Length != 0)
+                {
+                    var value = accessor(collection, trimmedValue.Value);
+                    result.Add(new KeyValuePair<string, string>(trimmedValue.Value, value ?? ""));
```
NTaylorMullen: `""` => `string.Empty`

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperService.cs
```diff
@@ -55,10 +72,13 @@ HtmlEncoder HtmlEncoder
 
                     try
                     {
-                        var value = await _storage.GetAsync(key);
-
+                        var serializedKey = Encoding.UTF8.GetBytes(key.GenerateKey());
```
NTaylorMullen: Add a comment stating why you put this here (to take advantage of the caching).

NTaylorMullen:
:watch:

sebastienros:
:up: :date:

Feedback applied. Still not broken.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/CacheTagKey.cs
```diff
@@ -0,0 +1,324 @@
...
+namespace Microsoft.AspNetCore.Mvc.TagHelpers.Cache
+{
+    /// <summary>
+    /// An instance of <see cref="CacheTagKey"/> represents the state of <see cref="CacheTagHelper"/>
+    /// or <see cref="DistributedCacheTagHelper"/> keys.
+    /// </summary>
+    public class CacheTagKey : IEquatable<CacheTagKey>
+    {
+        private static readonly char[] AttributeSeparator = new[] { ',' };
+        private static readonly Func<IRequestCookieCollection, string, string> _cookieAcccessor = (c, key) => c[key];
```
NTaylorMullen: Rename accessors to be pascal cased. Like the AttributeSeparator above.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/CacheTagKey.cs
```diff
@@ -0,0 +1,324 @@
...
+            return string.Equals(other._key, _key, StringComparison.Ordinal) &&
+                other._expiresAfter == _expiresAfter &&
+                other._expiresOn == _expiresOn &&
+                other._expiresSliding == _expiresSliding &&
+                string.Equals(other._varyBy, _varyBy, StringComparison.Ordinal) &&
+                AreSame(_cookies, other._cookies) &&
+                AreSame(_headers, other._headers) &&
+                AreSame(_queries, other._queries) &&
+                AreSame(_routeValues, other._routeValues) &&
+                _varyByUser == other._varyByUser && (!_varyByUser || string.Equals(other._username, _username, StringComparison.Ordinal)
```
NTaylorMullen: break line, bring below `);` up.

NTaylorMullen:
:shipit:

PR #4370, opened 2016-03-29
Fixing race condition in distributed cache helper unit test

sebastienros:


test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/DistributedCacheTagHelperTest.cs
```diff
@@ -764,7 +765,7 @@ public void UpdateCacheEntryOptions_SetsSlidingExpiration_IfExpiresSlidingIsSet(
             var tagHelperContext1 = GetTagHelperContext();
             var tagHelperContext2 = GetTagHelperContext();
 
-            TagHelperOutput tagHelperOutput = new TagHelperOutput(
+            TagHelperOutput tagHelperOutput1 = new TagHelperOutput(
```
NTaylorMullen: `var` just since you're touching this bit. Same below.

test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/DistributedCacheTagHelperTest.cs
```diff
@@ -777,6 +778,19 @@ public void UpdateCacheEntryOptions_SetsSlidingExpiration_IfExpiresSlidingIsSet(
                     return Task.FromResult<TagHelperContent>(tagHelperContent);
                 });
 
+            TagHelperOutput tagHelperOutput2 = new TagHelperOutput(
```
dougbu: :clap: Definitely `true` that two workers can never get the same `TagHelperOutput` in real life.

test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/DistributedCacheTagHelperTest.cs
```diff
@@ -753,6 +753,7 @@ public void UpdateCacheEntryOptions_SetsSlidingExpiration_IfExpiresSlidingIsSet(
             var childContent = "some-content";
             var resetEvent1 = new ManualResetEvent(false);
             var resetEvent2 = new ManualResetEvent(false);
+            var resetEvent3 = new ManualResetEvent(false);
```
dougbu: This covers the last test using `ManualResetEvent`s in this class. Does `CacheTagHelperTest` need a similar fix?

NTaylorMullen:
:shipit:

sebastienros:
:up: :date:

Feedback and also applied the same patch on the similar tests in `CacheTagHelper`

test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/DistributedCacheTagHelperTest.cs
dougbu: thx

dougbu:
:shipit:

PR #4371, opened 2016-03-29
@Html.Raw(null) throws exception

### LINKED ISSUE #4350: @Html.Raw(null) stops rendering the view
I have a view which outputs raw HTML for a text editor:

	@Html.Raw(Model.Description)

Description is often null. With RC1 this worked. With the nightly built from last week the output which is received by the browser stops some lines (some 100 bytes) above the code with @Html.Raw(null). No exception. Just missing HTML output. The output stops even within HTML tags or attributes.

Workaround:

	@Html.Raw(Model.Description ?? "")

ryanbrandenburg:
Fixes #4350.

Also made sure MvcSandbox includes its views as content.

### DIFF
--- a/samples/MvcSandbox/Startup.cs
+++ b/samples/MvcSandbox/Startup.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.IO;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.DependencyInjection;
@@ -34,6 +35,7 @@ namespace MvcSandbox
         {
             var host = new WebHostBuilder()
                 .UseDefaultHostingConfiguration(args)
+                .UseContentRoot(Directory.GetCurrentDirectory())
                 .UseIISPlatformHandlerUrl()
                 .UseKestrel()
                 .UseStartup<Startup>()
--- a/samples/MvcSandbox/project.json
+++ b/samples/MvcSandbox/project.json
@@ -18,6 +18,9 @@
     "Microsoft.Extensions.Logging.Console": "1.0.0-*",
     "Microsoft.NETCore.Platforms": "1.0.1-*"
   },
+  "content": [
+    "Views"
+  ],
   "tools": {
     "dotnet-razor-tooling": "1.0.0-*"
   },
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/PagedBufferedTextWriter.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/PagedBufferedTextWriter.cs
@@ -83,11 +83,21 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
 
         public override void Write(char[] buffer)
         {
+            if (buffer == null)
+            {
+                return;
+            }
+
             Write(buffer, 0, buffer.Length);
         }
 
         public override void Write(char[] buffer, int index, int count)
         {
+            if (buffer == null)
+            {
+                throw new ArgumentNullException(nameof(buffer));
+            }
+
             while (count > 0)
             {
                 var page = GetCurrentPage();
@@ -110,6 +120,11 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
 
         public override void Write(string value)
         {
+            if (value == null)
+            {
+                return;
+            }
+
             var index = 0;
             var count = value.Length;
 
--- a/test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Internal/PagedBufferedTextWriterTest.cs
+++ b/test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Internal/PagedBufferedTextWriterTest.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Buffers;
 using System.Collections.Generic;
 using System.IO;
@@ -9,11 +10,11 @@ using Xunit;
 
 namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
 {
-    public class PagedBufferedStringWriterTest
+    public class PagedBufferedTextWriterTest
     {
         private static readonly char[] Content;
 
-        static PagedBufferedStringWriterTest()
+        static PagedBufferedTextWriterTest()
         {
             Content = new char[4 * PagedBufferedTextWriter.PageSize];
             for (var i = 0; i < Content.Length; i++)
@@ -41,6 +42,24 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
 
             // Assert
             Assert.Equal<char>(Content, inner.ToString().ToCharArray());
+        }
+
+        [Fact]
+        public async Task Write_CharArray_Null()
+        {
+            // Arrange
+            var pool = new TestArrayPool();
+            var inner = new StringWriter();
+
+            var writer = new PagedBufferedTextWriter(pool, inner);
+
+            // Act
+            writer.Write((char[])null);
+
+            await writer.FlushAsync();
+
+            // Assert
+            Assert.Empty(inner.ToString());
         }
 
         [Fact]
@@ -83,6 +102,19 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
         }
 
         [Fact]
+        public void Write_CharArray_Bounded_Null()
+        {
+            // Arrange
+            var pool = new TestArrayPool();
+            var inner = new StringWriter();
+
+            var writer = new PagedBufferedTextWriter(pool, inner);
+
+            // Act & Assert
+            Assert.Throws<ArgumentNullException>("buffer", () => writer.Write(null, 0, 0));
+        }
+
+        [Fact]
         public async Task Write_CharArray_Bounded()
         {
             // Arrange
@@ -107,6 +139,24 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
 
             // Assert
             Assert.Equal<char>(Content, inner.ToString().ToCharArray());
+        }
+
+        [Fact]
+        public async Task Write_String_Null()
+        {
+            // Arrange
+            var pool = new TestArrayPool();
+            var inner = new StringWriter();
+
+            var writer = new PagedBufferedTextWriter(pool, inner);
+
+            // Act
+            writer.Write((string)null);
+
+            await writer.FlushAsync();
+
+            // Assert
+            Assert.Empty(inner.ToString());
         }
 
         [Fact]
@@ -158,7 +208,7 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
             Assert.Equal(3, pool.Returned.Count);
         }
 
-        private class TestArrayPool: ArrayPool<char>
+        private class TestArrayPool : ArrayPool<char>
         {
             public IList<char[]> Returned { get; } = new List<char[]>();
 


ryanbrandenburg:
CC @rynowak since it looks like he wrote this class.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Internal/PagedBufferedStringWriterTest.cs
```diff
@@ -44,6 +44,24 @@ static PagedBufferedStringWriterTest()
...
+
+            var writer = new PagedBufferedTextWriter(pool, inner);
+
+            // Act
+            writer.Write((char[])null);
+
+            await writer.FlushAsync();
+
+            // Assert
+            Assert.True(string.IsNullOrEmpty(inner.ToString()));
```
rynowak: Stop and think before you use `Assert.True` or `Assert.False` - it's almost always wrong.

Why is it null or empty?

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Internal/PagedBufferedStringWriterTest.cs
```diff
@@ -44,6 +44,24 @@ static PagedBufferedStringWriterTest()
         }
 
         [Fact]
+        public async Task Write_CharArray_Null()
```
rynowak: What about the other overload?

rynowak:
:watch:

ryanbrandenburg:
:up: :date:

samples/MvcSandbox/project.json
```diff
@@ -17,6 +17,9 @@
     "Microsoft.Extensions.Configuration.Json": "1.0.0-*",
     "Microsoft.Extensions.Logging.Console": "1.0.0-*"
   },
+  "content": [
+    "Views"
+  ],
```
dougbu: Why change the sandbox?

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/PagedBufferedTextWriter.cs
```diff
@@ -83,11 +83,21 @@ public override void Write(char value)
...
+
             Write(buffer, 0, buffer.Length);
         }
 
         public override void Write(char[] buffer, int index, int count)
         {
+            if (buffer == null)
+            {
+                return;
+            }
```
dougbu: This method should match `TextWriter` and / or `StringWriter`. Both of those implementations of this method will throw an `ArgumentNullException` if `buffer == null`. Do the same in this block.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Internal/PagedBufferedTextWriterTest.cs
```diff
@@ -83,6 +101,23 @@ static PagedBufferedStringWriterTest()
...
+        public async Task Write_CharArray_Bounded_Null()
+        {
+            // Arrange
+            var pool = new TestArrayPool();
+            var inner = new StringWriter();
+
+            var writer = new PagedBufferedTextWriter(pool, inner);
+
+            // Act
+            writer.Write(null, 0, 0);
```
dougbu: `Assert.Throws<ArgumentNullException>(() => writer.Write(null, 0, 0), ...)`

dougbu:
Curious: Where was the NRE in the #4350 scenario caught? I'm a bit surprised _any_ content reached the browser.

dougbu:
:watch:

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/PagedBufferedTextWriter.cs
ryanbrandenburg: Ah, you mean just this overload should throw. Seems reasonable since that's what the other writers do. I'll also take the other methods null checking structure. Functionally it'll be the same but it can't hurt to standardize.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Internal/PagedBufferedStringWriterTest.cs
ryanbrandenburg: Done.

samples/MvcSandbox/project.json
ryanbrandenburg: I was using the Sandbox to test the bug and my fix in the real world and figured reverting the stuff I did to get it to work in the new world was a waste.

ryanbrandenburg:
As for where the NRE was caught, it looks like everything above the offending function was written but nothing after.

dougbu:
> it looks like everything above the offending function was written but nothing after.

You've described the behaviour. I was asking why. For example, is there a `try` / `catch` / ignore (in MVC) on the stack, does the view buffer flush to the response when it's disposed, ...?

samples/MvcSandbox/project.json
dougbu: :ok:

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/PagedBufferedTextWriter.cs
```diff
@@ -110,25 +118,28 @@ public override void Write(char[] buffer, int index, int count)
 
         public override void Write(string value)
         {
-            var index = 0;
-            var count = value.Length;
-
-            while (count > 0)
+            if (value != null)
```
dougbu: Was more readable before: Exit early to avoid the extra nesting.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/PagedBufferedTextWriter.cs
```diff
@@ -83,11 +83,19 @@ public override void Write(char value)
...
+            {
+                Write(buffer, 0, buffer.Length);
+            }
         }
 
         public override void Write(char[] buffer, int index, int count)
         {
+            if (buffer == null)
+            {
+                throw new ArgumentNullException(nameof(buffer), Resources.ArgumentNull_Buffer);
```
dougbu: Get rid of the resource and just throw `ArgumentNullException(nameof(buffer))`. We fortunately don't use this constructor overload anywhere in MVC.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Resources.resx
```diff
@@ -280,4 +280,7 @@
   <data name="ValueInterfaceAbstractOrOpenGenericTypesCannotBeActivated" xml:space="preserve">
     <value>The type '{0}' cannot be activated by '{1}' because it is either a value type, an interface, an abstract class or an open generic type.</value>
   </data>
+  <data name="ArgumentNull_Buffer" xml:space="preserve">
```
dougbu: Revert changes to this file.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/PagedBufferedTextWriter.cs
```diff
@@ -83,129 +83,146 @@ public override void Write(char value)
...
         public override void Write(char[] buffer)
         {
+            if (buffer == null)
+            {
+                return;
+            }
+
             Write(buffer, 0, buffer.Length);
         }
+    }
```
dougbu: Huh?

samples/MvcSandbox/Startup.cs
```diff
@@ -5,6 +5,7 @@
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
+using System.IO;
```
dougbu: `System` `using`s first.

ryanbrandenburg:
@dougbu after stepping through this and consulting with @rynowak I've determined that the reason it behaves this way is that the exception is being thrown AFTER the response has already begun. Razor has constructed it's list of write operations, but not executed them and we've already sent all the headers (including 200 OK) to the client. Now we get to the list of operations from Razor which it thinks are safe, which includes our Write.Raw(null) and it throws an exception. Having already sent the response code we can't very well change it to a 500 now, so Kestrel cuts it's losses and we get an "ERR_INCOMPLETE_CHUNKED_ENCODING" and Chrome trying to show us the parts of the page it DID get.

So basically it behaves this way instead of giving us an error because the exception is thrown in a low enough level function that there's no good way to throw an error remaining. If this exception had been thrown by user code in the cshtml page rather than inside an actual Write we would have gotten the 500 and white page that I believe you expected.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/PagedBufferedTextWriter.cs
```diff
@@ -83,129 +83,146 @@ public override void Write(char value)
...
-                    page,
-                    _currentIndex,
-                    copyLength);
-
-                _currentIndex += copyLength;
-                index += copyLength;
-
-                count -= copyLength;
-            }
+            throw new ArgumentNullException(nameof(buffer));
```
ryanbrandenburg: @rynowak could you to confirm that having this throw is the correct behavior? It seems pretty dangerous because it causes the response to fail after we've already sent the headers but @dougbu is correct that that's what we do in `System.IO.TextWriter`.

rynowak: Argument validation is argument validation. If we have a condition that trips this, let's throw earlier and protect ourselves

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/PagedBufferedTextWriter.cs
```diff
@@ -83,11 +83,21 @@ public override void Write(char value)
...
+            }
+
             Write(buffer, 0, buffer.Length);
         }
 
         public override void Write(char[] buffer, int index, int count)
         {
+            if (buffer == null)
+            {
+                throw new ArgumentNullException(nameof(buffer));
```
ryanbrandenburg: @rynowak could you to confirm that having this throw is the correct behavior? It seems pretty dangerous because it causes the response to fail after we've already sent the headers but @dougbu is correct that that's what we do in System.IO.TextWriter.

P.S.: Sorry for the dupe, I accidentally staled the other comment to this effect.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/PagedBufferedTextWriter.cs
dougbu: Odd that this comment moved to another overload. But thanks for correcting the `Write(char[], int, int)` method.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/PagedBufferedTextWriter.cs
dougbu: @ryanbrandenburg your changes make this class act like almost every other `TextWriter`. That fixes inconsistencies as well as the `Raw()` problem. Having this method do anything other than `throw` for `buffer == null` creates inconsistencies and likely doesn't affect any scenario except user code calling this method directly. Even then, it just makes the behaviour of that code vary depending on exactly where in the `.cshtml` file it appears.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Internal/PagedBufferedTextWriterTest.cs
```diff
@@ -83,6 +102,19 @@ static PagedBufferedStringWriterTest()
...
+        public void Write_CharArray_Bounded_Null()
+        {
+            // Arrange
+            var pool = new TestArrayPool();
+            var inner = new StringWriter();
+
+            var writer = new PagedBufferedTextWriter(pool, inner);
+
+            // Act & Assert
+            Assert.Throws<ArgumentNullException>(() => writer.Write(null, 0, 0));
```
dougbu: Pass the `paramName` argument i.e. `Assert.Throws<ArgumentNullException>("buffer", () => writer.Write(null, 0, 0));`.

dougbu:
:shipit: with one last test fix.

PR #4372, opened 2016-03-30
[Fixes #4087] Add support for AddTagHelpersAsServices()

### LINKED ISSUE #4087: No equivalent of AddControllersAsServices() for VC's and TagHelpers

javiercn:


src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/AssemblyPart.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.AspNetCore.Mvc.ApplicationParts
     /// <summary>
     /// An <see cref="ApplicationPart"/> backed by an <see cref="Assembly"/>.
     /// </summary>
-    public class AssemblyPart : ApplicationPart
+    public class AssemblyPart : ApplicationPart, IApplicationPartTypeProvider
```
javiercn: Not for review, part of https://github.com/aspnet/Mvc/pull/4344 but I needed it here

src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/IApplicationPartTypeProvider.cs
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) .NET Foundation. All rights reserved.
```
javiercn: Not for review, part of #4344 but I needed it here

src/Microsoft.AspNetCore.Mvc.Razor/DependencyInjection/MvcRazorMvcCoreBuilderExtensions.cs
```diff
@@ -116,6 +152,11 @@ internal static void AddRazorViewEngineServices(IServiceCollection services)
                 return new DefaultChunkTreeCache(accessor.FileProvider);
             }));
 
+            services.TryAddTransient<ITagHelperDescriptorResolver, TagHelperDescriptorResolver>();
+
+            services.TryAddTransient<ITagHelperTypeResolver, TagHelperTypeResolver>();
+            services.TryAddTransient<ITagHelperDescriptorFactory>(s => new TagHelperDescriptorFactory(designTime: false));
```
javiercn: We might want to a add a parameterless constructor to TagHelperDescriptorFactory that calls into TagHelperDescriptorFactory(bool) with design time = false for use on DI and we can remove the use of the factory.

Also, these dependencies can be made singletons

src/Microsoft.AspNetCore.Mvc.Razor/DependencyInjection/MvcRazorMvcCoreBuilderExtensions.cs
```diff
@@ -123,7 +164,8 @@ internal static void AddRazorViewEngineServices(IServiceCollection services)
             // creating the singleton RazorViewEngine instance.
             services.TryAddTransient<IRazorPageFactoryProvider, DefaultRazorPageFactoryProvider>();
             services.TryAddTransient<IRazorCompilationService, RazorCompilationService>();
-            services.TryAddTransient<IMvcRazorHost, MvcRazorHost>();
+
```
javiercn: this line is just leftover

src/Microsoft.AspNetCore.Mvc.Razor.Host/MvcRazorHost.cs
```diff
@@ -143,9 +143,12 @@ public MvcRazorHost(string root)
         /// Initializes a new instance of <see cref="MvcRazorHost"/> using the specified <paramref name="chunkTreeCache"/>.
         /// </summary>
         /// <param name="chunkTreeCache">An <see cref="IChunkTreeCache"/> rooted at the application base path.</param>
-        public MvcRazorHost(IChunkTreeCache chunkTreeCache)
+        /// <param name="resolver">The <see cref="ITagHelperDescriptorResolver"/> used to resolve tag helpers on razor views.</param>
+        public MvcRazorHost(IChunkTreeCache chunkTreeCache, ITagHelperDescriptorResolver resolver)
             : this(chunkTreeCache, new RazorPathNormalizer())
         {
+
```
NTaylorMullen: extra line.

src/Microsoft.AspNetCore.Mvc.Razor/DependencyInjection/MvcRazorMvcBuilderExtensions.cs
```diff
@@ -42,6 +45,28 @@ public static class MvcRazorMvcBuilderExtensions
         }
 
         /// <summary>
+        /// Registers tag helpers as services and changes the existing <see cref="ITagHelperActivator"/>
```
NTaylorMullen: `changes` => `replaces`... `for an` => `with an`

src/Microsoft.AspNetCore.Mvc.Razor/TagHelpers/TagHelperFeature.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    /// <summary>
+    /// The list of tag helper types in an MVC application. The <see cref="TagHelperFeature"/> can be populated
+    /// using the <see cref="ApplicationPartManager"/> that is available during startup at <see cref="IMvcBuilder.PartManager"/>
+    /// and <see cref="IMvcCoreBuilder.PartManager"/> or at a later stage by requiring the <see cref="ApplicationPartManager"/>
+    /// as a dependency in a component.
+    /// </summary>
+    public class TagHelperFeature
+    {
+        public IList<TypeInfo> TagHelpers { get; } = new List<TypeInfo>();
```
NTaylorMullen: Needs docs. Also, I see all over you're doing `AsType`. Would it be simpler to just have this be an `IList<Type>`?

src/Microsoft.AspNetCore.Mvc.Razor/DependencyInjection/MvcRazorMvcBuilderExtensions.cs
```diff
@@ -42,6 +45,28 @@ public static class MvcRazorMvcBuilderExtensions
...
+        /// for an <see cref="ServiceBasedTagHelperActivator"/>.
+        /// </summary>
+        /// <param name="builder">The <see cref="IMvcBuilder"/> instance this method extends.</param>
+        /// <returns>The <see cref="IMvcBuilder"/> instance this method extends.</returns>
+        public static IMvcBuilder AddTagHelpersAsServices(this IMvcBuilder builder)
```
NTaylorMullen: This code is duplicated below for the corebuilder. Can we unify the logic and have both `AddTagHelperAsServices` call into a method that contains this? `DoSomethingUnified(PartManager, ServiceCollection)`

src/Microsoft.AspNetCore.Mvc.Razor/DependencyInjection/MvcRazorMvcCoreBuilderExtensions.cs
```diff
@@ -116,6 +152,11 @@ internal static void AddRazorViewEngineServices(IServiceCollection services)
                 return new DefaultChunkTreeCache(accessor.FileProvider);
             }));
 
+            services.TryAddTransient<ITagHelperDescriptorResolver, TagHelperDescriptorResolver>();
+
```
NTaylorMullen: No new line. Keep these services together. They all interact

src/Microsoft.AspNetCore.Mvc.Razor/DependencyInjection/MvcRazorMvcCoreBuilderExtensions.cs
NTaylorMullen: Why not make them singleton then?

src/Microsoft.AspNetCore.Mvc.Razor/TagHelpers/FeatureTagHelperTypeResolver.cs
```diff
@@ -0,0 +1,86 @@
...
+            try
+            {
+                assembly = Assembly.Load(parsedName);
+            }
+            catch (Exception ex)
+            {
+                errorSink.OnError(
+                    documentLocation,
+                    string.Format(
+                        "Cannot resolve TagHelper containing assembly '{0}'.Error: {1}",
```
NTaylorMullen: resource.

src/Microsoft.AspNetCore.Mvc.Razor/TagHelpers/FeatureTagHelperTypeResolver.cs
```diff
@@ -0,0 +1,86 @@
...
+                        parsedName.Name,
+                        ex.Message),
+                    assemblyName.Length);
+
+                return Type.EmptyTypes;
+            }
+
+            var feature = new TagHelperFeature();
+            _manager.PopulateFeature(feature);
```
NTaylorMullen: - If this class stays transient, this could be put in the ctor and happen once.
- If this class becomes a singleton, do we care about runtime changes to the feature? (If not we can move this to ctor)

src/Microsoft.AspNetCore.Mvc.Razor/DependencyInjection/MvcRazorMvcBuilderExtensions.cs
javiercn: I'm not sure is worth it, I'll let others chip in.

src/Microsoft.AspNetCore.Mvc.Razor/TagHelpers/FeatureTagHelperTypeResolver.cs
```diff
@@ -0,0 +1,86 @@
...
+
+            var results = new List<Type>();
+            for (int i = 0; i < feature.TagHelpers.Count; i++)
+            {
+                var tagHelper = feature.TagHelpers[i];
+                var tagHelperAssembly = tagHelper.Assembly;
+
+                if (tagHelperAssembly.Equals(assembly))
```
NTaylorMullen: Will this fall over with strong naming? We typically are looser with our assembly checks. [Example](https://github.com/dotnet/cli/blob/a81f1c08f0e87f8ee39e3d6e78bcc556bce7f38b/src/Microsoft.DotNet.ProjectModel.Loader/LoaderProjectContextExtensions.cs#L47-L73)

src/Microsoft.AspNetCore.Mvc.Razor/DependencyInjection/MvcRazorMvcCoreBuilderExtensions.cs
javiercn: We usually only make top level objects as singletons as the container has to hold on to them

src/Microsoft.AspNetCore.Mvc.Razor/TagHelpers/FeatureTagHelperTypeResolver.cs
```diff
@@ -0,0 +1,86 @@
...
+            {
+                throw new ArgumentNullException(nameof(manager));
+            }
+
+            _manager = manager;
+
+        }
+
+        /// <inheritdoc />
+        public IEnumerable<Type> Resolve(string assemblyName, SourceLocation documentLocation, ErrorSink errorSink)
```
NTaylorMullen: Any reason you can't just override the `GetExportedTypes` method? That way you would not need to have all the try catch error logic (could delegate to the base class)

src/Microsoft.AspNetCore.Mvc.Razor/TagHelpers/TagHelperFeature.cs
javiercn: This is meant for discovery, so Type info is better as it's the type that contains actual useful information in it. AsType is done only for registration on the service container and to coerce it to the interface required by the ITagHelperTypeResolver.

Id prefer to keep it as TypeInfo as it's consistent with other features and it's better for discovery scenarios (which is the main point of features).

src/Microsoft.AspNetCore.Mvc.Razor/TagHelpers/TagHelperFeatureProvider.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// Discovers tag helpers from a list of <see cref="ApplicationPart"/> instances.
+    /// </summary>
+    public class TagHelperFeatureProvider : IApplicationFeatureProvider<TagHelperFeature>
+    {
+        /// <inheritdoc />
+        public void PopulateFeature(IEnumerable<ApplicationPart> parts, TagHelperFeature feature)
+        {
+            foreach (var type in parts.OfType<IApplicationPartTypeProvider>().SelectMany(p => p.Types))
+            {
+                if (TagHelperConventions.IsTagHelper(type) && ! feature.TagHelpers.Contains(type))
```
NTaylorMullen: `!feature`

test/Microsoft.AspNetCore.Mvc.FunctionalTests/MvcTestFixture.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
```
NTaylorMullen: This class needs changes?

src/Microsoft.AspNetCore.Mvc.Razor/DependencyInjection/MvcRazorMvcCoreBuilderExtensions.cs
```diff
@@ -26,10 +29,19 @@ public static IMvcCoreBuilder AddRazorViewEngine(this IMvcCoreBuilder builder)
             }
 
             builder.AddViews();
+            AddRazorViewEngineFeatureProviders(builder);
             AddRazorViewEngineServices(builder.Services);
             return builder;
         }
 
+        private static void AddRazorViewEngineFeatureProviders(IMvcCoreBuilder builder)
```
NTaylorMullen: Move down. Wrong ordering :smile: [link](http://stackoverflow.com/questions/150479/order-of-items-in-classes-fields-properties-constructors-methods)

test/WebSites/ControllersFromServicesWebSite/TagHelpers/InServicesTagHelper.cs
```diff
@@ -0,0 +1,28 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Razor.TagHelpers;
+
+namespace ControllersFromServicesWebSite.TagHelpers
+{
+    [HtmlTargetElement("InServices", TagStructure = TagStructure.NormalOrSelfClosing)]
```
NTaylorMullen: Remove `TagStructure` bit. Doesn't need it.

test/Microsoft.AspNetCore.Mvc.Razor.Test/project.json
```diff
@@ -12,8 +12,10 @@
   ],
   "dependencies": {
     "Microsoft.AspNetCore.Http": "1.0.0-*",
+    "Microsoft.AspNetCore.Mvc": "1.0.0-*",
     "Microsoft.AspNetCore.Mvc.DataAnnotations": "1.0.0-*",
     "Microsoft.AspNetCore.Mvc.Formatters.Xml": "1.0.0-*",
+    "Microsoft.AspNetCore.Mvc.TagHelpers": "1.0.0-*",
```
NTaylorMullen: Pulled in transitively already.

test/Microsoft.AspNetCore.Mvc.Razor.Test/TagHelpers/TagHelperFeatureProviderTest.cs
```diff
@@ -0,0 +1,95 @@
...
+            manager.ApplicationParts.Add(new TestApplicationPart(typeof(DiscoveryTagHelper)));
+            manager.FeatureProviders.Add(new TagHelperFeatureProvider());
+
+            var feature = new TagHelperFeature();
+
+            // Act
+            manager.PopulateFeature(feature);
+
+            // Assert
+            Assert.Equal(new[] { typeof(DiscoveryTagHelper).GetTypeInfo() }, feature.TagHelpers.ToArray());
```
NTaylorMullen: var tagHelperType = Assert.Single(feature.TagHelpers);
Assert.Equal(typeof(DiscoveryTagHelper).GetTypeInfo(), tagHelperType);

test/Microsoft.AspNetCore.Mvc.Razor.Test/TagHelpers/TagHelperFeatureProviderTest.cs
```diff
@@ -0,0 +1,95 @@
...
+
+            public IEnumerable<TypeInfo> Types => new[] { typeof(AnotherTagHelper).GetTypeInfo() };
+        }
+    }
+}
+
+// These types need to be public for the test to work.
+namespace Microsoft.AspNetCore.Mvc.Razor.TagHelpers.TagHelperFeatureProviderTests
+{
+    public class DiscoveryTagHelper : TagHelper
```
NTaylorMullen: Remove process methods here and in the other taghelper

test/Microsoft.AspNetCore.Mvc.Razor.Test/TagHelpers/FeatureTagHelperTypeResolverTest.cs
```diff
@@ -0,0 +1,99 @@
...
+                foreach (var type in parts.OfType<IApplicationPartTypeProvider>().SelectMany(tp => tp.Types))
+                {
+                    feature.TagHelpers.Add(type);
+                }
+            }
+        }
+
+        private class TestTagHelper : TagHelper
+        {
+            public override void Process(TagHelperContext context, TagHelperOutput output)
```
NTaylorMullen: ditto

test/Microsoft.AspNetCore.Mvc.Razor.Test/TagHelpers/FeatureTagHelperTypeResolverTest.cs
```diff
@@ -0,0 +1,99 @@
...
+            // Arrange
+            var manager = new ApplicationPartManager();
+            var types = new[] { typeof(TestTagHelper) };
+            manager.ApplicationParts.Add(new TestApplicationPart(types));
+            manager.FeatureProviders.Add(new TestFeatureProvider());
+
+            var resolver = new FeatureTagHelperTypeResolver(manager);
+
+            // Act
+            var result = resolver.Resolve("Microsoft.AspNetCore.Mvc.Razor.Test", SourceLocation.Undefined, new ErrorSink());
```
NTaylorMullen: Any chance we can do typeof(...).Namespace or equivalent?

src/Microsoft.AspNetCore.Mvc.Razor/DependencyInjection/MvcRazorMvcCoreBuilderExtensions.cs
```diff
@@ -56,6 +70,28 @@ public static IMvcCoreBuilder AddRazorViewEngine(this IMvcCoreBuilder builder)
         }
 
         /// <summary>
+        /// Registers discovered tag helpers as services and changes the existing <see cref="ITagHelperActivator"/>
+        /// for an <see cref="ServiceBasedTagHelperActivator"/>.
+        /// </summary>
+        /// <param name="builder">The <see cref="IMvcCoreBuilder"/> instance this method extends.</param>
+        /// <returns>The <see cref="IMvcCoreBuilder"/> instance this method extends.</returns>
+        public static IMvcCoreBuilder AddTagHelpersAsServices(this IMvcCoreBuilder builder)
```
NTaylorMullen: Hmm, for this to be truly useful it requires RazorViewEngine bits to be added. Do we do things elsewhere to ensure dependencies have been added?

src/Microsoft.AspNetCore.Mvc.Razor/TagHelpers/TagHelperFeature.cs
NTaylorMullen: :ok:

src/Microsoft.AspNetCore.Mvc.Razor/DependencyInjection/MvcRazorMvcBuilderExtensions.cs
NTaylorMullen: O_o. We definitely want this bit to be consistent. If we need to add new pieces we should ensure that they're doing the same thing.

NTaylorMullen:
:watch:

src/Microsoft.AspNetCore.Mvc.Razor/TagHelpers/FeatureTagHelperTypeResolver.cs
javiercn: Will change for using the comparer, I was having issues getting this to work when I was using AssemblyNames.

src/Microsoft.AspNetCore.Mvc.Razor/TagHelpers/FeatureTagHelperTypeResolver.cs
javiercn: It's a good suggestion, I think it's possible if we also override IsType, make it return true and seal it. We want the source of truth to be the TagHelperFeatureProvider. I'll do it unless someone else has an objection to it.

test/Microsoft.AspNetCore.Mvc.Razor.Test/project.json
javiercn: My build was failing without this, but I'll give it a try.

test/Microsoft.AspNetCore.Mvc.Razor.Test/TagHelpers/TagHelperFeatureProviderTest.cs
javiercn: Will do

test/Microsoft.AspNetCore.Mvc.Razor.Test/TagHelpers/FeatureTagHelperTypeResolverTest.cs
javiercn: Good point

test/Microsoft.AspNetCore.Mvc.FunctionalTests/MvcTestFixture.cs
javiercn: I don't think so, it's most likely left over, we only add TagHelperFeature when you invoke AddTagHelpersAsServices()

test/Microsoft.AspNetCore.Mvc.Razor.Test/project.json
NTaylorMullen: With @sebastienros's changes this should be pulled in by `Microsoft.AspNetCore.Mvc` now.

test/Microsoft.AspNetCore.Mvc.Razor.Test/TagHelpers/FeatureTagHelperTypeResolverTest.cs
NTaylorMullen: Rather `typeof(...).Assembly.GetName`

src/Microsoft.AspNetCore.Mvc.Razor/DependencyInjection/MvcRazorMvcCoreBuilderExtensions.cs
javiercn: We do for AddXX methods on the service collection, but I'm not sure if we do it for things like this

test/Microsoft.AspNetCore.Mvc.Razor.Test/DependencyInjection/MvcRazorMvcBuilderExtensionsTest.cs
```diff
@@ -0,0 +1,101 @@
...
+
+            manager.FeatureProviders.Add(new TestFeatureProvider());
+
+            var builder = new MvcBuilder(services, manager);
+
+            // Act
+            builder.AddTagHelpersAsServices();
+
+            // Assert
+            var collection = services.ToList();
```
pranavkm: Much nicer if you use `Assert.Collection(services, service => { ...` especially if new services get added or the sequence gets re-ordered.

src/Microsoft.AspNetCore.Mvc.Razor/DependencyInjection/MvcRazorMvcCoreBuilderExtensions.cs
NTaylorMullen: :ok:

src/Microsoft.AspNetCore.Mvc.Razor/DependencyInjection/MvcRazorMvcBuilderExtensions.cs
rynowak: Yeah it seems pretty low effort to de-dupe this code

src/Microsoft.AspNetCore.Mvc.Razor/DependencyInjection/MvcRazorMvcCoreBuilderExtensions.cs
rynowak: I don't really have a strong feeling either way. It doesn't seem harmful to call `AddRazorViewEngine()` from here, they are all idempotent

src/Microsoft.AspNetCore.Mvc.Razor/TagHelpers/FeatureTagHelperTypeResolver.cs
```diff
@@ -0,0 +1,86 @@
...
+        /// <param name="manager">The <see cref="ApplicationPartManager"/> of the application.</param>
+        public FeatureTagHelperTypeResolver(ApplicationPartManager manager)
+        {
+            if (manager == null)
+            {
+                throw new ArgumentNullException(nameof(manager));
+            }
+
+            _manager = manager;
+
```
rynowak: extra line

src/Microsoft.AspNetCore.Mvc.Razor/TagHelpers/FeatureTagHelperTypeResolver.cs
rynowak: That seems fine to me

src/Microsoft.AspNetCore.Mvc.Razor/TagHelpers/FeatureTagHelperTypeResolver.cs
```diff
@@ -0,0 +1,86 @@
...
+                    assemblyName.Length);
+
+                return Type.EmptyTypes;
+            }
+
+            var feature = new TagHelperFeature();
+            _manager.PopulateFeature(feature);
+
+            var results = new List<Type>();
+            for (int i = 0; i < feature.TagHelpers.Count; i++)
```
rynowak: var

src/Microsoft.AspNetCore.Mvc.Razor/TagHelpers/TagHelperFeatureProvider.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    /// <summary>
+    /// Discovers tag helpers from a list of <see cref="ApplicationPart"/> instances.
+    /// </summary>
+    public class TagHelperFeatureProvider : IApplicationFeatureProvider<TagHelperFeature>
+    {
+        /// <inheritdoc />
+        public void PopulateFeature(IEnumerable<ApplicationPart> parts, TagHelperFeature feature)
+        {
+            foreach (var type in parts.OfType<IApplicationPartTypeProvider>().SelectMany(p => p.Types))
```
rynowak: How do the taghelpers include in MVC get in here?

rynowak:
:watch:

src/Microsoft.AspNetCore.Mvc.Razor/DependencyInjection/MvcRazorMvcCoreBuilderExtensions.cs
javiercn: I don't feel this is necessary, it'll be included when you call AddMvc and AddMvcCore is considered an advanced scenario, so you should know what you are doing and call AddRazorViewEngine to include those services.

javiercn:
@rynowak @NTaylorMullen Updated!!

NTaylorMullen:
:shipit:

src/Microsoft.AspNetCore.Mvc.Razor/TagHelpers/FeatureTagHelperTypeResolver.cs
```diff
@@ -0,0 +1,94 @@
...
+        protected sealed override bool IsTagHelper(TypeInfo typeInfo)
+        {
+            // Return true always as we have already decided what types are tag helpers when GetExportedTypes
+            // gets called.
+            return true;
+        }
+
+        private class AssemblyNameComparer : IEqualityComparer<AssemblyName>
+        {
+            public static readonly IEqualityComparer<AssemblyName> OrdinalIgnoreCase = new AssemblyNameComparer();
```
rynowak: Would be a little clearer to just call this `Instance` or similar. I was looking for the case sensitive version :laughing:

test/WebSites/ControllersFromServicesWebSite/Views/Another/InServicesTagHelper.cshtml
```diff
@@ -0,0 +1,2 @@
+@addTagHelper *, ControllersFromServicesWebSite
+<inservices></inservices>
```
rynowak: Would be really good to see this use one of the built-in taghelpers since that requires specific handling

rynowak:
Looks good :+1: :shipit:

PR #4379, opened 2016-03-30
[Proposal][Fixes #4187] Get rid of DNX services

### LINKED ISSUE #4187: Discussion: Microsoft.AspNetCore.Mvc.Dnx is required to allow Mvc in RC2 to work with Dnx
Discussion for https://github.com/aspnet/Announcements/issues/154

As part of our recent migration, we broke compatibility with DNX in Mvc. This prevents controller and view component discovery from working correctly and view compilation to fail in our latest sources and packages on the aspnetcidev \ aspnetvnext feeds. To address this issue, we've published `Microsoft.AspNetCore.Mvc.Dnx` to `https://www.myget.org/F/aspnetcidev` feed that would allow Mvc to continue working with DNX until we've tooling support for dotnet-cli tools.

To use this package:
1. Add the https://www.myget.org/F/aspnetcidev feed to your sources.
2. Add `Microsoft.AspNetCore.Mvc.Dnx` to your project.json.
3. In your startup under `ConfigureServices` add a call to `AddMvcDnx`:

``` C#
{
            ...
            services.AddMvc();
            services.AddMvcDnx();
            ....
}
```

javiercn:
We don't need IAssemblyProvider anymore as we have application parts, this commit removes support for IAssemblyProvider from MVC and DNX assembly provider. For customers running on RC1 using DNX that want to use nightly bits, the alternative is to add their assemblies manually to the IMvcBuilder.ApplicationPartManager during startup.


rynowak:
@javiercn - do they have to add the 'entry' assembly or do they get that for free?

src/Microsoft.AspNetCore.Mvc.Dnx/DnxRoslynCompilationService.cs
```diff
@@ -1,262 +0,0 @@
...
-using Microsoft.Dnx.Compilation.CSharp;
-using Microsoft.Extensions.CompilationAbstractions;
-using Microsoft.Extensions.Options;
-
-namespace Microsoft.AspNetCore.Mvc.Razor.Internal
-{
-    /// <summary>
-    /// A type that uses Roslyn to compile C# content and <see cref="ILibraryExporter"/> to find out references.
-    /// </summary>
-    public class DnxRoslynCompilationService : ICompilationService
```
javiercn: We might need to find an alternative for this. Can we make the default implementation make use of ApplicationPartManager to get the list of referenced assemblies?

rynowak: That's a different matter entirely. These are not the same.

javiercn:
@rynowak https://github.com/aspnet/Mvc/pull/4379/files#diff-163fbc17d95ee0fb70308a33731c7766R49

src/Microsoft.AspNetCore.Mvc.Dnx/DnxAssemblyProvider.cs
```diff
@@ -1,88 +0,0 @@
...
-using System;
-using System.Collections.Generic;
-using System.Diagnostics;
-using System.Linq;
-using System.Reflection;
-using Microsoft.Extensions.PlatformAbstractions;
-
-namespace Microsoft.AspNetCore.Mvc.Infrastructure
-{
-    public class DnxAssemblyProvider : IAssemblyProvider
```
pranavkm: Couldn't we just port this? The API doesn't seem significantly different and it avoids breaking people again.

PinpointTownes: Yes, please :+1:

pranavkm:
:shipit:

javiercn:
059ec0f3531cbbd04258c6ea5b4951ca27eecc92

PR #4380, opened 2016-03-30
Fix #4238 IFormFile model binder suppresses validation

### LINKED ISSUE #4238: Re-instate tests on Mono
In #4234 we disabled some tests on mono due to dotnet/cli#1613. Once that is resolved we should re-instate these tests.

rynowak:
This change no longer suppresses validation for IFormFile and
IFormFileCollection model values. This will allow the use of [Required] on an
IFormFile model, or a custom attribute for validating IFormFileCollection.

These types already have ValidateChildren = false, so we don't recurse
into them.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/FormCollectionModelBinder.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/FormCollectionModelBinder.cs
@@ -36,8 +36,7 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding.Binders
             {
                 model = new EmptyFormCollection();
             }
-
-            bindingContext.ValidationState.Add(model, new ValidationStateEntry() { SuppressValidation = true });
+            
             bindingContext.Result = ModelBindingResult.Success(bindingContext.ModelName, model);
         }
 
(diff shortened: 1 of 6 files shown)


dougbu:
:shipit: 

Any idea why we previously suppressed validation for the properties in addition to the type exclusion?

rynowak:
It seemed like the right thing to do at the time - I know we missed the idea of putting `[Required]` on a property.

RehanSaeed:
What's the best way to validate `IFormFile` to guard against zero byte files, unknown content types, invalid file names (This one could get complex and depends on the OS. Perhaps `Path.InvalidFileNameChars` could be used). I think these would be fairly common use cases for those using `IFormFile`. A custom `ValidationAttribute`? Would you be interested in a PR?

dougbu:
@RehanSaeed those types of validations should occur _before_ model binding. If you agree, use a middleware (which short-circuits and returns an error instead of letting MVC get involved) or an MVC resource filter (`IResourceFilter` or `IAsyncResourceFilter`).

RehanSaeed:
@dougbu So you mean looking at the raw parameters, in an `IResourceFilter` and adding errors to `ModelState`. If ModelState has errors, does that stop the `IFormFile` model binding taking place?

You would want the `IFormFile` related ModelState errors combined with any errors coming from the rest of your model if it had other properties too.

dougbu:
> So you mean looking at the raw parameters, in an `IResourceFilter` and adding errors to `ModelState`.

Yes

> If ModelState has errors, does that stop the `IFormFile` model binding taking place?

No. My suggestion was to short-circuit request processing if errors occurred. Model binding wouldn't happen at all.

> You would want the `IFormFile` related `ModelState` errors combined with any errors coming from the rest of your model if it had other properties too.

In that case, I suggest you create a custom version of `FormFileModelBinder` and `FormFileModelBinderProvider` and use those instead of the usual binder / provider. The binder would do the validation you require prior to and (in case of errors) instead of the usual binding.

PR #4381, opened 2016-03-30
[Fixes #4233] AuthorizationFilterAttribute returns UnauthorizedResult…

### LINKED ISSUE #4233: AuthorizationFilterAttribute returns UnauthorizedResult rather than ChallengeResult
RE: https://github.com/aspnet/Security/issues/720#issuecomment-192490540

This regressed at some point. Authorization failures should result in a challenge. The auth middleware will take over and decide what to do with it.
@blowdart @HaoK

kichalla:
… rather than ChallengeResult

@rynowak


test/WebSites/FiltersWebSite/Filters/AuthorizationFilterAttribute.cs
```diff
@@ -1,13 +1,15 @@
...
+using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.Authorization;
+using Microsoft.AspNetCore.Mvc.Filters;
 using Microsoft.AspNetCore.Mvc.Internal;
 
-namespace Microsoft.AspNetCore.Mvc.Filters
+namespace FiltersWebSite
 {
     [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = true, Inherited = true)]
     public abstract class AuthorizationFilterAttribute :
```
kichalla: I moved this filter attribute to test as it was being used in the filter tests.

rynowak: But what tests? IMO I'd rather see it deleted.

test/Microsoft.AspNetCore.Mvc.FunctionalTests/FiltersTest.cs
```diff
@@ -84,7 +84,7 @@ public FiltersTest(MvcTestFixture<FiltersWebSite.Startup> fixture)
             var response = await Client.GetAsync("http://localhost/Anonymous/GetHelloWorld");
 
             // Assert
-            Assert.Equal(HttpStatusCode.Unauthorized, response.StatusCode);
```
rynowak: @HaoK is this the 'right' behavior?

test/WebSites/FiltersWebSite/Filters/AuthorizationFilterAttribute.cs
kichalla: Following are some references of this attribute: https://github.com/aspnet/Mvc/search?utf8=%E2%9C%93&q=authorizationfilterattribute

rynowak: And of those, only one of them even calls into the base class. Just delete this

kichalla: Sure. :up: :date: now

rynowak:
:shipit: - please make sure commit message explains what you did, and update the bug to have that information as well

test/Microsoft.AspNetCore.Mvc.FunctionalTests/FiltersTest.cs
HaoK: I think so, your BlockAnonymous filter is kicking in which means you want a 403 Forbidden rather than 401 correct?

Issue #4287, opened 2016-03-14
AuthorizeFilter prevents unauthenticated requests to be handled by IAuthorizationService


malekpour:
Current implementation of `AuthorizeFilter` only calls `IAuthorizationService.AuthorizeAsync` when request is authenticated.

https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/Internal/AuthorizeFilter.cs#L78

I think `DenyAnonymousAuthorizationRequirement` should be used instead, otherwise there is no way to use authorization requirements without being authenticated.

Eilon:
@blowdart @Tratcher thoughts?

Tratcher:
@HaoK?

blowdart:
By design. Why go into policies when you don't have an identity to evaluate anyway?

PinpointTownes:
> By design.

Bad design then. The authorization service is already responsible of determining whether unauthenticated requests (= no principal/identity) are acceptable or not depending on the specific policy that applies, so there's no reason to have a duplicated logic in MVC.

What's the whole point of `DenyAnonymousAuthorizationRequirement` and `policy.RequiredAuthenticatedUser()` if the authorization service is not invoked when the user is not authenticated?

malekpour:
For example you can have a requirement to grant or deny access by IP address only or you may grant guest users access in restricted hours.

blowdart:
For your IP example you're now attempt to authorize on something that isn't an identity at all. So you either write an authentication middleware which builds an identity from an IP address, or you go the filter route.

malekpour:
This authentication vs. authorization has always been controversial. I've seen systems using IP, MAC address or computer name to identify and authorize group of users (e.g. Kiosk devices).

Considering the modular nature of AspNetCore, it does not make sense if MVC handles security, since the security project can take care of it very well.

In fact unauthenticated users have their own identity which is Anonymous. Lets have security project take care of them. The following code block functionality in the `AuthorizeFilter` can be replaced by a new requirement like `AllowAnonymousAuthorizationRequirement`

```c#
// Allow Anonymous skips all authorization
if (context.Filters.Any(item => item is IAllowAnonymousFilter))
{
    return;
}

PinpointTownes:
> What's the whole point of DenyAnonymousAuthorizationRequirement and policy.RequiredAuthenticatedUser() if the authorization service is not invoked when the user is not authenticated?

[...]

PR #4385, opened 2016-03-30
Filter should call through to AuthorizationService

HaoK:
Fixes https://github.com/aspnet/Mvc/issues/4287

cc @blowdart @Tratcher @rynowak

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/Authorization/AuthorizeFilter.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Authorization/AuthorizeFilter.cs
@@ -39,7 +39,7 @@ namespace Microsoft.AspNetCore.Mvc.Authorization
         public AuthorizationPolicy Policy { get; }
 
         /// <inheritdoc />
-        public virtual async Task OnAuthorizationAsync(Filters.AuthorizationFilterContext context)
+        public virtual async Task OnAuthorizationAsync(AuthorizationFilterContext context)
         {
             if (context == null)
             {
@@ -77,7 +77,6 @@ namespace Microsoft.AspNetCore.Mvc.Authorization
 
             // Note: Default Anonymous User is new ClaimsPrincipal(new ClaimsIdentity())
             if (httpContext.User == null ||
-                !httpContext.User.Identities.Any(i => i.IsAuthenticated) ||
                 !await authService.AuthorizeAsync(httpContext.User, context, Policy))
             {
                 context.Result = new ChallengeResult(Policy.AuthenticationSchemes.ToArray());
--- a/src/Microsoft.AspNetCore.Mvc.Core/Authorization/AuthorizeFilter.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Authorization/AuthorizeFilter.cs
@@ -76,8 +76,7 @@ namespace Microsoft.AspNetCore.Mvc.Authorization
             var authService = httpContext.RequestServices.GetRequiredService<IAuthorizationService>();
 
             // Note: Default Anonymous User is new ClaimsPrincipal(new ClaimsIdentity())
-            if (httpContext.User == null ||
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Mvc.Core/Authorization/AuthorizeFilter.cs
```diff
@@ -75,7 +75,6 @@ public AuthorizeFilter(AuthorizationPolicy policy)
 
             // Note: Default Anonymous User is new ClaimsPrincipal(new ClaimsIdentity())
             if (httpContext.User == null ||
```
PinpointTownes: I'm pretty sure @Tratcher told me once that this property couldn't be `null`, but maybe things have changed.

Tratcher: Still can't, at least not with our DefaultHttpContext, but who knows what implementation is under this HttpContext... :-)

PinpointTownes: Hehe yeah, people do unexpected things sometimes :smile: 

FWIW, I'd personally remove this `null` check and treat null principals like empty ones, specially since the authorization service is fine with null values.

src/Microsoft.AspNetCore.Mvc.Core/Authorization/AuthorizeFilter.cs
```diff
@@ -75,7 +75,6 @@ public AuthorizeFilter(AuthorizationPolicy policy)
...
-                !httpContext.User.Identities.Any(i => i.IsAuthenticated) ||
                 !await authService.AuthorizeAsync(httpContext.User, context, Policy))
             {
                 context.Result = new ChallengeResult(Policy.AuthenticationSchemes.ToArray());
```
PinpointTownes: Not directly related to this PR, but I'm still not very comfortable with such code. It's a decision that should be made at the authorization service level, not in the authentication middleware :-1: 

@HaoK curious: do you remember the outcome of the alternate design I had suggested for `IAuthorizationService`? https://github.com/PinpointTownes/Security/commit/7ac1a2515015a98dc6f4919f28baa9db33beb50a#commitcomment-11507157

Returning an `AuthorizationResult` class/struct/enum instead of a binary bool would allow much more flexibility IMHO, and would allow replacing the current crazy design, that consists in guessing whether an `AuthenticationTicket` can be resolved by the authentication middleware to determine if a 401 or a 403 response should be returned :smile:

src/Microsoft.AspNetCore.Mvc.Core/Authorization/AuthorizeFilter.cs
HaoK: I guess we could flow null into the auth service as well.

HaoK:
@Eilon who should sign off on this change?

Eilon:
Assigned to @kichalla to review.

test/Microsoft.AspNetCore.Mvc.Core.Test/Authorization/AuthorizeFilterTest.cs
```diff
@@ -27,6 +27,21 @@ public void InvalidUser()
         }
 
         [Fact]
+        public async Task AuthorizeFilterCanAuthorizeNonAuthenticatedUser()
+        {
+            // Arrange
+            var authorizeFilter = new AuthorizeFilter(new AuthorizationPolicyBuilder().RequireAssertion(_ => true).Build());
+            var authorizationContext = GetAuthorizationContext(services => services.AddAuthorization(), anonymous: true);
+            authorizationContext.HttpContext.User = new ClaimsPrincipal();
```
kichalla: add a variation for `null` user?

kichalla:
:shipit: after responding to the minor comment

HaoK:
f54a9648156d040fa455daffb680001a9928167d

PR #4387, opened 2016-03-30
Always clear buffers returned to the memory pool

rynowak:
Just a safety measure in case anything evades our 'flattening'. I don't
have any evidence that we're failing to clear something, but we want this
here as a safeguard.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/MemoryPoolViewBufferScope.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/MemoryPoolViewBufferScope.cs
@@ -120,7 +120,7 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
 
                 for (var i = 0; i < _leased.Count; i++)
                 {
-                    _viewBufferPool.Return(_leased[i]);
+                    _viewBufferPool.Return(_leased[i], clearArray: true);
                 }
 
                 _leased.Clear();


src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/MemoryPoolViewBufferScope.cs
```diff
@@ -120,7 +120,7 @@ public void Dispose()
 
                 for (var i = 0; i < _leased.Count; i++)
                 {
-                    _viewBufferPool.Return(_leased[i]);
+                    _viewBufferPool.Return(_leased[i], clearArray: true);
```
dougbu: What about the `Return()` in `GetPage()`?

dougbu:
Should the `Return()` calls in `PagedBufferedTextWriter` also change?

dougbu:
:watch:

dougbu:
Oops, also `JsonArrayPool` and / or `JsonInputFormatter`?

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/MemoryPoolViewBufferScope.cs
rynowak: That's only called when something really strange happens and code that will never throw throws. It's basically there as a safeguard in case someone adds more code to that block.

rynowak:
We don't have concerns about those because the data that's in there is just chars. This has objects and could result in a real leak.

dougbu:
:shipit:

PR #4388, opened 2016-03-30
Simplify BindAttribute - rename PredicateProvider

rynowak:
This change renames IPropertyBindingPredicateProvider to
IPropertyFilterProvider. The changes here are mostly renames of
parameters/variables from predicate -> propertyFilter. I did a
find+replace and left the term 'predicate' in some of the docs because it
refers to a predicate in the abstract sense.

This change also simplifies BindAttribute and removes support for type
activation.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/BindingInfo.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/BindingInfo.cs
@@ -33,7 +33,7 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
             BindingSource = other.BindingSource;
             BinderModelName = other.BinderModelName;
             BinderType = other.BinderType;
-            PropertyBindingPredicateProvider = other.PropertyBindingPredicateProvider;
+            PropertyFilterProvider = other.PropertyFilterProvider;
         }
 
         /// <summary>
@@ -52,9 +52,9 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
         public Type BinderType { get; set; }
 
         /// <summary>
-        /// Gets or sets the <see cref="ModelBinding.IPropertyBindingPredicateProvider"/>.
+        /// Gets or sets the <see cref="ModelBinding.IPropertyFilterProvider"/>.
         /// </summary>
-        public IPropertyBindingPredicateProvider PropertyBindingPredicateProvider { get; set; }
+        public IPropertyFilterProvider PropertyFilterProvider { get; set; }
 
         /// <summary>
         /// Constructs a new instance of <see cref="BindingInfo"/> from the given <paramref name="attributes"/>.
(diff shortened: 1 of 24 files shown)


src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/BindingInfo.cs
```diff
@@ -100,46 +100,45 @@ public static BindingInfo GetBindingInfo(IEnumerable<object> attributes)
...
-                bindingInfo.PropertyBindingPredicateProvider = new CompositePredicateProvider(
-                    predicateProviders);
+                bindingInfo.PropertyFilterProvider = new CompositePropertyFilterProvider(propertyFilterProviders);
             }
 
             return isBindingInfoPresent ? bindingInfo : null;
         }
 
-        private class CompositePredicateProvider : IPropertyBindingPredicateProvider
+        private class CompositePropertyFilterProvider : IPropertyFilterProvider
```
rynowak: @dougbu should we remove this and just chain?

```
predicate = (m) => p1(m) && p2(m)
```

dougbu: I'm not sure whether loop unrolling is necessary here and suggest the existing single closure might be sufficiently efficient. Further the existing code is pretty readable.

I do suggest a minor optimization for a likely-common case:
``` c#
if (propertyFilterProviders.Length == 1)
{
    bindingInfo.PropertyFilterProvider = propertyFilterProviders[0];
}
else if (propertyFilterProviders.Length > 1)
{
    bindingInfo.PropertyFilterProvider = new CompositePropertyFilterProvider(propertyFilterProviders);
}
```

If we then find `CompositePropertyFilterProvider` instances are expensive, we can always chain later.

src/Microsoft.AspNetCore.Mvc.Core/ControllerBase.cs
```diff
@@ -1221,13 +1221,13 @@ public virtual SignOutResult SignOut(AuthenticationProperties properties, params
         /// <param name="model">The model instance to update.</param>
         /// <param name="prefix">The prefix to use when looking up values in the current <see cref="IValueProvider"/>.
         /// </param>
-        /// <param name="predicate">A predicate which can be used to filter properties at runtime.</param>
+        /// <param name="propertyFilter">A predicate which can be used to filter properties at runtime.</param>
```
dougbu: :clap: liking the more-specific names.

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultBindingMetadataProvider.cs
```diff
@@ -107,34 +107,34 @@ public void CreateBindingMetadata(BindingMetadataProviderContext context)
             }
         }
 
-        private class CompositePredicateProvider : IPropertyBindingPredicateProvider
+        private class CompositePropertyFilterProvider : IPropertyFilterProvider
```
dougbu: Why do we need two of these? Should be simple-enough that making one or other version `public` won't be painful later.

rynowak: One of them is in abstractions, not making it public there

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBindingHelper.cs
```diff
@@ -607,42 +608,36 @@ internal static string GetPropertyName(Expression expression)
...
-            string prefix,
+        /// <returns>An expression which can be used with <see cref="IPropertyFilterProvider"/>.</returns>
+        public static Expression<Func<ModelMetadata, bool>> GetPropertyFilterExpression<TModel>(
             Expression<Func<TModel, object>>[] expressions)
         {
             if (expressions.Length == 0)
             {
-                // If nothing is included explcitly, treat everything as included.
-                return (context, propertyName) => true;
+                // If nothing is included explicitly, treat everything as included.
```
dougbu: No, no, the word was explicitly misspelled, I'm sure. :trollface:

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/IPropertyFilterProvider.cs
```diff
@@ -8,11 +8,11 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
...
     /// Provides a predicate which can determines which model properties should be bound by model binding.
     /// </summary>
-    public interface IPropertyBindingPredicateProvider
+    public interface IPropertyFilterProvider
     {
         /// <summary>
         /// Gets a predicate which can determines which model properties should be bound by model binding.
         /// </summary>
-        Func<ModelBindingContext, string, bool> PropertyFilter { get; }
+        Func<ModelMetadata, bool> PropertyFilter { get; }
```
dougbu: Very nice

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultBindingMetadataProvider.cs
dougbu: :frowning: ah well

dougbu:
:shipit:

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/BindingInfo.cs
rynowak: Sure

PR #4391, opened 2016-03-31
Remove IAssemblyProvider

javiercn:
* Removes IAssemblyProvider and their implementations from the code base.
* Adds DefaultAssemblyDiscoveryProvider to perform assembly discovery during startup and add the assemblies to the list of discovered parts.
* Updates AddMvcDnx to find the assemblies using the library manager and add them to the list of application parts.


javiercn:
@pranavkm @rynowak

src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/DefaultAssemblyPartDiscoveryProvider.cs
```diff
@@ -0,0 +1,94 @@
...
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using Microsoft.Extensions.DependencyModel;
+
+namespace Microsoft.AspNetCore.Mvc.ApplicationParts
+{
+    // Discovers assemblies that are part of the MVC application using the DependencyContext.
+    internal static class DefaultAssemblyPartDiscoveryProvider
```
pranavkm: Is this intentionally `internal`?

pranavkm:
Could we move the Provider to an `.Internal` namespace instead? Doesn't seem like there's anything in particular about it that would require it + methods in it to be `internal`.

Code looks :shipit:

src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/DefaultAssemblyPartDiscoveryProvider.cs
javiercn: It is, @rynowak

rynowak: This should be `.Internal` namespace not `internal static class`

javiercn:
@rynowak updated

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultAssemblyPartDiscoveryProvider.cs
```diff
@@ -0,0 +1,95 @@
...
+using Microsoft.Extensions.DependencyModel;
+
+namespace Microsoft.AspNetCore.Mvc.Internal
+{
+    // Discovers assemblies that are part of the MVC application using the DependencyContext.
+    public static class DefaultAssemblyPartDiscoveryProvider
+    {
+        private const string NativeImageSufix = ".ni";
+
+        internal static HashSet<string> ReferenceAssemblies { get; } = new HashSet<string>(StringComparer.Ordinal)
```
rynowak: These should be public or private. We don't really use the `internal` keyword

rynowak:
:shipit:

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultAssemblyPartDiscoveryProvider.cs
javiercn: They are internal for unit testing

rynowak: Nope. At least one of these methods is called by product code, it should be public.

PinpointTownes:
Thanks for keeping the MVC/DNX package updated :clap:

PR #4396, opened 2016-04-01
Quick fix: Address a couple of nits in / around `ViewDataDictionary`

dougbu:
- remove duplicate `null` checks in `ViewDataDictionary` constructors
- correct a couple of code comments
- use a `public` constructor for `ViewDataDictionary` in `ViewExecutor`

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ViewExecutor.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ViewExecutor.cs
@@ -130,11 +130,11 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
                 throw new ArgumentNullException(nameof(view));
             }
 
-            var services = actionContext.HttpContext.RequestServices;
             if (viewData == null)
             {
+                var services = actionContext.HttpContext.RequestServices;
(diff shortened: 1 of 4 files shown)


dougbu:
Mostly FYI but @ryanbrandenburg please glance through.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ViewExecutor.cs
```diff
@@ -130,11 +130,11 @@ public class ViewExecutor
                 throw new ArgumentNullException(nameof(view));
             }
 
-            var services = actionContext.HttpContext.RequestServices;
             if (viewData == null)
             {
+                var services = actionContext.HttpContext.RequestServices;
                 var metadataProvider = services.GetRequiredService<IModelMetadataProvider>();
-                viewData = new ViewDataDictionary(metadataProvider);
+                viewData = new ViewDataDictionary(metadataProvider, actionContext.ModelState);
```
dougbu: `ModelState` is a get-only property that can never be `null`. Therefore, no need to call the `internal` constructor.

ryanbrandenburg:
This all seems legit. :shipit:

dougbu:
384fd1f

Issue #4116, opened 2016-02-19
Changes to property ModelMetadata provided by a ModelMetadataDetailsProvider are lost for Nullable<T> in Display/Edit templates when DataType is used to change the template


rwasef1830:
Hello ASP.net team,
This bug is related to #2539, but the fix for that bug doesn't work when the nullable property has a DataType attribute that changes the display / edit templates.

```csharp
protected ViewDataDictionary(ViewDataDictionary source, object model, Type declaredModelType)
            : this(...)
        {
            // ...

            // This is the core constructor called when Model is known.
            var modelType = GetModelType(model);
            var metadataModelType = source.ModelMetadata.UnderlyingOrModelType;
            if (modelType == metadataModelType && model == source.ModelExplorer.Model)
            {
                // Preserve any customizations made to source.ModelExplorer.ModelMetadata if the Type
                // that will be calculated in SetModel() and source.Model match new instance's values.
                ModelExplorer = source.ModelExplorer;
            }
            else if (model == null)
            {
                // Ensure ModelMetadata is never null though SetModel() isn't called below.
                ModelExplorer = _metadataProvider.GetModelExplorerForType(_declaredModelType, model: null);
            }

            // If we're constructing a ViewDataDictionary<TModel> where TModel is a non-Nullable value type,
            // SetModel() will throw if we try to call it with null. We should not throw in that case.
            if (model != null)
            {
                SetModel(model);
            }
        }
```

The main part of that fix was this part:
```csharp
var modelType = GetModelType(model);
var metadataModelType = source.ModelMetadata.UnderlyingOrModelType;
if (modelType == metadataModelType && model == source.ModelExplorer.Model) { ... }
```
I suspect that the code does not handle nullable types correctly because ```metadataModelType``` will never be ```Nullable<T>``` whereas ```modelType``` can be, so the ModelExplorer is being always discarded in case of nullable properties.

https://github.com/rwasef1830/aspnet-mvc6-modelmetadata-repro

Hope that helps.

dougbu:
The problem is only partially about `Nullable<T>`. 

[...]

PR #4403, opened 2016-04-04
Stop using runtime type in `ViewDataDictionary` to determine `ModelMetadata`

dougbu:
- #4116
- greatly simplify rules for `ModelMetadata`; entirely remove metadata changes when Model is updated
- update `HtmlHelper.RenderPartialCoreAsync()` to ensure metadata is _not_ copied unless using existing model
- note existing functional tests did not need to change

nits:
- do not call `virtual SetModel()` method from constructor; now mostly redundant
- add some missing doc comments
- fix doc comment property versus type confusion; never need to specify `ViewDataDictionary.` prefix

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/HtmlHelper.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/HtmlHelper.cs
@@ -589,7 +589,7 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
                 // Determine which ViewData we should use to construct a new ViewData
                 var baseViewData = viewData ?? ViewData;
 
-                var newViewData = new ViewDataDictionary(baseViewData, model);
+                var newViewData = new ViewDataDictionary<object>(baseViewData, model);
                 var viewContext = new ViewContext(ViewContext, view, newViewData, writer);
 
                 await viewEngineResult.View.RenderAsync(viewContext);
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/HtmlHelper.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/HtmlHelper.cs
@@ -589,7 +589,15 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
                 // Determine which ViewData we should use to construct a new ViewData
                 var baseViewData = viewData ?? ViewData;
 
-                var newViewData = new ViewDataDictionary(baseViewData, model);
+                // Make an identical, defensive copy of the ViewDataDictionary to isolate changes.
+                ViewDataDictionary newViewData = new ViewDataDictionary(baseViewData);
+                if (!object.ReferenceEquals(model, baseViewData.Model))
+                {
+                    // Restart metadata from scratch since we know only the runtime type of the model. Of course an
+                    // @model directive in the partial view may give helpers in that view more information.
+                    newViewData.ModelExplorer = MetadataProvider.GetModelExplorerForType(typeof(object), model);
+                }
+
                 var viewContext = new ViewContext(ViewContext, view, newViewData, writer);
 
                 await viewEngineResult.View.RenderAsync(viewContext);
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/HtmlHelper.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/HtmlHelper.cs
@@ -590,7 +590,7 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
                 var baseViewData = viewData ?? ViewData;
 
                 // Make an identical, defensive copy of the ViewDataDictionary to isolate changes.
-                ViewDataDictionary newViewData = new ViewDataDictionary(baseViewData);
+                var newViewData = new ViewDataDictionary(baseViewData);
                 if (!object.ReferenceEquals(model, baseViewData.Model))
                 {
                     // Restart metadata from scratch since we know only the runtime type of the model. Of course an
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/HtmlHelper.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/HtmlHelper.cs
@@ -589,15 +589,7 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
                 // Determine which ViewData we should use to construct a new ViewData
                 var baseViewData = viewData ?? ViewData;
 
-                // Make an identical, defensive copy of the ViewDataDictionary to isolate changes.
-                var newViewData = new ViewDataDictionary(baseViewData);
-                if (!object.ReferenceEquals(model, baseViewData.Model))
-                {
-                    // Restart metadata from scratch since we know only the runtime type of the model. Of course an
-                    // @model directive in the partial view may give helpers in that view more information.
-                    newViewData.ModelExplorer = MetadataProvider.GetModelExplorerForType(typeof(object), model);
-                }
-
+                var newViewData = new ViewDataDictionary(baseViewData, model);
                 var viewContext = new ViewContext(ViewContext, view, newViewData, writer);
 
                 await viewEngineResult.View.RenderAsync(viewContext);
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ViewDataDictionary.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ViewDataDictionary.cs
@@ -58,23 +58,6 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
         }
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="ViewDataDictionary"/> class based in part on an existing
-        /// instance. This constructor is careful to avoid exceptions <see cref="SetModel"/> may throw when
-        /// <paramref name="model"/> is <c>null</c>.
-        /// </summary>
-        /// <param name="source"><see cref="ViewDataDictionary"/> instance to copy initial values from.</param>
-        /// <param name="model">Value for the <see cref="Model"/> property.</param>
-        /// <remarks>
-        /// For use when the new instance's declared <see cref="Model"/> <see cref="Type"/> is unknown but its
-        /// <see cref="Model"/> is known. In this case, <see cref="object"/> is the best possible guess about the
-        /// declared type.
-        /// </remarks>
-        public ViewDataDictionary(ViewDataDictionary source, object model)
-            : this(source, model, declaredModelType: typeof(object))
-        {
-        }
-
-        /// <summary>
         /// Initializes a new instance of the <see cref="ViewDataDictionary"/> class.
         /// </summary>
         /// <param name="metadataProvider">
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ViewDataDictionary.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ViewDataDictionary.cs
@@ -42,9 +42,15 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
         /// </summary>
         /// <param name="source"><see cref="ViewDataDictionary"/> instance to copy initial values from.</param>
         /// <remarks>
+        /// <para>
         /// For use when copying a <see cref="ViewDataDictionary"/> instance and the declared <see cref="Model"/>
         /// <see cref="Type"/> will not change e.g. when copying from a <see cref="ViewDataDictionary{TModel}"/>
         /// instance to a base <see cref="ViewDataDictionary"/> instance.
+        /// </para>
+        /// <para>
+        /// This constructor should not be used in any context where <see cref="Model"/> may be set to a value
+        /// incompatible with the declared type of <paramref name="source"/>.
+        /// </para>
         /// </remarks>
         public ViewDataDictionary(ViewDataDictionary source)
             : this(source, source.Model, source._declaredModelType)
@@ -230,6 +236,11 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
             {
                 // Source's ModelExplorer is already exactly correct.
                 ModelExplorer = source.ModelExplorer;
+            }
+            else if (model != null && !source.ModelMetadata.ModelType.IsAssignableFrom(model.GetType()))
+            {
+                // Base ModelMetadata on the declared type when new model has an incompatible type.
+                ModelExplorer = _metadataProvider.GetModelExplorerForType(declaredModelType, model);
             }
             else
             {
@@ -447,21 +458,37 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
         /// <param name="value">New <see cref="Model"/> value.</param>
         protected virtual void SetModel(object value)
         {
-            EnsureCompatible(value);
-
-            // Update ModelExplorer to reflect the new value. Preserve ModelMetadata to avoid losing property
-            // information or unexpectedly reflecting runtime information.
-            //
-            // Note EnsureCompatible() has already ensured ModelMetadata is correct.
+            // Update ModelExplorer to reflect the new value. When possible, preserve ModelMetadata to avoid losing
+            // property information or unexpectedly reflecting runtime information.
             if (object.ReferenceEquals(value, Model))
             {
-                // The metadata matches and the model is literally the same; nothing to do here.
+                // The metadata matches and the model is literally the same; usually nothing to do here.
+                if (value == null &&
+                    !ModelMetadata.IsReferenceOrNullableType &&
+                    _declaredModelType != ModelMetadata.ModelType)
+                {
+                    // Base ModelMetadata on the declared type when source VDD's Model was never set and it had a
+                    // non-Nullable value type (an unusual case). Though _declaredModelType might be another
+                    // non-Nullable value type, would otherwise need to duplicate logic behind
+                    // ModelMetadata.IsReferenceOrNullableType.
+                    ModelExplorer = _metadataProvider.GetModelExplorerForType(_declaredModelType, value);
+                }
+            }
+            else if (value != null && !ModelMetadata.ModelType.IsAssignableFrom(value.GetType()))
+            {
+                // Base ModelMetadata on the declared type when switching to an incompatible type. This should be
+                // possible only in cases where _declaredModelType==typeof(object), metadata was copied from another
+                // VDD, and user code can set the Model to a new type e.g. within a view component or a view that lacks
+                // an @model directive.
+                ModelExplorer = _metadataProvider.GetModelExplorerForType(_declaredModelType, value);
             }
             else
             {
                 // The metadata matches, but it's a new value.
                 ModelExplorer = new ModelExplorer(_metadataProvider, ModelExplorer.Container, ModelMetadata, value);
             }
+
+            EnsureCompatible(value);
         }
 
         // Throw if given value is incompatible with the declared Model Type.
@@ -486,6 +513,8 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
             }
         }
 
+        // Call after updating the ModelExplorer because this uses both _declaredModelType and ModelMetadata. May
+        // otherwise get incorrect compatibility errors.
         private bool IsCompatibleWithDeclaredType(object value)
         {
             if (value == null)
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ViewDataDictionary.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ViewDataDictionary.cs
@@ -222,7 +222,7 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
             var modelType = model?.GetType();
             var modelOrDeclaredType = modelType ?? declaredModelType;
             if (source.ModelMetadata.MetadataKind == ModelMetadataKind.Type &&
-                modelOrDeclaredType != source.ModelMetadata.ModelType &&
+                source.ModelMetadata.ModelType == typeof(object) &&
                 modelOrDeclaredType != typeof(object))
             {
                 // Base ModelMetadata on new type when there's no property information to preserve and type changes to
@@ -466,8 +466,8 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
             // property information.
             var modelType = value?.GetType();
             if (ModelMetadata.MetadataKind == ModelMetadataKind.Type &&
+                ModelMetadata.ModelType == typeof(object) &&
                 modelType != null &&
-                modelType != ModelMetadata.ModelType &&
                 modelType != typeof(object))
             {
                 // Base ModelMetadata on new type when there's no property information to preserve and type changes to
@@ -477,8 +477,8 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
             else if (modelType != null && !ModelMetadata.ModelType.IsAssignableFrom(modelType))
             {
                 // Base ModelMetadata on new type when new model is incompatible with the existing metadata. The most
-                // common case would be _declaredModelType==typeof(object), metadata was copied from another VDD, and
-                // user code sets the Model to a new type e.g. within a view component or a view that lacks an @model
+                // common case is _declaredModelType==typeof(object), metadata was copied from another VDD, and user
+                // code sets the Model to a new type e.g. within a view component or a view that lacks an @model
                 // directive.
                 ModelExplorer = _metadataProvider.GetModelExplorerForType(modelType, value);
             }
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ViewDataDictionary.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ViewDataDictionary.cs
@@ -9,6 +9,7 @@ using System.Globalization;
 using System.Reflection;
 #endif
 using Microsoft.AspNetCore.Mvc.ModelBinding;
+using Microsoft.AspNetCore.Mvc.ModelBinding.Metadata;
 using Microsoft.Extensions.Internal;
 
 namespace Microsoft.AspNetCore.Mvc.ViewFeatures
@@ -23,8 +24,8 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
         /// Initializes a new instance of the <see cref="ViewDataDictionary"/> class.
         /// </summary>
         /// <param name="metadataProvider">
-        /// <see cref = "IModelMetadataProvider" /> instance used to calculate
-        /// <see cref="ViewDataDictionary.ModelMetadata"/> values.
+        /// <see cref="IModelMetadataProvider"/> instance used to create <see cref="ViewFeatures.ModelExplorer"/>
+        /// instances.
         /// </param>
         /// <param name="modelState"><see cref="ModelStateDictionary"/> instance for this scope.</param>
         /// <remarks>For use when creating a <see cref="ViewDataDictionary"/> for a new top-level scope.</remarks>
@@ -60,7 +61,7 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
         /// <remarks>
         /// For use when the new instance's declared <see cref="Model"/> <see cref="Type"/> is unknown but its
         /// <see cref="Model"/> is known. In this case, <see cref="object"/> is the best possible guess about the
-        /// declared type when <paramref name="model"/> is <c>null</c>.
+        /// declared type.
         /// </remarks>
         public ViewDataDictionary(ViewDataDictionary source, object model)
             : this(source, model, declaredModelType: typeof(object))
@@ -71,8 +72,8 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
         /// Initializes a new instance of the <see cref="ViewDataDictionary"/> class.
         /// </summary>
         /// <param name="metadataProvider">
-        /// <see cref="IModelMetadataProvider"/> instance used to calculate
-        /// <see cref="ViewDataDictionary.ModelMetadata"/> values.
+        /// <see cref="IModelMetadataProvider"/> instance used to create <see cref="ViewFeatures.ModelExplorer"/>
+        /// instances.
         /// </param>
         /// <remarks>Internal for testing.</remarks>
         internal ViewDataDictionary(IModelMetadataProvider metadataProvider)
@@ -84,12 +85,11 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
         /// Initializes a new instance of the <see cref="ViewDataDictionary"/> class.
         /// </summary>
         /// <param name="metadataProvider">
-        /// <see cref = "IModelMetadataProvider" /> instance used to calculate
-        /// <see cref="ViewDataDictionary.ModelMetadata"/> values.
+        /// <see cref="IModelMetadataProvider"/> instance used to create <see cref="ViewFeatures.ModelExplorer"/>
+        /// instances.
         /// </param>
         /// <param name="declaredModelType">
-        /// <see cref="Type"/> of <see cref="Model"/> values expected. Used to set
-        /// <see cref="ViewDataDictionary.ModelMetadata"/> when <see cref="Model"/> is <c>null</c>.
+        /// <see cref="Type"/> of <see cref="Model"/> values expected. Used to set <see cref="ModelMetadata"/>.
         /// </param>
         /// <remarks>
         /// For use when creating a derived <see cref="ViewDataDictionary"/> for a new top-level scope.
@@ -105,17 +105,17 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
         /// Initializes a new instance of the <see cref="ViewDataDictionary"/> class.
         /// </summary>
         /// <param name="metadataProvider">
-        /// <see cref = "IModelMetadataProvider" /> instance used to calculate
-        /// <see cref="ViewDataDictionary.ModelMetadata"/> values.
+        /// <see cref="IModelMetadataProvider"/> instance used to create <see cref="ViewFeatures.ModelExplorer"/>
+        /// instances.
         /// </param>
         /// <param name="modelState"><see cref="ModelStateDictionary"/> instance for this scope.</param>
         /// <param name="declaredModelType">
-        /// <see cref="Type"/> of <see cref="Model"/> values expected. Used to set
-        /// <see cref="ViewDataDictionary.ModelMetadata"/> when <see cref="Model"/> is <c>null</c>.
+        /// <see cref="Type"/> of <see cref="Model"/> values expected. Used to set <see cref="ModelMetadata"/>.
         /// </param>
         /// <remarks>
         /// For use when creating a derived <see cref="ViewDataDictionary"/> for a new top-level scope.
         /// </remarks>
+        // This is the core constructor called when Model is unknown.
         protected ViewDataDictionary(
             IModelMetadataProvider metadataProvider,
             ModelStateDictionary modelState,
@@ -141,7 +141,7 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
                 throw new ArgumentNullException(nameof(declaredModelType));
             }
 
-            // This is the core constructor called when Model is unknown. Base ModelMetadata on the declared type.
+            // Base ModelMetadata on the declared type.
             ModelExplorer = _metadataProvider.GetModelExplorerForType(declaredModelType, model: null);
         }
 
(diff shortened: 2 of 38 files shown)


src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/HtmlHelper.cs
```diff
@@ -589,7 +589,15 @@ public Task RenderPartialAsync(string partialViewName, object model, ViewDataDic
                 // Determine which ViewData we should use to construct a new ViewData
                 var baseViewData = viewData ?? ViewData;
 
-                var newViewData = new ViewDataDictionary(baseViewData, model);
+                // Make an identical, defensive copy of the ViewDataDictionary to isolate changes.
+                ViewDataDictionary newViewData = new ViewDataDictionary(baseViewData);
+                if (!object.ReferenceEquals(model, baseViewData.Model))
+                {
+                    // Restart metadata from scratch since we know only the runtime type of the model. Of course an
+                    // @model directive in the partial view may give helpers in that view more information.
```
rynowak: Is the idea with this that the `@model` will **definitely** give more information?

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/HtmlHelper.cs
```diff
@@ -589,7 +589,15 @@ public Task RenderPartialAsync(string partialViewName, object model, ViewDataDic
                 // Determine which ViewData we should use to construct a new ViewData
                 var baseViewData = viewData ?? ViewData;
 
-                var newViewData = new ViewDataDictionary(baseViewData, model);
+                // Make an identical, defensive copy of the ViewDataDictionary to isolate changes.
+                ViewDataDictionary newViewData = new ViewDataDictionary(baseViewData);
```
rynowak: var

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ViewDataDictionary.cs
```diff
@@ -23,8 +24,8 @@ public class ViewDataDictionary : IDictionary<string, object>
         /// Initializes a new instance of the <see cref="ViewDataDictionary"/> class.
         /// </summary>
         /// <param name="metadataProvider">
-        /// <see cref = "IModelMetadataProvider" /> instance used to calculate
-        /// <see cref="ViewDataDictionary.ModelMetadata"/> values.
+        /// <see cref="IModelMetadataProvider"/> instance used to create <see cref="ViewFeatures.ModelExplorer"/>
+        /// instances.
```
rynowak: Do we really need to explain this? MMP is used for lots of things.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/HtmlHelper.cs
dougbu: Not definitely (in bold or otherwise) but that it _can_ provide more information and that `@model` describes intent better than `Model.GetType()`.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/HtmlHelper.cs
dougbu: Opps, I temporarily had an `if` and `else` that assigned to this variable. Forgot to revert the type.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ViewDataDictionary.cs
dougbu: Well, MMP is only used for this purpose. My intent here was to correct the wording and I don't see the extra explanation as a negative.

dougbu:
:up::date: addressed `var` comment and answered the others

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ViewDataDictionary.cs
rynowak: Yeah it's not wrong, I just don't think it's super useful either.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ViewFeatures/ViewDataDictionaryOfTModelTest.cs
```diff
@@ -124,33 +124,50 @@ public void CopyConstructors_InitalizeModelAndModelMetadataBasedOnSource_ModelOf
...
+            Assert.NotNull(viewData.TemplateInfo);
+            Assert.Equal("prefix", viewData.TemplateInfo.HtmlFieldPrefix);
+            Assert.NotSame(source.TemplateInfo, viewData.TemplateInfo);
+            Assert.Same(model, viewData.Model);
+            Assert.NotNull(viewData.ModelMetadata);
+            Assert.Equal(typeof(TestModel), viewData.ModelMetadata.ModelType);
+            Assert.NotSame(source.ModelMetadata, viewData.ModelMetadata);
+            Assert.Equal(source.Count, viewData.Count);
+            Assert.Equal("bar", viewData["foo"]);
+            Assert.IsType<CopyOnWriteDictionary<string, object>>(viewData.Data);
```
rynowak: Would be way clearer if these tests were focused on what's important about each distinct test. As it is, this is an insurmountable mountain of `Assert`s.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ViewFeatures/ViewDataDictionaryTest.cs
```diff
@@ -50,33 +49,26 @@ public void ConstructorInitalizesMembers()
...
             // Assert
             Assert.NotNull(viewData.ModelMetadata);
-            metadataProvider.Verify();
+            metadataProvider.Verify(m => m.GetMetadataForType(typeof(object)), Times.Once());
         }
 
-        // When SetModel is called, only GetMetadataForType from MetadataProvider is expected to be called.
         [Fact]
-        public void SetModelCallsGetMetadataForTypeExactlyOnce()
+        public void SetModel_DoesNotUsePassedInModelMetadataProvider()
```
rynowak: I don't think this is exactly what's being verified - this seems to imply that there's a second MMP.

It's just not doing anything...

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ViewFeatures/ViewDataDictionaryTest.cs
```diff
@@ -97,12 +85,9 @@ public void SetModelCallsGetMetadataForTypeExactlyOnce()
...
             // Assert
             Assert.NotNull(viewData.ModelMetadata);
-            // Verifies if the GetMetadataForType is called only once.
-            metadataProvider.Verify(
-                m => m.GetMetadataForType(typeof(object)), Times.Once());
-            // Verifies if GetMetadataForProperties and GetMetadataForProperty is not called.
-            metadataProvider.Verify(
-                m => m.GetMetadataForProperties(typeof(object)), Times.Never());
+
+            // Count is the same as in Constructor_UsesModelMetadataProvider().
```
rynowak: Neither old nor new comment is especially enlightening

rynowak:
:shipit: - changed here seem fine all my comments are nitpicks on tests and comments

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ViewDataDictionary.cs
dougbu: :see_no_evil:

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ViewFeatures/ViewDataDictionaryTest.cs
```diff
@@ -50,33 +49,26 @@ public void ConstructorInitalizesMembers()
...
             // Assert
             Assert.NotNull(viewData.ModelMetadata);
-            metadataProvider.Verify();
+            metadataProvider.Verify(m => m.GetMetadataForType(typeof(object)), Times.Once());
         }
 
-        // When SetModel is called, only GetMetadataForType from MetadataProvider is expected to be called.
         [Fact]
-        public void SetModelCallsGetMetadataForTypeExactlyOnce()
+        public void SetModel_DoesNotUseModelMetadataProvider()
```
dougbu: Note `SetModel()` may pass the `IModelMetadataProvider` into the `ModelExplorer` constructor. But nothing in `ViewDataDictionary` (in `SetModel()` or anywhere) does anything that involves the provider referenced in `ModelExplorer`.

dougbu:
:up::date: FYI

dougbu:
:up::date:

Taking all the tests and none of the source changes into a new branch where I'll build. That'll make it easy to comment on if / when the new tests fail in the old world.

In the meantime, a few general comments:
- It's almost-always Just Wrong<sup>:tm:</sup> to use runtime information when generating HTML. Unfortunately the previous `ViewDataDictionary` code updated `ModelMetadata` to match runtime types.
- In some cases, switching to runtime types overwrote property metadata and lost data annotations. That primarily affected templates for leaf properties e.g. `RazorPage<string>` subclasses.
- Overlapping this, we have a `ViewDataDictionary` copy constructor which preserves the declared type of a source VDD. This was unfortunately used when setting up a view component. In that scenario, the VDD was exposed to user code as a base `ViewDataDictionary` but would not allow assignment of values incompatible with the declared type.

Deeper background:
- `ViewDataDictionary` instances are created in only a few places and almost-always copied from another VDD instance. The general idea is that `TemplateInfo` and other data just flows through while `Model` and `ModelMetadata` information may or may not be copied from source instances.
- Type changes such as `ViewDataDictionary<string>` -> `ViewDataDictionary<int>` _never_ happen. The source type is always either loosened e.g. `ViewDataDictonary<string>` -> ViewDataDictionary<object>` inside `@Html.PartialAsync()` or tightened e.g. `ViewDataDictionary<object>` -> `ViewDataDictionary<int>` due to an `@model int` directive.
- Uninitialized non-Nullable value types are special cased.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ViewDataDictionary.cs
```diff
@@ -41,9 +42,15 @@ public class ViewDataDictionary : IDictionary<string, object>
...
         /// For use when copying a <see cref="ViewDataDictionary"/> instance and the declared <see cref="Model"/>
         /// <see cref="Type"/> will not change e.g. when copying from a <see cref="ViewDataDictionary{TModel}"/>
         /// instance to a base <see cref="ViewDataDictionary"/> instance.
+        /// </para>
+        /// <para>
+        /// This constructor should not be used in any context where <see cref="Model"/> may be set to a value
+        /// incompatible with the declared type of <paramref name="source"/>.
+        /// </para>
         /// </remarks>
         public ViewDataDictionary(ViewDataDictionary source)
```
dougbu: @Eilon this constructor is unsafe because it preserves the source's `_declaredModelType`. It's also no longer used in our product code. Any objection to removing it completely?

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ViewDataDictionary.cs
```diff
@@ -60,7 +67,7 @@ public ViewDataDictionary(ViewDataDictionary source)
         /// <remarks>
         /// For use when the new instance's declared <see cref="Model"/> <see cref="Type"/> is unknown but its
         /// <see cref="Model"/> is known. In this case, <see cref="object"/> is the best possible guess about the
-        /// declared type when <paramref name="model"/> is <c>null</c>.
+        /// declared type.
         /// </remarks>
         public ViewDataDictionary(ViewDataDictionary source, object model)
```
dougbu: @Eilon this constructor is almost identical to calling `new ViewDataDictionary<object>(source, model)` but it's less explicit. The only product code using it is within `@Html.PartialAsync()`. Any objection to removing it?

dougbu:
:up::date: to restore more of the old behaviour while still preserving property metadata where possible.

Still working on changing the tests.

dougbu:
:up::date: tests working again

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ViewDataDictionary.cs
```diff
@@ -395,37 +455,54 @@ public ViewDataInfo GetViewDataInfo(string expression)
...
-            // Reset or override ModelMetadata based on runtime value type. Fall back to declared type if value is
-            // null. When called from a constructor, current ModelExplorer may already be set to preserve
-            // customizations made in parent scope. But ModelExplorer is never null after instance is initialized.
-            var modelType = GetModelType(value);
-            Type metadataModelType = null;
-            if (ModelExplorer != null)
+            // Update ModelExplorer to reflect the new value. When possible, preserve ModelMetadata to avoid losing
+            // property information.
+            var modelType = value?.GetType();
+            if (ModelMetadata.MetadataKind == ModelMetadataKind.Type &&
```
dougbu: Note these conditions are now in the same order and identical to those in the copy-from-`source` constructor.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ControllerTest.cs
```diff
@@ -274,6 +276,48 @@ public void TempData_CanSetAndGetValues()
...
+                };
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(IncompatibleModelData))]
+        public void ViewDataModelSetter_DoesNotThrow(object model, Type expectedType)
+        {
+            // Arrange
+            var activator = new ViewDataDictionaryControllerPropertyActivator(new EmptyModelMetadataProvider());
```
dougbu: FYI, no, we didn't have any tests except the functional tests using `ViewDataDictionaryControllerPropertyActivator`.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/DefaultTemplatesUtilities.cs
```diff
@@ -335,6 +329,12 @@ private static ICompositeViewEngine CreateViewEngine()
             return viewEngine.Object;
         }
 
+        public static string FormatOutput(IHtmlHelper helper, object model)
+        {
+            var modelExplorer = helper.MetadataProvider.GetModelExplorerForType(model.GetType(), model);
+            return FormatOutput(modelExplorer);
+        }
+
```
dougbu: I could revert changes to this file but like leaving this method next to the other `FormatOutput()` overload.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ControllerTest.cs
```diff
@@ -274,6 +276,48 @@ public void TempData_CanSetAndGetValues()
             Assert.Equal(input, result);
         }
 
+        public static TheoryData<object, Type> IncompatibleModelData
+        {
+            get
+            {
+                // Small grab bag of instances and expected types with no common base except typeof(object).
```
dougbu: I could create a "for testing only" property exposing the `_declaredModelType` of a VDD. But the indirect tests changing the model type to a few different types seems sufficient. Any objection?

dougbu:
@rynowak: `ViewDataDictionary` has (and has had) a few almost-invariants (that are now a bit more visible in the code). Do you think the `ModelExplorer` setter should `throw` when the invariants below are violated? We don't even have a `null` check today.

1. `ModelExplorer` is never `null`
2. the declared model type can be assigned from `ModelMetadata.ModelType`
3. unless in a couple of special cases, `ModelMetadata.ModelType` can be assigned from `Model`
 1. a `ViewDataDictionary<{ValueType}>` where the `ValueType` is not `Nullable<T>` and `Model` has never been set is one special case. `Model` will remain `null` in that case.
 2. `@Html.Display...()` and `@Html.Editor...()` _may_ be a special case as well. I seem to remember `ModelExplorer` allows `Model` to be incompatible with the `ModelType` due to corner cases in the `ExpressionMetadataProvider` (where the `ModelType` ends up being `typeof(string)`) but I can't recall the details.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ViewComponents/ViewComponentContextTest.cs
```diff
@@ -0,0 +1,112 @@
...
+                    { "test string", typeof(string) },
+                    { new List<int>(), typeof(List<int>) },
+                    { new List<string>(), typeof(List<string>) },
+                };
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(IncompatibleModelData))]
+        public void ViewDataModelSetter_DoesNotThrow_IfValueIncompatibleWithSourceDeclaredType(
```
dougbu: This test fails for every item in the data set w/o the `src` fixes.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ViewFeatures/ViewDataDictionaryTest.cs
```diff
@@ -362,6 +475,59 @@ public void ModelSetter_SameType_UpdatesModelExplorer()
...
+                ModelExplorer = explorer,
+            };
+
+            // Act
+            var viewData = new ViewDataDictionary(source, model);
+
+            // Assert
+            Assert.NotNull(viewData.ModelExplorer);
+            Assert.NotSame(explorer, viewData.ModelExplorer);
+            Assert.Same(metadata, viewData.ModelMetadata);
```
dougbu: This assertion fails w/o the `src` changes. `viewData.ModelMetadata.ModelType` is `typeof(object)` when `model==null` and `typeof(int)` when `model==23`.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ViewFeatures/ViewDataDictionaryOfTModelTest.cs
```diff
@@ -223,6 +226,48 @@ public void CopyConstructors_ThrowInvalidOperation_IfModelIncompatible()
...
+                Model = -48,
+            };
+
+            // Act
+            var viewData = new ViewDataDictionary<int?>(source, model);
+
+            // Assert
+            Assert.NotNull(viewData.ModelExplorer);
+            Assert.NotSame(source.ModelExplorer, viewData.ModelExplorer);
+            Assert.Same(source.ModelMetadata, viewData.ModelMetadata);
```
dougbu: This assertion fails w/o the `src` changes when `model==23`. `viewData.ModelMetadata.ModelType` is `typeof(int)`.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ViewFeatures/ViewDataDictionaryTest.cs
```diff
@@ -253,6 +315,23 @@ public void CopyConstructorDoesNotThrowOnNullModel()
...
+            var metadataProvider = new EmptyModelMetadataProvider();
+            var source = new TestViewDataDictionary(metadataProvider, typeof(int));
+
+            // Act
+            var viewData = new ViewDataDictionary(source, model: null);
+
+            // Assert
+            Assert.Null(viewData.Model);
+            Assert.NotNull(viewData.ModelMetadata);
+            Assert.Same(source.ModelMetadata, viewData.ModelMetadata);
```
dougbu: This assertion fails w/o the `src` changes. `viewData.ModelMetadata.ModelType` is `typeof(object)`.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ViewFeatures/ViewDataDictionaryTest.cs
```diff
@@ -311,30 +390,64 @@ public void CopyConstructors_CopyModelMetadata_ForTypeObject()
...
+            // Act
+            var viewData1 = new ViewDataDictionary(source)
+            {
+                Model = model,
+            };
+            var viewData2 = new ViewDataDictionary(source, model: model);
+
+            // Assert
+            Assert.NotSame(source.ModelExplorer, viewData1.ModelExplorer);
+            Assert.Same(source.ModelMetadata, viewData1.ModelMetadata);
```
dougbu: This assertion fails for the `typeof(IEnumerable<string>), new string[] { "1", "2", "3" })` case w/o the `src` changes. `viewData1.ModelMetadata.ModelType` is `typeof(string[])` and the same for `viewData2`. Not exactly _wrong_ in this case (no property metadata was lost) but property metadata would be lost if it were available in `source.ModelMetadata`.

dougbu:
All tests that fail without my changes are now marked.

test/Microsoft.AspNetCore.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.CheckViewData.AtViewModel.html
```diff
@@ -0,0 +1,54 @@
...
+
+</div>
+<div class="row">
+    <h5>Check View Data view component</h5>
+<div class="col-md-3">MetadataKind: 'Type'</div>
+<div class="col-md-3">ModelType: 'SuperViewModel'</div>
+
+<h5>Check View Data view component's view</h5>
+<div class="col-md-3">MetadataKind: 'Type'</div>
+<div class="col-md-3">ModelType: 'SuperViewModel'</div>
```
dougbu: Without `src` changes, get `ModelType: 'ViewModel'` here. Minor case -- got metadata from the declared type rather than the instance.

test/Microsoft.AspNetCore.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.CheckViewData.AtViewModel.html
```diff
@@ -0,0 +1,54 @@
...
+<div class="col-md-3">ModelType: 'Nullable`1'</div>
+    <div class="col-md-3">PropertyName: 'NullableLong'</div>
+
+</div>
+<div class="row">
+    
+<h5>Template for TemplateModel</h5>
+<div class="col-md-3">MetadataKind: 'Property'</div>
+<div class="col-md-3">ModelType: 'TemplateModel'</div>
+    <div class="col-md-3">PropertyName: 'Template'</div>
```
dougbu: All property information is lost w/o `src` changes. Instead get
``` html
<div class="col-md-3">MetadataKind: 'Type'</div>
<div class="col-md-3">ModelType: 'SuperTemplateModel'</div>
```

test/Microsoft.AspNetCore.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.CheckViewData.NullViewModel.html
```diff
@@ -0,0 +1,54 @@
...
+<div class="col-md-3">MetadataKind: 'Type'</div>
+<div class="col-md-3">ModelType: 'ViewModel'</div>
+
+</div>
+<div class="row">
+    
+<h5>Template for Int32</h5>
+<div class="col-md-3">MetadataKind: 'Property'</div>
+<div class="col-md-3">ModelType: 'Int32'</div>
+    <div class="col-md-3">PropertyName: 'Integer'</div>
```
dougbu: All property information lost w/o `src` changes. Instead get
``` html
<div class="col-md-3">MetadataKind: 'Type'</div>
<div class="col-md-3">ModelType: 'Nullable`1'</div>
```

test/Microsoft.AspNetCore.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.CheckViewData.NullViewModel.html
```diff
@@ -0,0 +1,54 @@
...
+
+</div>
+<div class="row">
+    
+<h5>Template for Int64</h5>
+<div class="col-md-3">MetadataKind: 'Property'</div>
+<div class="col-md-3">ModelType: 'Nullable`1'</div>
+    <div class="col-md-3">PropertyName: 'NullableLong'</div>
```
dougbu: Type is correct but all other property information lost w/o `src` changes. Instead get
``` html
<div class="col-md-3">MetadataKind: 'Type'</div>
<div class="col-md-3">ModelType: 'Nullable`1'</div>
```

test/WebSites/HtmlGenerationWebSite/Components/CheckViewData - LackModel.cs
```diff
@@ -0,0 +1,42 @@
...
+            writer.WriteLine("<h5>Check View Data - LackModel view component</h5>");
+            writer.WriteLine($"<div class=\"col-md-3\">MetadataKind: '{ metadata.MetadataKind }'</div>");
+            writer.WriteLine($"<div class=\"col-md-3\">ModelType: '{ metadata.ModelType.Name }'</div>");
+            if (metadata.MetadataKind == ModelMetadataKind.Property)
+            {
+                writer.WriteLine($"<div class=\"col-md-3\">PropertyName: '{ metadata.PropertyName }'</div>");
+            }
+
+            // Confirm view component is able to set the model to anything.
+            ViewData.Model = 78.9;
```
dougbu: This line `throw`s an `InvalidOperationException` w/o `src` changes. The relevant `src` bit is the [one-line change](https://github.com/aspnet/Mvc/pull/4403/files#diff-9480b4257e8ee340006402a97f5afbb8R81) in `ViewComponentContext`.

dougbu:
Forgot about the functional tests last night. All tests that fail without my changes are now marked _for realz_.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ViewDataDictionary.cs
rynowak: :+1: good, that makes sense

rynowak: Is it worth trying to de-dupe this code?

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ControllerTest.cs
rynowak: I think what you've done here is fine, I don't know that we need to expose more from the VDD.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ControllerTest.cs
rynowak: I don't really get what this has to do with `Controller`. This really looks like two tests at the same time.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Rendering/DefaultTemplatesUtilities.cs
rynowak: :see_no_evil:

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ViewComponents/ViewComponentContextTest.cs
rynowak: Not super critical because integration, but this really seems like 2 tests in one.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ControllerTest.cs
dougbu: What would the two tests you're imagining do?

This test ensures the `ViewDataDictionary` exposed on `Controller` has no type restrictions i.e. the declared type is `typeof(object)`. Fits here (a) because we don't have a `ViewDataDictionaryControllerPropertyActivatorTest` class, (b) such a test class would need a `TestableController` and so on, and (c) it's about the `ViewData` property on `Controller`.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ViewDataDictionary.cs
dougbu: Thought about a `SetModelCore()` method but `SetModel()` and the main copy constructor each have unique branches in the logic. [One block](https://github.com/aspnet/Mvc/pull/4403/files/5fee305a5ddf8a0206e1945130290af389b4044f#diff-dd434cfe5fb0b1b08c4bdac8467917beR236) in that constructor could be done instead of calling `SetModelCore()` -- that's basically another branch in a large '||' condition. And could pass a `bool` into `SetModelCore()` for the [block below](https://github.com/aspnet/Mvc/pull/4403/files/5fee305a5ddf8a0206e1945130290af389b4044f#diff-dd434cfe5fb0b1b08c4bdac8467917beR497) if you feel strongly about it. LMK.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ViewComponents/ViewComponentContextTest.cs
dougbu: What are the two tests you're seeing? This is about ensuring the `ViewData` property on this class is instantiated w/o type restrictions.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ViewFeatures/ViewDataDictionaryOfTModelTest.cs
```diff
@@ -124,33 +124,36 @@ public void CopyConstructors_InitalizeModelAndModelMetadataBasedOnSource_ModelOf
...
-            Assert.Same(source.ModelMetadata, viewData2.ModelMetadata);
-            Assert.Equal(source.Count, viewData2.Count);
-            Assert.Equal("bar", viewData2["foo"]);
-            Assert.IsType<CopyOnWriteDictionary<string, object>>(viewData2.Data);
+        [Fact]
+        public void CopyConstructor_InitializesModelBasedOnSource_ModelMetadataBasedOnTModel()
+        {
+            // Arrange
+            var metadataProvider = new EmptyModelMetadataProvider();
+            var model = new SupremeTestModel();
```
rynowak: :heart_eyes_cat:

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ViewFeatures/ViewDataDictionaryOfTModelTest.cs
```diff
@@ -124,33 +124,36 @@ public void CopyConstructors_InitalizeModelAndModelMetadataBasedOnSource_ModelOf
...
+            var source = new ViewDataDictionary(metadataProvider)
+            {
+                Model = model,
+            };
+            source["foo"] = "bar";
+            source.TemplateInfo.HtmlFieldPrefix = "prefix";
```
rynowak: are these two lines needed for anything?

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ViewFeatures/ViewDataDictionaryOfTModelTest.cs
```diff
@@ -223,6 +226,48 @@ public void CopyConstructors_ThrowInvalidOperation_IfModelIncompatible()
...
+            Assert.NotNull(viewData.ModelExplorer);
+            Assert.NotSame(source.ModelExplorer, viewData.ModelExplorer);
+            Assert.Same(source.ModelMetadata, viewData.ModelMetadata);
+            Assert.Equal(typeof(int?), viewData.ModelMetadata.ModelType);
+            Assert.Equal(viewData.Model, viewData.ModelExplorer.Model);
+            Assert.Equal(model, viewData.Model);
+        }
+
+        [Fact]
+        public void CopyConstructor_UpdatesMetadata_IfDeclaredTypeChangesIncompatibly()
```
rynowak: Should we also have a case here where the `Model` is set and the type changes to something incompatible.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ControllerTest.cs
rynowak: Test 1: construct a controller and look at the metadata/type of the VDD
Test 2: construct a VDD matching what controller would have a set the model.

I think it's fine if you want to leave this here as an integration test, it just seems like it's covering a few different things.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ViewFeatures/ViewDataDictionaryTest.cs
```diff
@@ -50,33 +49,46 @@ public void ConstructorInitalizesMembers()
...
             // Assert
             Assert.NotNull(viewData.ModelMetadata);
-            metadataProvider.Verify();
+            metadataProvider.Verify(m => m.GetMetadataForType(typeof(TestModel)), Times.Once());
         }
 
-        // When SetModel is called, only GetMetadataForType from MetadataProvider is expected to be called.
         [Fact]
-        public void SetModelCallsGetMetadataForTypeExactlyOnce()
+        public void SetModel_UsesModelMetadataProvider_IfSourceTypeIsObject()
```
rynowak: Can we say something like "GetsNewMetadata" instead of "UsesModelMetadataProvider"?

rynowak:
:shipit: looks solid, I had a few comments but they are all nitpicks on tests and comments and whatnot

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ViewFeatures/ViewDataDictionaryOfTModelTest.cs
dougbu: Why no kind sir. :+1:

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ViewFeatures/ViewDataDictionaryOfTModelTest.cs
dougbu: :+1: should already exist because it's a core part of how `@model MyType` absolutely must work. If the case isn't tested, I'll do the needful.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ViewFeatures/ViewDataDictionaryTest.cs
dougbu: :ok:

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ControllerTest.cs
dougbu: :see_no_evil: because "test 1" determining the type isn't meaningful -- the type is `ViewDataDictionary` and not `ViewDataDictionary<object>`. IOW would need to expose `_declaredModelType` to write test 1 alone. So I'm leaving this as an integration test.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ViewDataDictionary.cs
dougbu: :see_no_evil:

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ViewComponents/ViewComponentContextTest.cs
dougbu: :see_no_evil:

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/ViewFeatures/ViewDataDictionaryOfTModelTest.cs
dougbu: Test already exists: `CopyConstructors_ThrowInvalidOperation_IfModelIncompatibleWithDeclaredType()`. I renamed it just above.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ViewDataDictionary.cs
dougbu: Decided offline to burn this w/ a small match and a large amount of napalm.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ViewDataDictionary.cs
dougbu: Decided offline to leave this alone because the same constructor existed in MVC 5.x and it does what it says on the tin.

dougbu:
:up::date: FYI

Working on squishing, squashing and moving to `release` branch since this is approved for RC2...

dougbu:
a0c8834 and merged into `dev`

Issue #4407, opened 2016-04-05
Occasional hanging test ProcessAsync_ExceptionInProcessing_DoesntBlockConcurrentRequests


JunTaoLuo:
I am seeing occasional hangs of this test https://github.com/aspnet/Mvc/blob/dev/test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/DistributedCacheTagHelperTest.cs#L606 on Linux (http://aspnetci/viewLog.html?buildId=79739&buildTypeId=XPlat_Linux_UbuntuUniverseCoreCLR&tab=buildLog#_focus=80061) and have been able to reproduce it on a local machine. 

I'm not too sure what the test does or why it's hanging. @sebastienros I see you have been fixing races related to cache tag helpers, care to take a look?

sebastienros:
I should take care of this. When you say locally, you still mean Linux?

JunTaoLuo:
Yes I reproduced it on my local linux vm.

BrennanConroy:
This is also happening on Windows
http://aspnetci/viewLog.html?buildId=80003&buildTypeId=XPlat_Windows_Win10Universe&tab=buildLog#_focus=97433

sebastienros:
Thanks, that helps. I already have a fix locally that I am testing, but it should just be an issue with the unit test itself, not the feature. The test is trying to test a specific multi-thread scenario and it appears it can go in a wrong state. I have already fixed a similar issue on the same unit test, apparently it's not enough.

PR #4415, opened 2016-04-06
Fixing race in distributed cache tag helper

sebastienros:
If an exception is thrown by an instance, another awaiting request could grab the worker task, and get into an infinite loop.
The fix removes the worker task before warning other requests to acquire another worker task or process it themselves.

Checked with @JunTaoLuo on his box it couldn't be repro-ed with the fix, but currently running it 100 times to be sure.


sebastienros:
#4407

JunTaoLuo:
Ran 100 iterations without hangs. I think this is good to go.

dougbu:
:shipit: discussed offline w/ @sebastienros and agree this is ready to go.

PR #4421, opened 2016-04-06
Remove using Task.FromResult where possible

rynowak:
Believe it or not, this one in ActionResult is showing in TechEmpower. I
did a pass on all of the product code since the cache task actually
improves the readability of the code :sunglasses:

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/ActionResult.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/ActionResult.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Threading.Tasks;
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.Mvc.Core/ActionResult.cs
```diff
@@ -22,7 +23,7 @@ public abstract class ActionResult : IActionResult
         public virtual Task ExecuteResultAsync(ActionContext context)
         {
             ExecuteResult(context);
-            return Task.FromResult(true);
```
rynowak: 80 bytes is too many

rynowak: This is actually around 8% of the allocated bytes in TechEmpower plaintext

benaadams:
:rocket:

dougbu:
:shipit:

Issue #4376, opened 2016-03-30
Lazy-Initialize HashSets in TempDataDictionary


rynowak:
For small-app scenarios that don't use TempData the dictionary is going to be instantiated but not used. We can save two hashsets per request by making them lazy.

Allocation data from HelloWorldMvc:
![image](https://cloud.githubusercontent.com/assets/1430011/14147183/4b58438c-f64f-11e5-9105-c94f3ed58479.png)

rynowak:
Poaching since I found 5 mins to fix this

rynowak:
d4c7dc3a834edc507865816a59960c5b31f9b10c

PR #4424, opened 2016-04-07
Make HashSets in TempData lazy

rynowak:
TempData is frequently created but not modified, this reduces the cost of
that without much complexity. #4376

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/TempDataDictionary.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/TempDataDictionary.cs
@@ -11,12 +11,14 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
     /// <inheritdoc />
     public class TempDataDictionary : ITempDataDictionary
     {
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/TempDataDictionary.cs
```diff
@@ -11,12 +11,14 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
...
+        // Perf: Everything here is lazy because the TempDataDictionary is frequently created and passed around
+        // without being manipulated.
         private Dictionary<string, object> _data;
         private bool _loaded;
         private readonly ITempDataProvider _provider;
         private readonly HttpContext _context;
-        private HashSet<string> _initialKeys = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
-        private HashSet<string> _retainedKeys = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+        private HashSet<string> _initialKeys;
+        private HashSet<string> _retainedKeys;
```
rynowak: This is safe because every actual method either calls `Load()` or checks `_isLoaded` and no-ops.

ajaybhargavb:
:shipit:

Issue #3659, opened 2015-11-28
`HtmlLocalizer` and `ViewLocalizer` mishandle arguments; pass `IHtmlContent` values to HTML encoder


dougbu:
MVC has moved away from bogus `ToString` overrides but the `HtmlLocalizer` implementation has not reacted to this change. The result is twofold:
- Many `IHtmlContent` values display their type names when used as `IHtmlLocalizer.Html()` or `IViewLocalizer.Html()` arguments. For example this occurs with view component results e.g. `@Localizer.Html("Hello {0}!", Component.Invoke("CurrentUser"))`.
- Arguments that are already HTML encoded and avoid the above issue are double-encoded. This occurs with `HtmlString` values passed as arguments to `IHtmlLocalizer.Html()` or `IViewLocalizer.Html()` e.g. `@Localizer.Html("Not that {0}!", new HtmlString("&lt;input&gt;"))`.

dougbu:
The issue here is likely the lack of `IHtmlContent` special-casing anywhere in `HtmlLocalizer`, especially [line 268](https://github.com/aspnet/Mvc/blob/490fcf1ab4d416ee6cc140797c2377e352677bda/src/Microsoft.AspNet.Mvc.Localization/HtmlLocalizer.cs#L268).

Eilon:
Can you clarify the scenario that doesn't work? What's broken?

dougbu:
Pretty much any argument implementing `IHtmlContent` won't work as expected. E.g.
- `@Localizer.Html("Hello {0}!", Component.Invoke("CurrentUser"))` generates `Hello Microsoft.AspNet.Mvc.ViewFeatures.StringCollectionTextWriter&#x2B;StringCollectionTextWriterContent!`
- `@Localizer.Html("Not that {0}!", new HtmlString("&lt;input&gt;"))` generates `Not that &amp;lt;input&amp;gt;!`.

In the above examples, `Localizer` is an injected `IViewLocalizer`. No matching resource but that's not relevant in this issue.

Eilon:
Well, `@Localizer.Html("Hello {0}!", Component.Invoke("CurrentUser"))` isn't even a thing. Not sure why anyone would even expect that to work (regardless of whether it *should* work).

But the second example looks legit and should work.

rynowak:
Load balancing for RC2

rynowak:
a55544681dfb8f23870c2144ed5618a41a1580fb

PR #4432, opened 2016-04-08
Fix for #3659 - handle IHtmlContent in localization

rynowak:
This change corrects handling of IHtmlContent instances in
LocalizedHtmlString by making using of FormattableHtmlString.


rynowak:
This change is for RC2 - ignore the fact that the compare branch is set to `dev`

src/Microsoft.AspNetCore.Mvc.Localization/LocalizedHtmlString.cs
```diff
@@ -100,117 +100,8 @@ public void WriteTo(TextWriter writer, HtmlEncoder encoder)
...
-        private static void AppendToOutput(
-            StringBuilder tokenBuffer,
-            object[] arguments,
-            TextWriter writer,
-            HtmlEncoder encoder)
-        {
-            if (tokenBuffer != null && tokenBuffer.Length > 0)
-            {
-                encoder.Encode(writer, string.Format(tokenBuffer.ToString(), arguments));
-            }
```
rynowak: :wave:

dougbu:
:shipit:

Issue #4435, opened 2016-04-09
Many unused instances of ModelErrorCollection during model validation


DavidObando:
Class `ModelStateDictionary` initializes a root `ModelStateNode` and then creates new child nodes during validation of a model. Class `ModelStateNode` inherits `ModelStateEntry` which has a non-null complex field of type `ModelErrorCollection`. This field is non-null as it's initialized on the constructor of `ModelStateEntry`, which is unnecessary.

We should lazily initialize the field with a ModelErrorCollection only when required. Each allocation spends 72 bytes. After only 1500 API calls we're already allocating and collecting 108,000 bytes.

dougbu:
`ModelStateDictionary` checks `Errors.Count` on `ModelStateNode` multiple times. Would be surprising if lazy initialization saved many `ModelErrorCollection` instances. Have you measured that part of this PR on its own?

DavidObando:
Yes, the count of elements went down to zero in my memory profile. I'll add more details when I get back to my workstation.

DavidObando:
@dougbu @rynowak I ran a test of 1500 POST calls with JSON payloads ranging between 14 KB and 500 KB. Here's the instance count of types with the text "Error" in it, before the change:

![image](https://cloud.githubusercontent.com/assets/950290/14436404/60174350-ffd1-11e5-9170-87b82407d1d7.png)

This is what I see after the change:

![image](https://cloud.githubusercontent.com/assets/950290/14436412/69c537ae-ffd1-11e5-8fff-e9de952c05ea.png)

I'm currently working to get this test into the Performance repository.

rynowak:
So 1 instance allocated/saved per request for what you're measuring :laughing:. I think you'd have to look at a model-binding (non-JSON) scenario to get any significant data.

My opinion: go ahead with the change you proposed, but leave the bug open. I think we need to measure more scenarios.

DavidObando:
Yes, 1 per request. It's not much, but it's an easy gain to put less pressure on the GC. I'll continue with the PR, leave this item open, and will sync up with you later offline regarding the additional scenarios you mentioned. Thanks!

Eilon:
Moving to post-v1; can do this fix at any time.

PR #4436, opened 2016-04-09
Reduce the number of allocations during model validation

### LINKED ISSUE #4434: DefaultCollectionValidationStrategy allocates RuntimeMethodInfos in function of the data being validated
DefaultCollectionValidationStrategy allocates RuntimeMethodInfos in function of the data being validated. That means that if we're sending a large payload, or large amounts of payloads, or a combination of both, where the data being validated contains collections, we'll allocate a very large amount of RuntimeMethodInfo objects.

This is caused by code that creates a new MethodInfo that calls GetEnumerable<T> for each collection instance in the data (see [here](https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultCollectionValidationStrategy.cs#L65)), as opposed to once per each collection type.

Given that DefaultCollectionValidationStrategy is a singleton, and that the amount of runtime types that will be used as part of the data model being validated is limited and not bound to grow past a few dozen, we should cache this instead of relying on the runtime type cache which allocates unnecessarily.

DavidObando:
When the service receives a model (say, via a POST message) MVC validates it to ensure the model is in a correct state. Validation currently incurs in many allocations that can be avoided. This tackles two of them:
  1. We're now caching the generic `GetEnumerator<T>` method infos generated on the fly during collection validation, and
  2. We're now only initializing `ModelErrorCollection` on demand.

The first one incurs in the additional allocation of 1 long-lived dictionary object, which will grow only to the amount of `Collection<T>` types used by the model being validated. This is expected to be a small to medium number.

The second change assumes that class `ModelStateEntry` isn't thread safe, as model validation isn't multithreaded.

This resolves #4434 and #4435.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelStateEntry.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelStateEntry.cs
@@ -10,6 +10,7 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
     /// </summary>
     public abstract class ModelStateEntry
     {
+        private ModelErrorCollection _errors;
         /// <summary>
         /// Gets the raw value from the request associated with this entry.
         /// </summary>
@@ -23,7 +24,17 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
         /// <summary>
         /// Gets the <see cref="ModelErrorCollection"/> for this entry.
         /// </summary>
-        public ModelErrorCollection Errors { get; } = new ModelErrorCollection();
+        public ModelErrorCollection Errors
+        {
+            get
+            {
+                if (_errors == null)
+                {
+                    _errors = new ModelErrorCollection();
+                }
+                return _errors;
+            }
+        }
 
         /// <summary>
         /// Gets or sets the <see cref="ModelValidationState"/> for this entry.
--- a/src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultCollectionValidationStrategy.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultCollectionValidationStrategy.cs
@@ -1,11 +1,14 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Reflection;
 using Microsoft.AspNetCore.Mvc.ModelBinding;
 using Microsoft.AspNetCore.Mvc.ModelBinding.Validation;
+using System.Linq.Expressions;
 
 namespace Microsoft.AspNetCore.Mvc.Internal
 {
@@ -44,7 +47,8 @@ namespace Microsoft.AspNetCore.Mvc.Internal
         /// <summary>
         /// Gets an instance of <see cref="DefaultCollectionValidationStrategy"/>.
         /// </summary>
-        public static readonly IValidationStrategy Instance = new DefaultCollectionValidationStrategy();
+        public static readonly DefaultCollectionValidationStrategy Instance = new DefaultCollectionValidationStrategy();
+        private readonly ConcurrentDictionary<Type, Func<object, IEnumerator>> _genericGetEnumeratorCache = new ConcurrentDictionary<Type, Func<object, IEnumerator>>();
 
         private DefaultCollectionValidationStrategy()
         {
@@ -60,10 +64,21 @@ namespace Microsoft.AspNetCore.Mvc.Internal
             return new Enumerator(metadata.ElementMetadata, key, enumerator);
         }
 
-        public static IEnumerator GetEnumeratorForElementType(ModelMetadata metadata, object model)
+        public IEnumerator GetEnumeratorForElementType(ModelMetadata metadata, object model)
         {
-            var getEnumeratorMethod = _getEnumerator.MakeGenericMethod(metadata.ElementType);
-            return (IEnumerator)getEnumeratorMethod.Invoke(null, new object[] { model });
+            Func<object, IEnumerator> getEnumerator = _genericGetEnumeratorCache.GetOrAdd(
+                key: metadata.ElementType,
+                valueFactory: (type) => {
+                    var getEnumeratorMethod = _getEnumerator.MakeGenericMethod(type);
+                    var parameter = Expression.Parameter(typeof(object), "model");
+                    var expression =
+                        Expression.Lambda<Func<object, IEnumerator>>(
+                            Expression.Call(null, getEnumeratorMethod, parameter),
+                            parameter);
+                    return expression.Compile();
+                });
+
+            return getEnumerator(model);
         }
 
         // Called via reflection.
--- a/src/Microsoft.AspNetCore.Mvc.Core/Internal/ExplicitIndexCollectionValidationStrategy.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Internal/ExplicitIndexCollectionValidationStrategy.cs
@@ -56,7 +56,7 @@ namespace Microsoft.AspNetCore.Mvc.Internal
             string key,
             object model)
         {
-            var enumerator = DefaultCollectionValidationStrategy.GetEnumeratorForElementType(metadata, model);
+            var enumerator = DefaultCollectionValidationStrategy.Instance.GetEnumeratorForElementType(metadata, model);
             return new Enumerator(metadata.ElementMetadata, key, ElementKeys, enumerator);
         }
 


src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultCollectionValidationStrategy.cs
```diff
@@ -45,6 +46,7 @@ public class DefaultCollectionValidationStrategy : IValidationStrategy
         /// Gets an instance of <see cref="DefaultCollectionValidationStrategy"/>.
         /// </summary>
         public static readonly IValidationStrategy Instance = new DefaultCollectionValidationStrategy();
+        private static readonly Dictionary<Type, MethodInfo> _genericGetEnumeratorCache = new Dictionary<Type, MethodInfo>();
```
dougbu: - Why is this `static` vs. a field inside the single `static` instance of `DefaultCollectionValidationStrategy`?
- Suggest this needs to be at least a `ConcurrentDictionary` before it can be shared across threads.

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultCollectionValidationStrategy.cs
```diff
@@ -62,7 +64,12 @@ private DefaultCollectionValidationStrategy()
 
         public static IEnumerator GetEnumeratorForElementType(ModelMetadata metadata, object model)
         {
-            var getEnumeratorMethod = _getEnumerator.MakeGenericMethod(metadata.ElementType);
+            MethodInfo getEnumeratorMethod;
+            if(!_genericGetEnumeratorCache.TryGetValue(metadata.ElementType, out getEnumeratorMethod))
```
dougbu: Spacing: `if (`

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelStateEntry.cs
```diff
@@ -23,7 +24,17 @@ public abstract class ModelStateEntry
...
+        {
+            get
+            {
+                if(_errors == null)
+                {
+                    _errors = new ModelErrorCollection();
+                }
+                return _errors;
+            }
+        }
```
dougbu: As mentioned in #4435, I doubt lazy-initialization saves much here.

rynowak: No, it's reasonable savings on larger models (from data I've collected in the past) and it doesn't add much complexity.

rynowak:
:watch:

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultCollectionValidationStrategy.cs
DavidObando: @dougbu I made the cache static as it's used from `GetEnumeratorForElementType()` which is itself static. I guess I could access it through the singleton, with something along the lines of:
```C#
public static IEnumerator GetEnumeratorForElementType(ModelMetadata metadata, object model)
{
    MethodInfo getEnumeratorMethod;
    var instance = (DefaultCollectionValidationStrategy)Instance; 
    if (!instance._genericGetEnumeratorCache.TryGetValue(metadata.ElementType, out getEnumeratorMethod))
    {
        getEnumeratorMethod = _getEnumerator.MakeGenericMethod(metadata.ElementType);
        instance._genericGetEnumeratorCache.Add(metadata.ElementType, getEnumeratorMethod);
    }
    return (IEnumerator)getEnumeratorMethod.Invoke(null, new object[] { model });
}
```

Is that a reasonable approach?

I agree on the `ConcurrentDictionary` suggestion, will send that change soon.

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelStateEntry.cs
dougbu: @DavidObando @rynowak and I talked offline and both need to see numbers demonstrating this change improves anything. If (as I expect due to numerous reads of this property) it doesn't help, need to change more in order to avoid these allocations.

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultCollectionValidationStrategy.cs
dougbu: Just make `GetEnumeratorForElementType()` non-`static`. This is an `.Internal` class and such minor changes are fair game.

DavidObando: Will do, thanks!

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelStateEntry.cs
DavidObando: @dougbu @rynowak I ran a test of 1500 POST calls with JSON payloads ranging between 14 KB and 500 KB. Here's the instance count of types with the text "Error" in it, before the change:

image

This is what I see after the change:

image

I'm currently working to get this test into the Performance repository.

rynowak:
@DavidObando - make sure to leave a comment when you update a PR or else no one will notice :laughing:

src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelStateEntry.cs
```diff
@@ -23,7 +24,17 @@ public abstract class ModelStateEntry
         /// <summary>
         /// Gets the <see cref="ModelErrorCollection"/> for this entry.
         /// </summary>
-        public ModelErrorCollection Errors { get; } = new ModelErrorCollection();
+        public ModelErrorCollection Errors
+        {
+            get
+            {
+                if(_errors == null)
```
rynowak: fix spacing

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultCollectionValidationStrategy.cs
```diff
@@ -60,9 +63,16 @@ private DefaultCollectionValidationStrategy()
...
-            var getEnumeratorMethod = _getEnumerator.MakeGenericMethod(metadata.ElementType);
+            MethodInfo getEnumeratorMethod;
+            if (!_genericGetEnumeratorCache.TryGetValue(metadata.ElementType, out getEnumeratorMethod))
+            {
+                getEnumeratorMethod = _genericGetEnumeratorCache.AddOrUpdate(
+                        key: metadata.ElementType,
+                        addValueFactory: (type) => _getEnumerator.MakeGenericMethod(type),
+                        updateValueFactory: (type, methodInfo) => methodInfo);
+            }
             return (IEnumerator)getEnumeratorMethod.Invoke(null, new object[] { model });
```
rynowak: Is this going to go far enough? This is still a reflection invoke. Might be another thing to look into after this change

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultCollectionValidationStrategy.cs
```diff
@@ -60,9 +63,16 @@ private DefaultCollectionValidationStrategy()
...
+        public IEnumerator GetEnumeratorForElementType(ModelMetadata metadata, object model)
         {
-            var getEnumeratorMethod = _getEnumerator.MakeGenericMethod(metadata.ElementType);
+            MethodInfo getEnumeratorMethod;
+            if (!_genericGetEnumeratorCache.TryGetValue(metadata.ElementType, out getEnumeratorMethod))
+            {
+                getEnumeratorMethod = _genericGetEnumeratorCache.AddOrUpdate(
+                        key: metadata.ElementType,
+                        addValueFactory: (type) => _getEnumerator.MakeGenericMethod(type),
+                        updateValueFactory: (type, methodInfo) => methodInfo);
```
rynowak: Is this better than `TryGetValue` followed by `TryAdd(key, value)`? Seems simpler that way.

rynowak:

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultCollectionValidationStrategy.cs
davidfowl: Yea, feels like a case where we should compile another expression tree. We're pretty good at that now 😄

rynowak: We might want to think more carefully about the issue that this code solves and see if there's a better way.

This is for cases where a type implements multiple kinds of `IEnumerator<T>`. We've already decided what we think `T` is, but there can be problems calling the non-generic `GetEnumerator` because we don't know which enumerator it's going to return.... Maybe we want to no-op if the results don't match our expectations....

We hit this case with `JObject`.

DavidObando: I'm currently testing the performance of changing this to an expression tree.

@rynowak I'm failing to understand if your comment is related to the cache we're introducing, or the overall design on how we invoke `GetEnumerator`. The code is only caching a call to a method that knows the type of `T` and operates on a live instance:
```C#
// Called via reflection.
private static IEnumerator GetEnumerator<T>(object model)
{
    return (model as IEnumerable<T>)?.GetEnumerator() ?? ((IEnumerable)model).GetEnumerator();
}
```

Are you saying that we can expect a situation in which the `metadata.ElementType` used to define `T` is inconsistent to the type obtained by `((IEnumerable)model).GetEnumerator()`? If so, is the cache affecting this situation?

Sorry if my questioning is too basic.

rynowak: My comment is related to the overall design.

> Are you saying that we can expect a situation in which the metadata.ElementType used to define T is inconsistent to the type obtained by ((IEnumerable)model).GetEnumerator()

Yes, this the exactly the reason why this method (and the perf problem) exist.

> If so, is the cache affecting this situation?

Nope

DavidObando: Thanks! I'll add a separate bug to track this.

DavidObando:
:up::date:

The change from reflection to compiled expression tree, and the simpler call to `ConcurrentDictionary.TryAdd` gave us a nice performance boost. In my local test, the cost of `GetEnumeratorForElementType` went from 6167 ms (153 ms own time) to 348 ms (53 ms own time).

The test consisted of 1500 POST calss with medium sized JSON payloads.

All tests passing.

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultCollectionValidationStrategy.cs
```diff
@@ -60,10 +64,22 @@ private DefaultCollectionValidationStrategy()
             return new Enumerator(metadata.ElementMetadata, key, enumerator);
         }
 
-        public static IEnumerator GetEnumeratorForElementType(ModelMetadata metadata, object model)
+        public IEnumerator GetEnumeratorForElementType(ModelMetadata metadata, object model)
         {
-            var getEnumeratorMethod = _getEnumerator.MakeGenericMethod(metadata.ElementType);
-            return (IEnumerator)getEnumeratorMethod.Invoke(null, new object[] { model });
+            Func<object, IEnumerator> getEnumerator;
+            if (!_genericGetEnumeratorCache.TryGetValue(metadata.ElementType, out getEnumerator))
```
davidfowl: Is it better to do GetOrAdd to avoid double hashing `metadata.ElementType`

rynowak:
@DavidObando was this merged?

rynowak:
Thanks :+1:

PR #4446, opened 2016-04-11
Repair failing functional test

rynowak:
I recently consolidated some code in routing that made the behavior for
attribute routes and conventional routes more consistent. So now, a
catch-all will add a null value to the RVD if it doesn't capture any text,
but will NOT trounce an existing value in the RVD.

This test depended on the behavior where attribute routes didn't ever add
a null value for a catch-all. The fix is to just add a default to the
catch-all.


rynowak:
FYI only - will merge to unbreak a failing test

PR #4448, opened 2016-04-11
Fix #4447 - Teach MVC to understand defaults

### LINKED ISSUE #4447: MVC not honoring attribute route default values for URL matching
See: https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/Internal/AttributeRoute.cs#L14

MVC isn't propagating the default values from a URL template. That means that code like:
```
        [HttpGet("/TeamName/{*Name=DefaultName}/")]
        public ActionResult GetTeam(string name)
        {
        }
```

will match `/TeamName` but will always have a default value of `null` for `name`

The fix here to pass `routeInfo.Defaults` + the secret route token.

rynowak:
The wireup for defaults in attribute routes was missing

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Abstractions/Abstractions/ActionDescriptor.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Abstractions/Abstractions/ActionDescriptor.cs
@@ -29,7 +29,7 @@ namespace Microsoft.AspNetCore.Mvc.Abstractions
 
         public AttributeRouteInfo AttributeRouteInfo { get; set; }
 
-        public IDictionary<string, object> RouteValueDefaults { get; }
+        public IDictionary<string, object> RouteValueDefaults { get; set; }
 
         /// <summary>
         /// The set of constraints for this action. Must all be satisfied for the action to be selected.
@@ -53,6 +53,6 @@ namespace Microsoft.AspNetCore.Mvc.Abstractions
         /// <summary>
         /// Stores arbitrary metadata properties associated with the <see cref="ActionDescriptor"/>.
         /// </summary>
-        public IDictionary<object, object> Properties { get; }
+        public IDictionary<object, object> Properties { get; set; }
     }
 }
--- a/src/Microsoft.AspNetCore.Mvc.Core/Internal/AttributeRoute.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Internal/AttributeRoute.cs
@@ -98,23 +98,28 @@ namespace Microsoft.AspNetCore.Mvc.Internal
             // it on startup.
             if (_router == null || _router.Version != actions.Version)
             {
-                _router = BuildRoute(actions);
+                var entries = GetEntries(actions);
+                _router = BuildRoute(entries, actions.Version);
             }
 
             return _router;
         }
 
-        private TreeRouter BuildRoute(ActionDescriptorCollection actions)
+        // internal for testing
+        internal AttributeRouteEntries GetEntries(ActionDescriptorCollection actions)
         {
-            var routeBuilder = new TreeRouteBuilder(_target, _loggerFactory);
+            var entries = new AttributeRouteEntries();
+
             var routeInfos = GetRouteInfos(_constraintResolver, actions.Items);
 
-            // We're creating one AttributeRouteGenerationEntry per action. This allows us to match the intended
+            // We're creating one TreeRouteLinkGenerationEntry per action. This allows us to match the intended
             // action by expected route values, and then use the TemplateBinder to generate the link.
             foreach (var routeInfo in routeInfos)
             {
-                routeBuilder.Add(new TreeRouteLinkGenerationEntry()
+                entries.LinkGenerationEntries.Add(new TreeRouteLinkGenerationEntry()
                 {
+                    // Using routeInfo.Defaults here WITHOUT adding the RouteGroupKey. We don't want to impact the
+                    // behavior of link generation.
                     Binder = new TemplateBinder(_urlEncoder, _contextPool, routeInfo.ParsedTemplate, routeInfo.Defaults),
                     Defaults = routeInfo.Defaults,
                     Constraints = routeInfo.Constraints,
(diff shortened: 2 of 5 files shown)


rynowak:
FYI this is an RC2 release bug - ignore the fact that I based the PR on dev.

src/Microsoft.AspNetCore.Mvc.Abstractions/Abstractions/ActionDescriptor.cs
```diff
@@ -53,6 +53,6 @@ public ActionDescriptor()
         /// <summary>
         /// Stores arbitrary metadata properties associated with the <see cref="ActionDescriptor"/>.
         /// </summary>
-        public IDictionary<object, object> Properties { get; }
```
rynowak: Made these get/set for unit testing. It's consistent with everything else in this class

src/Microsoft.AspNetCore.Mvc.Core/Internal/AttributeRoute.cs
```diff
@@ -133,24 +138,44 @@ private TreeRouter BuildRoute(ActionDescriptorCollection actions)
...
             foreach (var routeInfo in distinctRouteInfosByGroup)
             {
-                routeBuilder.Add(new TreeRouteMatchingEntry()
+                // Note that because we only support 'inline' defaults, each routeInfo group also has the same
+                // set of defaults. 
+                //
+                // We then inject the route group as a default for the matcher so it gets passed back to MVC
+                // for use in action selection.
+                var defaults = new RouteValueDictionary(routeInfo.Defaults);
+                defaults[TreeRouter.RouteGroupKey] = routeInfo.RouteGroup;
```
rynowak: This is the only new code. The rest is just a refactor for testability

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/AttributeRouteTest.cs
```diff
@@ -118,5 +103,513 @@ public class AttributeRouteTest
...
+                    Assert.Same(handler, e.Target);
+                    Assert.Collection(
+                        e.TemplateMatcher.Defaults.OrderBy(kvp => kvp.Key),
+                        kvp => Assert.Equal(new KeyValuePair<string, object>(TreeRouter.RouteGroupKey, "1"), kvp));
+                    Assert.Same(e.RouteTemplate, e.TemplateMatcher.Template);
+                });
+        }
+
+        [Fact]
+        public void AttributeRoute_GetEntries_CreatesMatchingEntry_WithDefault()
```
rynowak: This is the test that verifies the fix (along with the tweaked functional test). There are tests for defaults in attribute routes, but they life in the routing repo. What was missing here was the correct wire-up of values.

src/Microsoft.AspNetCore.Mvc.Core/Internal/AttributeRoute.cs
```diff
@@ -133,24 +138,44 @@ private TreeRouter BuildRoute(ActionDescriptorCollection actions)
             var distinctRouteInfosByGroup = GroupRouteInfosByGroupId(routeInfos);
             foreach (var routeInfo in distinctRouteInfosByGroup)
             {
-                routeBuilder.Add(new TreeRouteMatchingEntry()
+                // Note that because we only support 'inline' defaults, each routeInfo group also has the same
+                // set of defaults. 
+                //
+                // We then inject the route group as a default for the matcher so it gets passed back to MVC
+                // for use in action selection.
+                var defaults = new RouteValueDictionary(routeInfo.Defaults);
```
pranavkm: Does this need a `StringComparer.OrdinalIgnoreCase` comparer?

rynowak: RVD is always `OrdinalIgnoreCase `

pranavkm:
:shipit:

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/AttributeRouteTest.cs
```diff
@@ -118,5 +103,513 @@ public class AttributeRouteTest
...
+                    Assert.Empty(e.Constraints);
+                    Assert.Empty(e.Defaults);
+                    Assert.Equal(RoutePrecedence.ComputeGenerated(e.Template), e.GenerationPrecedence);
+                    Assert.Equal("BLOG_INDEX", e.Name);
+                    Assert.Equal(17, e.Order);
+                    Assert.Equal(actions[0].RouteValueDefaults, e.RequiredLinkValues);
+                    Assert.Equal("1", e.RouteGroup);
+                    Assert.Equal("api/Blog/{id}", e.Template.TemplateText);
+                },
+                e =>
```
javiercn: Is this repeated?

test/WebSites/RoutingWebSite/Controllers/TeamController.cs
```diff
@@ -63,8 +63,8 @@ public ActionResult GetAllTeams(int notRelevant)
             return Content(Url.Action(), "text/plain");
         }
 
-        [HttpGet("/TeamName/{*Name}/")]
-        public ActionResult GetTeam(string name = "DefaultName")
```
javiercn: This is not supported?

javiercn:
:shipit:

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/AttributeRouteTest.cs
rynowak: There are supposed to be two entries.

test/WebSites/RoutingWebSite/Controllers/TeamController.cs
rynowak: It won't be in RTM

rynowak: This is inconsistent with how conventional routes work with catchalls and inconsistent with our model binding works.

PR #4452, opened 2016-04-12
Fixing functional test random failure

sebastienros:
This code is waiting 1 second before requesting the same view/cached tag helper:

```html
<cache expires-after="TimeSpan.FromSeconds(1)" priority="CacheItemPriority.Low">
Cached content for @ViewBag.ProductId
</cache>
```

which expires after 1 second itself. Sometimes is returns the cached value where it expects a new value. I am proposing to wait for 2 seconds instead.


pranavkm:
:shipit:

NTaylorMullen:
:shipit:

JunTaoLuo:
Should this fix be merged to dev too?

sebastienros:
Done

Issue #3851, opened 2016-01-02
Add doc comments for filter attributes (e.g. RequireHttpsAttribute)


Eilon:
[RequireHttpsAttribute](https://github.com/aspnet/Mvc/blob/43226fe54de140db30746336f3d46130be47c0e4/src/Microsoft.AspNet.Mvc.Core/RequireHttpsAttribute.cs) is definitely missing it. Not sure about other attributes.

dougbu:
eda647e into `release`
Also merged into `dev`

PR #4455, opened 2016-04-12
Complete the doc comments for filters

dougbu:
- #3851
- add lots of words
- correct that vs. which confusion

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Abstractions/Filters/IFilterProvider.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Abstractions/Filters/IFilterProvider.cs
@@ -4,7 +4,8 @@
 namespace Microsoft.AspNetCore.Mvc.Filters
 {
     /// <summary>
-    /// A <see cref="FilterItem"/> provider.
+    /// A <see cref="FilterItem"/> provider. Implementations should update <see cref="FilterProviderContext.Results"/>
+    /// to make executable filters available.
     /// </summary>
     public interface IFilterProvider
     {
--- a/src/Microsoft.AspNetCore.Mvc.Abstractions/Filters/IFilterProvider.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Abstractions/Filters/IFilterProvider.cs
@@ -3,6 +3,9 @@
 
 namespace Microsoft.AspNetCore.Mvc.Filters
 {
+    /// <summary>
+    /// A <see cref="FilterItem"/> provider.
+    /// </summary>
     public interface IFilterProvider
     {
         /// <summary>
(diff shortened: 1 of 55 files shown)


dougbu:
Note this will likely go into RC2 but the PR is against `dev` for now.

dougbu:
FYI this is approved for RC2. I'll port to `release` branch after review is complete.

src/Microsoft.AspNetCore.Mvc.Abstractions/Filters/IFilterProvider.cs
```diff
@@ -26,8 +29,16 @@ public interface IFilterProvider
         /// </remarks>
         int Order { get; }
```
rynowak: Ima let you finish. But these remarks are some of the best remarks of all time :sunglasses:

src/Microsoft.AspNetCore.Mvc.Abstractions/Filters/IFilterProvider.cs
```diff
@@ -3,6 +3,9 @@
 
 namespace Microsoft.AspNetCore.Mvc.Filters
 {
+    /// <summary>
+    /// A <see cref="FilterItem"/> provider.
```
rynowak: I'd mention here that a filter provider can "provide an executable filter associated with an IFilterMetadata by setting FilterItem.Filter, or add FilterItems to add new filters"

src/Microsoft.AspNetCore.Mvc.Core/Authorization/AuthorizeFilter.cs
```diff
@@ -13,12 +13,13 @@
 namespace Microsoft.AspNetCore.Mvc.Authorization
 {
     /// <summary>
-    /// An implementation of <see cref="IAsyncAuthorizationFilter"/>
+    /// An implementation of <see cref="IAsyncAuthorizationFilter"/> which applies a specific
+    /// <see cref="AuthorizationPolicy"/>.
     /// </summary>
```
rynowak: I would add somewhere here that this is the filter that's added by the framework where you use `[Authorize(...)]`

src/Microsoft.AspNetCore.Mvc.Core/TypeFilterAttribute.cs
```diff
@@ -9,12 +9,20 @@
 
 namespace Microsoft.AspNetCore.Mvc
 {
+    /// <summary>
+    /// A filter that activates another filter of type <see cref="ImplementationType"/>.
+    /// </summary>
```
rynowak: This needs some kinds of description of 'activates'.

rynowak:
:shipit:

src/Microsoft.AspNetCore.Mvc.Core/TypeFilterAttribute.cs
dougbu: @rynowak we use "activates" in a few other places e.g. `Controller.ViewData` is `activated when <see cref="Controllers.IControllerActivator"/> activates`. Do those other uses (at least those outside `IRazorPageActivator` and similar) need more words? What words are you suggesting?

Without a specific suggestion, I'll switch to "creates" here.

dougbu:
:up::date:
@rynowak please have a quick look at the words I added to address your comments.

src/Microsoft.AspNetCore.Mvc.Core/TypeFilterAttribute.cs
rynowak: Specifically that it will retrieve the constructor parameters from DI

rynowak:
updates looks good :shipit: x2

dougbu:
eda647e into `release`
Also merged into `dev`

PR #4457, opened 2016-04-12
[Perf]:Cache TypeConverters in SimpleTypeModelBinder

### LINKED ISSUE #4179: Performance Scenario: Big ModelBinding
**This is top-level performance scenario for RC2. We're using this item to track the status and all of the individual improvements.**

Code here: https://github.com/aspnet/Performance/tree/dev/testapp/BigModelBinding
#### Summary

This is an API app that accepts a form post, updates a database, and renders a view for the updated model data. This is an area of known-badness in MVC and should help us put the cost of modelbinding/validation in context in a full application. This should be naively written, like a line-of-business app that's gotten quite large.
- Conventional routing
- Layout page from project template (script/link taghelpers)
- Startup code from no-auth project template
- View code using taghelpers for displaying the model values
# 
#### Variations

(None)
# 
#### Status

Scenario Status: incomplete
- [ ] Review startup code
- [ ] Add View

Known Investigations/Improvements:
- [x] #4361 - Cache TypeConverters in SimpleTypeModelBinder
- [ ]  #4362 - Carry the current ModelStateEntry on the ModelBindingContext
- [ ]  #4016 - Too much string allocation in Model Binding

### LINKED ISSUE #4361: Cache TypeConverters in SimpleTypeModelBinder
See: https://github.com/aspnet/Mvc/issues/4179

image

About 10% of the time in `SimpleTypeModelBinder` is spent looking up the type converter to use. We can close over the type converter in the constructor, since it can be resolved statically.

We can probably just get rid of `ValueProviderResultExtensions.ConvertTo`, or at least not use it in this code path. It does plenty of things that we don't need here and they have a cost.

mnltejaswini:
Fixes #4361 

Scenario under test : https://github.com/aspnet/Performance/tree/dev/testapp/BigModelBinding 
loadtest.ps1
Before the fix
----------------
image

After the fix
---------------
image

No overhead of looking up for converter each time.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/SimpleTypeModelBinder.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/SimpleTypeModelBinder.cs
@@ -2,6 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.ComponentModel;
+using System.Runtime.ExceptionServices;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Mvc.Internal;
 
@@ -12,6 +14,18 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding.Binders
     /// </summary>
     public class SimpleTypeModelBinder : IModelBinder
     {
+        private readonly TypeConverter _typeConverter;
+
+        public SimpleTypeModelBinder(Type type)
+        {
+            if (type == null)
+            {
+                throw new ArgumentNullException(nameof(type));
+            }
+
+            _typeConverter = TypeDescriptor.GetConverter(type);
+        }
+
         /// <inheritdoc />
         public Task BindModelAsync(ModelBindingContext bindingContext)
         {
@@ -32,7 +46,16 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding.Binders
 
             try
             {
-                var model = valueProviderResult.ConvertTo(bindingContext.ModelType);
+                var value = valueProviderResult.FirstValue;
+
+                object model = null;
+                if (!string.IsNullOrWhiteSpace(value))
+                {
+                    model = _typeConverter.ConvertFrom(
+                        context: null,
+                        culture: valueProviderResult.Culture,
+                        value: value);
+                }
 
                 if (bindingContext.ModelType == typeof(string))
                 {
@@ -65,6 +88,14 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding.Binders
             }
             catch (Exception exception)
             {
+                var isFormatException = exception is FormatException;
+                if (!isFormatException && exception.InnerException != null)
+                {
+                    // TypeConverter throws System.Exception wrapping the FormatException,
+                    // so we capture the inner exception.
+                    exception = ExceptionDispatchInfo.Capture(exception.InnerException).SourceException;
+                }
+
                 bindingContext.ModelState.TryAddModelError(
                     bindingContext.ModelName,
                     exception,
--- a/test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/Binders/DictionaryModelBinderTest.cs
+++ b/test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/Binders/DictionaryModelBinderTest.cs
@@ -35,7 +35,9 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding.Binders
             var bindingContext = GetModelBindingContext(isReadOnly, values);
             bindingContext.ValueProvider = CreateEnumerableValueProvider("{0}", values);
 
-            var binder = new DictionaryModelBinder<int, string>(new SimpleTypeModelBinder(), new SimpleTypeModelBinder());
+            var binder = new DictionaryModelBinder<int, string>(
+                new SimpleTypeModelBinder(typeof(int)), 
+                new SimpleTypeModelBinder(typeof(string)));
 
             // Act
             var result = await binder.BindModelResultAsync(bindingContext);
@@ -73,7 +75,9 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding.Binders
             var dictionary = new Dictionary<int, string>();
             bindingContext.Model = dictionary;
 
-            var binder = new DictionaryModelBinder<int, string>(new SimpleTypeModelBinder(), new SimpleTypeModelBinder());
+            var binder = new DictionaryModelBinder<int, string>(
+                new SimpleTypeModelBinder(typeof(int)), 
+                new SimpleTypeModelBinder(typeof(string)));
 
             // Act
             var result = await binder.BindModelResultAsync(bindingContext);
@@ -127,7 +131,9 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding.Binders
             IDictionary<string, string> dictionary)
         {
             // Arrange
-            var binder = new DictionaryModelBinder<string, string>(new SimpleTypeModelBinder(), new SimpleTypeModelBinder());
+            var binder = new DictionaryModelBinder<string, string>(
+                new SimpleTypeModelBinder(typeof(string)), 
+                new SimpleTypeModelBinder(typeof(string)));
 
             var context = CreateContext();
             context.ModelName = modelName;
@@ -163,7 +169,9 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding.Binders
                 { "three", "three" },
             };
 
-            var binder = new DictionaryModelBinder<string, string>(new SimpleTypeModelBinder(), new SimpleTypeModelBinder());
+            var binder = new DictionaryModelBinder<string, string>(
+                new SimpleTypeModelBinder(typeof(string)), 
+                new SimpleTypeModelBinder(typeof(string)));
 
             var context = CreateContext();
             context.ModelName = "prefix";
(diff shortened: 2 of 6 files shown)


src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/SimpleTypeModelBinder.cs
```diff
@@ -32,7 +45,11 @@ public Task BindModelAsync(ModelBindingContext bindingContext)
 
             try
             {
-                var model = valueProviderResult.ConvertTo(bindingContext.ModelType);
+                var value = valueProviderResult.FirstValue;
+
+                var model = !string.IsNullOrWhiteSpace(value)
+                    ? _typeConverter.ConvertFrom(null, valueProviderResult.Culture, value)
+                    : null;
```
rynowak: would be more readable without the ternary. Also why is the first parameter null?

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/SimpleTypeModelBinder.cs
```diff
@@ -67,7 +84,7 @@ public Task BindModelAsync(ModelBindingContext bindingContext)
             {
                 bindingContext.ModelState.TryAddModelError(
                     bindingContext.ModelName,
-                    exception,
+                    exception.InnerException ?? exception,
```
rynowak: Why is this here? I know this was calling into another method before, what's the **important** detail? We should capture that here.

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/Binders/DictionaryModelBinderTest.cs
```diff
@@ -163,7 +163,7 @@ public class DictionaryModelBinderTest
                 { "three", "three" },
             };
 
-            var binder = new DictionaryModelBinder<string, string>(new SimpleTypeModelBinder(), new SimpleTypeModelBinder());
+            var binder = new DictionaryModelBinder<string, string>(new SimpleTypeModelBinder(typeof(string)), new SimpleTypeModelBinder(typeof(string)));
```
rynowak: Do a pass here and break up long lines

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/SimpleTypeModelBinder.cs
```diff
@@ -12,6 +13,18 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding.Binders
...
+        private readonly TypeConverter _typeConverter;
+
+        public SimpleTypeModelBinder(Type type)
+        {
+            if (type == null)
+            {
+                throw new ArgumentNullException(nameof(type));
+            }
+
+            _typeConverter = TypeDescriptor.GetConverter(type);
```
rynowak: You need to check `CanConvert` here and fall back to the string type converter. Take a closer look at the code you're replacing.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/SimpleTypeModelBinder.cs
```diff
@@ -32,7 +45,11 @@ public Task BindModelAsync(ModelBindingContext bindingContext)
...
             try
             {
-                var model = valueProviderResult.ConvertTo(bindingContext.ModelType);
+                var value = valueProviderResult.FirstValue;
+
+                var model = !string.IsNullOrWhiteSpace(value)
+                    ? _typeConverter.ConvertFrom(null, valueProviderResult.Culture, value)
+                    : null;
 
                 if (bindingContext.ModelType == typeof(string))
```
rynowak: IIRC the code that you're replacing had special handling for enums. What's going on with that?

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/SimpleTypeModelBinder.cs
mnltejaswini: I would change it without ternary. The first parameter is ITypeDescriptorContext which is passed as null even in ModelBindingHelper.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/SimpleTypeModelBinder.cs
mnltejaswini: I think that check is done in ModelMetadata 
IsComplexType is true only when there is a converter and CanConvertFrom is true.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/SimpleTypeModelBinder.cs
mnltejaswini: If I understand the flow, that special handling is done only when CanConvertFrom is false which will never happen in SimpleTypeModelBinder.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/SimpleTypeModelBinder.cs
mnltejaswini: TryAddModelError rewrites the message for FormatExcpetions and OverflowException TypeConverter.ConvertFrom throws System exception but the inner exception is a FormatException.

rynowak: I'd expect this block to look very much like the existing code in https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBindingHelper.cs#L1066

For instance we should special case format exception, and it should include the comments about the fact that type converter wraps formatexceptions.

This code is really **replacing** the other code on the common execution path, so I expect it to have all the same bells and whistles.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/SimpleTypeModelBinder.cs
rynowak: Yeah actually I think you're right. I don't know a scenario where the code in ModelBindingHelper would behave differently.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/SimpleTypeModelBinder.cs
rynowak: That's interesting, I guess it makes sense since all VP data is strings.

rynowak:
:watch:

mnltejaswini:
:up::date:

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/SimpleTypeModelBinder.cs
```diff
@@ -32,7 +46,15 @@ public Task BindModelAsync(ModelBindingContext bindingContext)
 
             try
             {
-                var model = valueProviderResult.ConvertTo(bindingContext.ModelType);
+                var value = valueProviderResult.FirstValue;
+
+                object model = null;
+                if (!string.IsNullOrWhiteSpace(value))
+                {
+                    model = _typeConverter.ConvertFrom(context: null,
```
rynowak: context should be on the next line. Either put them each on their own line or not at all

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/SimpleTypeModelBinder.cs
```diff
@@ -65,6 +87,13 @@ public Task BindModelAsync(ModelBindingContext bindingContext)
             }
             catch (Exception exception)
             {
+                if (exception.InnerException != null)
+                {
+                    // TypeConverter throws System.Exception wrapping the FormatException,
+                    // so we capture the inner exception.
+                    exception = ExceptionDispatchInfo.Capture(exception.InnerException).SourceException;
```
rynowak: This should have another catch block to special case `FormatException`

rynowak:
:watch:

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/SimpleTypeModelBinder.cs
mnltejaswini: Not sure why we need that because TypeConverter throws System.Exception wrapping FormatException. (tried few samples through a separate Console application)

rynowak: I'd prefer that this be handled explicitly. We know this code can throw a `FormatException` or a wrapped `FormatException` - taking a dependency on the `InnerException` being null in the first case is brittle

mnltejaswini:
:up::date:

rynowak:
I looked through the tests for `ConvertTo`, and these are the ones that it seems relevant to port to the `SimpleTypeModelBinder`

ConvertToReturnsNullIfTrimmedValueIsEmptyString
ConvertToReturnsValueIfArrayElementIsStringValueAndDestinationTypeIsEnum
ConvertToReturnsValueIfArrayElementIsStringKeyAndDestinationTypeIsEnum
ConvertToReturnsValueIfElementIsStringAndDestinationIsNullableInteger
ConvertToReturnsValueIfElementIsStringAndDestinationIsNullableDouble
ConvertToUsesProvidedCulture

:watch:

mnltejaswini:
:up::date:

rynowak:
:shipit: - looks great!

PR #4463, opened 2016-04-13
Add web.config to test sites

JunTaoLuo:
Follow up for https://github.com/aspnet/Mvc/pull/4433. This time, I'm also updating the tests sites. Although the functional tests will not run behind IIS, we decided to add IISIntegration so the sites themselves can be launched from Visual Studios.

cc @ryanbrandenburg @Tratcher @NTaylorMullen

### DIFF
--- a/test/WebSites/ApiExplorerWebSite/Startup.cs
+++ b/test/WebSites/ApiExplorerWebSite/Startup.cs
@@ -47,6 +47,7 @@ namespace ApiExplorerWebSite
             var host = new WebHostBuilder()
                 .UseDefaultHostingConfiguration(args)
                 .UseKestrel()
+                .UseIISIntegration()
                 .UseStartup<Startup>()
                 .Build();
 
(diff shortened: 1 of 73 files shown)


ryanbrandenburg:
:shipit: Seems good to me, looking like it's missing some `wwwroot`'s in content, but my PR is already fixing those so no need to change it.

JunTaoLuo:
@ryanbrandenburg where am I missing `wwwroot`? I shouldn't be missing anything.

ryanbrandenburg:
Huh, I completely mis-remembered that, I could have sworn most of the test/Website sites had wwwroot folders but they don't. Ignore me!

JunTaoLuo:
@ryanbrandenburg I removed the ones that had only a placeholder `HelloWorld.htm` since we no longer need them. Maybe that's what you were thinking of?

ryanbrandenburg:
Probably.

PR #4472, opened 2016-04-14
Fixes #4471 - Don't call Enum.ToString()

### LINKED ISSUE #4471: GenerateForm calls Enum.ToString
Sampling data from BasicViews - percentages are relative to time spent in the view

image

Notice here that calling `Enum.ToString` and all of the things that it does are **more expensive** than generating the URL in routing. It's **half as expensive** as generating an antiforgery token, which does crypto....

rynowak:
Best perf fix ever :laughing:

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.TagHelpers/FormTagHelper.cs
+++ b/src/Microsoft.AspNetCore.Mvc.TagHelpers/FormTagHelper.cs
@@ -229,7 +229,7 @@ namespace Microsoft.AspNetCore.Mvc.TagHelpers
                     output.PostContent.AppendHtml(tagBuilder.InnerHtml);
                 }
 
-                if (string.Equals(Method, FormMethod.Get.ToString(), StringComparison.OrdinalIgnoreCase))
+                if (string.Equals(Method, "get", StringComparison.OrdinalIgnoreCase))
                 {
                     antiforgeryDefault = false;
                 }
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Mvc.TagHelpers/FormTagHelper.cs
```diff
@@ -229,7 +229,7 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
                     output.PostContent.AppendHtml(tagBuilder.InnerHtml);
                 }
 
-                if (string.Equals(Method, FormMethod.Get.ToString(), StringComparison.OrdinalIgnoreCase))
+                if (string.Equals(Method, "get", StringComparison.OrdinalIgnoreCase))
```
pranavkm: But what if `FormMethod.Get.ToString` returns `post`?

pranavkm:
:shipit:

src/Microsoft.AspNetCore.Mvc.TagHelpers/FormTagHelper.cs
dougbu: That won't happen 'til c#9 I'm sure.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
```diff
@@ -1104,7 +1104,7 @@ internal static object GetModelStateValue(ViewContext viewContext, string key, T
             if (string.IsNullOrEmpty(method))
             {
                 // Occurs only when called from a tag helper.
-                method = FormMethod.Post.ToString().ToLowerInvariant();
+                method = "post";
```
dougbu: Suggest sticking "post" in a `const string` in this class.

dougbu:
:shipit:

PR #4478, opened 2016-04-14
[Perf] Avoid Service Provider lookups when activating common Singleton properties of a Razor Page

### LINKED ISSUE #4244: RazorPage activation is too expensive
Sampling data, with 100% set to time executing the view (30% of request total)

Update: this was largely due to a temporary regression in GetRequiredService. We need to remeasure.

This is what it looks like now: 
image

Activation is about 7.5% of view execution, and of that half is the VDD creation which is being worked on actively. Still worth looking at.

Certain things that are part of the activator's workload are singletons that we can close over. https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Razor/RazorPageActivator.cs#L125

We should use the `GetRequiredService` calls as a guide for things we can cache and handle specially.
image

mnltejaswini:
Fixes #4244
Some of the common singleton properties that are activated on a Razor Page for every request are of types IJsonHelper, HtmlEncoder, DiagnosticSource, IUrlHelperFactory. For every request, these properties are activated by looking up in the service provider. The cost of these lookups can be avoided by special casing their activation in RazorPageActivator.

bottom up sampling of Hello World MVC app 
Before optimization
------------------------ 
image

After optimization
----------------------
image

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor/RazorPageActivator.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor/RazorPageActivator.cs
@@ -6,6 +6,7 @@ using System.Collections.Concurrent;
 using System.Diagnostics;
 using System.Linq.Expressions;
 using System.Reflection;
+using System.Text.Encodings.Web;
 using Microsoft.AspNetCore.Mvc.ModelBinding;
 using Microsoft.AspNetCore.Mvc.Razor.Internal;
 using Microsoft.AspNetCore.Mvc.Rendering;
@@ -29,13 +30,28 @@ namespace Microsoft.AspNetCore.Mvc.Razor
         private readonly ConcurrentDictionary<Type, PageActivationInfo> _activationInfo;
         private readonly IModelMetadataProvider _metadataProvider;
 
+        // Value accessors for common singleton properties activated in a RazorPage.
+        private Func<ViewContext, object> _urlHelperAccessor;
+        private Func<ViewContext, object> _jsonHelperAccessor;
+        private Func<ViewContext, object> _diagnosticSourceAccessor;
+        private Func<ViewContext, object> _htmlEncoderAccessor;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="RazorPageActivator"/> class.
         /// </summary>
-        public RazorPageActivator(IModelMetadataProvider metadataProvider)
+        public RazorPageActivator(
+            IModelMetadataProvider metadataProvider,
+            IUrlHelperFactory urlHelperFactory,
+            IJsonHelper jsonHelper,
+            DiagnosticSource diagnosticSource,
+            HtmlEncoder htmlEncoder)
         {
             _activationInfo = new ConcurrentDictionary<Type, PageActivationInfo>();
             _metadataProvider = metadataProvider;
+            _urlHelperAccessor = context => urlHelperFactory.GetUrlHelper(context);
+            _jsonHelperAccessor = context => jsonHelper;
+            _diagnosticSourceAccessor = context => diagnosticSource;
+            _htmlEncoderAccessor = context => htmlEncoder;
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Mvc.Razor/RazorPageActivator.cs
```diff
@@ -160,12 +184,51 @@ private PropertyActivator<ViewContext> CreateActivateInfo(PropertyInfo property)
...
+                valueAccessor = _urlHelperAccessor;
+            }
+            else if (property.PropertyType == typeof(IJsonHelper))
+            {
+                if (_jsonHelperAccessor == null)
+                {
+                    _jsonHelperAccessor = context => 
+                    {
+                        return _jsonHelper;
+                    };
```
dougbu: Shorten these simple expressions to e.g. `context => _jsonHelper`.

src/Microsoft.AspNetCore.Mvc.Razor/RazorPageActivator.cs
```diff
@@ -29,13 +31,35 @@ public class RazorPageActivator : IRazorPageActivator
...
+        private Func<ViewContext, object> _urlHelperAccessor;
+
+        private readonly IJsonHelper _jsonHelper;
+        private Func<ViewContext, object> _jsonHelperAccessor;
+
+        private readonly DiagnosticSource _diagnosticSource;
+        private Func<ViewContext, object> _diagnosticSourceAccessor;
+
+        private readonly HtmlEncoder _htmlencoder;
+        private Func<ViewContext, object> _htmlencoderAccessor;
```
dougbu: `htmlEncoder` in both of these

src/Microsoft.AspNetCore.Mvc.Razor/RazorPageActivator.cs
```diff
@@ -29,13 +31,35 @@ public class RazorPageActivator : IRazorPageActivator
...
+            IJsonHelper jsonHelper,
+            DiagnosticSource diagnosticSource,
+            HtmlEncoder htmlEncoder)
         {
             _activationInfo = new ConcurrentDictionary<Type, PageActivationInfo>();
             _metadataProvider = metadataProvider;
+            _urlHelperFactory = urlHelperFactory;
+            _jsonHelper = jsonHelper;
+            _diagnosticSource = diagnosticSource;
+            _htmlencoder = htmlEncoder;
```
dougbu: Suggest removing the lazy init of the accessors and doing those assignments here.

dougbu:
:watch: for slight cleanup

rynowak:
In your **after** measurement, there's still a significant chunk of time spent in `<CreateActivateInfo>__b_5` - can you include the rest of the data? What is that doing?

src/Microsoft.AspNetCore.Mvc.Razor/RazorPageActivator.cs
rynowak: :+1: to this suggestion

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorPageActivatorTest.cs
```diff
@@ -29,19 +28,26 @@ public class RazorPageActivatorTest
         public void Activate_ActivatesAndContextualizesPropertiesOnViews()
         {
             // Arrange
-            var activator = new RazorPageActivator(new EmptyModelMetadataProvider());
+            var urlHelperFactory = new UrlHelperFactory();
+            var jsonHelper = new JsonHelper(new Formatters.JsonOutputFormatter());
```
rynowak: Add a `using`

mnltejaswini:
They are for IViewComponentHelper and IHtmlHelper<T>  which are transient services.

mnltejaswini:
:up::date:

rynowak:
:shipit:

Issue #3978, opened 2016-01-23
ModelExpression equivalent of m => m?


newsash:
How to use label tag helper in the following scenario in a display template? `asp-for="@Model"` leads to invalid label text ("Model").
``` c#
@Html.LabelFor(m => m, new { @class = "control-label col-md-2" })
```

NTaylorMullen:
`<label asp-for="@Model" class="control-label col-md-2"></label>`

newsash:
In a display template (e.g. `@Html.DisplayFor(m => m.UserName`):
* `<label asp-for="@Model"></label>` results in <code>&lt;label for="UserName"&gt;**Model**&lt;/label&gt;</code>
* `@Html.LabelFor(m => m)` results in <code>&lt;label for="UserName"&gt;**UserName**&lt;/label&gt;</code>

dougbu:
I'm able to reproduce this as well. Haven't debugged what is going South but it does look strange, especially given the `for` attribute has the correct value.

dougbu:
FYI problem relates to how the `ModelMetadata` for the property is found. In the `@Html.LabelFor(m => m)` case, `PropertyName` is `"UserName"`. But in the `<label asp-for="@Model"></label>` case, `PropertyName` is `"Model"`. Looking a bit deeper...

dougbu:
:ok: the issue here is that `ExpressionMetadataProvider.FromLambdaExpression()` correctly handles `m => m` and other identity expressions (`ExpressionType.Parameter`) but does not understand that `m => {generated class}.Model` needs special handling. That is, `@Html.LabelFor(m => Model)` hits exactly the same naming issue.

Unfortunately tag helpers do not support lambda identity expressions and _always_ hit this issue if the `@Model` "fallback" is used. We probably didn't see this before because (a) the `ModelMetadata.PropertyName` property isn't used extensively in HTML generation and (b) `@Model.UserName`, `@Model[i].UserName`, and other lambda RHS that start with `@Model` work fine.

One caveat on (a): `string`s generated for client-side validation rules can be messed up when `PropertyName` is incorrect e.g. `data-val-required="The Model field is required."`.

dougbu:
And, no, I don't know of an RC1 workaround other than to use HTML helpers for this case.

dougbu:
`release`: 7a1ac03
also merged into `dev`

seraphx2:
So we can, or we cannot, use asp-for tag helpers in Display/Edit templates?
[...]

PR #4480, opened 2016-04-14
Special-case use of `razorPage.Model` property in `ExpressionMetadataProvider`

dougbu:
- #3978
- better-aligns `ExpressionMetadataProvider` with `ExpressionHelper`

nit: mock less in `RazorPageCreateModelExpressionTest`

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionMetadataProvider.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionMetadataProvider.cs
@@ -51,10 +51,12 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
                     var memberExpression = (MemberExpression)expression.Body;
                     propertyName = memberExpression.Member is PropertyInfo ? memberExpression.Member.Name : null;
                     if (string.Equals(propertyName, "Model", StringComparison.Ordinal) &&
-                        memberExpression.Type == typeof(TModel))
+                        memberExpression.Type == typeof(TModel) &&
+                        memberExpression.Expression.NodeType == ExpressionType.Constant)
                     {
-                        // Special case the Model property in RazorPage<TModel>. (m => Model) should behave
-                        // identically to (m => m).
+                        // Special case the Model property in RazorPage<TModel>. (m => Model) should behave identically
+                        // to (m => m). But do the more complicated thing for (m => m.Model) since that is a slightly
+                        // different beast.)
                         return FromModel(viewData, metadataProvider);
                     }
 
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionMetadataProvider.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionMetadataProvider.cs
@@ -50,6 +50,14 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
                     // Property/field access is always legal
                     var memberExpression = (MemberExpression)expression.Body;
                     propertyName = memberExpression.Member is PropertyInfo ? memberExpression.Member.Name : null;
+                    if (string.Equals(propertyName, "Model", StringComparison.Ordinal) &&
+                        memberExpression.Type == typeof(TModel))
+                    {
+                        // Special case the Model property in RazorPage<TModel>. (m => Model) should behave
+                        // identically to (m => m).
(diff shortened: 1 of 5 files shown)


test/Microsoft.AspNetCore.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.Order.html
```diff
@@ -70,7 +70,7 @@
         </div>
         <div>
             <label class="order" for="Customer_Gender">Gender</label>
-            <input type="radio" value="Male" data-val="true" data-val-required="The Model field is required." id="Customer_Gender" name="Customer.Gender" /> Male
+            <input type="radio" value="Male" data-val="true" data-val-required="The Gender field is required." id="Customer_Gender" name="Customer.Gender" /> Male
```
dougbu: That's right -- we already had a functional test for this scenario. It was just looking for a broken error message.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionMetadataProvider.cs
```diff
@@ -50,6 +50,14 @@ public static class ExpressionMetadataProvider
                     // Property/field access is always legal
                     var memberExpression = (MemberExpression)expression.Body;
                     propertyName = memberExpression.Member is PropertyInfo ? memberExpression.Member.Name : null;
+                    if (string.Equals(propertyName, "Model", StringComparison.Ordinal) &&
```
rynowak: Is this though enough? Should we check the expression on the LHS? What about a case like:
```
public class Thing
{
    public Thing Model { get; set; }
}
```

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Internal/ExpressionMetadataProviderTest.cs
```diff
@@ -1,14 +1,92 @@
...
 using Microsoft.AspNetCore.Mvc.ModelBinding;
+using Microsoft.AspNetCore.Mvc.ModelBinding.Metadata;
 using Xunit;
 
 namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Internal
 {
     public class ExpressionMetadataProviderTest
     {
         [Fact]
+        public void FromLambdaExpression_GetsExpectedMetadata_ForIdentityExpression()
```
rynowak: Good test adds here :+1:

rynowak:
:watch: for discussion on https://github.com/aspnet/Mvc/pull/4480/files#r59813906

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionMetadataProvider.cs
dougbu: Kind of a derivative case but possible. I guess this code could check whether the LHS is a `RazorPage<TModel>` instance. Lemme see how ugly that gets.

If clean, would break the `m => ViewData.Model` support we get for free right now. But `ExpressionHelper` gets the names wrong for that one anyway. No biggie either way.

rynowak: checking the type isn't really good enough (custom pages). What if you look for `this` or an instance of `VDD`?

rynowak: Yeah, we discussed this and their aren't great solutions. If we assume that you meant `m => m` when you meant `m => m.Model` the worst case is that we lose metadata on the `m.Model` property.

It's unfortunate we can't be completely precise, but this seems ok.

rynowak:
:shipit:

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ExpressionMetadataProvider.cs
dougbu: As discussed offline, can at least make this more specific and minimize the corner cases: Changed this condition to include `&& memberExpression.Expression.NodeType == ExpressionType.Constant`.

dougbu:
:up::date: FYI only unless @rynowak hates tests I added in last commit.

Will port to `release` branch and get this in tonight or over the weekend.

dougbu:
`release`: 7a1ac03
also merged into `dev`

PR #4488, opened 2016-04-15
[Perf] Reduce DefaultTagHelperContent allocations in LinkTagHelper and ScriptTagHelper

### LINKED ISSUE #4468: Reduce allocations in Link/Script TagHelper
Allocation data from BasicViews:

image

Looking here: https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.TagHelpers/LinkTagHelper.cs#L259

We create a new `DefaultTagHelperContent` object which we fill with content, and then ultimately add to the output builder. The output builder already gets recycled, we can just use it directly.

mnltejaswini:
Fixes #4468
Saves one DefaultTagHelperContent object allocation per link/script tag per request.


dougbu:
:shipit:

rynowak:
:shipit:

Issue #4409, opened 2016-04-05
Consider removing support for modifying the JsonSerializerSettings after startup


javiercn:
JsonOutputFormatter "allows" to change the settings at runtime, this is different from what JsonInputFormatter does, so It leaves me wondering, is there really an scenario for changing the settings after startup, or is this just the case that we wrote the code but didn't think the scenarios through?

```C#
public JsonSerializerSettings SerializerSettings
{
    get
    {
        return _serializerSettings;
    }
    set
    {
        if (value == null)
        {
            throw new ArgumentNullException(nameof(value));
        }

        _serializerSettings = value;

        // If the settings change, then invalidate the cached serializer.
        _serializer = null;
    }
}
```

I don't think we should support this scenario as it's inconsistent with the scenarios in JsonInputFormatter and I can't see an advantage of doing this over replacing the formatter in the list of formatters in the options. I believe Options has support for this built in https://github.com/aspnet/Options/blob/dev/test/Microsoft.Extensions.Options.Test/OptionsMonitorTest.cs

Eilon:
Yeah this seems a bit weird. I don't think you can effectively change any of these settings after startup. You'd cause crazy race conditions and the world would likely end.

@rynowak @dougbu any sense whether there was some scenario here at some point? I suspect we can delete some of this "guard" code, no?

dougbu:
@Eilon this code may exist only for unit testing scenarios. I don't recall a production scenario.

Then again, output formatters are more likely to be created from user code than input formatters e.g. within a custom `IActionResult`. I could imagine a user looking to customize an output formatter on, say, a per-action or per-response `Type` basis.

Eilon:
You might be able to customize formatters on a per-request basis, but *definitely* not by modifying the instance that's already in the `OutputFormatters` collection. Isn't that instance shared among all requests?

rynowak:
Yeah, we might want to just make it clearer what's supported and not by taking in the settings in the constructor.

Having the settings in a mutable property is a holdover from WebAPI2 and it might be a distraction from the intended pattern for configuring the formatter.

Eilon:
Yeah let's clean up the ctor's and make the properties get-only.

kichalla:
I am a bit unclear. So what is the final change that needs to be made here?

[...]

Issue #4339, opened 2016-03-24
Remove extra constructors in `JsonInputFormatter` and `JsonPatchInputFormatter`


dougbu:
The `JsonInputFormatter` and `JsonPatchInputFormatter` classes previously had a parameterless constructor for ease of use. This was lost when a required `ILogger` parameter was added. We have also since decided there's no important user scenario involving construction of these types. But `JsonInputFormatter` still has 3 constructor overloads while `JsonPatchInputFormatter` has 2. The extra overloads are used only in test code and are somewhat complicated i.e. don't just pass `null` to the longest overload.

Should remove all but the longest constructor overload in each class. Change the related tests to use the one remaining constructor for each class.

Eilon:
Is there any reasonably common scenario for new'ing up these types? (Aside from tests, which is mostly our own internal concern.)

rynowak:
I'm not really sure. There was in WebAPI - but that was often a workaround for other things being obtuse....

Only case I can think of is if you have two different sets of JSON settings.

rynowak:
There's still an action item to take here if we don't want to make them constructable, we should clean up the other ctors.

Eilon:
If there's nothing "common" then I'll put in Backlog. If there's something more urgent for v1, what is it that we should do?

dougbu:
Suggest that if there's no common scenario, we were wrong to add the extra constructors. They are a bit complicated and do odd things like `new` up a `DefaultObjectPoolProvider`. As @rynowak said, the action item would be to remove them. Could even change the bug title :smiley_cat:

Eilon:
Well, given what I've understood here, this would be backlog. So if you could distill this down to exactly what new proposed action we should take sooner, let's take a look at that.

dougbu:
Sure. I've updated the title and description. (In retrospect, should have posed this issue as a question on which way to go.)

rynowak:
> Suggest that if there's no common scenario, we were wrong to add the extra constructors

- Constructable
- DI
- Simple

Choose any two.

Eilon:
> Should remove all but the longest constructor overload in each class. Change the related tests to use the one remaining constructor for each class.

@rynowak are you onboard with that?

rynowak:
Yes

Eilon:
Ok, we'll do in RTM.

kwaclaw:
[...]

PR #4492, opened 2016-04-17
Remove extra options to manipulate `JsonSerializerSettings`

dougbu:
- #4339: remove non-recommended JSON formatter constructors
 - affects `JsonInputFormatter`, `JsonOutputFormatter`, `JsonPatchInputFormatter`
 - `JsonOutputFormatter` cleanup also impacts `JsonHelper`
 - make `SerializerSettingsProvider` class public and use it
- #4409: make `SerializerSetings` properties get-only and `protected`
 - affects `JsonInputFormatter`, `JsonOutputFormatter`

Recommended patterns:
- change `JsonSerializerSettings` values in `MvcJsonOptions` for almost all customizations
- find `JsonOutputFormatter` in `MvcOptions.OutputFormatters` when limiting per-result formatters
- start with `SerializerSettingsProvider.CreateSerializerSettings()` when customizing a per-result formatter

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Formatters.Json/JsonOutputFormatter.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Formatters.Json/JsonOutputFormatter.cs
@@ -7,13 +7,12 @@ using System.IO;
 using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Mvc.Formatters.Json.Internal;
-using Microsoft.AspNetCore.Mvc.Internal;
 using Newtonsoft.Json;
 
 namespace Microsoft.AspNetCore.Mvc.Formatters
 {
     /// <summary>
-    /// An output formatter that specializes in writing JSON content.
+    /// A <see cref="TextOutputFormatter"/> for JSON content.
     /// </summary>
     public class JsonOutputFormatter : TextOutputFormatter
     {
@@ -25,16 +24,11 @@ namespace Microsoft.AspNetCore.Mvc.Formatters
         // the serializer and invalidate it when the settings change.
         private JsonSerializer _serializer;
 
-        public JsonOutputFormatter()
-            : this(SerializerSettingsProvider.CreateSerializerSettings(), ArrayPool<char>.Shared)
-        {
-        }
-
-        public JsonOutputFormatter(JsonSerializerSettings serializerSettings)
-            : this(serializerSettings, ArrayPool<char>.Shared)
-        {
-        }
-
+        /// <summary>
+        /// Initializes a new <see cref="JsonOutputFormatter"/> instance.
+        /// </summary>
+        /// <param name="serializerSettings">The <see cref="JsonSerializerSettings"/>.</param>
+        /// <param name="charPool">The <see cref="ArrayPool{Char}"/>.</param>
         public JsonOutputFormatter(JsonSerializerSettings serializerSettings, ArrayPool<char> charPool)
         {
             if (serializerSettings == null)
@@ -157,7 +151,7 @@ namespace Microsoft.AspNetCore.Mvc.Formatters
                 WriteObject(writer, context.Object);
 
                 // Perf: call FlushAsync to call WriteAsync on the stream with any content left in the TextWriter's
-                // buffers. This is better than just letting dispose handle it (which would result in a synchronous 
+                // buffers. This is better than just letting dispose handle it (which would result in a synchronous
                 // write).
                 await writer.FlushAsync();
             }
--- a/src/Microsoft.AspNetCore.Mvc.Formatters.Json/JsonOutputFormatter.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Formatters.Json/JsonOutputFormatter.cs
@@ -25,7 +25,11 @@ namespace Microsoft.AspNetCore.Mvc.Formatters
         /// <summary>
         /// Initializes a new <see cref="JsonOutputFormatter"/> instance.
         /// </summary>
-        /// <param name="serializerSettings">The <see cref="JsonSerializerSettings"/>.</param>
+        /// <param name="serializerSettings">
+        /// The <see cref="JsonSerializerSettings"/>. Should be either the application-wide settings
+        /// (<see cref="MvcJsonOptions.SerializerSettings"/>) or an instance
+        /// <see cref="JsonSerializerSettingsProvider.CreateSerializerSettings"/> initially returned.
+        /// </param>
         /// <param name="charPool">The <see cref="ArrayPool{Char}"/>.</param>
         public JsonOutputFormatter(JsonSerializerSettings serializerSettings, ArrayPool<char> charPool)
         {
--- a/src/Microsoft.AspNetCore.Mvc.Formatters.Json/JsonOutputFormatter.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Formatters.Json/JsonOutputFormatter.cs
@@ -18,8 +18,6 @@ namespace Microsoft.AspNetCore.Mvc.Formatters
     {
         private readonly IArrayPool<char> _charPool;
 
-        private JsonSerializerSettings _serializerSettings;
-
         // Perf: JsonSerializers are relatively expensive to create, and are thread safe. We cache
         // the serializer and invalidate it when the settings change.
         private JsonSerializer _serializer;
@@ -41,7 +39,7 @@ namespace Microsoft.AspNetCore.Mvc.Formatters
                 throw new ArgumentNullException(nameof(charPool));
             }
 
-            _serializerSettings = serializerSettings;
+            SerializerSettings = serializerSettings;
             _charPool = new JsonArrayPool<char>(charPool);
 
             SupportedEncodings.Add(Encoding.UTF8);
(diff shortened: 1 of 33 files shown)


src/Microsoft.AspNetCore.Mvc.Formatters.Json/JsonOutputFormatter.cs
```diff
@@ -25,16 +24,11 @@ public class JsonOutputFormatter : TextOutputFormatter
...
 
-        public JsonOutputFormatter()
-            : this(SerializerSettingsProvider.CreateSerializerSettings(), ArrayPool<char>.Shared)
-        {
-        }
-
-        public JsonOutputFormatter(JsonSerializerSettings serializerSettings)
-            : this(serializerSettings, ArrayPool<char>.Shared)
-        {
-        }
```
dougbu: Removing these wasn't mentioned in #4339 but seems to fit. Let me know if you disagree.

test/WebSites/ApiExplorerWebSite/Startup.cs
```diff
@@ -23,9 +23,23 @@ public void ConfigureServices(IServiceCollection services)
...
+                JsonOutputFormatter jsonOutputFormatter = null;
+                for (var i = 0; i < options.OutputFormatters.Count; i++)
+                {
+                    var formatter = options.OutputFormatters[i];
+                    jsonOutputFormatter = formatter as JsonOutputFormatter;
+                    if (jsonOutputFormatter != null)
+                    {
+                        break;
+                    }
+                }
```
dougbu: Thoughts on adding a `FormatterCollection<TFormatter>.Find<T>() where T : TFormatter` method (or `Get()` or `FindType()` or ...)? I did this twice in this PR and that may indicate users would do the same.

test/WebSites/BasicWebSite/Controllers/ContentNegotiation/NormalController.cs
```diff
@@ -1,28 +1,70 @@
...
 namespace BasicWebSite.Controllers.ContentNegotiation
 {
     public class NormalController : Controller
     {
+        private JsonOutputFormatter _indentingFormatter;
+
+        public NormalController(IOptions<MvcJsonOptions> jsonOptions, ArrayPool<char> charPool)
+        {
+            var defaultSettings = jsonOptions.Value.SerializerSettings;
+            var settings = new JsonSerializerSettings
```
dougbu: This is tedious but `JsonSerializerSettings` lacks a copy constructor. Any objection to this test site simply using `new JsonSerializerSettings { Formatting = Formatting.Indented }`? Middle ground would be to copy over only what `SerializerSettingsProvider` changes.

test/WebSites/FiltersWebSite/Controllers/ResourceFilterController.cs
```diff
@@ -31,7 +33,9 @@ private class ShortCircuitWithFormatterAttribute : Attribute, IResourceFilter
 
             public ShortCircuitWithFormatterAttribute()
             {
-                _formatters = new IOutputFormatter[] { new JsonOutputFormatter() };
+                _formatters = new IOutputFormatter[] { new JsonOutputFormatter(
+                    new JsonSerializerSettings(),
```
dougbu: Hmm, should I do the Right Thing<sup>:tm:</sup> and get the settings from `IOptoins<MvcJsonOptions>`? That is, is it :ok: to demonstrate the lazy approach in one of our web sites? Same question for `JsonFormatterController` below.

src/Microsoft.AspNetCore.Mvc.Formatters.Json/JsonOutputFormatter.cs
kichalla: Users trying to do per-controller specific formatters was fairly common in Web API. In this scenario users would have to new up both input and output formatters. Removing this probably does not give the best experience IMO.

test/WebSites/FormatterWebSite/Controllers/JsonFormatterController.cs
```diff
@@ -21,8 +22,11 @@ public IActionResult ReturnsIndentedJson()
                 Name = "John Williams"
             };
 
-            var jsonFormatter = new JsonOutputFormatter();
-            jsonFormatter.SerializerSettings.Formatting = Formatting.Indented;
+            var settings = new JsonSerializerSettings
+            {
+                Formatting = Formatting.Indented,
+            };
+            var jsonFormatter = new JsonOutputFormatter(settings, ArrayPool<char>.Shared);
```
kichalla: This is a fairly common scenario. Users needing to know about `ArrayPool` is probably not a good experience.

test/WebSites/FiltersWebSite/Controllers/ResourceFilterController.cs
kichalla: (I suck at these emojis :-)...I did a +1 unintentionally...how do i revert that back?...doing  a -1 created one more entry :-))
The instance of serializer settings on `MvcJsonOptions` is shared across all controllers, so it cannot be used on `JsonFormatterController` as it is mutating it. Probably have a comment here about `MvcJsonOptions` here.

test/WebSites/BasicWebSite/Controllers/ContentNegotiation/NormalController.cs
kichalla: This is what I was not comfortable about with not having a way to create `JsonSerializerSettings` with all the defaults that the framework provides. I think we should provide an easier way for end users to get these settings.

test/WebSites/ApiExplorerWebSite/Startup.cs
kichalla: Use LINQ instead?

dougbu: > Use LINQ instead?

Ick

src/Microsoft.AspNetCore.Mvc.Formatters.Json/JsonOutputFormatter.cs
dougbu: > In this scenario users would have to new up both input and output formatters.

In what scenario would users need a per-controller _input_ formatter? And, despite our penchant for testing corner cases, when would users really need custom settings for a single output formatter?

test/WebSites/BasicWebSite/Controllers/ContentNegotiation/NormalController.cs
dougbu: > we should provide an easier way for end users to get these settings.

These settings come from very few places: `SerializerSettingsProvider.CreateSerializerSettings()` and the user's own code in `Startup` or a resource or action filter e.g. an `OnActionExecuting()` override in their controller. If needed, we could move `SerializerSettingsProvider` into a non-`.Internal` namespace.

test/WebSites/FiltersWebSite/Controllers/ResourceFilterController.cs
dougbu: I wasn't suggesting reusing the global settings in `JsonFormatterController`, just grabbing them as a starting point. In any case, the user is not mutating the settings here in `ResourceFilterController`. In `JsonFormatterController`, would need to either leave the code as I've changed or match whatever copying we choose in `NormalController` (above).

Separately the previous code in `JsonFormatterController` is fragile. Should not depend on the formatter's constructor to create a unique `JsonSerializerSettings` instance. Another good reason for #4409, killing the `SerializerSettings` property to death.

src/Microsoft.AspNetCore.Mvc.Formatters.Json/JsonOutputFormatter.cs
dougbu: That said, if others agree, I could restore the `JsonOutputFormatter(JsonSerializerSettings)` overload. Don't see any reason for the parameterless constructor.

test/WebSites/FormatterWebSite/Controllers/JsonFormatterController.cs
dougbu: I doubt users often need to change serializer settings outside their `Startup` code. But, as I [said](https://github.com/aspnet/Mvc/pull/4492/files#r59992980) in `JsonOutputFormatter`, I could restore the `JsonOutputFormatter(JsonSerializerSettings)` constructor overload.

test/WebSites/ApiExplorerWebSite/Startup.cs
javiercn: OfType<T>().FirstOrDefault(). I don't feel there's a need for a specifc method.

test/WebSites/BasicWebSite/Controllers/ContentNegotiation/NormalController.cs
javiercn: I wouldn't go to all the trouble of copying all the settings if the intent of the test is to prove that you can override the formatter per controller, I would just set anything that is not a default value.

src/Microsoft.AspNetCore.Mvc.Formatters.Json/JsonOutputFormatter.cs
dougbu: :see_no_evil: Decided creating a customized JSON output formatter is an uncommon scenario and doesn't need to be incredibly simple. The required parameters are easy to create (because `SerializerSettingsProvider` is public) or to grab from DI. The same is true for the JSON input formatters though they have a couple more values that users must grab from DI.

dougbu:
:up::date: also generalized the PR title and description

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/JsonHelper.cs
```diff
@@ -46,7 +57,7 @@ public IHtmlContent Serialize(object value, JsonSerializerSettings serializerSet
                 throw new ArgumentNullException(nameof(serializerSettings));
             }
 
-            var jsonOutputFormatter = new JsonOutputFormatter(serializerSettings);
+            var jsonOutputFormatter = new JsonOutputFormatter(serializerSettings, _charPool);
```
rynowak: :heart_eyes_cat:

rynowak:
It would be good to document as part of the docs for these constructors have to get both the *app-wide* and *secure-default* serializer settings.

Other than that :shipit:

dougbu:
:up::date: @rynowak I made the change we discussed offline...

rynowak:
:shipit:

dougbu:
d8d2e54

PR #4501, opened 2016-04-19
Remove OperationBindingContext

rynowak:
This change trims a few concepts that aren't really needed inside
ModelBinders anymore, and removes the OperationBindingContext class.


src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs
```diff
@@ -178,18 +178,22 @@ internal static void AddMvcCoreServices(IServiceCollection services)
...
+            services.TryAdd(ServiceDescriptor.Transient<ICompositeMetadataDetailsProvider>(s =>
             {
-                var options = serviceProvider.GetRequiredService<IOptions<MvcOptions>>().Value;
+                var options = s.GetRequiredService<IOptions<MvcOptions>>().Value;
                 return new DefaultCompositeMetadataDetailsProvider(options.ModelMetadataDetailsProviders);
             }));
             services.TryAddSingleton<IModelBinderFactory, ModelBinderFactory>();
-            services.TryAddSingleton<IObjectModelValidator, DefaultObjectValidator>();
-            services.TryAddSingleton<ValidatorCache>();
+            services.TryAdd(ServiceDescriptor.Singleton<IObjectModelValidator>(s =>
```
javiercn: Why not use the overload TryAddSingleton<T>(factory)?

src/Microsoft.AspNetCore.Mvc.Core/ControllerContext.cs
```diff
@@ -56,56 +52,6 @@ public new ControllerActionDescriptor ActionDescriptor
...
-                }
-
-                _inputFormatters = value;
-            }
-        }
-
-        /// <summary>
-        /// Gets or sets the list of <see cref="IModelValidatorProvider"/> instances for the current request.
-        /// </summary>
-        public virtual IList<IModelValidatorProvider> ValidatorProviders
```
dougbu: Wow, we were carting these two properties around for `TryUpdateModelAsync()` calls and that's it? No real scenario for controller code to change `ValidatorProviders` before calling `TryValidateModel()`?

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerArgumentBinder.cs
```diff
@@ -39,20 +39,20 @@ public class ControllerArgumentBinder : IControllerActionArgumentBinder
         }
 
         public Task<IDictionary<string, object>> BindActionArgumentsAsync(
-            ControllerContext context,
+            ControllerContext controllerContext,
```
dougbu: Renaming the parameters makes it harder to follow along at home.

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultModelBindingContext.cs
```diff
@@ -27,22 +30,24 @@ public DefaultModelBindingContext()
...
             string modelName)
         {
-            if (operationBindingContext == null)
+            if (actionContext == null)
             {
-                throw new ArgumentNullException(nameof(operationBindingContext));
+                throw new ArgumentNullException(nameof(actionContext));
             }
 
             if (metadata == null)
```
dougbu: Why is `valueProvider == null` allowed?

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultModelBindingContext.cs
```diff
@@ -113,7 +115,7 @@ public DefaultModelBindingContext()
             // to preserve the currrent state.
             if (modelMetadata.BindingSource != null && !modelMetadata.BindingSource.IsGreedy)
             {
-                ValueProvider = FilterValueProvider(_operationBindingContext.ValueProvider, modelMetadata.BindingSource);
+                ValueProvider = FilterValueProvider(_originalValueProvider, modelMetadata.BindingSource);
```
dougbu: Why isn't this using the `OriginalValueProvider` property?

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultModelBindingContext.cs
```diff
@@ -130,32 +132,16 @@ public DefaultModelBindingContext()
         }
 
         /// <inheritdoc />
-        public override NestedScope EnterNestedScope()
```
dougbu: Let's do a pass to reorder this class _separately_. This move doesn't do much more than make the review harder to read.

rynowak: I moved about 10 lines of code to comply with our coding standards. If I hadn't moved it, there's a good chance you might have suggested I move it :see_no_evil:

dougbu: 1. You split these methods from the other `EnterNestedScope()` overload.
2. There is no such guideline; no mention of member order in https://github.com/aspnet/Home/wiki/Engineering-guidelines#coding-guidelines. It's an informal policy we've imposed on new code and not something we rewrite classes to match.

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultModelBindingContext.cs
```diff
@@ -270,8 +269,15 @@ public override IValueProvider ValueProvider
         /// <inheritdoc />
         public override ValidationStateDictionary ValidationState
         {
-            get { return _state.ValidationState; }
-            set { _state.ValidationState = value; }
+            get { return _validationState; }
+            set
```
dougbu: Looks like this setter exists only for tests, when used together with the test-only parameterless constructor. Please add comments because it's too easy to mess the non-stacked property values up if the assumptions aren't stated.

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultObjectValidator.cs
```diff
@@ -14,34 +15,36 @@ public class DefaultObjectValidator : IObjectModelValidator
     {
         private readonly IModelMetadataProvider _modelMetadataProvider;
         private readonly ValidatorCache _validatorCache;
+        private readonly IModelValidatorProvider _validatorProvider;
 
         /// <summary>
         /// Initializes a new instance of <see cref="DefaultObjectValidator"/>.
         /// </summary>
         /// <param name="modelMetadataProvider">The <see cref="IModelMetadataProvider"/>.</param>
-        /// <param name="validatorCache">The <see cref="ValidatorCache"/>.</param>
```
dougbu: Strange to take the `ValidatorCache` out of DI. Why does the `DefaultObjectValidator` now own the instance?

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultObjectValidator.cs
```diff
@@ -51,14 +54,14 @@ public class DefaultObjectValidator : IObjectModelValidator
                 throw new ArgumentNullException(nameof(actionContext));
             }
 
-            if (validatorProvider == null)
+            if (validationState == null)
```
dougbu: The `ValidationVisitor` handles `validationState == null` just fine. Why does this condition now lead to an `ArgumentNullException`?

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/BodyModelBinder.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Linq;
```
dougbu: Can this `using` be removed now?

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/ComplexTypeModelBinder.cs
```diff
@@ -271,33 +271,6 @@ private bool CanValueBindAnyModelProperties(ModelBindingContext bindingContext)
...
-                if (rootValueProvider != null)
-                {
-                    valueProvider = rootValueProvider.Filter(bindingSource);
-                    if (valueProvider == null)
-                    {
-                        // Unable to find a value provider for this binding source. Binding will fail.
-                        return false;
-                    }
-                }
-            }
```
dougbu: Where did this block move to? Or, are you just no longer worried about the eventual binding failure?

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/DictionaryModelBinder.cs
```diff
@@ -81,8 +81,8 @@ public DictionaryModelBinder(IModelBinder keyBinder, IModelBinder valueBinder)
             }
 
             // Update the existing successful but empty ModelBindingResult.
-            var metadataProvider = bindingContext.OperationBindingContext.MetadataProvider;
-            var valueMetadata = metadataProvider.GetMetadataForType(typeof(TValue));
+            var elementMetadata = bindingContext.ModelMetadata.ElementMetadata;
+            var valueMetadata = elementMetadata.Properties[nameof(KeyValuePair<TKey, TValue>.Value)];
```
dougbu: Nice.

src/Microsoft.AspNetCore.Mvc.WebApiCompatShim/ApiController.cs
```diff
@@ -567,11 +567,10 @@ public void Validate<TEntity>(TEntity entity)
         /// </param>
         public void Validate<TEntity>(TEntity entity, string keyPrefix)
         {
-            var validatidationState = new ValidationStateDictionary();
+            var validationState = new ValidationStateDictionary();
```
dougbu: No, no. I'm positive we meant to introduce the new concept of `validatidation` states :trollface:

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/DefaultObjectValidatorTests.cs
```diff
@@ -466,12 +453,7 @@ public void Validate_ComplexType_IValidatableObject_CanUseRequestServices()
                 .Returns(new List<ValidationResult>());
 
             // Act
-            validator.Validate(
-                actionContext, 
-                validatorProvider, 
-                validationState, 
-                null, 
-                model.Object);
+            validator.Validate(actionContext, validationState, null, model.Object);
```
dougbu: minor: plz name the `null` guy

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/Binders/KeyValuePairModelBinderTest.cs
```diff
@@ -125,7 +125,7 @@ public class KeyValuePairModelBinderTest
             var binder = new KeyValuePairModelBinder<int, string>(innerBinder, innerBinder);
 
             // Act
-            var result = await binder.TryBindStrongModel<int>(bindingContext, innerBinder, "key");
+            var result = await binder.TryBindStrongModel<int>(bindingContext, innerBinder, "Key");
```
dougbu: Ah, previously case didn't matter but it does now. Correct?

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/Binders/KeyValuePairModelBinderTest.cs
```diff
@@ -212,21 +208,13 @@ private static DefaultModelBindingContext CreateContext()
             IValueProvider valueProvider,
             Type keyValuePairType)
         {
-            var metataProvider = new EmptyModelMetadataProvider();
+            var metadataProvider = new TestModelMetadataProvider();
```
dougbu: Does changing from an `EmptyModelMetadataProvider` to a `TestModelMetadataProvider` mean anything?

test/Microsoft.AspNetCore.Mvc.IntegrationTests/ModelBindingTestContext.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Mvc.ModelBinding;
+
+namespace Microsoft.AspNetCore.Mvc.IntegrationTests
+{
+    public class ModelBindingTestContext : ControllerContext
```
dougbu: Like it! Was getting concerned about reanimating `OperationBindingContext` as a test concept. But this is a nice hedge to reduce test code churn.

dougbu:
Suggest we'll need an announcement for those (few?) users that call `ModelBindingHelper` methods directly or who have written their own model binders.

dougbu:
:watch: for questions and lower-level comments. Liking the overall direction i.e. the boilerplate deletions.

src/Microsoft.AspNetCore.Mvc.Core/ControllerContext.cs
rynowak: Correct, and yes, no real scenario.

src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs
rynowak: Sure

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultModelBindingContext.cs
rynowak: fixed

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultModelBindingContext.cs
rynowak: Fixed

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultModelBindingContext.cs
rynowak: Added some comments. Really **all** of these are unit testing only btw.

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultObjectValidator.cs
rynowak: Because it would never be used anywhere else, and it can just construct the cache. Putting the cache in DI just adds complexity

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultObjectValidator.cs
rynowak: Sure, will revert this change.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/BodyModelBinder.cs
rynowak: yep

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/ComplexTypeModelBinder.cs
rynowak: This moved inside the binding context. I'm not worried about it because you'll fail to find any values in an empty collection

src/Microsoft.AspNetCore.Mvc.WebApiCompatShim/ApiController.cs
rynowak: :heart_eyes:

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/Binders/KeyValuePairModelBinderTest.cs
rynowak: Not really, I was just copy-pasting things

rynowak:
@dougbu updated

src/Microsoft.AspNetCore.Mvc.Core/ControllerBase.cs
```diff
@@ -1429,8 +1415,7 @@ public virtual SignOutResult SignOut(AuthenticationProperties properties, params
 
             ObjectValidator.Validate(
                 ControllerContext,
-                new CompositeModelValidatorProvider(ControllerContext.ValidatorProviders),
-                validationState: null,
+                validationState: null
```
dougbu: Won't compile unless you restore the comma.

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/Binders/KeyValuePairModelBinderTest.cs
dougbu: Unanswered. Am I missing anything here?

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultModelBindingContext.cs
dougbu: Sounds good.

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultModelBindingContext.cs
dougbu: This wasn't an important part of this PR but you haven't actually fixed anything yet. Either revert this part of the change or move `CreateBindingContext()` and the other `EnterNestedScope()` after the properties as well.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/ComplexTypeModelBinder.cs
dougbu: :ok:

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultObjectValidator.cs
dougbu: :ok:

dougbu:
:shipit: with a couple of fixes (especially [this](https://github.com/aspnet/Mvc/pull/4501#discussion-diff-60352310) and [this](https://github.com/aspnet/Mvc/pull/4501#discussion-diff-60858171)) -- as long as I'm not missing something [here](https://github.com/aspnet/Mvc/pull/4501#discussion-diff-60447549).

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/Binders/KeyValuePairModelBinderTest.cs
rynowak: Yeah sorry missed this one. There's no trick. `.Properties[string]` is case-sensitive (as all things should be)

PR #4502, opened 2016-04-19
[Perf] Check for Logger.IsEnabled in logging methods of MvcCoreLoggerExtensions 

### LINKED ISSUE #4466: MvcCoreLoggerExtensions.ActionMethodExecuted should check IsEnabled
`MvcCoreLoggerExtensions.ActionMethodExecuted` is doing some manipulation of its arguments without checking `logger.IsEnabled`.

This was showing up as around .5% of execution time for small MVC samples, so it's worth doing a pass.

We should fix this one and review the other logging methods to make sure they aren't doing unnecessary work

mnltejaswini:
Fixes #4466


rynowak:
:shipit:

Issue #4505, opened 2016-04-20
Model Binding goes into an infinite loop when the view model has a Func<>


sebastienros:
This is my model:

```
    public class ContentFieldInfo
    {
        public string FieldTypeName { get; set; }

        [BindNever]
        public Func<ContentPartFieldDefinition, object> Factory { get; set; }
    }
```

It's part of a bigger view model, but this is what makes the app fails when it's added to the view model.
If a request tries to model bind it in an action, the process never ends and keeps increasing memory. A memory dump shows a _cycle_ in `DefaultMetadataDetails`. It doesn't even breaks in the action, the controller is instantiated though.

I will provide a sample. Just filing now with the information I have in case something obvious comes to mind.

Eilon:
Any chance you want to take a crack at fixing it?

sebastienros:
Can confirm I can repro with MvcSandbox by just adding a `Func` property in a view model.

PR #4510, opened 2016-04-20
Prevent model binder resolution when not necessary

sebastienros:
1. `[BindNever]` was not taken into account while creating model binders.
2. Added a way to ignore model binding on specific types. By default adding 'Type' as it is very costly.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/Internal/ExcludeBindingMetadataProvider.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Internal/ExcludeBindingMetadataProvider.cs
@@ -49,26 +49,7 @@ namespace Microsoft.AspNetCore.Mvc.Internal
                 return;
             }
 
-            if (context.Key.MetadataKind == ModelMetadataKind.Property)
-            {
-                // BindingBehavior can fall back to attributes on the Container Type, but we should ignore
-                // attributes on the Property Type.
-                var bindingBehavior = context.PropertyAttributes.OfType<BindingBehaviorAttribute>().FirstOrDefault();
-                if (bindingBehavior == null)
-                {
-                    bindingBehavior =
-                        context.Key.ContainerType.GetTypeInfo()
-                        .GetCustomAttributes(typeof(BindingBehaviorAttribute), inherit: true)
-                        .OfType<BindingBehaviorAttribute>()
-                        .FirstOrDefault();
-                }
-
-                // Don't force the value if there is a specific attribute on it
-                if (bindingBehavior == null)
-                {
-                    context.BindingMetadata.IsBindingAllowed = false;
-                }
-            }
+            context.BindingMetadata.IsBindingAllowed = false;
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Mvc.Core/Internal/ExcludeBindingMetadataProvider.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Internal/ExcludeBindingMetadataProvider.cs
@@ -1,0 +1,74 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
+using Microsoft.AspNetCore.Mvc.ModelBinding.Metadata;
+
+namespace Microsoft.AspNetCore.Mvc.Internal
+{
+    /// <summary>
+    /// An <see cref="IBindingMetadataProvider"/> which configures <see cref="ModelMetadata.IsBindingAllowed"/> to
+    /// <c>false</c> for matching types.
+    /// </summary>
+    public class ExcludeBindingMetadataProvider : IBindingMetadataProvider
+    {
+        private readonly Type _type;
+
+        /// <summary>
+        /// Creates a new <see cref="ExcludeBindingMetadataProvider"/> for the given <paramref name="type"/>.
+        /// </summary>
+        /// <param name="type">
+        /// The <see cref="Type"/>. All properties of this <see cref="Type"/> will have
+        /// <see cref="ModelMetadata.IsBindingAllowed"/> set to <c>false</c>.
+        /// </param>
+        public ExcludeBindingMetadataProvider(Type type)
+        {
+            if (type == null)
+            {
+                throw new ArgumentNullException(nameof(type));
+            }
+
+            _type = type;
+        }
+
+        /// <inheritdoc />
+        public void CreateBindingMetadata(BindingMetadataProviderContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            // No-op if the metadata is not for the target type
+            if (!_type.IsAssignableFrom(context.Key.ModelType))
+            {
+                return;
+            }
+
+            if (context.Key.MetadataKind == ModelMetadataKind.Property)
+            {
+                // BindingBehavior can fall back to attributes on the Container Type, but we should ignore
+                // attributes on the Property Type.
+                var bindingBehavior = context.PropertyAttributes.OfType<BindingBehaviorAttribute>().FirstOrDefault();
+                if (bindingBehavior == null)
+                {
+                    bindingBehavior =
+                        context.Key.ContainerType.GetTypeInfo()
+                        .GetCustomAttributes(typeof(BindingBehaviorAttribute), inherit: true)
+                        .OfType<BindingBehaviorAttribute>()
+                        .FirstOrDefault();
+                }
+
+                // Don't force the value if there is a specific attribute on it
+                if (bindingBehavior == null)
+                {
+                    context.BindingMetadata.IsBindingAllowed = false;
+                }
+            }
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Mvc.Core/Internal/MvcCoreMvcOptionsSetup.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Internal/MvcCoreMvcOptionsSetup.cs
@@ -73,9 +73,9 @@ namespace Microsoft.AspNetCore.Mvc.Internal
             options.ValueProviderFactories.Add(new JQueryFormValueProviderFactory());
 
             // Set up metadata providers
+            options.ModelMetadataDetailsProviders.Add(new ExcludeBindingMetadataProvider(typeof(Type)));
             options.ModelMetadataDetailsProviders.Add(new DefaultBindingMetadataProvider(messageProvider));
             options.ModelMetadataDetailsProviders.Add(new DefaultValidationMetadataProvider());
-            options.ModelMetadataDetailsProviders.Add(new ExcludeBindingMetadataProvider(typeof(Type)));
 
             // Set up validators
             options.ModelValidatorProviders.Add(new DefaultModelValidatorProvider());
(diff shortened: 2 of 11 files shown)


sebastienros:
#4505

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
```diff
@@ -73,6 +73,11 @@ public IModelBinder CreateBinder(ModelBinderFactoryContext context)
 
         private IModelBinder CreateBinderCore(DefaultModelBinderProviderContext providerContext, object token)
         {
+            if (!providerContext.Metadata.IsBindingAllowed)
+            {
+                return NoOpBinder.Instance;
+            }
```
dougbu: Need tests to make sure `[BindNever]` is actually working. Suggest something in the integration tests project.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ExcludeBindingMetadataProvider.cs
```diff
@@ -0,0 +1,74 @@
...
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            // No-op if the metadata is not for the target type
+            if (!_type.IsAssignableFrom(context.Key.ModelType))
+            {
+                return;
+            }
+
+            if (context.Key.MetadataKind == ModelMetadataKind.Property)
```
dougbu: This means `Type[]` properties and similar collections of messy types are not excluded. Scary.

dougbu:
:watch:

src/Microsoft.AspNetCore.Mvc.Core/Internal/ExcludeBindingMetadataProvider.cs
sebastienros: For this specific `Type[]` example, the `Array` provider would then call into this one and get a no-op binder, same for collections of these types, and other complex types referencing `Type` like `delegates which triggered the issue in the first place.

What is scary in my opinion is that there might be other type that will trigger the same issue. So we might want to use this provider with more types for the default configuration, but I think the fact we have two solutions for these other cases (BindNever and this provider) is sufficient for any user scenario.

sebastienros:
:up::date:

Implemented the unit tests.
Not in [Design] phase anymore as it didn't get any "OMG don't do that"

rynowak:
> didn't get any "OMG don't do that"

You need to step up your game :laughing:

src/Microsoft.AspNetCore.Mvc.Core/Internal/ExcludeBindingMetadataProvider.cs
dougbu: :ok:

src/Microsoft.AspNetCore.Mvc.Core/Internal/ExcludeBindingMetadataProvider.cs
```diff
@@ -0,0 +1,74 @@
...
+using System.Reflection;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
+using Microsoft.AspNetCore.Mvc.ModelBinding.Metadata;
+
+namespace Microsoft.AspNetCore.Mvc.Internal
+{
+    /// <summary>
+    /// An <see cref="IBindingMetadataProvider"/> which configures <see cref="ModelMetadata.IsBindingAllowed"/> to
+    /// <c>false</c> for matching types.
+    /// </summary>
```
rynowak: IMO this shouldn't be `.Internal`. Users might hit similar issues, and we point them at this as the solution.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ExcludeBindingMetadataProvider.cs
```diff
@@ -0,0 +1,74 @@
...
+                // attributes on the Property Type.
+                var bindingBehavior = context.PropertyAttributes.OfType<BindingBehaviorAttribute>().FirstOrDefault();
+                if (bindingBehavior == null)
+                {
+                    bindingBehavior =
+                        context.Key.ContainerType.GetTypeInfo()
+                        .GetCustomAttributes(typeof(BindingBehaviorAttribute), inherit: true)
+                        .OfType<BindingBehaviorAttribute>()
+                        .FirstOrDefault();
+                }
```
rynowak: I don't get why any of this is needed. Ordering can solve this problem if the user needs to override this behavior. I'd rather not have scary reflection that doesn't follow the typical pattern.

src/Microsoft.AspNetCore.Mvc.Core/Internal/MvcCoreMvcOptionsSetup.cs
```diff
@@ -75,6 +75,7 @@ public void Configure(MvcOptions options)
             // Set up metadata providers
             options.ModelMetadataDetailsProviders.Add(new DefaultBindingMetadataProvider(messageProvider));
             options.ModelMetadataDetailsProviders.Add(new DefaultValidationMetadataProvider());
+            options.ModelMetadataDetailsProviders.Add(new ExcludeBindingMetadataProvider(typeof(Type)));
```
rynowak: Add a comment here explaining what this is for and what it does.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ExcludeBindingMetadataProvider.cs
rynowak: I don't see why this check is needed, checking `ModelType` should we all you need. In your example when you create a binder for `Type[]` the array binder will call back here with the metadata for `Type`, but for `ModelMetadataKind.Type`.

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/Metadata/ExcludeBindingMetadataProviderTest.cs
```diff
@@ -0,0 +1,96 @@
...
+            provider.CreateBindingMetadata(context);
+
+            // Assert
+            Assert.Equal(initialValue, context.BindingMetadata.IsBindingAllowed);
+        }
+
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public void IsBindingAllowed_IsFalse_ForMatchingTypeMetadata(bool initialValue)
```
rynowak: This is property metadata, not type metadata

test/Microsoft.AspNetCore.Mvc.IntegrationTests/SimpleTypeModelBinderIntegrationTest.cs
```diff
@@ -538,6 +538,42 @@ public class SimpleTypeModelBinderIntegrationTest
             Assert.Equal(new[] { "line 1", "line 2" }, entry.RawValue);
         }
 
+        [Theory]
+        [MemberData(nameof(PersonStoreData))]
+        public async Task BindParameter_WithBindNever_IsNotBound(Dictionary<string, StringValues> personStore)
```
rynowak: These tests aren't really related to `SimpleTypeModelBinder`- make a new test class

src/Microsoft.AspNetCore.Mvc.Core/Internal/ExcludeBindingMetadataProvider.cs
rynowak: We should add integration tests for:
- top-level `Type`
- `Type` as a property of a POCO
- `Type` as the element type of a collection
- `Type` as an element type of a dictionary

src/Microsoft.AspNetCore.Mvc.Core/Internal/MvcCoreMvcOptionsSetup.cs
```diff
@@ -73,6 +74,10 @@ public void Configure(MvcOptions options)
             options.ValueProviderFactories.Add(new JQueryFormValueProviderFactory());
 
             // Set up metadata providers
+
+            // Don't bind the Type class by default as it's expensive. A user can override this behavior
+            // by altering the collection of providers.
+            options.ModelMetadataDetailsProviders.Add(new ExcludeBindingMetadataProvider(typeof(Type)));
             options.ModelMetadataDetailsProviders.Add(new DefaultBindingMetadataProvider(messageProvider));
```
dougbu: A blank line might help make it clear the "Don't bind" comment applies only to one line of code.

dougbu:
:shipit: once Travis, AppVeyor, or your local command-line build succeeds.

rynowak:
:watch: maybe my comment was buried elsewhere but this deserves 1-2 integration tests

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/ModelBinderFactoryTest.cs
```diff
@@ -67,9 +67,46 @@ public void CreateBinder_CreatesNoOpBinder_WhenPropertyDoesntHaveABinder()
...
+
+            var context = new ModelBinderFactoryContext()
+            {
+                Metadata = metadataProvider.GetMetadataForProperty(typeof(Widget), nameof(Widget.Id)),
+            };
+
+            // Act
+            var result = factory.CreateBinder(context);
+
+            // Assert
```
dougbu: Suggest asserting that `result.GetType().FullName` contains "NoOp". Right now this test mainly demonstrates the `TestModelBinderProvider` isn't used.

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/ModelBinderFactoryTest.cs
```diff
@@ -67,9 +67,46 @@ public void CreateBinder_CreatesNoOpBinder_WhenPropertyDoesntHaveABinder()
...
         [Fact]
+        public void CreateBinder_CreatesNoOpBinder_WhenPropertyBindingIsNotAllowed()
+        {
+            // Arrange
+            var metadataProvider = new TestModelMetadataProvider();
+            metadataProvider
+                .ForProperty<Widget>(nameof(Widget.Id))
+                .BindingDetails(m => m.IsBindingAllowed = false);
+
+            var modelBinder = Mock.Of<IModelBinder>();
```
dougbu: Can avoid the mocking e.g. `new ByteArrayModelBinder()` since it just doesn't matter.

test/Microsoft.AspNetCore.Mvc.IntegrationTests/ExcludeBindingMetadataProviderIntegrationTest.cs
```diff
@@ -0,0 +1,110 @@
...
+
+namespace Microsoft.AspNetCore.Mvc.IntegrationTests
+{
+    public class ExcludeBindingMetadataProviderIntegrationTest
+    {
+        [Fact]
+        public async Task BindParameter_WithTypeProperty_IsNotBound()
+        {
+            // Arrange
+            var options = new TestOptionsManager<MvcOptions>();
```
dougbu: Don't need the options manager. Can use `new MvcOptions()`.

sebastienros:
:up::date:

Feedback applied

dougbu:
:shipit: if @rynowak is moderately pleased and once Travis, AppVeyor, or your local command-line build succeeds.

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/ModelBinderFactoryTest.cs
rynowak: We could also just make it a `.Internal` class - that would be less ugly than this.

rynowak:
:shipit:

PR #4511, opened 2016-04-21
[Fixes 4509] Stop registering disposable objects in our controller helper methods

javiercn:


javiercn:
Incorrect base for the PR

PR #4513, opened 2016-04-21
React to routing cleanup

rynowak:
/cc @pranavkm

### DIFF
--- a/test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/AttributeRouteTest.cs
+++ b/test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/AttributeRouteTest.cs
@@ -10,14 +10,12 @@ using Microsoft.AspNetCore.Mvc.Abstractions;
 using Microsoft.AspNetCore.Mvc.Infrastructure;
 using Microsoft.AspNetCore.Mvc.Routing;
 using Microsoft.AspNetCore.Routing;
-using Microsoft.AspNetCore.Routing.Constraints;
-using Microsoft.AspNetCore.Routing.Internal;
 using Microsoft.AspNetCore.Routing.Template;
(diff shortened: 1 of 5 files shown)


test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/AttributeRouteTest.cs
```diff
@@ -230,35 +224,33 @@ public void AttributeRoute_GetEntries_CreatesLinkGenerationEntry_WithDefault()
...
-                entries.LinkGenerationEntries,
+                builder.OutboundEntries,
                 e =>
                 {
-                    Assert.NotNull(e.Binder);
                     Assert.Empty(e.Constraints);
                     Assert.Equal(new RouteValueDictionary(new { slug = "hello" }), e.Defaults);
-                    Assert.Equal(RoutePrecedence.ComputeGenerated(e.Template), e.GenerationPrecedence);
-                    Assert.Equal("BLOG_INDEX", e.Name);
+                    Assert.Equal(RoutePrecedence.ComputeOutbound(e.RouteTemplate), e.Precedence);
```
pranavkm: Unrelated to this change - shouldn't this verify the actual value (e.g. `Assert.Equal(4, e.Precedence);`) rather than verify it against the computed value?

pranavkm:
:shipit:

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/AttributeRouteTest.cs
rynowak: There are already good tests for computing precedences, this is just to verify that the correct method was called to compute it. I don't want to hardcode the result of an algothm

PR #4515, opened 2016-04-21
[Design] Make ValueProviders lazy

rynowak:
/cc @eilon @dougbu

If you guys like it I'll add caching, tests, and fix the broken tests (just some tryupdatemodel tests)

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadataProvider.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadataProvider.cs
@@ -16,6 +16,7 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding.Metadata
     {
         private readonly TypeCache _typeCache = new TypeCache();
         private readonly Func<ModelMetadataIdentity, ModelMetadataCacheEntry> _cacheEntryFactory;
+        private readonly ModelMetadataCacheEntry _metadataCacheEntryForObjectType;
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Mvc.Core/ControllerBase.cs
```diff
@@ -1482,5 +1486,12 @@ public virtual SignOutResult SignOut(AuthenticationProperties properties, params
                 model: model);
             return ModelState.IsValid;
         }
+
+        private Task<CompositeValueProvider> GetValueProviderAsync()
```
rynowak: The other pivot here that I considered is to add `ICompositeValueProviderFactory` as a public property on `ControllerBase`. This would improve the testability and help make these overloads of `TryUpdateModel` that take a value provider more useful.

However it would add yet another thing to the controller surface area...

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ICompositeValueProviderFactory.cs
```diff
@@ -1,12 +1,25 @@
...
+
 namespace Microsoft.AspNetCore.Mvc.ModelBinding
 {
     /// <summary>
-    /// Represents an aggregate of <see cref="IValueProviderFactory"/>.
+    /// Uses configured <see cref="IValueProviderFactory"/> instances to create a
+    /// <see cref="CompositeValueProvider"/> for the current request.
     /// </summary>
-    public interface ICompositeValueProviderFactory : IValueProviderFactory
+    public interface ICompositeValueProviderFactory
```
rynowak: For some reason this interface already exists and wasn't used anywhere.

PR #4518, opened 2016-04-21
[Perf] Cache the ModelMetadata for known type "object"

### LINKED ISSUE #4377: Optimize creation of root VDD
Sampling data from HelloWorldMvc

image

We're creating this VDD with model metadata for `object` as a sensible default. We can do something smarter here like caching it to avoid the lookup.

mnltejaswini:
Fixes #4377 

Scenario : https://github.com/aspnet/Performance/tree/dev/testapp/HelloWorldMvc
Before Optimization
-----------------------
image

After Optimization
----------------------
image

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadataProvider.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadataProvider.cs
@@ -16,6 +16,7 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding.Metadata
     {
         private readonly TypeCache _typeCache = new TypeCache();
         private readonly Func<ModelMetadataIdentity, ModelMetadataCacheEntry> _cacheEntryFactory;
+        private readonly ModelMetadataCacheEntry _metadataCacheEntryForObjectType;
 
         /// <summary>
         /// Creates a new <see cref="DefaultModelMetadataProvider"/>.
@@ -26,6 +27,8 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding.Metadata
             DetailsProvider = detailsProvider;
 
             _cacheEntryFactory = CreateCacheEntry;
+
+            _metadataCacheEntryForObjectType = GetMetadataCacheEntryForObjectType();
         }
 
         /// <summary>
@@ -41,14 +44,13 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding.Metadata
                 throw new ArgumentNullException(nameof(modelType));
             }
 
-            var key = ModelMetadataIdentity.ForType(modelType);
-
-            var cacheEntry = _typeCache.GetOrAdd(key, _cacheEntryFactory);
+            var cacheEntry = GetCacheEntry(modelType);
 
             // We're relying on a safe race-condition for Properties - take care only
             // to set the value onces the properties are fully-initialized.
             if (cacheEntry.Details.Properties == null)
             {
+                var key = ModelMetadataIdentity.ForType(modelType);
                 var propertyDetails = CreatePropertyDetails(key);
 
                 var properties = new ModelMetadata[propertyDetails.Length];
@@ -71,10 +73,28 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding.Metadata
                 throw new ArgumentNullException(nameof(modelType));
             }
 
-            var key = ModelMetadataIdentity.ForType(modelType);
+            var cacheEntry = GetCacheEntry(modelType);
 
-            var cacheEntry = _typeCache.GetOrAdd(key, _cacheEntryFactory);
             return cacheEntry.Metadata;
+        }
+
+        private ModelMetadataCacheEntry GetCacheEntry(Type modelType)
+        {
+            ModelMetadataCacheEntry cacheEntry;
+
+            // Perf: We cached model metadata cache entry for "object" type to save ConcurrentDictionary lookups.
+            if (modelType == typeof(object))
+            {
+                cacheEntry = _metadataCacheEntryForObjectType;
+            }
+            else
+            {
+                var key = ModelMetadataIdentity.ForType(modelType);
+
+                cacheEntry = _typeCache.GetOrAdd(key, _cacheEntryFactory);
+            }
+
+            return cacheEntry;
         }
 
         private ModelMetadataCacheEntry CreateCacheEntry(ModelMetadataIdentity key)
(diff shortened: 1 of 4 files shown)


rynowak:
I don't think the data included here is sufficient to conclude anything, 6ms of sampling data is a tiny amount. I'll start reviewing the changes and let's talk tomorrow about how to gather more data.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadataProvider.cs
```diff
@@ -71,6 +74,11 @@ public virtual ModelMetadata GetMetadataForType(Type modelType)
                 throw new ArgumentNullException(nameof(modelType));
             }
 
+            if (modelType == typeof(object) && _cachedMetadataForObjectType != null)
```
rynowak: Why is the null check needed here?

rynowak: This at least needs a comment about why it's important

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/ViewDataDictionaryControllerPropertyActivator.cs
```diff
@@ -49,7 +49,6 @@ private PropertyActivator<ControllerContext>[] GetPropertiesToActivate(Type type
 
         private ViewDataDictionary GetViewDataDictionary(ControllerContext context)
         {
-            var serviceProvider = context.HttpContext.RequestServices;
```
rynowak: :heart_eyes_cat:

mnltejaswini:
:up::date:

rynowak:
Is there updated perf data for this? Come by if you want help collecting data

mnltejaswini:
@rynowak  I updated the profiles with more sampling data

dougbu:
Not getting a warm feeling from looking just at the time spent activating `ViewDataDictionary` properties in controllers. From #4377 we know that activation accounts for ~1.8% of execution time and the above data shows we save ~20% of that. But I'd prefer a more direct comparison.

On the other hand, I suspect we get the metadata for `typeof(object)` enough other places that there may be a larger change in the overall time spent in`GetMetadataForType()` relative to the rest of the app.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadataProvider.cs
```diff
@@ -84,6 +93,13 @@ private ModelMetadataCacheEntry CreateCacheEntry(ModelMetadataIdentity key)
             return new ModelMetadataCacheEntry(metadata, details);
         }
 
+        private ModelMetadata GetMetadataForObjectType()
+        {
+            var key = ModelMetadataIdentity.ForType(typeof(object));
+            var entry = CreateCacheEntry(key);
```
dougbu: Suggest changing `GetMetadataForProperties()` to call `GetMetadataForType()`. Right now, the cache entry for `typeof(object)` may be created anyhow, duplicating this work and uselessly hitting the cache despite the shortcut above.

dougbu:
:watch:

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Metadata/DefaultModelMetadataProvider.cs
rynowak: I don't understand this feedback - how would returning the cached type-metadata for object be correct when we need to return a property-metadata?

dougbu: `GetMetadataForProperties()` first gets `modelType`'s (the container's) metadata, duplicating most of `GetMetadataForType()`. My feedback is to maintain the consistency or just to use `GetMetadataForType()` for that part of `GetMetadataForProperties()`.

mnltejaswini: GetMetadataForProperties() doesn't use the DefaultModelMetadata for the container's ModelType. It uses DefaultModelMetadataDetails that has less info that the Metadata. Wondering if we can reuse the Object Metadata if any of the properties are of type object?

dougbu: > Wondering if we can reuse the Object Metadata if any of the properties are of type `object`?

No, property metadata is a completely different beast from type metadata. I'm wondering if we can save and reuse this `ModelMetadataCacheEntry` for when anything calls `GetMetadataForProperties(typeof(object))`. See [line 49](https://github.com/aspnet/Mvc/pull/4518/files#diff-c8fa65ecc8077ca10cd5f923c0082de2R49).

mnltejaswini:
:up::date:

rynowak:
Data looks much more convincing! Thanks :+1:

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/Metadata/DefaultModelMetadataProviderTest.cs
```diff
@@ -45,6 +45,24 @@ public void GetMetadataForType_Cached()
...
+            // Act
+            var metadata1 = Assert.IsType<DefaultModelMetadata>(provider.GetMetadataForType(typeof(object)));
+            var metadata2 = Assert.IsType<DefaultModelMetadata>(provider.GetMetadataForType(typeof(object)));
+
+            // Assert
+            Assert.Same(metadata1, metadata2);
+            Assert.Same(metadata1.Attributes, metadata2.Attributes);
+            Assert.Same(metadata1.BindingMetadata, metadata2.BindingMetadata);
+            Assert.Same(metadata1.DisplayMetadata, metadata2.DisplayMetadata);
+            Assert.Same(metadata1.ValidationMetadata, metadata2.ValidationMetadata);
```
dougbu: Don't bother with the additional assertions. Nothing can be different between properties of `metadata1` and `metadata2` once you've confirmed they are the same.

In addition, the `Assert.IsType()` checks definitely don't belong w/in `// Act`. Why are they needed at all?

dougbu:
:shipit: but with fewer assertions.

PR #4519, opened 2016-04-21
Filters provided by filter providers are made to never be cached

### LINKED ISSUE #4504: Possible double-execution of filter providers
We're seeing a filter provider called twice per-request in an orchard scenario

Stacks:

kichalla:
[Fixes #4504] Possible double-execution of filter providers

@rynowak @sebastienros


src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvokerCache.cs
```diff
@@ -85,7 +91,7 @@ public Entry GetCacheEntry(ControllerContext controllerContext)
             }
             else
             {
-                entry = new Entry(null, items, executor);
+                entry = new Entry(null, staticallyDefinedFilterItems, executor);
             }
```
rynowak: This logic, along with `allFiltersCached` needs to go away. We just want to store the items. This is a bit of a step back in perf, but we need to call the providers every request.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvokerCache.cs
```diff
@@ -56,22 +56,28 @@ public Entry GetCacheEntry(ControllerContext controllerContext)
                 return entry;
```
rynowak: Which method is called by the invoker? For instance calling `GetFilters` can still double-execute the providers. I think we've got more cleanup to do here.

src/Microsoft.AspNetCore.Mvc.Core/Internal/FilterItemOrderComparer.cs
```diff
@@ -1,34 +0,0 @@
-// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-using System;
-using System.Collections.Generic;
-using Microsoft.AspNetCore.Mvc.Filters;
-
-namespace Microsoft.AspNetCore.Mvc.Internal
-{
-    public class FilterItemOrderComparer : IComparer<FilterItem>
```
rynowak: Was this unused?

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvokerCache.cs
rynowak: `GetFilters` and `GetControllerActionMethodExecutor` should we deleted. We do not want to look up the cache entry more than once per request.

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ControllerActionInvokerCacheTest.cs
```diff
@@ -186,8 +186,80 @@ public void GetControllerActionMethodExecutor_CachesExecutor()
             Assert.Same(executor1, executor2);
         }
 
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public void GetFilters_FiltersFromFilterProviders_AreNeverCached(bool reusable)
```
rynowak: `GetFilters_FiltersAddedByAFilterProvider_AreNeverCached`

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ControllerActionInvokerCacheTest.cs
```diff
@@ -186,8 +186,80 @@ public void GetControllerActionMethodExecutor_CachesExecutor()
             Assert.Same(executor1, executor2);
         }
 
+        [Theory]
```
rynowak: I'd also like to see a test that verifies how many times a filter provider is called.

rynowak:
:watch:

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ControllerActionInvokerCacheTest.cs
kichalla: This test covers the scenario. The filter provider adds a new filter item always to the result. If there were duplicate items the count of item would be incorrect.

src/Microsoft.AspNetCore.Mvc.Core/Internal/FilterItemOrderComparer.cs
kichalla: yes right

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvokerCache.cs
kichalla: > but we need to call the providers every request.

The providers are getting called for every request via the `GetFiltersFromEntry` method.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvokerCache.cs
kichalla: https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/Internal/FilterActionInvoker.cs#L126-L127

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ControllerActionInvokerCacheTest.cs
rynowak: Right, but this is a unit test for a method that isn't called by product code :trollface:

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvokerCache.cs
rynowak: Which means that on the first request they are called twice.

kichalla:
:up: :date:

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvokerCache.cs
```diff
@@ -45,96 +45,59 @@ private InnerCache CurrentCache
...
-                entry = new Entry(filters, null, executor);
-            }
-            else
-            {
-                entry = new Entry(null, items, executor);
-            }
-
-            cache.Entries.TryAdd(actionDescriptor, entry);
-            return entry;
+            return new EntryInfo(allFilters, cacheEntry);
```
rynowak: This is looking good!

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvokerCache.cs
```diff
@@ -186,25 +147,38 @@ public InnerCache(int version)
...
                 ActionMethodExecutor = executor;
             }
-            public IFilterMetadata[] Filters { get; }
 
             public List<FilterItem> FilterItems { get; }
 
             public ObjectMethodExecutor ActionMethodExecutor { get; }
         }
+
+        public struct EntryInfo
```
rynowak: Let's call this something else ... `ControllerActionInvokerState`?

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvokerCache.cs
```diff
@@ -186,25 +147,38 @@ public InnerCache(int version)
...
+                IFilterMetadata[] allFilters,
+                Entry cacheEntry)
+            {
+                AllFilters = allFilters;
+                CacheEntry = cacheEntry;
+            }
+
+            public IFilterMetadata[] AllFilters { get; }
+
+            public Entry CacheEntry { get; set; }
```
rynowak: We don't need the cache entry here, just the objectmethodexecutor

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvokerCache.cs
```diff
@@ -186,25 +147,38 @@ public InnerCache(int version)
...
+        {
+            public EntryInfo(
+                IFilterMetadata[] allFilters,
+                Entry cacheEntry)
+            {
+                AllFilters = allFilters;
+                CacheEntry = cacheEntry;
+            }
+
+            public IFilterMetadata[] AllFilters { get; }
```
rynowak: Just call this `Filters`

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvokerCache.cs
```diff
@@ -186,25 +147,38 @@ public InnerCache(int version)
...
             }
 
-            public ConcurrentDictionary<ActionDescriptor, Entry> Entries { get; } = 
+            public ConcurrentDictionary<ActionDescriptor, Entry> Entries { get; } =
                 new ConcurrentDictionary<ActionDescriptor, Entry>();
 
             public int Version { get; }
         }
 
         public struct Entry
```
rynowak: This doesn't need to be public anymore

rynowak:
:watch: for minor feedback, looking good overall

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvokerCache.cs
kichalla: Sure

kichalla:
:up: :date:

rynowak:
:shipit: - do you know if this is approved for RC2?

kichalla:
@rynowak actually i haven't checked. will do now.

kichalla:
a259638d79ed5c02b8a2805234c93dd83d739cf5

PR #4522, opened 2016-04-22
Simplify assembly load on CoreCLR

### LINKED ISSUE #4328: Simplify assembly load on CoreCLR
Just use the default load context for razor views. We get no benefit from having a custom load context. Now that we can load things directly into the default load context, we don't need it.

/cc @pranavkm

pranavkm:
Fixes #4328

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultRoslynCompilationService.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultRoslynCompilationService.cs
@@ -2,17 +2,14 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Reflection.PortableExecutable;
-#if NETSTANDARD1_5
-using System.Runtime.Loader;
-#endif
 using System.Text;
+using System.Threading;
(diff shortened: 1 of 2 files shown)


pranavkm:
cc @davidfowl \ @NTaylorMullen

src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultRoslynCompilationService.cs
```diff
@@ -260,24 +261,27 @@ private List<MetadataReference> GetApplicationReferences()
                     continue;
                 }
 
-                metadataReferences.AddRange(referencePaths.Select(CreateMetadataFileReference));
+                foreach (var path in referencePaths)
+                {
+                    if (libraryPaths.Add(path))
```
NTaylorMullen: This a separate issue?

NTaylorMullen:
:shipit:

davidfowl:
:shipit:

src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultRoslynCompilationService.cs
```diff
@@ -247,6 +247,7 @@ private List<MetadataReference> GetApplicationReferences()
                 return metadataReferences;
             }
 
+            var libraryPaths = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```
davidfowl: Can you file an issue to make sure we don't hold onto the dependency context forever?

PR #4524, opened 2016-04-22
[Perf] Replace Activator.CreateInstance with cached delegate

### LINKED ISSUE #4470: Review use of Activator.CreateInstance

mnltejaswini:
Fixes #4470 
Places that are fixed are 
1. ControllerActionExecutor.cs(61):
2.DefaultRazorPageFactoryProvider.cs
3.ComplexTypeModelBinder.cs(348):
4.CollectionModelBinder.cs(150):

1. Benefit in using the cached delegate in .DefaultRazorPageFactoryProvider.cs
Before Optimization
--------------------------
image

After Optimization
-----------------------

image

2. Benefit in using cached delegate in ComplexTypeModelBinder.cs and ControllerActionExecutor.cs(61):
Scenario: https://github.com/aspnet/Performance/tree/dev/testapp/BigModelBinding (slight change in the controller to accept a value type argument with no default value)

Before Optimization
-------------------------
image

After Optimization
------------------------
image

In most cases, we do invoke default constructor which is relatively inexpensive as compared to parameterized constructors. However, Activator does have a ActivatorCache with size limit as 16. With large number of views or arguments/models there is a possibility of not hitting the cache and going through the slow path of security checks etc. This fix proactively prevents those costs.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionExecutor.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionExecutor.cs
@@ -16,7 +16,7 @@ namespace Microsoft.AspNetCore.Mvc.Internal
             object instance,
             IDictionary<string, object> actionArguments)
         {
-            var orderedArguments = PrepareArguments(actionArguments, actionMethodExecutor.MethodInfo.GetParameters());
+            var orderedArguments = PrepareArguments(actionArguments, actionMethodExecutor);
             return ExecuteAsync(actionMethodExecutor, instance, orderedArguments);
         }
 
@@ -30,8 +30,9 @@ namespace Microsoft.AspNetCore.Mvc.Internal
 
         public static object[] PrepareArguments(
             IDictionary<string, object> actionParameters,
-            ParameterInfo[] declaredParameterInfos)
+            ObjectMethodExecutor actionMethodExecutor)
         {
+            var declaredParameterInfos = actionMethodExecutor.MethodInfo.GetParameters();
             var count = declaredParameterInfos.Length;
             if (count == 0)
             {
@@ -58,7 +59,7 @@ namespace Microsoft.AspNetCore.Mvc.Internal
                         if (defaultValueAttribute?.Value == null)
                         {
                             value = parameterInfo.ParameterType.GetTypeInfo().IsValueType
-                                ? Activator.CreateInstance(parameterInfo.ParameterType)
+                                ? actionMethodExecutor.CreateParameterObject(index)
                                 : null;
                         }
                         else
--- a/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/CollectionModelBinder.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/CollectionModelBinder.cs
@@ -7,6 +7,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
+using System.Linq.Expressions;
 using System.Reflection;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Mvc.Internal;
@@ -20,6 +21,8 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding.Binders
     /// <typeparam name="TElement">Type of elements in the collection.</typeparam>
     public class CollectionModelBinder<TElement> : ICollectionModelBinder
     {
+        private Func<object> _modelCreate;
+
         /// <summary>
         /// Creates a new <see cref="CollectionModelBinder{TElement}"/>.
         /// </summary>
@@ -147,7 +150,14 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding.Binders
         /// <returns>An instance of <paramref name="targetType"/>.</returns>
         protected object CreateInstance(Type targetType)
         {
-            return Activator.CreateInstance(targetType);
+            if (_modelCreate == null)
+            {
+                _modelCreate = Expression.
+                    Lambda<Func<object>>(Expression.New(targetType)).
+                    Compile();
+            }
+
+            return _modelCreate();
         }
 
         // Used when the ValueProvider contains the collection to be bound as a single element, e.g. the raw value
--- a/src/Microsoft.AspNetCore.Mvc.Razor/Internal/CompilerCacheResult.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor/Internal/CompilerCacheResult.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq.Expressions;
 using Microsoft.AspNetCore.Mvc.Razor.Compilation;
 using Microsoft.Extensions.Primitives;
 
@@ -40,6 +41,7 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Internal
             CompilationResult = compilationResult;
             Success = true;
             ExpirationTokens = expirationTokens;
+            PageCreate = Expression.Lambda<Func<object>>(Expression.New(compilationResult.CompiledType)).Compile();
         }
 
         /// <summary>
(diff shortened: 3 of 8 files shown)


davidfowl:
What does before and after look like performance wise?

mnltejaswini:
@davidfowl  I updated the original comments with some sampling data.

rynowak:
@davidfowl - activator has a cache for 16 types. With Models, Views, and Controllers (all instantiated by us) you hit that limit pretty quickly and go down the slow path. We're fixing the ones that are on the super common path provided we don't have to implement new caches.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionExecutor.cs
```diff
@@ -58,7 +59,7 @@ public static class ControllerActionExecutor
                         if (defaultValueAttribute?.Value == null)
                         {
                             value = parameterInfo.ParameterType.GetTypeInfo().IsValueType
-                                ? Activator.CreateInstance(parameterInfo.ParameterType)
+                                ? actionMethodExecutor.CreateParameterObject(index)
```
rynowak: A better and simpler way to do this would probably be to add support to `ObjectMethodExecutor` for storing a 'default value' per parameter. We're just dealing with value types here so it should be safe. 

We should also include the result of the call to `MethodInfo.GetParameters()` which allocates.

Notice that this code is looking for `DefaultValueAttribute`, we should put that check up front as well.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/CollectionModelBinder.cs
```diff
@@ -20,6 +21,8 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding.Binders
     /// <typeparam name="TElement">Type of elements in the collection.</typeparam>
     public class CollectionModelBinder<TElement> : ICollectionModelBinder
     {
+        private Func<object> _modelCreate;
```
rynowak: `_modelCreator`

src/Microsoft.AspNetCore.Mvc.Razor/Internal/CompilerCacheResult.cs
```diff
@@ -75,5 +78,11 @@ public CompilerCacheResult(IList<IChangeToken> expirationTokens)
         /// Gets a value that determines if the view was successfully found and compiled.
         /// </summary>
         public bool Success { get; }
+
+        /// <summary>
+        /// Gets a Func to create the view.
+        /// </summary>
+        /// <remarks>This property is not available when <see cref="Success"/> is <c>false</c>.</remarks>
+        public Func<object> PageCreate { get; }
```
rynowak: `PageCreator` or `PageFactory`

src/Microsoft.AspNetCore.Mvc.Razor/Internal/CompilerCacheResult.cs
```diff
@@ -58,6 +60,7 @@ public CompilerCacheResult(IList<IChangeToken> expirationTokens)
             CompilationResult = default(CompilationResult);
             Success = false;
```
rynowak: Can we made `Success` a computed field? Worried about the size of this struct growing too large

src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultRazorPageFactoryProvider.cs
```diff
@@ -74,15 +74,15 @@ public RazorPageFactoryResult CreateFactory(string relativePath)
...
+        /// <returns>A factory for <see cref="IRazorPage"/>.</returns>
+        protected virtual Func<IRazorPage> GetPageFactory(Func<object> pageCreate, string relativePath)
         {
             return () =>
             {
-                var page = (IRazorPage)Activator.CreateInstance(compiledType);
+                var page = (IRazorPage)pageCreate();
                 page.Path = relativePath;
                 return page;
             };
```
rynowak: /cc @pranavkm - should we simplify this further now that the compilation result has a `Func<object>`

src/Microsoft.AspNetCore.Mvc.Razor/Internal/CompilerCacheResult.cs
rynowak: /cc @pranavkm - do you have a concern about coupling by introducing a `Func<object>` at this level? This basically stipulates that everything we return from compilation has the have a `new()`

dougbu: @rynowak `Activator` also requires the parameterless constructor. The only new (pardon the pun) restriction I see relates to the corner case where `Activator` has sufficient access to invoke a non-`public` constructor.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/CollectionModelBinder.cs
```diff
@@ -147,7 +150,14 @@ protected virtual object CreateEmptyCollection(Type targetType)
         /// <returns>An instance of <paramref name="targetType"/>.</returns>
         protected object CreateInstance(Type targetType)
         {
-            return Activator.CreateInstance(targetType);
+            if (_modelCreate == null)
+            {
+                _modelCreate = Expression.
+                    Lambda<Func<object>>(Expression.New(targetType)).
+                    Compile();
```
dougbu: It's not a rule but please follow the convention used in the rest of this repo: Place member access operators after the newlines.

src/Microsoft.AspNetCore.Mvc.Razor/Internal/CompilerCacheResult.cs
pranavkm: This is a cache specific type (`CompilerCacheResult`) and not the actual `CompilationResult` so this seems ok.

src/Microsoft.AspNetCore.Mvc.Razor/Internal/CompilerCacheResult.cs
```diff
@@ -75,5 +78,11 @@ public CompilerCacheResult(IList<IChangeToken> expirationTokens)
         /// Gets a value that determines if the view was successfully found and compiled.
         /// </summary>
         public bool Success { get; }
+
+        /// <summary>
+        /// Gets a Func to create the view.
```
pranavkm: > Get a delegate that creates an instance of the `<see cref="CompiledType" />`.

src/Microsoft.AspNetCore.Mvc.Razor/Internal/CompilerCacheResult.cs
mnltejaswini: @rynowak  and @pranavkm  I don't see any value of having the CompilationResult in the CompilerCacheResult once I create the PageCreator delegate. Is it safe to remove that property from the struct?

rynowak: Is the `Func<IRazorPage>` returned from pagefactory also cached? Who's ultimately holding on the to the instantiates the page?

src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultRazorPageFactoryProvider.cs
pranavkm: Yeah. We could cache this as part of the compiled expression.

src/Microsoft.AspNetCore.Mvc.Razor/Internal/CompilerCacheResult.cs
mnltejaswini: Yes, Func<IRazorPage> is set as a property of ViewLocationCacheItem which arte entries of a cache in RazorViewEngine.cs

mnltejaswini:
This PR is split into 3 separate PR for isolation. Hence this will be closed

PR #4531, opened 2016-04-26
Make ValueProvider creation lazy

rynowak:
We want this change to avoid MVC eagerly reading the form. This is good
for general perf and also for scenarios where you want read the body
yourself (large file uploads).

We DO have scenarios where you want to configure the value providers
per-request or also to change the limits on the value providers (form) so
it's worth keeping these around on the context.


src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerArgumentBinder.cs
```diff
@@ -111,10 +132,37 @@ public class ControllerArgumentBinder : IControllerActionArgumentBinder
...
+                controllerContext,
+                controllerContext.ValueProviderFactories);
+
+            return await BindModelAsync(parameter, controllerContext, valueProvider);
+        }
+
+        public async Task<ModelBindingResult?> BindModelAsync(
+            ParameterDescriptor parameter,
+            ControllerContext controllerContext,
+            IValueProvider valueProvider)
```
rynowak: This method needs to take the value provider as a parameter to avoid creating it over and over

src/Microsoft.AspNetCore.Mvc.Core/Internal/FilterActionInvoker.cs
```diff
@@ -80,14 +80,17 @@ public abstract class FilterActionInvoker : IActionInvoker
...
             _controllerActionInvokerCache = controllerActionInvokerCache;
             _inputFormatters = inputFormatters;
             _valueProviderFactories = valueProviderFactories;
             Logger = logger;
             _diagnosticSource = diagnosticSource;
             _maxModelValidationErrors = maxModelValidationErrors;
+
+            Context = new ControllerContext(actionContext);
+            Context.ModelState.MaxAllowedErrors = _maxModelValidationErrors;
+            Context.ValueProviderFactories = new List<IValueProviderFactory>(_valueProviderFactories);
```
rynowak: I felt like this code made more sense in the constructor since it's just doing initialization of the context

test/Microsoft.AspNetCore.Mvc.TestCommon/SimpleValueProviderFactory.cs
```diff
@@ -0,0 +1,36 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc.Internal;
+
+namespace Microsoft.AspNetCore.Mvc.ModelBinding
+{
+    public class SimpleValueProviderFactory : IValueProviderFactory
```
rynowak: Put this and `SimpleValueProvider` in the test common project because they are used multiple places

src/Microsoft.AspNetCore.Mvc.Core/ControllerBase.cs
```diff
@@ -1079,7 +1079,10 @@ public virtual SignOutResult SignOut(AuthenticationProperties properties, params
                 throw new ArgumentNullException(nameof(prefix));
             }
 
-            return TryUpdateModelAsync(model, prefix, new CompositeValueProvider(ControllerContext.ValueProviders));
+            var valueProvider = await CompositeValueProvider.CreateAsync(
+                ControllerContext,
+                ControllerContext.ValueProviderFactories);
```
dougbu: Suggest reducing this to `await CompositeValueProvider.CreateAsync(ControllerContext)`. No need for the additional parameter at least here.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerArgumentBinder.cs
dougbu: Sure but why is the value provider created at all in the case where all parameters have `[FromBody]` or similar? I was expecting a lazy-init property on the `ModelBindingContext` that's backed in the `ControllerContext`.

src/Microsoft.AspNetCore.Mvc.Core/ControllerBase.cs
rynowak: We would need both I think, but I'm not opposed to adding both.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerArgumentBinder.cs
rynowak: The argument binder will look for the prefix no matter what. We could try to change that, but that's another change.

Lazy property doesn't work because creating VPs is async.

dougbu: > that's another change.

Ah well :ok:

> Lazy property doesn't work 

Yeah, I meant a memoized `async` method.

Also didn't mean "expecting" since I knew the general direction of this PR :no_bell:

src/Microsoft.AspNetCore.Mvc.Core/ControllerBase.cs
dougbu: Where do we need to pass different factories?

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerArgumentBinder.cs
```diff
@@ -241,7 +289,8 @@ private static void CallPropertyAddRange<TElement>(object target, object source)
 
         private async Task<IDictionary<string, object>> PopulateArgumentsAsync(
             ControllerContext controllerContext,
-            IList<ParameterDescriptor> parameters)
+            IList<ParameterDescriptor> parameters,
+            IValueProvider valueProvider)
```
dougbu: Missing something here: The new parameter isn't used yet and this method still calls the older `BindModelAsync()` overload.

test/Microsoft.AspNetCore.Mvc.IntegrationTests/TryUpdateModelIntegrationTest.cs
```diff
@@ -1085,19 +1085,23 @@ System.Collections.IEnumerator System.Collections.IEnumerable.GetEnumerator()
             }
         }
 
-        private Task<bool> TryUpdateModel(
+        private async Task<bool> TryUpdateModel(
```
dougbu: `...Async()` :trollface:

src/Microsoft.AspNetCore.Mvc.Core/Internal/FilterActionInvoker.cs
```diff
@@ -336,18 +337,6 @@ private Task InvokeAllResourceFiltersAsync()
...
-                    // binding.
-                    var valueProviders = new List<IValueProvider>();
-                    var factoryContext = new ValueProviderFactoryContext(Context);
-
-                    for (var i = 0; i < _valueProviderFactories.Count; i++)
-                    {
-                        var factory = _valueProviderFactories[i];
-                        await factory.CreateValueProviderAsync(factoryContext);
-                    }
-                    Context.ValueProviders = factoryContext.ValueProviders;
```
dougbu: Previously the value providers were not created until this point, immediately before model binding. In the normal (non-`TryUpdateModelAsync()`) case, what is better with this delayed until w/in model binding?

dougbu: Answering my own question: This makes a difference when no controller properties or action parameters need to be bound. Also opens the door on avoiding the factories in additional cases e.g. the all-singing all-dancing all-`[FromBody]` case.

src/Microsoft.AspNetCore.Mvc.Core/Internal/FilterActionInvoker.cs
```diff
@@ -80,14 +80,17 @@ public abstract class FilterActionInvoker : IActionInvoker
                 throw new ArgumentNullException(nameof(diagnosticSource));
             }
 
-            Context = new ControllerContext(actionContext);
 
             _controllerActionInvokerCache = controllerActionInvokerCache;
             _inputFormatters = inputFormatters;
             _valueProviderFactories = valueProviderFactories;
```
dougbu: No longer need the `_valueProviderFactories` field.

src/Microsoft.AspNetCore.Mvc.Core/Internal/FilterActionInvoker.cs
dougbu: If `ValueProviderFactories` were also a `IReadOnlyList<IValueProviderFactory>`, would this defensive copy be necessary for any scenario?

src/Microsoft.AspNetCore.Mvc.Core/ControllerContext.cs
```diff
@@ -52,18 +52,18 @@ public new ControllerActionDescriptor ActionDescriptor
         }
 
         /// <summary>
-        /// Gets or sets the list of <see cref="IValueProvider"/> instances for the current request.
+        /// Gets or sets the list of <see cref="IValueProviderFactory"/> instances for the current request.
         /// </summary>
-        public virtual IList<IValueProvider> ValueProviders
+        public virtual IList<IValueProviderFactory> ValueProviderFactories
```
dougbu: In what scenarios could this list be changed? That is, could the property instead be an `IReadOnlyList<IValueProviderFactory>`?

dougbu:
:watch:

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerArgumentBinder.cs
dougbu: As discussed offline, let's create additional "no don't run the value provider factories" cases in follow-up PRs.

src/Microsoft.AspNetCore.Mvc.Core/Internal/FilterActionInvoker.cs
rynowak: No, the point of `Context.ValueProviderFactories` is that it is mutable.

src/Microsoft.AspNetCore.Mvc.Core/ControllerContext.cs
rynowak: We intend for this to be changed in scenarios where you want to prevent MVC model binding from reading the form data (large file uploads). We had a meeting about this and everything.

src/Microsoft.AspNetCore.Mvc.Core/ControllerBase.cs
rynowak: I'm going to just remove the action context overload, since the whole point of this method is to reduce boilerplate. We can add it back if anyone has an important need.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerArgumentBinder.cs
rynowak: Fixed, this was just a mistake.

rynowak:
@dougbu updated

dougbu:
:shipit:

PR #4536, opened 2016-04-26
Refactor Argument Binder

rynowak:
This change just rearranges some code in the argument binder with a mind
towards performance and clarity. We're removing a few Task<T>'s here as
well in certain cases, but not yet all of them. We additionally save a
dictionary in the case where you have bound properties.

Hopefully these changes break the code into more discrete and sensible
units without multiple levels of indirection without abstraction.
- Main 'driver' code
- BindModel
- ActivateProperty


src/Microsoft.AspNetCore.Mvc.Core/Controllers/IControllerArgumentBinder.cs
```diff
@@ -9,15 +9,20 @@ namespace Microsoft.AspNetCore.Mvc.Controllers
     /// <summary>
     /// Provides a dictionary of action arguments.
     /// </summary>
-    public interface IControllerActionArgumentBinder
+    public interface IControllerArgumentBinder
```
rynowak: Interface doesn't match file, this was supposed to have changed a while back

src/Microsoft.AspNetCore.Mvc.Core/Controllers/IControllerArgumentBinder.cs
```diff
@@ -9,15 +9,20 @@ namespace Microsoft.AspNetCore.Mvc.Controllers
...
         /// which can be used to invoke the action. Also binds properties explicitly marked properties on the 
         /// <paramref name="controller"/>.
         /// </summary>
-        /// <param name="context">The <see cref="ControllerContext"/> associated with the current action.</param>
+        /// <param name="controllerContext">The <see cref="ControllerContext"/> associated with the current action.</param>
         /// <param name="controller">The controller object which contains the action.</param>
-        Task<IDictionary<string, object>> BindActionArgumentsAsync(ControllerContext context, object controller);
+        /// <param name="arguments">The arguments dictionary.</param>
+        /// <returns>A <see cref="Task"/> which, when completed signals the completion of argument binding.</returns>
+        Task BindArgumentsAsync(
```
rynowak: shorter name, increased happiness

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerArgumentBinder.cs
```diff
@@ -65,57 +66,56 @@ public class ControllerArgumentBinder : IControllerActionArgumentBinder
...
-            var valueProvider = await CompositeValueProvider.CreateAsync(
-                controllerContext,
-                controllerContext.ValueProviderFactories);
-
-            var actionArguments = await PopulateArgumentsAsync(
-                controllerContext,
-                actionDescriptor.Parameters,
-                valueProvider);
-            return actionArguments;
+            return BindArgumentsCoreAsync(controllerContext, controller, arguments);
```
rynowak: Just one main 'driver' now instead of two overlapping code paths

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerArgumentBinder.cs
```diff
@@ -205,107 +205,79 @@ public class ControllerArgumentBinder : IControllerActionArgumentBinder
...
-                object value;
-                if (propertyHelper == null || !properties.TryGetValue(property.Name, out value))
+            if (propertyHelper.Property.CanWrite && propertyHelper.Property.SetMethod?.IsPublic == true)
+            {
+                // Handle settable property. Do not set the property to null if the type is a non-nullable type.
+                if (value != null || metadata.IsReferenceOrNullableType)
                 {
-                    continue;
+                    propertyHelper.SetValue(controller, value);
                 }
```
rynowak: These cases rolled into the calling code. Note we don't need to lookup the value in a dictionary anymore, we just bind it right away

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerArgumentBinder.cs
```diff
@@ -205,107 +205,79 @@ public class ControllerArgumentBinder : IControllerActionArgumentBinder
...
-                        propertyHelper = propertyHelpers[j];
-                        break;
-                    }
-                }
+            var propertyType = propertyHelper.Property.PropertyType;
+            var metadata = _modelMetadataProvider.GetMetadataForType(propertyType);
 
-                object value;
-                if (propertyHelper == null || !properties.TryGetValue(property.Name, out value))
+            if (propertyHelper.Property.CanWrite && propertyHelper.Property.SetMethod?.IsPublic == true)
```
rynowak: None of the code below here changed, it's showing up in the diff because it's not directly inside a loop anymore

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ControllerActionInvokerTest.cs
```diff
@@ -2037,10 +2037,13 @@ public void CreateActionResult_ReturnsObjectContentResult(Type type, object inpu
                 .Setup(fp => fp.OnProvidersExecuted(It.IsAny<FilterProviderContext>()))
                 .Verifiable();
 
-            var actionArgumentsBinder = new Mock<IControllerActionArgumentBinder>();
-            actionArgumentsBinder.Setup(
-                    b => b.BindActionArgumentsAsync(It.IsAny<ControllerContext>(), It.IsAny<object>()))
-                .Returns(Task.FromResult<IDictionary<string, object>>(new Dictionary<string, object>()));
+            var argumentBinder = new Mock<IControllerArgumentBinder>();
```
rynowak: Test changes are just reactions to API change

src/Microsoft.AspNetCore.Mvc.Core/Controllers/IControllerArgumentBinder.cs
dougbu: But now it's a `public` API change. Might be better to fix the filename instead. (Admittedly this isn't a prominent namespace nor an in-your-face method.)

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerArgumentBinder.cs
dougbu: Adding `?w=0` to the URL makes this pretty easy on the eyes :smiley_cat:

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerArgumentBinder.cs
dougbu: Sorry, what do you mean? Main changes to this method are removal of outer loop and `continue` -> `return`.

src/Microsoft.AspNetCore.Mvc.Core/Controllers/IControllerArgumentBinder.cs
dougbu: See previous comment about changing public API.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerArgumentBinder.cs
dougbu: A lot of this could happen w/o changing any other files. Seems like there's three overlapping things in this PR:

1. Remove duplication and strangeness inside `ControllerArgumentBinder`. All `private`, all good.
2. Avoiding a `Task<T>` return value in `IControllerActionArgumentBinder`. Probably enough goodness there to take the `public` API change hit.
3. Changing `IControllerActionArgumentBinder` name to match filename and name of this default implementation. I'm less sure about this part's worth.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerArgumentBinder.cs
```diff
@@ -205,107 +205,79 @@ public class ControllerArgumentBinder : IControllerActionArgumentBinder
             return modelBindingResult;
         }
 
-        // Called via reflection.
-        private static void CallPropertyAddRange<TElement>(object target, object source)
```
dougbu: Moving this method just makes the PR harder to review. I _think_ the method didn't change as you moved it. Please confirm.

dougbu:
:watch:

src/Microsoft.AspNetCore.Mvc.Core/Controllers/IControllerArgumentBinder.cs
rynowak: meh. This isn't something I've heard of anyone replacing

rynowak: Discussed with @Eilon and we're still making changes in this area until we're confident that we have it right.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerArgumentBinder.cs
rynowak: The mean that the special cases that used to be handled by `continue` are now part of the calling code and not part of this method.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerArgumentBinder.cs
rynowak: didn't change it.

rynowak:
@dougbu rebased and updated

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerArgumentBinder.cs
dougbu: Get the comments elsewhere about the `public` API changes. All fine on the 3 overlapping things in this PR.

dougbu:
:shipit: all :question:s answered.

Issue #4498, opened 2016-04-18
Consider unifying the list of assemblies included in the razor compilation with the of assembly parts of the application.


javiercn:
Right now application part manager is used for controller discovery, view component discovery and tag helper discovery (when doing AddTagHelpersAsServices()). However the assemblies included in the razor compilation are hard to configure. https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngineOptions.cs#L20

We could use application parts in order to define the list of assemblies to include as part of the compilation, this makes it easier to configure the list of assemblies that are part of the compilation (by just adding them to the list of application parts and having a feature that discovers them) and provides a single source of truth for configuring what assemblies are part of the application.

Eilon:
Sounds quite intriguing. @javiercn can you chat with @rynowak @NTaylorMullen @pranavkm briefly to think this through?

javiercn:
Discussed this with @pranavkm and seems like a good match. @rynowak suggested a quick design meeting for this.

Eilon:
Cool.

javiercn:
We had a quick discussion about this offline. We'll introduce a new application feature to represent compilation assets and populate it at the same time the application manager is created.

pranavkm:
I had a chat with @davidfowl offline and his suggestion was to not create a separate list for this - just reuse the assembly part we already have. https://github.com/aspnet/Mvc/commit/b3618f4b1c5e01b681f8dacac8a8f6d47553626e was a design based on that change.

DmitrySikorsky:
This would be really great feature!

PR #4540, opened 2016-04-27
Avoid holding on to DependencyContext instance

### LINKED ISSUE #4527: Avoid holding on to DependencyContext instances
They are pretty memory heavy and we should avoid keeping them around. Based on https://github.com/aspnet/Mvc/pull/4522#discussion_r60835905

pranavkm:
Fixes #4527

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultRoslynCompilationService.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultRoslynCompilationService.cs
@@ -29,12 +29,12 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Internal
     /// </summary>
     public class DefaultRoslynCompilationService : ICompilationService
     {
+        private readonly IHostingEnvironment _hostingEnvironment;
         private readonly IFileProvider _fileProvider;
         private readonly Action<RoslynCompilationContext> _compilationCallback;
         private readonly CSharpParseOptions _parseOptions;
         private readonly CSharpCompilationOptions _compilationOptions;
         private readonly ILogger _logger;
-        private readonly DependencyContext _dependencyContext;
         private object _applicationReferencesLock = new object();
         private bool _applicationReferencesInitialized;
         private List<MetadataReference> _applicationReferences;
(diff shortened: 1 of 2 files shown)


pranavkm:
These were the objects and their allocations in the gen2 heap prior to this change.
image

pranavkm:
cc @rynowak \ @Eilon

javiercn:
Could we use application parts for this instead to calculate the metadata references? That way we solve three problems. We don't hold on to the dependency context, we configure the list of assemblies in our application at a single point and we make easier to add or remove assemblies to the compilation rather than using the callback in the options.

pranavkm:
> Could we use application parts for this instead to calculate the metadata references?

What does that look like?

src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultRoslynCompilationService.cs
```diff
@@ -165,6 +151,22 @@ public CompilationResult Compile(RelativeFileInfo fileInfo, string compilationCo
...
         }
 
+        /// <summary>
+        /// Gets the <see cref="DependencyContext"/>.
+        /// </summary>
+        /// <param name="hostingEnvironment">The <see cref="IHostingEnvironment"/>.</param>
+        /// <returns>The <see cref="DependencyContext"/>.</returns>
+        protected virtual DependencyContext GetDependencyContext(IHostingEnvironment hostingEnvironment)
+        {
+            if (hostingEnvironment.ApplicationName != null)
```
davidfowl: Hmm, when is this ever null?

src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultRoslynCompilationService.cs
```diff
@@ -29,12 +29,12 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Internal
     /// </summary>
     public class DefaultRoslynCompilationService : ICompilationService
     {
+        private readonly IHostingEnvironment _hostingEnvironment;
         private readonly IFileProvider _fileProvider;
         private readonly Action<RoslynCompilationContext> _compilationCallback;
         private readonly CSharpParseOptions _parseOptions;
         private readonly CSharpCompilationOptions _compilationOptions;
         private readonly ILogger _logger;
-        private readonly DependencyContext _dependencyContext;
```
rynowak: Will we now load the DC for **every** individual compilation? Or are the metadata references cached indefinitely?

src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultRoslynCompilationService.cs
pranavkm: It's sometimes null in unit test scenarios. Although I could remove this check now that we only create it when we compile and not as part of the ctor.

src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultRoslynCompilationService.cs
pranavkm: Nope, we cache the result of it (`_applicationReferences`) the first time we need to compile.

rynowak: Ok, this just seems like a good change all around then

test/Microsoft.AspNetCore.Mvc.Razor.Test/Internal/DefaultRoslynCompilationServiceTest.cs
```diff
@@ -422,5 +413,33 @@ private DependencyContext GetDependencyContext()
...
+                      GetAccessor(viewEngineOptions),
+                      fileProviderAccessor,
+                      NullLoggerFactory.Instance)
+            {
+                _dependencyContext = dependencyContext;
+            }
+
+            private static IOptions<RazorViewEngineOptions> GetAccessor(RazorViewEngineOptions options)
+            {
+                var optionsAccessor = new Mock<IOptions<RazorViewEngineOptions>>();
```
rynowak: You can just use `TestOptionsManager<>` here

rynowak:
:shipit:

davidfowl:
:shipit: 

@javiercn, @pranavkm and @rynowak , we should discuss adding compilation things to application parts, or using it as the seed of assemblies to find references from.

rynowak:
We already talked about that today :laughing:

davidfowl:
Perfect, I'd like to know what the plan is 😊

pranavkm:
@davidfowl - https://github.com/aspnet/Mvc/issues/4498 tracks it. We'll introduce a new application feature to represent compilation assets and populate it at the same time the application manager is created.

PR #4545, opened 2016-04-27
[Perf] Reduce SelectListItem and other allocations when generating HTML for select lists

### LINKED ISSUE #3953: `SelectListItem` manipulation is quite expensive

mnltejaswini:
Fixes #3953 

The changes include
1. Cleaning up Linq and Linq to SQL like  usage in MultiSelectList.cs
2. Caching the List<SelectListItem> in MultiSelectList that will be used when getting the Enumerator
3. Removing the Groupby operation in DefaultHtmlGenerator.GenerateGroupsAndOptions
4. Remove the defensive deep copies of SelectListItem done in DefaultHtmlGenerator.GenerateSelect

Scenario: https://github.com/aspnet/Performance/tree/dev/testapp/BigViews 

Before Optimization
------------------------
image

After Optimization
-----------------------
image

Gain: ~30% reduction in total allocated bytes and total allocated objects.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Rendering/MultiSelectList.cs
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Rendering/MultiSelectList.cs
@@ -18,6 +18,7 @@ namespace Microsoft.AspNetCore.Mvc.Rendering
     public class MultiSelectList : IEnumerable<SelectListItem>
     {
         private IList<SelectListGroup> _groups;
+        private IList<SelectListItem> _selectListItems;
 
         public MultiSelectList(IEnumerable items)
             : this(items, selectedValues: null)
@@ -111,10 +112,15 @@ namespace Microsoft.AspNetCore.Mvc.Rendering
 
         public virtual IEnumerator<SelectListItem> GetEnumerator()
         {
-            return GetListItems().GetEnumerator();
+            if (_selectListItems == null)
+            {
+                _selectListItems = GetListItems();
+            }
+
+            return _selectListItems.GetEnumerator();
         }
 
-        internal IList<SelectListItem> GetListItems()
+        private IList<SelectListItem> GetListItems()
         {
             return (!string.IsNullOrEmpty(DataValueField)) ?
                 GetListItemsWithValueField() :
@@ -126,20 +132,29 @@ namespace Microsoft.AspNetCore.Mvc.Rendering
             var selectedValues = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
             if (SelectedValues != null)
             {
-                selectedValues.UnionWith(from object value in SelectedValues
-                                         select Convert.ToString(value, CultureInfo.CurrentCulture));
+                foreach (var value in SelectedValues)
+                {
+                    var stringValue = Convert.ToString(value, CultureInfo.CurrentCulture);
+                    selectedValues.Add(stringValue);
+                }
             }
 
-            var listItems = from object item in Items
-                            let value = Eval(item, DataValueField)
-                            select new SelectListItem
-                            {
-                                Group = GetGroup(item),
-                                Value = value,
-                                Text = Eval(item, DataTextField),
-                                Selected = selectedValues.Contains(value)
-                            };
-            return listItems.ToList();
+            var listItems = new List<SelectListItem>();
+            foreach (var item in Items)
+            {
+                var value = Eval(item, DataValueField);
+                var newListItem = new SelectListItem
+                {
+                    Group = GetGroup(item),
+                    Value = value,
+                    Text = Eval(item, DataTextField),
+                    Selected = selectedValues.Contains(value),
+                };
+
+                listItems.Add(newListItem);
+            }
+
+            return listItems;
         }
 
         private IList<SelectListItem> GetListItemsWithoutValueField()
@@ -150,14 +165,20 @@ namespace Microsoft.AspNetCore.Mvc.Rendering
                 selectedValues.UnionWith(SelectedValues.Cast<object>());
             }
 
-            var listItems = from object item in Items
-                            select new SelectListItem
-                            {
-                                Group = GetGroup(item),
-                                Text = Eval(item, DataTextField),
-                                Selected = selectedValues.Contains(item)
-                            };
-            return listItems.ToList();
+            var listItems = new List<SelectListItem>();
+            foreach (var item in Items)
+            {
+                var newListItem = new SelectListItem
+                {
+                    Group = GetGroup(item),
+                    Text = Eval(item, DataTextField),
+                    Selected = selectedValues.Contains(item),
+                };
+
+                listItems.Add(newListItem);
+            }
+
+            return listItems;
         }
 
         private static string Eval(object container, string expression)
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
```diff
@@ -1433,34 +1434,13 @@ private static string GetInputTypeString(InputType inputType)
...
-                newSelectList.Add(copy);
-            }
-
-            return newSelectList;
+            return GenerateGroupsAndOptions(optionLabel: optionLabel, selectList: selectList, currentValues: null);
         }
 
-        /// <inheritdoc />
-        public IHtmlContent GenerateGroupsAndOptions(string optionLabel, IEnumerable<SelectListItem> selectList)
+        private IHtmlContent GenerateGroupsAndOptions(string optionLabel, IEnumerable<SelectListItem> selectList, ICollection<string> currentValues)
```
dougbu: - Wrap this line w/ each parameter on its own line.
- `currentValues` should be either `HashSet<string>` or `ISet<string>` to avoid Linq use.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
```diff
@@ -1472,53 +1452,76 @@ public IHtmlContent GenerateGroupsAndOptions(string optionLabel, IEnumerable<Sel
                     Text = optionLabel,
                     Value = string.Empty,
                     Selected = false,
-                }));
+                }, currentValues: null));
```
dougbu: Rewrap to place the `GenerateOption` arguments on their own lines:
``` c#
listItemBuilder.AppendLine(GenerateOption(
    new SelectListItem
    {
        Text = optionLabel,
        Value = string.Empty,
        Selected = false,
    },
    currentValues: null));
```

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
```diff
@@ -1472,53 +1452,76 @@ public IHtmlContent GenerateGroupsAndOptions(string optionLabel, IEnumerable<Sel
...
+                        optionGenerated[i] = true;
                     }
                 }
             }
 
             return listItemBuilder;
         }
 
-        private IHtmlContent GenerateOption(SelectListItem item)
+        private IHtmlContent GenerateOption(SelectListItem item, ICollection<string> currentValues)
```
dougbu: `currentValues` should be either `HashSet<string>` or `ISet<string>` to avoid Linq in this method.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
```diff
@@ -1472,53 +1452,76 @@ public IHtmlContent GenerateGroupsAndOptions(string optionLabel, IEnumerable<Sel
...
-                else
-                {
-                    foreach (var item in group)
+                        for (var j = i; j < itemsList.Count; j++)
+                        {
+                            var groupItem = itemsList[j];
+
+                            if (!optionGenerated[j] &&
+                                groupItem.Group != null &&
+                                groupItem.Group.GetHashCode() == item.Group.GetHashCode())
```
dougbu: Use `object.ReferenceEquals(optGroup, groupItem.Group)` instead of `groupItem.Group != null && groupItem.Group.GetHashCode() == item.Group.GetHashCode()`. For one thing, you've got `optGroup` and don't need to reference `item.Group` again. For another, the previous use of `GetHashCode()` was a hack to maintain placement of items not in groups; no need for that now. See also the comments at the top of [`SelectListGroup`](https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Rendering/SelectListGroup.cs).

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
dougbu: One way to check: Remove the `using System.Linq;` statement from this file. The only compilation errors should relate to `ToArray()` and (perhaps) `ToList()`.

dougbu:
Why did the CI builds fail?

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Rendering/MultiSelectList.cs
```diff
@@ -111,7 +112,12 @@ public MultiSelectList(IEnumerable items, string dataValueField, string dataText
...
-            return GetListItems().GetEnumerator();
+            if (_selectListItems == null)
+            {
+                _selectListItems = GetListItems();
+            }
+
+            return _selectListItems.GetEnumerator();
         }
 
         internal IList<SelectListItem> GetListItems()
```
dougbu: This may have been `internal` for testing at some point. But nothing references it today. Since it's now less performant than using the enumerator directly, make it `private`.

dougbu:
:watch:

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
mnltejaswini: Contains is a method that ICollection provides.

mnltejaswini:
On the day I created the PR, the dev branch was having two failures in the functional tests which are what shown by the CI.

mnltejaswini:
:up::date:

src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs
dougbu: > Contains is a method that ICollection provides.

You're right. Never mind.

dougbu:
:shipit:

PR #4547, opened 2016-04-27
React to aspnet/Razor#738

ajaybhargavb:
React to aspnet/Razor#738

@NTaylorMullen FYI

### DIFF
--- a/test/Microsoft.AspNetCore.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.OrderUsingHtmlHelpers.Encoded.html
+++ b/test/Microsoft.AspNetCore.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.OrderUsingHtmlHelpers.Encoded.html
@@ -1,11 +1,10 @@
-﻿<html>
+<html>
 <head>
     <meta charset="utf-8" />
     <title></title>
 </head>
 <body>
-<form action="HtmlEncode[[/UrlEncode[[HtmlGeneration_Order]]/UrlEncode[[Submit]]]]" method="HtmlEncode[[post]]">
-        <div>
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNetCore.Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.OrderUsingHtmlHelpers.Encoded.html
```diff
@@ -1,11 +1,10 @@
-﻿<html>
+<html>
 <head>
     <meta charset="utf-8" />
     <title></title>
 </head>
 <body>
-<form action="HtmlEncode[[/UrlEncode[[HtmlGeneration_Order]]/UrlEncode[[Submit]]]]" method="HtmlEncode[[post]]">
-        <div>
+<form action="HtmlEncode[[/UrlEncode[[HtmlGeneration_Order]]/UrlEncode[[Submit]]]]" method="HtmlEncode[[post]]">        <div>
```
dougbu: This looks like a bug in the Razor change. There's no C# between the `<form>` and `<div>` elements and there _is_ whitespace. Is Razor now getting confused about transitions into and out of tag helpers?

Source is
``` html
    <form asp-controller="HtmlGeneration_Order" asp-action="Submit" asp-antiforgery=" true">
        <div>
```

ajaybhargavb: Actually, there is no tag helpers involved here. It is because of the `<text>` tag here.
```HTML
@* Use comments and <text/> elements to force whitespace in generated HTML. *@@using (
Html.BeginForm(actionName: "Submit", controllerName: "HtmlGeneration_Order"))
    {<text></text>
        <div>
```

PR #4548, opened 2016-04-28
How to make friends and influence people through the power of code duplication

rynowak:
This change rewrites the invoker state machine to avoid allocating
_anything_ unless necessary to call into a filter. We also avoid calling
into any `async` method unless necessary to avoid the state machine code
generated by the compiler.

This results in some code duplication, but the improvements are
significant and realized any time that you don't have a particular type of
filter.

===========

**The impact on techempower plaintext in our Smurflab :tm: 218krps -> 255krps**

Not including sampling data here because it's pretty hard to interpret the diff between the relatively deep callstacks in the old code and the flat ones in the new code. Now the user's action is about 25% of the total execution time, previously it was about 20%.

===========

Some allocation data

**Before**
image

image

**After**
image

image


src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
```diff
@@ -29,7 +28,6 @@ public class ControllerActionInvoker : FilterActionInvoker
             ControllerActionInvokerCache controllerActionInvokerCache,
             IControllerFactory controllerFactory,
             ControllerActionDescriptor descriptor,
-            IReadOnlyList<IInputFormatter> inputFormatters,
```
rynowak: Unused

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
```diff
@@ -82,34 +79,32 @@ protected override void ReleaseInstance(object instance)
         {
             _controllerFactory.ReleaseController(Context, instance);
         }
-
-        protected override async Task<IActionResult> InvokeActionAsync(ActionExecutingContext actionExecutingContext)
+        
+        protected override async Task<IActionResult> InvokeActionAsync(IDictionary<string, object> arguments)
```
rynowak: This is needed so we can call this method without allocating `ActionExecutingContext`

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvokerProvider.cs
```diff
@@ -37,7 +34,6 @@ public class ControllerActionInvokerProvider : IActionInvokerProvider
             _controllerFactory = controllerFactory;
             _controllerActionInvokerCache = controllerActionInvokerCache;
             _argumentBinder = argumentBinder;
-            _inputFormatters = optionsAccessor.Value.InputFormatters.ToArray();
```
rynowak: Unused

src/Microsoft.AspNetCore.Mvc.Core/Internal/MvcCoreDiagnosticSourceExtensions.cs
```diff
@@ -132,7 +132,6 @@ public static class MvcCoreDiagnosticSourceExtensions
 
         public static void AfterOnResourceExecution(
             this DiagnosticSource diagnosticSource,
-            ActionDescriptor actionDescriptor,
```
rynowak: These don't really need to be passed in here, it was just bloat

src/Microsoft.AspNetCore.Mvc.Core/Internal/MvcCoreLoggerExtensions.cs
```diff
@@ -269,7 +269,7 @@ public static void ActionMethodExecuting(this ILogger logger, ActionExecutingCon
             }
         }
 
-        public static void ActionMethodExecuted(this ILogger logger, ActionExecutingContext context, IActionResult result)
```
rynowak: Same as the binder - we might invoke the action without an `ActionExecutingContext`

src/Microsoft.AspNetCore.Mvc.Core/Internal/TaskCache.cs
```diff
@@ -8,22 +8,12 @@ namespace Microsoft.AspNetCore.Mvc.Internal
...
-        public static Task CompletedTask
-        {
-            get
-            {
-#if NET451
-                return _completedTask;
+        /// <summary>Gets a task that's already been completed successfully.</summary> 
+        public static readonly Task CompletedTask = Task.FromResult(0);
 #else
-                return Task.CompletedTask;
```
rynowak: This was actually showing up in the profile. It's not getting inlined, and the implementation of `Task.CompletedTask` does a volatile read.

src/Microsoft.AspNetCore.Mvc.Core/Internal/TaskCache.cs
```diff
@@ -8,22 +8,12 @@ namespace Microsoft.AspNetCore.Mvc.Internal
     public static class TaskCache
     {
 #if NET451
-        static readonly Task _completedTask = Task.FromResult(0);
-#endif
-
-        /// <summary>Gets a task that's already been completed successfully.</summary>
-        /// <remarks>May not always return the same instance.</remarks>
```
rynowak: Yeah. that's not really possible :laughing:

rynowak:
Hmm, github has given up on showing the diff of `FilterActionInvoker` - just as well, it's pretty much a rewrite.

DamianEdwards:
:100: :smiley: :+1: :speedboat:

src/Microsoft.AspNetCore.Mvc.Core/Internal/TaskCache.cs
davidfowl: We should move this helper to common. Kestrel has the same thing

davidfowl:
@rynowak I was going to ask where the changes are...

benaadams:
:rocket:

rynowak:
Closing this out in favor of https://github.com/aspnet/Mvc/pull/4687 which has a little better performance  and actually seems more maintainable once you get the gist of it.

PR #4546, opened 2016-04-27
[Perf] Replace Activator.CreateInstance with cached delegate in Model Binders

mnltejaswini:
Fixes Part1 of #4470
Refer to #4524  for before and after perf numbers

PR #4554, opened 2016-04-28
[Perf] Cache the default values for value type action arguments to avoid Activator.CreateInstance

### LINKED ISSUE #4470: Review use of Activator.CreateInstance

mnltejaswini:
Fixes #4470 
This will avoid Activator.CreateInstance per request when getting the default value for the value type parameter.
More details please refer to #4524

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionExecutor.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionExecutor.cs
@@ -16,7 +16,7 @@ namespace Microsoft.AspNetCore.Mvc.Internal
             object instance,
             IDictionary<string, object> actionArguments)
         {
-            var orderedArguments = PrepareArguments(actionArguments, actionMethodExecutor.MethodInfo.GetParameters());
+            var orderedArguments = PrepareArguments(actionArguments, actionMethodExecutor);
             return ExecuteAsync(actionMethodExecutor, instance, orderedArguments);
         }
 
@@ -30,8 +30,9 @@ namespace Microsoft.AspNetCore.Mvc.Internal
 
         public static object[] PrepareArguments(
             IDictionary<string, object> actionParameters,
-            ParameterInfo[] declaredParameterInfos)
+            ObjectMethodExecutor actionMethodExecutor)
         {
+            var declaredParameterInfos = actionMethodExecutor.ActionParameters;
             var count = declaredParameterInfos.Length;
             if (count == 0)
             {
@@ -46,26 +47,7 @@ namespace Microsoft.AspNetCore.Mvc.Internal
 
                 if (!actionParameters.TryGetValue(parameterInfo.Name, out value))
                 {
-                    if (parameterInfo.HasDefaultValue)
-                    {
-                        value = parameterInfo.DefaultValue;
-                    }
-                    else
-                    {
-                        var defaultValueAttribute = 
-                            parameterInfo.GetCustomAttribute<DefaultValueAttribute>(inherit: false);
-
-                        if (defaultValueAttribute?.Value == null)
-                        {
-                            value = parameterInfo.ParameterType.GetTypeInfo().IsValueType
-                                ? Activator.CreateInstance(parameterInfo.ParameterType)
-                                : null;
-                        }
-                        else
-                        {
-                            value = defaultValueAttribute.Value;
-                        }
-                    }
+                    value = actionMethodExecutor.GetDefaultValueForParameter(index);
                 }
 
                 arguments[index] = value;
--- a/src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
@@ -14,6 +15,7 @@ namespace Microsoft.AspNetCore.Mvc.Internal
 {
     public class ObjectMethodExecutor
     {
+        private object[] _parameterDefaultValues;
         private ActionExecutorAsync _executorAsync;
         private ActionExecutor _executor;
 
@@ -31,7 +33,6 @@ namespace Microsoft.AspNetCore.Mvc.Internal
 
         private static readonly MethodInfo _coerceMethod = ((MethodCallExpression)_coerceTaskExpression.Body).Method;
 
-
         private ObjectMethodExecutor(MethodInfo methodInfo)
         {            
             if (methodInfo == null)
@@ -39,6 +40,7 @@ namespace Microsoft.AspNetCore.Mvc.Internal
                 throw new ArgumentNullException(nameof(methodInfo));
             }
             MethodInfo = methodInfo;
+            ActionParameters = methodInfo.GetParameters();
         }
 
         private delegate Task<object> ActionExecutorAsync(object target, object[] parameters);
@@ -48,6 +50,8 @@ namespace Microsoft.AspNetCore.Mvc.Internal
         private delegate void VoidActionExecutor(object target, object[] parameters);
 
         public MethodInfo MethodInfo { get; }
+
+        public ParameterInfo[] ActionParameters { get; }
 
         public static ObjectMethodExecutor Create(MethodInfo methodInfo, TypeInfo targetTypeInfo)
         {
@@ -65,6 +69,18 @@ namespace Microsoft.AspNetCore.Mvc.Internal
         public object Execute(object target, object[] parameters)
         {
             return _executor(target, parameters);
+        }
+
+        public object GetDefaultValueForParameter(int index)
+        {
+            if (index < 0 || index > ActionParameters.Length - 1)
+            {
+                throw new ArgumentOutOfRangeException(nameof(index));
+            }
+
+            EnsureParameterDefaultValues();
+
+            return _parameterDefaultValues[index];
         }
 
         private static ActionExecutor GetExecutor(MethodInfo methodInfo, TypeInfo targetTypeInfo)
@@ -237,6 +253,44 @@ namespace Microsoft.AspNetCore.Mvc.Internal
         {
             var task = (Task<T>)taskAsObject;
             return CastToObject<T>(task);
+        }
+
+        private void EnsureParameterDefaultValues()
+        {
+            if (_parameterDefaultValues == null)
+            {
+                var count = ActionParameters.Length;
+                _parameterDefaultValues = new object[count];
+
+                for (var i = 0; i < count; i++)
+                {
+                    var parameterInfo = ActionParameters[i];
+                    object defaultValue;
+
+                    if (parameterInfo.HasDefaultValue)
+                    {
+                        defaultValue = parameterInfo.DefaultValue;
+                    }
+                    else
+                    {
+                        var defaultValueAttribute = parameterInfo
+                            .GetCustomAttribute<DefaultValueAttribute>(inherit: false);
+
+                        if (defaultValueAttribute?.Value == null)
+                        {
+                            defaultValue = parameterInfo.ParameterType.GetTypeInfo().IsValueType
+                                ? Activator.CreateInstance(parameterInfo.ParameterType)
+                                : null;
+                        }
+                        else
+                        {
+                            defaultValue = defaultValueAttribute.Value;
+                        }
+                    }
+
+                    _parameterDefaultValues[i] = defaultValue;
+                }
+            }
         }
     }
 }
--- a/test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ObjectMethodExecutorTest.cs
+++ b/test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ObjectMethodExecutorTest.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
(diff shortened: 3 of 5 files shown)


src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
```diff
@@ -67,6 +73,39 @@ public object Execute(object target, object[] parameters)
             return _executor(target, parameters);
         }
 
+        public object GetDefaultValueForParameter(int index)
+        {
+            Debug.Assert(ActionParameters.Length > 0);
```
dougbu: This is a `public` method and should not contain debug assertions related to properties initialized in the constructor. Instead `throw` an `ArgumentOutOfRangeException` if `index` is less than `0` or greater than `ActionParameters.Length - 1`.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
```diff
@@ -67,6 +73,39 @@ public object Execute(object target, object[] parameters)
...
+            if (_parameterDefaultValues == null)
+            {
+                _parameterDefaultValues = new object[ActionParameters.Length];
+            }
+
+            var defaultValue = _parameterDefaultValues[index];
+            if (defaultValue == null)
+            {
+                var defaultValueAttribute =
+                            parameterInfo.GetCustomAttribute<DefaultValueAttribute>(inherit: false);
```
dougbu: Use default indentation, nothing extra here.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
```diff
@@ -67,6 +73,39 @@ public object Execute(object target, object[] parameters)
...
+            Debug.Assert(ActionParameters.Length > 0);
+
+            var parameterInfo = ActionParameters[index];
+            if (_parameterDefaultValues == null)
+            {
+                _parameterDefaultValues = new object[ActionParameters.Length];
+            }
+
+            var defaultValue = _parameterDefaultValues[index];
+            if (defaultValue == null)
```
dougbu: Need a separate guard for this logic since repeating the same logic to come up w/ `null` isn't worthwhile.

dougbu: Actually, why not calculate all default values up front? `PrepareArguments` runs through all the parameters.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
```diff
@@ -49,6 +53,8 @@ private ObjectMethodExecutor(MethodInfo methodInfo)
 
         public MethodInfo MethodInfo { get; }
 
+        public ParameterInfo[] ActionParameters { get; }
```
dougbu: Why add this new `public` property? Looks like an implementation detail.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionExecutor.cs
```diff
@@ -52,19 +53,7 @@ public static class ControllerActionExecutor
...
-                        {
-                            value = parameterInfo.ParameterType.GetTypeInfo().IsValueType
-                                ? Activator.CreateInstance(parameterInfo.ParameterType)
-                                : null;
-                        }
-                        else
-                        {
-                            value = defaultValueAttribute.Value;
-                        }
+                        value = actionMethodExecutor.GetDefaultValueForParameter(index);
```
dougbu: Deal with `parameterInfo.DefaultValue` inside `GetDefaultValueForParameter()`. No need to expose `ActionParameters` or to split the default value logic into two places.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
mnltejaswini: MethodInfo.GetParameters() does allocation of ParameterInfo[] for every invocation. In order to avoid repeated allocations of the array for every request, this property is made available in ObjectMethodExecutor that can be used in PrepareArguments.

mnltejaswini:
:up:date:

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
```diff
@@ -31,6 +33,48 @@ public class ObjectMethodExecutor
 
         private static readonly MethodInfo _coerceMethod = ((MethodCallExpression)_coerceTaskExpression.Body).Method;
 
+        private object[] ParameterDefaultValues
```
dougbu: No need for the additional property. Instead create a `private void EnsureParameterDefaultValues()` method at the bottom of the file and call it before using `_parameterDefaultValues`.

Note that if the `private` property were necessary, it should be added just below the existing properties.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
```diff
@@ -31,6 +33,48 @@ public class ObjectMethodExecutor
...
+            {
+                if (_parameterDefaultValues == null)
+                {
+                    var count = ActionParameters.Length;
+                    _parameterDefaultValues = new object[count];
+
+                    for (var i = 0; i < count; i++)
+                    {
+                        var parameterInfo = ActionParameters[i];
+                        object defaultValue = null;
```
dougbu: Don't initialize this variable. That'll catch errors later.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
```diff
@@ -31,6 +33,48 @@ public class ObjectMethodExecutor
...
+
+                        if (parameterInfo.HasDefaultValue)
+                        {
+                            defaultValue = parameterInfo.DefaultValue;
+                        }
+                        else
+                        {
+                            var defaultValueAttribute = parameterInfo
+                            .GetCustomAttribute<DefaultValueAttribute>(inherit: false);
```
dougbu: Indent 4

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
dougbu: :ok:

dougbu:
What is the before / after perf change w/ this PR and whatever scenario it helps the most?

dougbu:
:watch:

mnltejaswini:
There is no specific scenario in Performance repo that can be used directly for measuring this PR benefit. I will create a scenario where action method has value types with no default values and collect a sampling profile. Similar to  #4557  and #4546  the effort here is to avoid the slower path of Activator.CreateInstance() that can happen when preparing the arguments of the controller actions

mnltejaswini:
Scenario : https://github.com/aspnet/Performance/tree/dev/testapp/BigModelBinding 

Changed the HomeController to have 4 value type parameters with no default values.

Before
--------
image

After
-----
image

mnltejaswini:
:up::date:

rynowak:
That's a really nice improvement in the profile

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
```diff
@@ -67,6 +71,18 @@ public object Execute(object target, object[] parameters)
...
         }
 
+        public object GetDefaultValueForParameter(int index)
+        {
+            if (index < 0 || index > ActionParameters.Length-1)
+            {
+                throw new ArgumentOutOfRangeException(nameof(index));
+            }
+
+            EnsureParameterDefaultValues();
```
rynowak: There's no benefit to making this lazy, this is already a cache

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
```diff
@@ -238,5 +254,43 @@ private static Task<object> Convert<T>(object taskAsObject)
             var task = (Task<T>)taskAsObject;
             return CastToObject<T>(task);
         }
+
+        private void EnsureParameterDefaultValues()
+        {
+            if (_parameterDefaultValues == null)
+            {
+                var count = ActionParameters.Length;
+                _parameterDefaultValues = new object[count];
```
rynowak: Just move all of this into the `Create` method

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
```diff
@@ -67,6 +71,18 @@ public object Execute(object target, object[] parameters)
             return _executor(target, parameters);
         }
 
+        public object GetDefaultValueForParameter(int index)
```
rynowak: We should probably just expose the array here as a property

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
```diff
@@ -67,6 +71,18 @@ public object Execute(object target, object[] parameters)
...
+        {
+            if (index < 0 || index > ActionParameters.Length-1)
```
pranavkm: Formatting. (space before and after `-`

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ObjectMethodExecutorTest.cs
```diff
@@ -239,6 +254,26 @@ public TaskOfTDerivedType<int> TaskActionWithCustomTaskOfTReturnType(int i, stri
...
+            {
+                return input;
+            }
+
+            public int EchoIntValueTypeWithNoDefaultValue(int input)
+            {
+                return input;
+            }
+
+            public bool EchoBoolValueTypeWithNoDefaultValue(bool input)
```
rynowak: Needs a test case for a default value (c#) like `Foo(bool value = true)`

rynowak:
:watch: getting closer here

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
```diff
@@ -238,5 +254,43 @@ private static Task<object> Convert<T>(object taskAsObject)
...
+                        defaultValue = parameterInfo.DefaultValue;
+                    }
+                    else
+                    {
+                        var defaultValueAttribute = parameterInfo
+                            .GetCustomAttribute<DefaultValueAttribute>(inherit: false);
+
+                        if (defaultValueAttribute?.Value == null)
+                        {
+                            defaultValue = parameterInfo.ParameterType.GetTypeInfo().IsValueType
```
pranavkm: @rynowak - `IsValueType` would be true for nullable types. Should we unwrap it and avoid the call to `Activator.CreateInstance` here if it's one?

For instance:
```
if (parameterInfo.ParameterType.GetTypeInfo().IsValueType && Nullable.GetUnderlyingType(parameterInfo.ParameterType) == null) 
{
   return Activator.CreateInstance(parameterInfo.ParameterType);
}

return null;
```

pranavkm:
LKGTM

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
rynowak: What's the concern? perf or correctness?

I'm not concerned about the perf of this, it's done once, that's the point of this change :laughing:

pranavkm: Just perf. The behavior of `Activator.CreateInstance` is correct for nullable types.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
mnltejaswini: I see the benefit of making it lazy because we don't compute default values always. We do it only if there aren't any action parameters for the corresponding method parameters which may not happen very often. We can save the object array allocation per controller action method and also querying the custom attributes  by making it lazy.

mnltejaswini:
What is LKGTM ?

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
mnltejaswini: As the default values are cached for the lifetime of the app, there should be much difference in the perf.

pranavkm:
Looks Good To Me spelt wrong

dougbu:
@pranavkm that should be LGTMSW

rynowak:
:watch: Please fix https://github.com/aspnet/Mvc/pull/4554/files#r62764757 and add a few more tests

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ObjectMethodExecutorTest.cs
rynowak: Would also like to see a test with multiple parameters

mnltejaswini:
:up::date:

rynowak:
:shipit: looking good!

PR #4556, opened 2016-04-28
[Fixes #4550] Add an Option to set the Razor view location formats

### LINKED ISSUE #4550: Add an Option to set the Razor view location formats
To customize the set of path formats that the Razor view engine will look at you need to customize the view engine itself. You need to either derive from RazorViewEngine, override the list of formats, and add that to the list of view engines, or you have to instantiate RazorViewEngine itself, set the list of formats, and add that to the list of view engines.

This is the property in question: https://github.com/aspnet/Mvc/blob/af58c2e6b619697a1a57736e9e996bb4d7e6e7ee/src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngine.cs#L76
(Plus the one for areas.)

We already have a Razor View Options class (https://github.com/aspnet/Mvc/blob/af58c2e6b619697a1a57736e9e996bb4d7e6e7ee/src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngineOptions.cs) so we should have the default list of view location formats there instead.

kichalla:
@Eilon @pranavkm

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngine.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngine.cs
@@ -8,7 +8,6 @@ using System.Globalization;
 using System.Linq;
 using System.Text.Encodings.Web;
 using Microsoft.AspNetCore.Mvc.Razor.Internal;
-using Microsoft.AspNetCore.Mvc.Routing;
 using Microsoft.AspNetCore.Mvc.ViewEngines;
 using Microsoft.Extensions.Caching.Memory;
 using Microsoft.Extensions.Logging;
@@ -21,12 +20,15 @@ namespace Microsoft.AspNetCore.Mvc.Razor
     /// Default implementation of <see cref="IRazorViewEngine"/>.
     /// </summary>
     /// <remarks>
-    /// For <c>ViewResults</c> returned from controllers, views should be located in <see cref="ViewLocationFormats"/>
-    /// by default. For the controllers in an area, views should exist in <see cref="AreaViewLocationFormats"/>.
+    /// For <c>ViewResults</c> returned from controllers, views should be located in
+    /// <see cref="RazorViewEngineOptions.ViewLocationFormats"/>
+    /// by default. For the controllers in an area, views should exist in
+    /// <see cref="RazorViewEngineOptions.AreaViewLocationFormats"/>.
     /// </remarks>
     public class RazorViewEngine : IRazorViewEngine
     {
-        private const string ViewExtension = ".cshtml";
+        public static readonly string ViewExtension = ".cshtml";
+
         private const string ControllerKey = "controller";
         private const string AreaKey = "area";
         private static readonly ViewLocationCacheItem[] EmptyViewStartLocationCacheItems =
@@ -34,10 +36,10 @@ namespace Microsoft.AspNetCore.Mvc.Razor
         private static readonly TimeSpan _cacheExpirationDuration = TimeSpan.FromMinutes(20);
 
         private readonly IRazorPageFactoryProvider _pageFactory;
-        private readonly IList<IViewLocationExpander> _viewLocationExpanders;
         private readonly IRazorPageActivator _pageActivator;
         private readonly HtmlEncoder _htmlEncoder;
         private readonly ILogger _logger;
+        private readonly RazorViewEngineOptions _options;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="RazorViewEngine" />.
@@ -49,9 +51,24 @@ namespace Microsoft.AspNetCore.Mvc.Razor
             IOptions<RazorViewEngineOptions> optionsAccessor,
             ILoggerFactory loggerFactory)
         {
+            _options = optionsAccessor.Value;
+
+            if (_options.ViewLocationFormats.Count == 0)
+            {
+                throw new ArgumentException(
+                    Resources.FormatViewLocationFormatsIsRequired(nameof(RazorViewEngineOptions.ViewLocationFormats)),
+                    nameof(optionsAccessor));
+            }
+
+            if (_options.AreaViewLocationFormats.Count == 0)
+            {
+                throw new ArgumentException(
+                    Resources.FormatViewLocationFormatsIsRequired(nameof(RazorViewEngineOptions.AreaViewLocationFormats)),
+                    nameof(optionsAccessor));
+            }
+
             _pageFactory = pageFactory;
             _pageActivator = pageActivator;
-            _viewLocationExpanders = optionsAccessor.Value.ViewLocationExpanders;
             _htmlEncoder = htmlEncoder;
             _logger = loggerFactory.CreateLogger<RazorViewEngine>();
             ViewLookupCache = new MemoryCache(new MemoryCacheOptions
@@ -59,47 +76,6 @@ namespace Microsoft.AspNetCore.Mvc.Razor
                 CompactOnMemoryPressure = false
             });
         }
-
-        /// <summary>
-        /// Gets the locations where this instance of <see cref="RazorViewEngine"/> will search for views.
-        /// </summary>
-        /// <remarks>
-        /// The locations of the views returned from controllers that do not belong to an area.
-        /// Locations are composite format strings (see http://msdn.microsoft.com/en-us/library/txafckwd.aspx),
-        /// which contains following indexes:
-        /// {0} - Action Name
-        /// {1} - Controller Name
-        /// The values for these locations are case-sensitive on case-sensitive file systems.
-        /// For example, the view for the <c>Test</c> action of <c>HomeController</c> should be located at
-        /// <c>/Views/Home/Test.cshtml</c>. Locations such as <c>/views/home/test.cshtml</c> would not be discovered
-        /// </remarks>
-        public virtual IEnumerable<string> ViewLocationFormats { get; } = new[]
-        {
-            "/Views/{1}/{0}" + ViewExtension,
-            "/Views/Shared/{0}" + ViewExtension,
-        };
-
-        /// <summary>
-        /// Gets the locations where this instance of <see cref="RazorViewEngine"/> will search for views within an
-        /// area.
-        /// </summary>
-        /// <remarks>
-        /// The locations of the views returned from controllers that belong to an area.
-        /// Locations are composite format strings (see http://msdn.microsoft.com/en-us/library/txafckwd.aspx),
-        /// which contains following indexes:
-        /// {0} - Action Name
-        /// {1} - Controller Name
-        /// {2} - Area name
-        /// The values for these locations are case-sensitive on case-sensitive file systems.
-        /// For example, the view for the <c>Test</c> action of <c>HomeController</c> should be located at
-        /// <c>/Views/Home/Test.cshtml</c>. Locations such as <c>/views/home/test.cshtml</c> would not be discovered
-        /// </remarks>
-        public virtual IEnumerable<string> AreaViewLocationFormats { get; } = new[]
-        {
-            "/Areas/{2}/Views/{1}/{0}" + ViewExtension,
-            "/Areas/{2}/Views/Shared/{0}" + ViewExtension,
-            "/Views/Shared/{0}" + ViewExtension,
-        };
 
         /// <summary>
         /// A cache for results of view lookups.
@@ -311,15 +287,15 @@ namespace Microsoft.AspNetCore.Mvc.Razor
                 isMainPage);
             Dictionary<string, string> expanderValues = null;
 
-            if (_viewLocationExpanders.Count > 0)
+            if (_options.ViewLocationExpanders.Count > 0)
             {
                 expanderValues = new Dictionary<string, string>(StringComparer.Ordinal);
                 expanderContext.Values = expanderValues;
 
                 // Perf: Avoid allocations
-                for (var i = 0; i < _viewLocationExpanders.Count; i++)
+                for (var i = 0; i < _options.ViewLocationExpanders.Count; i++)
                 {
-                    _viewLocationExpanders[i].PopulateValues(expanderContext);
+                    _options.ViewLocationExpanders[i].PopulateValues(expanderContext);
                 }
             }
 
@@ -385,13 +361,13 @@ namespace Microsoft.AspNetCore.Mvc.Razor
             ViewLocationCacheKey cacheKey)
         {
             // Only use the area view location formats if we have an area token.
-            var viewLocations = !string.IsNullOrEmpty(expanderContext.AreaName) ?
-                AreaViewLocationFormats :
-                ViewLocationFormats;
+            IEnumerable<string> viewLocations = !string.IsNullOrEmpty(expanderContext.AreaName) ?
+                _options.AreaViewLocationFormats :
+                _options.ViewLocationFormats;
 
-            for (var i = 0; i < _viewLocationExpanders.Count; i++)
+            for (var i = 0; i < _options.ViewLocationExpanders.Count; i++)
             {
-                viewLocations = _viewLocationExpanders[i].ExpandViewLocations(expanderContext, viewLocations);
+                viewLocations = _options.ViewLocationExpanders[i].ExpandViewLocations(expanderContext, viewLocations);
             }
 
             ViewLocationCacheResult cacheResult = null;
--- a/src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngineOptions.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngineOptions.cs
@@ -20,7 +20,7 @@ namespace Microsoft.AspNetCore.Mvc.Razor
         private Action<RoslynCompilationContext> _compilationCallback = c => { };
 
         /// <summary>
-        /// Get a <see cref="IList{IViewLocationExpander}"/> used by the <see cref="RazorViewEngine"/>.
+        /// Gets a <see cref="IList{IViewLocationExpander}"/> used by the <see cref="RazorViewEngine"/>.
         /// </summary>
         public IList<IViewLocationExpander> ViewLocationExpanders { get; }
             = new List<IViewLocationExpander>();
@@ -34,6 +34,61 @@ namespace Microsoft.AspNetCore.Mvc.Razor
         /// rooted at the application root.
         /// </remarks>
         public IList<IFileProvider> FileProviders { get; } = new List<IFileProvider>();
+
+        /// <summary>
+        /// Gets the locations where <see cref="RazorViewEngine"/> will search for views.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// The locations of the views returned from controllers that do not belong to an area.
+        /// Locations are composite format strings (see http://msdn.microsoft.com/en-us/library/txafckwd.aspx),
+        /// which may contain the following format items:
+        /// </para>
+        /// <list type="bullet">
+        /// <item>
+        /// <description>{0} - Action Name</description>
+        /// </item>
+        /// <item>
+        /// <description>{1} - Controller Name</description>
+        /// </item>
+        /// </list>
+        /// <para>
+        /// The values for these locations are case-sensitive on case-sensitive file systems.
+        /// For example, the view for the <c>Test</c> action of <c>HomeController</c> should be located at
+        /// <c>/Views/Home/Test.cshtml</c>. Locations such as <c>/views/home/test.cshtml</c> would not be discovered.
+        /// </para>
+        /// </remarks>
+        public IList<string> ViewLocationFormats { get; } = new List<string>();
+
+        /// <summary>
+        /// Gets the locations where <see cref="RazorViewEngine"/> will search for views within an
+        /// area.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// The locations of the views returned from controllers that belong to an area.
+        /// Locations are composite format strings (see http://msdn.microsoft.com/en-us/library/txafckwd.aspx),
+        /// which may contain the following format items:
+        /// </para>
+        /// <list type="bullet">
+        /// <item>
+        /// <description>{0} - Action Name</description>
+        /// </item>
+        /// <item>
+        /// <description>{1} - Controller Name</description>
+        /// </item>
+        /// <item>
+        /// <description>{2} - Area Name</description>
+        /// </item>
+        /// </list>
+        /// <para>
+        /// The values for these locations are case-sensitive on case-sensitive file systems.
+        /// For example, the view for the <c>Test</c> action of <c>HomeController</c> under <c>Admin</c> area should
+        /// be located at <c>/Areas/Admin/Views/Home/Test.cshtml</c>.
+        /// Locations such as <c>/areas/admin/views/home/test.cshtml</c> would not be discovered.
+        /// </para>
+        /// </remarks>
+        public IList<string> AreaViewLocationFormats { get; } = new List<string>();
 
         /// <summary>
         /// Gets or sets the callback that is used to customize Razor compilation
--- a/src/Microsoft.AspNetCore.Mvc.Razor/Resources.resx
+++ b/src/Microsoft.AspNetCore.Mvc.Razor/Resources.resx
@@ -203,4 +203,7 @@
   <data name="Compilation_DependencyContextIsNotSpecified" xml:space="preserve">
     <value>The Razor page '{0}' failed to compile. Ensure that your application's {1} sets the '{2}' compilation property.</value>
   </data>
+  <data name="ViewLocationFormatsIsRequired" xml:space="preserve">
+    <value>'{0}' cannot be empty. These locations are required to locate a view for rendering.</value>
+  </data>
 </root>
--- a/test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorViewEngineOptionsTest.cs
+++ b/test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorViewEngineOptionsTest.cs
@@ -1,6 +1,8 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+using System.Linq;
 using Microsoft.AspNetCore.Mvc.ApplicationParts;
 using Microsoft.AspNetCore.Mvc.Internal;
 using Microsoft.Extensions.DependencyInjection;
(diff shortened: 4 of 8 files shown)


src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngine.cs
```diff
@@ -46,60 +48,30 @@ public class RazorViewEngine : IRazorViewEngine
...
-        /// </remarks>
-        public virtual IEnumerable<string> AreaViewLocationFormats { get; } = new[]
-        {
-            "/Areas/{2}/Views/{1}/{0}" + ViewExtension,
-            "/Areas/{2}/Views/Shared/{0}" + ViewExtension,
-            "/Views/Shared/{0}" + ViewExtension,
-        };
+            if (_options.ViewLocationFormats.Count == 0)
+            {
+                throw new InvalidOperationException(Resources.ViewLocationFormatsIsRequired);
```
dougbu: Suggest `new ArgumentException(Resources.ViewLocationFormatsIsRequired, nameof(options))`

src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngine.cs
```diff
@@ -385,13 +357,13 @@ public string GetAbsolutePath(string executingFilePath, string pagePath)
             ViewLocationCacheKey cacheKey)
         {
             // Only use the area view location formats if we have an area token.
-            var viewLocations = !string.IsNullOrEmpty(expanderContext.AreaName) ?
-                AreaViewLocationFormats :
-                ViewLocationFormats;
+            IEnumerable<string> viewLocations = !string.IsNullOrEmpty(expanderContext.AreaName) ?
```
dougbu: Restore `var`

src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngine.cs
```diff
@@ -21,23 +20,26 @@ namespace Microsoft.AspNetCore.Mvc.Razor
...
-    /// by default. For the controllers in an area, views should exist in <see cref="AreaViewLocationFormats"/>.
+    /// For <c>ViewResults</c> returned from controllers, views should be located in
+    /// <see cref="RazorViewEngineOptions.ViewLocationFormats"/>
+    /// by default. For the controllers in an area, views should exist in
+    /// <see cref="RazorViewEngineOptions.AreaViewLocationFormats"/>.
     /// </remarks>
     public class RazorViewEngine : IRazorViewEngine
     {
-        private const string ViewExtension = ".cshtml";
+        public const string ViewExtension = ".cshtml";
```
pranavkm: `public static readonly`

src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngine.cs
```diff
@@ -46,60 +48,30 @@ public class RazorViewEngine : IRazorViewEngine
             IRazorPageFactoryProvider pageFactory,
             IRazorPageActivator pageActivator,
             HtmlEncoder htmlEncoder,
-            IOptions<RazorViewEngineOptions> optionsAccessor,
+            IOptions<RazorViewEngineOptions> options,
             ILoggerFactory loggerFactory)
         {
             _pageFactory = pageFactory;
             _pageActivator = pageActivator;
-            _viewLocationExpanders = optionsAccessor.Value.ViewLocationExpanders;
```
pranavkm: `_options = options.Value;`

src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngine.cs
```diff
@@ -46,60 +48,30 @@ public class RazorViewEngine : IRazorViewEngine
...
-        /// The locations of the views returned from controllers that do not belong to an area.
-        /// Locations are composite format strings (see http://msdn.microsoft.com/en-us/library/txafckwd.aspx),
-        /// which contains following indexes:
-        /// {0} - Action Name
-        /// {1} - Controller Name
-        /// The values for these locations are case-sensitive on case-sensitive file systems.
-        /// For example, the view for the <c>Test</c> action of <c>HomeController</c> should be located at
-        /// <c>/Views/Home/Test.cshtml</c>. Locations such as <c>/views/home/test.cshtml</c> would not be discovered
-        /// </remarks>
-        public virtual IEnumerable<string> ViewLocationFormats { get; } = new[]
```
dougbu: I expected `public virtual IEnumerable<string> ViewLocationFormats { get; }` with `ViewLocationFormats = _options.ViewLocationFormats;` in the constructor. Why is a subclass no longer able to extend / override the list of formats?

src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngine.cs
```diff
@@ -46,60 +48,30 @@ public class RazorViewEngine : IRazorViewEngine
             IRazorPageFactoryProvider pageFactory,
             IRazorPageActivator pageActivator,
             HtmlEncoder htmlEncoder,
-            IOptions<RazorViewEngineOptions> optionsAccessor,
+            IOptions<RazorViewEngineOptions> options,
```
pranavkm: Why change this?

src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngine.cs
pranavkm: Nevermind :)

src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngineOptions.cs
```diff
@@ -19,8 +19,20 @@ public class RazorViewEngineOptions
             new CSharpCompilationOptions(CodeAnalysis.OutputKind.DynamicallyLinkedLibrary);
         private Action<RoslynCompilationContext> _compilationCallback = c => { };
 
+        public RazorViewEngineOptions()
+        {
+            ViewLocationFormats = new List<string>();
```
pranavkm: Couldn't this declared inline like before?

src/Microsoft.AspNetCore.Mvc.Razor/Resources.resx
```diff
@@ -203,4 +203,10 @@
   <data name="Compilation_DependencyContextIsNotSpecified" xml:space="preserve">
     <value>The Razor page '{0}' failed to compile. Ensure that your application's {1} sets the '{2}' compilation property.</value>
   </data>
+  <data name="AreaViewLocationFormatsIsRequired" xml:space="preserve">
+    <value>AreaViewLocationFormats cannot be empty and is required to locate a view for rendering.</value>
```
pranavkm: Use  a format argument here and specify it in the calling code via `nameof`.

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorViewEngineOptionsTest.cs
```diff
@@ -12,6 +13,65 @@ namespace Microsoft.AspNetCore.Mvc.Razor
...
+                foreach (var location in areaViewLocations)
+                {
+                    options.AreaViewLocationFormats.Add(location);
+                }
+            });
+            var serviceProvider = services.BuildServiceProvider();
+
+            // Assert
+            var accessor = serviceProvider.GetRequiredService<IOptions<RazorViewEngineOptions>>();
+            Assert.Equal(areaViewLocations, accessor.Value.AreaViewLocationFormats.ToArray());
```
pranavkm: Does it require `ToArray`?

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorViewEngineOptionsTest.cs
```diff
@@ -12,6 +13,65 @@ namespace Microsoft.AspNetCore.Mvc.Razor
     public class RazorViewEngineOptionsTest
     {
         [Fact]
+        public void AreaViewLocationFormats_ContainsExpectedLocations()
+        {
+            // Arrange
+            var services = new ServiceCollection().AddOptions();
+            var areaViewLocations = new string[]
```
pranavkm: `var areaViewLocations = new []` (shouldn't need `string`)

src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngineOptions.cs
```diff
@@ -19,8 +19,20 @@ public class RazorViewEngineOptions
...
+            ViewLocationFormats.Add("/Views/{1}/{0}" + RazorViewEngine.ViewExtension);
+            ViewLocationFormats.Add("/Views/Shared/{0}" + RazorViewEngine.ViewExtension);
+
+            AreaViewLocationFormats = new List<string>();
+            AreaViewLocationFormats.Add("/Areas/{2}/Views/{1}/{0}" + RazorViewEngine.ViewExtension);
+            AreaViewLocationFormats.Add("/Areas/{2}/Views/Shared/{0}" + RazorViewEngine.ViewExtension);
+            AreaViewLocationFormats.Add("/Views/Shared/{0}" + RazorViewEngine.ViewExtension);
```
dougbu: The defaults should instead be in `RazorViewEngineOptionsSetup` and (@pranavkm ??) `DependencyContextRazorViewEngineOptionsSetup`.

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorViewEngineOptionsTest.cs
```diff
@@ -12,6 +13,65 @@ namespace Microsoft.AspNetCore.Mvc.Razor
...
+            {
+                "/Areas/{2}/Views/{1}/{0}.cshtml",
```
pranavkm: Can you use different strings than the defaults here?

src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngineOptions.cs
```diff
@@ -36,6 +48,38 @@ public class RazorViewEngineOptions
...
+        /// The locations of the views returned from controllers that do not belong to an area.
+        /// Locations are composite format strings (see http://msdn.microsoft.com/en-us/library/txafckwd.aspx),
+        /// which contains following indexes:
+        /// {0} - Action Name
+        /// {1} - Controller Name
+        /// The values for these locations are case-sensitive on case-sensitive file systems.
+        /// For example, the view for the <c>Test</c> action of <c>HomeController</c> should be located at
+        /// <c>/Views/Home/Test.cshtml</c>. Locations such as <c>/views/home/test.cshtml</c> would not be discovered
+        /// </remarks>
+        public IList<string> ViewLocationFormats { get; }
```
dougbu: Add `= new List<string>();` here and below and remove the explicit constructor.

src/Microsoft.AspNetCore.Mvc.Razor/Resources.resx
dougbu: Suggest two sentences: `... empty. These locations are required ...`. Same below.

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorViewEngineOptionsTest.cs
dougbu: Shouldn't need `ToArray()` since both are enumerable.

dougbu:
:watch:

src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngineOptions.cs
kichalla: > The defaults should instead be in RazorViewEngineOptionsSetup

Good point

src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngineOptions.cs
kichalla: I didn't want to do that as some code would be inline and some here, but based on @dougbu  comment, since i will be moving some code here to options setup, i can do inline now

src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngine.cs
kichalla: These no longer seem required. A user inheriting from this view engine anyway needs to supply razor view options to the base type in which case he still has the same experience as our engine i.e expecting the location formats from options.

kichalla:
:up: :date:

src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngine.cs
dougbu: Why change the `public` API and _remove_ a capability, especially after RC? If a user is subclassing `RazorViewEngine` today, we still want that to work tomorrow.

src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngine.cs
dougbu: +1 this is now more confusing because you're using the names `options` and `_options` for different types: `IOptions<RazorViewEngineOptions>` and `RazorViewEngineOptions`. Revert the parameter name change.

src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngine.cs
```diff
@@ -386,12 +362,12 @@ public string GetAbsolutePath(string executingFilePath, string pagePath)
         {
             // Only use the area view location formats if we have an area token.
             var viewLocations = !string.IsNullOrEmpty(expanderContext.AreaName) ?
-                AreaViewLocationFormats :
-                ViewLocationFormats;
+                _options.AreaViewLocationFormats.AsEnumerable() :
+                _options.ViewLocationFormats.AsEnumerable();
```
dougbu: What are the `AsEnumerable()` calls for? If you want to ensure the variable has type `IEnumerable<string>`, declare it with that type -- one of the few cases where not-`var` makes sense.

src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngineOptions.cs
```diff
@@ -36,6 +36,38 @@ public class RazorViewEngineOptions
...
+        /// </summary>
+        /// <remarks>
+        /// The locations of the views returned from controllers that do not belong to an area.
+        /// Locations are composite format strings (see http://msdn.microsoft.com/en-us/library/txafckwd.aspx),
+        /// which contains following indexes:
+        /// {0} - Action Name
+        /// {1} - Controller Name
+        /// The values for these locations are case-sensitive on case-sensitive file systems.
+        /// For example, the view for the <c>Test</c> action of <c>HomeController</c> should be located at
+        /// <c>/Views/Home/Test.cshtml</c>. Locations such as <c>/views/home/test.cshtml</c> would not be discovered
```
dougbu: Need some `<para>` and perhaps `<ul>` love or this whole remark will run together messily.

src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngineOptions.cs
```diff
@@ -36,6 +36,38 @@ public class RazorViewEngineOptions
         public IList<IFileProvider> FileProviders { get; } = new List<IFileProvider>();
 
         /// <summary>
+        /// Gets the locations where <see cref="RazorViewEngine"/> will search for views.
+        /// </summary>
+        /// <remarks>
+        /// The locations of the views returned from controllers that do not belong to an area.
+        /// Locations are composite format strings (see http://msdn.microsoft.com/en-us/library/txafckwd.aspx),
+        /// which contains following indexes:
```
dougbu: "which may contain"

dougbu: And the page you linked to describes these as index components of format items. Will be more understandable to use "format items" here.

src/Microsoft.AspNetCore.Mvc.Razor/Resources.resx
```diff
@@ -203,4 +203,10 @@
   <data name="Compilation_DependencyContextIsNotSpecified" xml:space="preserve">
     <value>The Razor page '{0}' failed to compile. Ensure that your application's {1} sets the '{2}' compilation property.</value>
   </data>
+  <data name="AreaViewLocationFormatsIsRequired" xml:space="preserve">
+    <value>'{0}' cannot be empty. These locations are required to locate a view for rendering.</value>
+  </data>
+  <data name="ViewLocationFormatsIsRequired" xml:space="preserve">
+    <value>'{0}' cannot be empty. These locations are required to locate a view for rendering.</value>
+  </data>
```
dougbu: Get rid of one of these. That is, follow the all-important Highlander principle :smiley_cat:

src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngine.cs
```diff
@@ -46,60 +48,34 @@ public class RazorViewEngine : IRazorViewEngine
...
+                    Resources.FormatViewLocationFormatsIsRequired(nameof(RazorViewEngineOptions.ViewLocationFormats)),
+                    nameof(options));
+            }
+
+            if (_options.AreaViewLocationFormats.Count == 0)
+            {
+                throw new ArgumentException(
+                    Resources.FormatAreaViewLocationFormatsIsRequired(nameof(RazorViewEngineOptions.AreaViewLocationFormats)),
+                    nameof(options));
+            }
```
dougbu: Move these checks to the top of the constructor.

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorViewEngineOptionsTest.cs
```diff
@@ -12,6 +13,65 @@ namespace Microsoft.AspNetCore.Mvc.Razor
...
+        public void AreaViewLocationFormats_ContainsExpectedLocations()
+        {
+            // Arrange
+            var services = new ServiceCollection().AddOptions();
+            var areaViewLocations = new[]
+            {
+                "/Areas/{2}/MvcViews/{1}/{0}.cshtml",
+                "/Areas/{2}/MvcViews/Shared/{0}.cshtml",
+                "/MvcViews/Shared/{0}.cshtml"
+            }.AsEnumerable();
```
dougbu: Throw `AsEnumerable()` into the nearest garbage heap. Not sure why the `Equal` assertion isn't working correctly with an array and a list but just `new`ing up a `List<string>` would be a much better workaround.

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorViewEngineTest.cs
```diff
@@ -1710,18 +1677,26 @@ private RazorViewEngine CreateSuccessfulViewEngine()
...
             pageFactory = pageFactory ?? Mock.Of<IRazorPageFactoryProvider>();
             return new TestableRazorViewEngine(
                 pageFactory,
-                GetOptionsAccessor(expanders));
+                GetOptions(expanders, viewLocationFormats, areaViewLocationFormats));
         }
 
-        private static IOptions<RazorViewEngineOptions> GetOptionsAccessor(
-            IEnumerable<IViewLocationExpander> expanders = null)
+        private static IOptions<RazorViewEngineOptions> GetOptions(
```
dougbu: Why rename this method? It does return an options accessory.

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorViewEngineTest.cs
```diff
@@ -1710,18 +1677,26 @@ private RazorViewEngine CreateSuccessfulViewEngine()
...
+            IEnumerable<IViewLocationExpander> expanders = null,
+            IEnumerable<string> viewLocationFormats = null,
+            IEnumerable<string> areaViewLocationFormats = null)
         {
+            var optionsSetup = new RazorViewEngineOptionsSetup(Mock.Of<IHostingEnvironment>());
```
dougbu: Feels strange to bring `RazorViewEngineOptionsSetup` into unit tests of `RazorViewEngine`. Suggest instead that this method should add the defaults when `viewLocationFormats` or `areaViewLocationFormats` are null. That also removes the need for `Clear()` calls below.

dougbu:
:watch:

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorViewEngineOptionsTest.cs
kichalla: I will rather keep the ToArray that i had earlier

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorViewEngineTest.cs
kichalla: > Suggest instead that this method should add the defaults when viewLocationFormats or areaViewLocationFormats are null

`RazorViewEngineOptionsSetup` is the one which has the default values, so I am using it here. Not sure what is weird about it

src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngine.cs
kichalla: > If a user is subclassing RazorViewEngine today, we still want that to work tomorrow.

Subclassing `RazorViewEngine` is a rarely occurring scenario, so I think this is fine.

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorViewEngineOptionsTest.cs
```diff
@@ -12,6 +13,65 @@ namespace Microsoft.AspNetCore.Mvc.Razor
...
+            builder.AddRazorOptions(options =>
+            {
+                options.AreaViewLocationFormats.Clear();
+
+                foreach (var location in areaViewLocations)
+                {
+                    options.AreaViewLocationFormats.Add(location);
+                }
+            });
+            var serviceProvider = services.BuildServiceProvider();
```
dougbu: What part of this is actually the `// Act`? I suspect the `AddRazorOptions()` call and the rest should be in `// Arrange` or `// Assert`. If so, this is a test of `MvcRazorMvcBuilderExtensions`.

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorViewEngineOptionsTest.cs
dougbu: As discussed offline, Roslyn must be getting confused. Though IntelliSense shows the `Assert.Equal()` overload is the `IEnumerable<T>` one, the test actually makes calls to `Assert.Equal<object>(array, list)` or something similar. Easy fixes include `Assert.Equal<string>(array, list)` or `Assert.Equal(array, list, StringComparer.Ordinal)`.

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorViewEngineTest.cs
dougbu: As I said, it's weird to use (and then override) the `RazorViewEngineOptionsSetup` instance in a test of `RazorViewEngine`. These are unit tests and we should be careful about adding unnecessary dependencies.

dougbu: Specifically what I'm suggesting is forgetting about `RazorViewEngineOptionsSetup` here. Instead,
``` c#
if (viewLocationFormats == null)
{
    // Do tests that don't pass in formats even care what the values are? If not
    options.ViewLocationFormats.Add("nowhere");
    // If they do, ...
    options.ViewLocationFormats.Add("/Views/{1}/{0}.cshtml");
    options.ViewLocationFormats.Add("/Views/Shared/{0}.cshtml");
}
else
{
     foreach (var location in viewLocationFormats)
     {
         options.ViewLocationFormats.Add(location);
     }
}
```

src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngine.cs
dougbu: Check w/ @Eilon since this is a completely optional breaking change.

dougbu: Plus we're not just talking about subclasses: These properties were `public` and available to anything that wanted to _read_ them.

src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngine.cs
dougbu: Never mind. This is for the [line 370](https://github.com/aspnet/Mvc/pull/4556/files#diff-557f4da6c8548e623754e6eaf781ea5dR370).

src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngineOptions.cs
```diff
@@ -36,6 +36,60 @@ public class RazorViewEngineOptions
         public IList<IFileProvider> FileProviders { get; } = new List<IFileProvider>();
 
         /// <summary>
+        /// Gets the locations where <see cref="RazorViewEngine"/> will search for views.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// The locations of the views returned from controllers that do not belong to an area.
+        /// Locations are composite format strings (see http://msdn.microsoft.com/en-us/library/txafckwd.aspx),
+        /// which may contain following indexes:
```
dougbu: `which may contain the following format items:` (missing "the" and use "format items").

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorViewEngineOptionsTest.cs
dougbu: Why is the service lookup now the `// Act`?

src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngine.cs
Eilon: So I thought this looked suspicious, but for a different reason. In MVC 5.x and earlier this property was *settable*, which provided good value (and is why I made a particular remark in the bug) - you could instantiate a view engine, set this property, and go along your merry way. But here it's read-only anyway, so its value of being even public is somewhat dubious, though it's perhaps nice for debugging. Now, it is *virtual*, which means a derived class can "set" it to return whatever it wants, but we're introducing a new way for that to happen anyway (via options) that a derived class can use just as easily.

So the question is: what real value does this property have, aside from perhaps debugging? (And that's somewhat of a dubious value as well.)

Eilon: Ok I should have scrolled down like 20 lines - I didn't realize we kept the properties. That's fine too 😄

dougbu: @kichalla didn't keep the properties. The only remaining property is `ViewLocationCache`.

> what real value does this property have, aside from perhaps debugging?

Allows any overrides or additions in a subclass to _use_ the formats w/o going through and storing in a separate field the options accessor.

Eilon: Oh sorry I didn't realize I was looking at the next *file* (the options class) where the properties moved to.

But anyway, only the view engine's existing logic *uses* the data. The "outsider's" job is to only *provide* the data. What other possible use is there?

dougbu: > What other possible use is there?

I said "subclass" in my most recent comment. But I understand your refutation of my more general "available to anything that wanted to _read_ them" comment above.

dougbu: Turning this around: What is the value in hiding the formats and breaking compat w/ RC2 bits?

Eilon: We already made this property nearly useless in ASP.NET Core MVC. It's key value historically is that it was settable, but as it is no longer settable, its value is nearly zero.

I did a few searches for these properties and every example is just trying to *set* them - none is trying to read them.

Eilon:
@kichalla / @dougbu what is the status of this PR?

kichalla:
:up: :date:

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorViewEngineTest.cs
dougbu: :ok: :see_no_evil:

dougbu:
:shipit: w/ the small change discussed offline to make the focus of the new tests [here](https://github.com/aspnet/Mvc/pull/4556#discussion-diff-61650152) more clear.

PR #4557, opened 2016-04-28
[Perf]  Replace Activator.CreateInstance with cached delegate in RazorPageFactoryProvider

### LINKED ISSUE #4470: Review use of Activator.CreateInstance

mnltejaswini:
Fixes #4470

Isolated changes corresponding to RazorPageFactoryProvider from #4524 for simplicity. For more details please refer to #4524 

Before Optimization
--------------------------
image

After Optimization
----------------------
image

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor/Internal/CompilerCacheResult.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor/Internal/CompilerCacheResult.cs
@@ -3,7 +3,9 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq.Expressions;
+using System.Reflection;
 using Microsoft.AspNetCore.Mvc.Razor.Compilation;
 using Microsoft.Extensions.Primitives;
 
@@ -18,9 +20,10 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Internal
         /// Initializes a new instance of <see cref="CompilerCacheResult"/> with the specified
         /// <see cref="Compilation.CompilationResult"/>.
         /// </summary>
+        /// <param name="relativePath">Application relative path to the view file.</param>
         /// <param name="compilationResult">The <see cref="Compilation.CompilationResult"/>.</param>
-        public CompilerCacheResult(CompilationResult compilationResult)
-            : this(compilationResult, new IChangeToken[0])
+        public CompilerCacheResult(string relativePath, CompilationResult compilationResult)
+            : this(relativePath, compilationResult, new IChangeToken[0])
         {
         }
 
@@ -28,10 +31,11 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Internal
         /// Initializes a new instance of <see cref="CompilerCacheResult"/> with the specified
         /// <see cref="Compilation.CompilationResult"/>.
         /// </summary>
+        /// <param name="relativePath">Application relative path to the view file.</param>
         /// <param name="compilationResult">The <see cref="Compilation.CompilationResult"/>.</param>
         /// <param name="expirationTokens">One or more <see cref="IChangeToken"/> instances that indicate when
         /// this result has expired.</param>
-        public CompilerCacheResult(CompilationResult compilationResult, IList<IChangeToken> expirationTokens)
+        public CompilerCacheResult(string relativePath, CompilationResult compilationResult, IList<IChangeToken> expirationTokens)
         {
             if (expirationTokens == null)
             {
@@ -39,7 +43,19 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Internal
             }
 
             ExpirationTokens = expirationTokens;
-            PageCreator = Expression.Lambda<Func<object>>(Expression.New(compilationResult.CompiledType)).Compile();
+            var compiledType = compilationResult.CompiledType;
+
+            var newExpression = Expression.New(compiledType);
+
+            var pathProperty = compiledType.GetProperty(nameof(IRazorPage.Path)) ?? 
+                (MemberInfo)compiledType.GetField(nameof(IRazorPage.Path));
+            Debug.Assert(pathProperty != null);
+
+            var propertyBindExpression = Expression.Bind(pathProperty, Expression.Constant(relativePath));
+            var objectInitializeExpression = Expression.MemberInit(newExpression, propertyBindExpression);
+            PageFactory =  Expression
+                .Lambda<Func<IRazorPage>>(objectInitializeExpression)
+                .Compile();
         }
 
         /// <summary>
@@ -56,7 +72,7 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Internal
             }
 
             ExpirationTokens = expirationTokens;
-            PageCreator = null;
+            PageFactory = null;
         }
 
         /// <summary>
@@ -67,12 +83,12 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Internal
         /// <summary>
         /// Gets a value that determines if the view was successfully found and compiled.
         /// </summary>
-        public bool Success => PageCreator != null;
+        public bool Success => PageFactory != null;
 
         /// <summary>
-        /// Gets a delegate that creates an instance of the <see cref="CompilationResult.CompiledType"/>.
+        /// Gets a delegate that creates an instance of the <see cref="IRazorPage"/>.
         /// </summary>
-        public Func<object> PageCreator { get; }
+        public Func<IRazorPage> PageFactory { get; }
 
     }
 }
--- a/src/Microsoft.AspNetCore.Mvc.Razor/Internal/CompilerCacheResult.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor/Internal/CompilerCacheResult.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq.Expressions;
 using Microsoft.AspNetCore.Mvc.Razor.Compilation;
 using Microsoft.Extensions.Primitives;
 
@@ -37,9 +38,8 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Internal
                 throw new ArgumentNullException(nameof(expirationTokens));
             }
 
-            CompilationResult = compilationResult;
-            Success = true;
             ExpirationTokens = expirationTokens;
+            PageCreator = Expression.Lambda<Func<object>>(Expression.New(compilationResult.CompiledType)).Compile();
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


mnltejaswini:
@pranavkm  can you please review.

dougbu:
> Isolated changes corresponding to RazorPageFactoryProvider from #4524 for simplicity.

Please close any no-long-relevant PRs. I would also appreciate before / after pics for the individual changes. That makes it much easier to see the perf / complexity trade-offs.

src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultRazorPageFactoryProvider.cs
```diff
@@ -74,18 +74,19 @@ public RazorPageFactoryResult CreateFactory(string relativePath)
         /// <summary>
         /// Creates a factory for <see cref="IRazorPage"/>.
         /// </summary>
-        /// <param name="compiledType">The <see cref="Type"/> to produce an instance of <see cref="IRazorPage"/>
+        /// <param name="pageCreator">The <see cref="Func{Object}"/> to produce an instance of <see cref="IRazorPage"/>
```
dougbu: Since we know the return type and are going to cast to it, why not `Func<IRazorPage>`?

test/Microsoft.AspNetCore.Mvc.Razor.Test/Internal/CompilerCacheTest.cs
```diff
@@ -58,7 +58,7 @@ public void GetOrAdd_ReturnsCompilationResultFromFactory()
 
             // Assert
             Assert.True(result.Success);
-            Assert.Same(type, result.CompilationResult.CompiledType);
+            Assert.Same(type, result.PageCreator().GetType());
```
dougbu: Change all of these to `Assert.IsType(type, result.PageCreator());`.

test/Microsoft.AspNetCore.Mvc.Razor.Test/Internal/CompilerCacheTest.cs
```diff
@@ -442,7 +442,7 @@ public void ConstructorNormalizesPrecompiledViewPath(string viewPath)
             // Assert
             var result1 = task1.Result;
             var result2 = task2.Result;
-            Assert.Same(result1.CompilationResult.CompiledType, result2.CompilationResult.CompiledType);
+            Assert.Same(result1.PageCreator().GetType(), result2.PageCreator().GetType());
```
dougbu: Will `Assert.Same(result1,PageCreator, result2.PageCreator);` work?

src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultRazorPageFactoryProvider.cs
dougbu: A `ViewLocationCacheItem` already contains a `Func<IRazorPage> PageFactory` property and `RazorPageFactoryResult` contains a similar `RazorPageFactory` property. May be caching similar things at two different levels i.e. before and after setting the `Path` property on the page. We should step back a bit e.g. create a single `Func<string, IRazorPage>` built using Linq expressions.

pranavkm: We could modify the creation of `CompilerCacheResult` to pass in the path and construct the result of `GetPageFactory` as a compiled expression.

mnltejaswini:
:up::date:

src/Microsoft.AspNetCore.Mvc.Razor/Internal/CompilerCacheResult.cs
```diff
@@ -17,29 +20,42 @@ public struct CompilerCacheResult
...
 
-            CompilationResult = compilationResult;
-            Success = true;
             ExpirationTokens = expirationTokens;
+            var compiledType = compilationResult.CompiledType;
+
+            var newExpression = Expression.New(compiledType);
+
+            var pathProperty = compiledType.GetProperty(nameof(IRazorPage.Path)) ?? 
+                (MemberInfo)compiledType.GetField(nameof(IRazorPage.Path));
```
pranavkm: Isn't this always going to be a property? Why look for a field?

src/Microsoft.AspNetCore.Mvc.Razor/Internal/CompilerCacheResult.cs
```diff
@@ -17,29 +20,42 @@ public struct CompilerCacheResult
         /// Initializes a new instance of <see cref="CompilerCacheResult"/> with the specified
         /// <see cref="Compilation.CompilationResult"/>.
         /// </summary>
+        /// <param name="relativePath">Application relative path to the view file.</param>
```
dougbu: Either hyphenate "application-relative" or switch to the much more common `Path of the view file relative to the application base.`

src/Microsoft.AspNetCore.Mvc.Razor/Internal/CompilerCacheResult.cs
```diff
@@ -17,29 +20,42 @@ public struct CompilerCacheResult
...
-            CompilationResult = compilationResult;
-            Success = true;
             ExpirationTokens = expirationTokens;
+            var compiledType = compilationResult.CompiledType;
+
+            var newExpression = Expression.New(compiledType);
+
+            var pathProperty = compiledType.GetProperty(nameof(IRazorPage.Path)) ?? 
+                (MemberInfo)compiledType.GetField(nameof(IRazorPage.Path));
+            Debug.Assert(pathProperty != null);
```
dougbu: In addition to not checking for a field, this is a user error, not something we should be asserting. If the user uses `@inherits NotAnIRazorPage` they'll get a fairly-understandable `InvalidCastException`. That behaviour should remain the same.

test/Microsoft.AspNetCore.Mvc.Razor.Test/Internal/CompilerCacheTest.cs
```diff
@@ -58,7 +58,8 @@ public void GetOrAdd_ReturnsCompilationResultFromFactory()
 
             // Assert
             Assert.True(result.Success);
-            Assert.Same(type, result.CompilationResult.CompiledType);
+            Assert.IsType<TestView>(result.PageFactory());
```
dougbu: Get rid of the `type` variables if you're not going to use them.

src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultRazorPageFactoryProvider.cs
```diff
@@ -62,30 +62,12 @@ public RazorPageFactoryResult CreateFactory(string relativePath)
...
         }
-
-        /// <summary>
-        /// Creates a factory for <see cref="IRazorPage"/>.
-        /// </summary>
-        /// <param name="compiledType">The <see cref="Type"/> to produce an instance of <see cref="IRazorPage"/>
-        /// from.</param>
-        /// <param name="relativePath">The application relative path of the page.</param>
-        /// <returns>A factory for <paramref name="compiledType"/>.</returns>
-        protected virtual Func<IRazorPage> GetPageFactory(Type compiledType, string relativePath)
```
pranavkm: This was meant to be analogous to ControllerActivator but we don't give the user a type now to do anything useful. I'm ok with removing this. cc @Eilon for API change

src/Microsoft.AspNetCore.Mvc.Razor/Internal/CompilerCacheResult.cs
```diff
@@ -3,6 +3,9 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
```
dougbu: Should not need this `using`.

dougbu:
@mnltejaswini in the future leave the process at 100% of the CPU i.e. don't double down on `FindView` for this PR. As is, can't tell whether spending less time in some _part_ of a method is significant.

dougbu:
:watch:

src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultRazorPageFactoryProvider.cs
Eilon: This is in an Internal namespace, so do we care what the API looks like beyond being functional? Or is there some deeper repercussion of this?

pranavkm: Nope. Just calling out that this used to be an extensibility point. This allowed "decorating" the page before returning to the view engine if you derived from this type.

src/Microsoft.AspNetCore.Mvc.Razor/Internal/CompilerCacheResult.cs
mnltejaswini: Added for extra guard.

src/Microsoft.AspNetCore.Mvc.Razor/Internal/CompilerCacheResult.cs
mnltejaswini: If there is no Path property, it will result in NullReferenceExpeption. Not sure if this will happen at all so added an assertion.

src/Microsoft.AspNetCore.Mvc.Razor/Internal/CompilerCacheResult.cs
pranavkm: But it's on the interface. Why would it not exist on the type?

pranavkm: https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Razor/IRazorPage.cs#L39

mnltejaswini:
:up::date:

pranavkm:
:shipit:

src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultRazorPageFactoryProvider.cs
Eilon: It's internal, not an extensibility point 😄

test/Microsoft.AspNetCore.Mvc.Razor.Test/Internal/CompilerCacheTest.cs
dougbu: thx

dougbu:
:shipit:

mnltejaswini:
Merged

PR #4565, opened 2016-04-30
Make ControllerArgumentBinder.BindActionArgumentsAsync override-able

imranbaloch:
When creating custom IControllerActionArgumentBinder, sometimes we just need to inherit from base (ControllerArgumentBinder) class and call the base implementation if required.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerArgumentBinder.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerArgumentBinder.cs
@@ -38,7 +38,7 @@ namespace Microsoft.AspNetCore.Mvc.Internal
             _validator = validator;
         }
 
-        public Task<IDictionary<string, object>> BindActionArgumentsAsync(
+        public virtual Task<IDictionary<string, object>> BindActionArgumentsAsync(
             ControllerContext controllerContext,
             object controller)
         {
@@ -97,7 +97,7 @@ namespace Microsoft.AspNetCore.Mvc.Internal
             return actionArguments;
         }
 
-        public async Task<ModelBindingResult?> BindModelAsync(
+        public virtual async Task<ModelBindingResult?> BindModelAsync(
             ParameterDescriptor parameter,
             ControllerContext controllerContext)
         {


Eilon:
cc @rynowak - thoughts?

@imranbaloch FYI this type is in a `.Internal` namespace, meaning it is not a supported API and could change at any time.

rynowak:
@imranbaloch - I don't think that there's any hard in your change, but as was mentioned here - this class is `.Internal` so it's not really a supported extensibility point?

What is it that you want to accomplish? We might find another way to support your needs that's going to have a long-term future.

imranbaloch:
@Eilon @rynowak thanks for reply. 

1) If it is internal namespace then why you guys making this class public? Lets assumed that if you guys RTM with same strategy(internal namespace and public classes). If a developer use a public class inside internal namespace then it will be not supported api?

2)  If you guys make this class internal then is it make sense to put this internal class in IOC container? Currently, it is registered in container  https://github.com/aspnet/Mvc/blob/756cd2dab8d12ab954e82328a5b9732114d32a27/src/Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs#L175

3) ControllerArgumentBinder class looks very important for me but sometimes I may need to return my own _IDictionary<string, object>_(may be from db or other place) depending upon a condition at runtime or call the the base(ControllerArgumentBinder class) implementation. The extensibility in this pull request allows me to achieve this. If you think louder you will find lot of use cases of using custom or base BindActionArgumentsAsync/BindModelAsync implementation at the same time.

Eilon:
@imranbaloch 

> 1) If it is internal namespace then why you guys making this class public? Lets assumed that if you guys RTM with same strategy(internal namespace and public classes). If a developer use a public class inside internal namespace then it will be not supported api?

We do it because there are often scenarios we weren't able to consider, so we err on the side of making things "public" in the C#/.NET sense, but they are indeed *not* supported. This allows customers to dig into internal data structures and behaviors to solve their own business needs. If we hear that something is useful, we will consider making things public.

> 2) If you guys make this class internal then is it make sense to put this internal class in IOC container? Currently, it is registered in container 

It is common for frameworks to have their own internal usage of DI, which is what the case here is. Using DI can be a good development pattern to follow, so MVC uses it internally, and app developers will also use the same container.

> 3) ControllerArgumentBinder class looks very important for me but sometimes I may need to return my own `IDictionary<string, object>` (may be from db or other place) depending upon a condition at runtime or call the the base(ControllerArgumentBinder class) implementation. The extensibility in this pull request allows me to achieve this. If you think louder you will find lot of use cases of using custom or base BindActionArgumentsAsync/BindModelAsync implementation at the same time.

We are certainly open to discussing making things more public, more extensible, etc.

What I recommend is logging an issue describing the particular scenario that you have, and then we can dive into the best way to implement this in the system.

Thanks!

Eilon:
Closing this PR for now; please log a new issue that goes into the particular scenario that you have and let's see if we can come up with an appropriate change for MVC that could satisfy that. Thanks!

PR #4577, opened 2016-05-02
[Fixes #4575] Logging ArgumentOutOfRangeException

### LINKED ISSUE #4575: Logging ArgumentOutOfRangeException
I am not sure if this should go here or in [aspnet/Logging](https://github.com/aspnet/Logging) but [this log message](https://github.com/aspnet/Mvc/blob/73dc7bf48b76c2b82ef09dd56714d1d3fe2948f1/src/Microsoft.AspNetCore.Mvc.Core/Internal/MvcCoreLoggerExtensions.cs#L177) causes a ArgumentOutOfRangeException when you attempt to enumerate over the values. Looks like another case of [aspnet/Logging@410](https://github.com/aspnet/Logging/issues/410) where the LoggingMessage definition is using Define<string> when it should be using Define().

kichalla:
@BrennanConroy @Eilon

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Formatters.Json/Internal/MvcJsonLoggerExtensions.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Formatters.Json/Internal/MvcJsonLoggerExtensions.cs
@@ -8,13 +8,13 @@ namespace Microsoft.AspNetCore.Mvc.Formatters.Json.Internal
 {
     internal static class MvcJsonLoggerExtensions
     {
-        private static readonly Action<ILogger, string, Exception> _jsonInputFormatterCrashed;
+        private static readonly Action<ILogger, Exception> _jsonInputFormatterCrashed;
 
         private static readonly Action<ILogger, string, Exception> _jsonResultExecuting;
 
         static MvcJsonLoggerExtensions()
         {
-            _jsonInputFormatterCrashed = LoggerMessage.Define<string>(
+            _jsonInputFormatterCrashed = LoggerMessage.Define(
                 LogLevel.Debug,
                 1,
                 "JSON input formatter threw an exception.");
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Mvc.Formatters.Json/Internal/MvcJsonLoggerExtensions.cs
```diff
@@ -8,13 +8,13 @@ namespace Microsoft.AspNetCore.Mvc.Formatters.Json.Internal
...
-        private static readonly Action<ILogger, string, Exception> _jsonInputFormatterCrashed;
+        private static readonly Action<ILogger, Exception> _jsonInputFormatterCrashed;
 
         private static readonly Action<ILogger, string, Exception> _jsonResultExecuting;
 
         static MvcJsonLoggerExtensions()
         {
-            _jsonInputFormatterCrashed = LoggerMessage.Define<string>(
+            _jsonInputFormatterCrashed = LoggerMessage.Define(
                 LogLevel.Debug,
```
kichalla: @Eilon Not related to this PR, but should this be `Error` log level?

BrennanConroy:
:shipit:

src/Microsoft.AspNetCore.Mvc.Formatters.Json/Internal/MvcJsonLoggerExtensions.cs
Eilon: Not sure, log a bug?

kichalla:
d30290c31068bec0267dc14852e5dd89a3e67ac2

src/Microsoft.AspNetCore.Mvc.Formatters.Json/Internal/MvcJsonLoggerExtensions.cs
rynowak: No, this is correct as-is, this is a failure to parse the request body. The error will *generally* be surfaced to the user as a 400

PR #4579, opened 2016-05-02
Change RequireHttpsAttribute to 302 by default

### LINKED ISSUE #4578: RequireHttps should be redirect 302 (not 301)
See https://github.com/aspnet/Mvc/issues/4561.

In this bug we just want to change the default, nothing more. In the other bug we'll make it configurable.

khellang:
Fixes #4578

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/RequireHttpsAttribute.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/RequireHttpsAttribute.cs
@@ -84,7 +84,7 @@ namespace Microsoft.AspNetCore.Mvc
                     request.QueryString.ToUriComponent());
 
                 // redirect to HTTPS version of page
-                filterContext.Result = new RedirectResult(newUrl, permanent: true);
+                filterContext.Result = new RedirectResult(newUrl, permanent: false);
             }
         }
     }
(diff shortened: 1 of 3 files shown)


kichalla:
@Eilon

src/Microsoft.AspNetCore.Mvc.Core/RequireHttpsAttribute.cs
```diff
@@ -76,15 +83,15 @@ protected virtual void HandleNonHttpsRequest(AuthorizationFilterContext filterCo
                     host = new HostString(host.Host);
                 }
 
-                var newUrl = string.Concat(
-                    "https://",
-                    host.ToUriComponent(),
-                    request.PathBase.ToUriComponent(),
-                    request.Path.ToUriComponent(),
-                    request.QueryString.ToUriComponent());
+                var newUrl = UriHelper.Encode(
```
Eilon: Why change this?

Eilon:
@khellang thanks for the PR! For `release` we'd like to take only the change from 301 to 302. The rest we can do for `dev`. It's a bit late in the game for us to take any bigger change for `release` unfortunately.

src/Microsoft.AspNetCore.Mvc.Core/RequireHttpsAttribute.cs
khellang: I thought it made sense to utilize the `UriHelper` from HttpAbstractions when I was through here anyway :smile:

khellang:
@Eilon Alright. Updated to just change the default. Will send a PR with the property to dev when this is merged back :smile:

Eilon:
Awesome, thanks for understanding!

kichalla:
Thanks @khellang. This PR has now been merged into release and back to dev.

3c2bdfd368f1b0ed81e92f9287f43a58be06e6e0

PR #4580, opened 2016-05-02
Quick fix: Simplify `OrderUsingHtmlHelpers.cshtml` layout

dougbu:
- intent was to make HTML helper output close to the tag helper output but it got over-complicated
- the output files for `Order.cshtml` and `OrderUsingHtmlHelpers.cshtml` now differ only in attribute order
 - i.e. HTML helpers output attribute alphabetically while tag helpers output them in order added


ajaybhargavb:
:shipit:

dougbu:
ad1375a

PR #4586, opened 2016-05-03
[Design][Perf] Avoid Coercing simple action methods to return Task<object>

### LINKED ISSUE #4539: Revisit generated expression for action method invoker
Sampling data from HelloWorldMvcCore
image

This method returns `Task` and we're forcing it down the `Task<object>` path which is the slowest path.

We should look at the different combination of factors here (`void`, `object`, `IActionResult`) X (`Task`, non-`Task`) and come up with a solution that works for each of these cases in a performant way.

mnltejaswini:
Targeted to fix #4539 

Before
---------
image

After
-----
image

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionExecutor.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionExecutor.cs
@@ -28,6 +28,14 @@ namespace Microsoft.AspNetCore.Mvc.Internal
             return actionMethodExecutor.ExecuteAsync(instance, orderedActionArguments);
         }
 
+        public static Task ExecuteVoidAsync(
+            ObjectMethodExecutor actionMethodExecutor,
+            object instance,
+            object[] orderedActionArguments)
+        {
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionExecutor.cs
```diff
@@ -28,6 +28,32 @@ public static class ControllerActionExecutor
...
+
+            if (actionMethodExecutor.CanExecuteInSync)
+            {
+                var result = actionMethodExecutor.Execute(instance, orderedActionArguments);
+                if (returnType == typeof(Task))
+                {
+                    await (Task)result;
+                }
+
+                return createActionResult(actionMethodExecutor, result);
```
rynowak: This is still going to result in allocating a `Task<T>` when the action method is synchronous. Maybe this would be a good place to use `ValueTask<T>`

mnltejaswini:
This is tracked by another PR https://github.com/aspnet/Mvc/pull/4661 and hence this is closed

PR #4600, opened 2016-05-09
Changed RequireHttpsAttribute to allow toggling of permanent requests.

### LINKED ISSUE #4561: RequireHttps - should it be redirect 302 by default?
RequireHttps uses a 301 permanent redirect.

https://github.com/aspnet/Mvc/blob/ee2cfa1963ee7f8195130d41330f94d4622ca89a/src/Microsoft.AspNetCore.Mvc.Core/RequireHttpsAttribute.cs#L87

It used to be 302 in MVC 5 AFAIK so this adds to the confusion.
All the major browsers cache 301 indefinitely so if you use RequireHttps in your code you can never remove it.
Otherwise you will have code that may not match the behaviour of your site depending on the state of the users browser cache.  This can cause some confusion without an in depth knowledge of browser behaviour and 301 vs. 302.

I realise I could override the attribute to say something like RequireHttpsNotPermanent but given the one way street that is 301 I propose making RequireHttps 302 (like it used to be) by default.  Make 301 an opt in so you are aware of the consequences.

Perhaps with an overridden RequireHttpsPermanent or RequireHttps("Permanent") attribute for 301.

Discussed at length here
https://github.com/aspnet/Security/issues/798

CaffeinatedCM:
Fixes #4561

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/RequireHttpsAttribute.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/RequireHttpsAttribute.cs
@@ -15,6 +15,18 @@ namespace Microsoft.AspNetCore.Mvc
     [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, Inherited = true, AllowMultiple = false)]
     public class RequireHttpsAttribute : Attribute, IAuthorizationFilter, IOrderedFilter
     {
+
+        private readonly bool _permanentRedirect;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RequireHttpsAttribute" /> class.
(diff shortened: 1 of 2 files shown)


Eilon:
@CaffeinatedCM thanks so much for the PR! We got another similar PR just shortly after this one at https://github.com/aspnet/Mvc/pull/4602.

I've looked at the design and code for both of them and I slightly prefer the approach taken at https://github.com/aspnet/Mvc/pull/4602. For example, the other PR just adds a get/set property, which I think would have improved readability when using the attribute, as compared to the constructor parameter approach.

Again, thank you for the PR, and we hope to be able to work with you again in the future.

PR #4602, opened 2016-05-09
Added Permanent property to configure redirect in RequireHttpsAttribute

### LINKED ISSUE #4561: RequireHttps - should it be redirect 302 by default?

khellang:
As promised in #4579

Fixes #4561

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/RequireHttpsAttribute.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/RequireHttpsAttribute.cs
@@ -15,6 +15,12 @@ namespace Microsoft.AspNetCore.Mvc
     [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, Inherited = true, AllowMultiple = false)]
     public class RequireHttpsAttribute : Attribute, IAuthorizationFilter, IOrderedFilter
     {
+        /// <summary>
+        /// Specifies whether a permanent redirect, <c>301 Moved Permanently</c>,
+        /// should be used instead of a temporary redirect, <c>302 Found</c>.
+        /// </summary>
+        public bool Permanent { get; set; }
+
         /// <inheritdoc />
         public int Order { get; set; }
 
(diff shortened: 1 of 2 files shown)


Eilon:
@khellang thanks for the PR! I've assigned this to an engineer on the team and we'll get it reviewed and merged ASAP.

sebastienros:
LGTM, I am still trying to run it locally, facing some unrelated issues right now.

ShadowDancer:
That's great, but how can I set Permanent property after adding filter globally?
Something like this:
`
            services.AddMvc(
                opt =>
                {
                    opt.Filters.Add(typeof (RequireHttpsAttribute));
                });`

Currently I have to create my own wrapper filter. Maybe such filter should be in framework(RequireHttpsPermanentAttribute) or configurable with IOptions?

pranavkm:
@ShadowDancer, you should be able to add an instance of the attribute to the filter collection:
`opt.Filters.Add(new RequireHttpsAttribute { Permanent = true })`

PR #4620, opened 2016-05-12
Change our cached task to a field

rynowak:
This makes it more inlinable and saves a volatile lookup on netstandard.


sebastienros:
:ship:

Issue #4612, opened 2016-05-11
ObjectResultExecutor should hold the Accept header in higher regard


tuespetre:
#### The problem behavior

Here is the rundown of the issue I'm seeing:

1. I have a controller action decorated with `[Produces("application/vnd.my-app+json; version=1")]`
2. I have a custom `OutputFormatter` that will say it `CanWriteResult` as long as the content type is `application/vnd.my-app+json`, regardless of media type parameter values
3. I have `RespectBrowserAcceptHeader` (why the 'browser' in the name?) set to `true` to make sure any `Accept` header is respected (Just in case, I guess? [It looks like maybe it only has to do with `*/*` wildcard media types...](https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectResultExecutor.cs#L262))
4. I send a `fetch` to the controller action from my browser's console with an `Accept` header value of `application/vnd.derp` and receive a 200 response with a `Content-Type` header value of `application/vnd.my-app; version=1`

#### The expected behavior

I would expect content negotiation to honor the `Accept` request header in all cases and respond with `406` whenever it cannot locate a formatter that applies to a media type specified in the `Accept` header, rather than responding with whatever it can muster up even when the client has explicitly declared what it will accept.

#### Possible workaround

In RC2, the `OutputFormatterWriteContext` exposes the `HttpContext`. A custom formatter could take advantage of that in order to properly respect the `Accept` header value, but this breaks the current negotiator/formatter pattern and causes the formatter to violate the Single Responsibility Principle.

In RC1, well, there's nothing that easy that I can see offhand.

khellang:
Have you added the `HttpNotAcceptableOutputFormatter`?

javiercn:
Hi, my guess is that there might be a formatter being chosen to serve the request. You can override https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectResultExecutor.cs#L401 call base and see if it returns a result to verify that is what is happening.

See https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectResultExecutor.cs#L180 for more details

And https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectResultExecutor.cs#L128 is where I expect to get a 406.

[...]

PR #4623, opened 2016-05-12
Alter content negotiation algorithm

tuespetre:
- Always respect explicit Accept headers (except for when they do contain the 'browser wildcard' of `*/*` unless `RespectBrowserAcceptHeader` is configured to `true`, as was the case already), and return `406` in the response when a formatter cannot be found that respects the header
- Remove `HttpNotAcceptableOutputFormatter` and its tests since it appeared to only exist to serve the old algorithm, plus it wasn't really a formatter
- Remove the `FailedContentNegotiation` property from `OutputFormatterCanWriteContext` since it was only used by `HttpNotAcceptableOutputFormatter`
- Remove test cases that confirmed the behavior of using whatever available formatter even if it does not align with an explicit Accept header
- Restructure/consolidate various conneg tests into a single test that receives a series of `ConnegScenario` objects via `[Theory][MemberData]`

### Other considerations

If it is 'too late in the game' to change the behavior this way, or if the current behavior needs to stick around for legacy reasons, it would now be very easy to add a flag along the lines of `RespectAcceptHeaderStrict` with a default of `false`. Then, while selecting a formatter, this flag can be checked at 'the last minute' and used to determine whether or not to grab the first available formatter without any regard to the Accept header.


Eilon:
@dougbu / @rynowak thoughts on this behavior?

rynowak:
The original design goal here was to make conneg a simple 'system' where as policy as possible was put into formatters rather than the 'system`. That's the line of thinking at gets us the `NoContentFormatter`, the `StringFormatter` and the `NotAcceptableFormatter`. The idea is that these formatters determine the behavior and that the intended way of configuring the system is to add and remove formatters.

Personally, I think the fact that we have `FailedContentNegotiation` and `IgnoreBrowserAcceptHeader` shows that this isn't super tenable. I like what removing `FailedContentNegotiation` does to the design - we can easily make the behavior of whether or not to a do a 'failed' conneg pass an option (as much as I hate boolean switches on options).

Whether or not this is the right default behavior, I want to get more input, but I think this moves the design to a better place.

Eilon:
I chatted with @rynowak and @danroth27 about this.

We feel that if we were to do something for v1.0.0 around this that the change would be more like this:
* Remove `HttpNotAcceptableOutputFormatter`, because no one likes it
* Add a new `bool` option called something like `FallbackToHttp406IfNoFormatterMatches` and have it be `false` by default

The reasons for this are that we think the current behavior is the most frequently requested behavior: you get JSON by default. It is also fully compliant with HTTP because all HTTP says is that the response's content must match the declared response `Content-Type` header. There is no rule that the response has to respect the request's `Accept` headers, or alternatively fail if it can't handle any of them. It's all a "best effort," which we already have.

@tuespetre any thoughts on this? Would you consider updating the PR with this behavior?

tuespetre:
@eilon true, the RFC (at least the 'good old' one) says 'should', not 'must'. I will update this PR with the switch as soon as I get the chance.

dougbu:
Much appreciated @tuespetre

tuespetre:
:up: :date:

@eilon I went the 'opposite' way with the switch and called it `UseFallbackFormatterWhenContentNegotiationFails` with a default of `true`, because it (in my opinion) better reflects the actual behavior of the component. I hope this is acceptable.

dougbu:
@tuespetre we tend to have `bool`s that default to `false`. That avoids having to set the default at all and makes it easier to reason about the code.

Eilon:
I'm not super concerned about whether it's `true` or `false` by default, but the current name one way or another doesn't lend itself well to being a Boolean value. Though the name `FallbackToHttp406IfNoFormatterMatches` was a bit tongue-in-cheek, it made it relatively clear what happens and when if you set it (though perhaps it isn't super clear what happens if it's `false`).

The name `UseFallbackFormatterWhenContentNegotiationFails` doesn't make it very clear to me either way what happens. Plus, it isn't clear what "fails" means in this context: content negotiation couldn't negotiate? Or it threw an exception?

tuespetre:
@eilon yeah this seems to be a prime example of 'naming is hard.' Everything that might describe it well enough seems really long and unwieldy.

**5 minutes later**

Actually, I may have found an applicable short one. We have `RespectBrowserAcceptHeader`, so how about `RespectUnsatisfiableAcceptHeader`? The default would be `false`.

dougbu:
@tuespetre suggest you go ahead w/ `RespectUnsatisfiableAcceptHeader` since @Eilon hasn't objected :smiley_cat:

Eilon:
Sorry, I don't quite get what `RespectUnsatisfiableAcceptHeader` is trying to convey.

tuespetre:
What if we "flip it" so to speak: "DisregardUnsatisfiableAcceptHeader"

"If we cannot satisfy the Accept header, we will disregard the Accept header."

Eilon:
Thanks for everyone's patience here.

I chatted with @danroth27 and we think something like this could work:

```c#
/// Gets or sets a value that decides whether an HTTP 406 Not Acceptable response
/// will be returned if no formatter has been selected to format the response.
public ReturnHttpNotAcceptable { get; set; } // false by default
```

tuespetre:
:up::date:

thanks @Eilon

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectResultExecutor.cs
```diff
@@ -175,71 +182,57 @@ public virtual Task ExecuteAsync(ActionContext context, ObjectResult result)
                 throw new ArgumentNullException(nameof(formatters));
             }
 
-            // Check if any content-type was explicitly set (for example, via ProducesAttribute
-            // or URL path extension mapping). If yes, then ignore content-negotiation and use this content-type.
-            if (contentTypes.Count == 1)
+            var request = formatterContext.HttpContext.Request;
+            var acceptableMediaTypes = GetAcceptableMediaTypes(contentTypes, request);
+            IOutputFormatter selectedFormatter = null;
+            bool selectFormatterWithoutRegardingAcceptHeader = false;
```
Eilon: Van use some more `var` here 😄

Eilon:
@tuespetre thanks!

I'll let @dougbu do the final review on this.

dougbu:
Thanks @Eilon and @tuespetre. @tuespetre I'll likely review this in detail sometime this evening.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectResultExecutor.cs
```diff
@@ -65,6 +66,11 @@ public class ObjectResultExecutor
         protected bool RespectBrowserAcceptHeader { get; }
 
         /// <summary>
+        /// Gets the value of <see cref="MvcOptions.ReturnHttpNotAcceptable"/>.
+        /// </summary>
+        protected bool ReturnHttpNotAcceptable { get; }
```
dougbu: Not for this PR but it's odd that `RespectBrowserAcceptHeader` and now `ReturnHttpNotAcceptable` are `protected` properties when both could be `private` fields. There's no user scenario for subclassing this `.Internal`-namespace type. @tuespetre you've been working w/ this code more recently that me. Am I missing anything?

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectResultExecutor.cs
```diff
@@ -131,6 +137,7 @@ public virtual Task ExecuteAsync(ActionContext context, ObjectResult result)
                 // No formatter supports this.
                 Logger.NoFormatter(formatterContext);
 
+                formatterContext.ContentType = default(StringSegment);
```
dougbu: Why is this necessary? Won't be using `formatterContext` when returning a `406` status.

test/Microsoft.AspNetCore.Mvc.Core.Test/Infrastructure/ObjectResultExecutorTest.cs
```diff
@@ -20,37 +20,6 @@ namespace Microsoft.AspNetCore.Mvc.Internal
 {
     public class ObjectResultExecutorTest
     {
-        [Fact]
```
dougbu: I'm having trouble understanding changes to existing tests due to this refactoring. Please restore the old approach so that changes forced by the `src` changes are obvious.

dougbu: No problem with `ConnegScenario` in general but would prefer to see it introduced w/o `src` changes in the same PR. (Well, "conneg" is a fairly ugly shorthand but whatever.)

test/Microsoft.AspNetCore.Mvc.Core.Test/Infrastructure/ObjectResultExecutorTest.cs
```diff
@@ -75,177 +43,110 @@ public void SelectFormatter_WithNoProvidedContentType_DoesConneg()
...
+                    },
+                    // Respects Accept header value by failing to negotiate an unsupported media type
+                    new ConnegScenario
+                    {
+                        ContentTypes = { "text/plain", "text/xml", "application/json"},
+                        AcceptHeaderValue = "application/vnd.custom",
+                        ResultingContentType = default(StringSegment),
+                        Formatters = { cannotWriteFormatter, jsonFormatter },
+                        ResultingFormatter = null,
+                        ReturnHttpNotAcceptable = true,
```
dougbu: Should check the `Response.StatusCode` as well in all of these scenarios -- generally not set but `406` in this scenario.

Should also include `ReturnHttpNotAcceptable = true` in additional variants of the above cases. That'll confirm this setting has no impact when an acceptable formatter is found.

dougbu:
:watch: for a few questions and test changes. Liking the improved `ObjectResultExecutor` readability.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectResultExecutor.cs
tuespetre: I'm not sure that they need to be `protected`; I just made the new property `protected` to match what was there :smile_cat:

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectResultExecutor.cs
tuespetre: Noted; how does the team feel about using `default` to facilitate `var` for `null` values?

test/Microsoft.AspNetCore.Mvc.Core.Test/Infrastructure/ObjectResultExecutorTest.cs
tuespetre: The status code is set by `ExecuteAsync` rather than `SelectFormatter`, and it is already verified by the `ExecuteAsync_NoFormatterFound_Returns406` test.

I will add explicit values for `ReturnHttpNotAcceptable` to verify that the settings does not negatively impact cases where there is an acceptable formatter.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectResultExecutor.cs
dougbu: > how does the team feel about using `default` to facilitate `var` for `null` values?

You won't see much use of `default` in MVC at all. Put another way, the `selectedFormatter` declaration is fine.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectResultExecutor.cs
tuespetre: This was for testing but since I'm dialing those changes back I will remove it.

tuespetre:
:up::date:

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectResultExecutor.cs
```diff
@@ -175,71 +181,57 @@ public virtual Task ExecuteAsync(ActionContext context, ObjectResult result)
                 throw new ArgumentNullException(nameof(formatters));
             }
 
-            // Check if any content-type was explicitly set (for example, via ProducesAttribute
-            // or URL path extension mapping). If yes, then ignore content-negotiation and use this content-type.
-            if (contentTypes.Count == 1)
+            var request = formatterContext.HttpContext.Request;
+            var acceptableMediaTypes = GetAcceptableMediaTypes(contentTypes, request);
+            var selectedFormatter = default(IOutputFormatter);
```
dougbu: FYI only: The main downside of this is some developers read `default` and think the type is a `struct`. In this case, they then have to _think_ about `struct`s and `interface`s :smiling_imp: 

In future I suggest sticking w/ `IOutputFormatter selectedFormatter = null;`, an approach that also works when the variable doesn't need to be initialized.

dougbu:
> I will add explicit values for `ReturnHttpNotAcceptable` to verify that the settings does not negatively impact cases where there is an acceptable formatter.

Where did you do this? I can't see it.

tuespetre:
> Where did you do this? I can't see it.

I stashed away the whole `ConnegScenario` thing and left the test suite with as little modification as possible due to your feedback:

> I'm having trouble understanding changes to existing tests due to this refactoring. Please restore the old approach so that changes forced by the `src` changes are obvious.
>
> No problem with `ConnegScenario` in general but would prefer to see it introduced w/o `src` changes in the same PR. (Well, "conneg" is a fairly ugly shorthand but whatever.)

Let me know if it's cool to add those changes as their own commit to this same pull request and I will do it.

dougbu:
> Let me know if it's cool to add those changes as their own commit to this same pull request and I will do it.

Let's get all of the tests working as-is. Right now a couple of functional tests are failing because `NotAcceptable` is returned where the test expects `Ok`. Should have mentioned this earlier, sorry.

I suggest running `build --quiet verify` on the command line.

Are you saying you need to add `ConnegScenario` in order to add a few tests? I was thinking more in terms of changing a few `[Fact]`s into `[Theory]`s with a `bool returnHttpNotAcceptable` parameter that's used in the options for those tests and shouldn't affect their assertions.

tuespetre:
:up::date: 

Sorry about that, I overlooked that when I was 'squashing' things earlier. The failing functional test was actually removed in the initial PR -- there's not really a good, immediate way to recreate what it did with the removal of the `HttpNotAcceptableOutputFormatter`. I've axed it again. Also tweaked the `default(IOutputFormatter)` bit.

dougbu:
:shipit: we're set and I'll get this in soon.

If we need additional assertions or tests, we can add them later. Same for refactoring the tests. Feel free to make suggestions in the form of test-only PRs.

Much appreciated.

dougbu:
Oops, missed fact the commit description doesn't mention #4612. Please add that reference and, if you have time, make sure your commit is up-to-date w/ `dev`.

Again, much appreciated.

tuespetre:
Both of those are done :smile_cat:

Eilon:
@tuespetre we'll get this on today, thanks!!

dougbu:
14f68b6

PR #4632, opened 2016-05-13
Fix #4631 and avoid Task<T> creation in invoker

### LINKED ISSUE #4631: DiagnosticSource extension methods in Invoker can nullref
The code for calling into DiagnosticSource *before* a synchronous filter's `*Exectued` method can nullref and make people sad.

```
_diagnosticSource.BeforeOnResourceExecuted(_resourceExecutedContext, item.Filter);

item.Filter.OnResourceExecuted(await InvokeResourceFilterAsync());

_diagnosticSource.AfterOnResourceExecuted(_resourceExecutedContext, item.Filter); 
```

The bug here is that `_resourceExecutedContext` will usually be null until `InvokeResourceFilterAsync` is called.

The correct code would look like:
```
await InvokeResourceFilterAsync()

_diagnosticSource.BeforeOnResourceExecuted(_resourceExecutedContext, item.Filter);

item.Filter.OnResourceExecuted(_resourceExecutedContext);

_diagnosticSource.AfterOnResourceExecuted(_resourceExecutedContext, item.Filter); 
```

rynowak:
This change corrects and ordering bug between the creation of the
'context' and the diagnostic source event that occurs before a synchronous
filter's 'after' stage.

Also made some simple changes to avoid allocating Task<T> in many common
cases. Now we'll only create the Task<T> when we really need it (async
filters).

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/Internal/FilterActionInvoker.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Internal/FilterActionInvoker.cs
@@ -227,7 +227,7 @@ namespace Microsoft.AspNetCore.Mvc.Internal
             return InvokeResourceFilterAsync();
         }
 
-        private async Task<ResourceExecutedContext> InvokeResourceFilterAsync()
+        private async Task<ResourceExecutedContext> InvokeResourceFilterAwaitedAsync()
         {
             Debug.Assert(_resourceExecutingContext != null);
 
@@ -243,6 +243,16 @@ namespace Microsoft.AspNetCore.Mvc.Internal
                 throw new InvalidOperationException(message);
             }
 
+            await InvokeResourceFilterAsync();
+
+            Debug.Assert(_resourceExecutedContext != null);
+            return _resourceExecutedContext;
+        }
+
+        private async Task InvokeResourceFilterAsync()
+        {
+            Debug.Assert(_resourceExecutingContext != null);
+
             var item = _cursor.GetNextFilter<IResourceFilter, IAsyncResourceFilter>();
             try
             {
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Mvc.Core/Internal/FilterActionInvoker.cs
```diff
@@ -294,9 +304,12 @@ private Task InvokeAllResourceFiltersAsync()
                     }
                     else
                     {
+                        await InvokeResourceFilterAsync();
+                        Debug.Assert(_resourceExecutedContext != null);
+
                         _diagnosticSource.BeforeOnResourceExecuted(_resourceExecutedContext, item.Filter);
 
-                        item.Filter.OnResourceExecuted(await InvokeResourceFilterAsync());
+                        item.Filter.OnResourceExecuted(_resourceExecutedContext);
```
rynowak: This is the bug - when `BeforeOnResourceExecuted` was called we hadn't yet created the `_resourceExecutedContext`. The same bug is present for Resource Action and Result filters.

rynowak: Found this while looking at the `Task<T>` optimization so I decided to include it as it's in the same few lines of code.

src/Microsoft.AspNetCore.Mvc.Core/Internal/FilterActionInvoker.cs
```diff
@@ -243,14 +243,24 @@ private Task InvokeAllResourceFiltersAsync()
                 throw new InvalidOperationException(message);
             }
 
+            await InvokeResourceFilterAsync();
+
+            Debug.Assert(_resourceExecutedContext != null);
+            return _resourceExecutedContext;
+        }
+
+        private async Task InvokeResourceFilterAsync()
```
rynowak: The perf improvement is to change the signature of this method

rynowak:
Allocation data for eliminating these tasks (Techempower Plaintext) - eliminates 240 bytes/request

**Before:**
image

image


**After:**
image

image

src/Microsoft.AspNetCore.Mvc.Core/Internal/FilterActionInvoker.cs
pranavkm: Could you add a test for this? Seems like we're missing DiagnosticSource tests for this.

pranavkm:
:shipit: once a couple tests are added.

src/Microsoft.AspNetCore.Mvc.Core/Internal/FilterActionInvoker.cs
rynowak: Discussed in person and we're going to add asserts to all of these methods. There's an extremely high cost/benefit ratio to adding unit tests for all of these diagnostic source methods.

Issue #4593, opened 2016-05-05
Performance Scenario: MusicStore 


mnltejaswini:
**This is top-level performance scenario for 1.0.0. We're using this item to track the status and all of the individual improvements.**

Code here: https://github.com/aspnet/MusicStore/tree/dev/src/MusicStore
#### Summary

This is a complex browser-based app that uses a variety of ASP.NET features. This is also our most prominent sample and had lots of eyes on it for coding style and feature set.

For simplicity we're measuring the home page.
- Authentication (cookie)
- Database (localdb)
- Session
- Cache TagHelper
- Layout Page
- View Components and Partials
- Diagnostics (logging)
# 
#### Variations

(None)
# 
#### Status

Scenario Status: **complete**

Known Investigations/Improvements:
- [ ] TBD: Update with known issues

mnltejaswini:
@rynowak  Didn't find any issue related to MusicStore scenario. Hence logged a new one.

mnltejaswini:
### 1. Analysis for MusicStore.GetCartAlbumTitles() 
https://github.com/aspnet/MusicStore/blob/dev/src/MusicStore/Models/ShoppingCart.cs#L102 
 Linq BinaryExpression compare is causing MethodInfo[] allocations which is 1.75% of total memory allocations. By changing the Where expression to  .Where(cart => cart.CartId.Equals(_shoppingCartId)) we got rid of MethodInfo[] allocations. 

The profile after the change is 
![image](https://cloud.githubusercontent.com/assets/8011073/15200108/c8310d16-1798-11e6-989f-9a88d7eabee7.png)

mnltejaswini:
2: ViewBufferTextWriter,  ViewBufferPage[], ViewBuffer allocations .
Most of these allocations are coming from 
![image](https://cloud.githubusercontent.com/assets/8011073/15200136/0442bce6-1799-11e6-9b32-b8e2c52e7e7d.png)

RazorPage.StartTagHelperWritingScope() is always followed by RazorPage.EndTagHelperWritingScope(). The code within Razor page executes synchronously and hence each RazorPage can have a simple List based pool for ViewBuffer and ViewBufferTextWriter. They can be rented in the StartTagHelperWritingScope() and can be returned in EndTagHelperWritingScope()

Prototype of these changes resulted in 3.4% saving in total bytes as shown below
[...]

PR #4634, opened 2016-05-13
[Perf]  Fast Generate Url for specific cases to avoid string allocations

mnltejaswini:
Reference Issue #4593 

loadtest with 3000 requests for the home page of the MusicStore app. 

Before 
--------
image

After
------
image

Total savings 1.18% savings in total bytes allocated. 
Saves 22 string allocations per request to the Home page.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
@@ -223,7 +223,7 @@ namespace Microsoft.AspNetCore.Mvc.Routing
         /// <param name="pathData">The <see cref="VirtualPathData"/>.</param>
         /// <param name="fragment">The URL fragment.</param>
         /// <returns>The generated URL.</returns>
-        internal virtual string GenerateUrl(string protocol, string host, VirtualPathData pathData, string fragment)
+        protected virtual string GenerateUrl(string protocol, string host, VirtualPathData pathData, string fragment)
         {
             if (pathData == null)
             {
@@ -294,14 +294,16 @@ namespace Microsoft.AspNetCore.Mvc.Routing
                 if (pathData.VirtualPath.Length == 0)
                 {
                     url = "/";
+                    return true;
                 }
                 else if (pathData.VirtualPath.StartsWith("/", StringComparison.Ordinal))
                 {
                     url = pathData.VirtualPath;
+                    return true;
                 }
             }
 
-            return (url != null);
+            return false;
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
@@ -223,7 +223,7 @@ namespace Microsoft.AspNetCore.Mvc.Routing
         /// <param name="pathData">The <see cref="VirtualPathData"/>.</param>
         /// <param name="fragment">The URL fragment.</param>
         /// <returns>The generated URL.</returns>
-        protected virtual string GenerateUrl(string protocol, string host, VirtualPathData pathData, string fragment)
+        internal virtual string GenerateUrl(string protocol, string host, VirtualPathData pathData, string fragment)
         {
             if (pathData == null)
             {
@@ -232,6 +232,15 @@ namespace Microsoft.AspNetCore.Mvc.Routing
 
             // VirtualPathData.VirtualPath returns string.Empty instead of null.
             Debug.Assert(pathData.VirtualPath != null);
+
+            // Perf: In most of the common cases, GenerateUrl is called with a null protocol, host and fragment. 
+            // In such cases, we might not need to build any URL as the url generated is mostly same as the virtual path available in pathData.
+            // For such common cases, this FastGenerateUrl method saves a string allocation per GenerateUrl call.
+            string url;
+            if (TryFastGenerateUrl(protocol, host, pathData, fragment, out url))
+            {
+                return url;
+            }
 
             var builder = GetStringBuilder();
             try
@@ -265,6 +274,34 @@ namespace Microsoft.AspNetCore.Mvc.Routing
                 // Clear the StringBuilder so that it can reused for the next call.
                 builder.Clear();
             }
+        }
+
+        private bool TryFastGenerateUrl(
+            string protocol, 
+            string host, 
+            VirtualPathData pathData, 
+            string fragment,
+            out string url)
+        {
+            var pathBase = HttpContext.Request.PathBase;
+            url = null;
+
+            if (string.IsNullOrEmpty(protocol) 
+                && string.IsNullOrEmpty(host) 
+                && string.IsNullOrEmpty(fragment)
+                && !pathBase.HasValue)
+            {
+                if (pathData.VirtualPath.Length == 0)
+                {
+                    url = "/";
+                }
+                else if (pathData.VirtualPath.StartsWith("/", StringComparison.Ordinal))
+                {
+                    url = pathData.VirtualPath;
+                }
+            }
+
+            return (url != null);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
```diff
@@ -233,6 +233,13 @@ protected virtual string GenerateUrl(string protocol, string host, VirtualPathDa
             // VirtualPathData.VirtualPath returns string.Empty instead of null.
             Debug.Assert(pathData.VirtualPath != null);
 
+            var url = FastGetUrl(protocol, host, pathData, fragment);
```
rynowak: Name this `FastGenerateUrl` or similar to make it clear that this handles a special case of `GenerateUrl`

src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
```diff
@@ -266,5 +273,33 @@ protected virtual string GenerateUrl(string protocol, string host, VirtualPathDa
...
+            VirtualPathData pathData, 
+            string fragment)
+        {
+            if (string.IsNullOrEmpty(protocol) 
+                && string.IsNullOrEmpty(host) 
+                && string.IsNullOrEmpty(fragment))
+            {
+                var pathBase = HttpContext.Request.PathBase;
+
+                if (!pathBase.HasValue)
```
rynowak: You can roll this into the `if` above

src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
```diff
@@ -233,6 +233,13 @@ protected virtual string GenerateUrl(string protocol, string host, VirtualPathDa
             // VirtualPathData.VirtualPath returns string.Empty instead of null.
             Debug.Assert(pathData.VirtualPath != null);
 
+            var url = FastGetUrl(protocol, host, pathData, fragment);
+
+            if (!string.IsNullOrEmpty(url))
```
rynowak: Why would this ever return the empty string?

src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
```diff
@@ -266,5 +273,33 @@ protected virtual string GenerateUrl(string protocol, string host, VirtualPathDa
...
+
+        private string FastGetUrl(
+            string protocol, 
+            string host, 
+            VirtualPathData pathData, 
+            string fragment)
+        {
+            if (string.IsNullOrEmpty(protocol) 
+                && string.IsNullOrEmpty(host) 
+                && string.IsNullOrEmpty(fragment))
```
rynowak: what happens today when these are empty? Generally we don't treat empty and null as equivalent unless they really are - or we have a legacy/compatibility requirement.

src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
rynowak: This could use some documentation about the finding and the insight that led to the optimization.

> It's pretty common that `GenerateUrl` will be called with a null protocol, host and fragment so in that case ...

rynowak:
Good perf improvement! :watch:

src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
```diff
@@ -266,5 +273,33 @@ protected virtual string GenerateUrl(string protocol, string host, VirtualPathDa
                 builder.Clear();
             }
         }
+
+        private string FastGetUrl(
```
rynowak: Do we have adequate tests for this new code path? I suspect we could stand to add a few

src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
sebastienros: Ryan, what about a `bool TryFastGenerateUrl` ? Or anything that returns a bool to check the result.

rynowak: Yeah that would also be good

src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
```diff
@@ -233,6 +233,16 @@ protected virtual string GenerateUrl(string protocol, string host, VirtualPathDa
             // VirtualPathData.VirtualPath returns string.Empty instead of null.
             Debug.Assert(pathData.VirtualPath != null);
 
+            // Perf: In most of the common cases, GenerateUrl is called with a null protocol, host and fragment. 
+            // In such cases, we might not need to build any Url as the url generated is mostly same as the virtual path available in pathData.
+            // For such common cases, this FastGenerateUrl method saves a string allocation per GenerateUrl call.
```
rynowak: always put `URL` in all caps in comments and documentation. @Eilon gets grumpy if you don't

rynowak:
:shipit: after fixing comment about comments

src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
mnltejaswini: These checks are copied from the original code path https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs#L239

mnltejaswini:
@rynowak  I added unit tests and has taken Sebastien's advise. One final review please

src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
```diff
@@ -266,5 +275,33 @@ protected virtual string GenerateUrl(string protocol, string host, VirtualPathDa
...
+                && !pathBase.HasValue)
+            {
+                if (pathData.VirtualPath.Length == 0)
+                {
+                    url = "/";
+                }
+                else if (pathData.VirtualPath.StartsWith("/", StringComparison.Ordinal))
+                {
+                    url = pathData.VirtualPath;
+                }
```
sebastienros: Easier to read if `return true` here and default path does `return false` without `url != null`

src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
```diff
@@ -223,7 +223,7 @@ private StringBuilder GetStringBuilder()
         /// <param name="pathData">The <see cref="VirtualPathData"/>.</param>
         /// <param name="fragment">The URL fragment.</param>
         /// <returns>The generated URL.</returns>
-        protected virtual string GenerateUrl(string protocol, string host, VirtualPathData pathData, string fragment)
+        internal virtual string GenerateUrl(string protocol, string host, VirtualPathData pathData, string fragment)
```
rynowak: This isn't right. This was intentionally protected so developers could override it.

If you need to call this method directly in tests, create a subclass and a public `new` method that calls base

test/Microsoft.AspNetCore.Mvc.Core.Test/Routing/UrlHelperTest.cs
```diff
@@ -939,6 +939,37 @@ public void Action_RouteValueInvalidation_DoesNotAffectActionAndController_Actio
...
+        [InlineData(null, null, null, "/Hello", null, "/Hello" )]
+        [InlineData(null, null, null, "/Hello", null, "/Hello")]
+        [InlineData("/", null, null, "", null, "/")]
+        [InlineData("/hello/", null, null, "/world", null, "/hello/world")]
+        [InlineData("/hello/", "https", "myhost", "/world", "fragment-value", "https://myhost/hello/world#fragment-value")]
+        public void GenerateUrl_FastAndSlowPathsReturnsExpected(
+            string appBase,
+            string protocol,
+            string host,
+            string virtualPath,
```
rynowak: How about cases where virtual path doesn't start with `/`? There's a check for it in the product code.

rynowak:
:watch:

src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
mnltejaswini: Got it. Will fix it.

mnltejaswini:
:up::date:

rynowak:
:shipit:

mnltejaswini:
:up::date:

PR #4636, opened 2016-05-13
Add asserts to DiagnosticSource

rynowak:


src/Microsoft.AspNetCore.Mvc.Core/Internal/FilterActionInvoker.cs
```diff
@@ -262,24 +262,24 @@ private Task InvokeAllResourceFiltersAsync()
 
                     await item.FilterAsync.OnResourceExecutionAsync(_resourceExecutingContext, InvokeResourceFilterAwaitedAsync);
 
-                    _diagnosticSource.AfterOnResourceExecution(_resourceExecutedContext, item.FilterAsync);
```
rynowak: I actually found another bug of the same variety by adding these asserts. If an async filter short-circuits, then `_resourceExecutedContext` hasn't been instantiated yet.

javiercn:
:shipit:

pranavkm:
LGTM

PR #4638, opened 2016-05-13
Merge invoker back into one class

rynowak:
This will help future generations maintain this class. Notice that the
protected methods that are going away all just call into another
extensibility point (other than the executor). If we need to we could make
that extensible in the future and then we have the same support with fewer
hooks and less complexity.


src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
```diff
@@ -17,108 +19,760 @@
 
 namespace Microsoft.AspNetCore.Mvc.Internal
 {
-    public class ControllerActionInvoker : FilterActionInvoker
+    public class ControllerActionInvoker : IActionInvoker
```
rynowak: Virtually all of this is copy-paste. The other thing I really changed here was the number and ordering of the constructor arguments.

rynowak:
Merged as 1e02fd2e6b46e78ab3f442752b59eb6f4f70e8e9

We review this in person

Issue #4558, opened 2016-04-28
Depend on `HtmlString` from `Microsoft.AspNetCore.Html`.


NTaylorMullen:
`HtmlString` today inherits from `HtmlEncodedString` and only adds the [Empty](https://github.com/aspnet/Mvc/blob/30ace9f35a63da9527281f903c71059accb33274/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Rendering/HtmlString.cs#L14) property. Aside from back compat it seems odd to have both available; might be nice to have just one right way of saying "this is already encoded".

/cc @Eilon

Muchiachio:
:+1: for leaving only `HtmlString` and moving it to abstractions.
Most of the time `IHtmlString` interface is used anyway, which is self explanatory and widely known.

dougbu:
@NTaylorMullen were you really proposing removing `HtmlString`? That's the one that's broadly known, in a commonly-referenced namespace, and what users should use. `HtmlEncodedString` exists primarily because some low-level classes had to be aware of the concept. And back-compat is a great reason to keep the 10 line subclass.

@Muchiachio `IHtmlString` does not exist in ASP.NET Core MVC.

NTaylorMullen:
@dougbu Yes, removing HtmlString.

Muchiachio:
oh sorry I meant `IHtmlContent` :)

NTaylorMullen:
Talked a bit offline. This issue should be "remove `HtmlString`" and i'll file a new issue to rename `HtmlEncodedString` => `HtmlString`

NTaylorMullen:
https://github.com/aspnet/NodeServices/pull/75/commits/24692f07cdba22fbbe420f3941a936ecfb0c9bcd
55922d68d49d0700732d23b8d53c75cb17cad81e

PR #4651, opened 2016-05-16
React to `HtmlEncodedString` rename to `HtmlString`.

NTaylorMullen:
- Removed the Mvc instance of `HtmlString` since it now exists in `HtmlAbstractions`.

#4558
aspnet/HtmlAbstractions#25

/cc @rynowak @dougbu @pranavkm


src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperFormatter.cs
```diff
@@ -14,18 +14,18 @@ namespace Microsoft.AspNetCore.Mvc.TagHelpers.Cache
...
+                throw new ArgumentNullException(nameof(context));
             }
-            
-            if (context.Html == null) 
+
+            if (context.Html == null)
             {
-                throw new ArgumentNullException(nameof(context.Html));    
+                throw new ArgumentNullException(nameof(context.Html));
             }
```
rynowak: This isn't the right exception/message

test/Microsoft.AspNetCore.Mvc.TagHelpers.Test/ImageTagHelperTest.cs
```diff
@@ -142,7 +143,7 @@ public void PreservesOrderOfSourceAttributesWhenRun()
             Assert.Equal(expectedOutput.TagName, output.TagName);
             Assert.Equal(4, output.Attributes.Count);
 
-            for(int i=0; i < expectedOutput.Attributes.Count; i++)
+            for (int i = 0; i < expectedOutput.Attributes.Count; i++)
```
rynowak: var

rynowak:
Is this namespace included with a default `using` in razor? should it be now?

rynowak:
:shipit:

NTaylorMullen:
> Is this namespace included with a default using in razor? should it be now?

It's not; however, I don't think it should be. If a user is creating `HtmlString`s in a view they're doing some really odd things :smile:.

rynowak:
eh, I'd be really surprised if we don't get feedback on compat about this one. Obiv `Html.Raw` is there, but I'm gonna bet `HtmlString` shows up in lots of peoples codez.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperFormatter.cs
NTaylorMullen: /cc @sebastienros I remember having the same comment when you wrote the code but there was a reason. What was the reason?

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperFormatter.cs
```diff
@@ -14,18 +14,13 @@ namespace Microsoft.AspNetCore.Mvc.TagHelpers.Cache
     /// </summary>
     public class DistributedCacheTagHelperFormatter : IDistributedCacheTagHelperFormatter
     {
-        
+
```
pranavkm: Remove?

pranavkm:
Would be nice to follow up on @rynowak 's suggestion on adding it to the default view namespaces.

src/Microsoft.AspNetCore.Mvc.TagHelpers/Cache/DistributedCacheTagHelperFormatter.cs
```diff
@@ -14,18 +14,13 @@ namespace Microsoft.AspNetCore.Mvc.TagHelpers.Cache
...
         /// <inheritdoc />
         public Task<byte[]> SerializeAsync(DistributedCacheTagHelperFormattingContext context)
         {
-            if (context == null) 
-            {
-                throw new ArgumentNullException(nameof(context));    
-            }
-            
-            if (context.Html == null)
```
dougbu: `context.Html == null` is still an error case. Will now `throw` a useless NRE.

Instead fix the `Exception` to be an `ArgumentException` and use our usual `FormatPropertyOfTypeCannotBeNull` resource.

dougbu:
I agree w/ @pranavkm and @rynowak's point about adding the new `HtmlString` to the default namespaces we use automatically in views.

I'm more concerned about user's tag helpers, HTML helper extensions, and similar e.g. this change broke a number of the MVC tag helpers. Not sure scaffolding needs to include `using Microsoft.AspNetCore.Html;` in every controller class. But this definitely needs an Announcement.

dougbu:
:shipit:

Eilon:
I seriously doubt almost anyone uses the `HtmlString` type directly from a CSHTML page. There's almost no good reason to use that type there.

From HTML helpers, sure, but there are much bigger changes one has to make from a CS file anyway. And with VS it's just a <kbd>Ctrl</kbd>+<kbd>.</kbd> away to get the namespace.

javiercn:
Agree with @NTaylorMullen we shouldn't encourage people to new up IHtmlString on their views

PR #4661, opened 2016-05-17
[Perf] Refactoring ControllerActionInvoker to avoid coercing return type to Task<object> 

### LINKED ISSUE #4539: Revisit generated expression for action method invoker

mnltejaswini:
Targeted to fix #4539 
Will update the PR with perf numbers.


test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ControllerActionExecutorTests.cs
```diff
@@ -1,509 +0,0 @@
-// Copyright (c) .NET Foundation. All rights reserved.
```
mnltejaswini: All these tests are moved to ControllerActionInvokerTest.cs

src/Microsoft.AspNetCore.Mvc.Core/Internal/ObjectMethodExecutor.cs
```diff
@@ -132,8 +149,16 @@ private static ActionExecutor WrapVoidAction(VoidActionExecutor executor)
...
-        private static ActionExecutorAsync GetExecutorAsync(MethodInfo methodInfo, TypeInfo targetTypeInfo)
+        private static ActionExecutorAsync GetExecutorAsync(Type taskInnerType, MethodInfo methodInfo, TypeInfo targetTypeInfo)
         {
+            if (taskInnerType == null)
+            {
+                // This will be the case for types which have derived from Task and Task<T> or non Task types.
+                throw new InvalidOperationException(Resources.FormatActionExecutor_UnexpectedTaskInstance(
+                    methodInfo.Name,
+                    methodInfo.DeclaringType));
+            }
```
rynowak: Why do we need to throw an exception in this private method - what code would a user write to hit this?

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ControllerActionExecutorTests.cs
rynowak: But why? Why can't the functionality on `ObjectMethodExecutor` be tested directly anymore?

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ControllerActionInvokerTest.cs
```diff
@@ -1868,39 +1872,19 @@ public void CreateActionResult_ReturnsSameActionResult()
         {
             // Arrange
             var mockActionResult = new Mock<IActionResult>();
+            var actionExecutor = ObjectMethodExecutor.Create(
+                typeof(TestController).GetMethod("Echo"),
```
rynowak: Use nameof

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
```diff
@@ -573,12 +573,51 @@ private Task InvokeAllExceptionFiltersAsync()
...
-                            arguments);
+                        var returnType = _executor.MethodReturnType;
 
-                        result = CreateActionResult( actionMethodInfo.ReturnType, actionReturnValue);
+                        if (returnType == typeof(void))
+                        {
+                            _executor.Execute(_controller, arguments);
+                            result = new EmptyResult();
+                        }
+                        else if (returnType == typeof(IActionResult))
```
rynowak: What if you check `IsMethodAsync` and `IsTypeAssignableFromIActionResult` right here, and throw the exception for null action results right here.

Same comment for  `_executor.TaskGenericType == typeof(IActionResult)` - it will read better and you don't have to add complexity to `CreateActionResult`

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ControllerActionInvokerTest.cs
```diff
@@ -1946,6 +1935,485 @@ public void CreateActionResult_ReturnsObjectContentResult(Type type, object inpu
...
+            // Act
+            await invoker.InvokeAsync();
+
+            // Assert
+            Assert.IsType<EmptyResult>(result);
+        }
+
+        [Fact]
+        // Async actions that has dynamic return type but return Task or Task<T> are not handled in any special way as,
+        // they are not supported. They are considered as regular methods that return object type.
```
rynowak: It's not "not supported" - it's just not a special case

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ControllerActionInvokerTest.cs
```diff
@@ -1946,6 +1935,485 @@ public void CreateActionResult_ReturnsObjectContentResult(Type type, object inpu
...
+                .Verifiable();
+
+            var invoker = CreateInvoker(new[] { filter.Object }, "ReturnTaskAsDynamicValue", actionParameters);
+
+            // Act
+            await invoker.InvokeAsync();
+
+            // Assert
+            var contentResult = Assert.IsType<ObjectResult>(result);
+            Assert.Equal(inputParam1, await (Task<int>)contentResult.Value);
```
rynowak: What's going on here with an await in an `Assert`?

rynowak:
:watch:

mnltejaswini:
:up::date:

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
```diff
@@ -573,12 +573,76 @@ private Task InvokeAllExceptionFiltersAsync()
...
+                        else if (_executor.TaskGenericType == typeof(IActionResult))
+                        {
+                            result = await (Task<IActionResult>)_executor.Execute(_controller, arguments);
+                            if (result == null)
+                            {
+                                throw new InvalidOperationException(
+                                    Resources.FormatActionResult_ActionReturnValueCannotBeNull(typeof(IActionResult)));
+                            }
+                        }
+                        else if (_executor.IsTypeAssignableFromIActionResult)
```
rynowak: `_executor.TaskGenericType == typeof(IActionResult)` and `returnType == typeof(IActionResult)` - why are these cases handled separately from `_executor.IsTypeAssignableFromIActionResult`?

mnltejaswini: Yes, there is one difference. if the _executor.TaskGenericType == typeof(IActionResult)  , we still execute it in sync because we can cast to Task<IActionResult> and await on it. We can get rid of the returntype == typeof(IActionResult)  as this is covered by IsTaskAssignableFromIActionResult and is sync

rynowak: Actually, I think I understand why. If the method is declared to return `Task<ContentResult>` - then we can either cast that to `Task` or `Task<ContentResult`. We can't use `Task<ContentResult>` here because it's a generic type that we don't know, and if we cast it to `Task` then there's no way to get the result at all.

This really seems like the best we can in that case, and it's worth optimizing for because we can do better in the case where the method returns `Task<IActionResult>`.

This actually seems like it's fine.

test/Microsoft.AspNetCore.Mvc.Core.Test/Internal/ControllerActionExecutorTests.cs
rynowak: Discussed in person, we simplified this functionality. These have become integration tests now

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
rynowak: > Yes, there is one difference. if the _executor.TaskGenericType == typeof(IActionResult) , we still execute it in sync because we can cast to Task and await on it. We can get rid of the returntype == typeof(IActionResult) as this is covered by IsTaskAssignableFromIActionResult and is sync

Yeah I'd like to see the case for sync `== typeof(IActionResult)` collapsed with the case for sync `typeof(IActionResult).IsAssignableFrom(...)`. Other than that :shipit:

rynowak:
:shipit: after addressing https://github.com/aspnet/Mvc/pull/4661#discussion_r63934361

PR #4665, opened 2016-05-18
[Fixes #4353] Cleanup application part discovery

### LINKED ISSUE #4353: Cleanup assembly discovery in ApplicationParts
We currently look up the IHostingEnvironment instance on AddMvcCore() from the ServicesCollection instance in the services collection that is passed during startup. This is fragile and we do it for assembly discovery purposes, we shoud find a better way of doing this.

javiercn:


javiercn:
@rynowak

Eilon:
Assigning to @kichalla to review.

kichalla:
:shipit:

Issue #4497, opened 2016-04-18
Using taghelpers in razor views using a model type which is loaded runtime


DavidEmanuelsen:
Hi,

I have created an application which uses a custom fileprovider for loading razor views from disk / database, and these views uses models coming from an assembly which is loaded runtime. I've used the RazorViewEngineOptions.CompilationCallback to load this assembly and the compilation of the views works - except for when using `asp-for` etc. (The compilation callback was implemented as per the example given in #4423)

The model object is created using Activator.CreateInstance where the type is coming from the external assembly.

The errors I'm getting when using taghelpers are:
- `Type expected`
- `Cannot convert lambda expression to intended delegate type because some of the return types in the block are not implicitly convertible to the delegate return type`

These errors don't contain much detail, but I'm assuming the problem is related to the model type which is loaded runtime. Any suggestions on how I can get this to work is greatly appreciated.

dougbu:
If any of the above errors are actually thrown exceptions, please provide the stack traces.

Eilon:
@DavidEmanuelsen indeed, where are you seeing these errors, and can you show us any relevant code from your app?

DavidEmanuelsen:
I have project containing the model to be used which implements an interface which is also know by the running application. This project / class library is build seperately, and in the main application I create an instance of the model like this:

```
Assembly externalAssembly = AssemblyLoadContext.Default.LoadFromAssemblyPath(assemblyPath);
var type = Type.GetType(serviceNamespace 
            + "." + serviceImplementationName 
            + "," + serviceNamespace);

return (IServiceModel)Activator.CreateInstance(type);
```

The compilation callback looks like this:
```
services.Configure<RazorViewEngineOptions>(options =>
{
        var previous = options.CompilationCallback;
        options.CompilationCallback = (context) =>
        {
               previous?.Invoke(context);
               MetadataReference[] references = serviceAssemblyReferences.ToArray();
                    
               context.Compilation = context.Compilation.AddReferences(references);
         };

         options.FileProviders.Add(new CustomFileProvider());
});
```

Where the serviceAssemblyReferences-list is populated with new items as needed before returning from the controller (creating new items using `MetadataReference.CreateFromFile(path)`)

[...]

PR #4672, opened 2016-05-18
Use dependency context from all application parts when compiling views

### LINKED ISSUE #4498: Consider unifying the list of assemblies included in the razor compilation with the of assembly parts of the application.

pranavkm:
Fixes #4498

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/ICompileLibrariesProvider.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/ICompileLibrariesProvider.cs
@@ -1,0 +1,19 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Microsoft.Extensions.DependencyModel;
+
+namespace Microsoft.AspNetCore.Mvc.ApplicationParts
+{
+    /// <summary>
+    /// Exposes <see cref="CompilationLibrary"/> instances from an <see cref="ApplicationPart"/>.
+    /// </summary>
+    public interface ICompileLibrariesProvider
+    {
+        /// <summary>
+        /// Gets the sequence of <see cref="CompilationLibrary"/> instances.
+        /// </summary>
+        IReadOnlyList<CompilationLibrary> GetCompilationLibraries();
+    }
+}
--- a/src/Microsoft.AspNetCore.Mvc.Razor/Compilation/MetadataReferenceFeature.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor/Compilation/MetadataReferenceFeature.cs
@@ -1,0 +1,19 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Microsoft.CodeAnalysis;
+
+namespace Microsoft.AspNetCore.Mvc.Razor.Compilation
+{
+    /// <summary>
+    /// Specifies the list of <see cref="MetadataReference"/> used in Razor compilation.
+    /// </summary>
+    public class MetadataReferenceFeature
+    {
+        /// <summary>
+        /// Gets the <see cref="MetadataReference"/> instances.
+        /// </summary>
+        public IList<MetadataReference> MetadataReferences { get; } = new List<MetadataReference>();
+    }
+}
--- a/src/Microsoft.AspNetCore.Mvc.Razor/Compilation/MetadataReferenceFeatureProvider.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor/Compilation/MetadataReferenceFeatureProvider.cs
@@ -1,0 +1,67 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Reflection.PortableExecutable;
+using Microsoft.AspNetCore.Mvc.ApplicationParts;
+using Microsoft.CodeAnalysis;
+using Microsoft.Extensions.DependencyModel;
+
+namespace Microsoft.AspNetCore.Mvc.Razor.Compilation
+{
+    /// <summary>
+    /// An <see cref="IApplicationFeatureProvider{TFeature}"/> for <see cref="MetadataReferenceFeature"/> that 
+    /// uses <see cref="DependencyContext"/> for registered <see cref="AssemblyPart"/> instances to create 
+    /// <see cref="MetadataReference"/>.
+    /// </summary>
+    public class MetadataReferenceFeatureProvider : IApplicationFeatureProvider<MetadataReferenceFeature>
+    {
+        /// <inheritdoc />
+        public void PopulateFeature(IEnumerable<ApplicationPart> parts, MetadataReferenceFeature feature)
+        {
+            if (parts == null)
+            {
+                throw new ArgumentNullException(nameof(parts));
+            }
+
+            if (feature == null)
+            {
+                throw new ArgumentNullException(nameof(feature));
+            }
+
+            var libraryPaths = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            foreach (var providerPart in parts.OfType<ICompileLibrariesProvider>())
+            {
+                var compileLibraries = providerPart.GetCompilationLibraries();
+
+                for (var i = 0; i < compileLibraries.Count; i++)
+                {
+                    var library = compileLibraries[i];
+                    var referencePaths = library.ResolveReferencePaths();
+                    foreach (var path in referencePaths)
+                    {
+                        if (libraryPaths.Add(path))
+                        {
+                            var metadataReference = CreateMetadataReference(path);
+                            feature.MetadataReferences.Add(metadataReference);
+                        }
+                    }
+                }
+            }
+        }
+
+        private static MetadataReference CreateMetadataReference(string path)
+        {
+            using (var stream = File.OpenRead(path))
+            {
+                var moduleMetadata = ModuleMetadata.CreateFromStream(stream, PEStreamOptions.PrefetchMetadata);
+                var assemblyMetadata = AssemblyMetadata.Create(moduleMetadata);
+
+                return assemblyMetadata.GetReference(filePath: path);
+            }
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultRoslynCompilationService.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultRoslynCompilationService.cs
@@ -7,17 +7,15 @@ using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Reflection;
-using System.Reflection.PortableExecutable;
 using System.Text;
 using System.Threading;
 using Microsoft.AspNetCore.Diagnostics;
-using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Mvc.ApplicationParts;
 using Microsoft.AspNetCore.Mvc.Razor.Compilation;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.Emit;
 using Microsoft.CodeAnalysis.Text;
-using Microsoft.Extensions.DependencyModel;
 using Microsoft.Extensions.FileProviders;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
@@ -29,7 +27,7 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Internal
     /// </summary>
     public class DefaultRoslynCompilationService : ICompilationService
     {
-        private readonly IHostingEnvironment _hostingEnvironment;
+        private readonly ApplicationPartManager _partManager;
         private readonly IFileProvider _fileProvider;
         private readonly Action<RoslynCompilationContext> _compilationCallback;
         private readonly CSharpParseOptions _parseOptions;
@@ -37,22 +35,22 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Internal
         private readonly ILogger _logger;
         private object _applicationReferencesLock = new object();
         private bool _applicationReferencesInitialized;
-        private List<MetadataReference> _applicationReferences;
+        private IList<MetadataReference> _applicationReferences;
 
         /// <summary>
         /// Initalizes a new instance of the <see cref="DefaultRoslynCompilationService"/> class.
         /// </summary>
-        /// <param name="environment">The <see cref="IHostingEnvironment"/>.</param>
+        /// <param name="partManager">The <see cref="ApplicationPartManager"/>.</param>
         /// <param name="optionsAccessor">Accessor to <see cref="RazorViewEngineOptions"/>.</param>
         /// <param name="fileProviderAccessor">The <see cref="IRazorViewEngineFileProviderAccessor"/>.</param>
         /// <param name="loggerFactory">The <see cref="ILoggerFactory"/>.</param>
         public DefaultRoslynCompilationService(
-            IHostingEnvironment environment,
+            ApplicationPartManager partManager,
             IOptions<RazorViewEngineOptions> optionsAccessor,
             IRazorViewEngineFileProviderAccessor fileProviderAccessor,
             ILoggerFactory loggerFactory)
         {
-            _hostingEnvironment = environment;
+            _partManager = partManager;
             _fileProvider = fileProviderAccessor.FileProvider;
             _compilationCallback = optionsAccessor.Value.CompilationCallback;
             _parseOptions = optionsAccessor.Value.ParseOptions;
@@ -60,7 +58,7 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Internal
             _logger = loggerFactory.CreateLogger<DefaultRoslynCompilationService>();
         }
 
-        private List<MetadataReference> ApplicationReferences
+        private IList<MetadataReference> ApplicationReferences
         {
             get
             {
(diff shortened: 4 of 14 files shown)


pranavkm:
cc @javiercn \ @rynowak

test/WebSites/ControllersFromServicesWebSite/AssemblyMetadataReferenceFeatureProvider.cs
```diff
@@ -0,0 +1,20 @@
...
+using Microsoft.AspNetCore.Mvc.Razor.Compilation;
+using Microsoft.CodeAnalysis;
+
+namespace ControllersFromServicesWebSite
+{
+    public class AssemblyMetadataReferenceFeatureProvider : IApplicationFeatureProvider<MetadataReferenceFeature>
+    {
+        public void PopulateFeature(IEnumerable<ApplicationPart> parts, MetadataReferenceFeature feature)
+        {
+            var currentAssembly = GetType().GetTypeInfo().Assembly;
```
davidfowl: ?

pranavkm: The test site is set up so that the only `AssemblyPart` that was added was for the ClassLibrary that it was referencing. I added this so as to not perturb the arrangement.

src/Microsoft.AspNetCore.Mvc.Razor/Compilation/MetadataReferenceFeatureProvider.cs
```diff
@@ -0,0 +1,80 @@
...
+                    continue;
+                }
+
+                for (var i = 0; i < dependencyContext.CompileLibraries.Count; i++)
+                {
+                    var library = dependencyContext.CompileLibraries[i];
+                    IEnumerable<string> referencePaths;
+                    try
+                    {
+                        referencePaths = library.ResolveReferencePaths();
```
javiercn: Under what scenarios can this fail?

src/Microsoft.AspNetCore.Mvc.Razor/Compilation/MetadataReferenceFeatureProvider.cs
```diff
@@ -0,0 +1,80 @@
...
+            {
+                var dependencyContext = DependencyContext.Load(part.Assembly);
+                if (dependencyContext == null)
+                {
+                    continue;
+                }
+
+                for (var i = 0; i < dependencyContext.CompileLibraries.Count; i++)
+                {
+                    var library = dependencyContext.CompileLibraries[i];
```
javiercn: Why do we use the CompileLibraries here instead of the RuntimeLibraries as in https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultAssemblyPartDiscoveryProvider.cs#L66?

davidfowl: Because we're trying to compile things 😄

src/Microsoft.AspNetCore.Mvc.Razor/Compilation/MetadataReferenceFeatureProvider.cs
```diff
@@ -0,0 +1,80 @@
...
+            {
+                throw new ArgumentNullException(nameof(parts));
+            }
+
+            if (feature == null)
+            {
+                throw new ArgumentNullException(nameof(feature));
+            }
+
+            var libraryPaths = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```
javiercn: If these are paths, I don't see a reason for these to be case insensitive. I believe in other places we are case sensitive.

test/WebSites/ControllersFromServicesWebSite/Startup.cs
```diff
@@ -25,10 +25,15 @@ public void ConfigureServices(IServiceCollection services)
...
-                    typeof(ComponentFromServicesViewComponent),
-                    typeof(InServicesTagHelper))))
+                .ConfigureApplicationPartManager(manager =>
+                {
+                    manager.ApplicationParts.Add(new TypesPart(
+                      typeof(AnotherController),
+                      typeof(ComponentFromServicesViewComponent),
+                      typeof(InServicesTagHelper)));
+
+                    manager.FeatureProviders.Add(new AssemblyMetadataReferenceFeatureProvider());
```
javiercn: Should this be part of https://github.com/aspnet/Mvc/blob/dev/test/Microsoft.AspNetCore.Mvc.FunctionalTests/MvcTestFixture.cs#L79

test/WebSites/ControllersFromServicesWebSite/project.json
```diff
@@ -1,7 +1,6 @@
 {
   "buildOptions": {
-    "emitEntryPoint": true,
-    "preserveCompilationContext": true
```
javiercn: Why are we changing this?

javiercn:
:watch:

src/Microsoft.AspNetCore.Mvc.Razor/Compilation/MetadataReferenceFeatureProvider.cs
pranavkm: We can remove this now.

src/Microsoft.AspNetCore.Mvc.Razor/Compilation/MetadataReferenceFeatureProvider.cs
pranavkm: It would rely on `DependencyContext` consistently casing this (including binaries that are compiled on different machines). Probably safer to leave it as is. Original code is here - https://github.com/aspnet/Mvc/pull/4672/files/549d113b37a6baed6871dee4be7721da0df6301e#diff-96cb3f263eeedd62fed8e95c300bb8b0L253

test/WebSites/ControllersFromServicesWebSite/Startup.cs
pranavkm: This site's application manager is wonky because of this line - https://github.com/aspnet/Mvc/pull/4672/files/549d113b37a6baed6871dee4be7721da0df6301e#diff-8b7f7cd9709495478436a4d926a77252L26. Other apps don't require this.

test/WebSites/ControllersFromServicesWebSite/project.json
pranavkm: The compilation context \ `DependencyContext` for this app isn't used. We use the one from the ClassLibrary. This enforces the expectation.

src/Microsoft.AspNetCore.Mvc.Razor/Compilation/MetadataReferenceFeatureProvider.cs
```diff
@@ -0,0 +1,80 @@
...
+            foreach (var part in parts.OfType<AssemblyPart>())
```
rynowak: You should not be casting to `AssemblyPart` here. Make a new interface that returns a DC and cast to that. 

This way a user can plug in a custom part that returns an arbitrary DC if desired.

src/Microsoft.AspNetCore.Mvc.Razor/Compilation/MetadataReferenceFeatureProvider.cs
```diff
@@ -0,0 +1,80 @@
...
+
+                for (var i = 0; i < dependencyContext.CompileLibraries.Count; i++)
+                {
+                    var library = dependencyContext.CompileLibraries[i];
+                    IEnumerable<string> referencePaths;
+                    try
+                    {
+                        referencePaths = library.ResolveReferencePaths();
+                    }
+                    catch (InvalidOperationException)
```
rynowak: What are some of the issues that might arise here - this gives me the heebie-jeebies, especially without a comment

src/Microsoft.AspNetCore.Mvc.Razor/Compilation/MetadataReferenceFeatureProvider.cs
rynowak: Or you can make the interface return metadata references directly... not sure which is better

src/Microsoft.AspNetCore.Mvc.Razor/Internal/DefaultRoslynCompilationService.cs
```diff
@@ -72,6 +70,13 @@ private List<MetadataReference> ApplicationReferences
             }
         }
 
+        private IList<MetadataReference> GetApplicationReferences()
```
rynowak: Should this be protected? `GetDependencyContext` was

rynowak:
:watch:

src/Microsoft.AspNetCore.Mvc.Razor/Compilation/MetadataReferenceFeatureProvider.cs
pranavkm: https://github.com/dotnet/cli/issues/1694 was the root cause for this bug. `ResolveReferencePaths` used to throw if you had any `type: build` dependencies

src/Microsoft.AspNetCore.Mvc.Razor/Compilation/MetadataReferenceFeatureProvider.cs
pranavkm: We'd have to make `Mvc.Core` depend on Roslyn \ DependencyModel? Is that fine by you?

rynowak: DependencyModel I'm not concerned about because the APM already does.... Roslyn we shouldn't do.

pranavkm:
:up: :date:

src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/IDependencyContextProvider.cs
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.DependencyModel;
+
+namespace Microsoft.AspNetCore.Mvc.ApplicationParts
+{
+    /// <summary>
+    /// Exposes a <see cref="DependencyContext"/> from an <see cref="ApplicationPart"/>.
+    /// </summary>
+    public interface IDependencyContextProvider
```
rynowak: Sorry in advance for being a jerk.

How hard is a DC to construct? Would it be better to expose a limited set of the stuff it provides from this interface instead?

I'm just asking the question. I feel like this is a good starting point at least.

pranavkm: It's not the nicest API to work with. See https://github.com/aspnet/Mvc/pull/4672/files/8651e71758e11828e66b9175e999f938fb6a3c2e#diff-c1746f8559d5a5341fe8a1b87c65d5daL304. We could boil this down to reference paths or push the CLI to make this API easier to construct.

cc @pakrym \ @davidfowl

pakrym: @pranavkm: How would you like API to look like?

pranavkm: Not entirely sure. Maybe a utility method that constructs it given reference paths?

rynowak: @pranavkm that doesn't seem entirely awful

davidfowl: We could do a builder API

pranavkm: Filed https://github.com/dotnet/cli/issues/3157. I'll leave this interface be in that case.

pakrym: Can you do `CompileLibrariesProvider`, you don't need anything else from context.

rynowak: Yeah that might be the best, and that gets to the heart of what you might actually want to do.

javiercn: @pakrym where does CompileLibrariesProvider live? Is that a suggestion?

src/Microsoft.AspNetCore.Mvc.Razor/Compilation/MetadataReferenceFeature.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Microsoft.CodeAnalysis;
+
+namespace Microsoft.AspNetCore.Mvc.Razor.Compilation
+{
+    /// <summary>
+    /// Sepcifies the list of <see cref="MetadataReference"/> used in Razor compilation.
```
dougbu: poTy

test/WebSites/ControllersFromServicesWebSite/Startup.cs
dougbu: What is the minimal code a user would write for the #4497 scenario? User wants an assembly (containing tag helpers) available when compiling a view. I think this PR provides a solution for that requirement -- without an additional compilation callback or exposing the `MetadataReference` collection directly. But I'm hoping our users won't need to write a custom `IApplicationFeatureProvider<MetadataReferenceFeature>` implementation for this.

BTW feel free to poach #4497 if this PR does provide its fix.

pranavkm: In their case, adding a `ApplicationPart`  \ `AssemblyPart` in Startup should suffice:

```
.AddMvc()
.ConfigureApplicationPartManager(manager => manager.ApplicationParts.Add(new AssemblyPart(typeof(MyTagHelper).GetTypeInfo().Assembly);
```

This test was specifically trying to avoid adding the website as an `ApplicationPart` which is why I had one of these.

dougbu: @pranavkm to clarify: The snippet you provided will work in RC2, in `dev` today, or in `dev` after this PR goes in? If it's anything but the last, I'll confirm w/ my #4497 repro then close that issue.

pranavkm: It should start working once this goes in. Prior to this change, we weren't looking for items in the assembly. It does bring up one issue - if the assembly was compiled without `preserveCompilationContext` we'll continue to ignore it. Perhaps we should add just the assembly if DependencyContext is null. We'd have to change the heuristic that tells people to add `preserveCompilationContext` to their application, but it might get views to compile more often.

cc @rynowak \ @davidfowl ?

src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/IDependencyContextProvider.cs
pakrym: @javiercn Yes, I suggest to create `ICompileLibrariesProvider` interface instead of `IDependencyContextProvider` because MVC only uses `CompilationLibraries` from `DependencyContext` and in this case no new API is required to simplify dependency context creation.

javiercn: I believe the interesting part of CompileLibraries is https://github.com/dotnet/cli/blob/rel/1.0.0/src/Microsoft.Extensions.DependencyModel/CompilationLibrary.cs#L39 that we use on https://github.com/aspnet/Mvc/pull/4672/files#diff-ab147a7292f31e46ebc6b39f1df7480bR49

Maybe we can have our own DTO with the interesting parts of compilelibrary and the resolved paths? I think that strikes a good balance. I think something like ApplicationReference { Name, Version and Paths } will be enough

@rynowak @pranavkm Thoughts?

pranavkm: Wouldn't it basically be the same type at that point?

pranavkm:
:up: :date:

src/Microsoft.AspNetCore.Mvc.Razor/Compilation/MetadataReferenceFeatureProvider.cs
javiercn: fair enough I guess, what is our strategy for paths and casing across platforms? I think this is something we should have an answer for that we can apply consistently through the stack. Only windows is case insensitive.

test/WebSites/ControllersFromServicesWebSite/Startup.cs
javiercn: I would add preserve compilation context always, what's the cost, size on the assembly? Adding just the assembly i'm not sure if it'll work. However, for apps that don't load assemblies dinamically I believe that everything will work as long as the top level assembly has preserveCompilationContext = true, as it will contain the transitive closure of all the dependencies.

javiercn:
:shipit: when @rynowak is happy

gdau:
Noted a small typo - GetCompilatonLibraries should be GetCompilationLibraries :)

pranavkm:
Nice catch @gdau

test/WebSites/ControllersFromServicesWebSite/Startup.cs
pranavkm: It's a gesture on the user's end that might be easy to miss

pranavkm:
Ping @rynowak

src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/ICompileLibrariesProvider.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    /// <summary>
+    /// Exposes <see cref="CompilationLibrary"/> instances from an <see cref="ApplicationPart"/>.
+    /// </summary>
+    public interface ICompileLibrariesProvider
+    {
+        /// <summary>
+        /// Gets the sequence of <see cref="CompilationLibrary"/> instances.
+        /// </summary>
+        IReadOnlyList<CompilationLibrary> GetCompilationLibraries();
```
rynowak: `ICompileLibrariesProvider` <-> `GetCompilationLibraries`

The nouns aren't really the same between these. Should it be `ICompileLibrariesProvider`?

pranavkm: Did you mean `ICompilationLibraryProvider`?

rynowak:
:shipit:

src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/ICompileLibrariesProvider.cs
davidfowl: Why are we exposing this as a public API?

pranavkm: The intent is to allow users to add additional compilation libraries

davidfowl: Show me a sample of that? Why wouldn't they just add a metadata reference?

rynowak: > Why wouldn't they just add a metadata reference?

You can ask this question about 100% of the application parts feature, but we still lots of requests for something like this and we still ended up building it. You can solve all of the problems that application model solves by just referencing everything from one project, but users have come up with plenty of cases where they don't want to do that.

As @pranavkm said this is here so you can author a custom part and have control over which assemblies are included as references. Providing interfaces for the various capabilities is part of the design of parts. 

Why would you do this? Well I suppose you could use it to work around the bug we have where a .csproj isn't part of the dependency graph :trollface:

PR #4683, opened 2016-05-19
[Perf] UrlHelper shares a single RouteValueDictionary across several calls to Action method

mnltejaswini:
The goal is to reduce the number of KeyValuePair<String, Object>[] allocations as part of determining the URL for the  action tags. Every call to Action method creates a new RouteValueDictionary as a temporary dictionary to hold the Route values. There is one URLhelper per request and hence we can share a single RouteValueDictionary across multiple calls to Action within a request execution. 

For scenario reference:  https://github.com/aspnet/Mvc/issues/4593 
loadtest -k -n 3000 http://localhost:5000 (home page of the MusicStore app)

Before 
--------
image

After 
------
image

Total savings in 1.4% in total bytes allocated. 
saves 1 RVD, List (and mostly the underlying array) per link to 1 per request.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Text;
 using Microsoft.AspNetCore.Http;
@@ -18,6 +19,8 @@ namespace Microsoft.AspNetCore.Mvc.Routing
 
         // Perf: Share the StringBuilder object across multiple calls of GenerateURL for this UrlHelper
         private StringBuilder _stringBuilder;
+        // Perf: Reuse the RouteValueDictionary across multiple calls of Action for this UrlHelper
+        private readonly RouteValueDictionary _routeValueDictionary;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="UrlHelper"/> class using the specified action context and
@@ -32,6 +35,7 @@ namespace Microsoft.AspNetCore.Mvc.Routing
             }
 
             ActionContext = actionContext;
+            _routeValueDictionary = new RouteValueDictionary();
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
```diff
@@ -51,7 +55,7 @@ public virtual string Action(UrlActionContext actionContext)
                 throw new ArgumentNullException(nameof(actionContext));
             }
 
-            var valuesDictionary = new RouteValueDictionary(actionContext.Values);
+            var valuesDictionary = GetValuesDictionary(actionContext);
```
rynowak: Should we do the same of `RouteUrl`? (see line 109). Both of these are commonly used and are pretty similar

rynowak: We could also elide the copy at line 109
```
var valuesDictionary = routeContext.Values as RouteValueDictionary() ?? new RouteValueDictionary(routeContext.Values);
```

src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
```diff
@@ -205,6 +209,25 @@ public virtual string Link(string routeName, object values)
             });
         }
 
+        private RouteValueDictionary GetValuesDictionary(UrlActionContext actionContext)
+        {
+            var dictionaryValues = actionContext.Values as IDictionary<string, object>;
+            if (dictionaryValues != null)
```
rynowak: It would be nice to also special case when `actionContext.Values` is a `RouteValueDictionary` - that can avoid the boxed enumerator.

what type is `actionContext.Values` usually?

rynowak:
:watch: - this seems like a good simple improvement

src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
mnltejaswini: Usually it is  Dictionary<string, object>

mnltejaswini:
:up::date:

src/Microsoft.AspNetCore.Mvc.Core/Routing/UrlHelper.cs
rynowak: Ok. I think what you have here is fine then.

rynowak:
:shipit:

PR #4687, opened 2016-05-19
Implement manual state machine

rynowak:
This change is a rewrite of the action invoker into a manual state machine. This makes techempower plaintext go from 245krps -> 265krps.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
@@ -33,6 +33,8 @@ namespace Microsoft.AspNetCore.Mvc.Internal
         // https://blogs.msdn.microsoft.com/ericlippert/2008/05/14/mutating-readonly-structs/
         private FilterCursor _cursor;
         private object _controller;
+        private Dictionary<string, object> _arguments;
+        private IActionResult _result;
 
         private AuthorizationFilterContext _authorizationContext;
 
@@ -124,48 +126,53 @@ namespace Microsoft.AspNetCore.Mvc.Internal
 
                     var startTimestamp = _logger.IsEnabled(LogLevel.Information) ? Stopwatch.GetTimestamp() : 0;
 
-                    await InvokeAllAuthorizationFiltersAsync();
+                    // The invoker is implemented using a 'Taskerator' or perhaps an 'Asyncerator' (both terms are correct
+                    // and in common usage). This method is the main 'driver' loop and will call into the `Next` method
+                    // (`await`ing the result) until a terminal state is reached.
+                    //
+                    // The `Next` method walks through the state transitions of the invoker and returns a `Task` when there's
+                    // actual async work that we need to await. As an optimization that Next method won't return a `Task`
+                    // that completes synchronously.
+                    //
+                    // Additionally the `Next` funtion will be called recursively when we're 'inside' a filter invocation.
+                    // Executing 'inside' a filter requires an async method call within a `try`/`catch` for error handling, so
+                    // we have to recurse. Each 'frame' calls into `Next` with a value of `Scope` that communicates what kind
+                    // of 'frame' is executing. This has an effect on the state machine transitions as well as what kinds of
+                    // contexts need to be constructed to communicate the result of execution of the 'frame'.
 
-                    // If Authorization Filters return a result, it's a short circuit because
-                    // authorization failed. We don't execute Result Filters around the result.
-                    Debug.Assert(_authorizationContext != null);
-                    if (_authorizationContext.Result != null)
-                    {
-                        await InvokeResultAsync(_authorizationContext.Result);
-                        return;
-                    }
+                    // When returning, the `Next` method will set `next` to the state to goto on the subsequent invocation.
+                    // This is similar to `Task.ContinueWith`, but since we have a fixed number of states we can avoid
+                    // the overhead of actually using `Task.ContinueWith`.
+                    var next = State.InvokeBegin;
+
+                    // The `scope` tells the `Next` method who the caller is, and what kind of state to initialize to
+                    // communicate a result. The outermost scope is `Scope.Invoker` and doesn't require any type
+                    // of context or result other than throwing.
+                    var scope = Scope.Invoker;
+
+                    // The `state` is used for internal state handling during transitions between states. In practice this
+                    // means storing a filter instance in `state` and then retrieving it in the next state.
+                    var state = (object)null;
+
+                    // `isCompleted` will be set to true when we've reached a terminal state.
+                    var isCompleted = false;
 
                     try
                     {
-                        await InvokeAllResourceFiltersAsync();
+                        while (!isCompleted)
+                        {
+                            await Next(ref next, ref scope, ref state, ref isCompleted);
+                        }
                     }
                     finally
                     {
-                        // Release the instance after all filters have run. We don't need to surround
-                        // Authorizations filters because the instance will be created much later than
-                        // that.
                         if (_controller != null)
                         {
                             _controllerFactory.ReleaseController(_controllerContext, _controller);
                         }
-                    }
 
-                    // We've reached the end of resource filters. If there's an unhandled exception on the context then
-                    // it should be thrown and middleware has a chance to handle it.
-                    Debug.Assert(_resourceExecutedContext != null);
-                    if (_resourceExecutedContext.Exception != null && !_resourceExecutedContext.ExceptionHandled)
-                    {
-                        if (_resourceExecutedContext.ExceptionDispatchInfo == null)
-                        {
-                            throw _resourceExecutedContext.Exception;
-                        }
-                        else
-                        {
-                            _resourceExecutedContext.ExceptionDispatchInfo.Throw();
-                        }
+                        _logger.ExecutedAction(_controllerContext.ActionDescriptor, startTimestamp);
                     }
-
-                    _logger.ExecutedAction(_controllerContext.ActionDescriptor, startTimestamp);
                 }
             }
             finally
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
```diff
@@ -112,363 +113,991 @@ public class ControllerActionInvoker : IActionInvoker
...
+                    {
+                        var filter = (IAsyncResourceFilter)state;
+                        if (_resourceExecutedContext == null)
+                        {
+                            // If we get here then the filter didn't call 'next' indicating a short circuit
+                            Debug.Assert(_resourceExecutingContext.Result != null);
+                            _resourceExecutedContext = new ResourceExecutedContext(_resourceExecutingContext, _filters)
+                            {
+                                Canceled = true,
+                                Result = _resourceExecutingContext.Result, // Can be null
```
dougbu: Comment incorrect. Have just asserted that `Result` is non-`null`.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
```diff
@@ -112,363 +113,991 @@ public class ControllerActionInvoker : IActionInvoker
...
-                    // We don't execute Result Filters around the result.
-                    Debug.Assert(_exceptionContext != null);
-                    if (_exceptionContext.Result != null)
+                case State.ResultInside:
                     {
-                        // This means that exception filters returned a result to 'handle' an error.
-                        // We're not interested in seeing the exception details since it was handled.
-                        await InvokeResultAsync(_exceptionContext.Result);
+                        // The empty result is always flowed back as the 'executed' result
+                        var result = _resultExecutingContext?.Result ?? _result ?? new EmptyResult();
```
javiercn: nit: The local can be inlined

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
```diff
@@ -112,363 +113,991 @@ public class ControllerActionInvoker : IActionInvoker
...
+                        {
+                            if (_resourceExecutingContext == null)
+                            {
+                                _resourceExecutingContext = new ResourceExecutingContext(_controllerContext, _filters);
+                            }
+
+                            state = current.Filter;
+                            goto case State.ResourceSyncBegin;
+                        }
+                        else if (scope == Scope.Resource)
```
javiercn: It would be nice to describe the transition here. For example, "All resource filters are on the stack, so we move on to put exception filters on the stack" but with better wording

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
```diff
@@ -112,363 +113,991 @@ public class ControllerActionInvoker : IActionInvoker
...
+                        {
+                            Debug.Assert(_resourceExecutingContext != null);
+                            goto case State.ResourceInside;
+                        }
+                        else
+                        {
+                            Debug.Assert(scope == Scope.Invoker);
```
javiercn: Similar comment here, something along the lines of "there were no resource filters to invoke"

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
```diff
@@ -112,363 +113,991 @@ public class ControllerActionInvoker : IActionInvoker
...
+
+                        _diagnosticSource.AfterOnResourceExecuted(resourceExecutedContext, filter);
+
+                        goto case State.ResourceEnd;
+                    }
+
+                case State.ResourceShortCircuit:
                     {
-                        await InvokeResourceFilterAsync();
                         Debug.Assert(_resourceExecutedContext != null);
```
javiercn: Debug.Assert(_resourceExecutedContext.Result != null)

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
```diff
@@ -112,363 +113,991 @@ public class ControllerActionInvoker : IActionInvoker
...
+                case State.ResourceEnd:
+                    {
+                        if (scope == Scope.Resource)
+                        {
+                            isCompleted = true;
+                            return TaskCache.CompletedTask;
+                        }
+
+                        Debug.Assert(scope == Scope.Invoker);
+                        Rethrow(_resourceExecutedContext);
```
javiercn: This makes it like it rethrows an exception unconditionally, it will be nice to be a bit more verbose and change the name to reflect that it will only throw if the previous scope ended in an unhandled exception. Something like RethrowIfUnhandledException

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
```diff
@@ -112,363 +113,991 @@ public class ControllerActionInvoker : IActionInvoker
...
+                        {
+                            next = State.ExceptionAsyncResume;
+                            return task;
+                        }
+
+                        goto case State.ExceptionAsyncResume;
+                    }
+
+                case State.ExceptionAsyncResume:
+                    {
```
javiercn: Debug.Assert(_exceptionContext != null);

javiercn: It would be nice to describe the possible scenarios related to exceptionContext here in a comment. exceptionContext == null -> no exception happened. exceptionContext != null && exceptionContext.Exception -> an exception happened but was handled. exceptionContext != null && exceptionContext.Exception == null -> An exception happened but was handled by an inner exception handler.

I'm making all of these kind of comments because with the current implementation of the invoker it's really hard to grasp these details.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
```diff
@@ -112,363 +113,991 @@ public class ControllerActionInvoker : IActionInvoker
...
+                    }
+
+                case State.ExceptionAsyncEnd:
+                    {
+                        var filter = (IAsyncExceptionFilter)state;
+                        var exceptionContext = _exceptionContext;
+
+                        _diagnosticSource.AfterOnExceptionAsync(exceptionContext, filter);
+
+                        if (exceptionContext.Exception == null)
```
javiercn: exceptionContext can be null here if no exceptions happen, can't it? If so, can you add a test to cover this case?

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
```diff
@@ -112,363 +113,991 @@ public class ControllerActionInvoker : IActionInvoker
...
+                            goto case State.ExceptionAsyncEnd;
+                        }
+
+                        goto case State.ExceptionEnd;
+                    }
+
+                case State.ExceptionAsyncEnd:
+                    {
+                        var filter = (IAsyncExceptionFilter)state;
+                        var exceptionContext = _exceptionContext;
```
javiercn: Debug.Assert(_exceptionContext != null)

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
```diff
@@ -112,363 +113,991 @@ public class ControllerActionInvoker : IActionInvoker
...
+                        if (task.Status != TaskStatus.RanToCompletion)
+                        {
+                            next = State.InvokeEnd;
+                            return task;
+                        }
+
+                        goto case State.ResourceEnd;
+                    }
+
+                case State.ExceptionEnd:
```
javiercn: Need to think what the right behavior is when an exception was handled by short-circuiting (setting Exception = null) but no result was set. (Need to take into account old MVC behavior)

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
```diff
@@ -112,363 +113,991 @@ public class ControllerActionInvoker : IActionInvoker
...
+                        {
+                            goto case State.ActionInside;
+                        }
+                    }
+
+                case State.ActionAsyncBegin:
+                    {
+                        Debug.Assert(_actionExecutingContext != null);
+
+                        var filter = (IAsyncActionFilter)state;
```
javiercn: Do we want to also do Debug.Assert(state != null or state = _cursor.CurrentFilter.AsyncFilter) (applicable to other filter stages). The goal is to ensure that we always set the state to the filter in State.ActionNext

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
```diff
@@ -112,363 +113,991 @@ public class ControllerActionInvoker : IActionInvoker
...
+                            };
+                        }
+
+                        _diagnosticSource.AfterOnActionExecution(_actionExecutedContext, filter);
+
+                        goto case State.ActionEnd;
+                    }
+
+                case State.ActionSyncBegin:
+                    {
```
javiercn: Debug.Assert(_actionExecutingContext != null)

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
```diff
@@ -112,363 +113,991 @@ public class ControllerActionInvoker : IActionInvoker
...
+                            goto case State.ResultSyncBegin;
+                        }
+                        else
+                        {
+                            goto case State.ResultInside;
+                        }
+                    }
+
+                case State.ResultAsyncBegin:
+                    {
```
javiercn: Debug.Assert(_resultExecutingContext != null)

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
```diff
@@ -112,363 +113,991 @@ public class ControllerActionInvoker : IActionInvoker
...
+                            {
+                                Canceled = true,
+                            };
+                        }
+
+                        _diagnosticSource.AfterOnResultExecution(_resultExecutedContext, filter);
+                        goto case State.ResultEnd;
+                    }
+
+                case State.ResultSyncBegin:
```
javiercn: Debug.Assert(_resultExecutingContext != null)

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
```diff
@@ -112,363 +113,991 @@ public class ControllerActionInvoker : IActionInvoker
...
-                        };
+                            next = State.ResultEnd;
+                            return task;
+                        }
+
+                        goto case State.ResultEnd;
                     }
-                    else if (_exceptionContext.Exception != null)
+
+                case State.ResultEnd:
```
javiercn: It would be good to add Debug.Assert(scope != <<scope>>) for some of these states, like in this case maybe Debug.Assert(scope != Scope.Exception). Also on the parts where the stack gets "Winded" and "Unwinded" its worth pointing out what happens with the state transitions, basically what the scenario is that we are covering on each of the branches. i.e, for scope == Scope.Resource indicate that the action being invoked contained resource filters and that we are unwinding those now.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
```diff
@@ -112,363 +113,991 @@ public class ControllerActionInvoker : IActionInvoker
...
-                    _exceptionContext.ExceptionDispatchInfo = ExceptionDispatchInfo.Capture(exception);
-                }
+                    ExceptionDispatchInfo = ExceptionDispatchInfo.Capture(exception),
+                };
             }
         }
 
-        private async Task InvokeAllActionFiltersAsync()
+        private async Task InvokeNextActionFilterAsync()
         {
```
javiercn: Debug.Assert(_actionExecutingContext != null);

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
```diff
@@ -483,219 +1112,132 @@ private Task InvokeAllExceptionFiltersAsync()
...
+        private async Task InvokeNextResultFilterAsync()
         {
-            _cursor.Reset();
-
-            _resultExecutingContext = new ResultExecutingContext(_controllerContext, _filters, result, _controller);
-            await InvokeResultFilterAsync();
-
-            Debug.Assert(_resultExecutingContext != null);
-            if (_resultExecutedContext.Exception != null && !_resultExecutedContext.ExceptionHandled)
+            try
```
javiercn: Debug.Assert(_resultExecutingContext != null);

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
```diff
@@ -709,125 +1251,177 @@ private Task InvokeAllExceptionFiltersAsync()
...
+            if (context.ExceptionHandled)
+            {
+                return;
+            }
 
-                    if (_resultExecutingContext.Cancel == true)
-                    {
-                        // Short-circuited by setting Cancel == true
-                        _logger.ResourceFilterShortCircuited(item.Filter);
+            if (context.ExceptionDispatchInfo != null)
```
javiercn: We only check for context.Exception in many of the states, is there a condition that when Exception != null ExceptionDispatchInfo != null too? If not, should we do a different check on the places where we check for Exception? My understanding is that ExceptionDispatchInfo is for async cases where Exception is for Sync cases.

javiercn:
:watch:

javiercn:
Overall looks good. Functionally everything looks alright compared to the old code (functionally equivalent). There are some small things we can improve (more Debug.Assert) and some comments at certain points of the state transitions to clarify what the algorithm does, as its hard to follow from the state transitions

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
rynowak: Fixed

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
rynowak: I decided to fix this and added some comments to describe the possibilities.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
rynowak: sure

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
rynowak: sure

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
rynowak: won't fix, I think it's clear enough.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
rynowak: sure

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
rynowak: When you do this, we treat it as a short-circuit, and flow out the `EmptyResult` as the result.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
rynowak: Added a bunch of these asserts.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
rynowak: Please take a look at what's there and add specific comments. I've added comments in a few places already so if you think something specific is missing point it out.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
rynowak: `context.Exception` cannot be null when `context.ExceptionDispatchInfo` is non-null, so it covers all possibilities.

rynowak:
@javiercn updated

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
```diff
@@ -510,219 +1197,133 @@ private Task InvokeAllExceptionFiltersAsync()
...
-            await InvokeActionFilterAsync();
+            await InvokeNextActionFilterAsync();
 
             Debug.Assert(_actionExecutedContext != null);
             return _actionExecutedContext;
         }
 
-        private async Task InvokeActionFilterAsync()
+        private async Task InvokeActionMethodAsync()
         {
```
javiercn: Does this include the fix for ```public object Action(){ return Content("Hello world")!}``` ?

javiercn:
:shipit:

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionInvoker.cs
rynowak: This is up to date as of when I updated the PR. If this is missing then I'll address it when I rebase.

PR #4688, opened 2016-05-19
[Fixes #4655] Make comparisons in DefaultAssemblyPartDiscoveryProvider case insensitive

### LINKED ISSUE #4655: RC2 Mvc not discovering controller in netcoreapp
I am having an issue where my controller is not being detected by Mvc in my barebones web app. The code for this problem can be found at https://github.com/awec/WeirdRC2Mvc. Also available at https://gist.github.com/awec/c4cb7be9dffdc5a0e29b813743628878 for easier reading. 

I'm using VSCode (VS is not installed on this machine) and dotnet cli. I've installed RC2 and Preview 1 of the sdk (dotnet --version is 1.0.0-preview1-002702). 

Other people who have looked at this issue report that it works in VS, however they also can repro the issue outside of VS (i.e. using the cli) on their machines (including non-Windows machines). 

**Repro steps:**
1. dotnet restore -> no issues
2. dotnet run -> no issues
3. http://localhost:5000 -> 404
4. http://localhost:5000/Home/Index -> 404

**Expected:** exception to be thrown (exception purposefully thrown in controller constructor to demonstrate the controller isn't being initialized). 

The trace logs show:

> dbug: Microsoft.AspNetCore.Routing.RouteBase[1]
>       Request successfully matched the route with name 'default' and template '{controller=Home}/{action=Index}/{id?}'.
> dbug: Microsoft.AspNetCore.Mvc.Internal.MvcRouteHandler[3]
>       No actions matched the current request

Things I have tried:

1. Adding an int id parameter to my action method to make sure the {id?} portion of the default route wasn't causing a problem
2. Renaming the controller from HomeController to Home
3. Moving the controller into a .Controllers namespace
4. Moving the controller class out of the Controllers folder
5. Explicitly adding the route in UseMvc instead of UseMvcWithDefaultRoute.

I definitely have the correct URL (localhost:5000) because I can get responses outside of MVC. 

As far as I can tell my controller meets the requirements for discovery from https://github.com/aspnet/Announcements/issues/165

javiercn:
project.lock.json and .deps.json file use the casing on the users project.json file which causes problems when we discover the default list of application parts. This fix relaxes our criteria for comparing library names so that we don't fail because of the casing used in project.json or a change in the casing of the library name between versions.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultAssemblyPartDiscoveryProvider.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultAssemblyPartDiscoveryProvider.cs
@@ -14,7 +14,7 @@ namespace Microsoft.AspNetCore.Mvc.Internal
     // Discovers assemblies that are part of the MVC application using the DependencyContext.
     public static class DefaultAssemblyPartDiscoveryProvider
     {
-        internal static HashSet<string> ReferenceAssemblies { get; } = new HashSet<string>(StringComparer.Ordinal)
+        internal static HashSet<string> ReferenceAssemblies { get; } = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
         {
             "Microsoft.AspNetCore.Mvc",
             "Microsoft.AspNetCore.Mvc.Abstractions",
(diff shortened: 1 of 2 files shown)


javiercn:
@pranavkm

pranavkm:
:shipit:

src/Microsoft.AspNetCore.Mvc.Core/Internal/DefaultAssemblyPartDiscoveryProvider.cs
```diff
@@ -74,7 +74,7 @@ private class CandidatesResolver
             public CandidatesResolver(IReadOnlyList<RuntimeLibrary> dependencies, ISet<string> referenceAssemblies)
             {
                 _dependencies = dependencies
-                    .ToDictionary(d => d.Name, d => CreateDependency(d, referenceAssemblies));
+                    .ToDictionary(d => d.Name, d => CreateDependency(d, referenceAssemblies), StringComparer.OrdinalIgnoreCase);
```
kichalla: This means the candidate assemblies are case insensitive too right? like `Foo` and `foo` should match?...if yes, can you add a `GetLibrary(..` thingy for this case

javiercn:
Updated @kichalla

kichalla:
:shipit:

PR #4699, opened 2016-05-20
Remove Key from ModelBindingResult

rynowak:
Removing the key from MBR because it's just bloat. It's not really used for anything. This simplifies a lot of code, next change will simplify more.

FYI a lot of the test changes are because now `default(ModelBindingResult) == ModelBindingResult.Failed()`

/cc @kichalla


kichalla:
:shipit:

rynowak:
Merged

PR #4701, opened 2016-05-20
Make ModelBindingContext.Result non-nullable

rynowak:
This change simplifies a bunch of code and fits more in line with the
current design of model binding.

Now, a model binder only has to do anything if it was successful.
'return' is enough to indicate failure.


src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerArgumentBinder.cs
```diff
@@ -189,13 +189,13 @@ public class ControllerArgumentBinder : IControllerArgumentBinder
             await binder.BindModelAsync(modelBindingContext);
 
             var modelBindingResult = modelBindingContext.Result;
-            if (modelBindingResult != null && modelBindingResult.Value.IsModelSet)
+            if (modelBindingResult != null && modelBindingResult.IsModelSet)
```
dougbu: Remove `modelBindingResult != null &&`.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerArgumentBinder.cs
```diff
@@ -85,9 +85,9 @@ public class ControllerArgumentBinder : IControllerArgumentBinder
                 var parameter = parameters[i];
 
                 var result = await BindModelAsync(parameter, controllerContext, valueProvider);
-                if (result != null && result.Value.IsModelSet)
+                if (result != null && result.IsModelSet)
```
dougbu: Remove `result != null &&` a few places in this class.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Internal/ModelBindingHelper.cs
```diff
@@ -450,13 +450,13 @@ public static class ModelBindingHelper
 
             await binder.BindModelAsync(modelBindingContext);
             var modelBindingResult = modelBindingContext.Result;
-            if (modelBindingResult != null && modelBindingResult.Value.IsModelSet)
+            if (modelBindingResult != null && modelBindingResult.IsModelSet)
```
dougbu: Remove `modelBindingResult != null &&`.

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/Binders/BodyModelBinderTests.cs
```diff
@@ -40,13 +40,13 @@ public class BodyModelBinderTests
...
             // Act
-            var binderResult = await binder.BindModelResultAsync(bindingContext);
+            await binder.BindModelAsync(bindingContext);
 
             // Assert
             mockInputFormatter.Verify(v => v.CanRead(It.IsAny<InputFormatterContext>()), Times.Once);
             mockInputFormatter.Verify(v => v.ReadAsync(It.IsAny<InputFormatterContext>()), Times.Once);
-            Assert.NotNull(binderResult);
-            Assert.True(binderResult.IsModelSet);
+            Assert.NotNull(bindingContext.Result);
```
dougbu: `null` isn't possible.

dougbu: Hmm, there's a fair number of these assertions in the new code.

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/Binders/ArrayModelBinderTest.cs
```diff
@@ -151,7 +151,7 @@ private static IModelBinder CreateIntBinder()
                     var model = value.ConvertTo(mbc.ModelType);
                     return ModelBindingResult.Success(model);
                 }
-                return null;
+                return ModelBindingResult.Failed();
```
dougbu: Why not `default(ModelBindingResult)` everywhere?

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/Binders/ArrayModelBinderTest.cs
```diff
@@ -44,23 +44,23 @@ public class ArrayModelBinderTest
             // Arrange
             var binder = new ArrayModelBinder<string>(new SimpleTypeModelBinder(typeof(string)));
 
-            var context = CreateContext();
-            context.IsTopLevelObject = true;
+            var bindingContext = CreateContext();
```
dougbu: Why rename these `context` variables, especially in an already-large PR?

kichalla:
:shipit: after addressing @dougbu's comments

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerArgumentBinder.cs
rynowak: :+1: - will do a find all references for `.Result` in the product and double check that I didn't miss any.

src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerArgumentBinder.cs
rynowak: :+1: - will do a find all references for `.Result` in the product and double check that I didn't miss any.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Internal/ModelBindingHelper.cs
rynowak: :+1: - will do a find all references for `.Result` in the product and double check that I didn't miss any.

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/Binders/BodyModelBinderTests.cs
rynowak: I'll do a global search for these.

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/Binders/ArrayModelBinderTest.cs
rynowak: I think this is a little semantically clearer. In general this code is a holdover from when you had to return an `MBR`. In practice neither product nor user code really needs to call `MBR.Failed()`.

Let's start a side discussion about removing `MBR.Failed`.

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/Binders/ArrayModelBinderTest.cs
rynowak: It was easier to just rename + find and replace. Tests often used different naming within the same file and this makes it a little bit better.

rynowak:
Updating this as an FYI - I removed a lot of extraneous null checks. if there's any more important feedback I'll address in a follow up

PR #4702, opened 2016-05-20
Remove usage of `async void` in unit tests

rynowak:
Yes I know xUnit supports it, but it's non-standard and it shouldn't show
up anywhere in our codebase.


rynowak:
FYI only - will merge

pranavkm:
Actually xunit doesn't support it. It doesn't have any magic to wait for the Task to complete, so these could be failing in the background and we would have never known.

rynowak:
Oh, I thought that was what their custom sync context was for.

pranavkm:
Huh, maybe xunit 2.0 does. The 1.x runner did not.

Issue #4711, opened 2016-05-23
Update Json.NET


JamesNK:
https://www.nuget.org/packages/Newtonsoft.Json/8.0.4-beta1

Targets netstandard1.0 - no more imports!

henkmollema:
Is there a reason it's still a beta version?

PinpointTownes:
FYI, JSON.NET is also used by a few other ASP.NET Core/.NET Core projects (e.g the OAuth2/OIDC middleware or IdentityModel). That would be great to update it everywhere at the same time to avoid package resolution warnings during `dotnet restore` :smile:

Eilon:
Yes, we will update it across as much as possible.

See https://github.com/dotnet/cli/issues/3198 for some info regarding some issues we have to resolve.

javiercn:
Done. Here is the command I used to verify that we updated all our dependencies across repos. Clone universe, run build pull and then ```Get-ChildItem -Recurse project.json | select-string "Newtonsoft"```

dougbu:
@javiercn `build pull` doesn't pull a number of repos. One of them that _does_ use Json.NET is Coherence.

Suggest using `build pull-all` instead. Probably also need to double-check at least the Performance repo.

javiercn:
I didn't know about pull-all. At least everything we need to build the product is in pull. I'll try pull all later and update any missing dependency, thanks for the tip!

javiercn:
@dougbu Did a pull all, the only one not in 9.0.1 was Coherence\src\CoherenceBuild\project.json:7:    "Newtonsoft.Json": "8.0.3", so it's fine

dougbu:
@javiercn suggest a PR to make Coherence consistent. It's low-priority but we don't have a reason to stick w/ 8.0.3.

javiercn:
@dougbu it's already ipsa

PR #4712, opened 2016-05-23
Update Json.NET version

henkmollema:
Json.NET 8.0.4-beta1 targets .NETStandard 1.0, no more imports required.

#4711

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/project.json
+++ b/src/Microsoft.AspNetCore.Mvc.ViewFeatures/project.json
@@ -47,7 +47,7 @@
       "type": "build"
     },
     "Microsoft.Extensions.WebEncoders": "1.0.0-*",
-    "Newtonsoft.Json": "8.0.3",
+    "Newtonsoft.Json": "8.0.4-beta1",
     "System.Buffers": "4.0.0-*"
   },
   "frameworks": {
@@ -55,10 +55,7 @@
     "netstandard1.5": {
       "dependencies": {
         "System.Runtime.Serialization.Primitives": "4.1.1-*"
-      },
-      "imports": [
-        "portable-net451+win8"
-      ]
+      }
     }
   }
 }


src/Microsoft.AspNetCore.Mvc.ViewFeatures/project.json
```diff
@@ -47,18 +47,15 @@
       "type": "build"
     },
     "Microsoft.Extensions.WebEncoders": "1.0.0-*",
-    "Newtonsoft.Json": "8.0.3",
+    "Newtonsoft.Json": "8.0.4-beta1",
```
pranavkm: I think nuget usually doesn't allow a non-prerelease package to depend on a release version. Would this be problematic when Mvc RTMs? cc @emgarten

emgarten: @pranavkm correct, NuGet.exe will fail on this. dotnet pack allows this I believe, but that will soon be replaced by nuget.exe's pack. I suggest waiting for a stable version of newtonsoft.json before updating.

pranavkm: cc @JamesNK

JamesNK:
I'll release a stable version that supports netstandard when .NET Core is stable so you'll be waiting a while.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/project.json
JamesNK: I'll release a stable version that supports netstandard when .NET Core is stable so you'll be waiting a while.

Eilon:
@henkmollema thanks for the PR. Unfortunately, updating to Json.NET is much more complex than this: we have to update dozens of repos, plus coordinate releases with @JamesNK .

henkmollema:
Yeah I understand. No worries :smile:

Eilon:
@henkmollema thanks 😄

PR #4720, opened 2016-05-23
Update the namespace for Razor generated code to AspNetCore

### LINKED ISSUE #4476: Update the namespace for Razor generated code to AspNetCore
Our generated Razor classes are in the `Asp` namespace. Should it be `AspNetCore`?

pranavkm:
Fixes #4476


pranavkm:
cc @rynowak

pranavkm:
cc @NTaylorMullen

NTaylorMullen:
:shipit:

Eilon:
Ok I'm late to the game here, but I'm not sure if this is a good change... What I'm wondering is if this name is likely to cause any conflicts with the `Microsoft.AspNetCore.*` namespaces in a user's CSHTML code or something. Is that likely to be a problem in any conceivable scenario? Or is this totally safe and won't conflict ever?

NTaylorMullen:
@Eilon They'd need to have some prettttty wonky class names to have conflicts. For example in [this](https://github.com/aspnet/Mvc/blob/69e029fc7903b011100279144507e5a05d7dd4d1/test/Microsoft.AspNetCore.Mvc.Razor.Host.Test/TestFiles/Output/Runtime/ModelExpressionTagHelper.cs#L12) output it generates the class name `TestFiles_Input_ModelExpressionTagHelper_cshtml`. Essentially the names correspond with the folder/filename of the view.

In the majority case users generated Razor class will end up looking something like `Views_XYZ_cshtml`

Eilon:
@NTaylorMullen sorry I wasn't very clear. I'm not concerned about the generated type names of the pages conflicting with anything. Rather, say within the page I have some code in `@{ ... }` or `@functions { ... }`, could code in there get some weird error/conflict if I was trying to reference something in a `Microsoft.AspNetCore` namespace? Or perhaps some code that I could write in there would easily cause a conflict elsewhere? Or is it all such that this could never conflict?

NTaylorMullen:
@Eilon Ah, I see. I can't envision that being a a problem since it's just `AspNetCore` and not `Microsoft.AspNetCore`. Could you give me a strawman you had in mind to make sure we be on the same page :smile: ?

Eilon:
I'm thinking of namespace resolution issues as mentioned here: http://stackoverflow.com/q/25061988/31668 (read the various answers)

Though, the more I think about it, I'm less sure that there's any chance for error. But I'd like a smart brain to evaluate my thoughts and let me know if I missed something 😄

NTaylorMullen:
Haha, fair enough. Ya I don't see any worrisome problems in transitioning to `AspNetCore` as far as type/namespace conflict goes. :+1:

Eilon:
Yeah basically *if* there was such a thing as partial namespace resolution in C# (which there *isn't*) then I think there would be an issue here. But there isn't. So there isn't. So good night 😄 (Or morning!)

pranavkm:
One possible scenario where this is broken is if you had a namespace in your app named `Microsoft` and were using it to reference types under the `AspNetCore` namespace e.g.
```
@using Microsoft
@inject AspNetCore.Mvc.Localization.IStringLocalizer<StringResources> Localizer
```

It's a bit far fetched though

Eilon:
@pranavkm how would that theoretically break?

pranavkm:
Nevermind. I was expecting it to partial namespace resolution which it doesn't.

Eilon:
Exactly 😄

PR #4721, opened 2016-05-23
Fix resource generation to work with dotnet test

pranavkm:


pranavkm:
cc @dougbu

test/Microsoft.AspNetCore.Mvc.TestCommon/ResourceFile.cs
```diff
@@ -190,10 +188,7 @@ public static void UpdateFile(Assembly assembly, string resourceName, string pre
 
             if (!string.Equals(normalizedPreviousContent, normalizedContent, StringComparison.Ordinal))
             {
-                // The DNX runtime compiles every file under the resources folder as a resource available at runtime
-                // with the same name as the file name. Need to update this file on disc.
```
dougbu: We still need _some_ comment explaining what's going on here.

dougbu:
:shipit:

test/Microsoft.AspNetCore.Mvc.TestCommon/ResourceFile.cs
pranavkm: > // The build system compiles every file under the resources folder as a resource available at runtime
> // with the same name as the file name. Need to update this file on disc.

Sound good?

PR #4729, opened 2016-05-24
Fix broken query string binding test

rynowak:
This test was relying on the parsing behavior of QueryString, that it
would omit the key and value when the value is empty. The expectations
weren't consistent with other tests for 'empty' values.

### DIFF
--- a/test/Microsoft.AspNetCore.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
+++ b/test/Microsoft.AspNetCore.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
@@ -10,6 +10,7 @@ using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Internal;
 using Microsoft.AspNetCore.Mvc.Abstractions;
 using Microsoft.AspNetCore.Mvc.ModelBinding;
+using Microsoft.AspNetCore.Testing;
 using Microsoft.Extensions.Primitives;
 using Xunit;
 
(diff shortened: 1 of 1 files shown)


rynowak:
FYI only will merge to unblock CI

test/Microsoft.AspNetCore.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
```diff
@@ -1887,10 +1888,11 @@ private class Order14
             Assert.Null(error.Exception);
         }
 
-        // This covers the case where a key is present, but has no value. The type converter
-        // will report an error.
+        // This covers the case where a key is present, but has no value. The model binder will
+        // report and error because it's a value type (non-nullable).
```
dougbu: "and" -> "an"

test/Microsoft.AspNetCore.Mvc.IntegrationTests/MutableObjectModelBinderIntegrationTest.cs
```diff
@@ -1887,10 +1888,11 @@ private class Order14
...
         [Fact]
-        public async Task MutableObjectModelBinder_BindsPOCO_TypeConvertedPropertyWithNoValue_NoError()
+        [ReplaceCulture]
+        public async Task MutableObjectModelBinder_BindsPOCO_TypeConvertedPropertyWithEmptyValue_Error()
```
dougbu: Why does this name still mention type conversion? Doesn't happen in this path through the `SimpleTypeModelBinder`.

dougbu:
:shipit: but added suggestions for a quick-n-dirty follow-up.

PR #4731, opened 2016-05-24
Inject the IRazorViewEngine directory instead of using GetRequiredService

pranavkm:


pranavkm:
cc @rynowak

src/Microsoft.AspNetCore.Mvc.Razor/Internal/MvcRazorMvcViewOptionsSetup.cs
```diff
@@ -15,32 +15,31 @@ public class MvcRazorMvcViewOptionsSetup : ConfigureOptions<MvcViewOptions>
         /// <summary>
         /// Initializes a new instance of <see cref="MvcRazorMvcViewOptionsSetup"/>.
         /// </summary>
-        /// <param name="serviceProvider">The application's <see cref="IServiceProvider"/>.</param>
-        public MvcRazorMvcViewOptionsSetup(IServiceProvider serviceProvider)
-            : base(options => ConfigureMvc(serviceProvider, options))
+        /// <param name="razorViewEngine">The <see cref="IRazorViewEngine"/>.</param>
+        public MvcRazorMvcViewOptionsSetup(IRazorViewEngine razorViewEngine)
```
pranavkm: There's two other places where we use `IServiceProvider`, but there's slightly more involved since they use `GetService` instead of `GetRequiredService` to get an optionally injected feature. We'd have to use constructor overloading to make those work correctly.

rynowak: Would our lives here be simpler if we got rid of the `ConfigureOptions` base class?

rynowak:
:shipit:

src/Microsoft.AspNetCore.Mvc.Razor/Internal/MvcRazorMvcViewOptionsSetup.cs
pranavkm: https://github.com/aspnet/Mvc/issues/4736

PR #4732, opened 2016-05-24
[Perf] Avoid ViewBuffers for writing bound TagHelper attribute values.

mnltejaswini:
This PR has changes that enables fixing  https://github.com/aspnet/Razor/issues/717 in Razor. Currently, writing TagHelper attribute uses the same code as the TagHelper content writing. This uses StartTagHelperWritingScope and EndTagHelperWritingScope which is expensive for writing attributes as these methods uses ViewBuffer, ViewBufferTextWriter. We an use simple StringWriter for writing attributes. Also, as attribute writing is not nested, we can use a shared writer within the RazorPage. The RazorPage has two new methods special cased for writing TagHelper attributes using the shared StringWriter.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor/RazorPage.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor/RazorPage.cs
@@ -40,6 +40,7 @@ namespace Microsoft.AspNetCore.Mvc.Razor
         private IViewBufferScope _bufferScope;
         private bool _ignoreBody;
         private HashSet<string> _ignoredSections;
+        private TextWriter _pageWriter;
 
         public RazorPage()
         {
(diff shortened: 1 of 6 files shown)


mnltejaswini:
@rynowak , @NTaylorMullen

src/Microsoft.AspNetCore.Mvc.Razor/RazorPage.cs
```diff
@@ -232,6 +233,55 @@ public TagHelperContent EndTagHelperWritingScope()
...
+        /// and <see cref="EndWriteTagHelperAttribute"/> method calls.
+        /// </remarks>
+        public string EndWriteTagHelperAttribute()
+        {
+            if (_pageWriter == null)
+            {
+                throw new InvalidOperationException(Resources.RazorPage_ThereIsNoActiveWritingScopeToEnd);
+            }
+
+            var content = _valueBuffer.ToString();
```
NTaylorMullen: Is `ToString`ing here effectively the same as calling `_valueBuffer.GetStringBuilder().ToString()`? Just ensuring this doesn't build up extra assets required to `ToString` the writer.

NTaylorMullen:
Looks good. Waiting on the Razor PR to sign off.

src/Microsoft.AspNetCore.Mvc.Razor/RazorPage.cs
mnltejaswini: Yes. StringWriter.ToString returns the underlying Stringbuilder.ToString()

src/Microsoft.AspNetCore.Mvc.Razor/RazorPage.cs
```diff
@@ -232,6 +233,55 @@ public TagHelperContent EndTagHelperWritingScope()
         }
 
         /// <summary>
+        /// Starts a new scope for writing <see cref="ITagHelper"/> attribute values.
+        /// </summary>
+        /// <remarks>
+        /// All writes to the <see cref="Output"/> or <see cref="ViewContext.Writer"/> after calling this method will
+        /// be buffered until <see cref="EndWriteTagHelperAttribute"/> is called.
+        /// The content will be buffered using a shared <see cref="StringWriter"/> within this <see cref="RazorPage"/> 
+        /// This method assumes that there will be no nesting of <see cref="BeginWriteTagHelperAttribute"/>
```
dougbu: Rather than assuming anything, suggest asserting that `_pageWriter == null` here and resetting `_pageWriter` in `EndWriteTagHelperAttribute()`.

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorPageTest.cs
```diff
@@ -225,6 +225,47 @@ public static TheoryData<HtmlEncoder> HtmlEncoderData
...
 
+        [Fact]
+        public async Task EndWriteTagHelperAttribute_RestoresPageWriter()
+        {
+            // Arrange
+            var page = CreatePage(v =>
+            {
+                v.BeginWriteTagHelperAttribute();
+                v.Write("Hello World!");
+                var returnValue = v.EndWriteTagHelperAttribute();
```
dougbu: Remove unused `returnValue`.

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorPageTest.cs
```diff
@@ -225,6 +225,47 @@ public static TheoryData<HtmlEncoder> HtmlEncoderData
...
+            var page = CreatePage(v =>
+            {
+                v.HtmlEncoder = new HtmlTestEncoder();
+                v.BeginWriteTagHelperAttribute();
+                v.Write("Hello World!");
+                var returnValue = v.EndWriteTagHelperAttribute();
+
+                // Assert
+                var content = Assert.IsType<string>(returnValue);
+                Assert.Equal("HtmlEncode[[Hello World!]]", content);
```
dougbu: Assertions in lambdas may lead to unintelligible test failures. Suggest checking quickly e.g. mess up the expected string, run test, and confirm the failure is useful.

rynowak: From what I've seen recently this isn't really the case. The test explorer will go right to the line.

NTaylorMullen:
Q: What's the perf results for this + the Razor change?

mnltejaswini:
Scenario: https://github.com/aspnet/Mvc/issues/4593 

Before Optimization
------------------------
image


After Optimization
----------------------
image


The total saving is 4.5% of total bytes allocated. 
It saves one allocation of ViewBuffer, ViewBufferTextWriter, List<ViewBufferPage> and underlying ViewBufferPage[] per attribute in a page to 0. Although it might increase char[] allocations because of StringBuilder  depending on the code content we add within the scope.

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorPageTest.cs
mnltejaswini: The failure is useful as pasted below

```
xUnit.net Runner (64-bit win10-x64)
  Discovering: Microsoft.AspNetCore.Mvc.Razor.Test
  Discovered:  Microsoft.AspNetCore.Mvc.Razor.Test
  Starting:    Microsoft.AspNetCore.Mvc.Razor.Test
    Microsoft.AspNetCore.Mvc.Razor.RazorPageTest.EndWriteTagHelperAttribute_ReturnsAppropriateContent [FAIL]
      Assert.Equal() Failure
                          ↓ (pos 10)
      Expected: HtmlEncode1[[Hello World!]]
      Actual:   HtmlEncode[[Hello World!]]
                          ↑ (pos 10)
```

NTaylorMullen:
:shipit:

test/Microsoft.AspNetCore.Mvc.Razor.Test/RazorPageTest.cs
dougbu: Excellent.

PR #4733, opened 2016-05-24
Remove IDisposable contract from IMvcRazorHost

### LINKED ISSUE #4709: Design problem: MvcRazorHost.Dispose
Hi,

MvcRazorHost class injects IChunkTreeCache and assigns it to _chunkTreeCache field (https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Razor.Host/MvcRazorHost.cs#L70). That's a good DI usage, no problem.
But then in MvcRazorHost.Dispose, it disposes the _chunkTreeCache (https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Razor.Host/MvcRazorHost.cs#L348).

Isn't that a design problem? I think a class should not dispose an object if it's injected (not created by this class).

With Dependency injection, IChunkTreeCache's lifetime may not be same as MvcRazorHost. Beside that, DI framework will already dispose IChunkTreeCache when needed. This design can lead to double Dispose of _chunkTreeCache. Maybe this works since MvcRazorHost is somehow singleton, but it's conceptually wrong.

So, a class should only dispose an object if this object is directly created by that class, not injected.

I come across to this situation while debugging my Castle Windsor DI adapter (https://github.com/volosoft/castle-windsor-ms-adapter/tree/master/src).

What do you think?

pranavkm:
Fixes #4709

### DIFF
--- a/test/Microsoft.AspNetCore.Mvc.Razor.Host.Test/Directives/ChunkInheritanceUtilityTest.cs
+++ b/test/Microsoft.AspNetCore.Mvc.Razor.Host.Test/Directives/ChunkInheritanceUtilityTest.cs
@@ -32,83 +32,81 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Directives
                 new UsingChunk { Namespace = "AppNamespace.Model" },
             };
             var cache = new DefaultChunkTreeCache(fileProvider);
-            using (var host = new MvcRazorHost(cache, new TagHelperDescriptorResolver(designTime: false)))
-            {
-                var utility = new ChunkInheritanceUtility(host, cache, defaultChunks);
+            var host = new MvcRazorHost(cache, new TagHelperDescriptorResolver(designTime: false));
+            var utility = new ChunkInheritanceUtility(host, cache, defaultChunks);
 
-                // Act
-                var chunkTreeResults = utility.GetInheritedChunkTreeResults(
-                    PlatformNormalizer.NormalizePath(@"Views\home\Index.cshtml"));
+            // Act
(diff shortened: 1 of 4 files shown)


pranavkm:
cc @NTaylorMullen \ @hikalkan

test/Microsoft.AspNetCore.Mvc.Razor.Host.Test/Directives/ChunkInheritanceUtilityTest.cs
```diff
@@ -32,83 +32,82 @@ @inherits MyBaseType
...
-                var utility = new ChunkInheritanceUtility(host, cache, defaultChunks);
+            var host = new MvcRazorHost(cache, new TagHelperDescriptorResolver(designTime: false));
 
-                // Act
-                var chunkTreeResults = utility.GetInheritedChunkTreeResults(
-                    PlatformNormalizer.NormalizePath(@"Views\home\Index.cshtml"));
+            var utility = new ChunkInheritanceUtility(host, cache, defaultChunks);
 
-                // Assert
-                Assert.Collection(chunkTreeResults,
```
pranavkm: All of this is just untabbed. The diff is a bit messed up. The split view helps a bit

NTaylorMullen:
:shipit: 

Might be worth an announcement.

Eilon:
Doubt we need an announcement. I'm sure 99.999% of users don't know about any of this. It's plumbing.

pranavkm:
cc @ToddGrun for tooling

PR #4735, opened 2016-05-24
Remove IControllerTypeProvider

pranavkm:
cc @javiercn


pranavkm:
Looks like we missed removing this when switching to ApplicationParts

javiercn:
:shipit:

PR #4739, opened 2016-05-25
Modify ICompilationLibrariesProvider to return a sequence of reference paths

### LINKED ISSUE #4738: Modify ICompilationLibrariesProvider to return a sequence of reference path strings
Based on offline discussions with @davidfowl  and @rynowak , the plan is to modify `ICompilationLibrariesProvider` to not expose `CompilationLibrary` instances instead returning a list of paths to reference assemblies. 

An additional follow up was to return the assembly path in the event the assembly was compiled without `preserveCompilationContext`. This will unblock csproj based class library scenarios which do not have deps record.

pranavkm:
Fixes #4738

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/AssemblyPart.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/AssemblyPart.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Reflection;
 using Microsoft.Extensions.DependencyModel;
 
@@ -11,7 +12,7 @@ namespace Microsoft.AspNetCore.Mvc.ApplicationParts
     /// <summary>
     /// An <see cref="ApplicationPart"/> backed by an <see cref="Assembly"/>.
     /// </summary>
-    public class AssemblyPart : ApplicationPart, IApplicationPartTypeProvider, ICompilationLibrariesProvider
+    public class AssemblyPart : ApplicationPart, IApplicationPartTypeProvider, ICompilationReferencesProvider
     {
         /// <summary>
         /// Initalizes a new <see cref="AssemblyPart"/> instance.
@@ -41,15 +42,19 @@ namespace Microsoft.AspNetCore.Mvc.ApplicationParts
         public IEnumerable<TypeInfo> Types => Assembly.DefinedTypes;
 
         /// <inheritdoc />
-        public IReadOnlyList<CompilationLibrary> GetCompilationLibraries()
+        public IEnumerable<string> GetReferencePaths()
         {
             var dependencyContext = DependencyContext.Load(Assembly);
             if (dependencyContext != null)
             {
-                return dependencyContext.CompileLibraries;
+                return dependencyContext.CompileLibraries.SelectMany(library => library.ResolveReferencePaths());
             }
 
-            return new CompilationLibrary[0];
+            // If an application has been compiled without preserveCompilationContext, return the path to the assembly
+            // as a reference. For runtime compilation, this will allow the compilation to succeed as long as it least
+            // one application part has been compiled with preserveCompilationContext and contains a super set of types
+            // required for the compilation to succeed.
+            return new[] { Assembly.Location };
         }
     }
 }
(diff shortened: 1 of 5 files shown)


pranavkm:
cc @davidfowl \ @rynowak

javiercn:
:shipit:

src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/AssemblyPart.cs
```diff
@@ -41,15 +42,15 @@ public AssemblyPart(Assembly assembly)
...
         {
             var dependencyContext = DependencyContext.Load(Assembly);
             if (dependencyContext != null)
             {
-                return dependencyContext.CompileLibraries;
+                return dependencyContext.CompileLibraries.SelectMany(library => library.ResolveReferencePaths());
             }
 
-            return new CompilationLibrary[0];
+            return new[] { Assembly.Location };
```
rynowak: This is probably worth a command to explain why and what.

rynowak:
:shipit:

src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/AssemblyPart.cs
javiercn: Did you mean a comment?

rynowak: I command you to add a comment

pranavkm: > sudo add comment

src/Microsoft.AspNetCore.Mvc.Core/ApplicationParts/AssemblyPart.cs
```diff
@@ -41,15 +41,37 @@ public AssemblyPart(Assembly assembly)
...
+                {
+                    for (var i = 0; i < compileLibraries.Count; i++)
+                    {
+                        var compileLibrary = compileLibraries[i];
+                        referencePaths.AddRange(compileLibrary.ResolveReferencePaths());
+                    }
+
+                    return referencePaths;
+                }
+                catch
```
pranavkm: cc @dougbu

pranavkm: @rynowak \ @javiercn we're adding an additional guard against `ResolveReferencePaths` failing. This happened in @dougbu's scenario for https://github.com/aspnet/Mvc/issues/4497 where the assembly was compiled with `preserveCompilationContext` but is dynamically loaded at runtime.

rynowak: This is problematic. Having a try/catch here means that a part is *all-or-nothing* with it's dependencies. 

What's the issue that causes `ResolveReferencePaths` to throw? Would a user rather see this exception?

dougbu: Still seem to hit issues even w/o `preserveCompilationContext: true`. See a FNF exception with `Message` "Could not load file or assembly 'ClassLibrary, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies. The system cannot find the file specified." Oddly the exception is `thrown` when `DefaultRoslynCompilationService` asks the successfully-generated for its exported types. Seems like the compilation dependencies (and their full paths) aren't preserved for runtime.

pranavkm: @rynowak I don't think there's much a user would be to able to do with the exception. I'll file a follow up on the CLI to figure out guidance on dealing with this.

rynowak: >  I don't think there's much a user would be to able to do with the exception.

But doesn't this indicate a deployment/publishing problem? We could possibly provide a better error

pranavkm: Spoke to @pakrym - he says as long as the dependencies are in the application base path (bin dir), `ResolveReferencePaths` should behave correctly.

dougbu: That only solves half the problem. What about packages referenced from the loaded assembly but not from the site?

pranavkm: They're going to address those two. @davidfowl and @pakrym were discussing a plan for it. @pakrym could you update #4497 with the CLI issue tracking your change?

Issue #4292, opened 2016-03-15
TypeActivate MvcRouteHandler


rynowak:
This class does a bunch of lazy service lookups. We should just activate it at startup to simplify things

ajaybhargavb:
3d0f436a06957501b3100e62af916bc857cbd39d

PR #4740, opened 2016-05-25
TypeActivate MvcRouteHandler

ajaybhargavb:
#4292 

@rynowak

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/Builder/MvcApplicationBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Builder/MvcApplicationBuilderExtensions.cs
@@ -89,7 +89,7 @@ namespace Microsoft.AspNetCore.Builder
 
             var routes = new RouteBuilder(app)
             {
-                DefaultHandler = new MvcRouteHandler(),
+                DefaultHandler = app.ApplicationServices.GetRequiredService<MvcRouteHandler>(),
             };
 
             configureRoutes(routes);
--- a/src/Microsoft.AspNetCore.Mvc.Core/Internal/MvcRouteHandler.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Internal/MvcRouteHandler.cs
@@ -10,20 +10,43 @@ using Microsoft.AspNetCore.Mvc.Core;
 using Microsoft.AspNetCore.Mvc.Infrastructure;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.AspNetCore.Routing.Tree;
-using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.AspNetCore.Mvc.Internal
 {
     public class MvcRouteHandler : IRouter
     {
-        private bool _servicesRetrieved;
-
         private IActionContextAccessor _actionContextAccessor;
         private IActionInvokerFactory _actionInvokerFactory;
         private IActionSelector _actionSelector;
         private ILogger _logger;
         private DiagnosticSource _diagnosticSource;
+
+        public MvcRouteHandler(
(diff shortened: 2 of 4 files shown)


src/Microsoft.AspNetCore.Mvc.Core/Internal/MvcRouteHandler.cs
```diff
@@ -10,21 +10,44 @@
...
+            : this(actionInvokerFactory, actionSelector, diagnosticSource, loggerFactory, actionContextAccessor: null)
+        {
+        }
+
+        public MvcRouteHandler(
+            IActionInvokerFactory actionInvokerFactory,
+            IActionSelector actionSelector,
+            DiagnosticSource diagnosticSource,
+            ILoggerFactory loggerFactory,
+            IActionContextAccessor actionContextAccessor)
```
rynowak: Let's talk about this more.

src/Microsoft.AspNetCore.Mvc.Core/Builder/MvcApplicationBuilderExtensions.cs
```diff
@@ -89,7 +89,7 @@ public static IApplicationBuilder UseMvcWithDefaultRoute(this IApplicationBuilde
 
             var routes = new RouteBuilder(app)
             {
-                DefaultHandler = new MvcRouteHandler(),
+                DefaultHandler = app.ApplicationServices.GetRequiredService<IRouter>(),
```
rynowak: Register this directly as `MvcRouteHandler` - it's not the be-all-end-all of routing

src/Microsoft.AspNetCore.Mvc.Core/Internal/MvcRouteHandler.cs
davidfowl: yes please. don't bring it back 😄

rynowak: It's not registered by default. I was more concerned with the correct DI behavior. Is this the correct way to take an 'optional' service? or do we need to take the `IServiceProvider` instead and try to resolve it in the ctor

ajaybhargavb: I spoke to @pranavkm. He said we can expect DI to behave correctly as it is in the spec.

ajaybhargavb:
:up: :date:

src/Microsoft.AspNetCore.Mvc.Core/Internal/MvcRouteHandler.cs
rynowak: Cool :+1:

test/Microsoft.AspNetCore.Mvc.Core.Test/Infrastructure/MvcRouteHandlerTests.cs
```diff
@@ -192,47 +204,16 @@ public class MvcRouteHandlerTests
...
-                optionsAccessor = new TestOptionsManager<MvcOptions>();
-            }
-
-            var diagnosticSource = new DiagnosticListener("Microsoft.AspNetCore");
-            if (diagnosticListener != null)
-            {
-                diagnosticSource.SubscribeWithAdapter(diagnosticListener);
-            }
+            return new MvcRouteHandler(
+                invokerFactory, actionSelector, diagnosticSource, loggerFactory, actionContextAccessor);
```
rynowak: gross break this up

rynowak:
:shipit: after addressing https://github.com/aspnet/Mvc/pull/4740/files/291d5d44857e871a9ba569cd9d95a4b9787b3519#r64504142

ajaybhargavb:
I addressed that in the second commit.

rynowak:
:shipit: again

Issue #4666, opened 2016-05-18
HttpPost with entity as parameter stuck in an infinite loop? (rc2-final)


as14220808:
Consider a call to a simple post action like this

        [HttpPost]
        public async Task<IActionResult> Create(Organization data)
        {
            return NotFound();
        }

The call is unsuccessful and cause memory consumption go up wildly (1 GB every 30 second) and it never stop and consume all my memory.

It looks like the call is stuck in an infinite loop of some sort.

After some experiment, I found out that only classes that I use in DbContext cause this infinite loop. Swapping "Organization" (which is a class I used in DbContext) with other class like the one I used as viewmodel dose not cause this.

I'm not sure what I did wrong because in rc1 everything works just fine.

![untitled](https://cloud.githubusercontent.com/assets/16411778/15355627/1b3cda46-1d1e-11e6-836e-b0e13c7f09d7.png)

xrkolovos:
I am having the same issue. I have a complex ef model, and i get as parameter in the post method one object of this model. It's like the framework trying to discover all the dependancies with this model and iterates through the whole model.

xrkolovos:
i am testing this bug for like 4 hours. After adding more complexity(i mean more objects with more connections between them) the time that takes to execute the method rises. At a given number of objects (with connections between them), the app breaks and it nevers hit the method inside the mvc controller. After that also we get the memory leak.

xrkolovos:
after installing the rc2 stuff, can we continue developing in rc1?

fglebel:
I am also stuck in a loop. Unable to advance. Was working fine in RC1.
![create](https://cloud.githubusercontent.com/assets/16037514/15362570/a1ac223e-1ce3-11e6-8826-22a9b00aa7e2.jpg)

Bartmax:
Just in case, a word from the MS team will be great.

rynowak:
> Just in case, a word from the MS team will be great.

Could one of you experiencing the issue grab a call-stack of where MVC is spending most of its time?

xrkolovos:
can you provide some info how to get one? Because i don't know how. After i am skipping the last line of the constuctor the stact trace window is empty.

xrkolovos:
you only need an complex model with about 20-25 classes and relations between them. if you need one, i can created for you.

fglebel:
Since I cannot hit the breakpoint, call stack is empty. I am trying to reproduce from scratch.

xrkolovos:
[...]

PR #4741, opened 2016-05-25
Fix for #4666 - reduce modelbinders created

rynowak:
This change to ModelBinderFactory makes the caching much more aggressive,
by caching all non-root binders. There's some trickiness here around
making sure we have the right behavior when all providers return null. See
the tests and comments.

I also kept the change I made for a temporary workaround to use a
dictionary rather than a "stack" for cycle breaking.  This seems like an
overall improvement in clarity.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Core/Internal/PlaceholderBinder.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/Internal/PlaceholderBinder.cs
@@ -2,8 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
 
-namespace Microsoft.AspNetCore.Mvc.ModelBinding.Internal
+namespace Microsoft.AspNetCore.Mvc.Internal
 {
     // Used as a placeholder to break cycles while building a tree of model binders in ModelBinderFactory.
     //
--- a/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
@@ -9,7 +9,7 @@ using System.Linq;
 using System.Runtime.CompilerServices;
 using Microsoft.AspNetCore.Mvc.Core;
 using Microsoft.AspNetCore.Mvc.Internal;
-using Microsoft.AspNetCore.Mvc.ModelBinding.Internal;
+using Microsoft.AspNetCore.Mvc.ModelBinding.Metadata;
 using Microsoft.Extensions.Internal;
 using Microsoft.Extensions.Options;
 
@@ -23,7 +23,7 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
         private readonly IModelMetadataProvider _metadataProvider;
         private readonly IModelBinderProvider[] _providers;
 
-        private readonly ConcurrentDictionary<object, IModelBinder> _cache;
+        private readonly ConcurrentDictionary<Key, IModelBinder> _cache;
 
         /// <summary>
         /// Creates a new <see cref="ModelBinderFactory"/>.
@@ -35,7 +35,7 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
             _metadataProvider = metadataProvider;
             _providers = options.Value.ModelBinderProviders.ToArray();
 
-            _cache = new ConcurrentDictionary<object, IModelBinder>(ReferenceEqualityComparer.Instance);
+            _cache = new ConcurrentDictionary<Key, IModelBinder>();
         }
 
         /// <inheritdoc />
@@ -46,31 +46,52 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
                 throw new ArgumentNullException(nameof(context));
             }
 
-            // We perform caching in CreateBinder (not in CreateBinderCore) because we only want to
-            // cache the top-level binder.
             IModelBinder binder;
-            if (context.CacheToken != null && _cache.TryGetValue(context.CacheToken, out binder))
+            if (TryGetCachedBinder(context.Metadata, context.CacheToken, out binder))
             {
                 return binder;
             }
-
+            
+            // Perf: We're calling the Uncached version of the API here so we can:
+            // 1. avoid allocating a context when the value is already cached
+            // 2. avoid checking the cache twice when the value is not cached
             var providerContext = new DefaultModelBinderProviderContext(this, context);
-            binder = CreateBinderCore(providerContext, context.CacheToken);
+            binder = CreateBinderCoreUncached(providerContext, context.CacheToken);
             if (binder == null)
             {
                 var message = Resources.FormatCouldNotCreateIModelBinder(providerContext.Metadata.ModelType);
                 throw new InvalidOperationException(message);
             }
 
-            if (context.CacheToken != null)
+            Debug.Assert(!(binder is PlaceholderBinder));
+            AddToCache(context.Metadata, context.CacheToken, binder);
+
+            return binder;
+        }
+
+        // Called by the DefaultModelBinderProviderContext when we're recursively creating a binder
+        // so that all intermediate results can be cached.
+        private IModelBinder CreateBinderCoreCached(DefaultModelBinderProviderContext providerContext, object token)
+        {
+            IModelBinder binder;
+            if (TryGetCachedBinder(providerContext.Metadata, token, out binder))
             {
-                _cache.TryAdd(context.CacheToken, binder);
+                return binder;
+            }
+
+            // We're definitely creating a binder for an non-root node here, so it's OK for binder creation
+            // to fail.
+            binder = CreateBinderCoreUncached(providerContext, token) ?? NoOpBinder.Instance;
+
+            if (!(binder is PlaceholderBinder))
+            {
+                AddToCache(providerContext.Metadata, token, binder);
             }
 
             return binder;
         }
 
-        private IModelBinder CreateBinderCore(DefaultModelBinderProviderContext providerContext, object token)
+        private IModelBinder CreateBinderCoreUncached(DefaultModelBinderProviderContext providerContext, object token)
         {
             if (!providerContext.Metadata.IsBindingAllowed)
             {
@@ -82,32 +103,34 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
             // happen looking at just model metadata.
             var key = new Key(providerContext.Metadata, token);
 
-            // If we're currently recursively building a binder for this type, just return
-            // a PlaceholderBinder. We'll fix it up later to point to the 'real' binder
-            // when the stack unwinds.
-            var stack = providerContext.Stack;
-            for (var i = 0; i < stack.Count; i++)
+            // The providerContext.Visited is used here to break cycles in recursion. We need a separate
+            // per-operation cache for cycle breaking because the global cache (_cache) needs to always stay
+            // in a valid state.
+            //
+            // We store null as a sentinel inside the providerContext.Visited to track the fact that we've visited
+            // a given node but haven't yet created a binder for it. We don't want to eagerly create a
+            // PlaceholderBinder because that would result in lots of unnecessary indirection and allocations.
+            var visited = providerContext.Visited;
+
+            IModelBinder binder;
+            if (visited.TryGetValue(key, out binder))
             {
-                var entry = stack[i];
-                if (key.Equals(entry.Key))
+                if (binder != null)
                 {
-                    if (entry.Value == null)
-                    {
-                        // Recursion detected, create a DelegatingBinder.
-                        var binder = new PlaceholderBinder();
-                        stack[i] = new KeyValuePair<Key, PlaceholderBinder>(entry.Key, binder);
-                        return binder;
-                    }
-                    else
-                    {
-                        return entry.Value;
-                    }
+                    return binder;
                 }
+
+                // If we're currently recursively building a binder for this type, just return
+                // a PlaceholderBinder. We'll fix it up later to point to the 'real' binder
+                // when the stack unwinds.
+                binder = new PlaceholderBinder();
+                visited[key] = binder;
+                return binder;
             }
 
-            // OK this isn't a recursive case (yet) so "push" an entry on the stack and then ask the providers
+            // OK this isn't a recursive case (yet) so add an entry and then ask the providers
             // to create the binder.
-            stack.Add(new KeyValuePair<Key, PlaceholderBinder>(key, null));
+            visited.Add(key, null);
 
             IModelBinder result = null;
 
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNetCore.Mvc.Core/Internal/PlaceholderBinder.cs
```diff
@@ -2,8 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
 
-namespace Microsoft.AspNetCore.Mvc.ModelBinding.Internal
+namespace Microsoft.AspNetCore.Mvc.Internal
```
rynowak: This was in the wrong namespace. Folder is right.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
```diff
@@ -47,31 +47,51 @@ public IModelBinder CreateBinder(ModelBinderFactoryContext context)
...
+        // so that all intermediate results can be cached.
+        private IModelBinder CreateBinderCoreCached(DefaultModelBinderProviderContext providerContext, object token)
+        {
+            IModelBinder binder;
+            if (TryGetCachedBinder(providerContext.Metadata, token, out binder))
+            {
+                return binder;
+            }
+
+            // We're definitely creating a binder for an interior node here, so it's OK for binder creation
```
rynowak: nit: this (and other places) should say "non-root" instead of "interior"

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
```diff
@@ -122,26 +138,47 @@ private IModelBinder CreateBinderCore(DefaultModelBinderProviderContext provider
                 }
             }
 
-            if (result == null && stack.Count > 1)
+            // If the DelegatingBinder was created, then it means we recursed. Hook it up to the 'real' binder.
```
rynowak: Should say "PlaceholderBinder"

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
```diff
@@ -122,26 +138,47 @@ private IModelBinder CreateBinderCore(DefaultModelBinderProviderContext provider
...
+            var delegatingBinder = collection[key] as PlaceholderBinder;
+            if (delegatingBinder != null)
             {
-                // Use a no-op binder if we're below the top level. At the top level, we throw.
-                result = NoOpBinder.Instance;
+                // It's also possible that user code called into `CreateBinder` but then returned null, we don't
+                // want to create something that will null-ref later so just hook this up to the no-op binder.
+                delegatingBinder.Inner = result ?? NoOpBinder.Instance;
```
rynowak: Wierd case. Now that we cache everything it would be possible for this to pollute the cache so we need to fix it. I wrote a test for it.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
```diff
@@ -122,26 +138,47 @@ private IModelBinder CreateBinderCore(DefaultModelBinderProviderContext provider
...
+            if (cacheToken == null)
+            {
+                return;
+            }
+
+            _cache.TryAdd(new Key(metadata, cacheToken), binder);
+        }
+
+        private bool TryGetCachedBinder(ModelMetadata metadata, object cacheToken, out IModelBinder binder)
+        {
```
rynowak: Will add assert for `metadata`

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
```diff
@@ -184,15 +221,33 @@ private class DefaultModelBinderProviderContext : ModelBinderProviderContext
             public override IModelMetadataProvider MetadataProvider { get; }
 
             // Not using a 'real' Stack<> because we want random access to modify the entries.
```
rynowak: Stale comment, will remove

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
dougbu: It isn't clear to me whether this PR is `[Design]` from this and similar comments.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
```diff
@@ -86,29 +106,25 @@ private IModelBinder CreateBinderCore(DefaultModelBinderProviderContext provider
...
-            var stack = providerContext.Stack;
-            for (var i = 0; i < stack.Count; i++)
+            var collection = providerContext.Collection;
+
+            IModelBinder binder;
+            if (collection.TryGetValue(key, out binder))
             {
-                var entry = stack[i];
-                if (key.Equals(entry.Key))
+                if (binder != null)
```
dougbu: Odd to have different handling for `null` entries in the cache and in this "stack". But is `binder == null` a valid case here? When exactly would a `null` value be found in the `providerContext.Collection`?

dougbu: Ah, I see. The second question should be "Why?" or "Can we use a different indicator for "currently looking for this binder?"

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
```diff
@@ -184,15 +221,33 @@ private class DefaultModelBinderProviderContext : ModelBinderProviderContext
...
         }
 
+        // This key allows you to specify a ModelMetadata which represents the type/property being bound
+        // and a 'token' which acts as an arbitrary discriminator.
+        //
+        // This is necessary because the same metadata might be bound as a top-level parameter (with BindingInfo on
+        // the ParameterDescriptor) or in a call to TryUpdateModel (no BindingInfo) or as a collection element.
+        //
+        // We need to be able to tell the difference between these things to avoid over-caching.
+        [DebuggerDisplay("{ToString(),nq}")]
```
dougbu: The debugger should fall back to the `ToString()` method w/o this attribute. Either remove this or rename the method and make it `private`.

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/ModelBinderFactoryTest.cs
```diff
@@ -235,6 +236,244 @@ public void CreateBinder_Caches_WhenTokenIsNotNull()
...
+
+            options.Value.ModelBinderProviders.Add(widgetProvider);
+            options.Value.ModelBinderProviders.Add(widgetIdProvider);
+
+            var factory = new ModelBinderFactory(metadataProvider, options);
+
+            var context = new ModelBinderFactoryContext()
+            {
+                Metadata = metadataProvider.GetMetadataForType(typeof(Widget)),
+                CacheToken = null, // We want the provider to run twice.
```
dougbu: Suggest "the outer binder provider" since this test involves three providers.

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/ModelBinderFactoryTest.cs
```diff
@@ -235,6 +236,244 @@ public void CreateBinder_Caches_WhenTokenIsNotNull()
...
+                    var binder = c.CreateBinder(c.Metadata.Properties[nameof(Widget.Id)]);
+                    Assert.IsType<NoOpBinder>(binder);
+                    if (inner == null)
+                    {
+                        inner = binder;
+                    }
+                    else
+                    {
+                        Assert.Same(inner, binder);
+                    }
```
dougbu: Don't other tests already confirm the `NoOpBinder` is a singleton?

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
```diff
@@ -122,26 +138,47 @@ private IModelBinder CreateBinderCore(DefaultModelBinderProviderContext provider
...
 
             return result;
         }
 
+        private void AddToCache(ModelMetadata metadata, object cacheToken, IModelBinder binder)
+        {
+            Debug.Assert(metadata != null);
+            Debug.Assert(binder != null);
+
+            if (cacheToken == null)
```
dougbu: Why support the `cacheToken == null` case at all? Not used for inner nodes and makes even less sense for top-level binders.

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/ModelBinderFactoryTest.cs
```diff
@@ -235,6 +236,244 @@ public void CreateBinder_Caches_WhenTokenIsNotNull()
...
+            Assert.Same(inner, result2);
+            Assert.Equal(1, widgetProvider.SuccessCount);
+            Assert.Equal(1, widgetIdProvider.SuccessCount);
+        }
+
+        // This is a really wierd case, but I wanted to make sure it's covered so it doesn't
+        // blow up in wierd ways.
+        //
+        // If a binder provider tries to recursively create itself, but then returns null, we've
+        // already returned and possibly cached the DelegatingBinder instance, we want to make sure that
```
dougbu: Placeholder, not delegating binder.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
dougbu: Also can `providerContext.Collection` be removed? That is, do we need more than the `_cache`?

dougbu:
:watch:

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
rynowak: Discussed in person, this is not a design PR

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
rynowak: Discussed in person, and agreed to improve the comments here.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
rynowak: Sure, removed

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/ModelBinderFactoryTest.cs
rynowak: This won't be an instance of `NoOpBinder`, it's the mock returned by the provider for `Widget.Id`

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
rynowak: Discussed in person and agreed to leave this for now. I think if we don't support the `cacheToken == null` case the alternatives become pretty surprising.

rynowak:
@dougbu - rebased and updated. Please take a look as this is approved for 1.0.0

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
dougbu: New comments look good.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
```diff
@@ -82,32 +102,34 @@ private IModelBinder CreateBinderCore(DefaultModelBinderProviderContext provider
...
-            var stack = providerContext.Stack;
-            for (var i = 0; i < stack.Count; i++)
+            // The providerContext.Visited is used here to break cycles in recursion. We need a separate
+            // per-operation cache for cycle breaking because the global cache (_cache) needs to always stay
+            // in a valid state.
+            //
+            // We store null as a sentinel inside the providerContext.Visited to track the fact that we've visited
+            // a given node but haven't yet created a binder for it. We don't want to eagerly create a
+            // PlaceholderBinder because that would result in lots of unnecessary indirection and allocations.
+            var collection = providerContext.Visited;
```
dougbu: Suggest `visited`

test/Microsoft.AspNetCore.Mvc.Core.Test/ModelBinding/ModelBinderFactoryTest.cs
dougbu: :ok:

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
```diff
@@ -121,26 +143,49 @@ private IModelBinder CreateBinderCore(DefaultModelBinderProviderContext provider
...
 
-            if (result == null && stack.Count > 1)
+            // If the PlaceholderBinder was created, then it means we recursed. Hook it up to the 'real' binder.
+            var placeholderBinder = collection[key] as PlaceholderBinder;
+            if (placeholderBinder != null)
             {
-                // Use a no-op binder if we're below the top level. At the top level, we throw.
-                result = NoOpBinder.Instance;
+                // It's also possible that user code called into `CreateBinder` but then returned null, we don't
+                // want to create something that will null-ref later so just hook this up to the no-op binder.
```
dougbu: Curious: Can't the `NoOpBinder` case be hit whenever a type w/o a binder is referenced recursively?

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
```diff
@@ -46,31 +46,51 @@ public IModelBinder CreateBinder(ModelBinderFactoryContext context)
...
-            binder = CreateBinderCore(providerContext, context.CacheToken);
+            binder = CreateBinderCoreUncached(providerContext, context.CacheToken);
             if (binder == null)
             {
                 var message = Resources.FormatCouldNotCreateIModelBinder(providerContext.Metadata.ModelType);
                 throw new InvalidOperationException(message);
             }
 
-            if (context.CacheToken != null)
+            AddToCache(context.Metadata, context.CacheToken, binder);
```
dougbu: If I'm reading this correctly, it's not possible for `binder` to be a `PlaceholderBinder` here. Might be worth an assertion. If I'm wrong, suggest adding `placeholder.Inner` (which _must_ not be `null`) to the cache instead.

dougbu:
:shipit:

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
rynowak: Sure

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
rynowak: Yes, but in that case `result != null`. The comment explains why `result` could be null.

src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/ModelBinderFactory.cs
rynowak: Sure, seems good to add an assert here

PR #4754, opened 2016-05-26
Modify test to look for assembly names rather than absolute paths

pranavkm:
This workaround an issue \ bug with the results returned by DependencyContext where
the path uses a mix of forward and back slashes which causes comparison with Assembly.Location that only
uses forward slashes to fail.


pranavkm:
cc @NTaylorMullen

NTaylorMullen:
:shipit:

PR #4755, opened 2016-05-26
Update Mvc.Core and dependencies to netstandard1.6

pranavkm:


pranavkm:
cc @ajaybhargavb \ @NTaylorMullen

javiercn:
Don't we need to also update test projects?

pranavkm:
@javiercn no, test projects target netcoreapp1.0 which is a super set.

NTaylorMullen:
:shipit:

ajaybhargavb:
:shipit:

javiercn:
I thought it was the oposite, 1.0 had less stuff than 1.1 and so on, isn't that the case? If I'm wrong just :shipit:

pranavkm:
I think netcoreapp1.0 targets netstandard1.6 (althought @davidfowl's sample says otherwise: https://github.com/davidfowl/NetStandard/blob/master/platforms.cs#L120)

NTaylorMullen:
> I think netcoreapp1.0 targets netstandard1.6 (althought @davidfowl's sample says otherwise: https://github.com/davidfowl/NetStandard/blob/master/platforms.cs#L120)

Didn't david create that before netstandard1.6 was a thing?

NTaylorMullen:
a

NTaylorMullen:
s

PR #4756, opened 2016-05-26
[Fixes #4506] Move and rename ActionDescriptor.Name to ControllerActionDescriptor.ActionName

### LINKED ISSUE #4506: Do we need Name and DisplayName on ActionDescriptor
It's not clear what the difference is between these. I think we probably want to remove `Name` and add `ActionName` to `ControllerActionDescriptor` - the more specialized type

javiercn:


javiercn:
@Eilon can you take a look at this one? @rynowak is not around and you seem to have an opinion on what needed to be done. I'll finish cleaning up the tests and the implementation once I know you are ok with the change.

test/Microsoft.AspNetCore.Mvc.ViewFeatures.Test/Internal/PartialViewResultExecutorTest.cs
```diff
@@ -53,28 +54,26 @@ public void FindView_UsesViewEngine_FromPartialViewResult()
...
-        public void FindView_UsesActionDescriptorName_IfViewNameIsNull()
-        {
-            // Arrange
-            var context = GetActionContext();
-            var executor = GetViewExecutor();
-
-            var viewName = "some-view-name";
-            context.ActionDescriptor.Name = viewName;
+        //[Fact]
+        //public void FindView_UsesActionDescriptorName_IfViewNameIsNull()
```
kichalla: ?

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Controller.cs
```diff
@@ -102,7 +102,7 @@ public dynamic ViewBag
         [NonAction]
         public virtual ViewResult View()
         {
-            return View(viewName: null);
+            return View(viewName: ControllerContext.ActionDescriptor.ActionName);
```
kichalla: Curious, the bug doesn't say about these changes..why are you making them? also one could return view result from a filter too in which case the earlier behavior of defaulting to the currently action name seems fine.

Eilon: We need to try to avoid doing this here. The whole idea is that the controller is simply manufacturing the results; it's not adding much flavor to them. Then it's up to the result's execution logic to determine that no view name means to use the normalized action name.

This current code will likely break customer unit tests either due to a NullRef in CC.AD.AN, or due to it producing a non-null value.

Eilon:
🕐

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Controller.cs
javiercn: @Eilon @kichalla Those were exactly my concerns, that's the reason I split the PR into two commits. If you look at the first commit alone, you'll see that by moving the ActionName down to ControllerActionDescriptor I'm forced to do a downcast on the result executor in order to access the action name. That looks bad to me as from a layering perspective ViewResult/ViewResultExecutor should know nothing about ControllerActionDescriptor as that's an implementation detail of controllers.

After talking to @pranavkm I've figured out that we can get the action name in the same way we get the controller name and area name (see https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngine.cs#L280).

We can move https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngine.cs#L210 to https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngine.cs#L282 and throw an exception at that point (or maybe even remove if the expanders handle it gracefully)

Both methods in  https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngine.cs#L233 can be easily changed to return false or we can just add a null check to the condition.

@Eilon  @kichalla @pranavkm Thoughts?

javiercn:
The approach I mentioned will work for all existing scenarios without change, I've done it on the ViewResultExecutor (just copy paste) to prove it works, but if you are ok with moving computing the viewName to razor view engine I think it's the best option

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/PartialViewResultExecutor.cs
```diff
@@ -158,5 +160,47 @@ public virtual Task ExecuteAsync(ActionContext actionContext, IView view, Partia
                 viewResult.ContentType,
                 viewResult.StatusCode);
         }
+
+        private static string GetActionName(ActionContext context)
```
kichalla: Not sure I like this change mainly because ActionName is something which is calculated only once by the application model builder and should be reused rather than be calculated for every request.
I see your point about wanting to avoid downcasting to ControllerActionDescriptor.

My 2 cents: So considering the above maybe we can just leave `ActionName` property on `ActionDescriptor` as it is and avoid making an changes?

javiercn: That argument could also be applied to controller and area, what's special about ActionName? I think it's more correct that the view engine looks for default views based on data from routing. The aim with this change is to stop treating ActionName specially.

src/Microsoft.AspNetCore.Mvc.ViewFeatures/Controller.cs
dougbu: @javiercn the current approach doesn't work end-to-end.

Your suggested approach _might_ work (if the action name is always correct in the route values, even when using attribute routing). But does the value of removing `ActionDescriptor.Name` justify that large a change?

kichalla:
:shipit:

scionwest:
FYI - This is a breaking change from RC2 to RTM without an Announcement to go along with it.

Eilon:
@scionwest we generally make announcements for only the most significant changes, or ones where the change is difficult to work around. Having said that, there's a discussion at https://github.com/aspnet/Home/issues/1571 where there are some good suggestions on ways that we can communicate these changes more clearly, so we're considering applying some of that in the future.

scionwest:
@Eilon I understand that for betas, previews and rc builds. When you tell the community there will be zero breaking changes, unless absolutely required, between RC2 and RTM, you should really announce them all. The community was caught off guard several times today by this, including Damian in ASP.net community standup.

Having said that, this was absolutely a minor fix on my end; it took all of 90 seconds to find this issue and change my code. Just wanted to pass along some feedback.

Eilon:
@scionwest understood, and your feedback is very helpful - we intend to do a better job communicating these types of changes in the future.

